Issue #173612, opened 2025-08-12
Add borderRadius parameter to Image widget to support rounded corners


houssemeddinefadhli81:
### Use case

Currently, to display images with rounded corners, Flutter developers need to wrap the Image widget with a ClipRRect or similar widget. This is a common use case but requires extra boilerplate and can lead to inconsistent UI code.

I propose adding a new optional borderRadius parameter directly to the Image widget and its named constructors (Image.asset, Image.network, Image.file, Image.memory). When this parameter is provided (non-null), the Image widget will clip its final rendered content with a ClipRRect using the given border radius.

### Proposal

Add an optional borderRadius parameter to the Flutter Image widget and its constructors (Image.asset, Image.network, etc.) to allow clipping images with rounded corners directly, without needing an extra ClipRRect wrapper.

```
Image.asset(
  'assets/avatar.png',
  width: 48,
  height: 48,
  borderRadius: BorderRadius.circular(24),
)
```

darshankawar:
@houssemeddinefadhli81 Thanks for the report and potential fix. 

> Flutter developers need to wrap the Image widget with a ClipRRect or similar widget. This is a common use case but requires extra boilerplate and can lead to inconsistent UI code.

As you mentioned correctly, using `ClipRRect` is the way to go to have rounded corners for Image widget but it is just one additional line of code that developers need to add, ie to wrap Image widget with `ClipRRect`. 
Also, there's already a community package https://pub.dev/packages/flutter_image_widget that offers the feature already, so can you check if that helps to serve your use case ?
Not all features can be included in the sdk and hence there are packages such like above that offers the feature which sdk can't so please check the same.

houssemeddinefadhli18:
Hi @darshankawar , Thanks for your feedback. Wrapping Image with ClipRRect works, but adding a borderRadius parameter to Image itself makes the code simpler and UI more consistent without extra packages. This is a small change but has a big impact. That’s why relying on native widgets is better than using packages in this case.

I’ve already made a PR for this: #173613 

Thanks again, @darshankawar !

dkwingsmt:
Whenever possible, composition is superior to properties. Composition allows for endless combinations of operations without exponential growth of APIs.

In terms of semantics, wrapping `Image` with `ClipRRect` is easy to understand, follows the convention that applies to other widgets, and allows many other variations such as `ClipRSuperellipse` or `CustomClipper` as well as all the other properties of these clipping widgets.

[...]

PR #173613, opened 2025-08-12
Add borderRadius parameter to Image widget for rounded corners

houssemeddinefadhli18:
## Description

This PR adds an optional borderRadius parameter to the Image widget and its named constructors (Image.asset, Image.network, Image.file, and Image.memory). When provided, the image is clipped directly in the render layer to a rounded rectangle with the specified radius, eliminating the need to wrap the widget in a separate ClipRRect.

This approach:
	•	Simplifies common UI patterns involving rounded images.
	•	Improves performance by keeping the widget tree shallower.
	•	Maintains API consistency with other core widgets that expose borderRadius.
	•	Keeps backward compatibility, as the parameter defaults to null.

## Issue fixed
Closes issue: #173612 

## Tests added
	•	Widget tests verifying that borderRadius correctly clips the image when set.
	•	Tests confirming no clipping occurs when borderRadius is null or BorderRadius.zero.

---

## Pre-launch Checklist

- [x] I signed the [CLA].


justinmc:
Not sure if this is worth adding versus wrapping Image in a ClipRRect yourself. What do you think @victorsanni ?

houssemeddinefadhli18:
@justinmc Thanks for your feedback! While ClipRRect achieves the same effect, adding borderRadius directly to Image reduces the need for an extra composited layer and widget in the tree. It allows the framework to handle clipping internally, which can be more efficient, especially in scrollable lists with many images. This keeps the widget tree shallower, improves readability, and aligns Image with other core widgets that expose borderRadius as a first-class property...

dkwingsmt:
> While ClipRRect achieves the same effect, adding borderRadius directly to Image reduces the need for an extra composited layer and widget in the tree. It allows the framework to handle clipping internally, which can be more efficient, especially in scrollable lists with many images. 

If I understand correctly this PR still implements the radius by wrapping with an extra `ClipRRect`. How does this PR improve the efficiency as you mentioned?

> This keeps the widget tree shallower, improves readability

I don't see this as a convincing merit. 

> aligns Image with other core widgets that expose borderRadius as a first-class property

Can you point out which widgets you're referring to? I would imagine that the widgets that we've added border radius properties to are those that can't be simply implemented with wrapping with an extra `ClipRRect`.

houssemeddinefadhli18:
@dkwingsmt Thanks for you reply! What about updating it to handle the clipping inside RenderImage by calling canvas.clipRRect() before paintImage(), which would draw the image with rounded corners in a single pass and remove the extra widget overhead?

dkwingsmt:
Let's move to https://github.com/flutter/flutter/issues/173612 for discussion.

houssemeddinefadhli18:
@dkwingsmt Thanks for the motivation and the detailed feedback! This PR no longer wraps Image in ClipRRect. Instead, clipping is done directly in RenderImage.paint using canvas.clipRRect, which keeps the widget tree shallower and improves rendering performance, especially in scrollable lists with many images. This approach also aligns Image with other core widgets like Container, DecoratedBox, and Card that expose borderRadius as a first-class property, making it easier and more consistent for developers to apply rounded corners without extra widgets.

dkwingsmt:
Closing this PR due to discussion under https://github.com/flutter/flutter/issues/173612

PR #173652, opened 2025-08-12
[web] Popping a nameless route should preserve the correct route name

### LINKED ISSUE #173356: The showDialog bug prevents Flutter from being applied to production-level web applications.
### Steps to reproduce

1. Jump from the home page to the detail page
2. Open the dialog on the detail page.
3. Click the back button that comes with the Edge Mobile browser.

### Expected results

The dialog will close, and the page will remain on the detail page, with the URL being "/detail"

### Actual results

The dialog will close, and the page will remain on the detail page, But the URL is /

### Code sample

part1

```dart
routes:  {
  '/': (context) => const HomePage(),
  '/detail': (context) => const DetailPage(),
},
```

part2

```dart
onTap: () {
  showDialog(
    context: context,
    builder:
        (ctx) => AlertDialog(
      title: const Text('Invalid Input'),
      content: const Text('Please enter all the fields'),
      actions: [
        TextButton(
          onPressed: () => Navigator.pop(ctx),
          child: const Text('Okay'),
        ),
      ],
    ),
  );
```
Do not click the "OK" button. Instead, click the back button built into the Edge browser (since you cannot prevent users from doing so). Clicking the "OK" button is normal. Only by clicking the back button provided by the browser can this bug be reproduced.

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

[√] Flutter (Channel stable, 3.32.8, on Microsoft Windows [版本 10.0.19045.6093], locale zh-CN)
[√] Windows Version (10 专业版 64 位, 22H2, 2009)
[!] Android toolchain - develop for Android devices (Android SDK version 35.0.1)
[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[√] Android Studio (version 2024.2)
[√] IntelliJ IDEA Ultimate Edition (version 2025.1)
[√] Connected device (4 available)
[√] Network resources

mdebbar:
Fixes https://github.com/flutter/flutter/issues/173356

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/navigation/history.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/navigation/history.dart
@@ -6,7 +6,6 @@ import 'package:meta/meta.dart';
 import 'package:ui/ui.dart' as ui;
 import 'package:ui/ui_web/src/ui_web.dart' as ui_web;
 
-import '../dom.dart';
 import '../platform_dispatcher.dart';
 import '../services/message_codec.dart';
 import '../services/message_codecs.dart';
@@ -258,18 +257,29 @@ class SingleEntryBrowserHistory extends BrowserHistory {
     _setupStrategy(strategy);
 
     final String path = currentPath;
-    if (!_isFlutterEntry(domWindow.history.state)) {
+    _currentRouteName = path;
+    if (!_isFlutterEntry(currentState)) {
       // An entry may not have come from Flutter, for example, when the user
       // refreshes the page. They land directly on the "flutter" entry, so
       // there's no need to set up the "origin" and "flutter" entries, we can
       // safely assume they are already set up.
       _setupOriginEntry(strategy);
-      _setupFlutterEntry(strategy, path: path);
+      _setupFlutterEntry(strategy);
     }
   }
 
   @override
   final ui_web.UrlStrategy? urlStrategy;
+
+  /// The route name of the current page.
+  ///
+  /// This is updated whenever the framework calls `setRouteName`. This is then
+  /// used when the user hits the back button to pop a nameless route, to restore
+  /// the route name from before the nameless route was pushed.
+  ///
+  /// This is also used to track the user-provided url when they change it
+  /// directly in the address bar.
+  String _currentRouteName = '/';
 
   static const MethodCall _popRouteMethodCall = MethodCall('popRoute');
   static const String _kFlutterTag = 'flutter';
@@ -303,11 +313,11 @@ class SingleEntryBrowserHistory extends BrowserHistory {
   @override
   void setRouteName(String? routeName, {Object? state, bool replace = false}) {
     if (urlStrategy != null) {
-      _setupFlutterEntry(urlStrategy!, replace: true, path: routeName);
+      _currentRouteName = routeName ?? currentPath;
+      _setupFlutterEntry(urlStrategy!, replace: true);
     }
   }
 
-  String? _userProvidedRouteName;
   @override
   void onPopState(Object? state) {
     if (_isOriginEntry(state)) {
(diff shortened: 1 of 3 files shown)


engine/src/flutter/lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -258,19 +257,30 @@ class SingleEntryBrowserHistory extends BrowserHistory {
     _setupStrategy(strategy);
 
     final String path = currentPath;
-    if (!_isFlutterEntry(domWindow.history.state)) {
+    _currentRouteName = path;
```
chunhtai:
```dart
_currentRouteName = currentPath
```

engine/src/flutter/lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -303,11 +313,11 @@ class SingleEntryBrowserHistory extends BrowserHistory {
   @override
   void setRouteName(String? routeName, {Object? state, bool replace = false}) {
     if (urlStrategy != null) {
-      _setupFlutterEntry(urlStrategy!, replace: true, path: routeName);
+      _currentRouteName = routeName ?? currentPath;
```
chunhtai, line 315: i think `routeName` will never be null given since `routeUpdated` method has been deprecated for ages.

engine/src/flutter/lib/web_ui/test/engine/history_test.dart
```diff
@@ -293,7 +295,7 @@ void testMain() {
       expect(spy.messages[0].methodName, 'pushRoute');
       expect(spy.messages[0].methodArguments, '/page3');
       spy.messages.clear();
-      // 2. The framework sends a `routePushed` platform message.
+      // 2. The framework sends a `routeUpdated` platform message.
```
chunhtai, line 298: not something need to be address in this pr, routeUpdated is not used by framework for a long time now

engine/src/flutter/lib/web_ui/test/engine/history_test.dart
```diff
@@ -341,6 +343,35 @@ void testMain() {
...
     });
+
+    test('popping a nameless route does not change url', () async {
+      final TestUrlStrategy strategy = TestUrlStrategy.fromEntry(
+        const TestHistoryEntry(null, null, '/home'),
+      );
+      await implicitView.debugInitializeHistory(strategy, useSingle: true);
+
+      // Go to a named route.
+      await routeUpdated('/named-route');
```
chunhtai, line 354: let's use routeInformationUpdated for accuracy

chunhtai (commented):
LGTM, just some minor comment

chunhtai (approved):

engine/src/flutter/lib/web_ui/lib/src/engine/navigation/history.dart
mdebbar: Do you think it's safe to remove the entire `routeUpdated` handler:
https://github.com/flutter/flutter/blob/1eb2f68da5a7d38e0c3bdef887d888a6594a6a9d/engine/src/flutter/lib/web_ui/lib/src/engine/window.dart#L628-L632

chunhtai: yes, I think so, it has not been used since we have introduced Router. I forgot to clean it up

mdebbar: Ok, I'll do this cleanup in a follow up PR.

Issue #173660, opened 2025-08-12
Add Linux tests to confirm GTK calls are on GTK thread


robert-ancell:
Add tests to the Linux embedder that check all mocked GTK calls are called from the GTK thread. This stops issues that might occur e.g. during rendering which can be done from another thread.

PR #173602, opened 2025-08-12
Fix GTK redraw call being called from non-GTK thread.

robert-ancell:
gtk_widget_queue_draw is not thread-safe, call it from an idle callback.

Fixes https://github.com/flutter/flutter/issues/173447

Issue #173447, opened 2025-08-08
Linux desktop app just stops updating its window after a while


Hixie:
### Steps to reproduce

Run the code below using `flutter run -d linux` and wait up to 15 minutes (usually a lot less, sometimes less than 10 seconds). All the code does is show a timer on the screen.

### Expected results

There should be a counter that increases forever.

### Actual results

The timer eventually just appears to hang.

In practice (as tested via debugger), the Dart part of the application is continuing to work fine. It's just the window that isn't updating.

### Code sample

```dart
import 'dart:async';

import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  final DateTime _start = DateTime.now();
  
  @override
  Widget build(BuildContext context) {
    scheduleMicrotask(() { setState(() { /* rebuild */ }); });
    return Center(
      child: Text(
        '${DateTime.now().millisecondsSinceEpoch - _start.millisecondsSinceEpoch}ms',
        textDirection: TextDirection.ltr,
      ),
    );
  }
}
```

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[!] Flutter (Channel [user-branch], 3.33.0-1.0.pre-1324, on Ubuntu 24.04.1 LTS 6.8.0-58-generic, locale C.UTF-8) [61ms]
    ! Flutter version 3.33.0-1.0.pre-1324 on channel [user-branch] at /home/ianh/dev/flutter
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions at https://flutter.dev/setup.
    ! Upstream repository unknown source is not a standard remote.
      Set environment variable "FLUTTER_GIT_URL" to unknown source to dismiss this error.
    • Framework revision 796c62bd76 (2 hours ago), 2025-08-07 15:42:16 -0700
    • Engine revision 9264b70f24
    • Dart version 3.10.0 (build 3.10.0-75.1.beta)
    • DevTools version 2.49.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[...]

PR #173667, opened 2025-08-13
[beta] Cherry pick fix GTK redraw call being called from non-GTK thread

robert-ancell:
Cherry pick of https://github.com/flutter/flutter/pull/173602

Impacted users: All Linux users of Flutter
Impact Description: Due to calling gtk_window_redraw on a Flutter thread a lock up may occur. The Flutter app will then become unresponsive.
Workaround: No workaround
Risk: Low - fix is to run the GTK call on the GTK thread which is what the correct behaviour should be.
Test coverage: Rendering covered by existing tests, use of thread not explicitly tested, but https://github.com/flutter/flutter/issues/173660 opened to add this in future.
Validation Steps: Run test program in https://github.com/flutter/flutter/issues/173447 which generates many frames and maximizes the chance of a lock up.

### DIFF
--- a/engine/src/flutter/shell/platform/linux/fl_view.cc
+++ b/engine/src/flutter/shell/platform/linux/fl_view.cc
@@ -95,11 +95,16 @@ G_DEFINE_TYPE_WITH_CODE(
         G_IMPLEMENT_INTERFACE(fl_plugin_registry_get_type(),
                               fl_view_plugin_registry_iface_init))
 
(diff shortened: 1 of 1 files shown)


matanlurey (approved):
LGTM, thanks!

loic-sharma (approved):
LGTM

camsim99:
The tree is green so not sure why tree status is red? Thinking about bypassing and landing this.

Issue #117973, opened 2023-01-04
Add Support for Android RISC-V


iapicca:
## Use case

Google recently announced the [android risc-v porting](https://youtu.be/70O_RmTWP58) (article [here](https://arstechnica.com/gadgets/2023/01/google-announces-official-android-support-for-risc-v/))
and `dart` already has [beta support](https://dart-review.googlesource.com/c/sdk/+/217289) for `risc-v`

## Proposal

Support `flutter` on `android-riscv`

## Related issues

- https://github.com/flutter/flutter/issues/99963
- https://github.com/dart-lang/sdk/issues/38587

maks:
I would love to see support RISC-V and starting with Android would probably be the best approach given Android's recently announced roadmap for support and that there is already Dart on Linux RISC-V (experimental) support.

I'd also like to chime in that even Android emulator support for RISC-V is still an unknown time away (but on roadmap) so it would also be great if there could be initial support for Flutter Linux (custom embedder not desktop) on RISC-V.

The main blocker with that though is unlike say a RPI4, the current cost-effective and widely available Allwinner D1 based boards [are just too slow for on board development as I found out](https://github.com/flutter/flutter/issues/99963#issuecomment-1336295347), so having remote development support for Flutter Linux I think would be essential to be able to do this in practical manner. Actually I think would be useful in many other embedded Flutter usecase, eg. for [the flutter-pi project]( https://github.com/ardera/flutter-pi) 
Which makes me wonder how people at Toyota do this? surely they have some sort of remote debugging facility for their embedded Flutter projects, though perhaps its proprietary?

iapicca:
@exaby73 triage me :)

enh-google:
> I'd also like to chime in that even Android emulator support for RISC-V is still an unknown time away (but on roadmap) so it would also be great if there could be initial support for Flutter Linux (custom embedder not desktop) on RISC-V.

you can build a minimal qemu environment --- https://github.com/google/android-riscv64/blob/main/README.md

as for speed, yeah, we're currently finding that qemu is the least worst choice (if you have a fast x86-64 machine to run it on!). we cross-compile from the x86-64 host and sync across. presumably you can do similar?

maks:
@enh-google Thanks for the pointer to Android's RISC-V doc, very useful! 👍🏻 
On:
>  we cross-compile from the x86-64 host and sync across

[...]

PR #173672, opened 2025-08-13
Engine build setup for Android RISCV64.

rmacnak-google:
Bug: https://github.com/flutter/flutter/issues/117973


jason-simmons (approved):
The tests in https://github.com/flutter/flutter/blob/master/engine/src/flutter/tools/engine_tool/test/run_target_test.dart using tables keyed on `TargetPlatform` need entries for `TargetPlatform.androidRiscv64`

Issue #6537, opened 2016-10-26
Some of our controls can't handle being built in 0x0 environments


Hixie:
We should create tests that put each UI control into a 0x0 container and make sure they don't crash.

See https://github.com/flutter/flutter/pull/6535.

Hixie:
See also https://github.com/flutter/flutter/issues/5259 which triggers this in release and debug --no-hot mode.

dishankjindal1:
Hey 👋, can you please describe What's this issue about? I am looking to contribute. Is this issue still pending or resolved?

ahmedsameha1:
Hi @Hixie, I am looking forward to handling this issue, so please assign it to me.

ahmedsameha1:
How should I handle this? A pull request for every UI widget?

Hixie:
@ahmedsameha1 That's probably the best way to do it, yeah!

ahmedsameha1:
Should I ignore the deprecated UI controls? For example, ButtonBar.
What about UI controls that are planned to be deprecated? For example, RawMaterialButton.

Hixie:
I would definitely defer them until last, at least!

ahmedsameha1:
While trying to handle this issue for AdaptiveTextSelectionToolbar, the following test fails:
```dart
  testWidgets('AdaptiveTextSelectionToolbar renders at zero area', (WidgetTester tester) async {
    await tester.pumpWidget(
      const MaterialApp(
        home: Center(
          child: SizedBox.shrink(
            child: Scaffold(
              body: AdaptiveTextSelectionToolbar(
                anchors: TextSelectionToolbarAnchors(primaryAnchor: Offset.zero),
                children: <Widget>[Text('X')],
              ),
            ),
          ),
        ),
      ),
    );
    final Finder xText = find.text('X');
    expect(tester.getSize(xText).isEmpty, isTrue);
  });
```
The failure output contains:
```
The following assertion was thrown during performLayout():
RenderPhysicalShape does not meet its constraints.
Constraints:
  BoxConstraints(w=0.0, h=0.0)
Size:
  Size(0.0, 20.0)
```
I need some guidance.

ahmedsameha1:
While trying to handle this issue for CalendarDatePicker, the following test fails:
```dart
  testWidgets('CalendarDatePicker renders at zero area', (WidgetTester tester) async {
    await tester.pumpWidget(
      MaterialApp(
        home: SizedBox.shrink(
          child: Material(
[...]

PR #173408, opened 2025-08-07
Make sure that CalendarDatePicker & YearPicker don't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the CalendarDatePicker & YearPicker UI controls.

PR #173677, opened 2025-08-13
Make sure that a DatePickerDialog doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the DatePickerDialog UI control.

### DIFF
--- a/packages/flutter/test/material/date_picker_test.dart
+++ b/packages/flutter/test/material/date_picker_test.dart
@@ -2750,6 +2750,16 @@ void main() {
       expect(lastDayText.data, equals('28'));
     });
   });
+
+  testWidgets('DatePickerDialog renders at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      MaterialApp(
+        home: SizedBox.shrink(
+          child: DatePickerDialog(firstDate: firstDate, lastDate: lastDate),
+        ),
+      ),
+    );
+  });
 }
 
 class _RestorableDatePickerDialogTestWidget extends StatefulWidget {


packages/flutter/test/material/date_picker_test.dart
```diff
@@ -2750,6 +2750,16 @@ void main() {
...
+
+  testWidgets('DatePickerDialog renders at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      MaterialApp(
+        home: SizedBox.shrink(
+          child: DatePickerDialog(firstDate: firstDate, lastDate: lastDate),
+        ),
+      ),
+    );
+  });
```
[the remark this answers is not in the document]
dkwingsmt: I agree. There are no `expect` statements.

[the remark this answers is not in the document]
dkwingsmt: For the record: https://github.com/flutter/flutter/pull/173408/files#r2263666483 explained that `expect(tester.takeException(), isNull);` is redundant.

dkwingsmt (approved):
LGTM

victorsanni (approved):
LGTM, thanks for the PR.

packages/flutter/test/material/date_picker_test.dart
```diff
@@ -2750,6 +2750,16 @@ void main() {
       expect(lastDayText.data, equals('28'));
     });
   });
+
+  testWidgets('DatePickerDialog renders at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      MaterialApp(
+        home: SizedBox.shrink(
```
LongCatIsLooong, line 2757: I suspect this test isn't actually testing an 0x0 dialog, if MaterialApp gives a tight BoxConstraints to the `SizedBox`. Could you double check the size of the `DatePickerDialog` after it's laid out and see if it's indeed 0x0?

ahmedsameha1: ```print(tester.getSize(find.byType(DatePickerDialog)));``` prints Size(800.0, 600.0)

Please check this: https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193

LongCatIsLooong: It's supposed to be 0x0 right?

ahmedsameha1: I'm not sure.

LongCatIsLooong: Could you add a `Scaffold` or a `Center` to the `DatePickerDialog`, and verifies in the test that the final size of the `DatePickerDialog` is indeed `Size.zero`?

ahmedsameha1: I think there is no need to do this, according to: https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193

LongCatIsLooong: > Make sure that a DatePickerDialog doesn't crash in 0x0 environment

The test you added is to verify that DatePickerDialog doesn't crash in 0x0 environment, I think you want to make sure its size is `Size.zero` right? Otherwise you're just verifying the wrong thing in the test.

ahmedsameha1: Sir, did you ever read that https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193? Currently, I am working on this issue according to that comment. If you see that my pull request doesn't follow that comment, your objections are welcome; otherwise, you could show your objections in the issue itself and not here.

dkwingsmt: We don't need to make sure the size is `Size.zero`. We just want to make sure that it doesn't crash. Verifying the size, or anything at all, is not needed.

dkwingsmt: No offense but I agree that @LongCatIsLooong has a point. It's not about verifying the final size of `DatePickerDialog` to be zero. It's that the current way the tests are written does not even apply a zero size constraint to the target at all, therefore nothing is being tested. The outer widget `MaterialApp` uses a _tight_ constraint `Size(800, 600)` to its child, meaning that the child will be of size 800x600 whatsoever even with `SizedBox.shrink`. In other words, the `SizedBox.shrink` here is a noop.

It's not technically necessary to add an `expect` that verifies the widget size to be zero, although adding one would make it extra safer. But we really need to add something between `MaterialApp` and the target widget, such as:
```dart
  testWidgets('DatePickerDialog renders at zero area', (WidgetTester tester) async {
    await tester.pumpWidget(
      MaterialApp(
        home: Center(
          child: SizedBox.shrink(
            child: DatePickerDialog(firstDate: firstDate, lastDate: lastDate),
          ),
        ),
      ),
    );
    // print(tester.getSize(find.byType(DatePickerDialog)));
  });
```
Only by adding the extra `Center` can we really create a 0x0 environment, which you can manually verify with that print - or automatically verify with an extra `expect`.

PR #173712, opened 2025-08-13
[ Widget Preview ] Move `--dtd-url` from a global flag to a `widget-preview start` option

bkonyi:
`--dtd-url` is only used by widget previews and has no current usage, so this is a safe change.


DanTup (approved):
lgtm, thanks!

PR #173742, opened 2025-08-13
Thread sub-builders for every engine-uploading builder

### LINKED ISSUE #173655: Release builders should thread builds
Similar to engine_v2  - the release builders should kick off builds and then prepare local checkouts before waiting for secondary builds to complete.

Image

matanlurey:
Closes https://github.com/flutter/flutter/issues/173655.

We should consider making these flags either the default (opt-out to `false`) or evergreen (no opt-out).


jtmcdole:
> We should consider making these flags either the default (opt-out to `false`) or evergreen (no opt-out).

Yep, the absence of flag cleanup was my fault.

jtmcdole (approved):

PR #173757, opened 2025-08-14
Return result of setting OpenGL contexts back to Flutter

robert-ancell:
Also check when using them internally that they have worked.


mattkae (approved):
Cool stuff

PR #173761, opened 2025-08-14
[Part 1] feat: updates text field maxLines to double

### LINKED ISSUE #35064: TextField maxLines should be a double, not an int.
We need to be able to define a fractional maximum number of lines for a TextField. When using an integer number of lines, and there is a lot of text, the user will only see a few of the lines and may think that there are no more.

On the contrary, if you define, say 5.5 lines, and there are in fact, say, 10 lines, the user immediately notices that there is more text to the bottom. In other words, half-lines are visual cues to the users that they are able to scroll to see the rest. 

It's a **small change** that **improves usability**. To be consistent, I would make minLines also double, although in terms of usability that one is probably not important.

This is NOT a breaking change, and preserves existing maxLines behavior. 

Note: I'm the author of https://github.com/flutter/flutter/issues/19105 which was fixed by https://github.com/flutter/flutter/pull/27205

gaganyadav80:
Part of #35064 

This PR aims to allow using double instead of int for the max/min line properties of Text Field which will include the following properties:

- maxLines
- minLines
- hintMaxLines
- errorMaxLines

I am aiming to raise separate PR these property changes and this is a part of them. I will keep updating description to include  links to other PR as well.

### DIFF
--- a/engine/src/flutter/lib/web_ui/skwasm/text/paragraph_style.cpp
+++ b/engine/src/flutter/lib/web_ui/skwasm/text/paragraph_style.cpp
@@ -38,7 +38,7 @@ SKWASM_EXPORT void paragraphStyle_setTextDirection(
 }
 
 SKWASM_EXPORT void paragraphStyle_setMaxLines(ParagraphStyle* style,
-                                              size_t maxLines) {
+                                              SkScalar maxLines) {
   style->skiaParagraphStyle.setMaxLines(maxLines);
 }
 
--- a/engine/src/flutter/lib/web_ui/skwasm/text/paragraph_style.cpp
+++ b/engine/src/flutter/lib/web_ui/skwasm/text/paragraph_style.cpp
@@ -38,7 +38,7 @@ SKWASM_EXPORT void paragraphStyle_setTextDirection(
 }
 
 SKWASM_EXPORT void paragraphStyle_setMaxLines(ParagraphStyle* style,
-                                              SkScalar maxLines) {
+                                              size_t maxLines) {
   style->skiaParagraphStyle.setMaxLines(maxLines);
 }
 
--- a/packages/flutter/lib/src/cupertino/text_field.dart
+++ b/packages/flutter/lib/src/cupertino/text_field.dart
@@ -637,7 +637,7 @@ class CupertinoTextField extends StatefulWidget {
   /// {@macro flutter.widgets.editableText.maxLines}
   ///  * [expands], which determines whether the field should fill the height of
   ///    its parent.
-  final double? maxLines;
+  final num? maxLines;
 
   /// {@macro flutter.widgets.editableText.minLines}
   ///  * [expands], which determines whether the field should fill the height of
@@ -945,7 +945,7 @@ class CupertinoTextField extends StatefulWidget {
     properties.add(
       DiagnosticsProperty<bool>('enableSuggestions', enableSuggestions, defaultValue: true),
     );
-    properties.add(DoubleProperty('maxLines', maxLines, defaultValue: 1));
+    properties.add(DiagnosticsProperty<num>('maxLines', maxLines, defaultValue: 1));
     properties.add(IntProperty('minLines', minLines, defaultValue: null));
     properties.add(DiagnosticsProperty<bool>('expands', expands, defaultValue: false));
     properties.add(IntProperty('maxLength', maxLength, defaultValue: null));
--- a/packages/flutter/lib/src/cupertino/text_field.dart
+++ b/packages/flutter/lib/src/cupertino/text_field.dart
@@ -268,7 +268,7 @@ class CupertinoTextField extends StatefulWidget {
     SmartDashesType? smartDashesType,
     SmartQuotesType? smartQuotesType,
     this.enableSuggestions = true,
-    this.maxLines = 1.0,
+    this.maxLines = 1,
     this.minLines,
     this.expands = false,
     this.maxLength,
@@ -315,7 +315,7 @@ class CupertinoTextField extends StatefulWidget {
            smartDashesType ?? (obscureText ? SmartDashesType.disabled : SmartDashesType.enabled),
        smartQuotesType =
            smartQuotesType ?? (obscureText ? SmartQuotesType.disabled : SmartQuotesType.enabled),
-       assert(maxLines == null || maxLines > 0.0),
+       assert(maxLines == null || maxLines > 0),
        assert(minLines == null || minLines > 0),
        assert(
          (maxLines == null) || (minLines == null) || (maxLines >= minLines),
@@ -325,17 +325,17 @@ class CupertinoTextField extends StatefulWidget {
          !expands || (maxLines == null && minLines == null),
          'minLines and maxLines must be null when expands is true.',
        ),
-       assert(!obscureText || maxLines == 1.0, 'Obscured fields cannot be multiline.'),
+       assert(!obscureText || maxLines == 1, 'Obscured fields cannot be multiline.'),
        assert(maxLength == null || maxLength > 0),
        // Assert the following instead of setting it directly to avoid surprising the user by silently changing the value they set.
        assert(
          !identical(textInputAction, TextInputAction.newline) ||
-             maxLines == 1.0 ||
+             maxLines == 1 ||
              !identical(keyboardType, TextInputType.text),
          'Use keyboardType TextInputType.multiline when using TextInputAction.newline on a multiline TextField.',
        ),
        keyboardType =
-           keyboardType ?? (maxLines == 1.0 ? TextInputType.text : TextInputType.multiline),
+           keyboardType ?? (maxLines == 1 ? TextInputType.text : TextInputType.multiline),
        enableInteractiveSelection = enableInteractiveSelection ?? (!readOnly || !obscureText);
 
   /// Creates a borderless iOS-style text field.
@@ -407,7 +407,7 @@ class CupertinoTextField extends StatefulWidget {
     SmartDashesType? smartDashesType,
     SmartQuotesType? smartQuotesType,
     this.enableSuggestions = true,
-    this.maxLines = 1.0,
+    this.maxLines = 1,
     this.minLines,
     this.expands = false,
     this.maxLength,
@@ -454,7 +454,7 @@ class CupertinoTextField extends StatefulWidget {
            smartDashesType ?? (obscureText ? SmartDashesType.disabled : SmartDashesType.enabled),
        smartQuotesType =
            smartQuotesType ?? (obscureText ? SmartQuotesType.disabled : SmartQuotesType.enabled),
-       assert(maxLines == null || maxLines > 0.0),
+       assert(maxLines == null || maxLines > 0),
        assert(minLines == null || minLines > 0),
        assert(
          (maxLines == null) || (minLines == null) || (maxLines >= minLines),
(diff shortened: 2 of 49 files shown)


packages/flutter/lib/src/cupertino/text_field.dart
```diff
@@ -1349,7 +1349,7 @@ class _CupertinoTextFieldState extends State<CupertinoTextField>
                       // text does not shrink vertically as you type (however in
                       // rare circumstances, the height may still change when
                       // there's no placeholder text).
-                      maxLines: hasText ? 1 : widget.maxLines,
+                      maxLines: (hasText ? 1 : widget.maxLines)?.ceil(),
```
gaganyadav80: This will be updated in a subsequent PR

packages/flutter/lib/src/material/selectable_text.dart
```diff
@@ -774,7 +774,7 @@ class _SelectableTextState extends State<SelectableText>
         autofocus: widget.autofocus,
         forceLine: false,
         minLines: widget.minLines,
-        maxLines: widget.maxLines ?? defaultTextStyle.maxLines,
+        maxLines: (widget.maxLines ?? defaultTextStyle.maxLines)?.toDouble(),
```
gaganyadav80: I have not updated the maxLines property for `Text/SelectableText/DefaultTextStyle` hence we need to convert them to double before passing to the `EditableText`

packages/flutter/lib/src/rendering/editable.dart
```diff
@@ -1931,10 +1931,10 @@ class RenderEditable extends RenderBox
   }
 
   double _preferredHeight(double width) {
-    final int? maxLines = this.maxLines;
-    final int? minLines = this.minLines ?? maxLines;
+    final double? maxLines = this.maxLines;
+    final int? minLines = this.minLines ?? maxLines?.floor();
```
gaganyadav80: Same here, `minLines` will be updated in other PR

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -8560,15 +8560,15 @@ void main() {
...
+      const double lines = 2;
       await tester.pumpWidget(
         MaterialApp(
           home: Align(
             alignment: Alignment.topLeft,
             child: SizedBox(
               width: 400,
               child: EditableText(
-                minLines: lines,
+                minLines: lines.floor(),
```
gaganyadav80: Same, will update in other PR

tirth-patel-nc:
@gaganyadav80 thanks for contributing. If you plan to raise the PRs in parts, you can use `part of #35064` instead of the `fixes` keyword, since fixes would close the issue.

loic-sharma:
FYI, you have some test failures in the CI checks.

<details>
<summary>Example error...</summary>

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8706578329954034593/+/u/run_test.dart_for_framework_tests_shard_and_subshard_libraries/stdout

```
/b/s/w/ir/x/w/flutter/packages/flutter/test/rendering/editable_intrinsics_test.dart: editable intrinsics [E]
  Expected: normalized value matches
            'RenderEditable#00000 NEEDS-LAYOUT NEEDS-PAINT NEEDS-COMPOSITING-BITS-UPDATE DETACHED\n'
              ' │ parentData: MISSING\n'
              ' │ constraints: MISSING\n'
              ' │ size: MISSING\n'
              ' │ cursorColor: null\n'
              ' │ showCursor: ValueNotifier<bool>#00000(false)\n'
              ' │ maxLines: 1\n'
              ' │ minLines: null\n'
              ' │ selectionColor: null\n'
              ' │ locale: ja_JP\n'
              ' │ selection: null\n'
              ' │ offset: _FixedViewportOffset#00000(offset: 0.0)\n'
              ' ╘═╦══ text ═══\n'
              '   ║ TextSpan:\n'
              '   ║   inherit: true\n'
              '   ║   size: 10.0\n'
              '   ║   height: 1.0x\n'
              '   ║   "12345"\n'
              '   ╚═══════════\n'
              ''
    Actual: 'RenderEditable#3da6c NEEDS-LAYOUT NEEDS-PAINT NEEDS-COMPOSITING-BITS-UPDATE DETACHED\n'
              ' │ parentData: MISSING\n'
              ' │ constraints: MISSING\n'
              ' │ size: MISSING\n'
              ' │ cursorColor: null\n'
              ' │ showCursor: ValueNotifier<bool>#6a6d0(false)\n'
              ' │ maxLines: 1.0\n'
              ' │ minLines: null\n'
              ' │ selectionColor: null\n'
              ' │ locale: ja_JP\n'
              ' │ selection: null\n'
              ' │ offset: _FixedViewportOffset#61187(offset: 0.0)\n'
              ' ╘═╦══ text ═══\n'
              '   ║ TextSpan:\n'
              '   ║   inherit: true\n'
              '   ║   size: 10.0\n'
              '   ║   height: 1.0x\n'
              '   ║   "12345"\n'
              '   ╚═══════════\n'
              ''
     Which: Lines 7 differed, expected: 
            ' │ maxLines: 1'
            but got
            ' │ maxLines: 1.0'
  
  package:matcher                                     expect
  package:flutter_test/src/widget_tester.dart 473:18  expect
  test/rendering/editable_intrinsics_test.dart 28:5   main.<fn>
```

</details>

packages/flutter/lib/src/cupertino/text_field.dart
loic-sharma: From my understanding, this feature doesn't work as expected until at least `Text` is updated. If so, I'd prefer we don't land this in a in-between state. How much bigger is this PR if you also update `Text`/`SelectableText`/`DefaultTextStyle` in this PR?

gaganyadav80: When I tried this code in the emulator it was working right though I am not sure about the exact working/rendering of `Text` widget. There was no overflow error even though `TextField` was using `1.5` as maxLines and `Text` had maxLines as `2` (due to the usage of `ceil`).

I tried to update the Text `maxLines` to double but it also required me to make changes to the `engine/src/flutter/lib/ui/text.dart`. I am not sure about the working or implementation of the engine. I can however just update the property to double of the method being used. Please correct me if I am wrong or if I need to update the same at other files too (i.e `engine/src/flutter/lib/web_ui/**`).

The usage of `int? maxLines` is in the method `_encodeParagraphStyle()` specifically at `LN 1936`. It would be a great help if you could guide me.

Updating the three classes you mentioned should not be a big change.

gaganyadav80:
Hello @loic-sharma @dkwingsmt , I wanted to ask if it is alright that I also update the **other** properties (maxLines, minLines, helperMaxLines, and hintMaxLines) to double in this PR.

Because I have already updated `TextField.maxLines` and `Text/SelectableText/DefaultTextStyle.maxLines` in this PR. The number of files will increase but most of these are one line changes. 

I am fine with raising separate PRs if that is preferred. Please let me know.

dkwingsmt:
I'd say yes, they should be updated in the same way. However I recommend we only work on one field in this PR to settle the details, and mirror the change to others after this PR.

gaganyadav80:
@dkwingsmt I updated the `Text`/`SelectableText`/`DefaultTextStyle` as @loic-sharma mentioned, along with the changes in the engine for `ui.ParagraphStyle`. 

The update is comprising 24 files change. Should I push in this PR or in a separate PR?

PS: I appreciate your patience if any of my questions seem overly basic. Please let me know if I should adjust my approach.

dkwingsmt:
Another question: I see like 80% of changes in this PR are changing int literals to doubles. Are they needed for compiling or just for aesthetic reasons? If it's the latter I would slightly prefer reverting these changes to make the PR cleaner and make subsequent PRs easier for you.

gaganyadav80:
Got it - I had assumed using double literals was a convention, but I realise now I was looking at some older code. I’ll go ahead and revert those changes.

dkwingsmt:
Hi!  @gaganyadav80 Are you able to address the comments above? I think it's pretty close to completion after that and I'm looking forward to it!

gaganyadav80:
Hello @dkwingsmt, Yes I will address the comments. I couldn't gather time due to my job. I will finish it at the end of week. I appreciate your patience.

dkwingsmt:
No worries, take your time. I just want to be sure it's not buried. (Just FYI if ever you don't feel comfortable working on it in the near term feel free to convert it to a draft.) :)

gaganyadav80:
Hi @dkwingsmt, I have updated the PR to remove the literal changes. But I have not pushed the changes on `Text`/`SelectableText`/`DefaultTextStyle` to update their `maxLines` to double as mentioned by @loic-sharma. Though I have them ready locally.

Let me know if I should push them in this same PR or you would like a new PR :)

dkwingsmt:
Thanks for your work! Let me take a look now.

dkwingsmt:
Given how small this PR is now, I think it no longer makes much sense to split the PR. Can you push the entire change? :)

gaganyadav80:
Yes, I will push the changes before EOD.

gaganyadav80:
Hi @dkwingsmt, I’ve pushed the changes, which also required updating `ParagraphStyle` in the Flutter engine. However, I must have done something wrong since all tests are failing.  

From the build logs, it looks like I’ll need to pass `maxLines` (now a double) separately through the `ParagraphBuilder`, similar to how `height` and `fontSize` are handled.  

Before I dig deeper into the code, I’d really appreciate your suggestions.

gaaclarke:
> From the build logs, it looks like I’ll need to pass `maxLines` (now a double) separately through the `ParagraphBuilder`, similar to how `height` and `fontSize` are handled.
> 
> Before I dig deeper into the code, I’d really appreciate your suggestions.

Yes, that's what you need to do.  ParagraphStyle stores an array of int32's that is most of its value encoded.  Max lines is no longer an int32 so it can not longer be stored in there and must be handled like the other double fields, like height.

engine/src/flutter/lib/web_ui/skwasm/text/paragraph_style.cpp
```diff
@@ -38,7 +38,7 @@ SKWASM_EXPORT void paragraphStyle_setTextDirection(
 }
 
 SKWASM_EXPORT void paragraphStyle_setMaxLines(ParagraphStyle* style,
-                                              size_t maxLines) {
+                                              SkScalar maxLines) {
   style->skiaParagraphStyle.setMaxLines(maxLines);
```
dkwingsmt, line 42: It seems that `setMaxLines` is a `size_t` ([link](https://skia.googlesource.com/skia/+/52130e5c3b34/modules/skparagraph/include/ParagraphStyle.h#103)). This call probably truncates the fraction parts. Should we to a `ceil` here instead?

dkwingsmt (commented):
I propose that we should instead keep engine-level `maxLines` as integers. The engine probably needs to render the full line anyway, and the framework works by clipping the render result. Some, if not all, backend renderers (such as Skia) only supports integral maxLines.

gaganyadav80:
@dkwingsmt Thanks for the suggestion. I’ll update the PR accordingly. To avoid future confusion about why `maxLines` is an integer in the engine but a double in the framework, would it make sense to add a comment in the framework code explaining this? I can include that if you think it would be helpful.

dkwingsmt:
Oh, sorry I thought I had replied to you (probably drafted but never sent out).

> To avoid future confusion about why maxLines is an integer in the engine but a double in the framework, would it make sense to add a comment in the framework code explaining this?

It can be useful, if you can think of a good place to explain it (maybe `Text`?). Not strongly necessary though, since it might be intuitive enough to realize that the engine has to draw full text lines but widgets can crop them by partial lines.

dkwingsmt:
Yeah, that's unfortunate. Can you try if it builds with `num`?

gaganyadav80:
Hello @dkwingsmt, hope you are doing well. I was just checking in if you got the time to review my last comment?

dkwingsmt:
I'm really sorry that your comment slipped through my review queue (I didn't check the "Reviewed" tab).

That's unfortunate indeed. In that case we might have to create a new parameter, for example with the name of `maxDisplayedLines`, that uses `double?`.
* `maxDisplayedLines` should be provided exclusively with `maxLines`, guarded by assertion.
* `get maxLines` should return `ceil(maxDisplayedLines)`.

The biggest downside I can think of is how many classes this will affect.

I thought about hard breaking, but it'll probably go bad considering how widespread these classes have been used for. 

Also cc @loic-sharma and @justinmc.

gaganyadav80:
@dkwingsmt @loic-sharma Thank you for the suggested approach. I will update the PR soon.

dkwingsmt:
I'm going to reluctantly close this pull request for now until we're confident enough about this change. I still agree that this is a feature we should support, and the existing choice of `int` parameters is unfortunate. However, given the  complexity (allowing packages to `implement` this class), it is indeed extremely difficult to make such a seemingly small change. We can't afford to break code for such a core feature either. As @loic-sharma pointed out, if you're interested in pushing for this change and would like to draft a design doc that evaluates a proposed solution, we're happy to look at it. If not, that's also okay. I sincerely appreciate the effort you've put into this PR. Thank you very much.

PR #173777, opened 2025-08-14
[flutter_tools] Use DWDS 25.0.1

srujzs:
DWDS 25.0.1 requires that a valid reloadedSourcesUri is passed and updated for both a hot restart and a hot reload. Therefore, the bootstrap scripts which use it as well as the code to write that file is updated. Note that this file is read in both DWDS and in the bootstrap script. Along with this, code is fixed to update modules and digests regardless of a full restart. Currently, it makes no difference as neither Flutter tools or DWDS makes use of the updated modules or digests with the new library bundle format, but it's updated for consistency.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/lib/src/web/bootstrap.dart
```diff
@@ -266,26 +269,27 @@ $_simpleLoaderScript
...
 
-              var el = document.getElementById(script.id);
+              var el = document.getElementById(module);
               if (el) el.remove();
               el = window.\$dartCreateScript();
               el.src = policy.createScriptURL(src);
               el.async = false;
               el.defer = true;
-              el.id = script.id;
+              el.id = module;
```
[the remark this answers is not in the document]
srujzs: Fair enough. Updated this code as well as any bootstrap code the DDC library bundle format uses to use `const` and `let` over `var`.

packages/flutter_tools/lib/src/web/bootstrap.dart
```diff
@@ -474,14 +478,14 @@ String generateDDCLibraryBundleMainModule({
 /* ENTRYPOINT_EXTENTION_MARKER */
 
 (function() {
-  let appName = "org-dartlang-app:/$entrypoint";
+  const appName = "org-dartlang-app:/$entrypoint";
```
jyameo, line 481: looks like some tests need to be updated to reflect these changes - ie. https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/general.shard/web/bootstrap_test.dart#L270

jyameo (approved):
LGTM! Thanks Srujan 🚀

packages/flutter_tools/lib/src/web/bootstrap.dart
srujzs: Done, thanks!

srujzs:
FYI - there are some hot restart tests that are failing that I'm debugging.

srujzs:
FYI again - I found a bug in the change I made to wait until scripts are parsed. I have a fix and am going to add tests to DWDS and then will publish 25.0.1 and use that.

Issue #173651, opened 2025-08-12
Reduce the usage of dependsOn in the flutter gradle code


reidbaker:
helpful blog post here

https://melix.github.io/blog/2021/10/gradle-quickie-dependson.html

No direct benefit to developers but the assumption that migrating our code to better align with gradles direction and the direction with AGP will be beneficial.

PR #173794, opened 2025-08-14
Migrate deeplink json creation to public AGP api

reidbaker:
Reviewers please pay special attention to the tests that were added and tests that were removed. If you see a set of functionality that is not covered and should be please say something. 

I read every test (and TBH also had to edit most of them) but I had gemini's agent mode help and I dont trust there is something I missed. 

Related to #173651

Newly added tests can be run from `packages/flutter_tools/gradle` with `./gradlew test --tests com.flutter.gradle.tasks.DeepLinkJsonFromManifestTaskTest` 

## Pre-launch Checklist

- [x] I signed the [CLA].


reidbaker:
@gemini-code-assist[bot] please review this PR

gmackall (approved):
LGMT. The gemini comments about switching to `any` vs `foreach` seem reasonable to me BTW.

I tested that the deeplink cookbook from https://docs.flutter.dev/cookbook/navigation/set-up-app-links still works properly. Have you tested any more involved examples? Just asking because this seems 
1. like a fragile area of the code
2. I don't know how well it is tested.

reidbaker:
> LGMT. The gemini comments about switching to `any` vs `foreach` seem reasonable to me BTW.
> 
> I tested that the deeplink cookbook from https://docs.flutter.dev/cookbook/navigation/set-up-app-links still works properly. Have you tested any more involved examples? Just asking because this seems
> 
> 1. like a fragile area of the code
> 2. I don't know how well it is tested.

I authored integration tests here https://github.com/flutter/flutter/blob/2265d94c6b0bbbf22193852e52e279e8c001d385/packages/flutter_tools/test/integration.shard/android_gradle_outputs_app_link_settings_test.dart#L4 during the intial conversion. Those tests use object based  xml that is then written to a string so we dont have to worry about authoring valid xml. 

These tests use strings because we are testing the file input and output and are intentionally not covering the full parsing. 

As far as the parsing code If I put up a follow up pr with those changes after this lands would that be ok with you? Right now the parsing code is a straight move after the xml file is read.

gmackall:
> > LGMT. The gemini comments about switching to `any` vs `foreach` seem reasonable to me BTW.
> > I tested that the deeplink cookbook from https://docs.flutter.dev/cookbook/navigation/set-up-app-links still works properly. Have you tested any more involved examples? Just asking because this seems
> > 
> > 1. like a fragile area of the code
> > 2. I don't know how well it is tested.
> 
> I authored integration tests here
> 
> https://github.com/flutter/flutter/blob/2265d94c6b0bbbf22193852e52e279e8c001d385/packages/flutter_tools/test/integration.shard/android_gradle_outputs_app_link_settings_test.dart#L4
> 
> during the intial conversion. Those tests use object based xml that is then written to a string so we dont have to worry about authoring valid xml.
> These tests use strings because we are testing the file input and output and are intentionally not covering the full parsing.
> 
> As far as the parsing code If I put up a follow up pr with those changes after this lands would that be ok with you? Right now the parsing code is a straight move after the xml file is read.

Yes I think a follow up pr makes sense, just surfacing them in case you weren't paying attention to the gemini comments 👍

PR #173799, opened 2025-08-14
Remove `luci_flags.parallel_download_builds` and friends

matanlurey:
These default to `true` and were removed in https://flutter-review.googlesource.com/c/recipes/+/68841.


jtmcdole (approved):
infrastructure bridge sunset outrun 2 lgtm_640_24

PR #173808, opened 2025-08-14
FlutterTextRange / FlutterTextPosition to swift

LongCatIsLooong:
extracted from #170823

## Pre-launch Checklist

- [ ] I signed the [CLA].


LongCatIsLooong:
No tests because there's no semantic changes / covered by existing tests and the tests to be added in https://github.com/flutter/flutter/pull/170823

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1936,7 +1866,7 @@ - (CGRect)firstRectForRange:(UITextRange*)range {
 }
 
 - (CGRect)caretRectForPosition:(UITextPosition*)position {
-  NSInteger index = ((FlutterTextPosition*)position).index;
+  NSInteger index = position.index;
   UITextStorageDirection affinity = ((FlutterTextPosition*)position).affinity;
```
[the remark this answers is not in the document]
LongCatIsLooong: The extensions don't introduce convenience getters for affinity because it's rarely needed unlike offset.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1820,12 +1754,8 @@ - (CGRect)localRectFromFrameworkTransform:(CGRect)incomingRect {
...
 // the range encompasses multiple lines.
 - (CGRect)firstRectForRange:(UITextRange*)range {
-  NSAssert([range.start isKindOfClass:[FlutterTextPosition class]],
-           @"Expected a FlutterTextPosition for range.start (got %@).", [range.start class]);
-  NSAssert([range.end isKindOfClass:[FlutterTextPosition class]],
-           @"Expected a FlutterTextPosition for range.end (got %@).", [range.end class]);
-  NSUInteger start = ((FlutterTextPosition*)range.start).index;
-  NSUInteger end = ((FlutterTextPosition*)range.end).index;
+  NSUInteger start = range.start.index;
+  NSUInteger end = range.end.index;
```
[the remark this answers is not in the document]
LongCatIsLooong: I don't know what are the benefits of those NSAsserts. If we're given a UITextPosition subclass that is not a FlutterTextPosition, that's a bug in the IME (or whatever text service the text field is interacting with). And the plugin will crash anyways because we're downcasting the to FlutterTextPosition.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
```diff
@@ -0,0 +1,113 @@
...
+  override var description: String {
+    "TextPosition(\(index))"
+  }
+}
+
+/// A UITextRange subclass that represents a text range within the text, in
+/// a Flutter text field.
+///
+/// Similar to the dart:ui TextRange class, this class represents a right open
+/// interval consists of two `TextPosition`s.
```
loic-sharma: The consists seems a bit odd here. Consider either switching to `that consists` or:

```suggestion
/// interval of two `TextPosition`s.
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
```diff
@@ -0,0 +1,113 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import UIKit
+
+/// A UITextPosition subclass that represents a location in the text, in
+/// a Flutter text field.
+@objc(FlutterTextPosition)
+final class TextPosition: UITextPosition, NSCopying {
+  // offset is exposed as `range` in an extension on UITextPosition.
```
loic-sharma: This took a second for me to click that this comment was explaining why this is `fileprivate`. Consider expanding:

```suggestion
  // fileprivate as offset is exposed as `range` in an extension on UITextPosition.
```

I don't feel strongly, feel free to skip. Same comment for other fields like `nsRange` below.

loic-sharma (dismissed):
LGTM, but please get another review from someone that has more experience in the iOS embedder.

Thanks for splitting into smaller PRs btw! :)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
hellohuanlin: > If we're given a UITextPosition subclass that is not a FlutterTextPosition, that's a bug in the IME (or whatever text service the text field is interacting with)

Or more likely, a bug in our code that provided such text position. 

> And the plugin will crash anyways because we're downcasting the to FlutterTextPosition.

One of the benefit of assert is that it provides a good crash log

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
hellohuanlin: what does "offset being exposed as `range`" mean? offset is a number, but range is 2 numbers.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
```diff
@@ -0,0 +1,113 @@
...
+  init(index: UInt, affinity: UITextStorageDirection = .forward) {
+    self.offset = index
+    self.affinity = affinity
+  }
+
+  override func isEqual(_ object: Any?) -> Bool {
+    guard let other = object as? TextPosition else {
+      return false
+    }
+    // Affinity is only visual.
```
hellohuanlin: This comment doesn't explain how something "only visual" is equivalent to "don't consider it in isEqual". I wouldn't be surprised if apple writes code like this: 

```
updateVisualIfNeeded(position: Position) {
  if position == self.position { return }
  // update visual
}
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  override var description: String {
+    "TextRange(\(range))"
+  }
+}
+
+// MARK: Class methods for initializing in Objective-C
+
+@available(swift, obsoleted: 1.0)
```
hellohuanlin, line 72: why do we need this

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
LongCatIsLooong: > a bug in our code that provided such text position.

IIRC `FlutterTextPosition` is the only UITextPosition with an accessible initializer in the codebase (same with text ranges), so I don't think that can happen today.  

I changed the `UITextInput` implementation so all covariant `UITextPosition`/`UITextRange` occurrences in the interface are now replaced with `FlutterTextPosition` / `FlutterTextRange` to make sure the compiler catches this.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
LongCatIsLooong: removed the override since it wasn't in the objc version.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
LongCatIsLooong: This is extension is for objective-c only. If you're calling from swift you can just use the initializer.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1366,17 +1303,17 @@ - (void)selectAll:(id)sender {
...
 
 // Change the range of selected text, without notifying the framework.
 - (void)setSelectedTextRangeLocal:(UITextRange*)selectedTextRange {
   if (_selectedTextRange != selectedTextRange) {
     if (self.hasText) {
-      FlutterTextRange* flutterTextRange = (FlutterTextRange*)selectedTextRange;
       _selectedTextRange = [[FlutterTextRange
-          rangeWithNSRange:fml::RangeForCharactersInRange(self.text, flutterTextRange.range)] copy];
+          rangeWithNSRange:fml::RangeForCharactersInRange(self.text, selectedTextRange.range)]
+          copy];
```
hellohuanlin, line 1316: this copy shouldn't be necessary.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1636,32 +1570,32 @@ - (UITextPosition*)positionFromPosition:(UITextPosition*)position offset:(NSInte
...
-                                 offset:(NSInteger)offset {
+- (FlutterTextPosition*)positionFromPosition:(UITextPosition*)position
+                                 inDirection:(UITextLayoutDirection)direction
+                                      offset:(NSInteger)offset {
   // TODO(cbracken) Add RTL handling.
   switch (direction) {
     case UITextLayoutDirectionLeft:
     case UITextLayoutDirectionUp:
-      return [self positionFromPosition:position offset:offset * -1];
+      return (FlutterTextPosition*)[self positionFromPosition:position offset:offset * -1];
```
hellohuanlin, line 1580: isn't positionFromPosition returning `FlutterTextPosition` already as part of your chnage?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1636,32 +1570,32 @@ - (UITextPosition*)positionFromPosition:(UITextPosition*)position offset:(NSInte
...
     case UITextLayoutDirectionRight:
     case UITextLayoutDirectionDown:
-      return [self positionFromPosition:position offset:1];
+      return (FlutterTextPosition*)[self positionFromPosition:position offset:1];
```
hellohuanlin, line 1583: same here

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
hellohuanlin: i'd prefer a consistent api for index and affinity.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1997,20 +1927,19 @@ - (CGRect)caretRectForPosition:(UITextPosition*)position {
...
-        ((FlutterTextPosition*)_selectedTextRange.start).affinity;
+    NSUInteger currentIndex = _selectedTextRange.start.index;
+    UITextStorageDirection currentAffinity = _selectedTextRange.start.affinity;
     return [FlutterTextPosition positionWithIndex:currentIndex affinity:currentAffinity];
   }
 
   FlutterTextRange* range = [FlutterTextRange
       rangeWithNSRange:fml::RangeForCharactersInRange(self.text, NSMakeRange(0, self.text.length))];
-  return [self closestPositionToPoint:point withinRange:range];
+  return (FlutterTextPosition*)[self closestPositionToPoint:point withinRange:range];
```
hellohuanlin, line 1942: same here

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TextRange.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// MARK: Convenience extensions on UITextRange and UITextPosition
+
+@objc extension UITextRange {
+  /// The UTF16 range this TextRange represents.
+  ///
+  /// This computed property throws if the receiver is not a TextRange.
+  /// This is a left closed right open interval.
+  var range: NSRange {
```
hellohuanlin, line 98: I have 2 concerns of having range as part of UITextRange extension. 

1. What if Apple introduces a field called `range` in UITextRange itself in future SDK? 

2. The caller has context around what type of UITextRange it is, so the decision belongs to the caller, not the callee. In this particular example the text input plugin is the only customer, but the extension visible for the whole module, so we may be surprised in the future, who knows? Rather than making the assumption for all UITextRange here, I'd just do a cast on the caller site.

LongCatIsLooong: > 
> 1. What if Apple introduces a field called `range` in UITextRange itself in future SDK? 

That's unlikely to happen. The whole point of having these two classes is that UiKit doesn't make assumptions about the document structure. The class is supposed to have very limited set of operations defined. Adding a range to the protocol will break a lot of use cases.

> 2. The caller has context around what type of UITextRange it is, so the decision belongs to the caller, not the callee. In this particular example the text input plugin is the only customer, but the extension visible for the whole module, so we may be surprised in the future, who knows? Rather than making the assumption for all UITextRange here, I'd just do a cast on the caller site. 

Yeah the big assumption is all objective-C UITextInput implementations in the iOS embedder use these two classes . I think that's a safe assumption because these two classes are modelled after the framework editable text document index/range model so I don't see why different subclasses are needed. For swift UITextInput implementations I think you should not import these classes or the extensions.

LongCatIsLooong: Oh I think I mixed up swift access control with dart access control. But still I don't see the need to introduce a UITextInput implementation that doesn't use FlutterTextPosition.

hellohuanlin: > Adding a range to the protocol will break a lot of use cases.

Why will adding a range break a lot of use cases? 

> But still I don't see the need to introduce a UITextInput implementation that doesn't use FlutterTextPosition.

I didn't meant introducing a UITextInput implementation. We have UITextField too in our codebase, and with this code, you are adding the functionality to UITextField's positions, which isn't what you intend to do if I'm reading it right. Well will that break? Not that I can think of right now, but we may be surprised in the future. 

What I am suggesting is to do the cast on caller site (like what the original code is doing), because that's the place where we have enough context to make the decision about the types.

LongCatIsLooong: > Why will adding a range break a lot of use cases?

Adding a new method to an abstract interface is always a breaking change. Also there can be `UITextPosition` classes that use "structured" representations (e.g., "chapter 3 section 2 paragraph 1 character 100") and for these implementations there may not be an easy way to compute a `NSRange` from 2 positions. 

> We have UITextField too in our codebase, and with this code, you are adding the functionality to UITextField's positions

Ah that's true. But the UITextInput interface is supposed to be used for the communication between system services (notably the IME) and editors. We as UITextField users aren't supposed to call those APIs. So you're not supposed to try to get a UITextPosition from a UITextField in the first place. Tests are exceptions but they're tests and they're ok / supposed to fail if the test author accidentally calls the wrong methods.

hellohuanlin: > Adding a new method to an abstract interface is always a breaking change.

Gotcha. I missed the part that UITextPosition itself doesn't contain anything (I mistakenly thought it contains position and affinity)

> We as UITextField users aren't supposed to call those APIs. 

There's a difference between "aren't supposed to call" vs "will not call". The best way to avoid calling a method of a type (that you don't want people to call) is simply not to add those methods to that type. 

> So you're not supposed to try to get a UITextPosition from a UITextField in the first place. 

Apple API supports getting a UITextPosition from a UITextField. We don't use it now doesn't mean that we won't use it in the future.

hellohuanlin: Ok let me explain it with an example why this is a bad pattern. 

What you are currently doing is basically: 

```
class Animal {}
class Bird: Animal {
  func fly() {}
}

extension Animal {
  func fly() {
    guard self is Bird else {
      // throw exception
    } 
    (self as! Bird).fly()
  }
}

class YourPlugin {
  func foo(animal: Animal) {
    animal.fly()
  }
}
```

This is clearly an anti-pattern: you are adding `fly()` to all `Animal`s. 

Then the follow-up argument is basically saying "we only have `Bird`, so we are safe". Well, we are safe until someone introduce another `Animal` and surprises us. Will it happen? I don't know. But why do we have to introduce `fly()` to `Animal` in the first place? 

What I was suggesting is follow the existing convention: (1) remove the extension, and (2) do the cast on the caller site. The caller (not the callee) is the one who knows that the `Animal` is a `Bird`: 

```
class YourPlugin {
  func foo(animal: Animal) {
    // The caller (not the callee) is the one who knows that the `Animal` is a `Bird`
    let bird = animal as! Bird
    bird.fly()
  }
}
```

Let me know if it makes sense.

LongCatIsLooong: Sorry for the late reply.

> Apple API supports getting a UITextPosition from a UITextField

I looked up the UITextField API, it doesn't have a selectedRange field like UITextView does (https://developer.apple.com/documentation/uikit/uitextview/selectedrange?language=objc), so you're actually supposed to use `UITextRange` and UITextInput method to figure out the selected range for a UITextField. But in the iOS embedder we rarely create UITextFields/ UITextViews (they are only used for accessing system APIs that are otherwise unavilable, e.g., https://github.com/flutter/flutter/blob/d328555151ca0a2e5455c3cc9900b43dc4867d54/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm#L432-L434, so I'm not too worried about misusing the extension on those text field instances. Even if they are misused, they'll crash in tests as long as it's tested.

> The caller (not the callee) is the one who knows that the Animal is a Bird:

I think it's a matter of what's the scope that you can make the assumption that a UITextRange is a FlutterTextRange.
In the case of UITextInput, I think the entire implementation can make that assumption, imo it would be safe to define the same extension in TextInputPlugin.mm. I can move the extension to that file (Objective C is already very verbose, I don't really want downcast it every time I reference it), but it would be nice to be able to use that in tests and alternative UITextInput implementations too. FlutterTextRange is the only UITextRange subclass used in the ios embedder today and I don't think that will change in the near future.

hellohuanlin: > Then the follow-up argument is basically saying "we only have Bird, so we are safe". Well, we are safe until someone introduce another Animal and surprises us. Will it happen? I don't know. But why do we have to introduce fly() to Animal in the first place?

I think your new comment is still saying "we only have Bird, so we are safe", so my respond is the same (unless I miss any new information).

LongCatIsLooong: > But why do we have to introduce fly() to Animal in the first place?

I don't think there's too big of a problem with that in our case. All `Animal`s are `Bird`s as far as the iOS embedder is concerned. It's unfortunate that Objective-C doesn't have associated types so the UITextInput protocol can only give you Animals instead of Birds and you have to fight the type system to downcast them to Bird and I think that's where the problem came from in the first place.

hellohuanlin: Unless i misread it, it seems your argument is once again still "we only have Bird, so we are safe", so my respond is the same: 

> Then the follow-up argument is basically saying "we only have Bird, so we are safe". Well, we are safe until someone introduce another Animal and surprises us. Will it happen? I don't know. But why do we have to introduce fly() to Animal in the first place?

hellohuanlin: I figured this may take even more back and forth, so I might as well just write up some additional notes. From top of my head, the current design: 

1. It's a misleading contract that violates the LSP principle. It creates a fragile API that's likely to cause bugs as the code grows. 

2. It subverts type safety. It's essentially hiding a type requirement. 

3. It's not sustainable and violates open/closed principle (open for extension, meaning you can add new features, while closed for modification, meaning you shouldn't have to go back and change exiting, working code to add those new features)

Your past replies above have been focusing on "all animals are birds *as of now*". The thing about using a good design pattern that they encourage you to build for the future, not just for the immediate state of your code. Choosing a design that's tightly coupled to the current state is a classic trap. While it might be true that all your `Animal`s happens to be Birds *today*, a robust system is designed with the understanding that tomorrow you might need to add a `Fish` or a `Dog`, for whatever unforeseeable reason that we do not anticipate today.

LongCatIsLooong: Yeah my point is still the same, I don't think this is going to happen in the foreseeable future. I would not be too worried about something that there's no good reason to add yet, I think trying to make a choice that is 100% future proof is usually unrealistic. If it ever happens, a proper test should catch this, then we decide whether to keep the extension or not.

Re:type safety, unfortunately in this case type safety is already subverted (and that's why we have to downcast in the first place) in the UITextInput protocol design which is limited by the very old language it's defined in. I can write a wrapper protocol for UITextInput in Swift that's more type safe but I don't think that's worth the effort because it's such a large protocol.

hellohuanlin: We can never achieve 100% future proof by choosing a good design pattern, however, that doesn't mean we should go with the other end of spectrum and purposely choose a bad design pattern that violates numerous design principles.

hellohuanlin: I think I have said everything I have in mind and I feel we are talking in circles. Feel free to chime in since you also reviewed this PR @loic-sharma

hellohuanlin: /gemini: is this a good design pattern? 

```
// MARK: Convenience extensions on UITextRange and UITextPosition

@objc extension UITextRange {
  /// The UTF16 range this TextRange represents.
  ///
  /// This computed property throws if the receiver is not a TextRange.
  /// This is a left closed right open interval.
  var range: NSRange {
    assert(self is TextRange, "Unexpected UITextRange type: \(type(of: self))")
    return (self as! TextRange).nsRange
  }
}

@objc extension UITextPosition {
  /// The offset from the start of the document to this TextPosition, in UTF16 code units.
  ///
  /// This computed property throws if the receiver is not a TextPosition.
  var index: UInt {
    assert(self is TextPosition, "Unexpected UITextPosition type: \(type(of: self))")
    return (self as! TextPosition).offset
  }
}
```

hellohuanlin: Gemini did a much better job explaining it:

s1

s2

s3

hellohuanlin: My take on the type safety part: 

The point is that you can't accidentally call a method that doesn't belong to a type. For example, 

```
animal.fly()
```

This won't compile. 

In the place where you are sure the animal is a `Bird` type, you have to **explicitly** downcast it: 

```
(animal as! Bird).fly()
```

This is telling compiler "I am sure the type is correct during runtime" and it is by design. 

However, if you add the `fly()` method to `Animal` type, you are removing this explicit casting requirement from the caller site:  

```
animal.fly()
```

I agree this is a convenient shortcut (I do the same for smaller personal projects), but for a larger project with multiple contributors, it only opens to potential problems for the future. Also this is a misleading API because `fly()` doesn't belong to `Animal` type.

LongCatIsLooong: I would not consider this a "design" and I think the scope in which the extension is defined matters: yes it's going to be a problem if this was exposed as a public extension, because we can't make the assumption about the exact type of a UITextRange outside of the scope of the iOS embedder, but in the iOS embedder I think it's a safe assumption (a proper test should cover the call path that misuses it, and it's very difficult to misuse in the first place). 

 > it only opens to potential problems for the future

Could you elaborate on what potential problems this could cause? As I mentioned I think this is pretty safe (unless we introduce a different UITextInput implementation that doesn't use FlutterTextRange in the future, which I don't see happening since it's modeled after the framework TextSelection. And if that ever happens we should move the extension to TextInputPlugin.mm so it's only accessible in the scope where the assumption still holds true.)

hellohuanlin: > I would not consider this a "design"

It is a classic OOP design issue, and the current design violates multiple design principles as I explained above (and explained by gemini)

> but in the iOS embedder I think it's a safe assumption

It is not a safe when someone adds a new implementation in the future. Your argument of "I don't expect this to happen" isn't a compelling argument - this argument could be used to justify abandoning *any* design principles, rendering them all moot. These principles exist precisely because the future is unpredictable and will bring surprises. 

In fact, even for today, we already use Apple's `UITextField` in a few places, which have its own ranges and positions. So you can't really say "`FlutterTextRange/Position` is the only `UITextRange/Position` in our embedder" (or "all Animals are Birds") even for today. If someone writes `let textPosition = textField.beginningOfDocument`, your UITextPosition extension would add methods to this `textPostion` too. 

> a proper test should cover the call path that misuses it

Writing tests does not justify choosing a bad design pattern.

> and it's very difficult to misuse in the first place

Quite the opposite - by adding `fly()` method to `Animal` type, you made it *easy* (rather than *difficult*) to misuse it. To make it *difficult* to misuse, you should add `fly()` method to `Bird` type. 

> And if that ever happens we should move the extension to TextInputPlugin.mm so it's only accessible in the scope where the assumption still holds true.

This violates Open/Closed principle (as I explained above, and also by gemini). Also note that when it happens, it may not be you and me working on Text Input plugin anymore. For any codebase that's expected to be maintained, scaled, or worked on by other people, it's a risky pattern, as whoever inherit this codebase may not be aware of your plan as they are not part of this conversation.

I agree that your extension brings some "convenience" as it removes the need to cast at the caller site. However, I think we should view this "inconvenience" as a "benefit", rather than a "drawback". As explained in my previous comment, this "inconvenience" is a type safety measure, and it's a good thing.

I also think that I have spent wayyyyy too much time on this, and we are still going in circles by repeating the same talking points, and I don't believe we are heading anywhere. I will hand it off to @vashworth and @jmagman and unassign myself.

hellohuanlin: Just talked to @jmagman  and @vashworth, and @vashworth will take over the PR review (thank you). 

FWIW, one last word before I go:

Basing a design on predictions or assumptions is an inherently fragile strategy, as the future may surprise you. A more robust, and ultimately simpler approach is to ground the design in established patterns and principles - they are not 100% future proof, but they are battle tested  and have proven their ability to handle a wide variety of unforeseen requirements, making them a much safer bet.

chinmaygarde:
May I mark this as a draft while progress is being made?

Piinks:
Welcome back to stale PR triage! Should we close this for now?

PR #172074, opened 2025-07-13
Make sure that a MaterialBanner doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle #6537 for the MaterialBanner UI control.

PR #173827, opened 2025-08-15
Make sure that a DesktopTextSelectionToolbarButton doesn't crash in 0…

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the DesktopTextSelectionToolbarButton UI control.


LongCatIsLooong:
Could you also do this https://github.com/flutter/flutter/pull/173677#discussion_r2279834884, in this PR?

ahmedsameha1:
I think there is no need to do this, according to: https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193

justinmc (approved):
LGTM 👍 I agree no takeException is needed when it's at the end of a test like this. I typically use it when it would be beneficial for the test to fail before the end of the test if there is an exception.

justinmc:
I guess you could expect that the size of DesktopTextSelectionToolbarButton is indeed 0x0?

ahmedsameha1:
> I guess you could expect that the size of DesktopTextSelectionToolbarButton is indeed 0x0?

Do you want me to add this expectation?

justinmc (changes requested):
Yes please add that expectation. I just talked with @LongCatIsLooong about this and we agree.

ahmedsameha1:
@justinmc Could you resolve this in https://github.com/flutter/flutter/issues/6537 because, according to https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193, there is no need to add that expectation. I think that this expectation is not specific to this widget, or is it?
Please consider this: https://github.com/flutter/flutter/pull/172074#issuecomment-3106174908

ahmedsameha1:
I have updated this pull request according to https://github.com/flutter/flutter/issues/6537#issuecomment-3216557050.

dkwingsmt (approved):
LGTM, thank you!

packages/flutter/test/material/desktop_text_selection_toolbar_button_test.dart
```diff
@@ -42,4 +42,19 @@ void main() {
...
+    await tester.pumpWidget(
+      const MaterialApp(
+        home: Center(
+          child: SizedBox.shrink(
+            child: DesktopTextSelectionToolbarButton(onPressed: null, child: Text('X')),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(DesktopTextSelectionToolbarButton)).isEmpty, isTrue);
```
LongCatIsLooong:
```suggestion
    expect(tester.getSize(find.byType(DesktopTextSelectionToolbarButton)), Size.empty);
```
`Size.isEmpty` is a necessary but not sufficient condition for `== Size.empty`. The new version would give a more detailed error message too.

LongCatIsLooong: Can you also add a comment that says this test should not assert / crash, or add back the expectation that `takeExceptions()` return `null`? Without the `takeExceptions()` or the comment readers will be confused as to what the test is trying to verify (since the new expectation is verifying the test itself, not "widget does not crash when it's 0x0").

dkwingsmt: > Size.isEmpty is a necessary but not sufficient condition for == Size.empty. The new version would give a more detailed error message too.

Previously we didn't use `expect(..., Size.empty);` because we wanted `Size(someNonZero, 0)` to pass as well. Maybe you're right that in this case we want exactly `== Size.empty`.

dkwingsmt: > Can you also add a comment that says this test should not assert / crash, or add back the expectation that takeExceptions() return null? 

I think it's pretty self-explanatory since the test name has said to avoid crashing.  https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193 also talked about why `expect(tester.takeException(), isNull);` is anti-pattern. I agree with the judgement, but if you disagree you're welcome to discuss under it.

LongCatIsLooong: The test name is "DesktopTextSelectionToolbarButton renders at zero area" and with that new expectation I would interpret that as "DesktopTextSelectionToolbarButton should be 0x0 if forced to be 0x0". I don't know about `takeException()` being an antipattern, it was just redundant since the test had no expectations previously.  

I think you can either add the issue link as a comment / add a comment at the end / update the test name, to help future readers understand the purpose of the test in case it fails.

LongCatIsLooong: > we wanted Size(someNonZero, 0) to pass as well

Hmm I think we actually want to make it 0x0 since most widgets have different layout algorithms horizontally and vertically. A row may work in 100 x 0 but will probably assert if you tell it to layout in 0x0.

dkwingsmt: > Hmm I think we actually want to make it 0x0 

Yes.

Re: Test name

What do you think of `DesktopTextSelectionToolbarButton does not crash at 0x0`?

justinmc (approved):
LGTM 👍

packages/flutter/test/material/desktop_text_selection_toolbar_button_test.dart
```diff
@@ -42,4 +42,19 @@ void main() {
...
```
dkwingsmt:
```suggestion
    expect(tester.getSize(find.byType(DesktopTextSelectionToolbarButton)), Size.zero);
```

dkwingsmt (commented):
Checklist:
- [ ] The target widget is expected to have a size of exactly `Size.zero`

PR #138757, opened 2023-11-20
Fix turbulence seed for all tests with ink sparkles

Piinks:
Fixes https://github.com/flutter/flutter/issues/138753

This makes sure that InkSparkles are always fixed for testing rather than random.  

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #173849, opened 2025-08-15
Reimplemented StretchingOverscrollIndicator with Simulation ported from Android 12.

MTtankkeo:
Fixed an issue #169659

This PR improves the StretchingOverscrollIndicator to better match native Android behavior.

- Previously, the stretch effect worked well only for slow dragging, and during fast flings it was absent or felt clipped.
- Implemented a Simulation-based animation that responds naturally to both slow drags and high-velocity flings.
- The result closely resembles the native Android overscroll behavior while maintaining Flutter's performance characteristics.

This change ensures that momentum-based gestures produce a more natural and intuitive overscroll experience, especially for users accustomed to native Android scroll views.

## Fling
[fling](https://github.com/user-attachments/assets/3944eef3-baaa-47df-82ca-9c49d4283fe4)

## Pull And Reduce
Also, a very similar animation is implemented when pulling as well.

[pull](https://github.com/user-attachments/assets/2a3025dc-0996-403d-b58b-d07f25487cda)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -18,6 +18,7 @@ import 'package:flutter/physics.dart' show Tolerance, nearEqual;
 import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
 
+import '../physics/spring_simulation.dart';
```
dkwingsmt:
```suggestion
import 'package:flutter/physics.dart';

```

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,142 +829,177 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
-    } else {
-      if (!_stretchController.isAnimating) {
-        assert(_stretchController.value == 1.0);
-        notifyListeners();
-      }
-    }
+  /// Creates a spring description with the given mass,
+  /// using predefined stiffness and damping values for the spring.
+  SpringDescription createSpringDescription(double mass) {
+    return SpringDescription(mass: mass, stiffness: _springStiffness, damping: _springDamping);
```
dkwingsmt: Instead of magic values for all three sping parameters mass, stiffness, and damping, I'd imagine that describing the springs with damping ratio `withDampingRatio` would give more consistent results, and the damping ratio is very likely just `1.0` (critically damped). Can you verify?

And in that case, since the mass can also be fixed on an arbitrary value (such as 1.0, the typical value used in UIKit springs), we only need to eyeball stiffness.

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,142 +829,177 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
-    switch (_state) {
-      case _StretchState.absorb:
-        _recede(_stretchDuration);
-      case _StretchState.recede:
-        _state = _StretchState.idle;
-        _pullDistance = 0.0;
-      case _StretchState.pull:
-      case _StretchState.idle:
-        break;
+    final double scaledVelocity = -(velocity * _flingVelocityFriction);
```
dkwingsmt: I'd imagine that the starting velocity to be _equal_ to the fling velocity. The velocity scaling is probably due to different unit system. If I remember correctly, `SpringSimulation` takes the full range (the distance from the starting point to the ending point) as 1.0, while `DragDetails` uses the number of pixels. Therefore the conversion rate is actually the height of the widget in pixels.

dkwingsmt:
Thank you for this amazing change! My first round of review focuses on the constants for the spring simulation.

packages/flutter/lib/src/widgets/overscroll_indicator.dart
MTtankkeo: @dkwingsmt I understand your point, but I want to implement an animation that is as close as possible to Android native. Using the approach you suggested causes it to noticeably differ from the existing Android native animation.

dkwingsmt: It's more like a refactor. The params convert with each other. What will the damping ratio be with the value you eyeballed? Let's put that it. By theory it's the damping ratio that decides the shape of the curve.

MTtankkeo: @dkwingsmt Thank you for the suggestion. I understand and will give it a try

MTtankkeo:
```dart
SpringDescription createSpringDescription(double dampingRatio) {
  final double damping = 2 * dampingRatio * math.sqrt(_springStiffness);
  return SpringDescription(mass: 1.0, stiffness: _springStiffness, damping: damping);
}
```

@dkwingsmt Is this the formula you meant? I'm not sure if I got it right, but I think the current formula feels a bit too mechanical.

dkwingsmt: Yes, this is the formula. Although you don't have to calculate them yourselves, instead you can use `SpringDescription.withDampingRatio`.

MTtankkeo:
```dart
class _StretchSimulation extends Simulation {
  _StretchSimulation(double distance, double velocity, Tolerance tolerance)
      : _dampedFreq = _naturalFrequency * math.sqrt(1 - _dampingRatio * _dampingRatio),
        _cosCoeff = distance,
        _sinCoeff = (1 / (_naturalFrequency * math.sqrt(1 - _dampingRatio * _dampingRatio))) *
            (_dampingRatio * _naturalFrequency * distance + velocity),
        _velocity = velocity,
        super(tolerance: tolerance);
  
  final double _dampedFreq;
  final double _cosCoeff;
  final double _sinCoeff;
  final double _velocity;

  // Physical constants ported directly from Android's EdgeEffect.java.
  static const double _naturalFrequency = 24.657;
  static const double _dampingRatio = 0.98;
  static const double _timeCorrectionFactor = 0.8;

  @override
  double x(double time) {
    final double correctedTime = time * _timeCorrectionFactor;

    // This is the direct port of the distance calculation from Android's
    // damped harmonic oscillator formula.
    return math.pow(math.e, -_dampingRatio * _naturalFrequency * correctedTime) *
        (_cosCoeff * math.cos(_dampedFreq * correctedTime) + _sinCoeff * math.sin(_dampedFreq * correctedTime));
  }

  @override
  double dx(double time) {
    // To calculate the velocity (the derivative of x(t)), we apply the same
    // time correction.
    final double correctedTime = time * _timeCorrectionFactor;

    // The derivative calculation ported from the original Android 12 code.
    final double position = x(time);
    final double uncorrectedDx = position * (-_naturalFrequency) * _dampingRatio +
        math.pow(math.e, -_dampingRatio * _naturalFrequency * correctedTime) *
            (-_dampedFreq * _cosCoeff * math.sin(_dampedFreq * correctedTime) +
              _dampedFreq * _sinCoeff * math.cos(_dampedFreq * correctedTime));

    return uncorrectedDx * _timeCorrectionFactor;
  }

  @override
  bool isDone(double time) {
    // The simulation is considered done when both position and velocity
    // are within the tolerance thresholds. This logic is similar to Android's
    // isAtEquilibrium().
    return x(time).abs() < tolerance.distance && dx(time).abs() < tolerance.velocity;
  }
}
```

@dkwingsmt I thought it wouldn’t be possible to achieve the exact same effect with Flutter’s built-in SpringSimulation—at best, it could only look somewhat similar. So I ported the original Android code and reworked it into a simulation form. The problem is that the motion feels about 1.5x faster (I can fix it by just multiplying by around 0.8), but I’m not sure why. Could this discrepancy be due to a difference in how Flutter and Android handle their simulation logic?

By the way, the simulation seems identical to the original Android one as long as I adjust the speed!

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,223 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
+  void removeListener(VoidCallback listener) {
+    _overscrollNotifier.removeListener(listener);
+  }
+
+  /// Creates a stretching-only [Simulation] ported from Android 12.
+  _StretchSimulation _createStretchSimulation(double velocity) {
+    return _StretchSimulation(overscroll, velocity, Tolerance.defaultTolerance);
+  }
+
+  /// Uses the saturation function approach to scale velocity
```
victorsanni:
```suggestion
  /// Uses the saturation function approach to scale velocity.
```

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,223 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
-    _stretchController.duration = duration;
-    _stretchController.forward(from: 0.0);
-    _state = _StretchState.recede;
+    final double pullDistance = normalizedOverscroll;
+    final double absDistance = pullDistance.abs();
+    final double linearIntensity = _stretchIntensity * absDistance;
+    final double exponentialIntensity =
+        _stretchIntensity * (1 - math.exp(-absDistance * _exponentialScalar));
+
+    // Maintain sign of overscroll for direction
```
victorsanni:
```suggestion
    // Maintain sign of overscroll for direction.
```

victorsanni (commented):
Failing customer tests: https://github.com/zulip/zulip-flutter/blob/f304ec4644ad8c3f454a11799acb4b403d8aff71/test/widgets/emoji_reaction_test.dart#L604-L652

victorsanni:
Oh I see you have https://github.com/zulip/zulip-flutter/pull/1833 out already. Thanks for that!

MTtankkeo:
In test for zulip, it's still being checked out to `c87de4169b2c79e38d73c9d79ecbcc63f2e603de`. Could you update it to the new commit `313f8caff7099b0991fc50cb116ae7145e459152`?

MTtankkeo:
@dkwingsmt When would you be able to review this?

dkwingsmt:
Hi! I'm very sorry for the late review. We've been a bit busy lately, especially with a summit last week. But I'll review it right now.

dkwingsmt:
Regarding simulation param: This is so cool! I'm so excited that you can find the Android code and replicate it in Flutter. Since your code is basically a damped spring, I was able to convert it to Flutter's spring simulation:

```dart
  final Simulation stretch = StretchSimulation(10, 1, Tolerance.defaultTolerance);
  for (double t = 0; t < 1; t += 0.1) {
    print('Stretch($t) = ${stretch.x(t)}');
  }

  // Physical constants ported directly from Android's EdgeEffect.java.
  const double kNaturalFrequency = 24.657;
  const double kDampingRatio = 0.98;
  const double kTimeCorrectionFactor = 0.8;

  const double kStiffness = kNaturalFrequency * kNaturalFrequency;
  final SpringDescription springDescription =
      SpringDescription.withDampingRatio(
          mass: 1, stiffness: kStiffness, ratio: kDampingRatio);

  final Simulation spring = SpringSimulation(springDescription, 10, 0, 1);
  for (double t = 0; t < 1; t += 0.1) {
    print('Spring($t) = ${spring.x(kTimeCorrectionFactor * t)}');
  }
```

They gave the same result as long as the Flutter version also uses the timeCorrectionFactor.
```
flutter: Stretch(0.0) = 10.0
flutter: Stretch(0.1) = 4.07408596499882
flutter: Stretch(0.2) = 0.8799738842349217
flutter: Stretch(0.30000000000000004) = 0.14999067992515963
flutter: Stretch(0.4) = 0.021680827237545313
flutter: Stretch(0.5) = 0.002656703351596544
flutter: Stretch(0.6) = 0.0002564212307430861
flutter: Stretch(0.7) = 0.000012938602458735777
flutter: Stretch(0.7999999999999999) = -0.0000019092137489869062
flutter: Stretch(0.8999999999999999) = -7.813953057815709e-7
flutter: Stretch(0.9999999999999999) = -1.689591067594363e-7
flutter: Spring(0.0) = 10.0
flutter: Spring(0.1) = 4.074085964998819
flutter: Spring(0.2) = 0.8799738842349216
flutter: Spring(0.30000000000000004) = 0.14999067992515955
flutter: Spring(0.4) = 0.021680827237545292
flutter: Spring(0.5) = 0.002656703351596539
flutter: Spring(0.6) = 0.0002564212307430852
flutter: Spring(0.7) = 0.00001293860245873564
flutter: Spring(0.7999999999999999) = -0.000001909213748986927
flutter: Spring(0.8999999999999999) = -7.813953057815737e-7
flutter: Spring(0.9999999999999999) = -1.689591067594365e-7
```

In order to bake the time correction factor into our simulation, we need to do the following things:
* Multiply `naturalFrequency` by `kTimeCorrectionFactor` (effectively multiplying `stiffness` by `kTimeCorrectionFactor^2`)
* Multiply starting velocity by `kTimeCorrectionFactor`.

This is the resulting code:
```dart
  final Simulation spring2 = SpringSimulation(
      SpringDescription.withDampingRatio(
          mass: 1,
          stiffness: kStiffness * kTimeCorrectionFactor * kTimeCorrectionFactor,
          ratio: kDampingRatio),
      10,
      0,
      /*velocity*/ 1 * kTimeCorrectionFactor);
  for (double t = 0; t < 1; t += 0.1) {
    print('Spring2($t) = ${spring2.x(t)}');
  }
```
The result proved the same:
```
flutter: Spring2(0.0) = 10.0
flutter: Spring2(0.1) = 4.07408596499882
flutter: Spring2(0.2) = 0.8799738842349218
flutter: Spring2(0.30000000000000004) = 0.14999067992515971
flutter: Spring2(0.4) = 0.02168082723754534
flutter: Spring2(0.5) = 0.0026567033515965495
flutter: Spring2(0.6) = 0.00025642123074308724
flutter: Spring2(0.7) = 0.00001293860245873592
flutter: Spring2(0.7999999999999999) = -0.00000190921374898688
flutter: Spring2(0.8999999999999999) = -7.813953057815676e-7
flutter: Spring2(0.9999999999999999) = -1.6895910675943604e-7
```

Can you rewrite the `_StretchSimulation` part with this? Also explain that the time correction factor is derived from eyeballing and is needed due to unknown reason. Here's my proposal:
```
  // A correction factor applied to the simulation time.
  //
  // The physical constants `_naturalFrequency` and `_dampingRatio` were ported
  // directly from Android's `EdgeEffect.java` source. However, using these
  // constants as-is resulted in an animation that was noticeably faster than
  // the native Android behavior. The underlying reason for this discrepancy is
  // unknown.
  //
  // This factor, determined by visual comparison ("eyeballing") to match the
  // platform's timing, is applied to the elapsed time `t` to slow down the
  // simulation.
  //
  // Based on the damped harmonic oscillator equations, an alternative,
  // mathematically equivalent, approach would be to multiply both the
  // `_naturalFrequency` and the initial velocity by this same factor,
  // rather than scaling the time input.
```

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,223 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
+  /// Creates a stretching-only [Simulation] ported from Android 12.
+  _StretchSimulation _createStretchSimulation(double velocity) {
+    return _StretchSimulation(overscroll, velocity, Tolerance.defaultTolerance);
+  }
+
+  /// Uses the saturation function approach to scale velocity.
+  /// Small velocities remain roughly proportional, while large
+  /// velocities gradually saturate. Returns the scaled velocity
+  /// with the original direction preserved.
+  double _calculateVelocityScale(double velocity) {
```
dkwingsmt: Can you explain where this part comes from? I can't find similar code from either `EdgeEffect` or the code before the PR.

dkwingsmt (commented):
In general it looks great!

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,238 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
+/// This class is a direct port of Android's EdgeEffect.java:
+/// https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/widget/EdgeEffect.java
+///
+/// It reproduces the damped harmonic oscillator behavior of Android's
+/// overscroll effect in Flutter, using the same physical constants:
+/// - natural frequency
+/// - damping ratio
+///
+/// The formulas and coefficients are derived directly from the Android source.
+class _StretchSimulation extends Simulation {
```
dkwingsmt: Can you replace this with `SpringSimulation` as in the `spring2` in [this comment](https://github.com/flutter/flutter/pull/173849#issuecomment-3310730196)? Since Flutter has a `SpringSimulation` it's weird to implement it again.

MTtankkeo: @dkwingsmt 

Is the code correct with what you said?

```dart
/// Creates a stretching-only [Simulation] ported from Android 12.
  SpringSimulation _createStretchSimulation(double velocity) {
    // Physical constants ported directly from Android's EdgeEffect.java.
    const double kNaturalFrequency = 24.657;
    const double kDampingRatio = 0.98;

    // A correction factor applied to the simulation time.
    //
    // The physical constants `_naturalFrequency` and `_dampingRatio` were ported
    // directly from Android's `EdgeEffect.java` source. However, using these
    // constants as-is resulted in an animation that was noticeably faster than
    // the native Android behavior. The underlying reason for this discrepancy is
    // unknown.
    //
    // This factor, determined by visual comparison ("eyeballing") to match the
    // platform's timing, is applied to the elapsed time `t` to slow down the
    // simulation.
    //
    // Based on the damped harmonic oscillator equations, an alternative,
    // mathematically equivalent, approach would be to multiply both the
    // `_naturalFrequency` and the initial velocity by this same factor,
    // rather than scaling the time input.
    const double kTimeCorrectionFactor = 0.8;

    const double kStiffness = kNaturalFrequency * kNaturalFrequency;
    return SpringSimulation(
      SpringDescription.withDampingRatio(
        mass: 1,
        stiffness: kStiffness * kTimeCorrectionFactor * kTimeCorrectionFactor,
        ratio: kDampingRatio,
      ),
      overscroll,
      0,
      velocity * kTimeCorrectionFactor
    );
  }
```

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,203 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
+
+  @override
+  void removeListener(VoidCallback listener) {
+    _overscrollNotifier.removeListener(listener);
+  }
+
+  /// Creates a stretching-only [Simulation] ported from Android 12.
+  _StretchSimulation _createStretchSimulation(double velocity) {
+    return _StretchSimulation(overscroll, velocity);
+  }
```
dkwingsmt:
```suggestion
  _StretchSimulation _createStretchSimulation(double velocity) {
    return SpringSimulation(
      _kStretchSpringDescription,
      overscroll,
      0,
      velocity * kTimeCorrectionFactor
    );
  }
```

While making the spring description a global constant:
```dart
final SpringDescription _kStretchSpringDescription = SpringDescription.withDampingRatio(
        mass: 1,
        stiffness: kStiffness * kTimeCorrectionFactor * kTimeCorrectionFactor,
        ratio: kDampingRatio,
      );
```

Then remove the `_StretchSimulation` class while moving the constants needed to global, such as `_kStiffness`.

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,203 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
   ///
-  /// The velocity must be positive.
-  void absorbImpact(double velocity, double totalOverscroll) {
-    assert(velocity >= 0.0);
-    velocity = clampDouble(velocity, _kMinVelocity, _kMaxVelocity);
-    _stretchSizeTween.begin = _stretchSize.value;
-    _stretchSizeTween.end = math.min(_stretchIntensity + (_flingFriction / velocity), 1.0);
-    _stretchController.duration = Duration(
-      milliseconds: math.max(velocity * 0.02, _kMinStretchDuration.inMilliseconds).round(),
+  /// The physical constants `_naturalFrequency` and `_dampingRatio` were ported
```
dkwingsmt:
```suggestion
  /// The physical constants `kNaturalFrequency ` and `kDampingRatio ` were ported
```

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -834,138 +826,203 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
+  /// the native Android behavior. The underlying reason for this discrepancy is
+  /// unknown.
+  ///
+  /// This factor, determined by visual comparison ("eyeballing") to match the
+  /// platform's timing, is applied to the elapsed time `t` to slow down the
+  /// simulation.
+  ///
+  /// Based on the damped harmonic oscillator equations, an alternative,
+  /// mathematically equivalent, approach would be to multiply both the
+  /// `_naturalFrequency` and the initial velocity by this same factor,
```
dkwingsmt:
```suggestion
  /// `kNaturalFrequency ` and the initial velocity by this same factor,
```

dkwingsmt (approved):
LGTM, thank you!

dkwingsmt:
I'm looking for someone more familiar with overscrolling to give a 2nd review.

packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -760,16 +749,23 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
...
-            _stretchController.pull(clampedOverscroll, _totalOverscroll);
+            final double distanceForPull = _totalOverscroll / viewportDimension;
+            final double clampedOverscroll = clampDouble(distanceForPull, -1.0, 1.0);
+            _stretchController.pull(clampedOverscroll);
           }
         }
       }
-    } else if (notification is ScrollEndNotification || notification is ScrollUpdateNotification) {
+    } else if (notification is ScrollEndNotification) {
+      final double velocity = widget.axis == Axis.vertical
```
Piinks: Could you make this a switch statement for clarity? https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-if-chains-or--or--with-enum-values

packages/flutter/test/material/paginated_data_table_test.dart
```diff
@@ -583,6 +583,7 @@ void main() {
   testWidgets('PaginatedDataTable footer scrolls', (WidgetTester tester) async {
     await tester.pumpWidget(
       MaterialApp(
+        scrollBehavior: const MaterialScrollBehavior().copyWith(overscroll: false),
```
Piinks: Why is this needed here?

packages/flutter/test/widgets/overscroll_stretch_indicator_test.dart
```diff
@@ -1310,6 +1308,50 @@ void main() {
       expect(box3.localToGlobal(Offset.zero).dy, 350.0);
     },
   );
+
+  testWidgets('Fling overscroll triggers stretching animation correctly', (
```
Piinks, line 1312: Is there a test here capturing a reversed scrollview to ensure it works as expected?

packages/flutter/test/material/paginated_data_table_test.dart
MTtankkeo: This test is intended to verify the scroll/layout position of the footer, and overscroll effects (stretch) introduce non-deterministic behavior that makes the test unstable (in fact, it failed due to being simulation-based). Therefore, I applied scrollBehavior: const MaterialScrollBehavior().copyWith(overscroll: false).

Another option would be to use await tester.pumpAndSettle() and wait for the overscroll animation to finish, but I believe disabling overscroll entirely is a simpler and more robust approach.

Piinks: > overscroll effects (stretch) introduce non-deterministic behavior that makes the test unstable

Hmm, this makes me wonder if in the future, unknowing contributors will end up writing unstable tests that are flaky because of this non-deterministic behavior. 

Is there a way we can stabilize the effect for the purposes of testing? I recalled doing something similar for #138757 to stabilize the ink sparkle when it is in tests.

MTtankkeo: @Piinks The approaches that come to mind right now are either explicitly disabling overscroll in exceptional cases, or using pumpAndSettle to ensure that the overscroll animation has finished before proceeding with the next test. If there are other approaches you’d suggest, I’d be happy to explore them.

gnprice: When this line isn't changed, is the test really non-deterministic — does it sometimes pass and sometimes fail? I would have expected the simulation's behavior to be deterministic.

It looks like what's happening in this test is: it wants to scroll all the way to one end, and it does so by doing a drag for 1000px, which presumably is a lot more than the total distance the scrollable can scroll. That causes overscroll, but then is sure to wind up at the end of the range afterward.

For this test, I think it'd be pretty reasonable to do the 1000px drag, and then `pumpAndSettle` to let the overscroll finish. If the behavior is deterministic, then I think a contributor newly writing a test like this one would figure out a solution like that without too much trouble, after a first attempt had the test fail. But if it really is non-deterministic, then that's a problem.

MTtankkeo: @gnprice From my perspective, calling it non-deterministic might have been a misunderstanding. The issue seems to have arisen because a single pump wasn’t sufficient to stabilize the overscroll effect. As you suggested, using pumpAndSettle should be able to sufficiently stabilize the overscroll behavior.

Piinks (approved):
LGTM

dkwingsmt:
The customer test failure seems to be real. Can you take a look?

MTtankkeo:
@dkwingsmt https://github.com/flutter/tests/pull/471 https://github.com/zulip/zulip-flutter/pull/1833
I already fixed the customer test failure, but I need to change the checkout because the zulip test was a problem.

Piinks:
> [flutter/tests#471](https://github.com/flutter/tests/pull/471) [zulip/zulip-flutter#1833](https://github.com/zulip/zulip-flutter/pull/1833) I already fixed the customer test failure, but I need to change the checkout because the zulip test was a problem.

Once [flutter/tests#471](https://github.com/flutter/tests/pull/471) lands (it looks like there is a conflict right now), we'll need another PR to bump the version of flutter/tests that flutter/flutter checks out for customer_testing - dev/customer_testing/tests.version

Piinks:
I can help with the Google test failures once customer_testing is set.

MTtankkeo:
The customer_testing and other test issues have been resolved.

dkwingsmt:
I'd let @Piinks double check one last time with everything. However @Piinks is out for the week and will return by Oct 20.

Piinks:
Thanks Tong! I am rebasing here since it looks like there is no Google testing instance currently. We need that to validate we have resolved any remaining failures before merging.

Piinks:
@MTtankkeo can you make this update in behavior opt-in? I have determined we won't be able to land this based on some of the failures coming from other dependencies that will have to update first once it gets to stable - https://pub.dev/packages/data_table_2 being one of them.

MTtankkeo:
> @MTtankkeo can you make this update in behavior opt-in? I have determined we won't be able to land this based on some of the failures coming from other dependencies that will have to update first once it gets to stable - https://pub.dev/packages/data_table_2 being one of them.

@Piinks How should I implement the opt-in behavior? Personally, I can’t think of a good approach right now. Could you provide an example?

Piinks:
One thing we could do is update data table to soften the affected tests, and roll the updated package into google, and then land this change. It looks like it's been 7 months since the package's last update.

MTtankkeo:
> @MTtankkeo which do you prefer, updating the package and then returning here, or introducing an opt in?

@Piinks Sounds good, I’ll look into fixing the failing tests in the data_table_2 package and open a PR soon.

PR #173852, opened 2025-08-15
Create Hot Restart over websocket test

### LINKED ISSUE #2669: Create e2e hot restart test

jyameo:
- Create hot restart over websocket test
- refactored websocket_dwds_test_common

Closes https://github.com/dart-lang/webdev/issues/2669

Blocked by hot restart bug: Fix in https://github.com/dart-lang/webdev/pull/2668 and https://github.com/flutter/flutter/pull/173777

- This PR can be landed once the bug above is fixed in dwds and the dwds version is updated


packages/flutter_tools/test/integration.shard/test_data/websocket_dwds_test_common.dart
```diff
@@ -0,0 +1,187 @@
...
+  }) async {
+    debugPrint('Step 1: Starting WebSocket DWDS connection setup...');
+
+    // Set up listening for app output before starting
+    final stdout = StringBuffer();
+    final sawDebugUrl = Completer<String>();
+    final StreamSubscription<String> subscription = flutter.stdout.listen((String e) {
+      stdout.writeln(e);
+      // Extract the debug connection URL
+      if (e.contains('Waiting for connection from Dart debug extension at http://')) {
```
biggs0125, line 62: If we have the regexp we can just use `firstMatch` and it'll be `null` if there isn't a match. It's harder to keep 2 copies of this string in sync.

packages/flutter_tools/test/integration.shard/test_data/websocket_dwds_test_common.dart
```diff
@@ -0,0 +1,187 @@
...
+        final debugUrlPattern = RegExp(
+          r'Waiting for connection from Dart debug extension at (http://[^\s]+)',
+        );
+        final Match? match = debugUrlPattern.firstMatch(e);
+        if (match != null && !sawDebugUrl.isCompleted) {
```
biggs0125, line 67: Is it possible to see the debug URL twice? Seems like something is off if that happens so perhaps we should throw a clear error. Otherwise it'll just look like a timeout.

packages/flutter_tools/test/integration.shard/test_data/websocket_dwds_test_common.dart
```diff
@@ -0,0 +1,187 @@
...
+          sawDebugUrl.complete(match.group(1)!);
+        }
+      }
+    });
+
+    io.Process? chromeProcess;
+    try {
+      // Step 1: Start Flutter app with web-server device (will wait for debug connection)
```
biggs0125, line 75: Comment step numbers are all off by 1. I actually think the comments are unnecessary since the `debugPrint` is self-commenting.

srujzs (approved):

packages/flutter_tools/test/integration.shard/test_data/websocket_dwds_test_common.dart
```diff
@@ -0,0 +1,187 @@
...
+      );
+      debugPrint('✓ Flutter app started successfully with WebSocket connection');
+
+      return WebSocketDwdsTestSetup(
+        stdout: stdout,
+        chromeProcess: chromeProcess,
+        subscription: subscription,
+      );
+    } catch (e) {
+      // Clean up on error
```
biggs0125, line 113: Should this call `cleanupWebSocketTestResources`?

packages/flutter_tools/test/integration.shard/hot_reload_websocket_test.dart
```diff
@@ -48,62 +44,17 @@ void testAll({List<String> additionalCommandArgs = const <String>[]}) {
...
+        final WebSocketDwdsTestSetup setup = await WebSocketDwdsTestUtils.setupWebSocketConnection(
+          flutter,
+          additionalCommandArgs: additionalCommandArgs,
+        );
 
-          // Step 5: Test hot reload functionality
-          debugPrint('Step 5: Testing hot reload with WebSocket connection...');
+        try {
+          // Test hot reload functionality
+          debugPrint('Step 6: Testing hot reload with WebSocket connection...');
```
biggs0125, line 57: Having step numbers here requires keeping all these files in sync. If the common file changes then we would have to update this one too. I think for the actual testing we can omit the step numbers.

PR #136239, opened 2023-10-10
[SingleChildScrollView] Correct the offset pixels if it is out of range during layout

xu-baolin:
Fixes https://github.com/flutter/flutter/issues/105733
FIxes https://github.com/flutter/flutter/issues/135865

The above two issues will not occur on the `ListView` widget, so after analyzing their differences, it was found that the `SingleChildScrollView` pixels were not corrected according to the content dimensions during layout.

### ListView correct the pixels codes
https://github.com/flutter/flutter/blob/5dfd78c2e30483a9a8668ba6b754d089b1e132c8/packages/flutter/lib/src/rendering/viewport.dart#L1462-L1465

We should correct the pixels to prevent the physics ballistic simulations.

This change makes them have the same behavior.

Issue #105733, opened 2022-06-09
Scroll with unnecessary animation and updates when maximizing window.


caduandrade:
Hi!

I recorded this GIF doing the horizontal resizing (to have a smaller image) but it generates the same effect when maximize the window.

![scroll](https://user-images.githubusercontent.com/8809484/172934140-2b4e16f0-7d53-49ab-a713-dee5a1b6c969.gif)

You can see that the scroll keeps changing its size and the content keeps moving.

This happens when the scroll offset is at the end.

In horizontal resizing, it is understandable because it must generate multiple events. But it gets really weird when maximizing the window (should be only one event) and seeing the entire content slowly move to the side.

I've also noticed that the scroll offset is updated numerous times when maximizing the window as if trying to adjust until it succeeds. Updates before the last one are also with offset values greater than the maximum allowed. This also happens when resizing horizontally.

## Steps to Reproduce

1. move scroll to the end
2. maximize the window

## Code

```dart
class HomePage extends StatefulWidget {
  const HomePage({Key? key}) : super(key: key);

  @override
  State<HomePage> createState() => _HomePageState();
}

class _HomePageState extends State<HomePage> {
  final ScrollController _scrollController = ScrollController();

  @override
  void initState() {
    _scrollController.addListener(_scrollListener);
    super.initState();
  }

  @override
  void dispose() {
    _scrollController.dispose();
    super.dispose();
  }

  void _scrollListener() {
    // offset bigger than maxScrollExtent?
    print(
        'offset: ${_scrollController.offset} / maxScrollExtent: ${_scrollController.position.maxScrollExtent}');
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
        body: Theme(
            data: ThemeData(
                scrollbarTheme: ScrollbarThemeData(
                    thumbVisibility: MaterialStateProperty.all(true),
                    thumbColor: MaterialStateProperty.all(Colors.red))),
            child: Scrollbar(
                controller: _scrollController,
                child: SingleChildScrollView(
                    controller: _scrollController,
                    scrollDirection: Axis.horizontal,
                    child:
                        const SizedBox(width: 2000, child: Placeholder())))));
  }
}
```

## Prints from the listener

```
...
offset: 82.05680889386778 / maxScrollExtent: 80
offset: 81.76819620001922 / maxScrollExtent: 80
offset: 81.51992893399559 / maxScrollExtent: 80
offset: 81.30657864605345 / maxScrollExtent: 80
[...]

PR #173857, opened 2025-08-15
Fix the issue where SingleChildScrollView resets its scroll position after relayout.

### LINKED ISSUE #145078: [SingleChildScrollView] Jitter when updating text during overscroll
### Steps to reproduce

In my prodution app I have a visible stopwatch on a `SingleChildScrollView`. I've included sample code which has the same issue. I think it's related to https://github.com/flutter/flutter/pull/136239, but that one is already closed.

1. run the example app
2. try to scroll up

### Expected results

The scroll position stays the same and there's no visible jitter.

### Actual results

The scroll position is reset everytime the text gets updated.

### Code sample

```dart
import 'dart:async';
import 'dart:math';

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return const MaterialApp(
      home: MyHomePage(),
    );
  }
}

class MyHomePage extends StatelessWidget {

  @override
    return Scaffold(
      appBar: AppBar(),
      body: SingleChildScrollView(
        child: Column(
          children: [
            const _Counter(),
            for (var i = 0; i < 15; i++)
              Container(
                height: 100,
                color: Color.fromRGBO(Random().nextInt(256), Random().nextInt(256), Random().nextInt(256), 1),
              ),
          ],
        ),
      ),
    );
  }
}

class _Counter extends StatefulWidget {
  const _Counter();

  @override
  State<_Counter> createState() => __CounterState();
}

class __CounterState extends State<_Counter> {
  int _count = 0;

  @override
  void initState() {
    Timer.periodic(const Duration(milliseconds: 500), (timer) {
      setState(() {
        _count++;
      });
    });
    super.initState();
  }

  @override
    return Text('$_count');
  }
}

```

</details>

### Screenshots or Video

<details open>

**Expected behaviour**

https://github.com/flutter/flutter/assets/27430651/a74ba7bb-9598-45f4-88bc-34be11faa897

**Actual behaviour**

https://github.com/flutter/flutter/assets/27430651/350a7881-96d3-4eb3-b035-6e0ff84982a6

</details>

### Logs

_No response_

### Flutter Doctor output

```console
flutter doctor
[✓] Flutter (Channel stable, 3.19.3, on macOS 14.3.1 23D60 darwin-arm64, locale nl-NL)
[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome)
[✓] Android Studio (version 2023.2)
[✓] VS Code (version 1.87.0)
    ! Error: Browsing on the local area network for iPhone van Govert. Ensure the device is unlocked and attached with a cable or associated with the same
      local area network as this Mac.
[✓] Network resources

```

</details>

### LINKED ISSUE #149018: Refresh Indicator With SingleChildScrollView With BouncingScrollPhysics, Page Offset Jump After Refresh
### Steps to reproduce

 When using Refresh indicator with SingleChildScrollView with physic BouncingScrollPhysics, Page jump after using refresh. this is not the case in version 3.16. but since 13.19 and now in version 3.22.1 it still happens. 
 
 1. Refresh indicator with SingleChildScrollView with physic BouncingScrollPhysics, it's error
 2. But Refresh indicator with ListView with physic BouncingScrollPhysics, it's fine

### Expected results

Refresh indicator with ListView with physic BouncingScrollPhysics, 

[Screen_recording_20240524_102754.webm](https://github.com/flutter/flutter/assets/36793730/8e187187-3e85-4051-917f-dcb35be4ab4c)

### Actual results

Refresh indicator with SingleChildScrollView with physic BouncingScrollPhysics,

[Screen_recording_20240524_101517.webm](https://github.com/flutter/flutter/assets/36793730/8649237d-0fc5-4542-a010-2c09cfbc9ae5)

### Code sample

<details open><summary>Code sample With SingleChildScrollView</summary>

```dart
import 'dart:async';

import 'package:flutter_hooks/flutter_hooks.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends HookWidget {
  const MyApp({super.key});

  @override
    final isLoading = useState(false);
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(title: const Text("Refresh Indicator")),
        body: RefreshIndicator(
          onRefresh: () async {
            isLoading.value = true;
            Future.delayed(
                const Duration(seconds: 3), () => isLoading.value = false);
          },
          child: SingleChildScrollView(
            physics: const AlwaysScrollableScrollPhysics(
                parent: BouncingScrollPhysics()),
            padding: const EdgeInsets.symmetric(horizontal: 20),
            child: Column(
              children: List.generate(
                100,
                (_) => isLoading.value
                    ? Row(children: [
                        const CircularProgressIndicator(),
                        const SizedBox(width: 5),
                        Text("loading $_")
                      ])
                    : Text("data $_"),
              )
                  .expand(
                    (_) => [_, const SizedBox(height: 10)],
                  )
                  .toList(),
            ),
          ),
        ),
      ),
    );
  }
}

```

</details>

<details open><summary>Code sample With ListView</summary>

```dart

import 'dart:async';

import 'package:flutter_hooks/flutter_hooks.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends HookWidget {
  const MyApp({super.key});

  @override
    final isLoading = useState(false);
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(title: const Text("Refresh Indicator")),
        body: RefreshIndicator(
          onRefresh: () async {
            isLoading.value = true;
            Future.delayed(
                const Duration(seconds: 3), () => isLoading.value = false);
          },
          child: ListView(
            physics: const AlwaysScrollableScrollPhysics(
                parent: BouncingScrollPhysics()),
            padding: const EdgeInsets.symmetric(horizontal: 20),
            children: List.generate(
              100,
              (_) => isLoading.value
                  ? Row(children: [
                      const CircularProgressIndicator(),
                      const SizedBox(width: 5),
                      Text("loading $_")
                    ])
                  : Text("data $_"),
            )
                .expand(
                  (_) => [_, const SizedBox(height: 10)],
                )
                .toList(),
          ),
        ),
      ),
    );
  }
}


```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.22.1, on macOS 14.4.1 23E224 darwin-arm64, locale en-ID)
    • Flutter version 3.22.1 on channel stable at /Users/User/Downloads/flutter
    • Framework revision a14f74ff3a (35 hours ago), 2024-05-22 11:08:21 -0500
    • Engine revision 55eae6864b
    • Dart version 3.4.1
    • DevTools version 2.34.3

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/User/Library/Android/sdk
    • Platform android-34, build-tools 33.0.0
    • Java version OpenJDK Runtime Environment (build 17.0.10+0-17.0.10b1087.21-11572160)

[✓] Xcode - develop for iOS and macOS (Xcode 15.0.1)
    • Build 15A507
    • CocoaPods version 1.14.3

[✓] Android Studio (version 2023.3)
    • Java version OpenJDK Runtime Environment (build 17.0.10+0-17.0.10b1087.21-11572160)

[✓] VS Code (version 1.89.1)
    • Flutter extension version 3.88.0

    • sdk gphone arm64 (mobile)       • emulator-5554         • android-arm64  • Android 11 (API 30) (emulator)
    • macOS (desktop)                 • macos                 • darwin-arm64   • macOS 14.4.1 23E224 darwin-arm64
    • Mac Designed for iPad (desktop) • mac-designed-for-ipad • darwin         • macOS 14.4.1 23E224 darwin-arm64
    • Chrome (web)                    • chrome                • web-javascript • Google Chrome 125.0.6422.76

[✓] Network resources

```

</details>

### LINKED ISSUE #173225: [ios] ValueListenableBuilder rebuild interrupts ongoing scroll gesture
### Steps to reproduce

1. Have a ValueListenableBuilder inside SingleChildScrollView.
2. Scroll so much that there is rubber banding effect.
3. Keep finger on display to "over scroll".
4. When ValueNotifier value changes ValueListenableBuilder builds, scrolling is ended and causes ScrollView to go to begining(or end) of view.

### Expected results

When ValueListenableBuilder rebuilds, scrolling should continue seamlessly.

### Actual results

Scroll gesture ends when ValueListenableBuilder rebuilds.

### Code sample

```dart

import 'dart:async';

  const MainApp({super.key});

  @override
}

class _MainAppState extends State<MainApp> {
  final titleValue = ValueNotifier<String>('0');
  int counter = 0;

  @override
  void initState() {
    super.initState();

    Timer.periodic(const Duration(milliseconds: 1500), (timer) {
      counter++;
      titleValue.value = counter.toString();
    });
  }

  @override
    return MaterialApp(
      home: Scaffold(
        body: SingleChildScrollView(
          child: ConstrainedBox(
            constraints: BoxConstraints(
              minHeight: MediaQuery.sizeOf(context).height * 1.2,
            ),
            child: Column(
              children: [
                const SizedBox(height: 300),
                Row(
                  children: [
                    const Text('Time: ', style: TextStyle(color: Colors.red)),
                    ValueListenableBuilder(
                      valueListenable: titleValue,
                      builder: (context, value, child) {
                        return Text(
                          value,
                          style: const TextStyle(
                            color: Colors.blue,
                            fontFeatures: [FontFeature.tabularFigures()],
                          ),
                        );
                      },
                    ),
                  ],
                ),
                Container(width: 200, height: 200, color: Colors.blue),
              ],
            ),
          ),
        ),
      ),
    );
  }
}

void main() {
  runApp(const MainApp());
}

```

</details>

### Screenshots or Video

<details open>

[[Upload media here]](https://github.com/user-attachments/assets/aec7daa4-5919-4086-8c2f-8bbbf0ef35b6)

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.32.8, on macOS 15.6 24G84 darwin-arm64, locale en-US) [257ms]
    • Flutter version 3.32.8 on channel stable at /Users/mm/Coding/flutter/flutter
    • Framework revision edada7c56e (10 days ago), 2025-07-25 14:08:03 +0000
    • Engine revision ef0cd00091
    • Dart version 3.8.1
    • DevTools version 2.45.1

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0) [973ms]
    • Android SDK at /Users/mm/Library/Android/sdk
    • Platform android-35, build-tools 34.0.0
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13368085-b895.109)

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [509ms]
    • Build 16F6
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [9ms]

[✓] Android Studio (version 2024.3) [9ms]
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13368085-b895.109)

[✓] VS Code (version 1.102.3) [8ms]
    • Flutter extension version 3.116.0

[✓] Connected device (4 available) [5.7s]
    • GoldenEye (2) (wireless) (mobile) • 00008110-000279691122801E            • ios            • iOS 18.6 22G86
    • iPhone 16 (mobile)                • AC5080C6-540F-4CBD-B01C-FC0146903FC7 • ios            • com.apple.CoreSimulator.SimRuntime.iOS-18-5 (simulator)
    • macOS (desktop)                   • macos                                • darwin-arm64   • macOS 15.6 24G84 darwin-arm64
    • Chrome (web)                      • chrome                               • web-javascript • Google Chrome 138.0.7204.184

[✓] Network resources [262ms]
```

</details>

yiiim:
Fixes: #145078
Fixes: #173225
Fixes: #149018

This PR adopts the scroll correction approach from `RenderViewport`, repeatedly calling `applyContentDimensions`. Additionally, it triggers a relayout on every scroll (as `RenderViewport` does) to update `_lastMetrics` in [scroll_position.dart](https://github.com/flutter/flutter/blob/da5c047a13d70a3797ca3edba3cc3220801fb84f/packages/flutter/lib/src/widgets/scroll_position.dart#L679).  
This PR also reverts several tests that were changed in #136239, because the code added in #136239 has now been replaced by this PR. Some tests modified by #136239 have been restored to their original state, but the new tests added by #136239 are still retained.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -806,10 +806,15 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
     // guard because the super call below would start a ballistic scroll activity.
     if (!_viewportDimensionWasNonZero || _needsPixelsCorrection) {
       _needsPixelsCorrection = false;
-      correctPixels(
-        tabBar._initialScrollOffset(viewportDimension, minScrollExtent, maxScrollExtent),
+      final double newPixels = tabBar._initialScrollOffset(
+        viewportDimension,
+        minScrollExtent,
+        maxScrollExtent,
       );
-      result = false;
+      if (newPixels != (hasPixels ? pixels : null)) {
+        correctPixels(newPixels);
+        result = false;
+      }
     }
     return super.applyContentDimensions(minScrollExtent, maxScrollExtent) && result;
   }
--- a/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
@@ -493,6 +493,8 @@ class _RenderSingleChildViewport extends RenderBox
     return constraints.constrain(childSize);
   }
 
+  static const int _maxLayoutCycles = 10;
+
   @override
   void performLayout() {
     final BoxConstraints constraints = this.constraints;
@@ -502,8 +504,7 @@ class _RenderSingleChildViewport extends RenderBox
       child!.layout(_getInnerConstraints(constraints), parentUsesSize: true);
       size = constraints.constrain(child!.size);
     }
-
-    while (true) {
+    for (int i = 0; i < _maxLayoutCycles; i++) {
       final bool didAcceptViewportDimension = offset.applyViewportDimension(_viewportExtent);
       final bool didAcceptContentDimension = offset.applyContentDimensions(
         _minScrollExtent,
--- a/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
@@ -503,8 +503,16 @@ class _RenderSingleChildViewport extends RenderBox
       size = constraints.constrain(child!.size);
     }
 
-    offset.applyViewportDimension(_viewportExtent);
-    while (!offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent)) {}
+    while (true) {
+      final bool didAcceptViewportDimension = offset.applyViewportDimension(_viewportExtent);
+      final bool didAcceptContentDimension = offset.applyContentDimensions(
+        _minScrollExtent,
+        _maxScrollExtent,
+      );
+      if (didAcceptViewportDimension && didAcceptContentDimension) {
+        break;
+      }
+    }
   }
 
   Offset get _paintOffset => _paintOffsetForPosition(offset.pixels);
--- a/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
@@ -504,7 +504,6 @@ class _RenderSingleChildViewport extends RenderBox
     }
 
     offset.applyViewportDimension(_viewportExtent);
-    // offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent);
     while (!offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent)) {}
   }
 
--- a/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
@@ -400,7 +400,7 @@ class _RenderSingleChildViewport extends RenderBox
   }
 
   void _hasScrolled() {
-    markNeedsPaint();
+    markNeedsLayout();
     markNeedsSemanticsUpdate();
   }
 
@@ -503,16 +503,9 @@ class _RenderSingleChildViewport extends RenderBox
       size = constraints.constrain(child!.size);
     }
 
-    if (offset.hasPixels) {
-      if (offset.pixels > _maxScrollExtent) {
-        offset.correctBy(_maxScrollExtent - offset.pixels);
-      } else if (offset.pixels < _minScrollExtent) {
-        offset.correctBy(_minScrollExtent - offset.pixels);
-      }
-    }
-
     offset.applyViewportDimension(_viewportExtent);
-    offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent);
+    // offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent);
+    while (!offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent)) {}
   }
 
   Offset get _paintOffset => _paintOffsetForPosition(offset.pixels);
--- a/packages/flutter/test/widgets/scroll_notification_test.dart
+++ b/packages/flutter/test/widgets/scroll_notification_test.dart
@@ -45,13 +45,14 @@ void main() {
     await gesture.up();
     await tester.pump(const Duration(seconds: 1));
 
-    expect(events.length, 5);
+    expect(events.length, 6);
     // user scroll do not trigger the ScrollMetricsNotification.
     expect(events[0] is ScrollStartNotification, true);
     expect(events[1] is UserScrollNotification, true);
     expect(events[2] is ScrollUpdateNotification, true);
-    expect(events[3] is ScrollEndNotification, true);
-    expect(events[4] is UserScrollNotification, true);
+    expect(events[3] is ScrollMetricsNotification, true);
+    expect(events[4] is ScrollEndNotification, true);
+    expect(events[5] is UserScrollNotification, true);
 
     events.clear();
     // Change the content dimensions again.
@@ -222,10 +223,12 @@ void main() {
     ScrollNotification? notification;
 
     void handleNotification(ScrollNotification value) {
-      if (value is ScrollStartNotification ||
-          value is ScrollUpdateNotification ||
-          value is ScrollEndNotification) {
-        notification = value;
+      switch (value) {
+        case ScrollStartNotification():
+        case ScrollUpdateNotification(dragDetails: DragUpdateDetails _):
+        case ScrollEndNotification():
+          notification = value;
+        default:
       }
     }
 
--- a/packages/flutter/test/widgets/scroll_notification_test.dart
+++ b/packages/flutter/test/widgets/scroll_notification_test.dart
@@ -46,7 +46,6 @@ void main() {
     await tester.pump(const Duration(seconds: 1));
 
     expect(events.length, 6);
-    // user scroll do not trigger the ScrollMetricsNotification.
     expect(events[0] is ScrollStartNotification, true);
     expect(events[1] is UserScrollNotification, true);
     expect(events[2] is ScrollUpdateNotification, true);
--- a/packages/flutter/test/widgets/clamp_overscrolls_test.dart
+++ b/packages/flutter/test/widgets/clamp_overscrolls_test.dart
@@ -103,7 +103,8 @@ void main() {
       await tester.pumpWidget(buildFrame(physics, scrollController: scrollController));
       final ScrollableState scrollable = tester.state(find.byType(Scrollable));
 
-      // The initialScrollOffset will be corrected during the first frame.
+      expect(scrollable.position.pixels, equals(initialOffset));
+      await tester.pump(const Duration(seconds: 1)); // Allow overscroll to settle
       expect(scrollable.position.pixels, equals(expectedOffset));
     }
 
--- a/packages/flutter/test/widgets/scrollable_test.dart
+++ b/packages/flutter/test/widgets/scrollable_test.dart
@@ -1275,7 +1275,7 @@ void main() {
 
     // Make the outer constraints larger that the scrollable widget is no longer able to scroll.
     await tester.pumpWidget(build(300.0));
-    expect(controller.position.pixels, 0.0);
+    expect(controller.position.pixels, 100.0);
     expect(controller.position.maxScrollExtent, 0.0);
 
     // Hover over the scroll view and create a zero offset pointer scroll.
--- a/packages/flutter/test/widgets/single_child_scroll_view_test.dart
+++ b/packages/flutter/test/widgets/single_child_scroll_view_test.dart
@@ -1125,4 +1125,62 @@ void main() {
 
     expect(tester.testTextInput.isVisible, isFalse);
   });
+
+  testWidgets('', (WidgetTester tester) async {});
+
+  // Regression test for https://github.com/flutter/flutter/issues/173225
+  testWidgets('ValueListenableBuilder rebuild does not interrupt ongoing scroll gesture', (
+    WidgetTester tester,
+  ) async {
+    final ValueNotifier<String> titleValue = ValueNotifier<String>('0');
+    addTearDown(titleValue.dispose);
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Scaffold(
+          body: SingleChildScrollView(
+            physics: const AlwaysScrollableScrollPhysics(parent: BouncingScrollPhysics()),
+            child: ValueListenableBuilder<String>(
+              valueListenable: titleValue,
+              builder: (BuildContext context, String value, Widget? child) {
+                return Text(value);
+              },
+            ),
+          ),
+        ),
+      ),
+    );
+
+    final ScrollableState scrollableState = tester.state(find.byType(Scrollable));
+    expect(scrollableState.position.pixels, 0.0);
+
+    final TestGesture gesture = await tester.startGesture(tester.getCenter(find.byType(Text)));
+    await tester.pump();
+
+    // Simulate initial downward drag to create overscroll
+    await gesture.moveBy(const Offset(0.0, 100.0));
+    await tester.pump();
+
+    // Verify we're in overscroll territory
+    final double initialOffset = scrollableState.position.pixels;
+    expect(initialOffset, lessThan(0.0), reason: 'Should be in overscroll territory');
+
+    // Now simulate ValueListenableBuilder rebuild during the gesture
+    titleValue.value = '1';
+    await tester.pump();
+
+    // Verify the scroll position hasn't been reset to 0
+    expect(scrollableState.position.pixels, lessThan(0.0));
+    expect(scrollableState.position.pixels, equals(initialOffset));
+
+    // Continue the gesture to verify it's still active
+    await gesture.moveBy(const Offset(0.0, 50.0));
+    await tester.pump();
+
+    // Verify gesture continues to work
+    expect(scrollableState.position.pixels, lessThan(initialOffset));
+
+    await gesture.up();
+    await tester.pumpAndSettle();
+  });
 }
--- a/packages/flutter/test/widgets/single_child_scroll_view_test.dart
+++ b/packages/flutter/test/widgets/single_child_scroll_view_test.dart
@@ -1126,8 +1126,6 @@ void main() {
     expect(tester.testTextInput.isVisible, isFalse);
   });
 
-  testWidgets('', (WidgetTester tester) async {});
-
   // Regression test for https://github.com/flutter/flutter/issues/173225
   testWidgets('ValueListenableBuilder rebuild does not interrupt ongoing scroll gesture', (
     WidgetTester tester,


packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -222,10 +223,12 @@ void main() {
     ScrollNotification? notification;
 
     void handleNotification(ScrollNotification value) {
-      if (value is ScrollStartNotification ||
-          value is ScrollUpdateNotification ||
-          value is ScrollEndNotification) {
-        notification = value;
+      switch (value) {
+        case ScrollStartNotification():
+        case ScrollUpdateNotification(dragDetails: DragUpdateDetails _):
```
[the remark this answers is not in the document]
yiiim: Yes, this is intentional. Since scrolling now triggers a relayout, the `ScrollNotificationObserver` will send an additional `ScrollUpdateNotification` at [this location](https://github.com/flutter/flutter/blob/515e753c6e5575be1a882bd0be6c13a0daa2c006/packages/flutter/lib/src/widgets/scroll_position.dart#L675), with `dragDetails` set to null.

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
```diff
@@ -503,16 +503,8 @@ class _RenderSingleChildViewport extends RenderBox
...
-      if (offset.pixels > _maxScrollExtent) {
-        offset.correctBy(_maxScrollExtent - offset.pixels);
-      } else if (offset.pixels < _minScrollExtent) {
-        offset.correctBy(_minScrollExtent - offset.pixels);
-      }
-    }
-
     offset.applyViewportDimension(_viewportExtent);
-    offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent);
+    while (!offset.applyContentDimensions(_minScrollExtent, _maxScrollExtent)) {}
```
Piinks: Should `applyViewportDimension` also be in the loop? I am looking at comparing this to RenderShrinkWrappingViewport (similar to SingleChildScrollView), which breaks out of the loop when both are return true.

Comparing to RenderViewport I think is a bit different than expected here, which used maxLayoutCycles to break the loop.

Piinks (commented):
Apologies for the delay here while I was OOO. 🙏

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -807,10 +807,15 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
     // guard because the super call below would start a ballistic scroll activity.
     if (!_viewportDimensionWasNonZero || _needsPixelsCorrection) {
       _needsPixelsCorrection = false;
-      correctPixels(
-        tabBar._initialScrollOffset(viewportDimension, minScrollExtent, maxScrollExtent),
+      final double newPixels = tabBar._initialScrollOffset(
```
Piinks, line 809: This makes me wonder if this will be a breaking change. Why does a change in SingleChildScrollView affect _TabBarScrollPosition? Will other custom scroll positions be broken?

yiiim: What you said makes a lot of sense. If use `SingleChildScrollView` with a custom `ScrollPosition` and always override `applyContentDimensions` to return `false`, it will indeed result in an infinite loop. However, always returning `false` is probably an incorrect usage of `applyContentDimensions`. According to the documentation, returning `false` will cause it to be called multiple times. I don't think it's possible to prevent this situation with an assertion. Do we have any conventional practices to handle this?

Piinks: It would be hard to try to turn it into a soft break, let me kick off Google testing to see if it uncovers any breaks.

Piinks:
Pinging back to the top of my inbox to check on the Google testing results.

Piinks:
Following up on Google testing, it looks like this breaks a couple of things in subtle ways that might be hard for developers to diagnose and fix. There is at least one custom scroll position class in the google testing set that looks to have caused some test failures. Separately there are some screen shot tests that show that the scroll position is not the same with this change.
Do you have thoughts on options to make this a non-breaking change?

dkwingsmt:
@yiiim Hi! Are you able to address the comments above? (From triage)

yiiim:
> Following up on Google testing, it looks like this breaks a couple of things in subtle ways that might be hard for developers to diagnose and fix. There is at least one custom scroll position class in the google testing set that looks to have caused some test failures. Separately there are some screen shot tests that show that the scroll position is not the same with this change. Do you have thoughts on options to make this a non-breaking change?

Thank you for your feedback. After carefully reviewing the implementation, I couldn't find an alternative approach to achieve the desired functionality without introducing a breaking change. If you have any suggestions or ideas, I would be happy to explore them further.

packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -45,13 +45,14 @@ void main() {
     await gesture.up();
     await tester.pump(const Duration(seconds: 1));
 
-    expect(events.length, 5);
+    expect(events.length, 6);
     // user scroll do not trigger the ScrollMetricsNotification.
```
Piinks: This comment indicates the metrics notification should not be fired?

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
Piinks: This was marked resolved, but I am still curious about whether or not maxLayoutCycles should apply here? Otherwise we could get stuck in this loop, right?

yiiim: If we use a fixed maxLayoutCycles to break out of the loop, perhaps we can avoid breaking changes? Would you agree with this approach?

packages/flutter/test/widgets/scroll_notification_test.dart
yiiim: It should be removed.

Piinks: I don't think ScrollMetricsNotification is meant to trigger in this way. This could introduce more unexpected behavior. @chunhtai do you recall from your work on metrics notifications?

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
```diff
@@ -503,16 +503,16 @@ class _RenderSingleChildViewport extends RenderBox
       size = constraints.constrain(child!.size);
     }
 
-    if (offset.hasPixels) {
-      if (offset.pixels > _maxScrollExtent) {
-        offset.correctBy(_maxScrollExtent - offset.pixels);
-      } else if (offset.pixels < _minScrollExtent) {
-        offset.correctBy(_minScrollExtent - offset.pixels);
+    while (true) {
```
chunhtai: while (!offset.applyViewportDimension(_viewportExtent) || !offset.applyContentDimensions(
        _minScrollExtent,
        _maxScrollExtent,
      ))

chunhtai: why is this convert to a while loop? are we expect the offset needs to be call several time until it stablized?

yiiim: Yes, this is mentioned in the PR description.

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
yiiim: I have added maxLayoutCycles. Could you rerun the Google tests and try again?

Piinks:
> I have added maxLayoutCycles. Could you rerun the Google tests and try again?

Rebasing to rerun 👍

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
```diff
@@ -400,7 +400,7 @@ class _RenderSingleChildViewport extends RenderBox
   }
 
   void _hasScrolled() {
-    markNeedsPaint();
+    markNeedsLayout();
```
chunhtai, line 403: Why do mark layout dirty? we don't use scroll offset during layout at all.

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
```diff
@@ -502,17 +504,16 @@ class _RenderSingleChildViewport extends RenderBox
...
       size = constraints.constrain(child!.size);
     }
-
-    if (offset.hasPixels) {
-      if (offset.pixels > _maxScrollExtent) {
-        offset.correctBy(_maxScrollExtent - offset.pixels);
-      } else if (offset.pixels < _minScrollExtent) {
-        offset.correctBy(_minScrollExtent - offset.pixels);
+    for (int i = 0; i < _maxLayoutCycles; i++) {
+      final bool didAcceptViewportDimension = offset.applyViewportDimension(_viewportExtent);
```
chunhtai, line 508: I still don't understand why we want to call this in a loop, the _viewportExtent, _minScrollExtent, and _maxScrollExtent will only change when child relayout, but that is not the case (and also doesn't need to) in this loop.

chunhtai (changes requested):
The renderViewport needs to do a while loop to correct the offset because rendersliver can report offset adjustment because it is lazy list, this is not the case for singlechildscrollview because its child is a not a sliver.

It feels to me the problem is that RenderSingleChildViewport's offset correction doesn't consider the overscroll cases where the RenderViewport uses a math.max(0, -centerOffset) to "lie" about overscroll so that the sliver won't attempt to correct it.

In RenderSingleChildViewport case, I don't think we need to correct offset at all because it doesn't work with sliver. I think we can just remove the offset correction and leave the applyContent/ViewportDimesion part alone

also, please use re-request review if you want me to take another look. I sometimes forget to follow up with pr after review

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
yiiim: https://github.com/flutter/flutter/blob/66c68b66d965138ac3038c78d7a3dc57d52377cd/packages/flutter/lib/src/widgets/scroll_position.dart#L679 If this is not done, _lastMetrics will be incorrect after scrolling.

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
yiiim: Even if minScrollExtent and maxScrollExtent do not change, applyContentDimensions should still be called repeatedly. This is described in the documentation for applyContentDimensions, and it is necessary for correct behavior. When [this line](https://github.com/flutter/flutter/blob/66c68b66d965138ac3038c78d7a3dc57d52377cd/packages/flutter/lib/src/widgets/scroll_position.dart#L656) returns false, it needs to keep being called. Otherwise, issues like #105733 may occur.

yiiim:
As I understand it, the while loop for `applyContentDimensions` is intended to correct the offset when `minScrollExtent` or `maxScrollExtent` changes. This is not related to whether the child is a Sliver. The way `applyContentDimensions` works requires it to be called repeatedly.

chunhtai:
Yeah I think repeatedly calling applycontentDimesion makes sense now after I read more into the code. the return value is actually used to decide who should drive the scroll offset correction, return false means the performalayout should drive by repeatedly calls applycontentDimesion, so that part of the code LGTM now.

packages/flutter/lib/src/widgets/single_child_scroll_view.dart
```diff
@@ -493,6 +493,8 @@ class _RenderSingleChildViewport extends RenderBox
     return constraints.constrain(childSize);
   }
 
+  static const int _maxLayoutCycles = 10;
```
chunhtai, line 496:
```suggestion
  static const int _kMaxLayoutCycles = 10;
```

Also either apply this to all RenderViewport or ignore this check

packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -45,13 +45,13 @@ void main() {
...
 
-    expect(events.length, 5);
-    // user scroll do not trigger the ScrollMetricsNotification.
+    expect(events.length, 6);
     expect(events[0] is ScrollStartNotification, true);
     expect(events[1] is UserScrollNotification, true);
     expect(events[2] is ScrollUpdateNotification, true);
-    expect(events[3] is ScrollEndNotification, true);
-    expect(events[4] is UserScrollNotification, true);
+    expect(events[3] is ScrollMetricsNotification, true);
```
chunhtai, line 52: Is this just change to the pixel location?

chunhtai (commented):
we should still avoid relayout when scroll in singlechildscrollview

Piinks:

Piinks:

dipsy:
I ran the reviewer's suggestion — dropping the offset correction from
`RenderSingleChildViewport` altogether — against the framework tests on master,
in case the numbers help:

- Removing the correction and leaving `applyViewportDimension` /
  `applyContentDimensions` in place fails two tests:
  `ClampingScrollPhysics handles out of bounds ScrollPosition - initialScrollOffset`
  and `ScrollMetricsNotification test`.
- The first one is because the physics is never consulted on the first layout:
  `ScrollPosition.applyContentDimensions` only calls `correctForNewDimensions`
  when `haveDimensions` is already true, so an out-of-range `initialScrollOffset`
  has nothing to bring it into range.
- Clamping the offset there instead — inside `applyContentDimensions` on the
  first pass — fixes that one, but then `ScrollMetricsNotification test` sees two
  notifications where it expects one, the same contract change this PR already
  absorbs by moving its expectation from 5 to 6 events.

Separately, a narrower patch that skips the clamp only while
`ScrollPosition.isScrollingNotifier.value` is true — leaving it in place for
layouts outside a scroll — passes 1860 tests locally (`test/widgets/scroll*`,
`scrollable*`, `list_view*`, `nested_scroll_view`, `page_view`,
`single_child_scroll_view`, `clamp_overscrolls`, plus `material/tabs_test.dart`,
`dropdown_menu_test.dart` and `text_field_test.dart`) without editing any
existing expectation. It does not address the asymmetry between the box and
sliver viewports, so it is only interesting as a stopgap.

One more trigger worth having in the tests: hover. A row that reveals a trailing
icon on hover relayouts as the pointer crosses it, which is enough to reset an
overscroll from a trackpad flick; changing only the row's colour does not, since
a repaint never reaches `performLayout`. Regression test for that path, driving a
real trackpad pan/zoom gesture:

```dart
// Regression test for https://github.com/flutter/flutter/issues/145078
testWidgets('overscroll survives a relayout of the content', (WidgetTester tester) async {
  final controller = ScrollController();
  late StateSetter rebuildRow;
  var revealed = false;

  await tester.pumpWidget(
    Directionality(
      textDirection: TextDirection.ltr,
      child: Align(
        child: SizedBox(
          width: 300.0,
          height: 300.0,
          child: SingleChildScrollView(
            controller: controller,
            child: Column(
              children: <Widget>[
                StatefulBuilder(
                  builder: (BuildContext context, StateSetter setState) {
                    rebuildRow = setState;
                    // Stands in for a row revealing an action on hover: the rebuild has to lay
                    // the row out again, a repaint alone does not reach performLayout.
                    return Row(
                      children: <Widget>[
                        const Expanded(child: SizedBox(height: 32.0)),
                        if (revealed) const SizedBox(width: 16.0, height: 32.0),
                      ],
                    );
                  },
                ),
                for (var i = 1; i < 30; i++) const SizedBox(height: 32.0),
              ],
            ),
          ),
        ),
      ),
    ),
  );
  await tester.pumpAndSettle();

  // Drag past the leading edge with a trackpad gesture and hold it there.
  const center = Offset(400.0, 300.0);
  final TestGesture gesture = await tester.createGesture(kind: PointerDeviceKind.trackpad);
  await gesture.panZoomStart(center);
  var pan = 0.0;
  for (var i = 0; i < 8; i++) {
    pan += 12.0;
    await gesture.panZoomUpdate(center, pan: Offset(0.0, pan));
    await tester.pump(const Duration(milliseconds: 16));
  }
  final double overscrolled = controller.position.pixels;
  expect(overscrolled, lessThan(-20.0));

  // The row rebuilds mid-gesture, as it would when hover reveals an action.
  rebuildRow(() => revealed = true);
  await tester.pump(const Duration(milliseconds: 16));

  expect(controller.position.pixels, overscrolled);

  await gesture.panZoomEnd();
  await tester.pumpAndSettle();
  expect(controller.position.pixels, 0.0);
}, variant: TargetPlatformVariant.only(TargetPlatform.macOS));
```

It passes with the changes in this PR, and fails on master — please take it if it
is useful.

@yiiim are you still planning to come back to this? If not, I am happy to open a
PR with the narrow variant plus this test as an interim fix, or to help get this
one over the line — whichever the reviewers prefer.

yiiim:
@dipsy Thanks for the detailed investigation! Please feel free to open a PR with the narrow variant and the regression test.

Piinks:
If you decide to return to this, material_ui and cupertino_ui are now accepting contributions.
The portions of this PR that affect the material and cupertino libraries can be ported over to flutter/packages, although this one could be tricky since it would be a breaking change.

yiiim:
This PR will be replaced by #185196

PR #173862, opened 2025-08-15
[Animation] Add granular frame forcing to animations

### LINKED ISSUE #133533: In MacOS 12.6.1, when I create a new window using desktop_multi_window, the switching application cuts back to this window, and the newly created window does not display properly

- [x] I have searched the [existing issues](https://github.com/flutter/flutter/issues)

### Steps to reproduce

1. Use Mac mini / macOS Monterey 12.6.1
2. Run application and create new window
3. When I switch between applications and go back to the sub-window created by the application, the sub-window is directly unresponsive, and I catch any errors, and there are no errors
4. I tried other versions such as 3.7.7 and didn't have this problem

### Expected results

The create child window does not respond

### Actual results

You can click to operate, normal interaction

### Code sample

<details>

```dart
void main(List<String> args) {
  pdebug("@@@@@@@@@@@@ main start");
  if (args.isNotEmpty && args.first == 'multi_window') {
    final arguments = jsonDecode(args[2]);
    final key = arguments['key'];
    switch (key) {
      case 'setting':
        MultiWindowRunner.runMaterialApp(arguments, const SettingHome());
        break;
      case 'chat_history':
        MultiWindowRunner.runMaterialApp(
            arguments,
            ChatHistory(
              chatId: arguments['chat_id'],
              messageId: arguments['message_id'],
            ));
        break;
    }
    return;
  }

  profileDebug = showProfileDebug;
  runZonedGuarded(() async {
    WidgetsFlutterBinding.ensureInitialized();
    await FileDragAndDropChannel.instance.initializedMainView();
    objcetMgr.multiWindowMgr.windowInit();
    runApp(MyApp());
  }, (error, stackTrace) async {
    if (inProduction) {
      await postException(error, stackTrace, userID: objcetMgr.mainUser.id);
    }
  });
}



static void runMaterialApp(dynamic arguments, Widget child) {
    BotToastNavigatorObserver botToastNavigatorObserver =
        BotToastNavigatorObserver();
    pageMgr.register();
    runApp(
      MaterialApp(
        localizationsDelegates: const [
          GlobalMaterialLocalizations.delegate,
          GlobalWidgetsLocalizations.delegate,
          GlobalCupertinoLocalizations.delegate,
        ],
        supportedLocales: const [Locale('zh', 'CN'), Locale('en')],
        builder: (context, widget) {
          return GestureDetector(
            onTap: () {
              FocusScopeNode currentFocus = FocusScope.of(context);
              if (!currentFocus.hasPrimaryFocus &&
                  currentFocus.focusedChild != null) {
                FocusManager.instance.primaryFocus?.unfocus();
              }
            },
            child: MediaQuery(
                data: MediaQuery.of(context).copyWith(textScaleFactor: 1.0),
                child: BotToastInit()(context, widget)),
          );
        },
        title: appName,
        navigatorObservers: [
          botToastNavigatorObserver,
        ],
        home: ThemeContainer(mainUser: arguments['mainUser'], child: child),
        theme: themeData(),
      ),
    );
  }
```
</details>

### Screenshots or Video

<details>

[Upload media here]

</details>

### Logs

no logs

### Flutter Doctor output
<details>
<summary> details</summary>

```
[!] Flutter (Channel unknown, 3.7.9, on macOS 12.6.1 21G217 darwin-arm64, locale zh-Hans-CN)
    ! Flutter version 3.7.9 on channel unknown at /Users/melody/Desktop/flutter
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
    ! Unknown upstream repository.
      Reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
[!] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more details.
[✓] Xcode - develop for iOS and macOS (Xcode 14.1)
[✓] Android Studio (version 2021.3)
[✓] VS Code (version 1.81.1)
[✓] HTTP Host Availability

```
</details>

### LINKED ISSUE #174356: [Desktop] Allow applications to continue frame rendering when hidden for real-time games and animation-driven audio
### Use case

We are building a real-time multiplayer game that uses Rive animations with audio event synchronization. When players perform actions remotely, the game continues running in the background, and animations must maintain their timeline to ensure audio plays at the correct positions. However, when the app window is hidden (e.g., minimized, covered by other windows, or in mission control on macOS), frame rendering stops, causing animations to freeze and audio to not play.

#### Proposed Solution
Add a `forceFramesEnabled` property to `SchedulerBinding` that allows developers to explicitly opt-in to continuous frame rendering when the app is hidden. This would be particularly useful for:

- Real-time games that continue running in the background
- Applications with animation-driven audio (like Rive)
- Multiplayer applications where game logic continues server-side

I have already implemented this feature in PR #173862.

#### Current Workaround
We've implemented a hacky solution by intercepting the `AppLifecycleState.hidden` state and replacing it with `AppLifecycleState.inactive` in a custom `DesktopWidgetsFlutterBinding`:

```dart
@override
void handleAppLifecycleStateChanged(AppLifecycleState state) {
  if (state == AppLifecycleState.hidden) {
    _isHidden = true;
    super.handleAppLifecycleStateChanged(AppLifecycleState.inactive);
  } else {
    if (state == AppLifecycleState.resumed) {
      _isHidden = false;
    }
    super.handleAppLifecycleStateChanged(state);
  }
}
```

tguerin:
This change adds opt-in, subtree-scoped frame forcing for Flutter animations as described in https://github.com/flutter/flutter/issues/174356#issuecomment-3287185284

Fixes https://github.com/flutter/flutter/issues/174356

This change could help fix #133533. 

## Pre-launch Checklist

- [x] I signed the [CLA].


justinmc:
We should make sure we are very certain that we want to support this use case before we add something like this due to the ability to shoot yourself in the foot here. @loic-sharma do you know if this is a valid concern for multiwindow/desktop apps?

loic-sharma:
> @loic-sharma do you know if this is a valid concern for multiwindow/desktop apps?

I don't know, however, I suspect that there's a better fix for https://github.com/flutter/flutter/issues/133533.

@chinmaygarde Would you have any thoughts on this change for this use case:

> Some desktop applications, such as games or apps with animation-driven audio (e.g., Rive) rely on continuous frame rendering to maintain smooth animation timing and audio playback.
> The current lifecycle behavior stops scheduling frames when the app is hidden or paused (e.g., mission control on MacOS), which can interrupt these use cases.

tguerin:
The fix was not originally made for this specific issue, but to address a problem we’re facing in our app. We are building a Desktop game, and when the window is completely covered, frame rendering stops, which prevents our animations from running.

To work around this, we currently have to intercept the hidden state and replace it with inactive, which is really hacky. The key requirement for us is to keep animations running even when the window is not visible.

I’m open to different approaches to achieve this, as long as we can meet that requirement.

dkwingsmt:
@tguerin Can you explain why you'd like the animation to keep running at background? What is broken by missing these animation frames? It would be great if you can open a new issue for this so we can discuss it and track it more clearly.

tguerin:
Here is the issue https://github.com/flutter/flutter/issues/174356

Piinks:
Does this comment in the linked issue provide what is needed here? https://github.com/flutter/flutter/issues/174356#issuecomment-3287185284

tguerin:
I will rework the PR to include the suggested approach.

tguerin:
@loic-sharma I managed to find some time to implement the suggested changes. However I still have one failing test, seems that the rendering_tester used by the linux_customer_testing still uses the version that doesn't have my modifications. How can i fix this? Is there some kind of dependency to bump?

Piinks:
> However I still have one failing test, seems that the rendering_tester used by the linux_customer_testing still uses the version that doesn't have my modifications. How can i fix this? Is there some kind of dependency to bump?

This means that as it currently stands, this is a breaking change. Our policy is documented here: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes

loic-sharma:
@tguerin That test failed because DevTools has a class that implements `Ticker`. We either need to update this class to add the new `forceFrames` member you added, or we need to remove this class (it appears to be unused code). I sent a PR to remove the class: https://github.com/flutter/devtools/pull/9459. Once that lands, it should unblock this PR.

packages/flutter/lib/src/scheduler/ticker.dart
```diff
@@ -282,6 +290,17 @@ class Ticker {
...
+      // Ensure a forced engine frame is scheduled, then register our callback
+      // without scheduling a new frame again here.
+      SchedulerBinding.instance.scheduleForcedFrame();
+      _animationId = SchedulerBinding.instance.scheduleFrameCallback(
+        _tick,
+        rescheduling: rescheduling,
+        scheduleNewFrame: false,
+      );
+      return;
+    }
```
loic-sharma, line 297: What do you think of something like this instead?

```dart
if (forceFrames) {
  ScheduleBinding.instance.scheduleForcedFrame();
} else {
  SchedulerBinding.instance.scheduleFrame();
}

_animationId = SchedulerBinding.instance.scheduleFrameCallback(
  _tick,
  rescheduling: rescheduling,
  scheduleNewFrame: false,
);
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -38,6 +44,13 @@ class TickerMode extends StatefulWidget {
   /// they just don't call their callbacks. Time still elapses.
   final bool enabled;
 
+  /// If true, tickers in this subtree will request frames using
+  /// SchedulerBinding.scheduleForcedFrame while active.
+  ///
+  /// Use sparingly, as this can increase battery usage when frames would
+  /// otherwise be suppressed by lifecycle.
```
loic-sharma: I think it'd be better to slightly explain what forced frames are here by reusing the docs on [`SchedulerBinding.scheduleForcedFrame`](https://api.flutter.dev/flutter/scheduler/SchedulerBinding/scheduleForcedFrame.html). 

Consider:

```suggestion
  /// If true, tickers in this subtree will force frames even if
  /// frames would normally not be scheduled (e.g. even if the
  /// device's screen is turned off).
  ///
  /// Use sparingly as this will cause significantly higher battery
  /// usage when the device should be idle.
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +109,73 @@ class TickerMode extends StatefulWidget {
...
+    'This feature was deprecated after v3.35.0-0.0.pre.',
+  )
   static ValueListenable<bool> getNotifier(BuildContext context) {
     final _EffectiveTickerMode? widget = context
         .getInheritedWidgetOfExactType<_EffectiveTickerMode>();
     return widget?.notifier ?? const _ConstantValueListenable<bool>(true);
   }
 
+  /// Returns the effective ticker mode values for this subtree and establishes
+  /// a dependency on the surrounding [TickerMode].
```
loic-sharma: Let's keep more of the docs from the `of` method:

```suggestion
  /// Returns the requested ticker mode values for this subtree and establishes
  /// a dependency on the ancestor [TickerMode], if any.  
  ///
  /// This is used automatically by [TickerProviderStateMixin] and
  /// [SingleTickerProviderStateMixin] to decide if their tickers should be
  /// enabled or disabled.
  ///
  /// In the absence of a [TickerMode] widget, this defaults to enabled
  /// tickers that don't force frames.
  ///
  /// Typical usage is as follows:
  ///
  /// ```dart
  /// bool tickingEnabled = TickerMode.valuesOf(context).enabled;
  /// ```
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +109,73 @@ class TickerMode extends StatefulWidget {
...
+  static TickerModeData valuesOf(BuildContext context) {
+    final _EffectiveTickerMode? widget = context
+        .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+    return widget?.values ?? const TickerModeData(enabled: true, forceFrames: false);
+  }
+
+  /// Obtains a [ValueListenable] of [TickerModeData] from the surrounding [TickerMode]
+  /// without establishing a dependency.
```
loic-sharma: Let's keep more of the docs from `getNotifier` method:

```suggestion
  /// Obtains a [ValueListenable] from the [TickerMode] surrounding the `context`,
  /// which indicates whether tickers are enabled in the given subtree.
  ///
  /// When that [TickerMode] enabled or disabled tickers, the listenable notifies
  /// its listeners.
  ///
  /// While the [ValueListenable] is stable for the lifetime of the surrounding
  /// [TickerMode], calling this method does not establish a dependency between
  /// the `context` and the [TickerMode] and the widget owning the `context`
  /// does not rebuild when the ticker mode data changes. This is preferable
  /// when the ticker mode does not impact what is currently rendered on screen,
  /// e.g. because it is only used to mute/unmute a [Ticker]. Since no dependency
  /// is established, the widget owning the `context` is also not informed when
  /// it is moved to a new location in the tree where it may have a different
  /// [TickerMode] ancestor. When this happens, the widget must manually
  /// unsubscribe from the old listenable, obtain a new one from the new ancestor
  /// [TickerMode] by calling this method again, and re-subscribe to it
  /// [StatefulWidget]s can, for example, do this in [State.activate],
  /// which is called after the widget has been moved to a new location.
  ///
  /// Alternatively, [of] can be used instead of this method to create a
  /// dependency between the provided `context` and the ancestor [TickerMode].
  /// In this case, the widget automatically rebuilds when the ticker mode
  /// changes or when it is moved to a new [TickerMode] ancestor, which
  /// simplifies the management cost in the widget at the expensive of some
  /// potential unnecessary rebuilds.
  ///
  /// In the absence of a [TickerMode] widget, this function returns a
  /// [ValueListenable], whose [ValueListenable.value] is
  /// [TickerModeData.fallback].
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +109,73 @@ class TickerMode extends StatefulWidget {
...
         .getInheritedWidgetOfExactType<_EffectiveTickerMode>();
     return widget?.notifier ?? const _ConstantValueListenable<bool>(true);
   }
 
+  /// Returns the effective ticker mode values for this subtree and establishes
+  /// a dependency on the surrounding [TickerMode].
+  static TickerModeData valuesOf(BuildContext context) {
+    final _EffectiveTickerMode? widget = context
+        .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+    return widget?.values ?? const TickerModeData(enabled: true, forceFrames: false);
```
loic-sharma, line 127: `DefaultTextStyle` also has a similar behavior where it defaults to sensible values, [`DefaultTextStyle.fallback`](https://api.flutter.dev/flutter/widgets/DefaultTextStyle/DefaultTextStyle.fallback.html), if there's no ancestor `DefaultTextStyle`.

Since both `valuesOf` and `getValuesNotifier` has similar fallback logic, let's also introduce a `TickerModeData.fallback` member.

loic-sharma: `DefaultTextStyle` also has a similar behavior where it defaults to sensible values, [`DefaultTextStyle.fallback`](https://api.flutter.dev/flutter/widgets/DefaultTextStyle/DefaultTextStyle.fallback.html), if there's no ancestor `DefaultTextStyle`.

Since both `valuesOf` and `getValuesNotifier` has similar fallback logic, what do you think of also introducing a `TickerModeData.fallback` member?

```suggestion
    return widget?.values ?? TickerModeData.fallback;
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +109,73 @@ class TickerMode extends StatefulWidget {
...
+  ///
+  /// The given `enabled` and `forceFrames` override the ambient values when not null;
+  /// otherwise the ambient values are preserved.
+  static Widget merge({Key? key, bool? enabled, bool? forceFrames, required Widget child}) {
+    return Builder(
+      builder: (BuildContext context) {
+        final _EffectiveTickerMode? parent = context
+            .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+        final bool parentEnabled = parent?.enabled ?? true;
+        final bool parentForce = parent?.forceFrames ?? false;
```
loic-sharma: If you do add `TickerModeData.fallback`, let's use it here as well:

```suggestion
        final bool parentEnabled = parent?.enabled ?? TickerModeData.fallback.enabled;
        final bool parentForce = parent?.forceFrames ?? TickerModeData.fallback.forceFrames;
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +109,73 @@ class TickerMode extends StatefulWidget {
...
+    const TickerModeData(enabled: true, forceFrames: false),
+  );
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _ancestorTicketMode = TickerMode.of(context);
+    final _EffectiveTickerMode? parent = context
+        .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+    _ancestorTicketMode = parent?.enabled ?? true;
```
loic-sharma: Could you rename this to `_ancestorTickerMode`? Looks like we made a typo here 😅

tguerin:
The customer testing is still failing, do i need to wait a release of the dev tools?

loic-sharma:
@tguerin Ah sorry, it looks like I have to roll DevTools into customer tests, and then roll customer tests into Flutter. Apologies for the delay, this is my first time doing this process.

I've started the DevTools to customer tests roll here: https://github.com/flutter/tests/pull/473

packages/flutter/lib/src/scheduler/ticker.dart
```diff
@@ -91,6 +91,14 @@ class Ticker {
 
   TickerFuture? _future;
 
+  /// If true, this ticker will request frames using
+  /// SchedulerBinding.scheduleForcedFrame instead of scheduleFrame.
```
loic-sharma:
```suggestion
  /// [SchedulerBinding.scheduleForcedFrame] instead of
  /// [SchedulerBinding.scheduleFrame].
```

packages/flutter/lib/src/scheduler/ticker.dart
```diff
@@ -91,6 +91,14 @@ class Ticker {
...
+  ///
+  /// This allows granular control to advance frames even when frames are
+  /// typically disabled (e.g. when the app is in the background) and should be
```
loic-sharma: This sentence is a little long. Consider splitting to improve readability:

```suggestion
  /// typically disabled (e.g. when the app is in the background). This should be
```

I don't feel strongly, feel free to skip :)

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +110,114 @@ class TickerMode extends StatefulWidget {
...
+  ///
+  /// The given `enabled` and `forceFrames` override the ambient values when not null;
+  /// otherwise the ambient values are preserved.
+  static Widget merge({Key? key, bool? enabled, bool? forceFrames, required Widget child}) {
+    return Builder(
+      builder: (BuildContext context) {
+        final _EffectiveTickerMode? parent = context
+            .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+        final bool parentEnabled = parent?.enabled ?? true;
+        final bool parentForce = parent?.forceFrames ?? false;
```
loic-sharma: Consider using `TickerModeData.fallback` here as well:

```suggestion
        final bool parentEnabled = parent?.enabled ?? TickerModeData.fallback.enabled;
        final bool parentForce = parent?.forceFrames ?? TickerModeData.fallback.forceFrames;
```

This would ensure `merge`'s logic doesn't get stale if we change the default values for whatever reason.

(The Dart formatting of this suggestion is likely wrong)

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -442,3 +579,61 @@ class _ConstantValueListenable<T> implements ValueListenable<T> {
...
+/// - [forceFrames]: When true, tickers in the subtree request frames using
+///   [SchedulerBinding.scheduleForcedFrame] while active. This value is
+///   combined across ancestors using logical OR, so any ancestor requesting
+///   forced frames enables it for the subtree.
+///
+/// For most widgets, reading these values is unnecessary; mixins such as
+/// [SingleTickerProviderStateMixin] and [TickerProviderStateMixin] apply them
+/// automatically to the [Ticker]s they vend. Use this class when you need to
+/// observe or react to ticker policy explicitly.
+class TickerModeData {
```
loic-sharma, line 606:
```suggestion
@immutable
class TickerModeData {
```

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -442,3 +579,61 @@ class _ConstantValueListenable<T> implements ValueListenable<T> {
...
+  /// [TickerMode.enabled] values.
+  final bool enabled;
+
+  /// Whether tickers should request forced frames while active.
+  ///
+  /// Effective value is the logical OR of all ancestor and local
+  /// [TickerMode.forceFrames] values. Forcing frames may increase battery
+  /// usage, so use sparingly.
+  final bool forceFrames;
+}
```
loic-sharma, line 641: Could you implement `operator ==` and `hashCode` for this type? This will allow  `ValueNotifier` to skip notifying listeners if its `value` is updated to a new `TickerModeData` that has the same values as the previous one.

loic-sharma: See: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#common-boilerplates-for-operator--and-hashcode

packages/flutter/test/scheduler/binding_test.dart
```diff
@@ -14,6 +14,16 @@ void main() {
...
 
+  test('Ticker.forceFrames requests forced frames', () async {
+    final Ticker t = Ticker((_) {});
+    t.forceFrames = true;
+    final TickerFuture f = t.start();
+    // A forced frame should be scheduled even if frames are otherwise disabled.
+    expect(SchedulerBinding.instance.hasScheduledFrame, isTrue);
+    t.stop();
+    await f;
+  });
```
loic-sharma, line 27: Could you also add more tests to `packages/flutter/test/widgets/ticker_mode_test.dart`, or update existing tests to cover the new force frame value?

loic-sharma: Thanks for adding these!

loic-sharma:
@tguerin Thanks for addressing all the feedback! It looks like you'll need to run `dart format` to pass the `Linux analyze` check.

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +110,114 @@ class TickerMode extends StatefulWidget {
...
+    final _EffectiveTickerMode? widget = context
+        .dependOnInheritedWidgetOfExactType<_EffectiveTickerMode>();
+    return widget?.values ?? TickerModeData.fallback;
+  }
+
+  /// Obtains a [ValueListenable] from the [TickerMode] surrounding the `context`,
+  /// which indicates whether tickers are enabled in the given subtree.
+  ///
+  /// When that [TickerMode] enabled or disabled tickers, the listenable notifies
+  /// its listeners.
```
justinmc, line 148: I think "enables or disables" would make more sense if I'm reading it right?

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -92,24 +110,114 @@ class TickerMode extends StatefulWidget {
...
+  /// [TickerMode] by calling this method again, and re-subscribe to it
+  /// [StatefulWidget]s can, for example, do this in [State.activate],
+  /// which is called after the widget has been moved to a new location.
+  ///
+  /// Alternatively, [of] can be used instead of this method to create a
+  /// dependency between the provided `context` and the ancestor [TickerMode].
+  /// In this case, the widget automatically rebuilds when the ticker mode
+  /// changes or when it is moved to a new [TickerMode] ancestor, which
+  /// simplifies the management cost in the widget at the expensive of some
+  /// potential unnecessary rebuilds.
```
justinmc, line 169: "at the expensive" => "at the expense"

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -442,3 +579,78 @@ class _ConstantValueListenable<T> implements ValueListenable<T> {
...
+///   forced frames enables it for the subtree.
+///
+/// For most widgets, reading these values is unnecessary; mixins such as
+/// [SingleTickerProviderStateMixin] and [TickerProviderStateMixin] apply them
+/// automatically to the [Ticker]s they vend. Use this class when you need to
+/// observe or react to ticker policy explicitly.
+@immutable
+class TickerModeData {
+  /// Creates a [TickerModeData].
+  const TickerModeData({required this.enabled, required this.forceFrames});
```
justinmc, line 608: Would someone ever set enabled false and forceFrames true?

packages/flutter/test/scheduler/binding_test.dart
```diff
@@ -14,6 +14,16 @@ void main() {
     expect(SchedulerBinding.instance.platformDispatcher.onBeginFrame, isNotNull);
   });
 
+  test('Ticker.forceFrames requests forced frames', () async {
+    final Ticker t = Ticker((_) {});
+    t.forceFrames = true;
+    final TickerFuture f = t.start();
+    // A forced frame should be scheduled even if frames are otherwise disabled.
+    expect(SchedulerBinding.instance.hasScheduledFrame, isTrue);
+    t.stop();
```
justinmc: Should this be done in an addTearDown so it's still stopped even if the previous expectation fails?

justinmc (commented):
@tguerin Thanks for your persistence on this and creating the issue and arguing your side of things. I'm glad to see this is just about to a state where everyone is happy. Just a few small comments below for me.

We should probably run the Google tests on this even though nothing appears to be breaking anymore.

packages/flutter/lib/src/widgets/ticker_provider.dart
loic-sharma: Yes, that's a legitimate use case. For example, let's say the subtree drives game audio using the animation system. I want this to run even if the app is backgrounded. I start with `TickerMode(enabled: true, forceFrames: true)`. Then, the player pauses the game. I want to pause the music. I update to `TickerMode(enabled: false, forceFrames: true)`.

packages/flutter/lib/src/widgets/ticker_provider.dart
tguerin: Indeed, so obvious I should have seen that, my bad!

tguerin:
> This is looking really good! Thanks for the wonderful contribution :)

Thank you for helping me land my first PR, really appreciate.

packages/flutter/test/scheduler/binding_test.dart
tguerin: Indeed It's safer and will avoid to impact other tests in case of failure

packages/flutter/lib/src/widgets/ticker_provider.dart
tguerin: "enables or disables" definitely sounds better

tguerin:
@loic-sharma @justinmc Thanks for all the discussion, this approach is definitely much cleaner than the first one I implemented, and it was interesting to dive into the Ticker code.

I’ll make sure to be more thorough next time to avoid some of the oversights in this PR. I was on holiday and trying to squeeze in time to address feedback, which led to a few things slipping through that I'd normally catch. Really appreciate everyone's patience and reviews!

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -87,29 +105,119 @@ class TickerMode extends StatefulWidget {
...
+  /// [TickerMode], calling this method does not establish a dependency between
+  /// the `context` and the [TickerMode] and the widget owning the `context`
+  /// does not rebuild when the ticker mode data changes. This is preferable
+  /// when the ticker mode does not impact what is currently rendered on screen,
+  /// e.g. because it is only used to mute/unmute a [Ticker]. Since no dependency
+  /// is established, the widget owning the `context` is also not informed when
+  /// it is moved to a new location in the tree where it may have a different
+  /// [TickerMode] ancestor. When this happens, the widget must manually
+  /// unsubscribe from the old listenable, obtain a new one from the new ancestor
+  /// [TickerMode] by calling this method again, and re-subscribe to it
```
justinmc: Missing a period at the end I think.

packages/flutter/lib/src/widgets/ticker_provider.dart
```diff
@@ -87,29 +105,119 @@ class TickerMode extends StatefulWidget {
...
+        final bool parentForce = parent?.forceFrames ?? TickerModeData.fallback.forceFrames;
+        return TickerMode(
+          key: key,
+          enabled: enabled ?? parentEnabled,
+          forceFrames: forceFrames ?? parentForce,
+          child: child,
+        );
+      },
+    );
+  }
```
justinmc, line 200: Nit: I wonder if this whole method would be better off as its own widget?

justinmc (approved):
LGTM 👍

packages/flutter/lib/src/widgets/ticker_provider.dart
tguerin: Good question! I followed the pattern here, which uses a static factory method rather than a separate widget. Since `TickerMode.merge` is just lightweight syntactic sugar (no custom state or special optimization) the static method keeps the API surface simpler.

loic-sharma (approved):
Thank you for the wonderful contribution! This is spectacular work 🎉

PR #173863, opened 2025-08-15
[ Tool ] Cleanup widget preview and frontend server shutdown

bkonyi:
Prevents the frontend server from throwing ToolExit with the "The Dart compiler exited unexpectedly." on shutdown via a signal.


matanlurey (approved):
Seems like code to me.

Just kidding - this looks good, thanks for the fix!

PR #173569, opened 2025-08-11
[ios][tools]do not log "bonjour not found" at all (unless verbose)

hellohuanlin:
The previous PR https://github.com/flutter/flutter/pull/172913 avoids compilation error by not having the "error: " prefix. However, it's still very confusing to have "key not found" printed out in the terminal (either stderr or stdout), despite that it doesn't cause compilation error anymore. 

This PR introduces a new flag "skipErrorLog" which skips logging the `stderr` if set to true. We set it for plist extraction, because "not having bonjour key" should be one of the expected "normal" states

<s> We don't want to pipe it to `stdout` either, because it would be confusing too. I figured people would still file issue if they see it in terminal, even if it's a completely normal state. But not a strong opinion, so let me know if you disagree. </s>

However, under verbose mode, we would pipe it to `stdout`. This will be consistent with our previous behavior before macOS 26.

*List which issues are fixed by this PR. You must list at least one issue. An issue is not required if the PR fixes something trivial like a typo.*

Fixes https://github.com/flutter/flutter/issues/172627

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[...]

PR #173879, opened 2025-08-15
Revert "[ios][tools]do not log "bonjour not found" at all (unless verbose)"

hellohuanlin:
Reverts flutter/flutter#173569

Sorry I have to revert this one as it didn't work. 

# Why didn't work

From the comment, it looks like `echoError` is only used for verbose mode, and `streamOutput` is used for non-verbose mode (I am still looking into how this is setup)

https://github.com/flutter/flutter/blob/e4f27cd09734db6c6ed94e104ab5333c48dfc544/packages/flutter_tools/bin/xcode_backend.dart#L151

# How to fix

## Possible Fix 1

I tested this works, by changing this line: 

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/bin/xcode_backend.dart#L159

into this: 

```
if (!verbose && exitCode == 0 && !skipErrorLog)
```

## Possible fix 2: 

Surprisingly this also works: 
```
if (!verbose && exitCode == 0) {
  streamOutput(errorOutput.string().replaceAll('error', '');
}
```

Although we made sure `errorOutput` doesn't start with `error:`, there's an `error:` in the middle of the string. 

Possibly because of this code in `mac.dart`: 
https://github.com/flutter/flutter/blob/e4f27cd09734db6c6ed94e104ab5333c48dfc544/packages/flutter_tools/lib/src/ios/mac.dart#L431-L433

# Why we missed it? 

I tested my initial code code and assumed updated code would work, because I incorrectly thought `echoError` was the one triggered the output in non-verbose mode. I should have verified it.

### DIFF
--- a/packages/flutter_tools/bin/xcode_backend.dart
+++ b/packages/flutter_tools/bin/xcode_backend.dart
@@ -102,14 +102,7 @@ class Context {
 
   Directory directoryFromPath(String path) => Directory(path);
 
-  /// Run given command ([bin]) in a synchronous subprocess.
-  ///
-  /// If [allowFail] is true, an exception will not be thrown even if the process returns a
-  /// non-zero exit code. Also, `error:` will not be prefixed to the output to prevent Xcode
-  /// complication failures.
-  ///
-  /// If [skipErrorLog] is true, `stderr` from the process will not be output unless in [verbose]
-  /// mode. If in [verbose], pipes `stderr` to `stdout`.
+  /// Run given command in a synchronous subprocess.
   ///
   /// Will throw [Exception] if the exit code is not 0.
   ProcessResult runSync(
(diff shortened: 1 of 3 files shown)


hellohuanlin:
> From the comment, it looks like echoError is only used for verbose mode, and streamOutput is used for non-verbose mode (I am still looking into how this is setup)

I think I get it now. 

For `echoError`: `mac.dart` calls `_DefaultProcessUtils::run`, which does `_logger.printTrace(runResult)`. So in echoError is only print out in verbose mode! 

For `streamOutput`: `mac.dart::listenToScriptOutputLine` loops through lines like this: 

```
if !verbose {
  only print lines containing "error:" and "warning:"
} else {
  print line
}
```


This is so confusing...

I think I will go with the fix 1 in the PR description, since fix 2's call path is a bit convoluted

jmagman (approved):
Clean revert LGTM.
@hellohuanlin can you file a new issue or open the old one and document your findings and suggested fixes?

hellohuanlin:
> @hellohuanlin can you file a new issue or open the old one and document your findings and suggested fixes?

re-opened the old issue.

PR #162908, opened 2025-02-07
[Android] Remove overlay when platform views are removed from screen.

jonahwilliams:
When there are no more platform views, make sure the overlay layer is hidden. When a platform view is added again, show the overlay. The show/hide allows us to avoid continually recreating and destroying the overlay surface + swapchain when a platform view slides in and out of frame.

To further reduce memory usage, we could do a delayed de-allocation of the overlay layer (say after 50 frames of no platform view, destroy it). But I'm leaving this to a follow up.

PR #173881, opened 2025-08-15
[HCPP] Clean up overlay layer when last frame had overlay content and current doesn't

### LINKED ISSUE #173635: [HCPP] Interactive Media Ads plugin example app does not work with HCPP
Non HCPP works fine:

https://github.com/user-attachments/assets/629fa98e-3512-4839-9f4d-5596b5655870

HCPP freezes when it should display the midroll (not preroll) at around 23 seconds:

https://github.com/user-attachments/assets/687fa567-afd8-4ca4-9b34-5d980efc8d3e

The video being frozen is likely not the problem - the IMA plugin intentionally pauses the video. But we aren't seeing the content that should now be overlayed on top (the ad). Also of note, backgrounding the app and then bringing it back in to focus resets the PV to the working state, with the midroll displaying as it should.

gmackall:
Fixes https://github.com/flutter/flutter/issues/173635

My current understanding of the issue is that we are never pushing a new empty frame for this layer, so the texture gets stuck as shown in the bug. If we keep track of when we had content last frame, and don't in the current frame, we can push a single empty frame for that layer when needed.

See example in https://github.com/flutter/flutter/issues/173635#issuecomment-3192787040

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -149,6 +149,17 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
     overlay_frame->Submit();
+    has_overlay_layers_ = true;
+  } else {
+    if (has_overlay_layers_) {
+      // Submit one frame to clean up the outstanding overlay content.
+      std::shared_ptr<OverlayLayer> layer = surface_pool_->GetLayer(
+          context, android_context_, jni_facade_, surface_factory_);
+      overlay_frame = layer->surface->AcquireFrame(frame_size_);
+      overlay_frame->set_submit_info({.frame_boundary = false});
+      overlay_frame->Submit();
```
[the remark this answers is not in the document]
gmackall: I considered doing this, it works without it but also if the goal is to clear it I don't mind doing it 🤷

gmackall:
Driver test with the change, texture and no_texture:

hybrid_composition_pp_platform_view texture

hybrid_composition_pp_platform_view no_texture

Without change:

hybrid_composition_pp_platform_view texture
hybrid_composition_pp_platform_view no_texture

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -149,6 +149,18 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
+    has_overlay_layers_ = true;
+  } else {
+    if (has_overlay_layers_) {
+      // Submit one frame to clean up the outstanding overlay content.
+      std::shared_ptr<OverlayLayer> layer = surface_pool_->GetLayer(
+          context, android_context_, jni_facade_, surface_factory_);
+      overlay_frame = layer->surface->AcquireFrame(frame_size_);
+      overlay_frame->Canvas()->Clear(flutter::DlColor::kTransparent());
+      overlay_frame->set_submit_info({.frame_boundary = false});
+      overlay_frame->Submit();
```
reidbaker: Does this submit need to be here since there is a submit outside the if statment? Are there costs to a double submit frame?

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -149,6 +149,18 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
+      has_overlay_layers_ = false;
```
reidbaker: Does this need to happen after the submit? If so can you add a comment explaining why? if not why do work after submit?

reidbaker:
@flar I added you as someone with expertise in the engine. If you are not the right person can you add the right person and remove yourself? For example, I don't know if there is a better way to clear the state without submitting a frame.

reidbaker (approved):
@gmackall Read back through the driver and app code and see if there are comments you could add that would help an author or review of future prs that handle framework migrations know if some future the change would break the value of this test. 

For example the screenshots you added in the pr were super helpful.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: That is a different frame that we submit below, so yes I believe we need to

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: I think it's fine for it to be anywhere in the if statement - it won't matter until the next frame, we just need to keep track of the state frame by frame. 

Only thing is if the frame submission in the line above can fail in a way that would prevent us getting to this line, it's better to have it in this order. But I don't know really know that flow 🤷

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
flar: It feels like we should experience the removal of an overlay (in this case does it really mean "the existence of a prior platform layer"?) more proactively, but I don't understand what happens when the overlays come and go.

Is this mechanism unique in having an internal structure that comes and goes when platforms appear and disappear?

Also, is there somewhere else that is responsible for whether or not the overlay layer is being displayed that we could "turn off"? That might be more efficient than just erasing it.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.h
```diff
@@ -112,6 +112,9 @@ class AndroidExternalViewEmbedder2 final : public ExternalViewEmbedder {
   // The task runners.
   const TaskRunners task_runners_;
 
+  // If there is currently content being displayed in the overlay layer.
+  bool has_overlay_layers_ = false;
```
flar: Nit - would "overlay_layer_has_content_" be a better name?

flar (approved):
Some suggestions and nits...

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.h
gmackall: changed to suggested

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: I agree that I did not expect to need to handle the removal of the overlay at this layer. For example, in the old PV code:
https://github.com/flutter/flutter/blob/2265d94c6b0bbbf22193852e52e279e8c001d385/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder.cc#L118
we don't need to submit an additional frame

I suspect this is because HCPP now (indirectly) uses a surface flinger which states in the documentation:
https://source.android.com/docs/core/graphics/surfaceflinger-windowmanager
> When the display is between refreshes, the display sends the VSync signal to SurfaceFlinger. The VSync signal indicates that the display can be refreshed without tearing. When SurfaceFlinger receives the VSync signal, SurfaceFlinger walks through its list of layers looking for new buffers. If SurfaceFlinger finds a new buffer, SurfaceFlinger acquires the buffer; **_if not, SurfaceFlinger continues to use the previously acquired buffer._** SurfaceFlinger must always display something, so it hangs on to one buffer. If no buffers have ever been submitted on a layer, the layer is ignored.

(emphasis mine)

Does that make sense?

flar: It's vague and doesn't explain how buffers go away. Also, there are buffers and there are layers. It sounds like you can remove them if they have the text "so it hangs on to one buffer". Do we have 2 buffers when we have PVs? Can we unregister them and leave only the buffer for the base scene?

gmackall: I believe we could remove them entirely, but we are already not doing that and instead managing the visibility of the overlay surface earlier (when there are no PVs visible), to avoid recreating and deleting it frequently:

https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc#L73-L80

See the description of https://github.com/flutter/flutter/pull/162908. This managing of the surfaces visibility was something that wasn't required in the old PV code path https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder.cc#L74
And I suspect we just are properly handling removing it when there are no platform views to overlay, but not when there are platform views BUT there are no overlays. And we can either update it to be empty, or hide it.

But after looking at it more perhaps it makes more sense for us to be hiding it instead of showing it as empty? Instead of submitting a frame, I've also tested something like:

```c++
void AndroidExternalViewEmbedder2::MaybeHideOverlaySurface() {
  if (overlay_layer_is_shown_) {
    jni_facade_->hideOverlaySurface2();
    overlay_layer_is_shown_ = false;
  }
}

void AndroidExternalViewEmbedder2::MaybeShowOverlaySurface() {
  if (!overlay_layer_is_shown_) {
    jni_facade_->showOverlaySurface2();
    overlay_layer_is_shown_ = true;
  }
}
```
And then we replace the lines here about submitting a frame with:
```c++
...
  if (overlay_frame != nullptr) {
    overlay_frame->set_submit_info({.frame_boundary = false});
    overlay_frame->Submit();
    MaybeShowOverlaySurface();
  } else {
    MaybeHideOverlaySurface();
  }
...
```

And this works as well

gmackall: To try to answer your question more directly, I see that the overlay surface has its own dedicated [`SurfaceControl`](https://developer.android.com/reference/android/view/SurfaceControl), which has docs that read:

> Handle to an on-screen Surface managed by the system compositor. The SurfaceControl is a combination of a buffer source, and metadata about how to display the buffers.

Based on this I believe that the overlay and the Platform views both have their own independent buffers. But this is all my reading on the fly, am new to this area of the code.

gmackall: I've changed to something like the above, mirroring the structure of how we hide the overlay layer when there are no platform views shown.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -147,8 +147,17 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
     }
   }
   if (overlay_frame != nullptr) {
+    if (!prev_frame_overlay_layer_shown_) {
+      jni_facade_->showOverlaySurface2();
+    }
     overlay_frame->set_submit_info({.frame_boundary = false});
     overlay_frame->Submit();
+    prev_frame_overlay_layer_shown_ = true;
```
flar: Looks good. I think it's cleaner if the boolean assignments could be moved inside of the ifs so they are associated with the action that changes the state itself.

flar: Also, the name could be "is_overlay_showing_" (or shown) since "prev_frame" isn't a relevant part of the state - just whether it is showing or not showing.

flar: Another idea - add private methods to do the on/off toggle of the overlay and consolidate the state management into dedicated methods near the state flag.

flar (approved):
Functionally looks fine - some suggestions for style/readability if you like...

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: I added private methods and renamed to `overlay_layer_has_content_`, and moved the reassignment inside of the if statements.

I also guarded the later call to `jni_facade_->showOverlaySurface2();` behind checking that the overlay layer has content to display
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc#L162

as otherwise I believe we could:
have a frame with
1. No PVs visible
2. Overlay content to display

followed by a frame with
1. PVs visible
2. No Overlay content to display

And we would get the same bug again, because the hiding we added in this pr would be overridden by the later call to show. Unlikely, but possible, so I made us only show if we also have overlay content to display. In that rare case we would make a double call to `jni_facade_->showOverlaySurface2();` but that doesn't seem like the biggest deal to me.

Hopefully this all looks right, sorry for the many iterations.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -167,7 +161,7 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
        slices = std::move(slices_), prev_frame_no_platform_views]() -> void {
         jni_facade->swapTransaction();
 
-        if (prev_frame_no_platform_views) {
+        if (prev_frame_no_platform_views && overlay_layer_has_content_) {
```
flar: I don't understand this change, if the has_content flag is true, then shouldn't the showOverlay method have been called in the showIfNeeded call?

flar: Doesn't this result in an extra JNI dispatch to "showOverlaySurface2"?

flar (changes requested):

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: > I don't understand this change, if the has_content flag is true, then shouldn't the showOverlay method have been called in the showIfNeeded call?

Yes, this was always the case. The reason I added it is that if it isn't true, before we would still have called dispatched to `showOverlaySurface2` when we have the following two frames:

one with
1. NO PVs visible
2. YES Overlay content to display

followed by a frame with
1. YES PVs visible
2. NO Overlay content to display

But in this case we still don't want to display the overlay surface (it would result in the same bug that the PR is fixing in the first place). So this addition guards against that

gmackall: Adding an and to the if statement doesn't result in an extra call, as it only makes the dispatch to `showOverlaySurface2` happen less often.

But before this change, and still after, we will get a double dispatch if we get two frames in a row with the following properties:
1. NO PVs visible
2. NO overlay content to display

followed by
1. YES PVs visible 
2. YES Overlay content to display

gmackall: Let me know if any of this needs clarifying

flar: What confuses me is that the only way that "overlay_layer_has_content_" can become true (the only assignment to it that sets it to true) is right after the ShowIfNeeded method calls `showOverlaySurface2()`.

So, if the variable being true means absolutely that someone called `showOverlaySurface2` then why does it need to be called here?

Oh, hey, there is a call in this file to hideOverlaySurface2 that doesn't go through the new methods. I think this might be cleaner if we made the private methods to hide and show it while setting the status variable true/false be the only way that we call those jni_facade methods.

flar: See lines 76/77.

flar: We might be able to also get rid of the prev_frame_no_pvs variable as well.

Up top, if we don't have anything to process, then call the internal hideIfNeeded method.

Then we show it if overlay_frame isn't null, or we show it in the callback, but only one of those should be needed, shouldn't it?

Also, is that PostTask run in a different thread? If it is, then we might want to think about whether the first case of noticing no more platform views shouldn't also be delegated to a Posted Task.

flar: We need to either always call it in this method, or always call it in a posted task on the other thread. Having it called by either thread could lead to race conditions.

flar (changes requested):

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: > Up top, if we don't have anything to process, then call the internal hideIfNeeded method.
>
> Then we show it if overlay_frame isn't null, or we show it in the callback, but only one of those should be needed, shouldn't it?

I don't believe this would work. The problem is that the visibility of the overlay frame is a function of: 1. if we have overlay content to display and 2. if the platform view layer has platform views to display

we can't model this with just one boolean/calls to the method that set just one boolean.

gmackall: Hmm maybe it would actually work, perhaps we can just remove the call that this comment is on and rely on the earlier call. Give me a moment to try it out

gmackall: I've changed to something like you suggested. I'm a bit apprehensive about the approach, as if the state gets out of sync with the actual visibility of the surface the problems are more persistent now. But can go with this approach if you prefer.

flar: A - do we really need an overlay
B - are we currently showing the overlay
C - who toggles B based on A.

A - I think the main discovery here is that we had 2 places where we decided whether or not overlay was needed - one at the top of the method, and then another in/after the loop where we think we might need overlays, but we might also be wrong and we only know as we run through the loop entries. So, cleaning up the decision about whether this frame should have an overlay was a good aspect of this fix.

B - Also, consolidating the JNI hide/show calls with a boolean that only makes the call if the state is different makes it easier to deploy the actual calls without conflict with each other's state.

C - But, we also have "who should do the actual JNI hide/show for best consistency and thread safety" and finding a good place for those "hide/showIfNeeded" calls - that's I think where we are right now. It looks like there is another thread involved - thus the PostTask - so we should make that thread be the one that makes the actual toggle of the overlay surface based on the information determined in A.

gmackall: I've swapped to toggling the visibility in the thread, PTAL

flar: We still do it in 2 different threads. The one at the top of SubmitFlutterView is doing it in the calling thread. The one in the PostTask is doing it in the thread that processes those tasks. I'm not sure when this method is called and by which thread so I'm not familiar with whether or not that causes a conflict.

@reidbaker Are you familiar with these threading issues?

gmackall: Hmm, this was always the case though. The way in which these threads are used is identical before and after the PR. Do you have a reason to suspect there was an issue with the threading model that Jonah originally wrote?

gmackall: And also if so, do you think it should be addressed in this PR? It sounds to me like that would be an independent problem from the one this PR is solving, given that this pr does not modify the threading approach.

flar: I think we are no worse than we were before on this front, it's more of a "as long as we are cleaning this up, let's try to deal with other issues that might arise". I'll approve and if @reidbaker has any suspicions about the threading here we can open up a new issue.

flar (approved):
LGTM

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
reidbaker: @flar at this point @gmackall understands this code more deeply than I do. Threading issues are tough and if we think we have found a race condition that would cause issues I support gmackall working on a change to that model. If you think that we are in no worse of a condition then I support landing this pr as it is an improvement.

reidbaker (approved):

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
flar: I don't think we are worse than we were before wrt threading. We fixed the bookkeeping, but the calls are still potentially spread between 2 threads. I would suggest to submit a new issue about this, but without a reproducible test case I'm not sure what value that would have. Is there anyone else around who understands the threading of the embedders?

gmackall: This code was mostly written by Jonah and reviewed by John McCutchan iirc. I don't think there is anyone who knows why it was written this way. 

I think in the absence of a known issue with it we'd (android team) prefer to avoid make changes here. But if you have a repro of a problem caused by the threading model I'd happily take a look.

reidbaker: @gmackall I dont think someone else is going to dig into a repro condition for us. I think we now own this code and the likely issues we will see will be race conditions which will be hard to debug. 

Consider adding a comment in the code about the thread model and maybe we spend some time this week drawing a diagram and figuring out if we think there is an issue.

flar: In particular, the thing I suppose might happen here is that a thread might be processing the rendering of a previous frame and show the overlay while another thread is processing a new frame without overlays and hide it. Those 2 threads would be in a race condition.

But, at least if we always us the new methods to hide/show the overlay then we have accurate state tracking and once we get an overlay for more than one frame, or it goes away for more than one frame, then we will catch up and it will appear or disappear on the second frame when both threads are now processing frames that both need or don't need the overlay. We might have the wrong state for a frame, but not permanently like we had before.

It's still possible in the worst case that the setting of the boolean and the call to change the state get interleaved on the two threads. We can't really solve that without adding the overhead of a mutex.

reidbaker: @flar how much overhead would a mutex be if we didnt interleave the calls very often? Basically since we know this is called from different threads and there is a boolean should we lock a mutex just to be safe?

reidbaker: I am guessing that a possible frame stutter would be the consequence instead of the wrong content which is probably a reasonable trade off if it only happened when we did interleave calls.

gmackall: I don't think we should add a mutex here. If we suspect there is an actual issue here, and we want to make a change I would think we would modify lines 73-79:
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc#L73-L79

to be something like
```c++
  if (!FrameHasPlatformLayers()) {
    frame->Submit();
    task_runners_.GetPlatformTaskRunner()->PostTask(fml::MakeCopyable(
        [&, jni_facade = jni_facade_]() {
          HideOverlayLayerIfNeeded();
          jni_facade->applyTransaction();
        }));
    return;
  }
```

so that all work is done on the platform thread. I believe all jni work is intended to be done on the platform thread anyways, based on 
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h#L31-L36

flar: I think we'd only need it if the state is actually changing, so the performance impact would be negligible, but it only solves the relatively rare case of mismatched boolean/overlay state which would require an extremely fine race condition.

The "wrong for a frame" issue has a much higher potential due to a higher chance for bad ordering and solving that would probably require fixing the threading issue which would obviate the need for the mutex.

I also feel like we have 3 asynchronous time frames here that might interfere and that maybe we need a better architecture.

One place in the code calls `SurfaceFrame::Submit` and then directly hides/shows the overlay. How do those interleave given our threading model?

Another place in the code calls `SurfaceFrame::Submit` and then posts a task to hide/show the overlay. How do those interleave? And, on top of how the Submit call compares to the posted task, how does the thread that processes the submitted frame compare to the posted task? Are those guaranteed to be back to back?

It feels like we need a way to say "Submit this frame, and when you do, call this other callback when you are finished with processing the submitted frame" to guarantee that the code that shows/hides the overlay is directly executed while processing a Submit call.

flar: Or, for all I know, is `SubmitFrame` synchronous? I somehow feel that it isn't because why then post the task after submitting it in the second case?

flar: Back to @gmackall - yes I think that using a posted task in both cases is safer.

But I still wonder how the timing of processing a call to Submit compares to processing a posted task. I feel like this should be `SubmitAnd(callback-to-be-called-when-the-actual-submit-finishes)`.

There used to be a concept of skipping a frame that was posted, but I'm not sure if that happens upstream of this code or downstream from it. It's in the rasterizer somewhere and in some cases trying to process a frame can return "do that again" or "skip that one and go on to the next" and I'm not sure if the separated "submit vs posttask" processing here survives those exceptional conditions.

flar: (See `DrawStatus` and `DrawSurfaceStatus` in rasterizer.h)

flar: It looks like the `Submit` calls simply call callbacks that were supplied when some implementation of `Surface` constructed a `SurfaceFrame` inside its `AcquireFrame` method. In some cases the callbacks just return true. In other cases they make a platform call with non-obvious threading and/or synchronicity implications.

It's not clear how the actions of those submit callbacks relate to either further code in the same thread, or in code posted as a task.

gmackall: Ok I spent more time looking in to this and am somewhat more confused, but it seems we maybe don't have a threading problem here? It definitely looks to me like the call to hideOverlaySurface isn't being made on the platform thread, but the method that it calls:
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java#L1338
Is annotated as being callable only on the ui thread.

And even if I locally add a line to ensure this:
```java
  @SuppressLint("NewApi")
  @UiThread
  public void hideOverlaySurface2() {
    ensureRunningOnMainThread(); // <-- new line here
    if (platformViewsController2 == null) {
      throw new RuntimeException(
          "platformViewsController must be set before attempting to destroy an overlay surface");
    }
    platformViewsController2.hideOverlaySurface();
  }
```

I still don't get an error, which tells me this must be running on the platform (ui) thread. This is extra confusing, because if I log from the c++ side:
```c++
  if (!FrameHasPlatformLayers()) {
    frame->Submit();
    // If the previous frame had platform views, hide the overlay surface.
    HideOverlayLayerIfNeeded();
    FML_LOG(ERROR) << "HI GRAY, hiding overlay layer because no platform views and current thread is "
                   << (task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread()
                           ? "platform"
                           : "not platform");
    jni_facade_->applyTransaction();
    return;
  }
```
I get logs:
```
HI GRAY, hiding overlay layer because no platform views and current thread is not platform
```

Which is extremely confusing! But my takeaway here is that we must eventually be passing this call to the platform thread, as otherwise we would be failing the call to `ensureRunningOnMainThread();`

So I think we don't have a race condition here (because it seems everything is being done on the platform thread), even if I don't understand the internals of why that is the case.

flar: Perhaps we could even move both calls out of the posted task back into the main method?

gmackall: Yeah, I had that thought as well but when I tried that for the task that we are posting below I DID get an error for making a call not on the platform thread:
```
E/flutter ( 7330): [ERROR:flutter/fml/platform/android/jni_util.cc(206)] java.lang.RuntimeException: Methods marked with @UiThread must be executed on the main thread. Current thread: 1.raster
E/flutter ( 7330): 	at io.flutter.embedding.engine.FlutterJNI.ensureRunningOnMainThread(FlutterJNI.java:1607)
E/flutter ( 7330): 	at io.flutter.embedding.engine.FlutterJNI.onDisplayPlatformView2(FlutterJNI.java:1373)
```

Again, very strange because it implies we are only passing the hideOverlaySurface calls but not the calls done in the task below. I really don't understand why that would be the case.

PR #173928, opened 2025-08-17
Make sure that a DesktopTextSelectionToolbar doesn't crash in 0x0 env…

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the DesktopTextSelectionToolbar UI control.

### DIFF
--- a/packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
+++ b/packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
@@ -27,4 +27,20 @@ void main() {
 
     expect(tester.getTopLeft(find.byType(DesktopTextSelectionToolbarButton)), anchor);
   });
+
+  testWidgets('DesktopTextSelectionToolbar renders at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Center(
+          child: SizedBox.shrink(
+            child: DesktopTextSelectionToolbar(
+              anchor: const Offset(10, 10),
+              children: const <Widget>[Text('X')],
+            ),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(DesktopTextSelectionToolbar)).isEmpty, isTrue);
+  });
 }


packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
```diff
@@ -27,4 +27,17 @@ void main() {
...
+      MaterialApp(
+        home: SizedBox.shrink(
+          child: DesktopTextSelectionToolbar(
+            anchor: const Offset(10, 10),
+            children: const <Widget>[Text('X')],
+          ),
+        ),
+      ),
+    );
+  });
```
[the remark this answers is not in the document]
justinmc: I think this is fine either way.

justinmc (approved):
LGTM 👍 

@dkwingsmt During triage you mentioned you were concerned about this test, were you wondering if the toolbar would end up in an Overlay or what was it?

dkwingsmt:
Yeah, I was saying I wanted someone familiar with this to verify its validity, such as making sure it's not laying out overlay.

justinmc:
In that case we are good, the Overlay thing is done in text_selection.dart, not as a part of DesktopTextSelectionToolbar itself 👍 . Thanks for clarifying!

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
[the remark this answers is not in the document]
justinmc: Actually, in all of these PRs, we should have an a `expect` that checks that the relevant widget (DesktopTextSelectionToolbar here) has a size of 0x0. Can you add that @ahmedsameha1?

Thanks @LongCatIsLooong for flagging this.

justinmc (changes requested):

ahmedsameha1:
@justinmc Could you resolve this in https://github.com/flutter/flutter/issues/6537 because, according to https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193, there is no need to add that expectation.
Please consider this: https://github.com/flutter/flutter/pull/172074#issuecomment-3106174908

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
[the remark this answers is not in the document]
dkwingsmt: @justinmc We don't need any `expect`. The original issue only wants to make sure that it doesn't crash, not even needing the actual size to be zero.

ahmedsameha1:
I have updated this pull request according to https://github.com/flutter/flutter/issues/6537#issuecomment-3216557050.

dkwingsmt (approved):
LGTM, thank you!

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
```diff
@@ -27,4 +27,20 @@ void main() {
...
+          child: SizedBox.shrink(
+            child: DesktopTextSelectionToolbar(
+              anchor: const Offset(10, 10),
+              children: const <Widget>[Text('X')],
+            ),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(DesktopTextSelectionToolbar)).isEmpty, isTrue);
```
justinmc, line 44: Nit: I think the error message will be more useful if you do it like this:

```dart
expect(tester.getSize(find.byType(DesktopTextSelectionToolbar)), Size.zero);
```

justinmc (approved):
LGTM 👍

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
justinmc: Eh this is arguable. Up to you @ahmedsameha1.

ahmedsameha1: Please consider this: https://github.com/flutter/flutter/pull/172074#issuecomment-3106174908

LongCatIsLooong: The framework doesn't allow a `RenderObject` to choose a size that doesn't satisfy the incoming constraints. If the incoming constraints is tight 0x0, the child `RenderBox` has no option but to size itself to `Size.empty`. I think the size being Size(0.0, 20.0) is illegal and that might be a bug.

dkwingsmt: I'm curious as well. @LongCatIsLooong can you take a look at what happened at https://github.com/flutter/flutter/pull/172074#issuecomment-3106174908 ?

LongCatIsLooong: > I'm curious as well. @LongCatIsLooong can you take a look at what happened at [#172074 (comment)](https://github.com/flutter/flutter/pull/172074#issuecomment-3106174908) ?

The `expect` is trying to measure the size of the text instead of that of the MaterialBanner widget.

dkwingsmt: Ah I see. Because Flutter doesn't enforce that the child to have large _constraints_ than the parent's size.

Therefore if we measure the size of the target widget (`MaterialBanner`) instead of the content the result should be exactly 0x0, which I've verified on the simulator.

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
```diff
@@ -27,4 +27,20 @@ void main() {
 
     expect(tester.getTopLeft(find.byType(DesktopTextSelectionToolbarButton)), anchor);
   });
+
+  testWidgets('DesktopTextSelectionToolbar renders at zero area', (WidgetTester tester) async {
```
dkwingsmt, line 31:
```suggestion
  testWidgets('DesktopTextSelectionToolbar does not crash at zero area', (WidgetTester tester) async {
```

packages/flutter/test/material/desktop_text_selection_toolbar_test.dart
dkwingsmt, line 44:
```suggestion
    expect(tester.getSize(find.byType(DesktopTextSelectionToolbar)), Size.zero);
```

dkwingsmt (commented):
Checklist:
- [ ] The test name goes “does not crash at zero area”
- [ ] The target widget is expected to have a size of exactly `Size.zero`

Piinks:
Rebasing to retest

PR #173982, opened 2025-08-18
Add java as a dependency of linux analyze since it is required by ktlint

reidbaker:
I did not find documentation for which java version was required but we use Java 21 for many tests and it passed locally for me with java 17 and 24 
https://github.com/pinterest/ktlint/releases  

https://github.com/flutter/flutter/pull/173869#issuecomment-3197941356 
A rerun fixed this issue which indicates some of our bots happen to have java and some do not. 

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #173988, opened 2025-08-18
Add `open_jdk` to `Linux analyze`

matanlurey:
Towards https://github.com/flutter/flutter/issues/173986.

This build _never_ declared (either directly or through `platform_properties`) that it required the Java JDK, and appears to have been "accidentally" getting it installed via a (now expired) cache of the Java JDK.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -348,7 +348,8 @@ targets:
       shard: analyze
       dependencies: >-
         [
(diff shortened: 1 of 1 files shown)


reidbaker (approved):
https://github.com/flutter/flutter/pull/173982 not sure what is different from my pr but approved.

Issue #174004, opened 2025-08-19
Edit the 3.35 blog to warn users that abiFilters is now set in Flutter apps


angcyo:
### Steps to reproduce

```
android {
    ...
    defaultConfig {
        ...

        ndk {
            abiFilters.add("arm64-v8a")
        }
    }
}
```

### Expected results

The abi should only include `arm64_v8a`.

### Actual results

The output abi will still include `"x86_64", "arm64_v8a" and "armeabi-v7a"`.

<img width="501" height="402" alt="Image" src="https://github.com/user-attachments/assets/38141b4f-b1c9-40da-bc2a-2c157ad80e79" />

### Code sample

Sample source code project:
https://github.com/angcyo/flutter_abi_filters_demo

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.35.1, on macOS 15.6 24G84 darwin-arm64, locale zh-Hans-CN) [310ms]
    • Flutter version 3.35.1 on channel stable at /Users/angcyo/Library/Android/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 20f8274939 (4 天前), 2025-08-14 10:53:09 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-lldb-debugging

[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [1,120ms]
    • Android SDK at /Users/angcyo/Library/Android/sdk
    • Emulator version 36.2.4.0 (build_id 13899551) (CL:N/A)
    • Platform android-36, build-tools 36.0.0
    • Java binary at: /Applications/Android Studio Preview.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.7+-13880790-b1038.58)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor --android-licenses

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [699ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16F6
[...]

PR #174008, opened 2025-08-19
[Android] Fix abiFilters invalid to Flutter Gradle Plugin

angcyo:
Fix Issues https://github.com/flutter/flutter/issues/174004

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
+++ b/packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
@@ -169,13 +169,17 @@ class FlutterPlugin : Plugin<Project> {
             //
             // If the user has specified abiFilters in their build.gradle file, those
             // settings will take precedence over these defaults.
-            FlutterPluginUtils.getAndroidExtension(project).buildTypes.forEach { buildType ->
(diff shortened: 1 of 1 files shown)


reidbaker (approved):
Thank you for this fix. To make sure we do not regress we will want an integration test. If you are willing to author that I would be appreciative but I do not want to block this pr. 

@gmackall review this assuming we will cherry pick into 3.35

gmackall (changes requested):
@angcyo you will need to sign the CLA before this PR can proceed.

This will need to pass tests, as well as have its own test, before it can land. Also, can I ask why in the original report you are adding to the set instead of defining/overriding the set as done in the documentation?
https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/dsl/Ndk#abiFilters:kotlin.collections.MutableSet
I believe if you instead set the `abiFilters` property to your desired set, this should work right?

If setting (instead of adding) as described above does not work then I think a change like this is warranted (but probably only in that case).

camsim99:
Note from Android triage: See last comment in https://github.com/flutter/flutter/issues/174004.

Issue #157470, opened 2024-10-23
There should be clarity and guidance on to address disposal of objects in exceptional situations.


polina-c:
### Steps to reproduce

-

### Expected results

-

### Actual results

-

### Code sample

<details open><summary>Code sample</summary>

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[Paste your output here]
```

</details>

PR #174014, opened 2025-08-19
Add a `RepeatingAnimationBuilder` API

### LINKED ISSUE #174011: TweenAnimationBuilder.repeat
### Use case

Implicit widgets are great, but they can't be used for repeating animations.

For example, here's a spinning green square:

```dart

void main() {
  runApp(const MaterialApp(home: SpinningSquarePage()));
}

class SpinningSquarePage extends StatefulWidget {
  const SpinningSquarePage({super.key});

  @override
  State<SpinningSquarePage> createState() => _SpinningSquarePageState();
}

class _SpinningSquarePageState extends State<SpinningSquarePage>
    with TickerProviderStateMixin {
  late final AnimationController _controller = AnimationController(
    duration: const Duration(seconds: 2),
    vsync: this,
  );

  @override
  void initState() {
    super.initState();
    _controller.repeat();
  }

  @override
  void dispose() {
    _controller.dispose();
    super.dispose();
  }

  @override
    return Scaffold(
      body: Center(
        child: RotationTransition(
          turns: _controller,
          child: const ColoredBox(
            color: Colors.green,
            child: SizedBox.square(dimension: 100),
          ),
        ),
      ),
    );
  }
}
```

### Proposal

Rough API:

```dart
class TweenAnimationBuilder {
  TweenAnimationBuilder.repeat({
    Key? key,
    required Tween<T> tween,
    required Duration duration,
    Curve curve,

    // If true, the animation flip-flips between forwards and reverse.
    bool reverse = false,

    // If true, the animation is paused.
    //
    // For example, LinearProgressIndicator would pause its animation
    // if it has a value, and would unpause the animation if its value
    // is null. 
    bool paused = false,

    required ValueWidgetBuilder builder,
    Widget? child,
  });
}
```

Example use:

```dart
import 'dart:math' as math;

void main() {
  runApp(const MaterialApp(home: SpinningSquarePage()));
}

class SpinningSquarePage extends StatelessWidget {
  const SpinningSquarePage({super.key});

  @override
    return Scaffold(
      body: Center(
        child: TweenAnimationBuilder<double>.repeat(
          tween: Tween<double>(begin: 0, end: 1),
          duration: const Duration(seconds: 2),
          builder: (context, rotation, _) {
            return Transform.rotate(
              angle: math.pi * rotation * 2,
              child: const ColoredBox(
                color: Colors.green,
                child: SizedBox.square(dimension: 100),
              ),
            );
          },
        ),
      ),
    );
  }
}
```

bernaferrari:
Implements / Close https://github.com/flutter/flutter/issues/174011

diff


packages/flutter/lib/src/material/progress_indicator.dart
```diff
@@ -1189,24 +1133,28 @@ class _RefreshProgressIndicatorState extends _CircularProgressIndicatorState {
...
           context,
           // Lengthen the arc a little
-          1.05 * _CircularProgressIndicatorState._strokeHeadTween.evaluate(_controller),
-          _CircularProgressIndicatorState._strokeTailTween.evaluate(_controller),
-          _CircularProgressIndicatorState._offsetTween.evaluate(_controller),
-          _CircularProgressIndicatorState._rotationTween.evaluate(_controller),
+          1.05 * _CircularProgressIndicatorState._strokeHeadTween.evaluate(animation),
+          _CircularProgressIndicatorState._strokeTailTween.evaluate(animation),
+          _CircularProgressIndicatorState._offsetTween.evaluate(animation),
+          _CircularProgressIndicatorState._rotationTween.evaluate(animation),
```
loic-sharma: Can we use [`transform`](https://api.flutter.dev/flutter/animation/Tween/transform.html) instead of `evaluate` here to remove the `AlwaysStoppedAnimation`?

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -117,7 +117,99 @@ class TweenAnimationBuilder<T extends Object?> extends ImplicitlyAnimatedWidget
...
+  ///
+  /// {@template flutter.widgets.TweenAnimationBuilder.repeat.paused}
+  /// The [paused] parameter controls whether the animation is running:
+  ///  * When `false` (default), the animation runs continuously.
+  ///  * When `true`, the animation pauses at its current value.
+  ///  * Changing from `true` to `false` resumes the animation from where it paused.
+  /// {@endtemplate}
+  ///
+  /// Unlike the default constructor, [onEnd] is not supported for repeating
+  /// animations since they never end.
```
loic-sharma: Nit: I'd consider dropping this line - calling out something that's not present might confuse users.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -117,7 +117,99 @@ class TweenAnimationBuilder<T extends Object?> extends ImplicitlyAnimatedWidget
...
+  /// )
+  /// ```
+  /// {@end-tool}
+  ///
+  /// See also:
+  ///
+  ///  * [AnimationController.repeat], which this constructor replaces for simple use cases.
+  ///  * [TweenAnimationBuilder], the default constructor for single animations.
+  ///
+  /// ## Ownership
```
loic-sharma: Could you move the Ownership section up above the code samples?

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -117,7 +117,99 @@ class TweenAnimationBuilder<T extends Object?> extends ImplicitlyAnimatedWidget
...
+  ///
+  /// The [TweenAnimationBuilder] takes full ownership of the provided [tween]
+  /// instance and mutates it. Once a [Tween] has been passed to a
+  /// [TweenAnimationBuilder], its properties should not be accessed or changed
+  /// anymore to avoid interference with the [TweenAnimationBuilder].
```
loic-sharma: This content mirrors but is slightly different from the "Ownership of the Tween" content on TweenAnimationBuilder. I'd consider making a template out of the original content and just use the template here.

https://github.com/flutter/flutter/blob/6cbd98adc517ca7a20c724c4f0c5a5a385b95f74/packages/flutter/lib/src/widgets/tween_animation_builder.dart#L58C1-L67C34

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -117,7 +117,99 @@ class TweenAnimationBuilder<T extends Object?> extends ImplicitlyAnimatedWidget
...
+  const TweenAnimationBuilder.repeat({
+    super.key,
+    required this.tween,
+    required super.duration,
+    super.curve = Curves.linear,
+    bool reverse = false,
+    bool paused = false,
+    required this.builder,
+    this.child,
+  }) : _isRepeating = true,
```
loic-sharma: Oof.. This is not as nice as I had hoped. Maybe this would be better as a separate `RepeatedTweenAnimationBuilder` widget?

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: the issue of making a template (now) where it is a separate component is that I can't repeat "The [TweenAnimationBuilder] takes full ownership" unless it accepts variable so might make sense to repeat now :( or  just omit or duplicate.

bernaferrari:
I need your help @loic-sharma, I have a leak test failing and I believe it is because of `assert()` which checks if tween is valid (has begin and end). I don't know how to solve that. I can move the assert to the constructor, but then it can't be `const`, or I can just remove the assert (but then if someone forgets begin or end will get runtime issues). What do you think? Shall I remove `const` or is there a simple good way to solve this problem?

loic-sharma:
@bernaferrari I would go ahead and make the widget non-`const` for now.

It looks like leaking objects in exceptional situations (like a test for error conditions) is a known issue: https://github.com/flutter/flutter/issues/157470. When we fix this, we can make this widget `const` :)

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,168 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  /// The curve to apply when animating the value.
+  ///
+  /// Defaults to [Curves.linear].
+  final Curve curve;
+
+  /// Called every time the animation value changes.
+  ///
+  /// The builder receives the animation object and the optional child.
+  /// The current value can be accessed via animation.value.
+  final Widget Function(BuildContext context, Animation<T> animation, Widget? child) builder;
```
loic-sharma: `TweenAnimationBuilder.builder` has type `ValueWidgetBuilder<T>` - should we use the same type here to be consistent?

bernaferrari: Good catch! Fixing

bernaferrari: This is tricky. ValueWidgetBuilder<T> takes the actual value T, while  RepeatingTweenAnimationBuilder is currently taking Animation<T>.
The reason is, many examples require animation, so it doesn't make sense for us to do `animation.value` and then create a new FixedAnimation(value) or something like that, I thought it made more sense to just use animation directly instead of animation.value, but it makes it different than TweenAnimationBuilder, so it is tricky.

I wish Repeating had value instead of animation, but then it will make the code for example worse than right now. Any thoughts?

Examples like this would suffer:
image

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,168 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    super.didUpdateWidget(oldWidget);
+
+    // Update duration if changed
+    if (widget.duration != oldWidget.duration) {
+      _controller.duration = widget.duration;
+    }
+
+    // Update animation if curve or tween changed
+    final bool tweenChanged =
+        widget.tween.begin != oldWidget.tween.begin || widget.tween.end != oldWidget.tween.end;
```
loic-sharma: Should we also check if the tween's runtime type changed? For example, I might switch to a different `Tween` implementation that has a different `lerp` function.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,168 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  late Animation<T> _animation;
+  CurvedAnimation? _curvedAnimation;
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = AnimationController(duration: widget.duration, vsync: this);
+    _updateAnimation();
+
+    // Start the animation if not paused and begin != end
```
loic-sharma: Please add a period to the end of all comments :)

```suggestion
    // Start the animation if not paused and begin != end.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: Yes!! I completely forgot about runtime type, thanks!

bernaferrari:
Well, I finished it, but I guess I took too long :( This completely breaks my PR: https://github.com/flutter/flutter/pull/174605

We can still merge it, but I think it will only be useful for samples. The biggest reason for this PR (progress indicator) is now even more complex with custom controller. I could do something like, if no custom controller is provided, use `RepeatingTweenAnimationBuilder` else do the standard behavior, but it will be duplicated code, so I'm not sure how to progress.

We can:
- When not using custom controller, use `RepeatingTweenAnimationBuilder` (but might have duplicated rendering logic)
- I guess allow custom `AnimationController` inside `RepeatingTweenAnimationBuilder`
- Merge but only use on examples, not on ProgressIndicator
- Don't merge.

Any thoughts? For now, I'm going for option 1, it is a bit more code but should work. I'll fix whatever is causing the golden later (interesting there was no golden issue before).

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: TLDR: We should use `ValueWidgetBuilder<T>` here as well.

Whether the builder accepts a `T` or a `Animation<T>` has implications on how many times the builder should be called:

```dart
FooAnimationBuilder(
  tween: Tween<double>(begin: 0, end: 1),
  builder: (context, double sizeFactor, child) {
    // This builder is called each time sizeFactor changes.
    // The subtree is rebuilt on each tick.
    return ...;
  },
);

FooAnimationBuilder(
  tween: Tween<double>(begin: 0, end: 1),
  builder: (context, Animation<double> sizeFactorAnimation, child) {
    // This builder should be called ONCE. You or some child widget needs an
    // AnimatedBuilder to trigger a callback each time sizeFactorAnimation
    // changes. This lets you minimize the subtree that is rebuilt each
    // time the animation ticks.
    return ...;
  },
);
```

Since `TweenAnimationBuilder` uses the former pattern (`T`), `RepeatingTweenAnimationBuilder` should use the same pattern - it'd be confusing otherwise. I do think the latter pattern (`Animation<T>`) is something we should consider, however let's punt that until after this PR as I'd like to come up with a new naming convention that makes the distinction clear. 

---

Counterintuitively, the `AlwaysStoppedAnimation` thing is a good thing. It's a code smell that indicates we're using some API incorrectly:

* **Option 1**: Switch to a child widget that doesn't accept an `Animation<T>`. For the `SizeTransition` example, we should consider switching to `ClipRect` + `Align` instead. For `RotationTransition`, we should consider switching to a `Transform` instead.
* **Option 2**: Consider whether the child widget actually needs to accept an `Animation<T>`. If the child widget rebuilds itself on each tick, it should might as well just accept a `T`.
* **Option 3**: Don't use `RepeatingTweenAnimationBuilder`. Go back to an explicit animation instead. Examples for this category: `AnimatedIcon` animations, `FadeTransition` animations (this has important optimizations over using `Opacity` directly), etc.

Let me know if you have any feedback or questions!

loic-sharma:
@bernaferrari Hey there, sorry for the long response time on my end. I think we should still merge this and use it only for examples:

1. Let's make `RepeatingTweenAnimationBuilder`'s builder accept a `T` instead of `Animation<T>`. See [this comment](https://github.com/flutter/flutter/pull/174014#discussion_r2377296222) for details.
2. Use `RepeatingTweenAnimationBuilder` conservatively where it is clearly the right choice. I wouldn't update the progress indicators if that requires multiple rendering

This is still a good first step! We can continue to make `RepeatingTweenAnimationBuilder` more powerful (by expanding it to the `Animation<T>` pattern) later to use this in more areas :)

dkwingsmt:
Hi @bernaferrari ! Are you able to address the comments above so we can work together to get this landed?

bernaferrari:
Yes, yes. I'll update it today

packages/flutter/lib/src/material/progress_indicator.dart
```diff
@@ -1614,4 +1614,4 @@ class _LinearProgressIndicatorDefaultsM3 extends ProgressIndicatorThemeData {
 }
 // dart format on
 
-// END GENERATED TOKEN PROPERTIES - ProgressIndicator
+// END GENERATED TOKEN PROPERTIES - ProgressIndicator
```
loic-sharma: This looks like a typo

```suggestion
// END GENERATED TOKEN PROPERTIES - ProgressIndicator

```

examples/api/lib/material/list_tile/list_tile.0.dart
```diff
@@ -116,16 +114,18 @@ class _ListTileExampleState extends State<ListTileExample> with TickerProviderSt
...
-                sizeFactor: _sizeAnimation,
-                axisAlignment: -1.0,
-                // Wrap the ListTile in a Material widget so the ListTile has someplace
-                // to draw the animated colors during the size transition.
-                child: const Material(
-                  child: ListTile(
-                    title: Text('ListTile with SizeTransition'),
-                    tileColor: Colors.red,
-                    minVerticalPadding: 25.0,
+              child: ClipRect(
```
loic-sharma: Could you expand on why the `ClipRect` here? (I suspect this is a typo)

bernaferrari: arggggh I pushed but I'm still fixing some things. I'm not very happy with the overall PR because with using everything we discussed, it feels `RepeatingTweenAnimationBuilder` is not useful for transitions, and therefore there are only very limited amount of places it is useful. I'll revert a lot of the examples soon, since they don't need it anymore.

examples/api/lib/material/progress_indicator/linear_progress_indicator.1.dart
```diff
@@ -61,9 +37,17 @@ class _ProgressIndicatorExampleState extends State<ProgressIndicatorExample>
...
           children: <Widget>[
             const Text('Linear progress indicator', style: TextStyle(fontSize: 20)),
-            LinearProgressIndicator(
-              value: determinate ? controller.value : null,
-              semanticsLabel: 'Linear progress indicator',
+            RepeatingTweenAnimationBuilder<double>(
+              tween: Tween<double>(begin: 0, end: 1),
+              duration: const Duration(seconds: 2),
+              reverse: true,
+              paused: determinate, // Pause when determinate to match original behavior
```
loic-sharma: This looks like a stale comment? :)

```suggestion
              paused: determinate,
```

examples/api/lib/painting/borders/border_side.stroke_align.0.dart
```diff
@@ -53,127 +28,135 @@ class _StrokeAlignExampleState extends State<StrokeAlignExample> with TickerProv
...
-                      color: borderColor,
-                      width: borderWidth,
-                      strokeAlign: (animation.value * 2) - 1,
+                    BorderedBox(
+                      shape: RoundedRectangleBorder(
+                        borderRadius: BorderRadius.circular(cornerRadius),
+                        side: BorderSide(
+                          color: borderColor,
+                          width: borderWidth,
+                          strokeAlign: (value * 2) - 1,
```
loic-sharma: Can this be `strokeAlign` too?

```suggestion
                          strokeAlign: strokeAlign,
```

examples/api/lib/painting/borders/border_side.stroke_align.0.dart
```diff
@@ -53,127 +28,135 @@ class _StrokeAlignExampleState extends State<StrokeAlignExample> with TickerProv
...
-                      strokeAlign: (animation.value * 2) - 1,
+                    BorderedBox(
+                      shape: StarBorder(
+                        pointRounding: 1,
+                        innerRadiusRatio: 0.5,
+                        points: 8,
+                        side: BorderSide(
+                          color: borderColor,
+                          width: borderWidth,
+                          strokeAlign: (value * 2) - 1,
```
loic-sharma: Here as well?

```suggestion
                          strokeAlign: strokeAlign,
```

examples/api/lib/material/list_tile/list_tile.0.dart
loic-sharma: Ahh ok, no worries. Feel free to tag me in a comment when you're ready for another pass :)

bernaferrari:
Ok. Sorry for the loooooong wait. I think I commited everything we talked about. The examples surface got (drastically) reduced and I'm using "double" so it matches the TweenBuilder. I am a bit 🤨 because I wish it could be more useful, but I don't know how. With `FadeTransition` (etc) it gets super nasty, and `ProgressIndicator` got their own implementation. So it feels we are still looking for a "better" and "clean" solution for them.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,170 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+/// {@end-tool}
+///
+/// See also:
+///
+///  * [TweenAnimationBuilder], which animates a tween value once.
+///  * [AnimationController.repeat], the underlying mechanism.
+class RepeatingTweenAnimationBuilder<T extends Object> extends StatefulWidget {
+  /// Creates a widget that repeats a tween animation.
+  ///
+  /// The [tween], [duration], and [builder] arguments must not be null.
```
loic-sharma: These arguments aren't nullable, so I don't think this comment is necessary. Let me know if I'm missing something though!

```suggestion
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,170 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    _curvedAnimation?.dispose();
+    _curvedAnimation = CurvedAnimation(parent: _controller, curve: widget.curve);
+    _animation = widget.tween.animate(_curvedAnimation!);
+  }
+
+  @override
+  void didUpdateWidget(RepeatingTweenAnimationBuilder<T> oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    // Update duration if changed
```
loic-sharma: Nit:

```suggestion
    // Update the duration if it changed.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,170 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    if (widget.duration != oldWidget.duration) {
+      _controller.duration = widget.duration;
+    }
+
+    // Update animation if curve or tween changed
```
loic-sharma:
```suggestion
    // Update the animation if the curve or the tween changed.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: Good! Fixed

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,170 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    // If no animation needed (begin equals end), just build with static value
+    if (widget.tween.begin == widget.tween.end) {
+      return widget.builder(context, _animation.value, widget.child);
+    }
+
```
loic-sharma: Could you expand on this logic? Does this result in a measurable performance improvement?

I would consider removing this. Changing the tree's depth means that the widget returned by `widget.builder` can lose state unless the subtree has a `GlobalKey`. This behavior wouldn't be immediately obvious to newbies, I expect this might become a usability hiccup.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,168 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    final bool shouldAnimate = widget.tween.begin != widget.tween.end;
+    final bool wasAnimating = oldWidget.tween.begin != oldWidget.tween.end;
+
+    if (!shouldAnimate || widget.paused) {
+      _controller.stop(canceled: false);
+    } else if (!wasAnimating ||
+        widget.paused != oldWidget.paused ||
+        widget.reverse != oldWidget.reverse) {
+      _controller.repeat(reverse: widget.reverse);
```
loic-sharma: Hm a slight drawback here is that this will always goes forward, even if we paused while the animation was running in reverse. Unfortunately, this seems to be a limitation of the `AnimationController` API (we might need to add a `startReversed` parameter to `AnimationController.repeat`?). I think that's OK for now, we can always fix this later.

loic-sharma:
This is looking really good. That sample app is a work of art. Excellent work! :)

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: Probably no measurable performance improvement. I can remove. It came because Progress Indicator does something similar. But I agree, it is probably a super tiny optimization we don't need.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: The AnimationController API is very VERY confusing, and I guess part of that went for our new Tween. Right now, "reverse" means: when it reaches the end it "reverses" until zero. Like 0->1->0, instead of 0->1, 0->1. It doesn't mean, like, "please revert the movement in the opposite direction"

At some point I considered cleaning/improving the AnimationController, but it would be breaking. Not sure if we should just (eventually) make InfiniteAnimationController or something like that, which Compose does (and is much cleaner).

packages/flutter/test/widgets/tween_animation_builder_test.dart
```diff
@@ -393,4 +393,195 @@ void main() {
...
+          duration: duration,
+          tween: Tween<double>(begin: 0.0, end: 1.0),
+          builder: (BuildContext context, double value, Widget? child) {
+            values.add(value);
+            return const Placeholder();
+          },
+        ),
+      );
+
+      // Initial value should be 0.0
```
loic-sharma: Nit: could you do a pass on this file's comment and add periods at the end of all comments?

loic-sharma (approved):
LGTM, nice work!

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,163 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+///
+/// See also:
+///
+///  * [TweenAnimationBuilder], which animates a tween value once.
+///  * [AnimationController.repeat], the underlying mechanism.
+class RepeatingTweenAnimationBuilder<T extends Object> extends StatefulWidget {
+  /// Creates a widget that repeats a tween animation.
+  RepeatingTweenAnimationBuilder({
+    super.key,
+    required this.tween,
```
dkwingsmt: Do you think we can loosen the limitation of `tween` and instead allow `Animatable` here? Reasoning:
* I expect that this class is mostly used to play looping animations. Currently a smooth loop needs either `reverse: true` or special periodic properties (such as angles). This class can not be used for, for example, a TranslateY that increases ease-in and then decreases ease-in, since the animation is asymmetrical and has the same beginning and end.
* By roughly examining the code I think the only place that use the special properties of `Tween` are using its `begin` and `end` in `shouldChange`.

We can delegate `reverse` to `Animatable`, and delegate the comparison of `begin` and `end` to `Tween`.

bernaferrari: Could you review my last commit? (or I guess re-review everything) https://github.com/flutter/flutter/pull/174014/commits/86e67f31a00ee0f9470c70ac8dfa2a7e692f6874

I still need to test a bit more locally + see if tests pass. But can give you a clear picture if that's what you want or not.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  @override
+  void initState() {
+    super.initState();
+    _controller = AnimationController(duration: widget.duration, vsync: this);
+    final T startValue = widget.animatable.transform(0.0);
+    final T endValue = widget.animatable.transform(1.0);
+    _updateAnimation();
+
+    // Start the animation if not paused and the animation spans different values.
+    if (!widget.paused && startValue != endValue) {
```
loic-sharma: (GitHub seems to have lost my comment?)

Now that this accepts an `Animatable`, I can pass a `TweenSequence<Offset>` that goes right, then goes left back to where it started. We should run the animation even if the start and end values are the same.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    if (widget.duration != oldWidget.duration) {
+      _controller.duration = widget.duration;
+    }
+
+    final T startValue = widget.animatable.transform(0.0);
+    final T endValue = widget.animatable.transform(1.0);
+
+    // Update the animation if the curve or the animatable changed.
+    final bool animatableChanged =
+        !identical(widget.animatable, oldWidget.animatable) ||
```
loic-sharma: I think I prefer the previous `runtimeType` approach over `identical`.

With `identical`, I'll need to pass a `const` value or store the `Animatable` in a field or something to ensure the reference stays the same. I think this might be annoying.

bernaferrari: Do you have any suggestion in how to compare them? Not sure == will work well either.

chunhtai: runtimeType may not work correctly if they change tweens, the best way may be implementing equality in Animatable and subclass.

I am not sure why we didn't make animatable immutable

bernaferrari: I can't add equality to Animatable because Tween is not const (requires @immutable). What do I do? I want to scream. I remember some big discussions in the past on why Tween wasn't const and that was a regret.

chunhtai: > equality to Animatable because Tween is not const

do you need const and immurable to implement equality and hashcode?

bernaferrari: image
seems like yes

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+/// This example shows a continuously rotating square that can be paused and resumed.
+///
+/// ** See code in examples/api/lib/widgets/tween_animation_builder/repeating_tween_animation_builder.0.dart **
+/// {@end-tool}
+///
+/// See also:
+///
+///  * [TweenAnimationBuilder], which animates a tween value once.
+///  * [AnimationController.repeat], the underlying mechanism.
+class RepeatingTweenAnimationBuilder<T extends Object> extends StatefulWidget {
```
dkwingsmt: Now that it accepts an `Animatable` we might as well rename it to `RepeatingAnimationBuilder` and move it to a separate file.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
     return widget.builder(context, _currentTween!.evaluate(animation), widget.child);
   }
 }
+
+/// Widget that animates an [Animatable] value and repeats indefinitely.
+///
+/// The animation continuously cycles from the [Animatable] evaluated at 0.0 to
+/// the value at 1.0.
+/// When [reverse] is true, the animation reverses direction when it reaches
```
dkwingsmt: We might as well delegate `reverse` to some kind of `Animatable`, such as `TweenSequence`. (I'm not sure if there's a better class, and we can create one if needed, but I don't think it should be part of this class.)

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+/// the end. When [paused] is true, the animation stops at its current value.
```
dkwingsmt: I suggest removing `pause` and add an optional `AnimationController`.
* We have constantly received PRs that add animation controllers to existing widgets. 
* The `pause` is replaceable by an animation controller.
* Controlling animation status with propagating widget property seems to go against the principle of performance optimization as described in `@macro flutter.widgets.TweenAnimationBuilder.optimizations`.
* IMO `pause` is an antipattern, because it's a imperative operation, which feels wrong when passed in as a widget property.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+///
+/// {@macro flutter.widgets.TweenAnimationBuilder.optimizations}
```
dkwingsmt: Now that this widget isn't bound with `Tween` we might as well remove this paragraph. Developers should refer to `Tween`'s document for usage details when they use `Tween`s.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+///
+/// {@tool dartpad}
+/// This example shows a continuously rotating square that can be paused and resumed.
+///
+/// ** See code in examples/api/lib/widgets/tween_animation_builder/repeating_tween_animation_builder.0.dart **
```
dkwingsmt: Rename the example file (if we're renaming this class)

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+/// {@end-tool}
+///
+/// See also:
+///
+///  * [TweenAnimationBuilder], which animates a tween value once.
+///  * [AnimationController.repeat], the underlying mechanism.
+class RepeatingTweenAnimationBuilder<T extends Object> extends StatefulWidget {
+  /// Creates a widget that repeats an animatable animation.
```
dkwingsmt:
```suggestion
  /// Creates a widget that repeats an animation.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
chunhtai: discussed with the team offline, looks like the class is meant to be mutable, nothing we can do here I think. though we should convert to use == instead of identical in case some day we figure out a magic way to make it immutable and override == in the future.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    required this.builder,
+    this.child,
+  });
+
+  /// The animatable to drive repeatedly.
+  ///
+  /// The animation continuously cycles from the value returned by evaluating
+  /// [animatable] at 0.0 to the value returned at 1.0.
+  /// In most cases this will be a [Tween] or [TweenSequence], but any
+  /// [Animatable] that produces values of type [T] is accepted.
```
dkwingsmt: I'd suggest something that isn't as strong, such as:
```suggestion
  /// Typically this is a [Tween] or a [TweenSequence].
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  ///
+  /// ## Ownership
```
dkwingsmt: Now that it's an animatable this widget doesn't seem to mutate it anymore, and we should no longer discourage storing it. This instance should now be managed by the parent (which might not needs special mentioning any more).

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+    this.reverse = false,
+    this.paused = false,
+    required this.builder,
+    this.child,
+  });
+
+  /// The animatable to drive repeatedly.
+  ///
+  /// The animation continuously cycles from the value returned by evaluating
+  /// [animatable] at 0.0 to the value returned at 1.0.
```
dkwingsmt: I suggest removing this paragraph, since it's already mentioned in the class's doc.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
     return widget.builder(context, _currentTween!.evaluate(animation), widget.child);
   }
 }
+
+/// Widget that animates an [Animatable] value and repeats indefinitely.
+///
+/// The animation continuously cycles from the [Animatable] evaluated at 0.0 to
+/// the value at 1.0.
```
dkwingsmt: I'd like to add a bit more detail to this paragraph. (The last sentence assumed the removal of `reverse`; feel free to tweak it otherwise.)
```dart
/// The animation continuously cycles from the value produced by the
/// [Animatable] at 0.0 to the value at 1.0. The [builder] receives the
/// current value produced by this [Animatable] and builds the child from it.
/// When the animation reaches 1.0 it jumps immediately back to the value at
/// 0.0.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  ///
+  /// It is good practice to never store an [Animatable] provided to a
+  /// [RepeatingTweenAnimationBuilder] in an instance variable to avoid accidental
+  /// modifications of the [Animatable].
+  final Animatable<T> animatable;
+
+  /// The duration of the animation.
+  final Duration duration;
+
+  /// The curve to apply when animating the value.
```
dkwingsmt: I suggest a bit more details on how the curve affects the algorithm:
```dart
  /// The curve applied to the animation input before it is passed to the
  /// [animatable].
  ///
  /// In other words, the curve transforms the controller's 0.0..1.0 timeline
  /// and the resulting value is then fed into the [animatable].
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  ///
+  /// Defaults to [Curves.linear].
+  final Curve curve;
+
+  /// Called every time the animation value changes.
+  ///
+  /// The builder receives the current animation value and the optional child.
```
dkwingsmt: I suggest a different first paragraph to introduce this property more directly.
```suggestion
  /// A builder that creates the animated widget subtree.
  ///
  /// Invoked on every frame with the current animation value and the optional
  /// prebuilt child. The child can be incorporated to reuse a subtree that
  /// doesn't depend on the animation, avoiding unnecessary rebuilds.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  final ValueWidgetBuilder<T> builder;
+
+  /// The child widget to pass to the builder.
```
dkwingsmt:
```suggestion
  /// An optional widget to pass to the builder.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+  /// [RepeatingTweenAnimationBuilder] in an instance variable to avoid accidental
+  /// modifications of the [Animatable].
+  final Animatable<T> animatable;
+
+  /// The duration of the animation.
+  final Duration duration;
+
+  /// The curve to apply when animating the value.
+  ///
+  /// Defaults to [Curves.linear].
```
dkwingsmt: I suggest a bit more details on how the curve affects the algorithm:
```dart
  /// The curve applied to the animation input before it is passed to the
  /// [animatable].
  ///
  /// In other words, the curve transforms the controller's 0.0..1.0 timeline
  /// and the resulting value is then fed into the [animatable].
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: +1. Start values and end values shouldn't matter any more.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: A lot of values in the animation subsystem are designed to be mutable for performance consideration, and for the fact that it's not the class instance that matters, but the value that they evaluate to every frame.

What about this:
* Changing `curve` should not recreate the `CurvedAnimation` instance, since `CurvedAnimation.curve` is mutable.
* We should not evaluate the final animation with `Animatable.evaluate`. Instead, we should manually evaluate the result every frame by calling `Animatable.transform` with the result of  `CurvedAnimation`.

This way, neither new `curve` nor new `animatable` should lead to rebuilding `Animation` instances.

chunhtai: > but the value that they evaluate to every frame

That go against the animation abstraction we created in the first place, but I am ok with it.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: `RepeatingAnimationBuilder` sounds good to me too 👍

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: I strongly disagree. This would massively explode the boilerplate for common scenarios. 

For example, imagine the common curved "ping pong" effect:

```dart
RepeatingAnimationBuilder(
  animatable: TweenSequence<Offset>([
    TweenSequenceItem<Offset>(
      tween: Tween<Offset>(begin: Offset.zero, end: Offset(0, 100))
        .chain(CurveTween(curve: Curves.ease)),
      weight: 1.0,
    ),
    TweenSequenceItem<Offset>(
      tween: Tween<Offset>(begin: Offset(0, 100), end: Offset.zero)
        .chain(CurveTween(curve: Curves.ease)),
      weight: 1.0,
    ),
  builder: ...,
),
```

This widget's entire purpose is to reduce boilerplate for common scenarios. In my mind, this is exactly like `Column(spacing: )`.

Also, `AnimationController.repeat` already has a `reverse` argument. `RepeatingAnimationBuilder` would be better aligned to `AnimationController` if it also has a `reverse` argument. This would make it easier to migrate from `AnimationController` to `RepeatingAnimationBuilder`.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: Can you explain what kind of animation abstraction you're talking about?

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: The `paused` argument makes `RepeatingAnimationBuilder` useful for common patterns. For example, a loading indicator that loops repeatedly unless a `value` is provided.

> We have constantly received PRs that add animation controllers to existing widgets.

Could you expand on how this affects `RepeatingAnimationBuilder`?

`RepeatingAnimationBuilder` is not meant to be the most flexible API in the world. It's meant to be a convenient API for a common subset of repeating animations. If you need the full power of an animation controller, you should not use `RepeatingAnimationBuilder`.

> The `pause` is replaceable by an animation controller.

Yes, but switching to an animation controller adds **significant** boilerplate (you need a stateful widget, ticker provider, init state, dispose). In my mind, we should add a `paused` argument if this is a common scenario that we can easily support with an ergonomic API.

Anything you can do with `RepeatingAnimationBuilder`, you can do with a `AnimationController`. I don't think that is good reason to remove features from `RepeatingAnimationBuilder` :)

> Controlling animation status with propagating widget property seems to go against the principle of performance optimization as described in @macro flutter.widgets.TweenAnimationBuilder.optimizations.
> IMO pause is an antipattern, because it's a imperative operation, which feels wrong when passed in as a widget property.

I disagree, but I suspect I don't fully understand your pushback - could expand on this? The animation controller is an imperative API. The `RepeatingAnimationBuilder` is a declarative API, that maps the user's intent down to the imperative API. In my mind, this nicely mirrors other patterns, like the declarative widget tree vs the imperative render tree.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: What do you think if we create something like `ReversingAnimatable`? It will work like:
```dart
   RepeatingAnimationBuilder(
      animatable: ReversingAnimatable<Offset>(
        Tween<Offset>(begin: Offset.zero, end: Offset(0, 100))
            .chain(CurveTween(curve: Curves.ease))),
      builder: ...,
    );
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: Strictly speaking, this might not be called on every frame. For example, things like `Visibility` can mute a running animation.

Maybe:

```suggestion
  /// A builder that creates the animated widget subtree.
  ///
  /// The builder is called every time the animation value changes.
  ///
  /// The optional child can be used to avoid unnecessary rebuilds
  /// if a part of the subtree doesn't depend on the animation.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: > I disagree, but I suspect I don't fully understand your pushback - could expand on this? The animation controller is an imperative API. The RepeatingAnimationBuilder is a declarative API, that maps the user's intent down to the imperative API. In my mind, this nicely mirrors other patterns, like the declarative widget tree vs the imperative render tree.

Pardon me if I sounds like just echoing `@macro flutter.widgets.TweenAnimationBuilder.optimizations`, or if I'm understanding it wrong.

The animation pipeline is designed so that when you want to configure the animation (such as the begin and end of a tween) you don't have to pass the number all the way down through widget properties, which causes rebuilds. Instead you just drill in this change through mutable containers, such as `Tween`. The job of the animation pipeline is to pass in the correct mutable containers that contain configurations, not the correct configurations themselves.

In my opinion, the `pause` flag is similar to `begin` and `end`.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
chunhtai: that everything can be composed into a single animation<T>: `vsync -> apply curve ->drive animatable-> animation`

now that we use animatable as a function instead of a composable unit. which also strengthen that animatable should probably be immutable

chunhtai: I think maybe animatable should be like curve, where you can have AnimatableAnimation and you can swap out the animatable instead of mutating the animatable internal.

Anyway, this is out of scope, your suggestion sgtm

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: We do have [`ReverseTween`](https://api.flutter.dev/flutter/animation/ReverseTween-class.html) already. However, using `ReverseTween` would still be verbose:

```dart
RepeatingAnimationBuilder(
  animatable: TweenSequence<Offset>([
    TweenSequenceItem<Offset>(
      tween: Tween<Offset>(begin: Offset.zero, end: Offset(0, 100))
        .chain(CurveTween(curve: Curves.ease)),
      weight: 1.0,
    ),
    TweenSequenceItem<Offset>(
      tween: ReverseTween(
        Tween<Offset>(begin: Offset.zero, end: Offset(0, 100))
          .chain(CurveTween(curve: Curves.ease)),
      ),
      weight: 1.0,
    ),
  builder: ...,
),
```

I still strongly prefer just adding a `reverse` argument to `RepeatingAnimationBuilder`. (I do like the idea of a `ReverseAnimatable` though to allow reversing any `Animatable`).

packages/flutter/lib/src/widgets/tween_animation_builder.dart
loic-sharma: To summarize this thread, we're in agreement with this solution:

> * Changing `curve` should not recreate the `CurvedAnimation` instance, since `CurvedAnimation.curve` is mutable.
> * We should not evaluate the final animation with `Animatable.evaluate`. Instead, we should manually evaluate the result every frame by calling `Animatable.transform` with the result of `CurvedAnimation`.
> 
> This way, neither new `curve` nor new `animatable` should lead to rebuilding `Animation` instances.
```

File: packages/flutter/lib/src/widgets/tween_animation_builder.dart (hunk 24 above)
Comment (bernaferrari) [reply packages/flutter/lib/src/widgets/tween_animation_builder.dart]: Not sure if I can help much on the discussion. On Android it would be:
```kt
val infiniteTransition = rememberInfiniteTransition(label = "position")

val offsetY by infiniteTransition.animateFloat(
    initialValue = 0f,
    targetValue = 100f,
    animationSpec = infiniteRepeatable(
        animation = tween(
            durationMillis = 1000,
            easing = FastOutSlowInEasing // equivalent to Curves.ease
        ),
        repeatMode = RepeatMode.Reverse
    ),
    label = "offsetY"
)

Box(
    modifier = Modifier.offset(y = offsetY.dp)
) {
    // Your content
}
```
or finite:
```kt
var trigger by remember { mutableStateOf(false) }

val offsetY by animateFloatAsState(
    targetValue = if (trigger) 100f else 0f,
    animationSpec = repeatable(
        iterations = 2, // 0→100→0 (one full cycle)
        animation = tween(1000),
        repeatMode = RepeatMode.Reverse
    ),
    label = "offsetY"
)
```

on SwiftUI it would be:
```swift
struct ContentView: View {
    @State private var isAnimating = false
    
    var body: some View {
        Text("Hello World")
            .offset(y: isAnimating ? 100 : 0)
            .animation(
                .easeInOut(duration: 1)
                    .repeatForever(autoreverses: true),
                value: isAnimating
            )
            .onAppear {
                isAnimating = true
            }
    }
}
```
or finite:
```swift
.animation(
    .easeInOut(duration: 1)
        .repeatCount(3, autoreverses: true), // Repeats 3 times
    value: isAnimating
)
```

dkwingsmt: I talked with Loic offline and he misunderstood me for what I referred to for `ReversingAnimatable` (I referred to a `Animatable` that is basically a `TweenSequence` concatenating the target Animatable and the same Animatable reversed). Loic agreed that this class would be useful, but we struggled to come up with a good name.

We also agreed that an enum just like Android's `RepeatMode` would be good enough. Although it's technically the same as `bool reverse`, I think a repeat mode makes much more sense on this class semantically. With this option available, while a new `ReversingAnimatable` class would be awesome, it's no longer necessary and can wait for another PR.

What do you think of changing the boolean to a variable with a `RepeatMode` enum?

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: I discussed with Loic offline. Since this class is meant for simpler occasions and should never have an animation controller, I agree that a `paused` is reasonable. Thanks for the discussion.

packages/flutter/lib/src/widgets/tween_animation_builder.dart
dkwingsmt: With `reverse` or `reverseMode` I've rephrased it as follows:
```dart
/// The animation continuously cycles from the value produced by the
/// [Animatable] at 0.0 to the value at 1.0. The [builder] receives the
/// current value produced by this [Animatable] and builds the child from it.
/// When the animation reaches 1.0, it either restarts from 0.0 or reverses
/// direction, based on the [repeatMode].
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
bernaferrari: yes, sure, I'll do it!

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -212,3 +216,175 @@ class _TweenAnimationBuilderState<T extends Object?>
...
+/// the end. When [paused] is true, the animation stops at its current value.
+///
+/// {@macro flutter.widgets.TweenAnimationBuilder.optimizations}
+///
+/// {@tool dartpad}
+/// This example shows a continuously rotating square that can be paused and resumed.
+///
+/// ** See code in examples/api/lib/widgets/tween_animation_builder/repeating_tween_animation_builder.0.dart **
+/// {@end-tool}
+///
```
dkwingsmt: Since this widget is only meant for simpler animations and technically counts as API cliff, developers might want to convert this to a full-blown explicit animation widget as they expand their app. It would be great if we can add a paragraph to explain it with an example of an equivalent app that implements it explicitly. Not necessary for this PR if you don't feel like it.

bernaferrari: I think I'm unable to make a PR without causing dozens of comments and google testing breaking and other issues. I'll try to address this, thanks!

dkwingsmt: It's because you're creating the critical stuff that everyone will use :) Thank you for your contribution. Let me know if you need any help. I'm glad to write docs or examples for you if you feel overwhelmed.

bernaferrari: <p>I pushed your PR comments fixed. While this PR should help a bit, I keep wondering how we can make Flutter animation as pleasant to write as Compose:</p>

<table>
  <tr>
    <td align="center">
      image
    </td>
    <td align="center">
      image
    </td>
  </tr>
</table>

bernaferrari:
I think I addressed every comment. Thanks for all the suggestions. I think I rewrote this PR two or three times already. It is getting better every time.

bernaferrari:
I didn't re-request anything but I guess that's automated when I force pushed to latest. But feel free to review again if/when you want.

Piinks:
Flyby comment from triage: it looks like there are a number of failing checks here.

bernaferrari:
Oh, I think tree was down. I'll rebase.

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
```diff
@@ -0,0 +1,167 @@
...
+import 'ticker_provider.dart';
+import 'transitions.dart';
+import 'value_listenable_builder.dart';
+
+/// Configures how [RepeatingAnimationBuilder] loops its animation.
+enum RepeatMode {
+  /// Each iteration starts over from the beginning once 1.0 is reached.
+  restart,
+
+  /// Each iteration runs forward, then reverses back to the beginning.
```
dkwingsmt, line 17: Can you document whether this means the `duration` is effectively doubled for each actual iteration, or that the curve is calculated at double speed?

(Alternatively we can document it in `RepeatingAnimationBuilder` or even in `RepeatingAnimationBuilder.duration` but I think documenting it here is more useful if this enum is to be used in other widgets. But feel free to do it this way if you find it easier.)

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
```diff
@@ -0,0 +1,167 @@
...
+    this.curve = Curves.linear,
+    this.repeatMode = RepeatMode.restart,
+    this.paused = false,
+    required this.builder,
+    this.child,
+  });
+
+  /// The animatable to drive repeatedly. Typically this is a [Tween] or a
+  /// [TweenSequence], but any [Animatable] that produces values of type [T]
+  /// is accepted.
```
dkwingsmt:
```suggestion
  /// The animatable to drive repeatedly.
  /// 
  /// Typically this is a [Tween] or a [TweenSequence], but any [Animatable]
  /// that produces values of type [T] is accepted.
```

packages/flutter/lib/src/widgets/tween_animation_builder.dart
```diff
@@ -13,6 +13,7 @@ import 'package:flutter/animation.dart';
 
 import 'framework.dart';
 import 'implicit_animations.dart';
+import 'transitions.dart';
```
dkwingsmt: Why is this needed?

dkwingsmt (approved):
Comprehensive tests. Clear documents. Thank you so much!

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
loic-sharma: FYI, this is needed because the first paragraph is what shows up in the IDE. The IDE tooltip is small, so for readability a single sentence is best.

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
```diff
@@ -0,0 +1,167 @@
...
+  final Animatable<T> animatable;
+
+  /// The duration of the animation.
```
loic-sharma, line 63: Consider this to address this comment above: https://github.com/flutter/flutter/pull/174014/files#r2516709026

```suggestion
  /// The duration of the animation.
  ///
  /// If [repeatMode] is [RepeatMode.restart], this is the
  /// duration of the entire animation sequence from 0.0 to 1.0.
  ///
  /// If [repeatMode] is [RepeatMode.reverse], both the
  /// forward segment (0.0 to 1.0) and the backward segment
  /// (1.0 to 0.0) will each take this duration separately.
  /// The total time for one complete forward-and-reverse cycle
  /// will be twice this value.
```

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
loic-sharma: I drafted a suggestion below: https://github.com/flutter/flutter/pull/174014/files#r2519860585

packages/flutter/lib/src/widgets/repeating_animation_builder.dart
```diff
@@ -0,0 +1,167 @@
...
+  final Curve curve;
+
+  /// A builder that creates the animated widget subtree.
+  ///
+  /// The builder is called every time the animation value changes. The optional
+  /// child can be used to avoid unnecessary rebuilds if a part of the subtree
+  /// does not depend on the animation.
+  final ValueWidgetBuilder<T> builder;
+
+  /// An optional widget to pass to the builder.
```
loic-sharma, line 91: `TweenAnimationBuilder.child` has some helpful docs that would be helpful here as well: https://api.flutter.dev/flutter/widgets/TweenAnimationBuilder/child.html

Maybe:

```suggestion
  /// An optional widget to pass to the builder.
  ///
  /// If a builder callback's return value contains a subtree that does not
  /// depend on the animation, it's more efficient to build that subtree once
  /// instead of rebuilding it on every animation tick.
  ///
  /// If the pre-built subtree is passed as the child parameter, the
  /// [RepeatingAnimationBuilder] will pass it back to the [builder]
  /// function so that it can be incorporated into the build.
  ///
  /// Using this pre-built child is entirely optional, but can improve
  /// performance significantly in some cases and is therefore a good practice.
```

loic-sharma (approved):
I left some minor documentation suggestions.

Thank you for doing this! This is truly outstanding work :)

bernaferrari:
I need to find the answer for this. Don't worry. I'll get that later today. Thanks for collaborating over 3 months for this.
image

Issue #166550, opened 2025-04-03
Remove code that manipulates AGP tasks by discovering them via their name


gmackall:
The Flutter Gradle plugin is (by necessity) built on top of AGP. 

AGP today has exposed APIs designed for extending it https://developer.android.com/build/extend-agp#variant-api-artifacts-tasks.

This was not the case when the Flutter Gradle plugin was written. Instead, we discovered tasks that AGP added by guessing their name as a string and then doing things with them (typically making them depend on the Flutter tasks that we were adding). We should remove this string based guessing, and migrate to their intended API.

e.g.
https://github.com/flutter/flutter/blob/e971379436d426324f4a02fe2b1bfdb24a261764/packages/flutter_tools/gradle/src/main/groovy/flutter.groovy#L474-L485

The good news is that I think these attempts to get tasks might already be failing? They seemed to be always null in my brief testing, which means it would be easy to remove them. But that is a guess.

gmackall:
More generally speaking, it is a common pattern that Gradle will let you get something (a task, an extension, etc)
1. by name (string)
2. by type

We should probably generally speaking stop using (1). In Groovy we would sometimes get something by string, and then because Groovy is dynamic we would just call a method that we hoped would exist (without casting). In Kotlin we can't do that, and we probably shouldn't have been doing that anyways. We should just get it in a typesafe way so that we know the methods/fields, and then call them accordingly.

jmagman:
Example: https://github.com/flutter/flutter/pull/173794 migrated deeplink json creation:

https://github.com/flutter/flutter/pull/173794/files#diff-4e0ce399158c429624a8733adc044e057ede914d6f33729d06e6d14ab26003e5L817
https://github.com/flutter/flutter/pull/173794/files#diff-4e0ce399158c429624a8733adc044e057ede914d6f33729d06e6d14ab26003e5L846-L847

PR #174070, opened 2025-08-19
Increase testing coverage and maintainability of android manifest parsing logic

reidbaker:
Related to: #166550
Follow up to https://github.com/flutter/flutter/pull/173794 to implement suggested improvements and with strictly human authored tests changes. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/gradle/src/main/kotlin/IntentFilterCheck.kt
```diff
@@ -8,13 +8,12 @@ import kotlinx.serialization.json.JsonObject
...
-    var hasActionView = false
-    var hasDefaultCategory = false
-    var hasBrowsableCategory = false
-
+class IntentFilterCheck(
+    var hasAutoVerify: Boolean = false,
+    var hasActionView: Boolean = false,
+    var hasDefaultCategory: Boolean = false,
+    var hasBrowsableCategory: Boolean = false
+) {
```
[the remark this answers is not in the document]
camsim99: Is there some reason not to take this change?

packages/flutter_tools/gradle/src/test/kotlin/tasks/DeepLinkJsonFromManifestTaskTest.kt
```diff
@@ -90,68 +95,50 @@ class DeepLinkJsonFromManifestTaskTest {
...
-        val appLinkSettings = DeepLinkJsonFromManifestTaskHelper.createAppLinkSettings(defaultNamespace, manifestFile)
-
-        assertEquals(defaultNamespace, appLinkSettings.applicationId)
-        assertFalse(appLinkSettings.deeplinkingFlagEnabled)
-        assertTrue(appLinkSettings.deeplinks.isEmpty())
+        try {
+            val appLinkSettings = DeepLinkJsonFromManifestTaskHelper.createAppLinkSettings(defaultNamespace, manifestFile)
+            assertEquals(defaultNamespace, appLinkSettings.applicationId)
+            assertFalse(appLinkSettings.deeplinkingFlagEnabled)
+            assertTrue(appLinkSettings.deeplinks.isEmpty())
```
camsim99: Nit: Is the `SAXParseException` so unclear that it deserves a try/catch? Just curious because I'm a bit worried that someone may try to copy the try/catch pattern you use here, forget to not just specifically catch `SAXParseException`s, and then accidentally swallow assertion errors from the assertion tests.

camsim99 (approved):
Overall looks good to me but left some food for thought!

packages/flutter_tools/gradle/src/test/kotlin/tasks/DeepLinkJsonFromManifestTaskTest.kt
reidbaker: Yeah it is super unhelpful and gives you basically nothing to debug with. I kept it focused on the specific exception that happens when the xml is invalid because the xml is being created by test tooling.

reidbaker:
@camsim99 in https://github.com/flutter/flutter/pull/174070#discussion_r2298554446 it requires changing the parsing code  which means they are not val's as currently authored.

packages/flutter_tools/gradle/src/test/kotlin/tasks/DeepLinkJsonFromManifestTaskTest.kt
reidbaker: Thinking of ways to make this more clear. I could duplicate the parse call from the helper in the "build" function such that when we called build it validated the xml was at least parsable. Then if it failed in the builder I could call fail and print. Do you like that modification?

camsim99: I like that structure better!

reidbaker:
@camsim99 I ended up trying out https://github.com/flutter/flutter/pull/174070#discussion_r2298583900 and liked it a lot more.

PR #174088, opened 2025-08-20
Use an alternative to `git describe` for `master` version resolution

### LINKED ISSUE #173904: Flutter `stable` is 3.35.1 while `main` is 3.33.0-1.0.pre-1446
### Steps to reproduce

1. `git clone https://github.com/flutter/flutter -b main`
2. `flutter --version`

### Expected results

I expect the version of Flutter on the `main` branch to be higher than `stable` and `beta`

### Actual results

```console
$ flutter --version
Flutter 3.33.0-1.0.pre-1446 • channel main • https://github.com/flutter/flutter
Framework • revision 5c8c174f3b (12 hours ago) • 2025-08-15 22:03:40 -0400
Engine • hash ae5fc5cddb31f37e8805e57d8385ee43b8e0cced (revision 5c8c174f3b) (12 hours ago) • 2025-08-16 02:03:40.000Z
Tools • Dart 3.10.0 (build 3.10.0-104.0.dev) • DevTools 2.49.0
```

### Code sample

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
flutter doctor -v
[✓] Flutter (Channel main, 3.33.0-1.0.pre-1446, on macOS 15.6 24G84 darwin-arm64, locale en-AU) [1,110ms]
    • Flutter version 3.33.0-1.0.pre-1446 on channel main at /Users/brett/flutter
    • Upstream repository https://github.com/flutter/flutter
    • Framework revision 5c8c174f3b (12 hours ago), 2025-08-15 22:03:40 -0400
    • Engine revision 5c8c174f3b
    • Dart version 3.10.0 (build 3.10.0-104.0.dev)
    • DevTools version 2.49.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging

[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [802ms]
    • Android SDK at /Users/brett/Library/Android/sdk
    • Java version OpenJDK Runtime Environment (build 21.0.3+-79915917-b509.11)

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [491ms]
    • Build 16F6
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [5ms]

[✓] Connected device (2 available) [5.6s]
    • macOS (desktop) • macos  • darwin-arm64   • macOS 15.6 24G84 darwin-arm64
    • Chrome (web)    • chrome • web-javascript • Google Chrome 139.0.7258.128
    ! Error: Browsing on the local area network for Brett’s iPad. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for Brett’s iPhone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.

[✓] Network resources [977ms]

```

</details>

matanlurey:
Closes https://github.com/flutter/flutter/issues/173904.

It's not clear to me how `git describe --tags HEAD` ever ... worked to determine a fallback. From what I can tell, the _intent_ was to use the latest (newest? closest?) tag as the base version, and then append `-{commitCount}-{shortHash}` as the fallback version number when on `master` (or any non-published branch).

So, I rewrote the implementation, unfortunately with 4 separate calls out to `git ...` instead of a single one.

There are 20+ tests that fail as a result of this change, mostly because they make specific expectations around `git describe` being invoked, and of course that is no longer the case - putting those aside, I'd like to double check that:

1. I understand what the original command was _intending_ to do
2. We like the _output_ of the updated command
3. ... we either are okay with the implementation, or have an alternative in mind (I'm no `git` master)

Wdyt?

At this commit:
```sh
$ flutter-dev --version                                                            
Flutter 3.36.0-1.0.pre-170 • channel [user-branch] • https://github.com/matanlurey/flutter
Framework • revision 250381a185 (5 minutes ago) • 2025-08-19 18:57:36 -0700
Engine • hash f278b0aa3b8c6732ab636563eb8e896c35fc9c79 (revision 9ac4facf7e) (2 hours ago) • 2025-08-19 23:42:28.000Z
Tools • Dart 3.10.0 (build 3.10.0-115.0.dev) • DevTools 2.49.0
```

/cc @zanderso @jmagman for historics.


zanderso:
(1) Yes, I think you have it right. When the original logic was written, tagging happened much more frequently since there was a tag for every roll to g3. That was the 'dev' channel, which no longer exists.

(2) The new output is definitely less confusing! AFAIK that string is not load bearing, and it's mostly just us maintainers on master channel.

(3) Only concern from me about the implementation is whether the additional synchronous calls out to git impact tool startup time (or any other operation where we compute the version). Maybe Gemini can suggest a shorter sequence of commands? =)

bkonyi:
This LGTM, although I'm also no Git master and have no idea if there's a better way to do this.

matanlurey:
I was able to reduce the number of git calls by one, and updated tests. PTAL @bkonyi @jtmcdole

reidbaker (approved):

matanlurey:
@stuartmorgan-g brings up a good point (in another chat/thread):

> Well, without being able to jump to before a branch point all I know is it worked at head on my machine. If you gave me a version that just always picked the numerically highest tag in the repo it would have passed the manual testing I did.

@matanlurey:

> I mean you bring up a point - that is essentially what it does - and I think if you did checkout an older commit it would still pretend to be the latest tag. I don't know if there is a way to say, given a ref, give me the closest published tag that my ref is otherwise not associated with.
>
> Maybe there is a way to say, "list all tags older than my current commit" - let me take a look

bkonyi (approved):

matanlurey:
Re-thinking this. Will send out for review again when ready.

matanlurey:
Ok here is my attempt based on some offline discussion with @stuartmorgan-g and others.

It's a lot less hacky, but will require a single manual step by the release team. If we like this change, I'll modify the release team documentation and let the team know (in practice I think this is fine, and we could even help automate it at a future point).

matanlurey:
Ugh, I missed something.

My original workaround ("use the highest tag number") _only_ kicks in if a commit is so new it's not part of any release. That means, it _was_ valid. I'm going to revert back to that point as it in general was more automated.

bkonyi (approved):

PR #174090, opened 2025-08-20
Refactor: Migrate page transition builder to widgets

rkishan516:
Changes:
* Move PageTransitionsBuilder and its implementation classes like FadeUpwardsPageTransitionsBuilder, OpenUpwardsPageTransitionsBuilder, FadeForwardsPageTransitionsBuilder, ZoomPageTransitionsBuilder, CupertinoPageTransitionsBuilder to widgets
* Move PredictiveBackPageTransitionsBuilder to widgets

part of: #172929 

## Pre-launch Checklist

- [x] I signed the [CLA].


rkishan516:
Current implementation still contains
```dart
import '../material/color_scheme.dart';
import '../material/theme.dart';
```

as import in `page_transition_theme.dart`.

victorsanni:
> Current implementation still contains
> 
> ```dart
> import '../material/color_scheme.dart';
> import '../material/theme.dart';
> ```
> 
> as import in `page_transition_theme.dart`.

`widgets/` can't import `material/`, so this can't be a straightforward move to `widgets/`. I think the transition builders aren't tied to `material/`, but the page transitions theme (and any other theming stuff) probably is, so should remain.

justinmc (commented):
I appreciate you jumping on this @rkishan516!

+1 to what @victorsanni said: https://github.com/flutter/flutter/pull/174090#issuecomment-3207420616

I just wanted to chime in to say that I'm exploring adding a widgets/system_ui directory for this kind of stuff. Depending on if or what you are able to move out of Material, that might be the long term destination for it.

rkishan516:
> widgets/ can't import material/, so this can't be a straightforward move to widgets/. I think the transition builders aren't tied to material/, but the page transitions theme (and any other theming stuff) probably is, so should remain.

Actually @victorsanni, Its both PageTransitionBuilder and PageTransitionTheme that depends on MaterialTheme right now.

> I just wanted to chime in to say that I'm exploring adding a widgets/system_ui directory for this kind of stuff. Depending on if or what you are able to move out of Material, that might be the long term destination for it.

Sure @justinmc, I can keep this MR on hold until `system_ui` is created.

MitchellGoodwin:
> Its both PageTransitionBuilder and PageTransitionTheme that depends on MaterialTheme right now.

It's a hard requirement that the Widgets library can't import Material or Cupertino, or else it creates a circular dependency. I do not think there is a way to move PageTransitionTheme down without causing issues right now, so that should stay in Material. Ongoing investigations are happening around theming options in the Widget layer and that may provide a way to do that later.

So what should move down is PageTransitionsBuilder and the classes that extend from it. That still leaves the dependency on Theme and ColorScheme. Both are used to get a default background color for the transitions that need it. That gets partially resolved by adding a `backgroundColor` property to two of the builders. I suggest moving the `backgroundColor` property to PageTransitionBuilder so it can be used by all extended classes, and removing the theme look up from the builders. Then PageTransitionTheme, which is still in Material can see if `backgroundColor` is null and if so, it can set the `backgroundColor` using the Material theme.

Also the Cupertino transition builder will need the relevant code from Cupertino to move down as well.

rkishan516:
@MitchellGoodwin I have updated PR based on your suggestion, but when backgroundColor is not given in PageTransitionBuilder, I had to fallback to transparent. 
Instances where transparent was used :-
1. FadeForwardsPageTransitionsBuilder.delegatedTransition
2. FadeForwardsPageTransitionsBuilder.buildTransitions
3. ZoomPageTransitionsBuilder.delegatedTransition
4. ZoomPageTransitionsBuilder.buildTransitions

victorsanni:
Hi @rkishan516, can you draft this PR and open separate PRs migrating these builders individually? Will make it much easier to review.

rkishan516:
Sure @victorsanni, Will do.

rkishan516:
@victorsanni, 1st PR for migrate through breaking this PR.
https://github.com/flutter/flutter/pull/174321

rkishan516:
@victorsanni Next PR for this PR train is https://github.com/flutter/flutter/pull/175560.

victorsanni:
Next PR for this is https://github.com/flutter/flutter/pull/177080.

PR #174156, opened 2025-08-20
Report a correct display ID in the window metrics event on win32

mattkae:
## What's new?
Now that we're sending down "true" display information to the framework, we also need to send down the correct display id in the WindowMetricsEvent for the view so that the display resolves properly. This pull request adds that :)

## How to test
Here's a silly app that makes sure that our `Display` is legit:

```dart
// Copyright 2014 The Flutter Authors. All rights reserved.
// found in the LICENSE file.

void main() => runApp(W());

class W extends StatelessWidget {
  @override
    final x = View.of(context);
    print(x.display.id);
    return const Center(
      child: Text('Hello, world!', key: Key('title'), textDirection: TextDirection.ltr),
    );
  }
}
```

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/windows/flutter_window.cc
```diff
@@ -305,6 +308,22 @@ PointerLocation FlutterWindow::GetPrimaryPointerLocation() {
...
+    if (display.display_id ==
+        reinterpret_cast<FlutterEngineDisplayId>(monitor)) {
+      return display;
+    }
+  }
+
+  FML_LOG(ERROR) << "This window is not associated with any monitor";
+  const auto display = display_monitor_->FromMonitor(monitor);
+  FML_CHECK(display != std::nullopt);
+  return display.value();
```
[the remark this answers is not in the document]
mattkae: ty gemini

engine/src/flutter/shell/platform/windows/flutter_window.cc
```diff
@@ -76,9 +77,11 @@ static uint64_t ConvertWinButtonToFlutterButton(UINT button) {
 FlutterWindow::FlutterWindow(
     int width,
     int height,
+    std::shared_ptr<DisplayMonitor> const& display_monitor,
```
loic-sharma, line 80: Hm maybe we should rename this to `DisplayManager`. "Monitor" is a bit on an overloaded term here, and we're using it for more than just listening to changes.

mattkae: Could we do that in another PR? Just want to limit the scope, I can add that to my list as a follow up

engine/src/flutter/shell/platform/windows/flutter_window.cc
```diff
@@ -305,6 +308,22 @@ PointerLocation FlutterWindow::GetPrimaryPointerLocation() {
...
+  HMONITOR monitor =
+      MonitorFromWindow(GetWindowHandle(), MONITOR_DEFAULTTONEAREST);
+  for (auto const& display : display_monitor_->GetDisplays()) {
+    if (display.display_id ==
+        reinterpret_cast<FlutterEngineDisplayId>(monitor)) {
+      return display;
+    }
+  }
+
+  FML_LOG(ERROR) << "This window is not associated with any monitor";
```
loic-sharma: Just curious, if this happens would it be because Flutter Windows has a bug? If so, should we just crash?

loic-sharma: On second thought, what do you think of:

1. Changing this method to return just a `FlutterEngineDisplayId`
2. Add a `std::optional<FlutterEngineDisplay> DisplayMonitor::FindById(FlutterEngineDisplayId id)` method
3. In debug mode, use the display monitor to verify the display ID is a known value. If it is not, report the error message

I think that makes the fallback case a little less funky.

engine/src/flutter/shell/platform/windows/flutter_window.cc
loic-sharma: Yup that sounds good to me 👍

engine/src/flutter/shell/platform/windows/display_monitor.h
```diff
@@ -31,6 +31,9 @@ class DisplayMonitor {
                            LPARAM lparam,
                            LRESULT* result);
 
+  // Finds the display infomration associated with the id.
```
loic-sharma:
```suggestion
  // Finds the display information associated with the id.
```

engine/src/flutter/shell/platform/windows/flutter_window.cc
```diff
@@ -305,6 +308,18 @@ PointerLocation FlutterWindow::GetPrimaryPointerLocation() {
...
+FlutterEngineDisplayId FlutterWindow::GetDisplayId() {
+  FlutterEngineDisplayId const display_id =
+      reinterpret_cast<FlutterEngineDisplayId>(
+          MonitorFromWindow(GetWindowHandle(), MONITOR_DEFAULTTONEAREST));
+  if (auto const display = display_monitor_->FindById(display_id)) {
+    return display->display_id;
+  }
+
+  FML_LOG(ERROR) << "Current monitor not found in display list.";
+  return display_id;
```
loic-sharma: Can this be simplified?

```suggestion
  if (!display_monitor_->FindById(display_id)) {
    FML_LOG(ERROR) << "Current monitor not found in display list.";
  }

  return display_id;
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -381,12 +381,14 @@ void FlutterWindowsView::SendWindowMetrics(size_t width,
...
   PhysicalWindowBounds bounds = binding_handler_->GetPhysicalWindowBounds();
   double pixel_ratio = binding_handler_->GetDpiScale();
+  FlutterEngineDisplayId display_id = binding_handler_->GetDisplayId();
 
   FlutterWindowMetricsEvent event = {};
   event.struct_size = sizeof(event);
   event.width = bounds.width;
   event.height = bounds.height;
   event.pixel_ratio = pixel_ratio;
+  event.display_id = display_id;
```
loic-sharma, line 391: Is there a test somewhere that verifies viewport metrics get a display ID?

loic-sharma (approved):
Thanks for fixing this! LGTM but please make sure there's a test somewhere that the display ID is plumbed to the window metric: https://github.com/flutter/flutter/pull/174156/files#r2289414502

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
mattkae: I will add one!

Issue #104117, opened 2022-05-18
Provide a way for plugins to know what window to interact with


stuartmorgan:
Several of our 1P plugins present view controllers over the Flutter view; `GIDSignIn` needs this information, for example. Currently, we do this using `UIApplication.sharedApplication.keyWindow`, but in iOS 13 `keyWindow` is deprecated because of the introduction of scenes on iPads.

To make Flutter scene-friendly, we should provide an alternate way of getting that information. On the macOS embedding, for instance, I added a way to get the `NSView` of the Flutter content, which allows plugins to get the window they are part of for things like showing sheets. (I keep finding 3P plugins that ignore that and use the key window instead, but 🤷🏻 ). We should consider doing the same for iOS, or providing some alternate solution.

/cc @jmagman

stuartmorgan:
I'm going to temporarily suppress these warnings with a TODO referencing this issue, so once we have a solution we should look for this issue number in flutter/plugins and update the relevant code.

stuartmorgan:
> On the macOS embedding, for instance, I added a way to get the `NSView` of the Flutter content via the plugin registrar

This API is being adjusted as part of multi-window support; we should take those adjustments into account when designing a corresponding iOS API.

cyanglaz:
> On the macOS embedding, for instance, I added a way to get the NSView of the Flutter content via the plugin registrar, which allows plugins to get the window they are part of for things like showing sheets.

This is the API on macOS: https://github.com/flutter/engine/blame/main/shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h#L45

Multi-view version: 
https://github.com/flutter/engine/blame/main/shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h#L50

jmagman:
See cl/760657609 for an internal use-case where some hackery of plugins was needed.

stuartmorgan-g:
If we're okay adding the non-multi-friendly version of this on iOS (`view`, vs the newer `viewWithIdentifier:`), this should be trivial to add.

stuartmorgan-g:
Based on offline discussion, for iOS we probably want an API to get the view controller rather than the view, since the APIs we're dealing with mostly want view controllers.

vashworth:
macOS version:

[...]

PR #174168, opened 2025-08-20
Add a `viewController` property to the ios/macOS `FlutterPluginRegistrar` protocol 

LongCatIsLooong:
#104117

Adds a `viewController` (per https://github.com/flutter/flutter/issues/104117#issuecomment-3114772381, presumably because iOS plugins want to presenting new view controllers) getter that makes the single-view assumption. ~This is a method instead of a weak + readonly property like the macOS registrar protocol because existing getters are also methods, e.g.,~
```objc
- (NSObject<FlutterBinaryMessenger>*)messenger;
- (NSObject<FlutterTextureRegistry>*)textures;
```

~Unfortunately swift callers will have to use a preprocess macro and add different code paths for macOS and iOS.~

#### Single view assumption

I've considered adding a `- (nullable UIViewController*)viewControllerForViewID: ViewIDType;` method but that will force plugins to make breaking changes in order to migrate off of the deprecated `keyWindow` property. Take the text input plugin as an example, it has to add a `viewID` argument to the `setTextInputClient` call when establishing a new text input connection, so the text input plugin knows which view hierarchy to add the text input view to. It is unclear to me whether more `FlutterPluginRegistrar` changes are needed to completely get rid of the single-view assumption in the protocol (the views `FlutterPlatformViewFactory` creates are always added as subviews to `flutterView` so I guess that interface may have to change in the future? According to according to [Flutter’s multi-window status](https://docs.google.com/document/d/13E27tD8_9f6lDgwg3MpGNTV8XIRCZH3ByI-t9kI9IUM/edit?tab=t.0#heading=h.8lxwn1tiky70), only a `viewForIdentifier` method will be added to registrar), so IMO it would be a better experience to not introduce the `viewID` concept right now and force plugins to make break changes, since they may have to make more breaking changes in the future.

Alternatively, we could also ignore the `viewID` args for now and always return the view controller of the implicit view in the implementation, so that plugin authors can use made-up view IDs to get the implicit view controller. But that's going to make future multiview migrations difficult since the compiler won't be able to catch unmigrated code if the plugin is using a bogus view ID.

##  Pre-launch Checklist

- [ ] I signed the [CLA].


LongCatIsLooong:
Hi @loic-sharma according to [Multi-view runner APIs](https://docs.google.com/document/d/1WkJuzifgJXyceaSoCCu8qTETYmYBOYKoz7Qaxtsore4/edit?tab=t.0#heading=h.le3phqrnhtzi) the only plugin registrar change required to add multiview support to the macOS plugin registrar is to add a get view by id method (which is pretty similar to what this PR does but for macOS). But adding that method now will require some plugins to add a `viewID` parameter to their API (in order to remove the deprecated `keyWindow` usages from their codebase), are you OK with adding the single view version first and deprecate this in favor of the multi view version in the future?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPlatformViews.h
```diff
@@ -34,7 +34,7 @@ FLUTTER_DARWIN_EXPORT
  * @param frame The rectangle for the newly created `UIView` measured in points.
  * @param viewId A unique identifier for this `UIView`.
  * @param args Parameters for creating the `UIView` sent from the Dart side of the Flutter app.
- *   If `createArgsCodec` is not implemented, or if no creation arguments were sent from the Dart
+ *   If `createArgsCodec` is not implemented, or if no creation arguments were sent from the Dart
```
LongCatIsLooong, line 37: (unrelated change, removed an invisible character)

loic-sharma:
> ... are you OK with adding the single view version first and deprecate this in favor of the multi view version in the future?

For iOS that seems reasonable, yes. Plumbing view IDs around is a significant/painful migration, so it seems reasonable to do separate migrations for `keyWindow` and single view assumptions.

LongCatIsLooong:
To answer my own question regarding whether the platform view factory registration methods on the registrar interface would need a view ID parameter: looks like each layer (including those representing platform views) knows which view it is supposed to render on, so it probably doesn't need a view ID.

https://github.com/flutter/flutter/blob/5c23f4c7c3ea9139eec2a05ec1cd01ff6dd3f419/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm#L71-L132

vashworth:
> Unfortunately swift callers will have to use a preprocess macro and add different code paths for macOS and iOS.

@stuartmorgan-g @LongCatIsLooong What do you think of adding a `viewController` property to the macOS version as well to keep them uniform?

stuartmorgan-g:
> This is a method instead of a weak + readonly property like the macOS registrar protocol because existing getters are also methods

AFAIK that's only because the API predates properties, and nobody thought to change it before the introduction of Swift made it a breaking change to do so. I would recommend making it a property; IMO new code should follow current best practices even when older code doesn't. We already have an open issue for trying to migrate the old methods to properties in a non-breaking way.

stuartmorgan-g:
> What do you think of adding a `viewController` property to the macOS version as well to keep them uniform?

The obvious argument against is that it adds a new API surface that will be quasi-deprecated from the start. That said, it doesn't add anything conceptually new over `view`, so as long as we also add `viewControllerForViewIdentifier:` on macOS at the same time as `viewController`, I think the consistency argument is stronger than the deprecation counter-argument.

LongCatIsLooong:
> > What do you think of adding a `viewController` property to the macOS version as well to keep them uniform?
> 
> The obvious argument against is that it adds a new API surface that will be quasi-deprecated from the start. That said, it doesn't add anything conceptually new over `view`, so as long as we also add `viewControllerForViewIdentifier:` on macOS at the same time as `viewController`, I think the consistency argument is stronger than the deprecation counter-argument.

Ah adding `viewController` and `viewControllerForViewID:` together to both platforms so people can decide which one to use / when to migrate? Sounds good.

stuartmorgan-g:
> Ah adding `viewController` and `viewControllerForViewID:` together to both platforms so people can decide which one to use / when to migrate? Sounds good.

I thought the plumbing didn't exist on the iOS side to properly implement the ID-based version, in which case we shouldn't add it there yet.

Actually, looking at macOS headers, do we not have the view-by-identifier method yet? I thought it had been added, but if not, then we don't need to add it here. I just didn't want to get into a situation where the view controller version was *less* capable than the existing API.

LongCatIsLooong:
> > Ah adding `viewController` and `viewControllerForViewID:` together to both platforms so people can decide which one to use / when to migrate? Sounds good.
> 
> I thought the plumbing didn't exist on the iOS side to properly implement the ID-based version, in which case we shouldn't add it there yet.
> 
> Actually, looking at macOS headers, do we not have the view-by-identifier method yet? I thought it had been added, but if not, then we don't need to add it here. I just didn't want to get into a situation where the view controller version was *less* capable than the existing API.

IIRC the macos multi view change was  reverted. I'll remove the multi view stuff then.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngineTest.mm
```diff
@@ -189,6 +190,31 @@ - (void)testNotifyPluginOfDealloc {
...
+  FlutterEngine* engine = [[FlutterEngine alloc] initWithName:@"engine" project:project];
+  id mockEngine = OCMPartialMock(engine);
+  NSObject<FlutterPluginRegistrar>* registrar = [mockEngine registrarForPlugin:@"plugin"];
+
+  // Verify accessing the viewController getter calls FlutterEngine.viewController.
+  (void)[registrar viewController];
+  OCMVerify(times(1), [mockEngine viewController]);
+}
+
+- (void)testGetViewControllerByViewIDFromRegistrar {
```
vashworth: This test should be removed since `viewControllerForViewIdentifier` was removed

vashworth:
Looks like `FlutterMenuPluginTest.mm` needs to be updated:

```
../../../flutter/shell/platform/darwin/macos/framework/Source/FlutterMenuPluginTest.mm:27:17: error: auto property synthesis will not synthesize property 'viewController' declared in protocol 'FlutterPluginRegistrar' [-Werror,-Wobjc-protocol-property-synthesis]
    27 | @implementation FakePluginRegistrar
       |                 ^
 ../../../flutter/shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h:64:49: note: property declared here
    64 | @property(nullable, readonly) NSViewController* viewController;
       |                                                 ^
 ../../../flutter/shell/platform/darwin/macos/framework/Source/FlutterMenuPluginTest.mm:61:1: note: add a '@synthesize' directive
    61 | @end
       | ^
 1 error generated.
 ```

vashworth (approved):
LGTM after tests are fixed

LongCatIsLooong:
~Waiting for cl/800143229 and cl/800146189 to be merged.~

waiting on https://github.com/flutter/packages/pull/9922 and https://github.com/flutter/packages/pull/9923

Issue #173446, opened 2025-08-08
Add semantics for CupertinoExpansionTile


victorsanni:
### Use case

Similar to ExpansionTile, CupertinoExpansionTile should announce its expansion state for assistive technologies (voiceover).

### Proposal

Announce expansion state when expanded/collapsed.

chunhtai:
Maybe we should move the semantics along with the localized label to Expandable widget in widget layer.

PR #174169, opened 2025-08-20
[A11y] Announce Expansible expansion status

victorsanni:
Part of [Add semantics for CupertinoExpansionTile](https://github.com/flutter/flutter/issues/173446)


victorsanni:
This is not a breaking change for localization.

packages/flutter/lib/src/widgets/expansible.dart
```diff
@@ -340,9 +349,34 @@ class _ExpansibleState extends State<Expansible> with SingleTickerProviderStateM
...
+    _timer = null;
     super.dispose();
   }
 
+  void _announceSemantics() {
+    if (!MediaQuery.supportsAnnounceOf(context)) {
+      return;
+    }
+    final WidgetsLocalizations localizations = WidgetsLocalizations.of(context);
+    final String stateHint = widget.controller.isExpanded
```
chunhtai: Consider pass-in the boolean or message directly. or refactor out something like
```
bool get _effectiveExpanded => widget.controller.isExpanded;
```
in case we change how we decide _effectiveExpanded later

chunhtai (approved):
LGTM, just a minor suggestion

victorsanni:
Blocked on https://github.com/flutter/flutter/pull/174365

hannah-hyj (approved):
LGTM

packages/flutter/lib/src/widgets/expansible.dart
```diff
@@ -340,13 +351,36 @@ class _ExpansibleState extends State<Expansible> with SingleTickerProviderStateM
     widget.controller.removeListener(_toggleExpansion);
     _animationController.dispose();
     _heightFactor.dispose();
+    _timer?.cancel();
+    _timer = null;
     super.dispose();
   }
 
+  void _announceSemantics() {
```
ash2moon, line 359: Instead of using the announcement service for this, can we use a combination of live region and the `expanded` hint? I am also working on this in another PR to do exactly that but in the `ExpansionTile`.

Piinks:

PR #174184, opened 2025-08-21
Skip wasm build when dry run is disabled and --wasm is not specified.

biggs0125:
This fixes an issue where the wasm build was still getting executed when `--no-wasm-dry-run` was specified.

Also adds a test to verify the expected behavior.


mdebbar (approved):
Thanks for the quick fix!

This probably needs to be CP'd, right?

biggs0125:
> Thanks for the quick fix!
> 
> This probably needs to be CP'd, right?

Yes, it should be.

mdebbar:
Apparently, the `cp: stable` doesn't works if it's added before the PR is merged 🤷‍♂️

I removed and re-added it. Here you go: https://github.com/flutter/flutter/pull/174263 @biggs0125

PR #172314, opened 2025-07-17
[skwasm] Port to `DisplayList` objects

eyebrowsoffire:
This PR refactors the skwasm renderer to use `DisplayList` objects as its main model objects instead of using Skia objects directly. Then, at render time, we dispatch the display list commands to the skia surface. This is a preparatory step for impeller on web.
* Some build rules were reworked in order to allow `DisplayList` to compile via emscripten
* Some pieces of the display list library were further refactored to allow us to compile it without actually building and linking the impeller shaders. The two major classes that needed to be separated out were `DlRuntimeEffect` and the text drawing system.
* `SkPath` and `SkImage` are still used as the main model objects in skwasm. As of right now, `DisplayList` just thinly wraps these objects, so this is the minimal possible change for now. I will have to refactor this somewhat further when preparing for actual impeller adoption.
* Several special cased code paths in skwasm were removed, as they are taken care of by `DisplayList` itself. This includes shadow drawing, determining when to enable dithering, and determining the right clamp value for filters.

PR #174219, opened 2025-08-21
Refactor text and runtime effect to separate skia and impeller implementations.

eyebrowsoffire:
This makes two refactors to the display list architecture:
* The concrete implementations of `DlRuntimeEffect` for skia and impeller are separated, and the impeller implementation put into the `impeller/display_list` target. This makes sure that a client can link against the main `display_list` library without actually pulling in all of impeller. (This is needed for https://github.com/flutter/flutter/pull/172314)
* The `DrawTextBlob` and `DrawTextFrame` methods are consolidated into one `DrawText` call, and that takes a `DlText` object. The `DlText` object has two implementations, one for skia and one for impeller, and the impeller one is moved into `impeller/display_list` for the same reason mentioned above.


.vscode/settings.json
```diff
@@ -5,5 +5,57 @@
     "githubPullRequests.ignoredPullRequestBranches": [
         "master"
     ],
-    "files.trimTrailingWhitespace": true
+    "files.trimTrailingWhitespace": true,
+    "files.associations": {
```
flar: Is this part of the other PR? It doesn't make sense as part of this PR unless I've missed something...

engine/src/flutter/display_list/dl_op_records.h
```diff
@@ -1001,35 +1000,24 @@ struct DrawDisplayListOp final : DrawOpBase {
...
 
-  const DlScalar x;
-  const DlScalar y;
-  const std::shared_ptr<impeller::TextFrame> text_frame;
-
-  void dispatch(DlOpReceiver& receiver) const {
-    receiver.drawTextFrame(text_frame, x, y);
+  DisplayListCompare equals(const DrawTextOp* other) const {
+    return text->getTextBlob() == other->text->getTextBlob() &&
+                   text->getTextFrame() == other->text->getTextFrame() &&
```
flar: Perhaps put this pointer comparison inside a `DlText::==` operator to hide the details (and in case we eventually have some better way to compare them than by (pair of) subsumed pointer(s)?)

engine/src/flutter/display_list/dl_text.h
```diff
@@ -0,0 +1,33 @@
...
+namespace impeller {
+class TextFrame;
+}
+
+class SkTextBlob;
+
+namespace flutter {
+class DlText {
+ public:
+  virtual DlRect getBounds() const = 0;
```
flar: I know there is a mixed style in DL objects, but I think we are moving towards capitalized getters like GetBounds, GetTextFrame, etc. I wasn't doing that until I started integrating with Impeller, but I think it is better matched to the C++ style guidelines so I'd like to move in that direction.

Technically you can have a strict getter that is lower case, but I don't think either of getTextFoo() are "strict getters" since they are only defined for specific subclasses...

engine/src/flutter/display_list/dl_text.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_DISPLAY_LIST_DL_TEXT_H_
+#define FLUTTER_DISPLAY_LIST_DL_TEXT_H_
+
+#include "flutter/display_list/geometry/dl_geometry_types.h"
+#include "third_party/skia/include/core/SkRefCnt.h"
```
flar: I think we can change the getter to SkTextBlob* and we won't need SkRefCnt included here...?

engine/src/flutter/display_list/dl_text.h
```diff
@@ -0,0 +1,33 @@
...
+}
+
+class SkTextBlob;
+
+namespace flutter {
+class DlText {
+ public:
+  virtual DlRect getBounds() const = 0;
+  virtual std::shared_ptr<impeller::TextFrame> getTextFrame() const = 0;
+  virtual sk_sp<SkTextBlob> getTextBlob() const = 0;
```
flar: One of the magic things about sk_sp is that you can just use a raw pointer in cases where you are just passing it along (because the ref count is built-in and there is free "shared from this"). So, this could return SkTextBlob*. Indeed, if you look at SkCanvas::drawTextBlob, the real method takes the ptr and the version that takes an sk_sp does a get() on it and passes it to the pointer version...

engine/src/flutter/display_list/dl_text_skia.cc
```diff
@@ -0,0 +1,14 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "flutter/display_list/dl_text_skia.h"
+
+namespace flutter {
+std::shared_ptr<DlTextSkia> DlTextSkia::Make(const sk_sp<SkTextBlob>& blob) {
```
flar, line 8: We could consider having Blob* factory and avoid bringing sk_sp in, but callers would have to do a `.get()` if they have an sk_sp already. We could have both, but, well, I think I'm overthinking this all. I think this is only ever called when we have the sk_sp anyway.

This is fine. Just maybe pointing out some sk_sp-isms for general knowledge...

engine/src/flutter/display_list/effects/dl_runtime_effect.h
```diff
@@ -8,21 +8,17 @@
...
 #include "flutter/fml/macros.h"
-#include "flutter/impeller/runtime_stage/runtime_stage.h"
-
 #include "third_party/skia/include/core/SkRefCnt.h"
-#include "third_party/skia/include/effects/SkRuntimeEffect.h"
+
+class SkRuntimeEffect;
+namespace impeller {
+class RuntimeStage;
+}
```
flar, line 16: This is not new to your changes, but it occurs to me that we can get away with SkRE* here in the getter and avoid infecting all users of this wrapper class with the need to understand SkRefCnt. It's only used in one place (2 calls) in dk_sk_conversions which work whether we have a pointer or an sk_sp and we save ourselves 2 inc/decrefs...

engine/src/flutter/display_list/skia/dl_sk_dispatcher.cc
```diff
@@ -315,19 +315,14 @@ void DlSkCanvasDispatcher::drawDisplayList(
...
   canvas_->restoreToCount(restore_count);
 }
-void DlSkCanvasDispatcher::drawTextBlob(const sk_sp<SkTextBlob> blob,
-                                        DlScalar x,
-                                        DlScalar y) {
+void DlSkCanvasDispatcher::drawText(const std::shared_ptr<DlText>& text,
+                                    DlScalar x,
+                                    DlScalar y) {
+  auto blob = text->getTextBlob();
+  FML_CHECK(blob) << "Impeller DlText cannot be drawn to a Skia canvas.";
```
flar, line 322: Like the more descriptive comment.

engine/src/flutter/display_list/testing/dl_test_snippets.cc
```diff
@@ -1084,25 +1089,46 @@ std::vector<DisplayListInvocationGroup> CreateAllRenderingOps() {
...
+              r.drawText(DlTextSkia::Make(GetTestTextBlob(1)), 10, 20);
             }},
-           {1, 24, 1,
+           {1, 32, 1,
+            [](DlOpReceiver& r) {
+              r.drawText(DlTextSkia::Make(GetTestTextBlob(2)), 10, 10);
+            }},
+       }},
+#if IMPELLER_SUPPORTS_RENDERING
+      {"DrawText",
```
flar: (Still while being `#if`d), move these cases into the previous block so we get ==/!= checks between calls that differ in Skia/Impeller-ism. Also, the fact that both blocks are labeled "DrawText" makes it hard to know which test case failed when it reports "bad thingy in DrawText variant 3".

flar: So instead of 2 blocks both named "DrawText", we have 1 block that only has 4 cases for Skia and 8 cases for Skia+Impeller...

engine/src/flutter/flow/surface.h
```diff
@@ -12,6 +12,8 @@
 #include "flutter/flow/surface_frame.h"
 #include "flutter/fml/macros.h"
 
+#include "third_party/skia/include/core/SkData.h"
```
flar, line 15: Do we need this for the DlText changes?

engine/src/flutter/impeller/display_list/dl_dispatcher.cc
```diff
@@ -1061,44 +1043,52 @@ void FirstPassDispatcher::skew(DlScalar sx, DlScalar sy) {
...
-    matrix_ = matrix_ * Matrix::MakeColumn(
-        mxx,  myx,  0.0f, 0.0f,
-        mxy,  myy,  0.0f, 0.0f,
-        0.0f, 0.0f, 1.0f, 0.0f,
-        mxt,  myt,  0.0f, 1.0f
-    );
-  }
+// 2x3 2D affine subset of a 4x4 transform in row major order
+void FirstPassDispatcher::transform2DAffine(
+  DlScalar mxx, DlScalar mxy, DlScalar mxt,
```
flar: I think your indent is off by 2 for these. Argument continuations are indented 4 spots, not 2. Same for FullPerspective below...

engine/src/flutter/display_list/display_list_unittests.cc
```diff
@@ -1224,7 +1230,8 @@ TEST_F(DisplayListTest, SingleOpsMightSupportGroupOpacityBlendMode) {
     static auto display_list = builder.Build();
     RUN_TESTS2(canvas.DrawDisplayList(display_list);, false);
   }
-  RUN_TESTS2(canvas.DrawTextBlob(GetTestTextBlob(1), 0, 0, paint);, false);
+  RUN_TESTS2(canvas.DrawText(DlTextSkia::Make(GetTestTextBlob(1)), 0, 0, paint);
```
flar, line 1233: There is no test for Impeller Text here (and the one test for it down below is testing a different condition). While the builder code is the same for both, it couldn't hurt to have both tested here so that when we eventually delete Skia we aren't left without any "text" case for this test.

engine/src/flutter/display_list/display_list_unittests.cc
```diff
@@ -3516,7 +3523,7 @@ TEST_F(DisplayListTest, NopOperationsOmittedFromRecords) {
           builder.DrawAtlas(kTestImage1, xforms, rects, nullptr, 2,
                             DlBlendMode::kSrcOver, DlImageSampling::kLinear,
                             nullptr, &paint);
-          builder.DrawTextBlob(GetTestTextBlob(1), 10, 10, paint);
+          builder.DrawText(DlTextSkia::Make(GetTestTextBlob(1)), 10, 10, paint);
```
flar, line 3531: Also need a test to make sure that DrawText(Impeller) is also nop'd.

engine/src/flutter/display_list/display_list_unittests.cc
```diff
@@ -5505,13 +5514,16 @@ TEST_F(DisplayListTest, BoundedRenderOpsDoNotReportUnbounded) {
     ASSERT_LT(blob->bounds().width(), draw_rect.GetWidth());
     ASSERT_LT(blob->bounds().height(), draw_rect.GetHeight());
 
+    auto text = DlTextImpeller::Make(frame);
```
flar: Or, makefromblob above instead of blob->frame->text.

flar (changes requested):
Some things I should have noted before, but most of them are style and better testing related...

engine/src/flutter/display_list/dl_text_skia.cc
flar: (In other words, read for edification, but ignore in practice)

engine/src/flutter/txt/src/skia/paragraph_skia.cc
```diff
@@ -120,8 +120,10 @@ class DisplayListParagraphPainter : public skt::ParagraphPainter {
     }
     std::shared_ptr<DlText> text;
     if (impeller_enabled_) {
+#if IMPELLER_SUPPORTS_RENDERING
```
flar, line 122: Should this have the if inside it as well? Otherwise we might have a mismatch? Or is that possible?

.vscode/settings.json
eyebrowsoffire: Gahhh this freaking file keeps changing itself and I accidentally added it. I will unstage this change.

engine/src/flutter/flow/surface.h
eyebrowsoffire: `SurfaceData` below uses an `SkData`. I think we previously got away with it through some kind of transitive include.

engine/src/flutter/txt/src/skia/paragraph_skia.cc
eyebrowsoffire: I don't think it's possible for that to happen, but I can switch this up to cover the whole if statement.

engine/src/flutter/display_list/effects/dl_runtime_effect.h
eyebrowsoffire: The wrapper class inherits from `SkRefCnt` itself though, so just changing the getter won't allow me to get rid of that inclusion. We'd have to switch `DlRuntimeEffect` to use a shared_ptr everywhere instead, and that feels a bit out-of-scope for this.

flar: Oooh. Didn't see that. D'oh! Well, we're headed that way anyway...

engine/src/flutter/display_list/dl_op_records.h
```diff
@@ -1013,9 +1013,7 @@ struct DrawTextOp final : DrawOpBase {
   void dispatch(DlOpReceiver& receiver) const { receiver.drawText(text, x, y); }
 
   DisplayListCompare equals(const DrawTextOp* other) const {
-    return text->GetTextBlob() == other->text->GetTextBlob() &&
-                   text->GetTextFrame() == other->text->GetTextFrame() &&
-                   x == other->x && y == other->y
+    return text == other->text && x == other->x && y == other->y
```
flar: I think this only compares the shared pointers. You can use the Equals mechanism (see SetSharedImageFilterOp) which does pointer comparisons and then also *foo == *bar comparisons if needed. You could also do *text == *(other->text), but that gets dangerous if one of them is null. The Equals family of methods protects against that.

flar: `Equals(text, other->text)`

engine/src/flutter/impeller/display_list/dl_dispatcher.cc
flar: These still look off by 2. You got the ones inside the method body, but these argument methods should also be indented by 4 as well.

engine/src/flutter/impeller/display_list/dl_dispatcher.cc
```diff
@@ -1061,44 +1043,52 @@ void FirstPassDispatcher::skew(DlScalar sx, DlScalar sy) {
...
-        mxx, myx, mzx, mwx,
-        mxy, myy, mzy, mwy,
-        mxz, myz, mzz, mwz,
-        mxt, myt, mzt, mwt
-    );
-  }
+// clang-format off
+// full 4x4 transform in row major order
+void FirstPassDispatcher::transformFullPerspective(
+  DlScalar mxx, DlScalar mxy, DlScalar mxz, DlScalar mxt,
```
flar: These argument methods are also indented off by 2.

engine/src/flutter/display_list/dl_op_records.h
```diff
@@ -1013,7 +1013,7 @@ struct DrawTextOp final : DrawOpBase {
   void dispatch(DlOpReceiver& receiver) const { receiver.drawText(text, x, y); }
 
   DisplayListCompare equals(const DrawTextOp* other) const {
-    return text == other->text && x == other->x && y == other->y
+    return *text == *other->text && x == other->x && y == other->y
```
flar: I'd prefer `Equals(text, other->text)`, but this is likely safe in practice.

flar (approved):

engine/src/flutter/display_list/dl_op_records.h
eyebrowsoffire: Yeah I just reread your comment, I switched it to use `Equals()` instead

PR #174229, opened 2025-08-21
Make sure that SimpleDialog and SimpleDialogOption do not crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for SimpleDialog and SimpleDialogOption UI controls.


dkwingsmt (commented):
Can you split them into two tests so that we can verify both size to be exactly zero?

dkwingsmt (approved):
LGTM, thank you!

victorsanni (approved):

PR #174236, opened 2025-08-21
`_downloadArtifacts` (Web SDK) uses content-aware hashing in post-submit

matanlurey:
Towards https://github.com/flutter/flutter/issues/174225.

Will need to get cherry-picked into 3.35 and 3.36.

### DIFF
--- a/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
+++ b/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
@@ -57,7 +57,7 @@ class CopyArtifactsStep implements PipelineStep {
     };
     final Uri url = Uri.https(
       'storage.googleapis.com',
-      '${realmComponent}flutter_infra_release/flutter/${isPostsubmit ? contentHash : gitRevision}/flutter-web-sdk.zip',
+      '${realmComponent}flutter_infra_release/flutter/${realm != LuciRealm.Try ? contentHash : gitRevision}/flutter-web-sdk.zip',
     );
     final http.Response response = await http.Client().get(url);
     if (response.statusCode != 200) {
--- a/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
+++ b/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
@@ -57,7 +57,7 @@ class CopyArtifactsStep implements PipelineStep {
     };
     final Uri url = Uri.https(
       'storage.googleapis.com',
-      '${realmComponent}flutter_infra_release/flutter/$contentHash/flutter-web-sdk.zip',
+      '${realmComponent}flutter_infra_release/flutter/${isPostsubmit ? contentHash : gitRevision}/flutter-web-sdk.zip',
     );
     final http.Response response = await http.Client().get(url);
     if (response.statusCode != 200) {
--- a/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
+++ b/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
@@ -57,7 +57,7 @@ class CopyArtifactsStep implements PipelineStep {
     };
     final Uri url = Uri.https(
       'storage.googleapis.com',
-      '${realmComponent}flutter_infra_release/flutter/${realm == LuciRealm.Prod ? contentHash : gitRevision}/flutter-web-sdk.zip',
+      '${realmComponent}flutter_infra_release/flutter/${realm == LuciRealm.Try ? gitRevision : contentHash}/flutter-web-sdk.zip',
     );
     final http.Response response = await http.Client().get(url);
     if (response.statusCode != 200) {
--- a/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
+++ b/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
@@ -57,7 +57,7 @@ class CopyArtifactsStep implements PipelineStep {
     };
     final Uri url = Uri.https(
       'storage.googleapis.com',
-      '${realmComponent}flutter_infra_release/flutter/${realm != LuciRealm.Try ? contentHash : gitRevision}/flutter-web-sdk.zip',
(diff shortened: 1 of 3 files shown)


eyebrowsoffire (approved):
LGTM, assuming it works!

harryterkelsen (approved):
LGTM

matanlurey:
Looks like this is not working. [Example failure](https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20Engine%20Drone/3531721/overview):

```txt
Running `dart pub get` in 'engine/src/flutter/lib/web_ui'
Pipeline experienced the following failures:
  "copy_artifacts": Could not download flutter-web-sdk.zip from cloud bucket at URL: https://storage.googleapis.com/flutter_archives_v2/flutter_infra_release/flutter/1dc06b550ced0908e59a43c06a26353b1aa2536a/flutter-web-sdk.zip. Response status code: 404
#0      CopyArtifactsStep._downloadArtifacts (file:///b/s/w/ir/cache/builder/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart:64:7)
<asynchronous suspension>
#1      CopyArtifactsStep.run (file:///b/s/w/ir/cache/builder/engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart:96:37)
<asynchronous suspension>
#2      Pipeline.run (file:///b/s/w/ir/cache/builder/engine/src/flutter/lib/web_ui/dev/pipeline.dart:136:9)
<asynchronous suspension>
#3      TestCommand.run (file:///b/s/w/ir/cache/builder/engine/src/flutter/lib/web_ui/dev/test_runner.dart:407:7)
<asynchronous suspension>
#4      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#5      main (file:///b/s/w/ir/cache/builder/engine/src/flutter/lib/web_ui/dev/felt.dart:41:26)
<asynchronous suspension>

Test pipeline failed.
```

The content hash was `1dc06b550ced0908e59a43c06a26353b1aa2536a`.

I'm trying to figure out what previously running process was expected to have uploaded this, but did not.

matanlurey:
This is set:
https://github.com/flutter/flutter/blob/86327198ff0678f46cbdfb3e12f1759d03aa768d/engine/src/flutter/ci/builders/linux_web_engine_build.json#L10-L12

But I don't see where it is uploading

matanlurey:
Here is the [original CL that added support](https://flutter-review.googlesource.com/c/recipes/+/66780) for content-hashed uploads:

```py
if flags.get('upload_content_hash', False):
    content_hash_url = api.repo_util.get_content_hash_url(
        api, path.remote, api.repo_util.content_hash()
    )
    api.archives.upload_artifact(path.local, content_hash_url)
    api.flutter_bcid.upload_provenance(path.local, content_hash_url)
```

... which makes sense. The value we care about comes from `properties.content_hash`:

```py
def content_hash(self):
  """Returns the content hash if present, or None

      The content hash is a source-derived value that fingerprints the artifacts
      with the engine source that produced it. By moving to a content hash, we
      build less copies of the same engine source code and speed up CI.

      See: https://github.com/flutter/flutter/blob/master/bin/internal/content_aware_hash.sh
  """
  return self.m.properties.get('content_hash', None)
```

... but I don't see [that property passed in anywhere here](https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20Production%20Engine%20Drone/1255091/infra).

engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
```diff
@@ -57,7 +57,7 @@ class CopyArtifactsStep implements PipelineStep {
     };
     final Uri url = Uri.https(
       'storage.googleapis.com',
-      '${realmComponent}flutter_infra_release/flutter/$gitRevision/flutter-web-sdk.zip',
+      '${realmComponent}flutter_infra_release/flutter/${isPostsubmit ? contentHash : gitRevision}/flutter-web-sdk.zip',
```
eyebrowsoffire: You might be able to just do `realm == LuciRealm.Try ? gitRevision : contentHash` instead of reading the `LUCI_PR` environment variable. It's probably equivalent functionally, so it's up to you.

eyebrowsoffire (approved):
I have an alternative suggestion that might be simpler for detecting presubmit, but either way probably works.

engine/src/flutter/lib/web_ui/dev/steps/copy_artifacts_step.dart
matanlurey: Ah that's better.

I made it a negative so that, if you use this tool locally, it defaults to git revision instead of content hash.

mdebbar: The way I'm reading the code, only `LuciRealm.Try` will use `gitRevision`. Does local development use `LuciRealm.Try` or `LuciRealm.Unknown`?

Basically:
```
realm == LuciRealm.Try ? gitRevision : contentHash
```
and
```
realm != LuciRealm.Try ? contentHash : gitRevision
```
are the same thing.

mdebbar: What you probably want to do is `realm == LuciRealm.Prod ? contentHash : gitRevision`

eyebrowsoffire: This code path is never going to be exercised locally (and your actual local build will not ever be in GCS) so that point is moot.

I think the staging realm is postsubmit + `bringup: true` so I think that should be contentHash as well. Hence, only the try realm should actually use `gitRevision` right now, IMO

mdebbar: Ok that makes sense.

My comment was a response to:

> I made it a negative so that, if you use this tool locally, it defaults to git revision instead of content hash.

So if that point is moot, please carry on 🙂

matanlurey: Makes sense. Done as `${realm == LuciRealm.Try ? gitRevision : contentHash}`

matanlurey:
I am excited to land this, and find out if it works!11!

jtmcdole (approved):

matanlurey:
Skipping remaining presubmits, the one test that could have been impacted (in presubmit) has passed.

On postsubmit, this might fail. If it does, I'll revert it.

Issue #136769, opened 2023-10-17
Expose `OverlayPortal`'s declarative API


LongCatIsLooong:
Currently `OverlayPortal`'s declarative API is hidden behind the imperative `OverlayPortalController.open()` and `OverlayPortalController.close()` API. We couldn't find a declarative API that we could all agree on to describe the z-order of an `Overlay`'s render children. Under the hood, the `OverlayPortalController.open()` method uses the current time as the z-order value, such that the last controller to call `open` gets to show its overlay child at the top. 

One example where the declarative API becomes desirable is for implementing the menu system. All open menus should dismiss when the containing window changes size. Window resizing is propagated in the widget tree via `MediaQuery` changes, but `OverlayPortalController.open()` and `OverlayPortalController.close()` can't be called during widget tree rebuild because they both call `setState`.

LongCatIsLooong:
> We couldn't find a declarative API that we could all agree on to describe the z-order of an Overlay's render children. 

The `controller` API solved all known use cases at the time (see the review comments in https://github.com/flutter/flutter/pull/105335) by sorting the z-index (the paint order) of different active overlay children by the time `controller.open` was called. This works well because overlay portals usually "open" in response to user interactions that happen between frames.  

I also recall that in the discussion we agreed that directly taking an integer as the z-index is probably not the best approach, as it will be difficult for different widgets to find THE right index to use (I want my widget to be above widget a from package A and below widget b from package B, but a is given a larger z-index by the author of A, or the z indices of a and b happen to be consecutive integers).

LongCatIsLooong:
Workaround for the menu system case (or more generally when you want to remove the overlay child during build): return an empty `SizedBox` (or anything that cannot be perceived by the user).

LongCatIsLooong:
Semi related:

OverlayPortal internally has the ability to target any ancestor Overlays (not just the closest one or the root Overlay). Currently the API allows the user to target root overlay via a named constructor `targetRootOverlay` but if the developer wants to make that a run-time decision then the code gets a bit verbose:
```dart
final overlayPortal = targetRootOverlay
  ? OverlayPortal.targetRootOverlay(...)
  : OverlayPortal(...);
```

The original plan was to expose a `location` parameter that encodes the info about the target overlay and the z-index.

davidhicks980:
@LongCatIsLooong 

[...]

Issue #2901, opened 2016-03-25
Reactivate should guarantee rebuild


Hixie:
We've been assuming that if you're ever reactivated, you will be rebuilt, but actually if you're reprojected that might not be true.

This leads to things like Inherited not properly notifying descendants when they move around and get a new Inherited ancestor.

Hixie:
also means that if you disconnect from an ancestor when you deactivate (e.g. disconnecting from a Material that's doing your splashes), you won't reconnect.

PR #174239, opened 2025-08-21
Allow OverlayPortal.overlayChildLayoutBuilder to choose root Overlay

### LINKED ISSUE #168785: Allow OverlayPortal.overlayChildLayoutBuilder to choose which overlay to use.
### Use case

OverlayPortal.overlayChildLayoutBuilder was recently landed, and it has proven useful for implementing animated nested menus (this is a huge understatement. I had written a gigantic follower chain layer that worked but was  painfully complex. This will save me many hours). However, AFAIK, there is no option to target the root overlay. 

### Proposal

Allow developers to choose which overlay to target.

chunhtai:
fixes https://github.com/flutter/flutter/issues/168785

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1707,21 +1707,17 @@ class OverlayPortalController {
 /// entry to.
 ///
 /// This is typically used in [OverlayPortal].
-final class OverlayLocation {
-  const OverlayLocation._(this._isRoot);
-
-  final bool _isRoot;
-
+enum OverlayLocation {
   /// The [OverlayPortal] attaches to the immediate [Overlay] above the widget
   /// tree.
-  static const OverlayLocation nearest = OverlayLocation._(false);
+  nearest,
 
   /// The [OverlayPortal] attaches to the root [Overlay] above the widget
   /// tree.
   ///
   /// In case of multi-view apps, the root [Overlay] refers to the first Overlay
   /// below the View.
-  static const OverlayLocation root = OverlayLocation._(true);
+  root,
 }
 
 /// A widget that renders its overlay child on an [Overlay].
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1707,7 +1707,7 @@ class OverlayPortalController {
 /// child on.
 ///
 /// This is typically used in [OverlayPortal].
-enum OverlayLocation {
+enum OverlayChildLocation {
   /// The [OverlayPortal] renders its overlay child on the closest ancestor
   /// [Overlay] above the widget tree.
   nearestOverlay,
@@ -1804,12 +1804,12 @@ class OverlayPortal extends StatefulWidget {
   ///
   /// The [overlayLocation] sets which [Overlay] this widget attaches the widget
   /// returned by [overlayChildBuilder] to. Defaults to
-  /// [OverlayLocation.nearest].
+  /// [OverlayChildLocation.nearestOverlay].
   const OverlayPortal({
     super.key,
     required this.controller,
     required this.overlayChildBuilder,
-    this.overlayLocation = OverlayLocation.nearest,
+    this.overlayLocation = OverlayChildLocation.nearestOverlay,
     this.child,
   });
 
@@ -1825,7 +1825,7 @@ class OverlayPortal extends StatefulWidget {
     required this.controller,
     required this.overlayChildBuilder,
     this.child,
-  }) : overlayLocation = OverlayLocation.root;
+  }) : overlayLocation = OverlayChildLocation.rootOverlay;
 
   /// Creates an [OverlayPortal] that renders the widget `overlayChildBuilder`
   /// builds on the closest [Overlay] when [OverlayPortalController.show] is
@@ -1851,12 +1851,12 @@ class OverlayPortal extends StatefulWidget {
   ///
   /// The [overlayLocation] sets which [Overlay] this widget attaches the widget
   /// returned by `overlayChildBuilder` to. Defaults to
-  /// [OverlayLocation.nearest].
+  /// [OverlayChildLocation.nearestOverlay].
   OverlayPortal.overlayChildLayoutBuilder({
     Key? key,
     required OverlayPortalController controller,
     required OverlayChildLayoutBuilder overlayChildBuilder,
-    OverlayLocation overlayLocation = OverlayLocation.nearest,
+    OverlayChildLocation overlayLocation = OverlayChildLocation.nearestOverlay,
     required Widget? child,
   }) : this(
          key: key,
@@ -1892,7 +1892,7 @@ class OverlayPortal extends StatefulWidget {
 
   /// The [Overlay] that the widget returns from [overlayChildBuilder] is
   /// attached to.
-  final OverlayLocation overlayLocation;
+  final OverlayChildLocation overlayLocation;
 
   @override
   State<OverlayPortal> createState() => _OverlayPortalState();
@@ -1917,11 +1917,11 @@ class _OverlayPortalState extends State<OverlayPortal> {
         locationCache._theater == marker.theater;
   }
 
-  _OverlayEntryLocation _getLocation(int zOrderIndex, OverlayLocation overlayLocation) {
+  _OverlayEntryLocation _getLocation(int zOrderIndex, OverlayChildLocation overlayLocation) {
     final _OverlayEntryLocation? cachedLocation = _locationCache;
     late final _RenderTheaterMarker marker = _RenderTheaterMarker.of(
       context,
-      targetRootOverlay: overlayLocation == OverlayLocation.root,
+      targetRootOverlay: overlayLocation == OverlayChildLocation.rootOverlay,
     );
     final bool isCacheValid =
         cachedLocation != null &&
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1703,21 +1703,21 @@ class OverlayPortalController {
   }
 }
 
-/// The location of the [Overlay] that a [OverlayPortal] attaches its overlay
-/// entry to.
+/// The location of the [Overlay] that an [OverlayPortal] renders its overlay
+/// child on.
 ///
 /// This is typically used in [OverlayPortal].
 enum OverlayLocation {
-  /// The [OverlayPortal] attaches to the immediate [Overlay] above the widget
-  /// tree.
-  nearest,
+  /// The [OverlayPortal] renders its overlay child on the closest ancestor
+  /// [Overlay] above the widget tree.
+  nearestOverlay,
 
-  /// The [OverlayPortal] attaches to the root [Overlay] above the widget
-  /// tree.
+  /// The [OverlayPortal] renders its overlay child on the root [Overlay] above
+  /// the widget tree.
   ///
   /// In case of multi-view apps, the root [Overlay] refers to the first Overlay
   /// below the View.
-  root,
+  rootOverlay,
 }
 
 /// A widget that renders its overlay child on an [Overlay].
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1703,6 +1703,27 @@ class OverlayPortalController {
   }
 }
 
+/// The location of the [Overlay] that a [OverlayPortal] attaches its overlay
+/// entry to.
+///
+/// This is typically used in [OverlayPortal].
+final class OverlayLocation {
+  const OverlayLocation._(this._isRoot);
+
+  final bool _isRoot;
+
+  /// The [OverlayPortal] attaches to the immediate [Overlay] above the widget
+  /// tree.
+  static const OverlayLocation nearest = OverlayLocation._(false);
+
+  /// The [OverlayPortal] attaches to the root [Overlay] above the widget
+  /// tree.
+  ///
+  /// In case of multi-view apps, the root [Overlay] refers to the first Overlay
+  /// below the View.
+  static const OverlayLocation root = OverlayLocation._(true);
+}
+
 /// A widget that renders its overlay child on an [Overlay].
 ///
 /// The overlay child is initially hidden until [OverlayPortalController.show]
@@ -1784,22 +1805,31 @@ class OverlayPortal extends StatefulWidget {
   /// Creates an [OverlayPortal] that renders the widget [overlayChildBuilder]
   /// builds on the closest [Overlay] when [OverlayPortalController.show] is
   /// called.
+  ///
+  /// The [overlayLocation] sets which [Overlay] this widget attaches the widget
+  /// returned by [overlayChildBuilder] to. Defaults to
+  /// [OverlayLocation.nearest].
   const OverlayPortal({
     super.key,
     required this.controller,
     required this.overlayChildBuilder,
+    this.overlayLocation = OverlayLocation.nearest,
     this.child,
-  }) : _targetRootOverlay = false;
+  });
 
   /// Creates an [OverlayPortal] that renders the widget [overlayChildBuilder]
   /// builds on the root [Overlay] when [OverlayPortalController.show] is
   /// called.
+  @Deprecated(
+    'Use OverlayPortal with root overlay instead. '
+    'This feature was deprecated after v3.33.0-0.0.pre.',
+  )
   const OverlayPortal.targetsRootOverlay({
     super.key,
     required this.controller,
     required this.overlayChildBuilder,
     this.child,
-  }) : _targetRootOverlay = true;
+  }) : overlayLocation = OverlayLocation.root;
 
   /// Creates an [OverlayPortal] that renders the widget `overlayChildBuilder`
   /// builds on the closest [Overlay] when [OverlayPortalController.show] is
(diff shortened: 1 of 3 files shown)


packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1789,18 +1793,24 @@ class OverlayPortal extends StatefulWidget {
...
     required this.controller,
     required this.overlayChildBuilder,
+    this.targetsOverlay,
     this.child,
   }) : _targetRootOverlay = false;
 
   /// Creates an [OverlayPortal] that renders the widget [overlayChildBuilder]
   /// builds on the root [Overlay] when [OverlayPortalController.show] is
   /// called.
+  @Deprecated(
```
chunhtai, line 1819: I am neutral on whether to remove this. I did it now to remove redundancy and also can help clean up the code `_targetRootOverlay` if this eventually gets removed

LongCatIsLooong: I'm leaning towards deprecating this (and re-implement the underlying logic using the new API).

chunhtai: not possible since we need context to do Overlay.of(rootOverlay: true)

There are other alternatives, but that will make constructor non null (by using a OverlayState getter)

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1785,22 +1797,30 @@ class OverlayPortal extends StatefulWidget {
   /// Creates an [OverlayPortal] that renders the widget [overlayChildBuilder]
   /// builds on the closest [Overlay] when [OverlayPortalController.show] is
   /// called.
+  ///
+  /// {@macro flutter.widgets.overlayPortal.targetsOverlay}
   const OverlayPortal({
     super.key,
     required this.controller,
     required this.overlayChildBuilder,
+    this.targetsOverlay,
```
LongCatIsLooong: `targetOverlay` as a noun instead of a phrase?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1723,6 +1727,14 @@ class OverlayPortalController {
 /// widget rebuilds. Stateful descendants in the overlay child subtree may lose
 /// states as a result.
 ///
+/// {@template flutter.widgets.overlayPortal.targetsOverlay}
+/// Use [targetsOverlay] to choose which [Overlay] this overlay portal
+/// attaches to. Consider using [GlobalKey] or [Overlay.of] with appropriate
+/// [BuildContext] to get the [OverlayState]. One can also use
+/// `Overlay.of(context, rootOverlay: true)` to pick the root [Overlay]. If
```
LongCatIsLooong: Hmm calling `Overlay.of` doesn't establish a dependency (it's not using an inheritedWidget). So if an OverlayPortal gets reparented to a different subtree where it can't see the old target Overlay, it's not guaranteed that the OverlayPortal's parent will rebuild to update the `targetsOverlay` parameter (e.g., if you cache the widget).

LongCatIsLooong: See: https://main-api.flutter.dev/flutter/widgets/Overlay/maybeOf.html

LongCatIsLooong: (I thought I had a test for this but apparently all tests are passing)

chunhtai: In that case, it is the consumer of this widget to ensure it grabs the correct overlay state when it change? (in State.activate)

not entirely sure what else we can do. using globalkey is not feasible unless we expose root overlay's globalkey all the way to WidgetsApp

LongCatIsLooong: >  it is the consumer of this widget to ensure it grabs the correct overlay state when it change

Yeah if the creator of the OverlayPortal widget rebuilds. But the problem is what if that widget doesn't rebuild because the user cached the widget, or the widget is a `const` instance? I think we need an alternative version of `Overlay.of` that establishes a dependency (i.e. an InheritedWidget for `Overlay`). To prevent these kind of bugs maybe the new `of` method should use a different type from `OverlayState` so people can't use `Overlay.of` for this API.

packages/flutter/lib/src/widgets/overlay.dart
LongCatIsLooong: Ah I see yeah you don't have access to the root overlay until mounted.

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: Looking at the code, I think the statefulElement will be rebuild when globalkey reparent. but not StatelessWidget. not sure why this is the case.

https://github.com/flutter/flutter/blob/7130981fe1ca12481f0bd5c4bdf6dbcee84afe43/packages/flutter/lib/src/widgets/framework.dart#L6019

Can the overlay.of create dependency be another pr?

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: yeah, has been reading too much objective c code

packages/flutter/lib/src/widgets/overlay.dart
LongCatIsLooong: If you had dependencies before getting reparented, you'll rebuilt on activation (via `didChangeDependencies`):

https://github.com/flutter/flutter/blob/f0f208deae4f50973fd332afcbd55113da46ef3d/packages/flutter/lib/src/widgets/framework.dart#L4758-L4772

> I think the statefulElement will be rebuild when globalkey reparent.

Looks like the issue was https://github.com/flutter/flutter/issues/2901, but I'm not sure as to why that was needed (given that elements with dependencies already rebuild on activation, and that both the old parent Element and new parent Element rebuild)?

LongCatIsLooong: > Can the overlay.of create dependency be another pr?

Yeah but I think the `Overlay.of` that establishes dependencies a prerequisite for this PR?

chunhtai: > that both the old parent Element and new parent Element rebuild)?

If you don't have dependecy, but have some specialize logic to update state in activate, then this makes sense. My guess is that since State has access to to activate callback, this is necessary to make it not a footgun.

> Yeah but I think the Overlay.of that establishes dependencies a prerequisite for this PR?

sgtm

chunhtai:
a reminder to myself on how to write test that swap overlay
https://gist.github.com/chunhtai/6107ecfd30c95f00a6e4da00e6cbe86e

davidhicks980:
Before this lands, how are you thinking this could be integrated into a widget like `RawMenuAnchor`? 

For RawMenuAnchor, there currently is only one constructor:

```dart
const RawMenuAnchor({
    super.key,
    this.childFocusNode,
    this.consumeOutsideTaps = false,
    this.onOpen,
    this.onClose,
    this.onOpenRequested = _defaultOnOpenRequested,
    this.onCloseRequested = _defaultOnCloseRequested,
    this.useRootOverlay = false,
    this.builder,
    required this.controller,
    Widget Function(BuildContext context, RawMenuOverlayInfo info)  required this.overlayBuilder,
    this.child,
  });
  ```  

Some users of `RawMenuAnchor` support a follower layer (e.g. `MenuAnchor`), so using `OverlayPortal.overlayChildLayoutBuilder` within `RawMenuAnchor` would require an additional constructor on `RawMenuAnchor` -- one using overlayChildLayoutBuilder and one using the default constructor.

The reason why I mentioned making `_OverlayChildLayoutBuilder` public ([see here](https://github.com/flutter/flutter/issues/168785#issuecomment-2943705989)) is because only a single OverlayPortal AND a single `RawMenuAnchor` constructor would be required -- there would no longer be the need for `OverlayPortal.overlayChildLayoutBuilder` since users of `RawMenuAnchor` could just insert a `OverlayChildLayoutBuilder` in their overlay. Choosing whether to use the `OverlayChildLayoutBuilder` would be the responsibility of downstream widgets (e.g. `MenuAnchor`) rather than upstream widgets (e.g. `RawMenuAnchor`). 

Either way works for me, but I figure it's better to get your opinions on this now rather than after this lands.

**Comment by LongCatIsLooong** [comment]
> Before this lands, how are you thinking this could be integrated into a widget like `RawMenuAnchor`?
> 
> For RawMenuAnchor, there currently is only one constructor:
> 
> ```dart
> const RawMenuAnchor({
>     super.key,
>     this.childFocusNode,
>     this.consumeOutsideTaps = false,
>     this.onOpen,
>     this.onClose,
>     this.onOpenRequested = _defaultOnOpenRequested,
>     this.onCloseRequested = _defaultOnCloseRequested,
>     this.useRootOverlay = false,
>     this.builder,
>     required this.controller,
>     Widget Function(BuildContext context, RawMenuOverlayInfo info)  required this.overlayBuilder,
>     this.child,
>   });
> ```
> 
> Some users of `RawMenuAnchor` support a follower layer (e.g. `MenuAnchor`), so using `OverlayPortal.overlayChildLayoutBuilder` within `RawMenuAnchor` would require an additional constructor on `RawMenuAnchor` -- one using overlayChildLayoutBuilder and one using the default constructor.
> 
> The reason why I mentioned making `_OverlayChildLayoutBuilder` public ([see here](https://github.com/flutter/flutter/issues/168785#issuecomment-2943705989)) is because only a single OverlayPortal AND a single `RawMenuAnchor` constructor would be required -- there would no longer be the need for `OverlayPortal.overlayChildLayoutBuilder` since users of `RawMenuAnchor` could just insert a `OverlayChildLayoutBuilder` in their overlay. Choosing whether to use the `OverlayChildLayoutBuilder` would be the responsibility of downstream widgets (e.g. `MenuAnchor`) rather than upstream widgets (e.g. `RawMenuAnchor`).
> 
> Either way works for me, but I figure it's better to get your opinions on this now rather than after this lands.

`OverlayChildLayoutBuilder` doesn't work outside of OverlayPortal. I think exposing that as a generic widget makes it harder to discover and much easier to misuse (like the way parentData can be misused in which case you'll get a runtime error).

**Comment by davidhicks980** [comment]
@LongCatIsLooong Yeah, that makes sense. IIRC only calculating/passing layout info if dependence is established (e.g. `OverlayPortal.childLayoutOf(context)`) also wouldn't work since `_computeNewLayoutInfo()` would need to have been called prior to dependence. So, the current approach sounds good.

**Comment by chunhtai** [comment]
Turns out, the _RenderTheaderMarker.of already register dependencies, so swapping out overlay won't be a problem. I added more test regardless

File: packages/flutter/lib/src/widgets/overlay.dart [hunk 4]
```diff
@@ -1703,6 +1703,23 @@ class OverlayPortalController {
   }
 }
 
+/// The location of the [Overlay] that a [OverlayPortal] attaches its overlay
```
LongCatIsLooong: nit: an [OverlayPortal].

LongCatIsLooong: Maybe something like "A type that represents the target [Overlay], which an [OverlayPortal] should render its overlay child on"?

I'm a bit torn on wether we should mention OverlayEntry here. Currently the overlay child is paint above the OverlayPortal's OverlayEntry and beneath the next OverlayEntry in the paint order, but it's already documented in the OverlayPortal class documentation, so I guess we can keep it simple for now (since the API that allows you to target a different OverlayEntry doesn't exist yet)?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1703,6 +1703,23 @@ class OverlayPortalController {
...
+/// entry to.
+///
+/// This is typically used in [OverlayPortal].
+enum OverlayLocation {
+  /// The [OverlayPortal] attaches to the immediate [Overlay] above the widget
```
LongCatIsLooong: The [OverlayPortal] renders its overlay child on the closest ancestor [Overlay], or some such?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1703,6 +1703,23 @@ class OverlayPortalController {
...
+  /// tree.
+  nearest,
```
LongCatIsLooong: nearestOverlay in case we expose the ability to change the z-offset within an Overlay?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1703,6 +1703,23 @@ class OverlayPortalController {
   }
 }
 
+/// The location of the [Overlay] that a [OverlayPortal] attaches its overlay
+/// entry to.
+///
+/// This is typically used in [OverlayPortal].
+enum OverlayLocation {
```
LongCatIsLooong: maybe: OverlayChildLocation?

LongCatIsLooong (commented):
The implementation lgtm.

I'm not very sure how best to evolve the API. I'm hoping we can add new `location`s to the `OverlayLocation` and eventually enable https://github.com/flutter/flutter/issues/136769, but now the controller argument gets in the way since it was introduced to track the z-index.

packages/flutter/lib/src/widgets/overlay.dart
LongCatIsLooong: (also `rootOverlay`)

deminearchiver:
Hello. I am not a contributor, but a developer using Flutter. When I first spotted this PR, the changes looked like they would allow passing an arbitrary `OverlayState` value to `OverlayPortal.childLayoutBuilder`. There even were tests for that, if I recall correctly. But now only the nearest or the root ancestor overlay locations are supported, which makes this PR less useful, probably. I was thought that it would be possible to pass a custom overlay that resides, for example, between the root overlay and the nearest ancestor, via a `GlobalKey`, or even in an arbitrary location in the tree since `_OverlayChildLayoutBuilder` establishes paint transform to the target overlay. I suppose there were some concerns about this implementation? I would be thankful if somebody explained why the API was nerfed so drastically from the initial changes.

chunhtai:
> the controller argument gets in the way since it was introduced to track the z-index.

If we want to be declarative, do we still need a controller?

chunhtai:
@deminearchiver 

The problem is that we still not sure how to implement it correctly. The previous implementation uses OverlayState as input, but the Overlay.of won't listen to overlayState changes in case of global key reparenting. We can make Overlay.of() to use dependsOnInheritedWidgetOfExactType to listen for OverlayState change, but it would be a breaking change if someone call Overlay.of in initState. This also brought the question whether takes an OverlayState as input is the best way to solve this problem.

Also the https://github.com/flutter/flutter/issues/136769, if we want to be fulling declarative, we not only need the OverlayState, we also need a way to specify z-index. That is why we decided to hold it back until we figure that out.

For now, we decided to solve the immediate issue first that the `overlayChildLayoutBuilder` to be able to choose the root overlay and keep the option open (using an enum class) for future improvement.

packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -1344,6 +1340,198 @@ void main() {
...
+    ).single;
+
+    final OverlaySwapsState swaps = tester.state<OverlaySwapsState>(find.byType(OverlaySwaps));
+    swaps.swaps();
+    await tester.pumpAndSettle();
+
+    final RenderObject newParentTheater = _ancestorRenderTheaters(
+      tester.renderObject(find.byKey(overlayPortal)),
+    ).single;
+    expect(parentTheater != newParentTheater, isTrue);
```
LongCatIsLooong: nit: `expect(parentTheater, isNot(newParentTheater))` gives a better error message.

packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -1344,6 +1340,198 @@ void main() {
...
+
+    await tester.pumpWidget(
+      Directionality(
+        textDirection: TextDirection.ltr,
+        child: OverlaySwaps(overlayEntry1, overlayEntry2),
+      ),
+    );
+
+    controller1.show();
+    await tester.pumpAndSettle();
```
LongCatIsLooong: why `pumpAndSettle`? Shouldn't `pump` suffice here?

LongCatIsLooong (approved):
LGTM with nits

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -823,19 +823,14 @@ class _RawMenuAnchorState extends State<RawMenuAnchor> with _RawMenuAnchorBaseMi
...
-      return OverlayPortal(
-        controller: _overlayController,
-        overlayChildBuilder: _buildOverlay,
-        child: child,
-      );
-    }
+    return OverlayPortal(
+      controller: _overlayController,
+      overlayChildBuilder: _buildOverlay,
+      overlayLocation: useRootOverlay
```
davidhicks980, line 829: At some point we should replace the useRootOverlay boolean with an overlayChildLocation param on RawMenuAnchor. I can also do this on a future PR since it's a breaking change.

chunhtai:
migration guide https://github.com/flutter/website/pull/12397

LongCatIsLooong:
Oh does this change break any registered test?

chunhtai:
no, but deprecating the constructor must have broken someone's code.

LongCatIsLooong:
> no, but deprecating the constructor must have broken someone's code.

Would it be possible to apply a dart fix?

chunhtai:
I don't think the dart fix can change name **and** add parameter. but I am not too familiar with the syntax

PR #174253, opened 2025-08-22
Load fonts in the order addFont is called

jiahaog:
Before this change, `load` loads fonts in a non-deterministic order. After this change, it loads the fonts sequentially.

See cl/797583429 for the full context. We found that the order of how fonts load for a specific font family actually affects runtime behavior, and the current non-deterministic order of loading fonts leads to flakiness.

Consider the following:

- foo.ttf has glyphs 'A' and 'B'
- foo-italic.ttf only has glyph 'B'

If `addFont(foo-italic.ttf)` is called before `addFont(foo.ttf)`, the font loader will load `foo-italic.ttf` first and use that as the base font. When looking for the 'A' glyph, it may will look for the glyph only in `foo-italic.ttf`, which doesn't have it, and will show an incorrect empty placeholder glyph.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/services/font_loader.dart
+++ b/packages/flutter/lib/src/services/font_loader.dart
@@ -31,6 +31,8 @@ class FontLoader {
   ///
   /// The [bytes] argument specifies the actual font asset bytes. Currently,
   /// only OpenType (OTF) and TrueType (TTF) fonts are supported.
+  ///
+  /// [load] will load fonts in the order this is called.
   void addFont(Future<ByteData> bytes) {
     if (_loaded) {
       throw StateError('FontLoader is already loaded');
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/services/font_loader.dart
```diff
@@ -31,6 +31,8 @@ class FontLoader {
   ///
   /// The [bytes] argument specifies the actual font asset bytes. Currently,
   /// only OpenType (OTF) and TrueType (TTF) fonts are supported.
+  ///
+  /// [load] will load fonts in the order this is called.
```
Piinks:
```suggestion
  /// The [load] method will load fonts in the order this is called.
```

Piinks (approved):
LGTM, thanks for this. Sorry it appeared to have fallen through the review queue. 🙏

jiahaog:
Thanks!

jason-simmons:
Note that this was necessary in the cl/797583429 case because the italic TTF file used there did not contain the correct data indicating that it is an italic font.

The scoring algorithm in `SkFontStyleSet::matchStyleCSS3` will search for the typeface whose attributes most closely match the specified text style.  Ordinarily, if the text style does not request italics, then the algorithm will prefer the regular typeface over the italic typeface.

However, in this situation the italic font asset was missing data like the `OS/2.fsSelection` italic bit and the `post.italicAngle` value.  So the scoring algorithm did not know it was an italic font and assigned it the same score as the regular font.  If both fonts have the same score, then `matchStyleCSS3` will default to the font that comes first in the iteration order.

Ideally the tables within that font should be fixed.  But this PR will ensure that the results are consistent if there is an error in the font that produces the wrong scores.

jiahaog:
Thanks for looking it and sharing the root cause!

Issue #27587, opened 2019-02-06
GestureDetector doesn't respond correctly after transition with Transform widget


MariaMelnik:
Transformed widget (GestureDetector with inner child) doesn't respond correctly.
Part of widget what stays in original boundaries is good, but outside original boundaries - doesn't respond.
Flutter inspector shows correct boundaries of transformed widget.

main.dart with this issue: [https://gist.github.com/MariaMelnik/a8a3cfaeaf7dbc11ead0dd472790040d](url)

flutter inspector:
![image](https://user-images.githubusercontent.com/20188655/52337667-52032e80-2a19-11e9-88ed-5f6e472e696f.png)

flutter doctor:
```
[✓] Flutter (Channel beta, v1.0.0, on Mac OS X 10.13.6 17G65, locale en-RU)
    • Flutter version 1.0.0 at /Users/mariamelnik/work/flutter
    • Framework revision 5391447fae (10 weeks ago), 2018-11-29 19:41:26 -0800
    • Engine revision 7375a0f414
    • Dart version 2.1.0 (build 2.1.0-dev.9.4 f9ebf21297)

[!] Android toolchain - develop for Android devices (Android SDK 28.0.3)
    • Android SDK at /Users/mariamelnik/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)
    ! Some Android licenses not accepted.  To resolve this, run: flutter doctor --android-licenses

[!] iOS toolchain - develop for iOS devices (Xcode 9.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 9.4.1, Build version 9F2000
    ✗ Verify that all connected devices have been paired with this computer in Xcode.
      If all devices have been paired, libimobiledevice and ideviceinstaller may require updating.
      To update with Brew, run:
        brew update
        brew uninstall --ignore-dependencies libimobiledevice
        brew uninstall --ignore-dependencies usbmuxd
        brew install --HEAD usbmuxd
        brew unlink usbmuxd
        brew link usbmuxd
        brew install --HEAD libimobiledevice
        brew install ideviceinstaller
    • ios-deploy 1.9.2
[...]

Issue #75747, opened 2021-02-10
Opt a subtree out of hitTestSelf culling


Hixie:
We often get questions about why hit testing "doesn't work" around transforms. The short answer is that ancestors don't ask their children to hit test if their own hitTestSelf fails (no point asking your children if they were hit if the hit wasn't inside your bounds).

@csells suggested that it would be nice if a subtree could opt-in to not doing that culling. Obviously that would in general be expensive but it would be less expensive than doing a full tree walk (unless someone opted out the entire tree, obviously). I think we would still avoid doing this ever in the Flutter framework since it's better to adjust the widget hierarchy to not have ancestors of the transformed widget be smaller than that widget, but it might be useful in especially elaborate scenes where that is impractical.

One way to do that would be to have HitTestResult have a flag that says whether to apply the hitTestSelf check. Having to check that flag every step of the walk would itself be not cheap, but the cost (which affects every app) might still be worth it?

Another option might be to have a way for a widget to hook into the hit testing of a far ancestor (e.g. via an inherited widget). That would be better in that it would allow the whole thing to be done as a package. Maintaining paint order correctness would be distinctly non-trivial though (maybe even intractable).

Any other ideas?

cc @goderbauer

goderbauer:
A solution for this would also help with the often raised complained that you can't hit-test children drawn in the visual overflow of a stack (see discussion at https://github.com/flutter/flutter/issues/19445).

goderbauer:
> it would be nice if a subtree could opt-in to not doing that culling. Obviously that would in general be expensive but it would be less expensive than doing a full tree walk (unless someone opted out the entire tree, obviously).

I believe choosing the root of the subtree that you want to opt out will not not be easy. The root will not be the transform widget that moves a child out of its bounds. It has to be an ancestor, who has the transformed descendant still in its bounds. Otherwise, the hit test would never reach the transform widget or the transformed child. I am wondering, if in practice that would encourage people to basically opt-out the entire tree...

goderbauer:
Would be interesting to study a use case, that we would enable (or make significantly simpler over e.g. re-organizing the widget tree) by having this mechanism. Do we have one?

esDotDev:
I can certainly provide a number of use cases, just slammed right now. I'll circle back in a cpl weeks when things calm down a bit. 

[...]

PR #174286, opened 2025-08-22
Update transformHitTests documentation for clarity

### LINKED ISSUE #78120: Transforms transformHitTests description is not clear enough
## Use case

As seen in e.g. issue [#27587](https://github.com/flutter/flutter/issues/27587), many developers are confused about hit testing not working as expected when using `Transform`.

Take for example the following:
```dart

void main() {
  runApp(MyApp());
}

  @override
    return MaterialApp(
      theme: ThemeData.dark(),
      home: Scaffold(
        body: Center(
          child: Container(
            width: 100,
            height: 100,
            color: Colors.yellow,
            child: Transform.translate(
              offset: Offset(50, 50),
              transformHitTests: true,
              child: GestureDetector(
                onTap: () => print("hit"),
                child: Container(
                  height: 100,
                  width: 100,
                  color: Colors.blue.withAlpha(125),
                ),
              ),
            ),
          ),
        ),
      ),
    );
  }
}
```
2021031319:10:21_screenshot_sel

Hitting the overlapping area will work as expected, but hitting the overhanging area of the blue square will not result in a hit.
This happens since the parent can not register a hit outside its bounds and therefore does not inform its children.

`transformHitTests` description reads a little confusing in that regard:
```
/// Whether to apply the transformation when performing hit tests.
```
Which seems to suggest that the hit area will be transformed along with the widget.

The actual role of `transformHitTests` is to transform the coordinates of hits that the parent does register to the coordinate system of the child.
(See https://github.com/flutter/flutter/pull/32192)

## Proposal

Changing the description might clear up some confusion and hint to the reason why hits outside the parent can not be registered.
It might also be a good place to hint to the prefered usage of Overlay as mentioned [here](https://github.com/flutter/flutter/issues/27587#issuecomment-618071879)

Possible alternative description:
```
/// Whether to transformation registered hits into the childs resulting coordinate system.
/// Hits can only be registered if they are within the bounds of the parent.
///
/// Consider increasing the bounds of the parent to fit the childs new position.
/// Alternatively consider placing the child using `Overlay` instead.
```

Or something along those lines.
Potentially reference `RenderBox` `hitTestChildren` as well?

Rushikeshbhavsar20:
## Description 📝

This PR updates the documentation for the transformHitTests parameter in the Transform widget to resolve ambiguity in the API documentation that has caused hit testing issues for developers.

The current documentation simply states "Whether to apply the transformation when performing hit tests" which is ambiguous. As seen in issues like #27587, developers expect transformed widgets to receive touch events across their entire visual area, but hits only register within the parent's original bounds. 

**The core issue**:  The documentation lacks clarity on:
❌ Hit test coordinate system transformation vs. hit test bounds expansion
🔄 Parent widget bounds as the constraint for hit test registration
⚠️ The distinction between visual rendering bounds and interactive bounds

### Before (Current Documentation) 

*Current documentation is vague: "Whether to apply the transformation when performing hit tests"*

Screenshot 2025-08-22 214745

### After (Improved Documentation) 
Screenshot 2025-08-22 214808

*New documentation clearly explains behavior, limitations, and solutions*

**Changes made:** ✅
- 🔧 Clarified that `transformHitTests` transforms registered hits into the child's resulting coordinate system
- 🚨 Explicitly documented the critical limitation: only hits within parent bounds can be registered
- 📚 Explained the behavior difference between `true` and `false` settings with concrete examples of transformation types
- 💡 Added guidance for common workarounds: expanding parent bounds or using Overlay for out-of-bounds elements

## Issues Fixed 🐛

Fixes #78120

**Related issues that motivated this change:** 🔗
- #27587 - Developers confused why GestureDetector doesn't respond after Transform
-  Multiple developers reported spending significant time debugging this behavior
-  Common misconception that visual transformation expands the hit test area

**Note**: This PR is test-exempt as it only updates documentation comments without changing any functionality.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -1778,7 +1778,18 @@ class Transform extends SingleChildRenderObjectWidget {
...
+  /// When `true`: Hit coordinates within the parent's bounds are transformed to match
+  /// where the child appears visually after any transformation such as (translation, rotation, scaling, or skewing etc).
+  ///
+  /// When `false`: Hit coordinates are not transformed, potentially causing taps to
+  /// register in the wrong location relative to the child's visual position.
+  ///
+  /// **Important:** Only hits within the parent's bounds can be registered and transformed,
+  /// part of the child extending outside the parent will not receive touch events.
+  ///
+  /// For interactive elements outside parent bounds, consider using an [Overlay].
```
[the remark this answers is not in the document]
Rushikeshbhavsar20: Thanks for the feedback! I've updated the documentation to address your suggestions:

✅ Removed informal language ("etc.")
✅ Added `{@tool snippet}` with practical scaling example
✅ Enhanced clarity about coordinate transformation and parent bounds limitation

The code example now clearly demonstrates the hit testing behavior when a transformed child extends beyond its parent's bounds.

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -1778,7 +1778,45 @@ class Transform extends SingleChildRenderObjectWidget {
...
+  ///         onTap: () => debugPrint('Tapped!'),
+  ///         child: const ColoredBox(
+  ///           color: Colors.purple,
+  ///         ),
+  ///       ),
+  ///     ),
+  ///   ),
+  /// )
+  /// ```
+  /// {@end-tool}
```
justinmc, line 1825: I would also add: "Defaults to true."

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -1778,7 +1778,45 @@ class Transform extends SingleChildRenderObjectWidget {
...
 
-  /// Whether to apply the transformation when performing hit tests.
+  /// Whether to transform registered hits into the child's resulting coordinate system.
+  ///
+  /// When `true`, hit coordinates within the parent's bounds are transformed to match
+  /// where the child appears visually after any transformation such as translation,
+  /// rotation, scaling, or skewing.
+  ///
+  /// When `false`, hit coordinates are not transformed, potentially causing taps to
+  /// register in the wrong location relative to the child's visual position.
```
justinmc: Instead of "wrong", I would say "...in a different location relative...".

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -1778,7 +1778,45 @@ class Transform extends SingleChildRenderObjectWidget {
...
+  ///
+  /// **Important:** Only hits within the parent's original bounds can be registered
+  /// and transformed. Any part of the child extending outside the parent's bounds
+  /// will not receive touch events.
```
justinmc: Thanks for including this, I think it's tricky. See https://github.com/flutter/flutter/issues/75747.

justinmc (commented):
Thanks for improving Flutter's docs! This is a tricky topic.

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -1778,7 +1778,45 @@ class Transform extends SingleChildRenderObjectWidget {
...
+  ///
+  /// For interactive elements that need to be tappable outside their parent's
+  /// original bounds, consider using an [Overlay].
```
victorsanni: Is `Overlay` the right widget to use here? My understanding is that `Overlay` is more static, and `OverlayEntry`/`OverlayPortal` are used to insert widgets into the `Overlay`.

packages/flutter/lib/src/widgets/basic.dart
Rushikeshbhavsar20: Thanks for catching that! I’ve added “Defaults to true.” at the end of the property docs.

packages/flutter/lib/src/widgets/basic.dart
Rushikeshbhavsar20: Good point, thanks , I updated the wording .

packages/flutter/lib/src/widgets/basic.dart
Rushikeshbhavsar20: Thanks for the reference! I went through it and made some changes to the Important note. Please review when you get a chance.

packages/flutter/lib/src/widgets/basic.dart
Rushikeshbhavsar20: Thanks for pointing that out . I’ve revised the docs to suggest [OverlayEntry] or [OverlayPortal] instead of referring to Overlay directly. Please review and let me know if any changes are needed.

victorsanni (approved):
LGTM. Thanks for the PR.

loic-sharma (approved):
Thanks for the wonderful doc!

PR #174291, opened 2025-08-22
NavigatorPopScope examples no longer use deprecated onPop.

justinmc:
I just noticed that these example are using a deprecated member.


justinmc:
Gemini's style recommendations go against the Flutter styleguide.

justinmc:
I don't think this change is testable.

stuartmorgan-g:

chunhtai (approved):
LGTM

PR #174300, opened 2025-08-22
Correctly implement PlatformViews' cursors on Web

dkwingsmt:
Fix https://github.com/flutter/flutter/issues/174246

Before: 

https://github.com/user-attachments/assets/33d2a476-d9c4-46e1-a5c5-ea8e602060d6

After: 

https://github.com/user-attachments/assets/383e4a01-4b8e-4c0e-a5a4-fa0d1fc28b67

This PR is not a result of a thorough fix for mouse cursors on all platforms, but simply matching the behavior of `SelectionArea` on Web to that on non-Web.

A central question lies how a platform view widget should control the mouse cursor. On the surface, the answer is simple: it doesn't, and let the view content to control (hence the `MouseCursor.uncontrolled` constant). But what if the empty region of the view content doesn't control the cursor at all? This would be a problem even on non-Web, because if neither the view content nor the view controls the cursor on non-Web, then the user will find the cursor remaining the one used by the last region, which would be different when entering and when leaving, the same bug as  https://github.com/flutter/flutter/issues/174246 . On the other hand, what is it supposed to fall back to if the view content doesn't define a cursor? Is the default cursor the logically correct option? ...

I decided that I did not want to dig too much into the details, and instead focus on this specific issue. I tried the repro app on macOS and found that the empty region of `SelectionArea` falls back to the widget behind it, the same behavior as `MouseCursor.defer`. Therefore in this PR I made `MouseCursor.uncontrolled` essentially the same as `.defer` on Web, matching the behavior across platforms. 

## Pre-launch Checklist

- [ ] I signed the [CLA].


mdebbar:
This new behavior makes sense for the specific platform view used by `SelectionArea`, but I worry that it may not be ideal to generalize it for all platform views. On the web, it's better to let the contents of platform views decide their cursor (at least by default).

dkwingsmt:
I've tested the cursor behavior. The following app is a `MouseRegion(cursor: SystemMouseCursors.grab)` wrapping around a `WebViewWidget`. As you can see, as soon as the cursor moves within the webview it turns into the default cursor. Does this resolve your concern?

https://github.com/user-attachments/assets/1c40f15a-19ce-4411-9c19-44e56edd1d4b
<details>
<summary>
Code
</summary>
```dart
import 'package:webview_flutter/webview_flutter.dart';

const kStartUrl = 'https://example.com';

void main() {
  runApp(const CenteredWebViewApp());
}

class CenteredWebViewApp extends StatelessWidget {
  const CenteredWebViewApp({super.key});

  @override
    return const MaterialApp(
      home: CenteredWebViewPage(),
    );
  }
}

class CenteredWebViewPage extends StatefulWidget {
  const CenteredWebViewPage({super.key});

  @override
  State<CenteredWebViewPage> createState() => _CenteredWebViewPageState();
}

class _CenteredWebViewPageState extends State<CenteredWebViewPage> {
  late final WebViewController _controller;

  @override
  void initState() {
    super.initState();
    _controller = WebViewController()
      ..loadRequest(Uri.parse(kStartUrl));
  }

  @override
    return Scaffold(
      body: MouseRegion(
        cursor: SystemMouseCursors.grab,
        child: Center(
          child: LayoutBuilder(
            builder: (context, constraints) {
              final double width =
                  (constraints.maxWidth * 0.8).clamp(480, 1100);
              final double height =
                  (constraints.maxHeight * 0.8).clamp(360, 800);

              return Container(
                width: width,
                height: height,
                decoration: BoxDecoration(
                  color: Theme.of(context).colorScheme.surface,
                  borderRadius: BorderRadius.circular(14),
                  boxShadow: const [
                    BoxShadow(
                        blurRadius: 20, spreadRadius: 2, offset: Offset(0, 6))
                  ],
                  border: Border.all(color: Colors.black12),
                ),
                clipBehavior: Clip.antiAlias,
                child: WebViewWidget(controller: _controller),
              );
            },
          ),
        ),
      ),
    );
  }
}

```

</details>

dkwingsmt:
I talked with @mdebbar offline. For the record, Mouad was curious it would work even though the webview region of the Flutter view was assigned with a different cursor (such as the `.grab`), and my theory was that the DOM of the webview obstructed hit testing at an OS level and therefore Flutter view's cursor is not even used.

Mouad then pointed out that changing`.uncontrolled`'s behavior so that it defers on Web will make its document wrong, and instead of changing the document as well, he suggested changing `_PlatformViewGestureMixin`, which I adopted.

mdebbar (approved):
LGTM

PR #174315, opened 2025-08-22
Implement Overlay.of with inherited widget

chunhtai:
for https://github.com/flutter/flutter/pull/174239

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/framework.dart
+++ b/packages/flutter/lib/src/widgets/framework.dart
@@ -2449,6 +2449,7 @@ abstract class BuildContext {
   /// Returns the nearest widget of the given [InheritedWidget] subclass `T` or
   /// null if an appropriate ancestor is not found.
   ///
+  /// {@template flutter.widgets.BuildContext.getInheritedWidgetOfExactType}
   /// This method does not introduce a dependency the way that the more typical
   /// [dependOnInheritedWidgetOfExactType] does, so this context will not be
   /// rebuilt if the [InheritedWidget] changes. This function is meant for those
@@ -2465,6 +2466,7 @@ abstract class BuildContext {
   /// that value is not cached and/or reused later.
   ///
   /// Calling this method is O(1) with a small constant factor.
+  /// {@endtemplate}
   T? getInheritedWidgetOfExactType<T extends InheritedWidget>();
 
   /// Obtains the element corresponding to the nearest widget of the given type `T`,
--- a/packages/flutter/lib/src/widgets/lookup_boundary.dart
+++ b/packages/flutter/lib/src/widgets/lookup_boundary.dart
@@ -105,6 +105,34 @@ class LookupBoundary extends InheritedWidget {
     return candidate.widget as T;
   }
 
+  /// Obtains the nearest widget of the given type `T` within the current
+  /// [LookupBoundary] of `context`, which must be the type of a concrete
+  /// [InheritedWidget] subclass.
+  ///
+  /// This method behaves exactly like
+  /// [BuildContext.getInheritedWidgetOfExactType], except it only
+  /// considers [InheritedWidget]s of the specified type `T` between the
+  /// provided [BuildContext] and its closest [LookupBoundary] ancestor.
+  /// [InheritedWidget]s past that [LookupBoundary] are invisible to this
+  /// method. The root of the tree is treated as an implicit lookup boundary.
+  ///
+  /// {@macro flutter.widgets.BuildContext.getInheritedWidgetOfExactType}
+  static T? getInheritedWidgetOfExactType<T extends InheritedWidget>(
+    BuildContext context, {
+    Object? aspect,
+  }) {
+    // The following call makes sure that context depends on something so
+    // Element.didChangeDependencies is called when context moves in the tree
+    // even when requested dependency remains unfulfilled (i.e. null is
+    // returned).
+    context.dependOnInheritedWidgetOfExactType<LookupBoundary>();
+    final InheritedElement? candidate = getElementForInheritedWidgetOfExactType<T>(context);
+    if (candidate == null) {
+      return null;
+    }
+    return candidate.widget as T;
+  }
+
   /// Obtains the element corresponding to the nearest widget of the given type
   /// `T` within the current [LookupBoundary] of `context`.
   ///
--- a/packages/flutter/lib/src/widgets/lookup_boundary.dart
+++ b/packages/flutter/lib/src/widgets/lookup_boundary.dart
@@ -121,11 +121,6 @@ class LookupBoundary extends InheritedWidget {
     BuildContext context, {
     Object? aspect,
   }) {
-    // The following call makes sure that context depends on something so
-    // Element.didChangeDependencies is called when context moves in the tree
-    // even when requested dependency remains unfulfilled (i.e. null is
-    // returned).
-    context.dependOnInheritedWidgetOfExactType<LookupBoundary>();
     final InheritedElement? candidate = getElementForInheritedWidgetOfExactType<T>(context);
     if (candidate == null) {
       return null;
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -421,7 +421,9 @@ class _OverlayEntryWidgetState extends State<_OverlayEntryWidget> {
       child: _RenderTheaterMarker(
         theater: _theater,
         overlayEntryWidgetState: this,
-        child: widget.entry.builder(context),
+        // Use a Builder so that the `widget.entry.builder` can have access to
+        // _RenderTheaterMarker.of
+        child: Builder(builder: widget.entry.builder),
       ),
     );
   }
@@ -540,10 +542,6 @@ class Overlay extends StatefulWidget {
   /// If `rootOverlay` is set to true, the state from the furthest instance of
   /// this class is given instead. Useful for installing overlay entries above
   /// all subsequent instances of [Overlay].
-  ///
-  /// This method registers dependency on the Overlay element. The widget
-  /// associated with the given context rebuilds if it's reparented to a
-  /// different Overlay.
   /// {@endtemplate}
   ///
   /// See also:
@@ -611,20 +609,22 @@ class Overlay extends StatefulWidget {
   ///    and throws if an [Overlay] is not found.
   static OverlayState? maybeOf(BuildContext context, {bool rootOverlay = false}) {
     InheritedElement? element =
-        LookupBoundary.getElementForInheritedWidgetOfExactType<_OverlayScope>(context);
+        LookupBoundary.getElementForInheritedWidgetOfExactType<_RenderTheaterMarker>(context);
     if (rootOverlay) {
       InheritedElement? walker = element;
       while (walker != null) {
         element = walker;
         walker!.visitAncestorElements((Element ancestor) {
-          walker = LookupBoundary.getElementForInheritedWidgetOfExactType<_OverlayScope>(ancestor);
+          walker = LookupBoundary.getElementForInheritedWidgetOfExactType<_RenderTheaterMarker>(
+            ancestor,
+          );
           return false;
         });
         assert(element != walker); // can only happen if Overlay element is the root.
       }
     }
     if (element != null) {
-      return (element.widget as _OverlayScope).state;
+      return (element.widget as _RenderTheaterMarker).overlayEntryWidgetState.widget.overlayState;
     }
 
     return null;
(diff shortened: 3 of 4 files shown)


packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -536,11 +536,15 @@ class Overlay extends StatefulWidget {
   /// OverlayState overlay = Overlay.of(context);
   /// ```
   ///
+  /// {@template flutter.widgets.Overlay.of}
   /// If `rootOverlay` is set to true, the state from the furthest instance of
   /// this class is given instead. Useful for installing overlay entries above
   /// all subsequent instances of [Overlay].
   ///
-  /// This method can be expensive (it walks the element tree).
+  /// This method register dependency on the Overlay element. A rebuild on the
```
LongCatIsLooong: registeres

LongCatIsLooong: maybe a bit more specific like "the widget associated with the given `context` rebuilds if it's reparented to a different Overlay"?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -599,21 +603,32 @@ class Overlay extends StatefulWidget {
...
   static OverlayState? maybeOf(BuildContext context, {bool rootOverlay = false}) {
-    return rootOverlay
-        ? LookupBoundary.findRootAncestorStateOfType<OverlayState>(context)
-        : LookupBoundary.findAncestorStateOfType<OverlayState>(context);
+    InheritedElement? element =
+        LookupBoundary.getElementForInheritedWidgetOfExactType<_OverlayScope>(context);
+    if (rootOverlay) {
+      InheritedElement? walker = element;
+      while (walker != null) {
+        element = walker;
```
LongCatIsLooong: What if I subscribe to the root Overlay then a new Overlay is added so the old root is no longer the root?

LongCatIsLooong:
The google testing failures seem legit too. Looks like you may not be able to call `Overlay.of` in `initState` if you want to establish a dependency?

chunhtai:
> The google testing failures seem legit too. Looks like you may not be able to call `Overlay.of` in `initState` if you want to establish a dependency?

yeah I have to migrate them, you are not suppose to look up ancestor chain during initState anyway. I will see how bad it is and may do a migration guide

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: right, let me think about how to fix this

LongCatIsLooong:
> > The google testing failures seem legit too. Looks like you may not be able to call `Overlay.of` in `initState` if you want to establish a dependency?
> 
> yeah I have to migrate them, you are not suppose to look up ancestor chain during initState anyway. I will see how bad it is and may do a migration guide

For `OverlayEntry` related use cases the new dependency established via the `Overlay.of` API after this change isn't really necessary tho (since those APIs are mostly imperative). According to the `initState` doc it seems it's OK to look up ancestors (however you're not supposed to use inherited widgets yet, not sure why these two are different).

However `Overlay.of` establishing dependency does make it more consistent with other `.of` APIs, I wish we had given the no dependency version a different name so we could keep both.

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: After looking into the code this is not needed, so if a root is no longer a root without the entirely tree to be build from scratch, it must be keep alive by a globalkey.

The globalkey reparent will cause the `activate` to be called in the entire subtree, which will causes didChangeDependencies and markNeedsBuild(if it has any dependencies) of the entire to be called.

I will add a test to ensure this behavior

chunhtai:
> For `OverlayEntry` related use cases the new dependency established via the `Overlay.of` API after this change isn't really necessary tho (since those APIs are mostly imperative). According to the `initState` doc it seems it's OK to look up ancestors (however you're not supposed to use inherited widgets yet, not sure why these two are different).
> 
> However `Overlay.of` establishing dependency does make it more consistent with other `.of` APIs, I wish we had given the no dependency version a different name so we could keep both.

Yes this is unfortunate, but I can't think of a good way to mitigate this. I still plan to make the hard switch, Do you have any other suggestion?

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: this is actually quite funny, this means any dependency will be fine, not even need to be on Overlay

LongCatIsLooong: > The globalkey reparent will cause the `activate` to be called in the entire subtree, which will causes didChangeDependencies and markNeedsBuild(if it has any dependencies) of the entire to be called.

`Element.activate` doesn't guarantee rebuild iirc (however `State.activate` does, for historical reasons I guess).

chunhtai: It will, through didChangeDependencies

https://github.com/flutter/flutter/blob/a8bd12229b0961c7293e82ada91123d731383c54/packages/flutter/lib/src/widgets/framework.dart#L4771

LongCatIsLooong:
> > For `OverlayEntry` related use cases the new dependency established via the `Overlay.of` API after this change isn't really necessary tho (since those APIs are mostly imperative). According to the `initState` doc it seems it's OK to look up ancestors (however you're not supposed to use inherited widgets yet, not sure why these two are different).
> > However `Overlay.of` establishing dependency does make it more consistent with other `.of` APIs, I wish we had given the no dependency version a different name so we could keep both.
> 
> Yes this is unfortunate, but I can't think of a good way to mitigate this. I still plan to make the hard switch, Do you have any other suggestion?

Maybe leave `Overlay.of` as-is and add the new static method to a class that's more closely associated with `OverlayPortal` (which is the only known use case of the new API it seems)?

chunhtai:
I am concern about having a different static method that does what `.of` usually do, but the actual `.of` does something else.

Do you think allow looking up ancestor in initstate is a good idea in the first place? Last time I talked with Michael, there is really no good reason that we use findAncestorState any more since we have inherited widget, that means all the .of method should be implement with inherited widget. WDYT?

LongCatIsLooong:
> Do you think allow looking up ancestor in initstate is a good idea in the first place? 

I don't think it's the best place to call OverlayEntry APIs. I know there are apps currently doing that, iirc one of them want the OverlayEntry to have the same lifespan as the main widget.

> Last time I talked with Michael, there is really no good reason that we use findAncestorState any more since we have inherited widget, that means all the .of method should be implement with inherited widget. WDYT?

Are you deprecating `findAncestorState`? I think findAncestorState probably has its own use cases: if you don't use `Overlay.of` there's no additional overhead whatsoever, as compared to inherited widgets with which you'll have to add entries to `_dependencies` and `_inheritedElements`. With `_inheritedElements` now being a persistent hashmap this should be negligible, especially for Overlays since you typically don't have a lot of Overlays in a widget tree so it definitely makes sense to switch over to inherited widgets. But if you know an ancestor widget is only one parent away from you (which is kinda common if you're writing you own widget) I think `findAncestorState` is the more common choice since you don't want to go out of the way to introduce a new InheritedWidget type (ParentDataWidget is one that I can think of).

chunhtai:
> I know there are apps currently doing that, iirc one of them want the OverlayEntry to have the same lifespan as the main widget.

They can and probably should initialize it in didChangeDependencies, and dispose it in `dispose`. We do this all the time.

> Are you deprecating findAncestorState?

not deprecating entirely, I am pretty sure there will be use case that only findAncestorState can do. e.g. you need to own the ancestor widget in order to migrate findAncestorState to inheritedWidget.

instead, avoid using findAncestorState to implement `.of` method

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -942,6 +959,16 @@ class _WrappingOverlayState extends State<_WrappingOverlay> {
   }
 }
 
+class _OverlayScope extends InheritedWidget {
```
LongCatIsLooong: I think there's already the `_RenderTheaterMarker` widget that can give you the `OverlayState`?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -536,11 +536,15 @@ class Overlay extends StatefulWidget {
   /// OverlayState overlay = Overlay.of(context);
   /// ```
   ///
+  /// {@template flutter.widgets.Overlay.of}
   /// If `rootOverlay` is set to true, the state from the furthest instance of
   /// this class is given instead. Useful for installing overlay entries above
   /// all subsequent instances of [Overlay].
   ///
-  /// This method can be expensive (it walks the element tree).
+  /// This method registers dependency on the Overlay element. The widget
```
LongCatIsLooong: Looks like it doesn't register dependency in the implementation?

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -599,21 +601,33 @@ class Overlay extends StatefulWidget {
...
-
   static OverlayState? maybeOf(BuildContext context, {bool rootOverlay = false}) {
-    return rootOverlay
-        ? LookupBoundary.findRootAncestorStateOfType<OverlayState>(context)
-        : LookupBoundary.findAncestorStateOfType<OverlayState>(context);
+    InheritedElement? element =
+        LookupBoundary.getElementForInheritedWidgetOfExactType<_RenderTheaterMarker>(context);
+    if (rootOverlay) {
+      InheritedElement? walker = element;
+      while (walker != null) {
```
LongCatIsLooong: it looks like there's already a `_rootRenderTheaterMarkerOf` method for looking up the root Overlay?

LongCatIsLooong (approved):

packages/flutter/lib/src/widgets/overlay.dart
chunhtai: good call

packages/flutter/lib/src/widgets/framework.dart
```diff
@@ -2449,6 +2449,7 @@ abstract class BuildContext {
   /// Returns the nearest widget of the given [InheritedWidget] subclass `T` or
   /// null if an appropriate ancestor is not found.
   ///
+  /// {@template flutter.widgets.BuildContext.getInheritedWidgetOfExactType}
   /// This method does not introduce a dependency the way that the more typical
   /// [dependOnInheritedWidgetOfExactType] does, so this context will not be
```
LongCatIsLooong, line 2454: Does this work if you use this template on a non-`Element` class member, since this is referring to `dependOnInhritedWidgetOfExactType`?

packages/flutter/lib/src/widgets/lookup_boundary.dart
```diff
@@ -105,6 +105,34 @@ class LookupBoundary extends InheritedWidget {
...
+  /// {@macro flutter.widgets.BuildContext.getInheritedWidgetOfExactType}
+  static T? getInheritedWidgetOfExactType<T extends InheritedWidget>(
+    BuildContext context, {
+    Object? aspect,
+  }) {
+    // The following call makes sure that context depends on something so
+    // Element.didChangeDependencies is called when context moves in the tree
+    // even when requested dependency remains unfulfilled (i.e. null is
+    // returned).
+    context.dependOnInheritedWidgetOfExactType<LookupBoundary>();
```
LongCatIsLooong: Does the context have to depend on the LookupBoundary for `getInheritedX` too? I thought this was an one-off query?

LongCatIsLooong (commented):
LGTM with 2 nits.

packages/flutter/lib/src/widgets/lookup_boundary.dart
LongCatIsLooong: Oh LookupBoundary has a static method with the same name so that works.

packages/flutter/lib/src/widgets/framework.dart
chunhtai: The doc will try to find the reference from the context in the lookupBoundary scope, which would reference LookupBoundary's version

packages/flutter/lib/src/widgets/lookup_boundary.dart
chunhtai: yeah you are right, we shouldn't need this depends. since there shouldn't be reflective in any case

packages/flutter/lib/src/widgets/framework.dart
LongCatIsLooong: I replied but to the wrong comment 🤦 https://github.com/flutter/flutter/pull/174315#discussion_r2305365047. Yeah sg.

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -2192,14 +2180,39 @@ class _RenderTheaterMarker extends InheritedWidget {
...
+  }) {
+    if (targetRootOverlay) {
+      final InheritedElement? ancestor = _rootRenderTheaterMarkerOf(
+        LookupBoundary.getElementForInheritedWidgetOfExactType<_RenderTheaterMarker>(context),
+      );
+      assert(ancestor == null || ancestor.widget is _RenderTheaterMarker);
+      if (ancestor == null) {
+        return null;
+      }
+      return context.dependOnInheritedElement(ancestor) as _RenderTheaterMarker;
```
LongCatIsLooong: it looks like google testing is failing because of this creating a dependency regardless of the `createDependency` flag?

PR #174321, opened 2025-08-23
Refactor: Migrate page transition builder class to widgets

rkishan516:
Changes
* Move PageTransitionsBuilder from `material/page_transitions_theme.dart` to `widget/page_transitions_builder.dart`

part of: #172929

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/page_transitions_builder.dart
+++ b/packages/flutter/lib/src/widgets/page_transitions_builder.dart
@@ -1,0 +1,88 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// @docImport 'package:flutter/material.dart';
+library;
+
+import 'basic.dart';
+import 'framework.dart';
+import 'pages.dart';
+import 'transitions.dart';
+
+/// Defines a page transition animation for a [PageRoute].
+///
+/// PageTransitionsBuilder can be used directly with widget layer primitives
+/// in any design system. Custom [PageRoute] subclasses can accept a
+/// PageTransitionsBuilder and delegate to its [buildTransitions] method when
+/// overriding [ModalRoute.buildTransitions]. This enables reusable transition
+/// animations that work with [Navigator] and other navigation primitives.
+///
+/// ## Example usage
+///
+/// {@tool dartpad}
+/// This example shows how to create a custom [PageTransitionsBuilder] that
+/// slides the new page in from the right while fading it in, and how to use
+/// it with a custom [PageRoute].
+///
+/// ** See code in examples/api/lib/widgets/page_transitions_builder/page_transitions_builder.0.dart **
+/// {@end-tool}
+///
+/// As an example of usage, in the Material library this class is used by
+/// [PageTransitionsTheme] to define a [MaterialPageRoute] page transition
+/// animation. Apps can configure the map of builders for
+/// [ThemeData.pageTransitionsTheme] to customize the default
+/// [MaterialPageRoute] page transition animation for different platforms.
+///
+/// See also:
+///
+///  * [PageTransitionsTheme], which uses this class to configure page transitions.
+///  * [MaterialPageRoute], which uses this class to build its transition.
+///  * [FadeUpwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android O.
+///  * [OpenUpwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android P.
+///  * [ZoomPageTransitionsBuilder], which defines the default page transition
+///    that's similar to the one provided in Android Q.
+///  * [CupertinoPageTransitionsBuilder], which defines a horizontal page
+///    transition that matches native iOS page transitions.
+///  * [FadeForwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android U.
+abstract class PageTransitionsBuilder {
+  /// Abstract const constructor. This constructor enables subclasses to provide
+  /// const constructors so that they can be used in const expressions.
+  const PageTransitionsBuilder();
+
+  /// Provides a secondary transition to the previous route.
+  ///
+  /// {@macro flutter.widgets.delegatedTransition}
+  DelegatedTransitionBuilder? get delegatedTransition => null;
+
+  /// {@macro flutter.widgets.TransitionRoute.transitionDuration}
+  ///
+  /// Defaults to 300 milliseconds.
+  Duration get transitionDuration => const Duration(milliseconds: 300);
+
+  /// {@macro flutter.widgets.TransitionRoute.reverseTransitionDuration}
+  ///
+  /// Defaults to 300 milliseconds.
+  Duration get reverseTransitionDuration => transitionDuration;
+
+  /// Wraps the child with one or more transition widgets which define how [route]
+  /// arrives on and leaves the screen.
+  ///
+  /// Subclasses override this method to create a transition animation.
+  ///
+  /// The [MaterialPageRoute.buildTransitions] method is an example of a method
+  /// that uses this to build a transition. It looks up the current
+  /// [PageTransitionsTheme] with `Theme.of(context).pageTransitionsTheme`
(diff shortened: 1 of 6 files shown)


victorsanni (commented):
I would avoid design-language-specific documentation and references (Material, Theme and ThemeData etc). Try to make this as standalone and as design-language-agnostic as possible. Instead, move the Material-specific documentation to the relevant Material widgets.

victorsanni:
> I would avoid design-language-specific documentation and references (Material, Theme and ThemeData etc). Try to make this as standalone and as design-language-agnostic as possible. Instead, move the Material-specific documentation to the relevant Material widgets.

I asked offline and found this is not entirely true. Referencing `Material` APIs in `widgets` documentation is okay, as long as it is used as an example of a design-language-specific implementation (vs. treating `Material` like a first-class citizen).

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -0,0 +1,66 @@
...
+///  * [FadeUpwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android O.
+///  * [OpenUpwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android P.
+///  * [ZoomPageTransitionsBuilder], which defines the default page transition
+///    that's similar to the one provided in Android Q.
+///  * [CupertinoPageTransitionsBuilder], which defines a horizontal page
+///    transition that matches native iOS page transitions.
+///  * [FadeForwardsPageTransitionsBuilder], which defines a page transition
+///    that's similar to the one provided by Android U.
```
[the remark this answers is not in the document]
victorsanni: +1

victorsanni (commented):
The documentation suggestions from gemini look like a good place to start. 

Also are there tests for this class that can be moved here? I don't think we want to add a new file + class in `widgets/` without some sort of test.

rkishan516:
I don't think there are tests for `PageTransitionsBuilder`. But what kind of new tests we want to add?

victorsanni:
> I don't think there are tests for PageTransitionsBuilder. But what kind of new tests we want to add?

A test could make a mock transitions builder that extends `PageTransitionsBuilder` and verify that it works.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -0,0 +1,66 @@
...
+  /// Defaults to 300 milliseconds.
+  Duration get reverseTransitionDuration => transitionDuration;
+
+  /// Wraps the child with one or more transition widgets which define how [route]
+  /// arrives on and leaves the screen.
+  ///
+  /// The [MaterialPageRoute.buildTransitions] method looks up the
+  /// current [PageTransitionsTheme] with `Theme.of(context).pageTransitionsTheme`
+  /// and delegates to this method with a [PageTransitionsBuilder] based
+  /// on the theme's [ThemeData.platform].
```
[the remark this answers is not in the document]
victorsanni: +1 here

victorsanni:
Approach with test so far is looking good. Can you adopt the gemini doc recommendations and fix the failing tests in CI?

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -0,0 +1,107 @@
...
+          },
+        ),
+      ),
+      '/test': (BuildContext context) => const Material(child: Text('test page')),
+    };
+
+    await tester.pumpWidget(
+      MaterialApp(
+        theme: ThemeData(
+          pageTransitionsTheme: PageTransitionsTheme(
```
victorsanni, line 60: It would be nice to have an additional test not as tightly coupled to `Material`. Maybe without using `PageTransitionsTheme`, something more bare-bones. Just to show how to integrate `PageTransitionsBuilder` into alternative design languages.

rkishan516: I have added another test agnostic to design system, I don't know that this is enough to show, how to integrate `PageTransitionsBuilder`  into alternative design languages.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -0,0 +1,71 @@
...
+
+/// @docImport 'package:flutter/material.dart';
+library;
+
+import 'basic.dart';
+import 'framework.dart';
+import 'pages.dart';
+import 'transitions.dart';
+
+/// Defines a page transition animation for a [PageRoute].
```
victorsanni, line 13: It would be nice if there was a paragraph after this line with more explanations on how `PageTransitionsBuilder` can be used in an arbitrary (non-Material) design system. With Navigator etc. and more widget layer primitives. Like a textual description of the test with custom navigator and page route.

victorsanni (commented):
Looking good so far, one final comment asking for more documentation.

rkishan516:
@victorsanni I have updated the paragraph, please have a look.

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -0,0 +1,247 @@
...
+            Animation<double> secondaryAnimation,
+            Widget child,
+          ) {
+            return FadeTransition(
+              opacity: animation,
+              child: ScaleTransition(
+                scale: Tween<double>(
+                  begin: 0.5,
+                  end: 1.0,
+                ).animate(CurvedAnimation(parent: animation, curve: Curves.easeInOut)),
```
victorsanni: this might be causing a memory leak, consider using `CurveTween` or something similar instead.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -0,0 +1,78 @@
...
+/// This class is used by [PageTransitionsTheme] to define a [MaterialPageRoute]
+/// page transition animation. Apps can configure the map of builders for
+/// [ThemeData.pageTransitionsTheme] to customize the default
+/// [MaterialPageRoute] page transition animation for different platforms.
+///
+/// PageTransitionsBuilder can also be used directly with widget layer primitives
+/// in any design system. Custom [PageRoute] subclasses can accept a
+/// PageTransitionsBuilder and delegate to its [buildTransitions] method when
+/// overriding [ModalRoute.buildTransitions]. This enables reusable transition
+/// animations that work with [Navigator] and other navigation primitives without
```
victorsanni: Not sure we need to explicitly say "without requiring Material Design components". The idea is the class should be understandable as a standalone concept for developers who don't even know Material exists.

victorsanni: So this should precede the paragraph on `PageTransitionsTheme,` while that paragraph should be prefixed with an annotation that it is an example of how this class can be used.

rkishan516: Yup, Make sense.

rkishan516: I have pushed snipped directly in doc, right now. Do you feel, we should create a example file for this?

victorsanni (commented):
To fix `linux analyze`, run `dart format packages/flutter/test/widgets/page_transitions_builder_test.dart`

victorsanni (approved):

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -0,0 +1,87 @@
...
+///
+/// {@tool dartpad}
+/// This example shows how to create a custom [PageTransitionsBuilder] that
+/// slides the new page in from the right while fading it in, and how to use
+/// it with a custom [PageRoute].
+///
+/// ** See code in examples/api/lib/widgets/page_transitions_builder/page_transitions_builder.0.dart **
+/// {@end-tool}
+///
+/// This class is used by [PageTransitionsTheme] to define a [MaterialPageRoute]
```
MitchellGoodwin: Turbo nit: add to the beginning of this, "As an example of usage, in the Material library this class is used by...."

MitchellGoodwin (approved):
LGTM, just left one small nit. Feel free to ignore.

PR #174336, opened 2025-08-24
Add PredictiveBackGestureBuilder

TheLastFlame:
This PR introduces a public PredictiveBackGestureBuilder, providing developers with a simple way to implement predictive back animations for custom transitions.

The Material animation variant has been extracted into a separate widget.

Usage:
PredictiveBackPageTransitionsBuilder: https://github.com/flutter/flutter/pull/174337
Dialogs / BottomSheets: \<soon\>

**Critical Issue:**
During the implementation of PredictiveBackPageTransitionsBuilder, it was discovered that the current approach using multiple _predictiveBackObservers in WidgetsBindingObserver only works if we use the same types of transitions and do not consider dialogs or bottom sheets.

If the top page doesn't have an observer, but the one below it does, the back gesture stops working because the default handler is disabled.
I don't have any ideas on how to solve this yet.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/predictive_back_transition.dart
+++ b/packages/flutter/lib/src/material/predictive_back_transition.dart
@@ -1,3 +1,5 @@
+import 'dart:math';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
@@ -94,7 +96,7 @@ class PredictiveBackTransition extends StatelessWidget {
         ? renderObject.size.width
         : MediaQuery.widthOf(context);
 
-    final double maxShift = (width / _kDivisionFactor) - _kMargin;
+    final double maxShift = max(0, (width / _kDivisionFactor) - _kMargin);
 
     final SwipeEdge? swipeEdge = currentBackEvent?.swipeEdge;
 
@@ -112,7 +114,7 @@ class PredictiveBackTransition extends StatelessWidget {
     final double startTouchY = startBackEvent?.touchOffset?.dy ?? 0;
     final double currentTouchY = currentBackEvent?.touchOffset?.dy ?? 0;
 
-    final double yShiftMax = (height / _kDivisionFactor) - _kMargin;
+    final double yShiftMax = max(0, (height / _kDivisionFactor) - _kMargin);
 
     // Apply the decelerated gesture progress to the Y-shift so the preview is
     // more apparent at the start (matches Android docs recommendation).
--- a/packages/flutter/lib/src/material/predictive_back_transition.dart
+++ b/packages/flutter/lib/src/material/predictive_back_transition.dart
@@ -89,8 +89,8 @@ class PredictiveBackSharedElementTransition extends StatelessWidget {
       useInterpolation ? kCurve.transform(progress.value) : progress.value;
 
   double _calcXShift(BuildContext context) {
-    final RenderBox? renderObject = context.findRenderObject() as RenderBox?;
-    final double width = renderObject?.size.width ?? MediaQuery.widthOf(context);
+    final RenderObject? renderObject = context.findRenderObject();
+    final double width = renderObject is RenderBox ? renderObject.size.width : MediaQuery.widthOf(context);
 
     final double maxShift = (width / _kDivisionFactor) - _kMargin;
 
--- a/packages/flutter/lib/src/material/predictive_back_transition.dart
+++ b/packages/flutter/lib/src/material/predictive_back_transition.dart
@@ -126,7 +126,7 @@ class PredictiveBackSharedElementTransition extends StatelessWidget {
 
   double _calcScale() => 1 - (1 - _kMinScale) * _gestureProgress;
 
-  Matrix4 _createTransformMatrix(BuildContext context, double animationValue) {
+  Matrix4 _createTransformMatrix(BuildContext context) {
     final double scale = _calcScale();
     final double xShift = useXShift ? _calcXShift(context) : 0;
     final double yShift = useYShift ? _calcYShift(context) : 0;
@@ -141,7 +141,7 @@ class PredictiveBackSharedElementTransition extends StatelessWidget {
     return MatrixTransition(
       animation: progress,
       alignment: alignment ?? Alignment.center,
-      onTransform: (double animationValue) => _createTransformMatrix(context, animationValue),
+      onTransform: (_) => _createTransformMatrix(context),
       child: child,
     );
   }
--- a/packages/flutter/lib/src/material/predictive_back_transition.dart
+++ b/packages/flutter/lib/src/material/predictive_back_transition.dart
@@ -1,0 +1,148 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+/// A widget that applies Android-style predictive-back transform to `child`.
+class PredictiveBackSharedElementTransition extends StatelessWidget {
+  /// Creates a predictive back shared element transition.
+  ///
+  /// The [progress] animation drives the transform (scale + translation).
+  /// Use [startBackEvent] and [currentBackEvent] to compute Y-shift from the
+  /// user's touch movement and the [SwipeEdge] from [currentBackEvent] to
+  /// determine the X direction. Defaults: [useXShift]=true, [useYShift]=true,
+  /// [useInterpolation]=true. If [alignment] is omitted, the transform uses
+  /// `Alignment.center`.
+  const PredictiveBackSharedElementTransition({
+    super.key,
+    required this.progress,
+    required this.startBackEvent,
+    required this.currentBackEvent,
+    this.alignment,
+    this.useXShift = true,
+    this.useYShift = true,
+    this.useInterpolation = true,
+    required this.child,
+  });
+
+  /// Animation that drives the transform applied to [child].
+  ///
+  /// The animation's current `value` (typically in the range `0.0`..`1.0`)
+  /// is used to compute scale and shifts. When [useInterpolation] is `true`
+  /// the implementation applies the Android back-gesture curve to the raw
+  /// `value` before computing the transform.
+  final Animation<double> progress;
+
+  /// The initial back gesture event when the gesture began.
+  ///
+  /// Used as a reference point for computing the Y-axis shift based on how far
+  /// the touch point has moved since the gesture started. May be `null` if
+  /// such data is not available.
+  final PredictiveBackEvent? startBackEvent;
+
+  /// The most recent back gesture event.
+  ///
+  /// Provides the current touch offset and swipe edge (left/right) which are
+  /// used to compute direction and magnitude of X/Y shifts. May be `null`.
+  final PredictiveBackEvent? currentBackEvent;
+
+  /// If non-null, the alignment for the transform applied to [child].
+  ///
+  /// Defaults to `Alignment.center` when omitted.
+  final Alignment? alignment;
+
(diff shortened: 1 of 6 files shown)


packages/flutter/lib/src/material/predictive_back_transition.dart
```diff
@@ -0,0 +1,148 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+/// A widget that applies Android-style predictive-back transform to `child`.
+class PredictiveBackSharedElementTransition extends StatelessWidget {
```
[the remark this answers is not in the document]
justinmc: Take it up with the Material team, Gemini.

In seriousness though, I guess the name of the class should match the name of the file.

justinmc: How does this class relate to _PredictiveBackSharedElementPageTransition? Is it for the incoming page?

packages/flutter/lib/src/widgets/predictive_back_builder.dart
```diff
@@ -0,0 +1,189 @@
...
+
+import 'binding.dart';
+import 'framework.dart';
+import 'routes.dart';
+
+/// Phases of a predictive-back gesture.
+///
+/// These phases describe the lifecycle of the platform back gesture as it
+/// relates to widgets that want to respond visually.
+enum PredictiveBackPhase {
```
justinmc, line 13: Looks like this has been moved from predictive_back_page_transitions_builder.dart. I think this PR is kind of in a partial state since it doesn't include the removal of this enum from that file? Is there a better way to organize this?

I do agree it's probably right to make some of the stuff in that file public now. I want it to be easy for people to write their own predictive back page transitions.

justinmc (commented):
Thanks so much for taking this on! The video in the other PR (https://github.com/flutter/flutter/pull/174337) looks great. I was really hoping we'd be able to get the incoming page transition working via delegatedTransition and it looks like it works!

Is the critical issue mentioned in the description involving _predictiveBackObservers resolved now, or can you elaborate?

We will need tests for this if we want to merge it separately from https://github.com/flutter/flutter/pull/174337.

Can you reorganize this PR and https://github.com/flutter/flutter/pull/174337 so that it's easier to see what was moved where, etc.?

packages/flutter/lib/src/material/predictive_back_transition.dart
TheLastFlame: @justinmc, I've created a generic builder for PredictiveBack in the Widgets layer so that developers can easily create custom implementations without a dependency on Material.
However, this specific animation is part of the Material spec, so it has been extracted into a separate widget within the Material library.

This is a generic animation builder that can be used for pages as well as dialogs, bottom sheets, and drawers.

In _PredictiveBackSharedElementPageTransition, it's used for both the bottom and top pages, as their transformations are identical and synchronous.

packages/flutter/lib/src/widgets/predictive_back_builder.dart
TheLastFlame: @justinmc, yes, this has been moved and partially reworked. I will describe the partial state separately below.

TheLastFlame:
@justinmc, regarding how these PRs are organized:

I felt they were too large to be reviewed easily in a single PR, especially since I suspect they might break a number of automated tests. That's why I decided to split them into three parts:

1. The core refactor: creating public builders that everyone, including us, can use.
2. Implementing the full Predictive Back transition based on these builders.
3. Implementing Predictive Back for modal elements like dialogs, bottom sheets, etc.

I might have been wrong in thinking this was the best approach, so if you'd prefer, we can reorganize it.

TheLastFlame:
The old versions of _PredictivePhase and PredictiveBackGestureDetector are kept for now, as they are used by two transitions that are being addressed in the second PR.

Also, I'm not sure it's worth working on the old Fullscreen transition, as it's unlikely that anyone will use it going forward. Besides, it needs fixing – its gesture interpolation is pretty broken.

TheLastFlame:
And now for the bad news.
I still can't figure out how to fix the core conceptual problem with multiple observers.
This is a critical issue, and we can't move forward without solving it.
I haven't bothered with tests or anything else because the current implementation is simply not viable.

We need to somehow link different PredictiveBackBuilders together in a clean, organic way.
Because when they aren't aware of each other, this is what happens:

https://github.com/user-attachments/assets/a00be541-6700-44ff-88de-fe7a8978bab8


The predictive back events are being caught by the bottom PredictiveBackBuilder. However, only the top one should be handling navigation commands. And if the top one doesn't exist, nothing happens.

And we can't just pass this responsibility to the bottom PredictiveBackBuilder because:
a) for some reason, that didn't work (the predictive transition stopped working entirely, I haven't investigated why yet)
b) it won't let us use Hero animations, which is highly desirable for things like cards in Gmail. (To do that, we need to ensure that the screen's animation controller is not affected during the pre-commit phase of the Predictive Back animation, so that the Hero animation can run after the commit).
@justinmc

TheLastFlame:
Well, it seems that I was able to successfully resolve this issue.

I've also added predictive animations for DialogRoute and ModalBottomSheetRoute.

@justinmc, you talked about reorganizing these pr.... would you prefer me to merge them into one?

[screen-20250903-013521-1756827312455.webm](https://github.com/user-attachments/assets/758054bb-e8aa-419d-9e81-4adfb380eee8)

packages/flutter/lib/src/widgets/predictive_back_builder.dart
```diff
@@ -0,0 +1,196 @@
+import 'package:flutter/services.dart';
+
+import 'binding.dart';
+import 'framework.dart';
+import 'routes.dart';
+
+/// Phases of a predictive-back gesture.
+///
+/// These phases describe the lifecycle of the platform back gesture as it
+/// relates to widgets that want to respond visually.
```
justinmc, line 12: I think you should add a "See also" section that points to something like PredictiveBackPageTransitionsBuilder, so that users who don't even know what predictive back is will be able to figure it out without us explaining it over and over again.  Same for any other classes related to predictive back.

packages/flutter/lib/src/widgets/predictive_back_builder.dart
```diff
@@ -0,0 +1,196 @@
...
+enum PredictiveBackPhase {
+  /// No back gesture is in progress.
+  idle,
+
+  /// The back gesture has started
```
justinmc, line 17: Missing a period.

packages/flutter/lib/src/widgets/predictive_back_builder.dart
```diff
@@ -0,0 +1,196 @@
...
+/// back events via the framework binding (it implements
+/// [WidgetsBindingObserver]) and calls `transitionBuilder` with the current
+/// gesture phase and event data so the subtree can render gesture-driven
+/// UI (for example, a preview/transform of the page's content).
+///
+/// The widget does not itself implement a visual transition — that work is
+/// performed by the `transitionBuilder` callback which receives the current
+/// `PredictiveBackPhase`, the initial `startBackEvent` and the most recent
+/// `currentBackEvent`, as well as the `child` to embed.
+class PredictiveBackGestureBuilder extends StatefulWidget {
```
justinmc, line 43: This is replacing the old private _PredictiveBackGestureDetector?

packages/flutter/lib/src/widgets/predictive_back_builder.dart
```diff
@@ -0,0 +1,196 @@
...
+
+  @override
+  State<PredictiveBackGestureBuilder> createState() => _PredictiveBackGestureBuilderState();
+}
+
+class _PredictiveBackGestureBuilderState extends State<PredictiveBackGestureBuilder>
+    with WidgetsBindingObserver {
+  PredictiveBackPhase get phase => _phase;
+  PredictiveBackPhase _phase = PredictiveBackPhase.idle;
+  set phase(PredictiveBackPhase phase) {
```
justinmc, line 114: Would it make sense to make these setters and getters private? Now that the widget is public, it makes me nervous seeing this public API here. I expect something else to use it, but it's only used internally.

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -952,15 +955,36 @@ mixin WidgetsBinding
...
+            'Both $_controllingObserver and $observer tried to take control.',
+          );
+          _controllingObserver = observer;
+          _backGestureObservers.add(observer);
+        case PredictiveBackObserverBehavior.updateOnly:
+          _backGestureObservers.add(observer);
+        case PredictiveBackObserverBehavior.updateIfControlled:
+          updateIfControlledObservers.add(observer);
+        case null:
+          {}
```
justinmc, line 982: It looks like this is your solution to the multiple observer problem that you mentioned in https://github.com/flutter/flutter/pull/174336#issuecomment-3235608000.

This is actually a problem that pops up in a few other places in Flutter where we have this pattern where an event from the engine notifies observers. It works well when all observers are equal and all need to hear about the event. But sometimes, like in this case, you want only one specific observer to claim the event.

Here is an example where I think we have handled this poorly in the past:

https://github.com/flutter/flutter/blob/569ee71e35d79eef4fbf8bf2211711d0e161dcff/packages/flutter/lib/src/widgets/binding.dart#L943-L951

Really the observer that receives the event should depend on some hierarchy of Navigators in the widget tree.

So I want to make sure we're not enshrining a similar mistake here. Is there any case of a strange navigation setup where your solution won't be correct?

justinmc (commented):
Quick review here. Myself and most people working on Flutter at Google will be out next week, but I'll review this again when we're back.

I'm on board with doing this and https://github.com/flutter/flutter/pull/174337 separately as long as you're on board with adding tests to this PR separately. The way you've divided it up makes sense to me.

My biggest question mark is probably the way you solved the observer problem (comment about this below). We are going to have to do something fancy there like what you did, but because this is an existing problem, I want to make sure we get the solution right and set a good precedent.

Thank you again for helping Flutter nail this feature!

justinmc:
@TheLastFlame Are you still available to get this PR landed? I definitely want to get this and https://github.com/flutter/flutter/pull/174337 in. Sorry for the low communication here. If you can't get ahold of me you're welcome to ping me on Discord.

TheLastFlame:
@justinmc, sorry for disappearing, I've been very busy these past few weeks. I'll try to get back to these PRs in the next few days.

packages/flutter/lib/src/widgets/predictive_back_builder.dart
TheLastFlame: @justinmc, As far as I understand, it is designed this way so that the phase in which it is located cannot be changed from outside the widget. I cannot imagine a scenario in which this would be a good idea.
Given that this phase is simply transferred to the builder available to the user, they can make all the necessary modifications there.

packages/flutter/lib/src/widgets/predictive_back_builder.dart
TheLastFlame: @justinmc, technically, yes. Initially, I created PredictiveBackObserver, which combines _PredictiveBackGestureDetector and _PredictiveBackGestureListener, but then I renamed it to better reflect the final essence.

Piinks:

Piinks:

Piinks:

Piinks:
Since this is still blocked, I am going to close it for now. Please feel free to re-open this, or create a fresh PR when it is able to move forward. Thank you!

PR #174440, opened 2025-08-26
Remove dead code from snippet_generator

sigurdm:
Randomly stumbled on code that seems to no longer have clients.


sigurdm:
Yeah, I think it still makes sense

Piinks (approved):
LGTM!

PR #174463, opened 2025-08-26
[impeller] Support partitioned host buffer

eyebrowsoffire:
This change adds support for host buffer that keeps index buffers separate from other general-purpose data. This is a requirement for WebGL: https://registry.khronos.org/webgl/specs/latest/2.0/#BUFFER_OBJECT_BINDING

No current backends use the partitioned implementation, but once I add WebGL support it will use it to avoid https://github.com/flutter/flutter/issues/149603


engine/src/flutter/impeller/core/partitioned_host_buffer.h
```diff
@@ -0,0 +1,40 @@
...
+                             const EmplaceProc& cb) override;
+  virtual size_t GetMinimumUniformAlignment() const override;
+  virtual void Reset() override;
+  virtual TestStateQuery GetStateForTest(BufferCategory category) override;
+
+ private:
+  SimpleHostBuffer& _buffer(BufferCategory category);
+
+  SimpleHostBuffer dataBuffer_;
+  SimpleHostBuffer indexBuffer_;
```
gaaclarke: The linter called this out but this isn't proper naming for instance variables.

https://google.github.io/styleguide/cppguide.html#Variable_Names

engine/src/flutter/impeller/core/partitioned_host_buffer.h
```diff
@@ -0,0 +1,40 @@
...
+  virtual BufferView Emplace(size_t length,
+                             size_t align,
+                             BufferCategory category,
+                             const EmplaceProc& cb) override;
+  virtual size_t GetMinimumUniformAlignment() const override;
+  virtual void Reset() override;
+  virtual TestStateQuery GetStateForTest(BufferCategory category) override;
+
+ private:
+  SimpleHostBuffer& _buffer(BufferCategory category);
```
gaaclarke: Improper name for a function.

https://google.github.io/styleguide/cppguide.html#Function_Names

engine/src/flutter/impeller/entity/contents/host_buffer_unittests.cc
```diff
@@ -28,7 +28,7 @@ TEST_P(HostBufferTest, IdleWaiter) {
   auto mock_idle_waiter = std::make_shared<MockIdleWaiter>();
   {
     auto buffer = HostBuffer::Create(GetContext()->GetResourceAllocator(),
-                                     mock_idle_waiter, 256);
+                                     mock_idle_waiter, 256, false);
```
gaaclarke: You should have an argument comment when passing in literals as rvalues.

```suggestion
                                     mock_idle_waiter, 256, /*partition_by_category=*/false);
```

gaaclarke (changes requested):
Introducing polymorphism in HostBuffer adds complication, overhead, and hurts locality.  It's not necessary for what we are trying to do.

All you need to do is introduce a new datatype:

```c++
struct HostBuffers {
  HostBuffer data;
  std::optional<HostBuffer> indices;
};
```

eyebrowsoffire:
> Introducing polymorphism in HostBuffer adds complication, overhead, and hurts locality.  It's not necessary for what we are trying to do.
> 
> 
> 
> All you need to do is introduce a new datatype:
> 
> 
> 
> ```c++
> 
> struct HostBuffers {
> 
>   HostBuffer data;
> 
>   std::optional<HostBuffer> indices;
> 
> };
> 
> ```

The idea here is that most backends still should use a unified host buffer under the hood. Separation of indices into another buffer is pretty much specific to WebGL, so for other backends the specification of the buffer category is basically a no-op. But the backend-agnostic code needs to specify whether it is going to use the buffer for indexes or for other data, so that if the backend needs to separate it it can. Hence the polymorphism, we need different behavior for different backends.

However, it doesn't necessarily have to be dynamic polymorphism, I'm happy to refactor this a different way. But I'm not sure your suggestion on its own solves the problem. We can jump on a call to discuss the details and get aligned on a solution.

engine/src/flutter/impeller/renderer/renderer_unittests.cc
```diff
@@ -1040,9 +1060,16 @@ TEST_P(RendererTest, ArrayUniforms) {
...
+  auto data_host_buffer = HostBuffer::Create(
+      GetContext()->GetResourceAllocator(), GetContext()->GetIdleWaiter(),
+      GetContext()->GetCapabilities()->GetMinimumUniformAlignment());
+  auto indexes_host_buffer =
+      GetContext()->GetCapabilities()->NeedsPartitionedHostBuffer()
+          ? HostBuffer::Create(
+                GetContext()->GetResourceAllocator(),
+                GetContext()->GetIdleWaiter(),
+                GetContext()->GetCapabilities()->GetMinimumUniformAlignment())
+          : data_host_buffer;
```
gaaclarke: nit: We can probably pull this into a function since it happens a fair bit.

engine/src/flutter/impeller/entity/contents/content_context.h
```diff
@@ -290,18 +290,33 @@ class ContentContext {
   void ClearCachedRuntimeEffectPipeline(
       const std::string& unique_entrypoint_name) const;
 
-  /// @brief Retrieve the currnent host buffer for transient storage.
+  /// @brief Retrieve the current host buffer for transient storage of indexes
+  ///        used for indexed draws.
```
gaaclarke, line 294: It's worth noting that GetTransientsIndexesBuffer may return the same value as GetTransientsDataBuffer in certain circumstances.

engine/src/flutter/impeller/entity/contents/content_context.cc
```diff
@@ -555,10 +555,17 @@ ContentContext::ContentContext(
...
           context_->GetResourceAllocator(),
           context_->GetIdleWaiter(),
           context_->GetCapabilities()->GetMinimumUniformAlignment())),
+      indexes_host_buffer_(
+          context_->GetCapabilities()->NeedsPartitionedHostBuffer()
+              ? HostBuffer::Create(
+                    context_->GetResourceAllocator(),
+                    context_->GetIdleWaiter(),
+                    context_->GetCapabilities()->GetMinimumUniformAlignment())
+              : data_host_buffer_),
```
gaaclarke: Please move this into the constructor.  The order in which initializers are executed is controlled by the order in the header.  I think we have a lint that forces them to be in the same order but I think it's safer just to avoid order issues in initializers.

gaaclarke (approved):
LGTM, thanks, this is a much better approach I think.  It might be worth telling the gemini bot to double check it since there is a lot of small refactoring which it's good at checking.  I reviewed it too and looked all good.

gaaclarke:
fyi the test failure looks apropos, `ResetHostBufferBasedOnFrameBoundary`

PR #174480, opened 2025-08-27
[A11y] Add semantics for CupertinoExpansionTile

### LINKED ISSUE #173446: Add semantics for CupertinoExpansionTile

victorsanni:
Fixes [Add semantics for CupertinoExpansionTile](https://github.com/flutter/flutter/issues/173446)


packages/flutter/lib/src/cupertino/expansion_tile.dart
```diff
@@ -168,12 +169,26 @@ class _CupertinoExpansionTileState extends State<CupertinoExpansionTile> {
...
-      title: widget.title,
-      trailing: _buildIcon(context, animation),
-      backgroundColorActivated: CupertinoColors.transparent,
+    final CupertinoLocalizations localizations = CupertinoLocalizations.of(context);
+    final String onTapHint = _tileController.isExpanded
+        ? localizations.expansionTileExpandedTapHint
+        : localizations.expansionTileCollapsedTapHint;
+    String? semanticsHint;
+    if (MediaQuery.supportsAnnounceOf(context)) {
+      semanticsHint = _tileController.isExpanded
```
chunhtai: Can this hint also move to widget layer? for example, the wraps the semantics with hint with widget.headerBuilder of Expansible

victorsanni: I'm not sure here. The hint says 'double tap to collapse'. But that is an implementation detail on how the headerBuilder in Expansible is implemented. What if for example a developer has an Expansible that only expands or collapses if a FAB is tapped? Then double tapping the expansible itself will have no effect.

chunhtai: yeah, I think you are right since the ontap is implemented in this widget.

packages/flutter/lib/src/cupertino/localizations.dart
```diff
@@ -308,6 +310,30 @@ abstract class CupertinoLocalizations {
   /// The label for the back button, used in [CupertinoNavigationBar] and [CupertinoSliverNavigationBar].
   String get backButtonLabel;
 
+  /// The semantics hint to describe the tap action on an expanded
+  /// [CupertinoExpansionTile] on iOS and macOS. This is appended to the [collapsedHint]
+  /// hint to provide a more detailed description of the action, e.g. "Expanded
+  /// double tap to collapse".
+  String get expansionTileExpandedHint => 'double tap to collapse';
```
chunhtai, line 317: For iOS i remember it will already announce `double tap to activate` if you have onTap action. Can you check whether this will cause it to pronounce both this and `double tap to activate` back to back?

packages/flutter/lib/src/material/expansion_tile.dart
```diff
@@ -584,23 +584,15 @@ class _ExpansionTileState extends State<ExpansionTile> {
...
-      case TargetPlatform.macOS:
-        semanticsHint = _tileController.isExpanded
-            ? '${localizations.collapsedHint}\n ${localizations.expansionTileExpandedHint}'
-            : '${localizations.expandedHint}\n ${localizations.expansionTileCollapsedHint}';
-      case TargetPlatform.android:
-      case TargetPlatform.fuchsia:
-      case TargetPlatform.linux:
-      case TargetPlatform.windows:
-        break;
+    if (MediaQuery.supportsAnnounceOf(context)) {
```
chunhtai: Why? I think the reason that we only apply this to ios and macos is that android supports onTapHint while ios and other platform ignores it

victorsanni: Ah I see

packages/flutter/lib/src/cupertino/localizations.dart
victorsanni: I tested, it doesn't announce `double tap to activate`. Is this intended?

chunhtai: I am not sure what exact mechanism triggers that to be announce, but sounds good to me now

chunhtai (approved):
LGTM

PR #174488, opened 2025-08-27
Check GTK calls are done on the same thread.

### LINKED ISSUE #173660: Add Linux tests to confirm GTK calls are on GTK thread

robert-ancell:
Fixes https://github.com/flutter/flutter/issues/173660

### DIFF
--- a/engine/src/flutter/shell/platform/linux/testing/mock_gtk.cc
+++ b/engine/src/flutter/shell/platform/linux/testing/mock_gtk.cc
@@ -24,9 +24,15 @@ static void fl_mock_keymap_class_init(FlMockKeymapClass* klass) {
 
 static void fl_mock_keymap_init(FlMockKeymap* self) {}
 
+// Override GdkKeymap
+GType gdk_keymap_get_type() {
+  return fl_mock_keymap_get_type();
+}
+
 static MockGtk* mock = nullptr;
 
 MockGtk::MockGtk() {
+  thread = g_thread_self();
   mock = this;
 }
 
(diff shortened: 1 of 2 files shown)


engine/src/flutter/shell/platform/linux/testing/mock_gtk.cc
```diff
@@ -36,116 +42,232 @@ MockGtk::~MockGtk() {
   }
 }
 
+// Check we are running on the GTK thread.
+static void check_thread() {
+  if (mock != nullptr) {
```
mattkae, line 47: Why would `mock` be null when calling this? And does that invalidate any of the tests?

mattkae (approved):
One question, but approving it if you have good reason!

engine/src/flutter/shell/platform/linux/testing/mock_gtk.cc
robert-ancell: If you don't make a MockGtk object at the start of your test then this will be null - this is the case where you don't want to mock any of the GTK calls. Feels a bit clunky but couldn't think of a simpler solution.

flar:
This appears to be causing build errors due to conflicting definitions of `gdk_cairo_draw_from_gl`. Reverting.

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8705343407724739553/+/u/build_ci_host_release_benchmarks_flutter_build_dart:copy_dart_sdk_flutter_display_list:display_list_benchmarks_flutter_display_list:display_list_builder_benchmarks_flutter_display_list:display_list_region_benchmarks_flutter_display_list:display_list_transform_benchmarks_flutter_fml:fml_benchmarks_flutter_impeller_geometry:geometry_benchmarks_flutter_lib_ui:ui_benchmarks_flutter_shell_common:shell_benchmarks_flutter_shell_testing_flutter_txt:txt_benchmarks_flutter_tools_path_ops_flutter_build_archives:flutter_patched_sdk_flutter:unittests/stdout

flar:
Revert has landed. Feel free to rebase and reland now.

PR #174491, opened 2025-08-27
fix: findChildIndexCallback to take seperators into account for seperated named constructor in ListView and SliverList

### LINKED ISSUE #174261: ListView.separated findChildIndexCallback doesn't take separators into account
### Steps to reproduce

[`ListView.separated`](https://api.flutter.dev/flutter/widgets/ListView/ListView.separated.html) (as well as [`SliverList.separated`](https://api.flutter.dev/flutter/widgets/SliverList/SliverList.separated.html) accept [`findChildIndexCallback`](https://api.flutter.dev/flutter/widgets/SliverChildBuilderDelegate/findChildIndexCallback.html) parameter which can be used to preserve the state of its children when their position change. 

When this callback is implemented to return the actual item index (based on what's built by `itemBuilder`), the states aren't correctly preserved.

### Expected results

Indices returned by `findChildIndexCallback` should be treated as item indices to match indices passed to `itemBuilder`.

### Actual results

Indices returned by `findChildIndexCallback` are passed directly to `SliverChildBuilderDelegate` without taking the added separators into account. Currently, to achieve the desired behavior, indices returned by the callback need to be multiplied by 2.

### Code sample

In the following example ListView's items are reordered but their states aren't preserved: https://dartpad.dev/?id=13e666900223cef2b9d89ea8f925bad0

### Flutter Doctor output

```
flutter doctor -v
[✓] Flutter (Channel stable, 3.35.1, on macOS 15.6.1 24G90 darwin-arm64, locale en-PL) [653ms]
    • Flutter version 3.35.1 on channel stable at /Users/jakub/fvm/versions/3.35.1
    • Framework revision 20f8274939 (8 days ago), 2025-08-14 10:53:09 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0

[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0-rc5) [2.3s]
    • Android SDK at /Users/jakub/Library/Android/sdk
    • Emulator version 35.3.11.0 (build_id 12836668) (CL:N/A)
    • Platform android-36, build-tools 36.0.0-rc5
    • Java binary at: /opt/homebrew/opt/openjdk@17/bin/java
    • Java version OpenJDK Runtime Environment Homebrew (build 17.0.15+0)

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [1,470ms]
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [147ms]

[!] Android Studio (not installed) [146ms]
      (or visit https://flutter.dev/to/macos-android-setup for detailed instructions).

[✓] IntelliJ IDEA Ultimate Edition (version 2025.2) [145ms]
    • IntelliJ at /Users/jakub/Applications/IntelliJ IDEA Ultimate.app
    • Flutter plugin version 87.1
    • Dart plugin version 252.24322.5

[✓] Connected device (3 available) [6.7s]
    • SM A525F (mobile) • R58R60ABD8R • android-arm64  • Android 14 (API 34)
    • macOS (desktop)   • macos       • darwin-arm64   • macOS 15.6.1 24G90 darwin-arm64
    • Chrome (web)      • chrome      • web-javascript • Google Chrome 139.0.7258.139
    ! Error: Browsing on the local area network for iPhone (Marcel). Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for iPhone (Jakob). Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.

[✓] Network resources [508ms]

```

rkishan516:
FindChildIndexCallback to take seperators into account for seperated named constructor in ListView and SliverList
fixes: #174261

## Migration guide

https://github.com/flutter/website/pull/12636

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
+++ b/packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
@@ -952,4 +952,26 @@ transforms:
       - kind: 'rename'
         newName: 'markForComplete'
 
+  # Changes made in https://github.com/flutter/flutter/pull/174491
+  - title: "Migrate to 'findItemIndexCallback'"
+    date: 2025-10-16
+    element:
+      uris: [ 'widgets.dart', 'material.dart', 'cupertino.dart' ]
+      constructor: 'separated'
+      inClass: 'ListView'
+    changes:
+      - kind: 'renameParameter'
+        oldName: 'findChildIndexCallback'
+        newName: 'findItemIndexCallback'
+  - title: "Migrate to 'findItemIndexCallback'"
+    date: 2025-10-16
+    element:
+      uris: [ 'widgets.dart', 'material.dart', 'cupertino.dart' ]
+      constructor: 'separated'
+      inClass: 'SliverList'
+    changes:
+      - kind: 'renameParameter'
+        oldName: 'findChildIndexCallback'
+        newName: 'findItemIndexCallback'
+
 # Before adding a new fix: read instructions at the top of this file.
--- a/packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
+++ b/packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
@@ -952,26 +952,4 @@ transforms:
       - kind: 'rename'
         newName: 'markForComplete'
 
-  # Changes made in https://github.com/flutter/flutter/pull/174491
-  - title: "Migrate to 'findItemIndexCallback'"
-    date: 2025-10-16
-    element:
-      uris: [ 'widgets.dart', 'material.dart', 'cupertino.dart' ]
-      constructor: 'separated'
-      inClass: 'ListView'
-    changes:
-      - kind: 'renameParameter'
-        oldName: 'findChildIndexCallback'
-        newName: 'findItemIndexCallback'
-  - title: "Migrate to 'findItemIndexCallback'"
-    date: 2025-10-16
-    element:
-      uris: [ 'widgets.dart', 'material.dart', 'cupertino.dart' ]
-      constructor: 'separated'
-      inClass: 'SliverList'
-    changes:
-      - kind: 'renameParameter'
-        oldName: 'findChildIndexCallback'
-        newName: 'findItemIndexCallback'
-
 # Before adding a new fix: read instructions at the top of this file.
--- a/packages/flutter/lib/src/widgets/scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/scroll_view.dart
@@ -1458,6 +1458,8 @@ class ListView extends BoxScrollView {
       'Use findItemIndexCallback instead. '
       'findChildIndexCallback returns child indices (which include separators), '
       'while findItemIndexCallback returns item indices (which do not). '
+      'If you were multiplying results by 2 to account for separators, '
+      'you can remove that workaround when migrating to findItemIndexCallback. '
       'This feature was deprecated after v3.37.0-1.0.pre.',
     )
     ChildIndexGetter? findChildIndexCallback,
--- a/packages/flutter/lib/src/widgets/scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/scroll_view.dart
@@ -1419,6 +1419,20 @@ class ListView extends BoxScrollView {
   ///
   /// {@macro flutter.widgets.PageView.findChildIndexCallback}
   ///
+  /// {@template flutter.widgets.ListView.separated.findItemIndexCallback}
+  /// The [findItemIndexCallback] returns item indices (excluding separators),
+  /// unlike the deprecated [findChildIndexCallback] which returns child indices
+  /// (including both items and separators).
+  ///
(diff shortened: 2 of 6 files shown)


victorsanni (commented):
While our tests pass, this is probably going to break anyone currently using `ListView.separated.findChildIndexCallback` because they are probably multiplying that result by 2 already to workaround this.

I think we should deprecate this property but only on the separated constructor. And replace with a `ListView.separated.findItemIndexCallback`. Is this possible? What do you think @rkishan516?

rkishan516:
> I think we should deprecate this property but only on the separated constructor. And replace with a `ListView.separated.findItemIndexCallback`. Is this possible? What do you think @rkishan516?

Yes, it is possible. I was just worried about `seperated` constructor being inconsistent with others.

packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
```diff
@@ -952,4 +952,26 @@ transforms:
       - kind: 'rename'
         newName: 'markForComplete'
 
+  # Changes made in https://github.com/flutter/flutter/pull/174491
+  - title: "Migrate to 'findItemIndexCallback'"
```
victorsanni: Can we factor in the calculation when dart-fixing to the new property? To preserve the current behavior.

rkishan516: I haven't seen any data driven fix like this. Can we do this level of code transformation with data driven fixes? Any example which I can follow?

victorsanni: I agree, I don't think this is doable.

Maybe all we can do here is alert users in the deprecation messages that they no longer need to employ a workaround to multiply the results of `findChildIndexCallback` by 2 to account for separators.

What do you think?

rkishan516: Yup, makes sense. Let me update the deprecation message then.

Piinks (commented):
Is it possible we might want to keep the original callback since it is used the keep track of children when they are re-ordered? There is an open pending PR that add a separated constructor to the reorderable list classes.

packages/flutter/lib/src/widgets/scroll_view.dart
```diff
@@ -1478,7 +1487,12 @@ class ListView extends BoxScrollView {
            }
            return separatorBuilder(context, itemIndex);
          },
-         findChildIndexCallback: findChildIndexCallback,
+         findChildIndexCallback: findItemIndexCallback != null
```
chunhtai, line 1509: we should add an assert to ensure only one of them is provided

packages/flutter/lib/fix_data/fix_widgets/fix_widgets.yaml
chunhtai: we shouldn't add this data driven fix if we can't migrate the calculation, this will break customer's app

chunhtai (commented):
change lgtm, but we need a migration guide and follows https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes

chunhtai:
I see no good reason to preserve the original callback. since we don't provide a way to build different separator based on index, and you can't swap a separator's location with other item.

packages/flutter/lib/src/widgets/scroll_view.dart
rkishan516: Sure, I will add.

packages/flutter/lib/src/widgets/scroll_view.dart
```diff
@@ -1454,7 +1454,16 @@ class ListView extends BoxScrollView {
...
+    @Deprecated(
+      'Use findItemIndexCallback instead. '
+      'findChildIndexCallback returns child indices (which include separators), '
+      'while findItemIndexCallback returns item indices (which do not). '
+      'If you were multiplying results by 2 to account for separators, '
+      'you can remove that workaround when migrating to findItemIndexCallback. '
+      'This feature was deprecated after v3.37.0-1.0.pre.',
+    )
     ChildIndexGetter? findChildIndexCallback,
+    ChildIndexGetter? findItemIndexCallback,
```
chunhtai, line 1480: Can you add some doc around findItemIndexCallback in the constructor api doc and how/why it is different from findChildIndexCallback of the other constructor

chunhtai: also consider using doc template `{@template` to also put the doc in sliverList's constructor

chunhtai (approved):
LGTM

packages/flutter/lib/src/widgets/scroll_view.dart
rkishan516: Sure, I will update.

victorsanni (approved):

PR #174580, opened 2025-08-27
[ Tool ] Serve DevTools from DDS, remove ResidentDevToolsHandler

bkonyi:
This PR contains the initial work to prepare for the removal of instances of DevTools being served manually by the flutter_tool, instead relying on DDS to serve DevTools. This will be consistent with how the standalone Dart VM serves DevTools, tying the DevTools lifecycle to a live DDS instance. This will allow for the removal of much of the logic needed to properly manage the lifecycle of the DevTools server in a future PR. Also, by serving DevTools from DDS, users will no longer need to forward a secondary port in remote workflows as DevTools will be available on the DDS port.

There's two remaining circumstances that will prevent us from removing `DevtoolsRunner` completely:

  - The daemon's `devtools.serve` endpoint
  - `flutter drive`'s `--profile-memory` flag used for recording memory profiles


bkonyi:
Friendly bump. Can I get a review on this please? :-)

helin24 (approved):
I don't have a deep understanding of this code and implications, but I did read over the files and I tested the internal mobile path.

PR #174659, opened 2025-08-28
[web] Cleanup opportunities post renderer unification

mdebbar:
- `injectClientICUIfNeeded` as a method on `SkParagraphBuilder` for convenience.
- Move `CanvasKitVariant` outside of `canvaskit/` because it's needed outside of that folder.
- Make `resourceCacheMaxBytes` a method on all renderers, not only CanvasKit.
- Some optimizations in `OcclusionMap`.
- Remove unused CanvasKit test utils.


engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -20,6 +20,12 @@ class SkwasmRenderer extends Renderer {
 
   SkwasmPathConstructors pathConstructors = SkwasmPathConstructors();
 
+  @override
+  set resourceCacheMaxBytes(int bytes) {
+    // ?
+    // rasterizer.setResourceCacheMaxBytes(bytes);
+  }
```
mdebbar: @eyebrowsoffire @harryterkelsen should we do this in Skwasm? I don't see any usages of this in the framework, so I'm not sure who uses it.

harryterkelsen: It is documented in the framework as something that can be called here: 

https://github.com/flutter/flutter/blob/fdab2850235b4a3a5ef1447f002a34902329a596/packages/flutter/lib/src/services/system_channels.dart#L417

I added code to enable this in Skwasm as part of the unification, but forgot to plumb it all the way through. I think for consistency, we should add this here. I also don't know who actually uses this but it's even enabled in the mobile engine.

engine/src/flutter/lib/web_ui/lib/src/engine/renderer.dart
```diff
@@ -59,6 +59,9 @@ abstract class Renderer {
   // Listens for view disposal events from the view manager.
   late StreamSubscription<int> _onViewDisposedListener;
 
+  /// Set the maximum number of bytes that can be held in the GPU resource cache.
+  set resourceCacheMaxBytes(int bytes);
```
harryterkelsen: If skwasm also has `resourceCacheMaxBytes` then the implementation should be moved into this class since it will be the same for both renderers.

engine/src/flutter/lib/web_ui/lib/src/engine/occlusion_map.dart
```diff
@@ -83,7 +104,17 @@ class OcclusionMapBranch implements OcclusionMapNode {
 class OcclusionMap {
   OcclusionMapNode root = OcclusionMapEmpty();
 
-  void addRect(ui.Rect rect) => root = root.insert(rect);
+  void addRect(ui.Rect rect) {
+    if (rect.width <= 0 || rect.height <= 0) {
```
harryterkelsen: nit: replace this condition and the one below with `rect.isEmpty`

engine/src/flutter/lib/web_ui/lib/src/engine/occlusion_map.dart
```diff
@@ -55,17 +65,28 @@ class OcclusionMapBranch implements OcclusionMapNode {
 
   @override
   OcclusionMapNode insert(ui.Rect other) {
+    if (rectContainsOther(other, boundingBox)) {
```
harryterkelsen, line 72: Add a comment saying something like "If `other` contains the entire bounding box of the left and right branch, then replace this branch with a leaf that just contains `other`". I think this line is a bit too clever to go without a comment.

harryterkelsen (changes requested):
I think the Skwasm renderer should support `setResourceCacheMaxBytes`.

mdebbar:
@harryterkelsen I addressed all comments. I won't rebase this PR until https://github.com/flutter/flutter/pull/174588 relands.

eyebrowsoffire (approved):
LGTM

mdebbar:
Waiting for https://github.com/flutter/flutter/pull/175392/ to land.

harryterkelsen (approved):
LGTM

PR #168866, opened 2025-05-14
CupertinoSliverNavigationBar respects accessibility text scaling

victorsanni:
Dupe of https://github.com/flutter/flutter/pull/167313 (closed because of issues with Google testing)

Fixes [Cupertino nav bars should respect accessibility text scaling](https://github.com/flutter/flutter/issues/167248)
Fixes [Title jumps during a CupertinoNavigationBar -> CupertinoNavigationBar hero transition](https://github.com/flutter/flutter/issues/90589)

## Before

| Regular | Scaled | 
| --- | --- |
| Image | Image |

## After

| Regular | Scaled | 
| --- | --- |
| flutter nav bar unscaled | flutter after 2 textscale |

## Native

| Regular | Scaled | 
| --- | --- |
| Image | Image |

PR #174673, opened 2025-08-28
Nav bar static components respect ambient MediaQueryData

### LINKED ISSUE #174642: CupertinoSlidingSegmentedControl in CupertinoNavigationBar renders incorrectly in dark brightness (starting 3.35.0)
### Steps to reproduce

1. Render a CupertinoSlidingSegmentedControl in a CupertinoNavigationBar and set the device brightness to dark.
2. Leave all non-essential params unspecified

The screenshots below are from iOS 18.5.

This appears starting with 3.35.0 (3.32.8 doesn't exhibit the issue).

### Expected results

This is rendered on 3.32.8

Image

https://github.com/user-attachments/assets/716f2379-6533-425f-a008-3ea92ad7445b

### Actual results

This is rendered on 3.35.0. The thumb renders white where it previously rendered Color(0xFF636366) (_kThumbColor).

The segment background colors also appear to have changed.

Image

https://github.com/user-attachments/assets/200ab665-8753-42aa-9f7a-ec214187dc9e

### Code sample

```dart

void main() {
  runApp(const SlidingControlDemo());
}

class SlidingControlDemo extends StatelessWidget {
  const SlidingControlDemo({super.key});

  @override
  Widget build(BuildContext context) => const CupertinoApp(
    home: CupertinoPageScaffold(
      navigationBar: CupertinoNavigationBar(middle: SlidingControl()),
      child: Center(child: SlidingControl()),
    ),
  );
}

class SlidingControl extends StatefulWidget {
  const SlidingControl({super.key});

  @override
  State<SlidingControl> createState() => _SlidingControlState();
}

class _SlidingControlState extends State<SlidingControl> {
  var _value = 1;

  void _onValueChange(int? value) => value != null
      ? setState(() {
          _value = value;
        })
      : null;

  @override
  Widget build(BuildContext context) => CupertinoSlidingSegmentedControl(
    children: const {1: Text('1'), 2: Text('2')},
    groupValue: _value,
    onValueChanged: _onValueChange,
  );
}
```

</details>

### Screenshots or Video

<details open>

See Expected and Actual Results above

</details>

### Logs

n/a

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.0, on macOS 15.5 24F74 darwin-arm64, locale en-US) [480ms]
    • Flutter version 3.35.0 on channel stable at /Users/ryanhanks/fvm/versions/3.35.0
    • Framework revision b896255557 (2 weeks ago), 2025-08-13 17:14:08 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-lldb-debugging

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [1,278ms]
    • Build 16F6
    • CocoaPods version 1.16.2
```

</details>

victorsanni:
Fixes [CupertinoSlidingSegmentedControl in CupertinoNavigationBar renders incorrectly in dark brightness (starting 3.35.0)](https://github.com/flutter/flutter/issues/174642)
Fixes [[Cupertino] MediaQuery.of(context) returns Size(0,0) inside CupertinoNavigationBar trailing since Flutter 3.35](https://github.com/flutter/flutter/issues/174771#top)

Bug was introduced in [CupertinoSliverNavigationBar respects accessibility text scaling](https://github.com/flutter/flutter/pull/168866)

See https://github.com/flutter/flutter/issues/174642#issuecomment-3235042874


packages/flutter/test/cupertino/nav_bar_test.dart
```diff
@@ -451,6 +451,34 @@ void main() {
...
+          ),
+          child: CupertinoNavigationBar(
+            leading: CupertinoButton(
+              onPressed: () {},
+              child: const _ExpectCustomMediaQuery(value: value, index: 0x000001),
+            ),
+            middle: const _ExpectCustomMediaQuery(value: value, index: 0x000100),
+            trailing: CupertinoButton(
+              onPressed: () {},
+              child: const _ExpectCustomMediaQuery(value: value, index: 0x010000),
```
dkwingsmt: While the current way is acceptable, I'd prefer if you can explore a better style that writes the `expect`s in the test body instead of in the class, which is hard to find.

Can you make `_ExpectCustomMediaQuery` accept a void callback instead and define the `expect`s in the callback?

dkwingsmt (approved):
LGTM with a minor suggestion

packages/flutter/test/cupertino/nav_bar_test.dart
dkwingsmt: Also probably add a "regression test" comment explaining why these expectations make sense

victorsanni: > While the current way is acceptable, I'd prefer if you can explore a better style that writes the expects in the test body instead of in the class, which is hard to find.

Ah. I implemented it that way because it was consistent with the `_ExpectStyles` class in the same file. But then the `_ExpectStyles` class is reused across multiple tests. I can explore a solution contained entirely within the test itself.

packages/flutter/test/cupertino/nav_bar_test.dart
```diff
@@ -451,6 +451,52 @@ void main() {
     expect(count, 0x010101);
   });
 
+  testWidgets('Nav bar static components respect MediaQueryData', (WidgetTester tester) async {
+    // This is a regression test for https://github.com/flutter/flutter/issues/174642
+    const double value = 10.0;
+
+    void expectCustomMediaQueryData(BuildContext context) {
+      expect(MediaQuery.platformBrightnessOf(context), Brightness.dark);
```
dkwingsmt, line 459: This looks great, but is it ok if it doesn't check the `value` anymore? (Which I'm ok with since the original issue wasn't about custom values)

victorsanni: The problem with that is a future change might copy over only the mediaQueryData brightness leaving the other properties out. And we want to prevent against that regression. But MediaQueryData has many properties, so I'm thinking of testing with a subset of those properties numerous enough to protect against regression.

What do you think is the optimal approach here?

dkwingsmt: Yeah I think this is sufficient. Just a few is enough to verify our approach here, and thoroughly testing `copyWith` should be done in `MediaQueryData`'s tests.

MitchellGoodwin (approved):
LGTM as well

PR #174679, opened 2025-08-29
RenderTable adoptChild before calling setFlatChildren.

### LINKED ISSUE #174133: OverlayPortal in Table/TableRow crashes when OverlayPortalController.show is called before the OverlayPortalController is assigned to the OverlayPortal
I have a Flutter project that runs fine with Flutter SDK 3.32.X, but crashes at a specific point after upgrading to Flutter 3.35.1. I am not allowed to disclose this project and therefore derived a smaller version from it:

[overlay_portal_assertion.zip](https://github.com/user-attachments/files/21898634/overlay_portal_assertion.zip)

The output from the debug console is contained in the "debug console.txt" within the folder of the project.

To cause the exception:

1. Execute the application in debug mode,
2. Click the Button (with label "Login") and
3. Click the button on the right (with label "Tools"). Then the exception is triggered.

Flutter doctor output when having and using Flutter 3.35.1:
```
[√] Flutter (Channel stable, 3.35.1, on Microsoft Windows [Version 10.0.26100.4946], locale de-DE) [230ms]
    • Flutter version 3.35.1 on channel stable at C:\Progs\Flutter\flutter
    • Framework revision 20f8274939 (6 days ago), 2025-08-14 10:53:09 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0

[√] Windows Version (11 Pro 64-bit, 24H2, 2009) [2,1s]

[√] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [1.645ms]
    • Android SDK at C:\Progs\Android\SDK
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java

[√] Chrome - develop for the web [70ms]

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.14.12) [69ms]
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community

[√] Android Studio (version 2025.1.2) [13ms]

[√] VS Code, 64-bit edition (version 1.103.1) [11ms]
    • VS Code at C:\Program Files\Microsoft VS Code
    • Flutter extension version 3.116.0

[√] Connected device (3 available) [174ms]
    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [Version 10.0.26100.4946]
    • Chrome (web)      • chrome  • web-javascript • Google Chrome 139.0.7258.128
    • Edge (web)        • edge    • web-javascript • Microsoft Edge 137.0.3296.93

[√] Network resources [267ms]

• No issues found!
```

For running the app with Flutter 3.32.8:

1. Go to the folder with the Flutter SDK,
2. Switch the working copy to branch "3.32.8",
3. Run "flutter doctor -v" to let adjust to the switched branch,
4. Change the version for the Dart sdk under "environment:" "sdk:" to "^3.8.1" in file pubspec.yaml of the project,
5. Run "flutter pub get" to update the packages.
6. Execute the app in debug mode and execute the same steps as above.

To revert to the latest SDK:

1. Execute "flutter channel stable",
2. Execute "flutter doctor -v"
3. Change the version for the Dart SDK back to 3.9.0 in file pubspec.yaml of the project,
4. Run "flutter pub get" to update the packages.

yiiim:
Fixes: #174133

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/rendering/table.dart
+++ b/packages/flutter/lib/src/rendering/table.dart
@@ -841,7 +841,18 @@ class RenderTable extends RenderBox {
         if (cells[xyNew] != null &&
             (x >= _columns || y >= _rows || _children[xyOld] != cells[xyNew])) {
           if (!lostChildren.remove(cells[xyNew])) {
-            adoptChild(cells[xyNew]!);
+            assert(() {
+              if (!_shelteredChildren.remove(cells[xyNew])) {
+                throw FlutterError.fromParts(<DiagnosticsNode>[
+                  ErrorSummary('The newly added child did not call shelterChild.'),
+                  ErrorHint(
+                    'If you are not writing your own RenderTable subclass, then this is not '
+                    'your fault. Contact support: https://github.com/flutter/flutter/issues/new?template=02_bug.yml',
+                  ),
+                ]);
+              }
+              return true;
+            }());
           }
         }
       }
@@ -912,6 +923,23 @@ class RenderTable extends RenderBox {
     if (value != null) {
       adoptChild(value);
     }
+  }
+
+  final List<RenderBox> _shelteredChildren = <RenderBox>[];
+
+  /// Before using [setFlatChildren], you need to shelter the new child with this method
+  /// to attach it to the tree. Allowing children that are not attached to the tree to
+  /// participate in building is dangerous.
+  ///
+  /// See this issue for reference: https://github.com/flutter/flutter/issues/174133.
+  void shelterChild(RenderBox child) {
+    assert(child != this);
+    assert(!_children.contains(child));
+    assert(() {
+      _shelteredChildren.add(child);
+      return true;
+    }());
+    adoptChild(child);
   }
 
   @override
@@ -1306,6 +1334,7 @@ class RenderTable extends RenderBox {
 
   @override
   void performLayout() {
+    assert(_shelteredChildren.isEmpty);
     final BoxConstraints constraints = this.constraints;
     final int rows = this.rows;
     final int columns = this.columns;
@@ -1452,6 +1481,7 @@ class RenderTable extends RenderBox {
 
   @override
   void paint(PaintingContext context, Offset offset) {
+    assert(_shelteredChildren.isEmpty);
     assert(_children.length == rows * columns);
     if (rows * columns == 0) {
       if (border != null) {
--- a/packages/flutter/lib/src/rendering/table.dart
+++ b/packages/flutter/lib/src/rendering/table.dart
@@ -394,6 +394,10 @@ class RenderTable extends RenderBox {
        _textBaseline = textBaseline,
        _defaultVerticalAlignment = defaultVerticalAlignment,
        _configuration = configuration {
+    assert(() {
+      _debugShelteredChildren = <RenderBox>[];
+      return true;
+    }());
     _children = <RenderBox?>[]..length = _columns * _rows;
     this.rowDecorations = rowDecorations; // must use setter to initialize box painters array
     children?.forEach(addRow);
@@ -842,7 +846,7 @@ class RenderTable extends RenderBox {
             (x >= _columns || y >= _rows || _children[xyOld] != cells[xyNew])) {
           if (!lostChildren.remove(cells[xyNew])) {
             assert(() {
-              if (!_shelteredChildren.remove(cells[xyNew])) {
+              if (!_debugShelteredChildren.remove(cells[xyNew])) {
                 throw FlutterError.fromParts(<DiagnosticsNode>[
                   ErrorSummary('The newly added child did not call shelterChild.'),
                   ErrorHint(
@@ -925,7 +929,7 @@ class RenderTable extends RenderBox {
     }
   }
 
-  final List<RenderBox> _shelteredChildren = <RenderBox>[];
+  late final List<RenderBox> _debugShelteredChildren;
 
   /// Before using [setFlatChildren], you need to shelter the new child with this method
   /// to attach it to the tree. Allowing children that are not attached to the tree to
@@ -935,8 +939,9 @@ class RenderTable extends RenderBox {
   void shelterChild(RenderBox child) {
     assert(child != this);
     assert(!_children.contains(child));
+    assert(!_debugShelteredChildren.contains(child));
     assert(() {
-      _shelteredChildren.add(child);
+      _debugShelteredChildren.add(child);
       return true;
     }());
     adoptChild(child);
@@ -1334,7 +1339,7 @@ class RenderTable extends RenderBox {
 
   @override
   void performLayout() {
-    assert(_shelteredChildren.isEmpty);
+    assert(_debugShelteredChildren.isEmpty);
     final BoxConstraints constraints = this.constraints;
     final int rows = this.rows;
     final int columns = this.columns;
@@ -1481,7 +1486,7 @@ class RenderTable extends RenderBox {
 
   @override
   void paint(PaintingContext context, Offset offset) {
-    assert(_shelteredChildren.isEmpty);
+    assert(_debugShelteredChildren.isEmpty);
     assert(_children.length == rows * columns);
     if (rows * columns == 0) {
       if (border != null) {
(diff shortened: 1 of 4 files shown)


filiph:
When this lands, I'd vote for cherry-picking this into stable. I'm hitting the issue (https://github.com/flutter/flutter/issues/174133) and there's no workaround that I know of.

packages/flutter/lib/src/rendering/table.dart
```diff
@@ -914,6 +925,23 @@ class RenderTable extends RenderBox {
     }
   }
 
+  final List<RenderBox> _shelteredChildren = <RenderBox>[];
```
LongCatIsLooong: is this debug only? If so, consider adding a `debug` prefix and nil initialize it when not in debug mode (the dart compiler can't eliminate the list initializer even if the field is not used).

packages/flutter/lib/src/rendering/table.dart
```diff
@@ -914,6 +925,23 @@ class RenderTable extends RenderBox {
...
+
+  /// Before using [setFlatChildren], you need to shelter the new child with this method
+  /// to attach it to the tree. Allowing children that are not attached to the tree to
+  /// participate in building is dangerous.
+  ///
+  /// See this issue for reference: https://github.com/flutter/flutter/issues/174133.
+  void shelterChild(RenderBox child) {
```
LongCatIsLooong, line 939: Does this have to be exposed as a public method? This is just `adoptChild` (plus asserts) right?

packages/flutter/lib/src/widgets/table.dart
```diff
@@ -318,6 +318,8 @@ class _TableElement extends RenderObjectElement {
     // in [_updateRenderObjectChildren].
     if (!_doingMountOrUpdate) {
       renderObject.setChild(slot.column, slot.row, child);
+    } else {
+      renderObject.shelterChild(child);
```
LongCatIsLooong, line 322: So this calls `renderObject.adopt(child)`. But at the end of the mount method, wouldn't `renderObject.adopt(child)` be called again by `setFlatChildren`? That's going to cause some assert to fire I think (the same child is adopted a second time).

yiiim: setFlatChildren will not adopt the child again; the original adopt child call has already been replaced by an assertion in this PR.

packages/flutter/lib/src/rendering/table.dart
yiiim: Yes, it is just `adoptChild` with an added assertion. It is made public because it needs to be called by `_TableElement`.

yiiim: Assertions can ensure that this public method is called correctly.

LongCatIsLooong:
(May not be able to review until Friday. Sorry for the delay)

LongCatIsLooong:
I looked at TableElement's implementation and my hunch is that we should remove `assert(child.owner == owner)` from `redepthChild`. AFAIK the notion of "depth" has little to do with who the pipeline owners is, so I'm not sure if that's the perfect place to put that assert, and clearly you can change the depth without the render object being attached like in the Table case. 

So I'm leaning towards removing the assert from `redepthChild` (or maybe move it to `adoptChild` instead) since it's not clear to me why the invariant has to be enforced in `redepthChild`. The git [history](https://github.com/flutter/flutter/blame/e074af8099b1a3ad5b909258a4ab532b73c96045/packages/flutter/lib/src/rendering/node.dart#L54-L59) shows that the assert is there from day one. @Hixie WDYT?

Hixie:
At the time there was only ever one owner but if that's changed then yeah, removing the assert is fine.

Hixie:
(Or just in general, if the assert isn't asserting something that's accurate at that place in the code, it should be removed. Sorry, I'm not quite sure I understand the issue here so it's hard for me to usefully comment on it off the top of my head.)

yiiim:
Even if we remove the assertions, can we still keep this PR? Ideally, during the mount process, the RenderObject should already be in the tree.

LongCatIsLooong:
> Even if we remove the assertions, can we still keep this PR? Ideally, during the mount process, the RenderObject should already be in the tree.

Hmm I think that's true, after `inflateChild` i would expect the child's render object to know who the root render object is (but today if the child is in a Table that won't be true). That said, I'm not sure about the new public method though, since it's just adoptChild, would it be ok if we just call adoptChild instead?

> At the time there was only ever one owner but if that's changed then yeah, removing the assert is fine.

I think even with multiview each tree has (each view would have its own root node and pipeline owner if I recall) only one PipelineOwner, so that's still true (I'll update my comment in the issue). OverlayPortal's RenderObject is the only thing in the framework that has two parent render objects, and in this case, RenderTheater and RenderTable. And when RenderTheater tries to adopt the OverlayPortal, the implementation calls `RenderTable.redepthChild(overlayPortal)` to make sure the depth of the overlayPortal is greater than that of both the theater and the table. So the assert fired because the OverlayPortal's render object and the RenderTheater is attached but RenderTable's child is not.


I think what I'm not sure is what the assert is trying to verify? If it was `assert(child.parent == this)` then I get it since the depth depends on the tree structure and you want to make sure you can only call `redpeth` on your child.

yiiim:
> Hmm I think that's true, after `inflateChild` i would expect the child's render object to know who the root render object is (but today if the child is in a Table that won't be true). That said, I'm not sure about the new public method though, since it's just adoptChild, would it be ok if we just call adoptChild instead?

We cannot call adoptChild directly because it is marked as protected.

packages/flutter/lib/src/rendering/table.dart
```diff
@@ -914,6 +929,24 @@ class RenderTable extends RenderBox {
     }
   }
 
+  late final List<RenderBox> _debugShelteredChildren;
+
+  /// Before using [setFlatChildren], you need to shelter the new child with this method
```
LongCatIsLooong, line 934: API docs must begin with a simple description of what this method does.

packages/flutter/lib/src/rendering/table.dart
```diff
@@ -914,6 +929,24 @@ class RenderTable extends RenderBox {
...
+  /// to attach it to the tree. Allowing children that are not attached to the tree to
+  /// participate in building is dangerous.
+  ///
+  /// See this issue for reference: https://github.com/flutter/flutter/issues/174133.
```
LongCatIsLooong, line 938: I would make this a non-public (`//`) comment

packages/flutter/lib/src/rendering/table.dart
```diff
@@ -1306,6 +1339,7 @@ class RenderTable extends RenderBox {
 
   @override
   void performLayout() {
+    assert(_debugShelteredChildren.isEmpty);
```
LongCatIsLooong, line 1342: This check should probably be moved to whenever the child model finishes updating (at the end of setFlatChildren maybe), since we want to throw as early as possible. Same for the assert in paint.

LongCatIsLooong (approved):
LGTM with nits. 

The new public method is a bit weird since `dropChild` is done by `setFlatChildren` but for adopting a child you'll now have to call a different method. So this can be a bit tricky to get right. But even without `shelterChild`, the TableElement implementation already has to keep track of whether `insertRenderObject` is called from mount/update to determine whether to call `setFlatChildren` or do nothing so this already feels very tightly coupled / you really need to know what's going on in RenderTable to use it to implement a custom Table widget. 

My understanding is that `setFlatChildren` exists for performance purposes (to reduce the number of times we have to traverse the child list). Hopefully there's a way to restructure the child model so the element doesn't have to call `setFlatChildren`

packages/flutter/lib/src/rendering/table.dart
victorsanni: +1 here. A short one-sentence description to quickly grasp the purpose of the method would be nice.

packages/flutter/lib/src/rendering/table.dart
victorsanni: +1 here as well. And the preceding newline.

victorsanni (commented):
Hi @yiiim, this PR is ready to go once final comments are addressed.

yiiim:
> Hi @yiiim, this PR is ready to go once final comments are addressed.

Sorry for the delay. I think @LongCatIsLooong  makes a valid point. `shelterChild` and `setFlatChildren` are tightly coupled.. Since `shelterChild` would be a public method of a public class, it would be difficult to remove once this PR is merged. We might want to consider other approaches. Sorry for not resolving this issue.

PR #174709, opened 2025-08-29
Make sure that a Divider doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the Divider widget.


dkwingsmt (approved):
Checklist:

MitchellGoodwin (approved):
LGTM as well!

MitchellGoodwin:
@ahmedsameha1 could you resolve the conflicts? Then this should be clear to merge.

Issue #141899, opened 2024-01-19
Add test that covers Flutter.groovy ABI code and/or migrate the code to kotlin 


reidbaker:
https://github.com/flutter/flutter/issues/141897 found that type errors are not caught unless the code is executed. Add a test covering the abi behavior or make flutter.groovy compile static or migrate the abi code to kotlin so that is statically checked.

reidbaker:
Code was migrated to kotlin and unit tests were added.

PR #174710, opened 2025-08-29
add tests for FlutterPluginConstants

AbdeMohlbi:
towards #141899

## Pre-launch Checklist

- [x] I signed the [CLA].


reidbaker (commented):
@gmackall What do you think about these? On one hand these feel like change detector tests on the other they are at a layer where even a change might break assumptions that developers care about.

gmackall:
> @gmackall What do you think about these? On one hand these feel like change detector tests on the other they are at a layer where even a change might break assumptions that developers care about.

I would prefer not to land tests for these constants, as I agree they are basically change detector tests. 

I agree that it's important that we are somehow testing what platforms we support, but I think the behavior that we really don't want to regress is that the build actually outputs the architectures we support (not that the constants don't change). So I think tests that instead verify the existence of the build artifacts for our supported platforms, and the nonexistence of build artifacts for unsupported platforms, would be a better approach. Which is, fwiw, basically what https://github.com/flutter/flutter/blob/0c258ce6c5d1b0bcaf69469527ca8b5b5b5603fe/dev/devicelab/bin/tasks/gradle_plugin_fat_apk_test.dart#L28 does I believe

mboetger:
From Android triage: thank you for taking a look at this.  From @gmackall's comment, we will not be adding this test.

PR #162062, opened 2025-01-23
Make `DropdownMenu` TextField reactive to label changes

ahmedrasar:
## Description

The current behavior of Flutter regarding `DropdownMenu` text field is very basic: `DropdownMenu` updates the text field once a selection has been made with its coordinating `label`, and nothing more.

If the selection's `label` changes, the text field will still show the old value, although the menu buttons have been updated.

This not only results in the need to write a lot of boilerplate code to match the label to the current selection, but it's also counterintuitive. The `DropdownMenu` should handle rematching by itself.

Issue #155660, touched on this, and a solution was introduced in #155757. It was later reverted due to it restricting some use cases. That issue & solution, however, only address `initialSelection`. If another selection was made and its corresponding `label` was changed, the text field would still show the old value, bringing us back to square one. 

This PR should not conflict with any previous behavior of `DropdownMenu`: any new value provided by `initialSelection` or via controller would not be overwritten by rematching. However, entries with the same value will be mapped to a single label (the first matched entry's label). 

## Related Issues

- Fixes #155660.

## Tests

Added 3 tests.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #155757, opened 2024-09-26
Fix DropdownMenu does not rematch initialSelection when entries have changed

bleroux:
## Description

This PR makes DropdownMenu rematching the initialSelection when the entries are updated.
If the new entries contains one entry whose value matches `initialSelection` this entry's label is used to initialize the inner text field, if no entries matches `initialSelection` the text field is emptied.

## Related Issue

Fixes [DropdownMenu.didUpdateWidget should re-match initialSelection when dropdownMenuEntries have changed](https://github.com/flutter/flutter/issues/155660).

## Tests

Adds 3 tests.

Issue #155660, opened 2024-09-25
DropdownMenu.didUpdateWidget should re-match initialSelection when dropdownMenuEntries have changed


pamtbaau:
### Use case

In my use case, the ` DropdownMenu.dropdownMenuEntries` are asynchronously fetched from the database and assigned after the Future has been resolved. The `initialSelection`, however, has already been set in the first build cycle.

The problem lays in that in `DropdownMenu.didUpdateWidget()`, where the `initialSelection` is not matched again against the updated menu entries. See https://github.com/flutter/flutter/blob/9d42f826d71b1b9beb2d8c07bcdad731ba78968d/packages/flutter/lib/src/material/dropdown_menu.dart#L552-L557

This will lead to an empty field not showing the matching `initialSelection`, but yet showing a correct list of items in the dropdown. Quite confusing...

Finding out the cause after debugging, I use the following, hopefully temporary, workaround:
```
DropdownMenu<String>(
  initialSelection: items.isEmpty ? null : _selectedItem ?? widget.item,
  ... etc
)
```

### Proposal

I would like the `initialSelection` to be matched again against the new value set for `dropdownMenuEntries` as is done when the intialSelection has changed. See https://github.com/flutter/flutter/blob/9d42f826d71b1b9beb2d8c07bcdad731ba78968d/packages/flutter/lib/src/material/dropdown_menu.dart#L561-L569

IMHO this should be provided by Flutter directly.

bleroux:
Using this code sample:

<details><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  List<DropdownMenuEntry<String>> entries = const [
    DropdownMenuEntry(value: "1", label: "one"),
    DropdownMenuEntry(value: "2", label: "two"),
    DropdownMenuEntry(value: "3", label: "three"),
    DropdownMenuEntry(value: "4", label: "four"),
  ];

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      debugShowCheckedModeBanner: false,
      home: Scaffold(
        body: Padding(
          padding: const EdgeInsets.all(32.0),
          child: Column(
            children: [
              DropdownMenu<String>(
                label: const Text("test"),
                initialSelection: '2.5',
[...]

PR #174757, opened 2025-08-30
Fix `DropdownMenu` entry selection with enableFilter

ahmedrasar:
## Issues
 - Fixes #174609

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -912,7 +912,9 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
                     text: entry.label,
                     selection: TextSelection.collapsed(offset: entry.label.length),
                   );
-                  _selectedEntryIndex = i;
+                  _selectedEntryIndex = widget.dropdownMenuEntries.indexWhere(
+                    (DropdownMenuEntry<T> entry) => entry.value == filteredEntries[i].value,
+                  );
                   currentHighlight = widget.enableSearch ? i : null;
                   widget.onSelected?.call(entry.value);
                   _enableFilter = false;
@@ -1005,7 +1007,9 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
           text: entry.label,
           selection: TextSelection.collapsed(offset: entry.label.length),
         );
-        _selectedEntryIndex = currentHighlight;
+        _selectedEntryIndex = widget.dropdownMenuEntries.indexWhere(
+          (DropdownMenuEntry<T> entry) => entry.value == filteredEntries[currentHighlight!].value,
+        );
         widget.onSelected?.call(entry.value);
       }
     } else {
--- a/packages/flutter/test/material/dropdown_menu_test.dart
+++ b/packages/flutter/test/material/dropdown_menu_test.dart
@@ -2599,6 +2599,64 @@ void main() {
     },
   );
 
+  testWidgets('Entry selection in filtered entries should persist after the filter is cleared', (
+    WidgetTester tester,
+  ) async {
+    final TextEditingController controller = TextEditingController();
+    addTearDown(controller.dispose);
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: StatefulBuilder(
+          builder: (BuildContext context, StateSetter setState) {
+            String getFirstItemLabel() => 'A1';
+
+            return Scaffold(
+              body: Column(
+                children: <Widget>[
+                  DropdownMenu<TestMenu>(
+                    requestFocusOnTap: true,
+                    enableFilter: true,
+                    controller: controller,
+                    dropdownMenuEntries: <DropdownMenuEntry<TestMenu>>[
+                      DropdownMenuEntry<TestMenu>(
+                        value: TestMenu.mainMenu0,
+                        // The use of a function is necessary to make the list mutable,
+                        // since immutable lists' selection does not rematch.
+                        label: getFirstItemLabel(),
+                      ),
+                      const DropdownMenuEntry<TestMenu>(value: TestMenu.mainMenu2, label: 'B1'),
+                      const DropdownMenuEntry<TestMenu>(value: TestMenu.mainMenu3, label: 'A2'),
+                    ],
+                  ),
+                ],
+              ),
+              floatingActionButton: FloatingActionButton(onPressed: () => setState(() {})),
+            );
+          },
+        ),
+      ),
+    );
+
+    // Open the menu.
+    await tester.tap(find.byType(DropdownMenu<TestMenu>));
+    await tester.pump();
+
+    // Filter the entries to only show the ones with 'A'.
+    await tester.enterText(find.byType(TextField), 'A');
+    await tester.pump();
+
+    // Select the second item in the filtered list.
+    await tester.tap(findMenuItemButton('A2'));
+    await tester.pump();
+
+    // Rebuild the widgets.
+    await tester.tap(find.byType(FloatingActionButton));
+    await tester.pump();
+
+    expect(controller.text, 'A2');
+  });
+
   testWidgets('The default text input field should not be focused on mobile platforms '
       'when it is tapped', (WidgetTester tester) async {
     final ThemeData themeData = ThemeData();


ahmedrasar:
I forgot to rebase with master first, my bad.

bleroux:
@ahmedrasar I filed https://github.com/flutter/flutter/pull/174778 which fixes the same issue. 
To improve readability and avoid subtle bugs related to storing the index, I opted for a solution that stores the selected value instead of the selected index. WDYT?

ahmedrasar:
@bleroux. I had gone through many iterations on this, and one of them was your suggestion.

It does indeed improve readability and unlike storing the index it doesn't loop the entries to get the value.

However it has two drawbacks. The first is that storing a `T` value could be expensive. The second, is that `T` itself could be nullable. So if an entry with null value was clicked, the widget would treat it as if there is no entry selected.

I comared every possible soultion I can think of [in this comment](https://github.com/flutter/flutter/pull/162062#discussion_r2081010517)

Your test is better than mine, though.

bleroux:
Thanks for the clarification.
Let's close https://github.com/flutter/flutter/pull/174778 in favor of this PR.

> Your test is better than mine, though.

Feel free to use it.

reidbaker (approved):

QuncCccccc (commented):
I think the original issue(https://github.com/flutter/flutter/issues/155660) was when dropdownMenuEntries are changed, didUpdateWidget needs rematch. But here and previous PR, we just introduced extra logic to keep the behavior correct.

I think the most simple solution that I can think of is the comment [here](https://github.com/flutter/flutter/pull/162062#discussion_r1990504037). It fixed the original issue and won't introduce regression. Please do consider that solution as I just feel the previous PR and this one added new logics in several places.

ahmedrasar:
@QuncCccccc. Yes, #155660 was about initial selection not rematching. But the issue is present with any selection not just initial selection. Fixing the issue only for initial selection would be more bizarre  IMO. 

> I think the most simple solution that I can think of is the comment https://github.com/flutter/flutter/pull/162062#discussion_r1990504037. It fixed the original issue and won't introduce regression. Please do consider that solution as I just feel the previous PR and this one added new logics in several places.

It would be more straightforward for sure, but for it to work as expected labels should be unique, otherwise it could cause mismatching. Also, an additional variable would be needed to counter the case of using an external controller.  That was the reason for reverting #155757. 

A third option would be something like
```dart
class _DropdownMenuState<T> {
late T _lastSleectedValue;
late bool _hasSelection;
}
```
It would eliminate the frequent looping to get the value since it's already stored, with the trade off having to maintain an additional variable. 

`_lastSelectedValue` could cause a slight inconvenience since it can't be assigned to `null` to mean 'none'.

QuncCccccc (changes requested):
Thanks a lot for your contribution.

> It would be more straightforward for sure, but for it to work as expected labels should be unique, otherwise it could cause mismatching.

Could you provide an example for this? I don't think it doesn't work when the labels are the same. Here's the demo. The value 4 and 5 have the same label "Five" but the search highlight can find the correct "Five".

https://github.com/user-attachments/assets/1ba4f58d-cac4-4f6b-869f-407547c7b62c

However, with this PR and previous PR change, the demo shows a wrong search highlight (When we select the second "Five", next time when we open the menu, it should highlight the second "Five").

https://github.com/user-attachments/assets/69b4a8db-b032-4344-abf6-ae96450b5a3d

Please do consider to revert the previous PR and apply the change I suggested([here](https://github.com/flutter/flutter/pull/162062#discussion_r1990504037)). Please let me know if there are any concerns:)

bleroux:
> Please do consider to revert the previous PR

I agree with this, it is always sad to have to revert previous work but this is usually the best option as it gives time to find a proper solution on master while the revert will be cherry-picked quickly.
That's particularly true when the regression is more problematic than the original issue.
In that particular case, the PR to revert would be https://github.com/flutter/flutter/pull/162062 which was a fix for https://github.com/flutter/flutter/issues/155660.
@ahmedrasar Do you have the bandwidth to file the revert? Let me know, if not I will do. Sorry for the inconvenience.

ahmedrasar:
What I meant by 'it could cause mismatching' is that when two or more entries have the same label, the looping function (`firstWhere`) would always return the first match.

Take this example

https://github.com/user-attachments/assets/7dca8613-611c-4868-af2e-8af149265d11

Although both green and blue have the same label (green) toggling localization wouldn't mismatch blue to green's label (since it's first in the last). This is because both have different values. If we would rematch based on the current textfield aganist the previous list's labels, blue (with green label) would map to verde (green in espanol).

<details>

```dart

void main(List<String> args) {
  runApp(MaterialApp(home: DropdownPage()));
}

class DropdownPage extends StatefulWidget {
  const DropdownPage({super.key});

  @override
  _DropdownPageState createState() => _DropdownPageState();
}

class _DropdownPageState extends State<DropdownPage> {
  var _selectedValue = Colors.white;
  Localization _localization = EnglishLocalization();

  @override
    return Scaffold(
      appBar: AppBar(title: Text('Dropdown Page')),
      backgroundColor: _selectedValue,
      body: Row(
        children: [
          DropdownMenu(
            initialSelection: _selectedValue,
            onSelected: (Color? newValue) {
              setState(() {
                _selectedValue = newValue!;
              });
            },
            dropdownMenuEntries: [
              DropdownMenuEntry<Color>(
                value: Colors.red,
                label: _localization.redLabel,
              ),
              DropdownMenuEntry<Color>(
                value: Colors.green,
                label: _localization.greenLabel,
              ),
              DropdownMenuEntry<Color>(
                value: Colors.blue,
                label: _localization.blueLabel,
              ),
            ].toList(),
          ),

          IconButton(
            onPressed: () {
              // Toggle between English and Spanish localization
              setState(() {
                _localization = _localization is EnglishLocalization
                    ? EspanolLocalization()
                    : EnglishLocalization();
              });
            },
            icon: Icon(Icons.language),
          ),
        ],
      ),
    );
  }
}

abstract class Localization {
  String get redLabel;
  String get greenLabel;
  String get blueLabel;
}

class EnglishLocalization implements Localization {
  @override
  String get redLabel => 'Red';

  @override
  String get greenLabel => 'Green';

  @override
  String get blueLabel => 'Green';
}

class EspanolLocalization implements Localization {
  @override
  String get redLabel => 'Rojo';

  @override
  String get greenLabel => 'Verde';

  @override
  String get blueLabel => 'Azul';
}
```
</details>

It's not really a big deal since it wouldn't make sense to use the same label anyway. ~~Also the use of additional variable is not necessary since checking `_localTextEditingController` could do the trick, sorry for not noticing this at first.~~ Additional variable is **STILL** needed.

I'll file a revert and apply the new changes asap.

ahmedrasar:
@QuncCccccc. Hi! I did some tests on what you said here on 3.35.2, 3.27.3, this PR, and master. And it's reproducible on all of them. It seems that the problem is not about highlighting the current entry but with searching and highlighting the first match of the search function. 

In this clicking the dropdown menu itself has no effect on the current entry is highlighted despite having the same label. The first match is highlighted when you press the textfield itself because it apply the search function.

This test is on 3.27.3 before the previous PR.

https://github.com/user-attachments/assets/5c11f777-370c-42b0-ad3d-c0584101caad

<details>

<summary> code</summary>

```dart

void main() {
  runApp(DropdownSameLabels());
}

class DropdownSameLabels extends StatelessWidget {
  const DropdownSameLabels({super.key});

  @override
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(title: Text('Dropdown Menu with Same Labels')),
        body: DropdownMenuExample(),
      ),
    );
  }
}

class DropdownMenuExample extends StatefulWidget {
  const DropdownMenuExample({super.key});

  @override
  _DropdownMenuExampleState createState() => _DropdownMenuExampleState();
}

class _DropdownMenuExampleState extends State<DropdownMenuExample> {
  int? _selectedItem;

  @override
    return Row(
      spacing: 10,
      children: <Widget>[
        Text('Selected Item: ${_selectedItem ?? 'None'}'),
        DropdownMenu(
          // requestFocusOnTap: false,
          onSelected: (int? newValue) {
            setState(() {
              _selectedItem = newValue;
            });
          },
          dropdownMenuEntries: const <DropdownMenuEntry<int>>[
            DropdownMenuEntry(value: 1, label: 'One'),
            DropdownMenuEntry(value: 2, label: 'Two'),
            DropdownMenuEntry(value: 3, label: 'Three'),
            DropdownMenuEntry(value: 4, label: 'Four'),
            DropdownMenuEntry(value: 5, label: 'Five'),
            DropdownMenuEntry(value: 6, label: 'Five'),
            DropdownMenuEntry(value: 7, label: 'Five'),
            DropdownMenuEntry(value: 8, label: 'Five'),
            DropdownMenuEntry(value: 9, label: 'Nine'),
            DropdownMenuEntry(value: 10, label: 'Ten'),
          ],
          initialSelection: _selectedItem,
        ),
      ],
    );
  }
}
```
</details>

It could be worth creating an issue for it, but as you suggested in the previous PR, having duplicate values in the entries list doesn't make sense, so I would assume that having duplicate labels also doesn't make sense.

And about this PR, I'm okay with either storing the index or compering the textfield. But compering the textfield would result in the subtle bug I mentioned in the previous comment, so WDYT?

QuncCccccc:
> And about this PR, I'm okay with either storing the index or compering the textfield. But compering the textfield would result in the subtle bug I mentioned in the previous comment, so WDYT?

Oh I thought we are going to revert the PR that caused regression and start a new fix?

ahmedrasar:
I see no reason for the revert since the bug you mentioned is not related to this or the previous PR. The new bug that this PR fixes is a simple mistake of storing the index of selected entry from the filtered list instead of the absolute index. Even when applying your changes, it's about the same since the only thing that would differ is getting the index from a loop instead of storing it locally.

bleroux:
> I see no reason for the revert since the bug you mentioned is not related to this or the previous PR

You put a lot of effort on analyzing this problem, this is much appreciated. :pray: 
One reason to advocate for the revert was to give more time to study other solutions. We were expecting to be able to discuss in details in the new PR while users would have the revert being cherry-picked. Caching index is often a source of bugs, so it would be great to find another way.

For instance, in https://github.com/flutter/flutter/pull/174757#issuecomment-3241481754 you mentioned that storing values would be better but not possible because  "So if an entry with null value was clicked". This is true, but maybe the problem is that this is a bug which has to be fixed and we should not accept entries with null value (see https://github.com/flutter/flutter/pull/176711/ which might address this problem).

bleroux:
@QuncCccccc I think the most actionable path is to get this PR merged. The additional code is somewhat small and it will be easy to cherry-pick this fix. (Cherry-picking the revert might be more difficult as the original PR landed long ago).
Once this is merged, I will revisit the solution and try to remove index caching (https://github.com/flutter/flutter/pull/176711 might help for this).

ahmedrasar:
Hi, I'm sorry for taking to long to response to you, but I'm busy and won't be available for 2 months. Wish for the best.

QuncCccccc:
Thanks for submitting this PR and for your contribution! Given the need for further discussion on the solution and the two-month availability gap, we're going to close this PR for now. I'll pick up the fix with @bleroux to ensure we fix the issue soon. We appreciate the initial work and hope to see your future contributions!

PR #174761, opened 2025-08-30
Make sure that a VerticalDivider doesn't crash at 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the VerticalDivider widget.


victorsanni (approved):

dkwingsmt (approved):
Checklist:

PR #174776, opened 2025-09-01
Fix CupertinoNavigationBar MediaQuery regression #174771

realmeylisdev:
## Problem

Since Flutter 3.35, `MediaQuery.of(context)` inside CupertinoNavigationBar components (trailing, leading, middle, etc.) returns `Size(0.0, 0.0)` instead of the correct screen dimensions. This regression breaks responsive logic and layout decisions made within navigation bar components.

## Root Cause

The issue occurs because CupertinoNavigationBar components wrap their child widgets with `MediaQuery` widgets that only provide `textScaler` data using `MediaQueryData(textScaler: ...)`. When creating a `MediaQueryData` with only specific properties, all other properties default to their initial values, causing `size` to default to `Size.zero`.

Affected components:
- Leading widget
- Trailing widget 
- Middle widget
- Back chevron
- Back label
- Large title
- Bottom widget

## Solution

Replace `MediaQueryData()` constructors with `MediaQuery.of(context).copyWith()` to preserve all existing MediaQuery properties (size, padding, devicePixelRatio, etc.) while only modifying the text scaling behavior.

### Core Fix
- Modified 6 `MediaQueryData` constructor calls in `packages/flutter/lib/src/cupertino/nav_bar.dart`
- Preserves all inherited MediaQuery properties while maintaining text scaling limits
- Ensures `size`, `padding`, `devicePixelRatio` and other properties are not reset to defaults

### Testing
- Added comprehensive test suite with 3 focused test cases in `packages/flutter/test/cupertino/nav_bar_test.dart`
- Tests verify MediaQuery size preservation in trailing/leading widgets
- Regression test specifically prevents Size(0,0) from returning
- Tests are minimal and focused on the core issue

### Fixed Issues
- ✅ MediaQuery.of(context).size returns correct screen dimensions
- ✅ MediaQuery.of(context).padding returns correct safe area padding  
- ✅ MediaQuery.of(context).devicePixelRatio returns correct pixel ratio
- ✅ Responsive navigation bar components work correctly
- ✅ No impact on existing text scaling behavior

### Compatibility
- ✅ Backward compatible - no breaking changes
- ✅ No impact on Material design widgets
- ✅ All existing CupertinoNavigationBar functionality preserved
- ✅ Performance neutral - same number of MediaQuery widgets

## Testing Instructions

To verify the fix:

```dart
CupertinoNavigationBar(
  trailing: Builder(
    builder: (context) {
      final size = MediaQuery.of(context).size;
      print('Size: $size'); // Should print actual screen size, not Size(0.0, 0.0)
      return Icon(CupertinoIcons.info);
    },
  ),
)
```

Fixes flutter/flutter#174771


reidbaker (approved):
Need approval from someone on the framework team before landing. 

@justinmc I think this should be a candidate for a 3.35 cherry pick.

reidbaker:
@meylis1998 thank you for your contribution!

victorsanni:
Hi @meylis1998, thanks for the contribution and the quick fix! A similar [issue](https://github.com/flutter/flutter/issues/174642) was filed last week, for which I added the same [fix](https://github.com/flutter/flutter/pull/174673) as in this PR. I'll go ahead and dedupe to that PR since it is closer to completion. Sorry to take this one off your plate.

PR #127137, opened 2023-05-18
Adds SemanticsNode Finders for searching the semantics tree

pdblasi-google:
* Pulled `FinderBase` out of `Finder`
  * `FinderBase` can be used for any object, not just elements
  * Terminology was updated to be more "find" related
* Re-implemented `Finder` using `FinderBase<Element>`
  * Backwards compatibility maintained with `_LegacyFinderMixin`
* Introduced base classes for SemanticsNode finders
* Introduced basic SemanticsNode finders through `find.semantics`
* Updated some relevant matchers to make use of the more generic `FinderBase`

Closes #123634
Closes #115874

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #174787, opened 2025-09-01
Fix InputDecoration does not apply errorStyle to error

bleroux:
## Description

This PR fixes `InpuDecorator` not applying `InputDecoration.errorStyle` to `InputDecoration.error`.

## Before

Text from `InputDecoration.error` is not styled correctly:

image

## After

Text from `InputDecoration.error` is styled correctly:

image

## Related Issue

Fixes [InputDecorator does not apply errorStyle to error](https://github.com/flutter/flutter/issues/174784) 

## Tests

Adds 1 test.

### DIFF
--- a/packages/flutter/test/material/input_decorator_test.dart
+++ b/packages/flutter/test/material/input_decorator_test.dart
@@ -6210,6 +6210,21 @@ void main() {
           );
         }, throwsAssertionError);
       });
+
+      // Regression test for https://github.com/flutter/flutter/issues/174784.
+      testWidgets('InputDecorator error widget text style defaults to errorStyle', (
+        WidgetTester tester,
+      ) async {
+        await tester.pumpWidget(
+          buildInputDecorator(decoration: const InputDecoration(error: Text(errorText))),
+        );
+
+        expect(findError(), findsOneWidget);
+        final ThemeData theme = Theme.of(tester.element(findDecorator()));
+        final Color expectedColor = theme.colorScheme.error;
+        final TextStyle expectedStyle = theme.textTheme.bodySmall!.copyWith(color: expectedColor);
+        expect(getErrorStyle(tester), expectedStyle);
+      });
     });
 
     testWidgets('InputDecorator with counter does not crash when given a 0 size', (
--- a/packages/flutter/lib/src/material/input_decorator.dart
+++ b/packages/flutter/lib/src/material/input_decorator.dart
@@ -401,8 +401,11 @@ class _HelperErrorState extends State<_HelperError> with SingleTickerProviderSta
 
   Widget _buildError() {
     assert(widget.error != null || widget.errorText != null);
-    final Widget? capturedError = widget.error;
     final String? capturedErrorText = widget.errorText;
+    Widget? capturedError = widget.error;
+    if (capturedError != null && widget.errorStyle != null) {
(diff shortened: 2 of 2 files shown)


bleroux:
@justinmc If you get a chance to review this PR, it would be great if it can get merged before or not long after https://github.com/flutter/flutter/pull/174831 (which relies on this fix to get a proper visual rendering).

packages/flutter/test/material/input_decorator_test.dart
```diff
@@ -6210,6 +6210,21 @@ void main() {
...
+
+      // Regression test for https://github.com/flutter/flutter/issues/174784.
+      testWidgets('InputDecorator error widget text style defaults to errorStyle', (
+        WidgetTester tester,
+      ) async {
+        await tester.pumpWidget(
+          buildInputDecorator(decoration: const InputDecoration(error: Text(errorText))),
+        );
+
+        expect(findError(), findsOneWidget);
```
ValentinVignal, line 6222: Reading the documentation of `findsOneWidget`, it is written that

> This is equivalent to the preferred [findsOne] method.

Should we be using `findsOne` instead of `findsOneWidget`?

ValentinVignal (approved):
LGTM

packages/flutter/test/material/input_decorator_test.dart
bleroux: > Should we be using findsOne instead of findsOneWidget?

Interesting question!
Searching through the codebase:
- findsOne: 8218
- findsOneWidget: 7610

In input_decorator_test.dart file, there are no occurences of `findsOne`, only `findsOneWidget`.
The reason is probably because `findsOne` was introduced somewhat recently, see https://github.com/flutter/flutter/pull/127137.
I would say using findsOneWidget when we are looking for a Widget is ok, but maybe @justinmc has some thinking about this.

justinmc: I did some major archaeology here and my conclusion is that we should actually be using findsOne and not findsOneWidget! See these two comment threads:

  * https://github.com/flutter/flutter/pull/127137#discussion_r1213744575
  * https://github.com/flutter/flutter/pull/127137#discussion_r1278034663

It seems like there was a long-term plan to deprecate findsOneWidget, but the people involved are no longer working on Flutter.

So I guess strictly this PR should be changed to use findsOne. And I should stop using findsOneWidget like I have been for all my time on Flutter.

But in reality, both are equivalent and it doesn't really matter. I'll just keep this fact in my head in case anyone ever comes along with the motivation to clean up these duplicate methods and do the deprecation.

justinmc (approved):
LGTM 👍. Sorry for the delay on this! Treat my comment below as a total nit.

packages/flutter/test/material/input_decorator_test.dart
ValentinVignal: I see, thank you for those insights @justinmc. Is there something blocking us from deprecated `findsOneWidget` and other similar finders today? 

If not, I could open PRs to deprecate them and migrate to `findsOne` and similar finders

bleroux:
> LGTM 👍. Sorry for the delay on this! Treat my comment below as a total nit.

Thanks for the review 🙏 
I will let the findsOneWidget for the moment as it would be better to replace all occurrences in this test file at once.

Issue #174366, opened 2025-08-25
`ListTile.selectedTileColor` not visible when parent `Container` has explicit background color


cnayan:
### Steps to reproduce

ListTile's selection highlighting (`selectedTileColor`) becomes invisible or non-functional when the `ListTile` is nested inside a `Container` with an explicit color property set.

Steps to Reproduce:
1. Create a ListView with ListTile items inside a Container
2. Set an explicit background color on the parent Container (e.g., `Color.fromARGB(255, 27, 27, 27)`)
3. Set `selectedTileColor: Colors.blue` on the ListTile
4. Set `selected: true` on a ListTile item
5. Observe that the selection highlighting is not visible

Expected Behavior:
- ListTile should display the specified selectedTileColor (blue background) when selected: true, regardless of parent Container background color.

Actual Behavior:
- ListTile selection highlighting is not visible when parent Container has an explicit color property. The blue selection background does not appear.

Workaround:
- Remove the color property from the parent Container. The selection highlighting works correctly when no explicit color is set on parent containers.

---

Environment:

Flutter version: 3.35.1
Dart version: 3.9.0 (stable) (Mon Aug 11 07:58:10 2025 -0700) on "windows_x64"
Test Platform: Desktop

---

Code Description

```dart
// This DOESN'T work - selection not visible
Container(
  color: Color.fromARGB(255, 27, 27, 27),
  child: ListView.builder(
    itemBuilder: (context, index) => ListTile(
      title: Text('Item $index'),
      selected: selectedIndex == index,
      selectedTileColor: Colors.blue,
    ),
  ),
)

// This WORKS - selection visible
Container(
  // color: Color.fromARGB(255, 27, 27, 27), // Commented out
  child: ListView.builder(
    itemBuilder: (context, index) => ListTile(
      title: Text('Item $index'),
      selected: selectedIndex == index,
      selectedTileColor: Colors.blue,
    ),
  ),
)
```

### Expected results

- ListTile should display the specified selectedTileColor (blue background) when selected: true, regardless of parent Container background color.

### Actual results

- ListTile selection highlighting is not visible when parent Container has an explicit color property. The blue selection background does not appear.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

class Model {
  final String name;
  Model({required this.name});
}

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
[...]

PR #174811, opened 2025-09-02
docs(ListTile): clarify tileColor/selectedTileColor visibility with colored ancestors

jihanurrahman33:
Clarify in ListTile API docs that tileColor and selectedTileColor are painted
by the nearest Material ancestor and can be obscured by a colored ancestor
(e.g., Container/ColoredBox) placed between ListTile and that Material.

This adds a short note to both tileColor and selectedTileColor docs and
points readers to the existing guidance showing how to wrap ListTile with
Material when needed.

Refs: flutter/flutter#174366

### DIFF
--- a/packages/flutter/lib/src/material/list_tile.dart
+++ b/packages/flutter/lib/src/material/list_tile.dart
@@ -691,10 +691,22 @@ class ListTile extends StatelessWidget {
   /// When that is also null, the [ListTileTheme.selectedTileColor] is used, otherwise
   /// [Colors.transparent] is used.
   ///
-  /// {@endtemplate}
+  /// ///
+/// Note: These colors are painted by the nearest Material ancestor.
+/// If a colored ancestor (for example, a Container/ColoredBox with
+/// a non-null color) sits between the ListTile and its Material, the
+/// tileColor may be obscured. Wrap the ListTile with a Material or apply
+/// the background color to the Material instead.
+{@endtemplate}
   final Color? tileColor;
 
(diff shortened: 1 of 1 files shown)


jihanurrahman33:
Thanks for the note about tests.

This PR is documentation-only (API docs for ListTile), clarifying existing behavior rather than changing code. ListTile’s background painting behavior is already covered by existing tests in the repo, for example:

- packages/flutter/test/material/list_tile_test.dart:
  - testWidgets('ListTile respects tileColor & selectedTileColor', …) — verifies tileColor/selectedTileColor are painted by the Material ancestor.
  - testWidgets('ListTile shows Material ripple effects on top of tileColor', …) — ensures ink is painted over the tile’s Material.

Given this is a docs change that aligns with current behavior and existing tests, I’m requesting a test exemption for this PR.

If you’d prefer an additional targeted test (e.g., a scenario where a ColoredBox sits between ListTile and Material, demonstrating the background can be obscured), I’m happy to add one — though that would test general compositing rather than ListTile logic.

cnayan:
@jihanurrahman33 Can images be included? As mentioned in https://github.com/flutter/flutter/issues/174366#issuecomment-3235792083

QuncCccccc:
Thank you for your contribution! I created a fix(https://github.com/flutter/flutter/pull/174858) so when we use Container to wrap ListTile, ListTile's background color is still visible. For now, I'm closing this PR because once the fix lands, the doc seems a bit conflictive. Please feel free to take a look and leave comments in the fix PR. Let me know if there're any concerns:)

Issue #172405, opened 2025-07-19
showPerformanceOverlay=true in MaterialApp throws UnimplementedError on web with no clear “not supported” message


RocksteadyDog:
### Use case

Hello!

A small proposal

Enabling the performance overlay in a Flutter web app by setting:
```dart
MaterialApp(
  showPerformanceOverlay: true,
  // …
)
```

results in the following UnimplementedError:
```dart
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following UnimplementedError was thrown during a scheduler callback:
UnimplementedError

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 266:3       throw_
errors.dart:266
dart-sdk/lib/_internal/js_dev_runtime/private/profile.dart 117:39                 addPerformanceOverlay
profile.dart:117
package:flutter/src/rendering/layer.dart 1056:12                                  addToScene
layer.dart:1056
package:flutter/src/rendering/layer.dart 711:5                                    [_addToSceneWithRetainedRendering]
layer.dart:711
package:flutter/src/rendering/layer.dart 1372:12                                  addChildrenToScene
layer.dart:1372
package:flutter/src/rendering/layer.dart 2083:5                                   addToScene
layer.dart:2083
package:flutter/src/rendering/layer.dart 1120:5                                   buildScene
layer.dart:1120
package:flutter/src/rendering/view.dart 358:30                                    compositeFrame
view.dart:358
package:flutter/src/rendering/binding.dart 634:19                                 drawFrame
binding.dart:634
package:flutter/src/widgets/binding.dart 1242:13                                  drawFrame
binding.dart:1242
package:flutter/src/rendering/binding.dart 495:5                                  [_handlePersistentFrameCallback]
binding.dart:495
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 117:77  tear
operations.dart:117
package:flutter/src/scheduler/binding.dart 1438:7                                 [_invokeFrameCallback]
binding.dart:1438
package:flutter/src/scheduler/binding.dart 1351:9                                 handleDrawFrame
binding.dart:1351
package:flutter/src/scheduler/binding.dart 1060:9                                 <fn>
binding.dart:1060
lib/_engine/engine/frame_service.dart 147:9                                       <fn>
frame_service.dart:147
[...]

PR #174812, opened 2025-09-02
fix(widgets): guard showPerformanceOverlay on web with no-op and debug message

jihanurrahman33:
Web currently throws an UnimplementedError when `showPerformanceOverlay` is
enabled (MaterialApp/WidgetsApp) because the engine does not implement the
performance overlay layer. This change makes the framework:

- No-op on web (do not insert PerformanceOverlay), and
- In debug mode, emit a single error report message explaining it is not
  supported on web.

This avoids a crash while providing a clearer signal to developers.

Tests: platform-specific (web) behavior; existing non-web overlay tests remain
valid. Requesting test exemption for the web-only path.

Refs: flutter/flutter#172405

### DIFF
--- a/packages/flutter/lib/src/widgets/app.dart
+++ b/packages/flutter/lib/src/widgets/app.dart
@@ -232,7 +232,7 @@ Locale basicLocaleListResolution(
   // supported locale.
   final Locale resolvedLocale = matchesLanguageCode ?? matchesCountryCode ?? supportedLocales.first;
   return resolvedLocale;
-}
+      }
 
 /// The signature of [WidgetsApp.onGenerateTitle].
 ///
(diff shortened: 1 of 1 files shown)


jihanurrahman33:
Thanks for the quick review!

- Addressed feedback: fixed indentation of the block within `showPerformanceOverlay` and removed `const` from `Positioned` (latest commit: b2d1924).
- Behavior unchanged on non‑web; web path now guards without crashing and reports a clear debug message.

Tests: this change is a web‑only guard in `WidgetsApp`. Requesting a test exemption; happy to add a small framework test if preferred (e.g., guard that no `PerformanceOverlay` is inserted on web and that a single debug error is reported).

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -232,7 +232,7 @@ Locale basicLocaleListResolution(
   // supported locale.
   final Locale resolvedLocale = matchesLanguageCode ?? matchesCountryCode ?? supportedLocales.first;
   return resolvedLocale;
-}
+      }
```
justinmc, line 235: FYI there are weird formatting changes that shouldn't have been made here and below.

justinmc:
@jihanurrahman33 Are you still planning to fix the formatting/analyzer problems here?

Piinks:

Issue #173302, opened 2025-08-05
[Google3 Bug]: State information not announced for AM/PM in Material DateTimePicker


evantea:
### Help us understand the severity of this issue

- [ ] causing severe production issues e.g. malfunctions or data loss
- [ ] blocking next binary release
- [x] blocking a client feature launch within a quarter
- [ ] nice-to-have but does not block a launch within the next quarter

### Steps to reproduce

1. Turn on Voicover
2. Navigate to the DateTimePicker
3. Select a date
4. Navigate to the AM/PM buttons and select one of them

### Expected results

VoiceOver announces the AM/PM buttons with state information (e.g. selected, PM, button).

### Actual results

VoiceOver announces the AM/PM buttons without state information.

### Code sample
<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      debugShowCheckedModeBanner: false,
      home: Scaffold(body: Center(child: MyWidget())),
    );
  }
}

class MyWidget extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    WidgetsBinding.instance.addPostFrameCallback((timestamp) {
      showTimePicker(
        context: context,
        initialTime: TimeOfDay.now(),
        orientation: Orientation.portrait,
      );
    });

    return Container(height: 300, width: 100);
  }
}
```

</details>

### Screenshots or Video

<img width="550" height="573" alt="Image" src="https://github.com/user-attachments/assets/453799a3-8929-40a4-b25c-70a8a6b4acc0" />

### Flutter Doctor output

n/a

evantea:
This bug affects the "chalk" app

evantea:
The semantics are currently being announced as RadioButton and "checked" and "not-checked", closing for now to verify with our accessibility team that this is enough. I'll re-open with more clear instructions if any action is necessary :)

evantea:
Re-opening! The information provided above is only happening on Android, for iOS it simply says the following no matter the state of the buttons:
- "AM, button"
- "PM, button"

tirth-patel-nc:
@evantea, thanks for the report. We are seeing the same behavior with the latest SDK versions on iOS. My colleague verified it on iOS as well:

```
stable : 3.32.8
master : 3.33.0-1.0.pre-1299
```

| Android | iOS |
| --- | --- |
[...]

PR #174816, opened 2025-09-02
a11y(material): make AM/PM buttons live regions so state change is announced

jihanurrahman33:
Improve accessibility of the Material TimePicker AM/PM control by marking the
AM/PM buttons as live regions so their state change is announced by screen
readers when toggled.

Details:
- In `_AmPmButton` (packages/flutter/lib/src/material/time_picker.dart), add
  `liveRegion: true` to the Semantics node wrapping the AM/PM button. The node
  already provides `button`, `inMutuallyExclusiveGroup`, and `checked` flags;
  making it a live region prompts an announcement when its checked state
  changes.

Tests:
- Existing semantics tests validate AM/PM semantics (label, flags, actions).
  This change augments semantics and should not break them. If preferred, I can
  add an extra assertion for the live region flag.

Refs: flutter/flutter#173302

### DIFF
--- a/packages/flutter/lib/src/material/time_picker.dart
+++ b/packages/flutter/lib/src/material/time_picker.dart
@@ -809,6 +809,7 @@ class _AmPmButton extends StatelessWidget {
       checked: selected,
       inMutuallyExclusiveGroup: true,
       button: true,
+      liveRegion: true,
       child: Padding(
         padding: padding,
         child: Material(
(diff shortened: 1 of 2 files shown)


jihanurrahman33:
Added a test asserting AM/PM buttons are live regions (commit 73c293b). This accompanies the Semantics change and helps prevent regressions. Happy to extend the check to PM as well if desired.

dkwingsmt:
cc @chunhtai

chunhtai:
The linked issue is about when the button is focused, it doesn't pronounce the state when focused in iOS. For example, if am button was already selected, when you swipe the a11y focus to the button, it should announce something like `selected, am, button`. I think this pr may not address the original issue.

chunhtai:
Hi @jihanurrahman33 Do you still plan to continue work on this issue?

dkwingsmt:

Issue #170825, opened 2025-06-18
☔ Privacy sandbox on android (.asb) b/322362589


reidbaker:
[Privacy Sandbox](https://privacysandbox.com/intl/en_us/android/) sdks are a piece of android functionality that feature flutter wishes to make easy for developers to use. 

Privacy sandbox relies on tooling that depends on android app bundles. `flutter build aab` is a supported flutter command that builds an app bundle but `build apk` and `run` do not go through the android app bundle phase. 

Additionally the fluter android team considers not building apks from app bundles technical debt. 

The umbrella bug will track the work the flutter team does to make privacy sandbox sdks easy to consume while not regressing in existing functionality. 

Googler only supplemental documentation [go/flutter-issue-170825](go/flutter-issue-170825)

- [ ] Milestone 1: A new flutter command exists that will build a universal apk from an aab `flutter build aab-apk` https://github.com/flutter/flutter/issues/166751 
- [ ] Milestone 2: Flutter build apk is replaced by flutter build aab-apk with an opt in flag
- [ ] Milestone 3: Flutter build apk supports building privacy sandbox  dependencies when opt in flag is used
- [ ] Milestone 4: Flutter run supports building privacy sandbox dependencies when opt in flag is used
- [ ] Milestone 5: Opt in flag is changed to Opt out flag. 
- [ ] Milestone 6: Opt out flag is removed and legacy apk and run code is removed . 

https://buganizer.corp.google.com/issues/390131309

jmagman:
Instead of a new command, can we add the needed flags to existing commands `flutter build apk` or `flutter build aab`? The suite of build commands is getting confusing, we could branch on what to actually do on whether there's a flag like `--asb=123.asb,456.asb` to either command?

reidbaker:
Possible dupe of https://github.com/flutter/flutter/issues/142750

jmagman:
This work has been paused. Closing, we can re-open if we decide this needs to be done in the future.

PR #174839, opened 2025-09-02
Add --enable-gradle-managed-install flag for flutter run

ash2moon:
Part of https://github.com/flutter/flutter/issues/170825

This adds a new `--enable-gradle-managed-install` flag which delegates the responsibility of APK installation to Gradle, allowing for improved privacy sandbox support.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/lib/src/android/android_device.dart
```diff
@@ -562,25 +562,48 @@ class AndroidDevice extends Device {
...
-      );
-      // Package has been built, so we can get the updated application ID and
-      // activity name from the .apk.
-      builtPackage =
-          await ApplicationPackageFactory.instance!.getPackageForPlatform(
-                devicePlatform,
-                buildInfo: debuggingOptions.buildInfo,
-              )
-              as AndroidApk?;
+      if (debuggingOptions.enableGradleManagedInstall) {
```
reidbaker, line 566: This method is already pretty long. Is there a reasonable extraction of this logic in the if and else blocks?

packages/flutter_tools/lib/src/android/android_device.dart
```diff
@@ -562,25 +562,48 @@ class AndroidDevice extends Device {
...
+        // Package has not been built yet, but we can still get the package information
+        // from the manifest.
+        builtPackage = await ApplicationPackageFactory.instance!.getPackageForPlatform(
```
reidbaker: I think this is copy of previous code but why do we know that this is a safe to "!"?

packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -222,6 +222,11 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
         FlutterOptions.kWebWasmFlag,
         help: 'Compile to WebAssembly rather than JavaScript.\n$kWasmMoreInfo',
         negatable: false,
+      )
+      ..addFlag(
+        'enable-gradle-managed-install',
+        negatable: false,
```
reidbaker: I think we want this to be negatable so that we can modify the default later if we choose to and let people opt out.

packages/flutter_tools/lib/src/device.dart
```diff
@@ -1346,6 +1351,7 @@ class DebuggingOptions {
         ipv6: (json['ipv6'] as bool?) ?? false,
         google3WorkspaceRoot: json['google3WorkspaceRoot'] as String?,
         printDtd: (json['printDtd'] as bool?) ?? false,
+        enableGradleManagedInstall: (json['enableGradleManagedInstall'] as bool?) ?? false,
```
reidbaker: Can this use a constant defined somewhere instead of the string? I recognize that is not the pattern here but am wondering why.

packages/flutter_tools/lib/src/android/android_device.dart
ash2moon: Should be safe because the flutter commands that use `android_device.dart` are all wrapped in a [`runInContext`](https://github.com/flutter/flutter/blob/6fd30fd25af047fba43b4538982d7677bb0a229d/packages/flutter_tools/lib/src/context_runner.dart#L127) block inside the [`runner.run`](https://github.com/flutter/flutter/blob/599ed5252e806c98fc869d2bb6f34f61e67410a0/packages/flutter_tools/lib/runner.dart#L51) command.

packages/flutter_tools/lib/src/device.dart
ash2moon: Added a new section for constants within this file. Will create a child PR to fix the current set of strings in here.

packages/flutter_tools/test/general.shard/android/android_device_test.dart
```diff
@@ -546,6 +549,223 @@ Uptime: 441088659 Realtime: 521464097
...
+              'shell',
+              '-x',
+              'logcat',
+              '-v',
+              'time',
+              '-T',
+              '0',
+            ],
+          ),
+          const FakeCommand(
```
reidbaker: These commands seem fragile to changes. Could we make this easier to maintain by replacing the serial number with something from the fake android device (or maybe get the string parameters from `adbCommandForDevice`). 

SImilar questions for package id and the directory "/data/local/tmp/sky.com.example.app.sha1'"

packages/flutter_tools/test/general.shard/android/android_device_test.dart
```diff
@@ -546,6 +549,223 @@ Uptime: 441088659 Realtime: 521464097
...
+              '1234',
+              'shell',
+              'am',
+              'start',
+              '-a',
+              'android.intent.action.MAIN',
+              '-c',
+              'android.intent.category.LAUNCHER',
+              '-f',
+              '0x20000000',
```
reidbaker: What is this value?

packages/flutter_tools/test/general.shard/android/android_device_test.dart
```diff
@@ -546,6 +549,223 @@ Uptime: 441088659 Realtime: 521464097
...
+            command: <String>['adb', '-s', '1234', 'shell', 'getprop'],
+            stdout: '[ro.product.cpu.abi]: [arm64-v8a]\n[ro.build.version.sdk]: [99]',
+          ),
+          const FakeCommand(
+            command: <String>['adb', '-s', '1234', 'shell', 'am', 'force-stop', 'com.example.app'],
+          ),
+          const FakeCommand(
+            command: <String>['adb', '-s', '1234', 'install', '-t', '-r', 'dummy.apk'],
+          ),
+          const FakeCommand(
```
reidbaker: Are these commands across tests similar enough you could extract them to into a class final or a setup method?

packages/flutter_tools/test/general.shard/android/android_device_test.dart
ash2moon: This comes from [here](https://github.com/flutter/flutter/blob/9124bdff820097c2a0b1aa50913b7223b16797d0/packages/flutter_tools/lib/src/android/android_device.dart#L622); it basically does not launch the activity if it is already running at the top of the history stack. See https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_SINGLE_TOP

I added this as a comment on the line.

packages/flutter_tools/test/general.shard/android/android_device_test.dart
ash2moon: Sure, I made some builder methods to make it easier to maintain and extracted out/reused some of those string constants.

packages/flutter_tools/test/general.shard/android/android_device_test.dart
ash2moon: I moved my tests into `android_device_start_test.dart` which specifically covers testing the `startApp` function so that it's a bit more cohesive. I also refactored the hardcoded commands into builder* methods to make it easier to reuse across the other tests as well.

ash2moon (commented):
Refactored the test from `android_device_test.dart` to `android_device_start_test.dart`

Issue #171060, opened 2025-06-24
Enable to ignore certain children of a column when determining the width for the column


TB-StarcMed:
### Use case

I am writing an App where I have a Column on the left next to the content panel. The Column has Text and Image children. Currently, the Column's width is determined to be the maximum of the minimum widths of all children. I want that the width of the column is just determined based on the maximum of the minimum widths of only the Image children and Text children should just take that resulting width and word wrap. To achieve this, I currently have to wrap the Text children in a Container with the width property set to the usual width of the images + some experimentially determined offset. 

Just some notes: To achieve the word-wrapping on the Text components, I have softWrap = true, textWidthBasis = TextWidthBasis.parent and overflow = TextOverflow.visible on these Text components.

From my research, I have found https://stackoverflow.com/questions/58468915/flutter-have-child-match-another-childs-width-inside-of-column, where other users have a similar situation. The first solution given in there is not suitable to my situation since I have multiple children that should be used to determine the width of the Column. The second solution is not suitable either since it sizes the texts differently and puts all characters of the texts vertically above each other using. The third solution is a bit questionable since it relies on determining the wanted width based on the first layout and needs a refresh of the layout then. I also think that a Flutter vanilla solution would be quite useful.

### Proposal

I would like that there was a way to make the Column, Row and Flex components be able to ignore certain children when determining the size for the cross-axis. I imagine, there was a property on the Flexible component that would define that respective child should be ignored when determining the size of the cross-axis.

TB-StarcMed:
An example made for this issue: [column_sizing.zip](https://github.com/user-attachments/files/20880644/column_sizing.zip).

dkwingsmt:
I might not have understood your proposal perfectly, but can you achieve this by wrapping each of the specific children with a custom widget that lay outs its child regardless of the parent width?

TB-StarcMed:
[...]

Issue #137546, opened 2023-10-30
SizedBox calculates intrinsic dimensions incorrectly


Albert221:
Let's start with example, you can paste it into https://dartpad.dev

```dart
import 'package:flutter/material.dart';
import 'package:flutter/rendering.dart';

void main() {
  debugPaintSizeEnabled = true;   
  
  runApp(
    const MaterialApp(
      home: Scaffold(
        body: Center(
          child: IntrinsicHeight(
            child: SizedBox(
              width: 100,
              child: Column(
                children: [
                  Text(lipsum),
                ],
              ),
            ),
          ),
        ),
      ),
    ),
  );
}

const lipsum =
    'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam malesuada vulputate efficitur. Curabitur a nulla et nulla bibendum rhoncus. Duis nec porttitor nunc, vitae pharetra sapien. Ut et vulputate velit, eget porttitor nulla. Curabitur maximus dui id nunc ornare lacinia.';
```

The `Column` has height equal to intrinsic heigth of its child. But it's wrong. The `computeMinIntrinsicHeight` doesn't take into account the `SizedBox`'s `width` when it should. It calculates the intrinsic height with the height it receives from the parent. You can see this height by commenting out the `width` line.

Reproducible both on Flutter stable (`3.13.7`) and master (`3.16.0-11.0.pre.3`).

| Expected | Actual | What intrinsic width "sees" |
|-|-|-|
| ![image](https://github.com/flutter/flutter/assets/4085280/eade5a08-7351-4518-a278-365a24d9c1e0) | ![image](https://github.com/flutter/flutter/assets/4085280/a39183d4-5c3b-4238-87e8-2b007e459ba0) | ![image](https://github.com/flutter/flutter/assets/4085280/e34bbf33-49bf-4291-838e-0c567cdc89bf) |

The faulty code is in `RenderConstrainedBox`, so it touches not only `SizedBox`, but also `ConstrainedBox` and possibly other widgets.

danagbemava-nc:
Reproducible using the code sample provided above. 

<details>
<summary>flutter doctor -v</summary>

```
[!] Flutter (Channel stable, 3.13.9, on macOS 14.0 23A344 darwin-arm64, locale en-GB)
    • Flutter version 3.13.9 on channel stable at /Users/nexus/dev/sdks/flutter
[...]

PR #174849, opened 2025-09-02
[ Widget Preview] Add `group` property to `Preview`

### LINKED ISSUE #173871: Widget previews: allow grouping/organizing previews
### Use case

Currently, all detected previews are presented as one grid/list. It would be helpful to group previews into collapsible groups or sidebar tabs.

### Proposal

Allow specifying a group (or some other "path") on a preview. An API for this could look something like:

```dart
@Preview(name: 'MyText', group: 'components')
Widget myTextPreview() {...}

@Preview(name: 'MyButton', group: 'components')
Widget myButtonPreview() {...}

@Preview(name: 'HomeScreen', group: 'screens')
Widget homeScreenPreview() {...}
```

Here, `myTextPreview` and `myButtonPreview` would display together, and `homeScreenPreview` would be shown separetely.

bkonyi:
This change adds support for creating groups of widget previews. Previews with the same `group` parameter will be grouped together in a collapsable tile in widget previewer. Previews without a provided `group` will be placed in the `Default` group.

This change also slightly reworks the UI to make individual previews more easily distinguishable from preview groups, moves the controls for the previewer to the bottom of the preview window, and adds a divider between the preview name and preview content if a name is provided.

Fixes https://github.com/flutter/flutter/issues/173871

**Example:**

image


packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
```diff
@@ -340,60 +418,114 @@ class WidgetPreviewWidgetState extends State<WidgetPreviewWidget> {
...
 
-    preview = Column(
-      crossAxisAlignment: CrossAxisAlignment.start,
-      children: [
-        if (widget.preview.name != null) ...[
-          Text(
-            widget.preview.name!,
-            style: fixBlurryText(
-              TextStyle(fontSize: 16, fontWeight: FontWeight.w300),
+    // This IntrinsicWidth is here to prevent the Divider from forcing the Column
```
LongCatIsLooong: By divider you mean the `VerticalSpacer`? Generally I would not recommend using `Intrinsic{Width,Height}` since it doesn't work with some very common widgets (doesn't work with LayoutBuilders by design, and currently having issues with RenderConstrainedBoxes: https://github.com/flutter/flutter/issues/137546)

LongCatIsLooong: So you may see the `Error` widget if the developer puts the annotation on a widget that has a `LayoutBuilder` descendant. Instead the `VerticalSpacer` should be as narrow as possible since its horizontal dimension doesn't matter?

packages/flutter_tools/lib/src/widget_preview/preview_details.dart
```diff
@@ -49,6 +53,11 @@ final class PreviewDetails {
   /// instead of a `Widget`.
   final bool isBuilder;
 
+  /// The group the preview belongs to.
+  ///
+  /// If not provided, the preview will be part of the "empty" group.
```
LongCatIsLooong: I thought the default was "Default"?

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
bkonyi: Ah, I forgot to move this comment. It applies to the `Divider` in `_WidgetPreviewNameHeader`.

Yeah, I wasn't too sure about this, but I couldn't seem to get the `Divider` to stop making the `Column` expand to fill the width of the previewer rather than using the width of the previewed widget as a constraint. Maybe I'm missing something simple?

packages/flutter_tools/lib/src/widget_preview/preview_details.dart
bkonyi: Yes, it is. I wrote this comment before there was actually a group named `Default` This line can be removed.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
LongCatIsLooong: Ah I see, yeah it's unfortunate that it's not easy to achieve with a Column (https://github.com/flutter/flutter/issues/171060) since the divider shouldn't contribute to the cross size but also wants to be as wide as the Column. Is`Divider` only used in `_WidgetPreviewNameHeader`? The implementation looks pretty straightforward:
https://github.com/flutter/flutter/blob/52d5c5dc93c5372e1e895a3537173927030fbbfb/packages/flutter/lib/src/material/divider.dart#L185-L212

I think the easiest workaround might be taking the BoxDecoration + Padding from Divider, and adding them to the Text widget (so we don't need the Column).

LongCatIsLooong (approved):
LGTM except for the use of IntrinsicWidth.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
bkonyi: Unfortunately that doesn't work as it only results in the contents of Text being underlined. I'm trying to get a horizontal line dividing the title from the rest of the widget contents.

bkonyi: Ah, but it will work if I apply the decoration to the Column containing the actual preview instead of the Text! Thanks for the suggestion :)

PR #174860, opened 2025-09-03
FontAssets from the build hook

mosuem:
Allows to add fonts via build hooks with https://pub.dev/packages/font_asset, see https://github.com/mosuem/font_asset_test as an example.

This now also calls the icon treeshaker via a link hook.

TODO:
- [ ] Make adding via pubspec or build hook compatible
- [ ] Allow for more fonts per family
- [ ] Set up flutter config to hook input containing `app.dill` and other binaries.
- [ ] Use `@RecordUse` instead of custom logic.

Long term: Unbundle icontreeshaker from flutter, remove fonts via pubspec adding.

## Pre-launch Checklist

- [ ] I signed the [CLA].


mosuem:
@bkonyi @dcharkes @goderbauer wdyt? Just a proof of concept.

The next step would be to give `IconData` the `@RecordUse` annotation to replace the icon tree shaker with a build helper.

packages/flutter_tools/pubspec.yaml
```diff
@@ -23,6 +23,7 @@ dependencies:
   ffi: 2.1.4
   file: 7.0.1
   flutter_template_images: 5.0.0
+  font_asset: 0.0.4
```
dcharkes: This should have been `font_assets` to be consistent with `code_assets` and `data_assets`.

packages/flutter_tools/lib/src/asset.dart
```diff
@@ -103,6 +106,17 @@ class HookAsset {
   }
 }
 
+class FontHookAsset extends HookAsset {
+  FontHookAsset({
+    required super.file,
```
dcharkes, line 111: This doesn't seem to mirror the way the yaml is set up. The yaml has more than one file per font.

https://docs.flutter.dev/cookbook/design/fonts#define-fonts-in-the-pubspec-yaml-file

Would it make more sense to make a `FontFamilyAsset` with a list of `Font`s? In the current setup you'll end up repeating the family name. Especially if you end up writing a `package:font_assets` `FontAsset` object literal, it would make sense if it 1-to-1 translates to the yaml structure.

Also, I see a weight, but not an optional style?

dcharkes (commented):
Interesting!

I haven't looked at all the wiring, and if the caching works correctly with the targets etc.

This PR would benefit from an example (or better, an integration test as example).

I left some more remarks on the API design and naming.

mosuem:
> This PR would benefit from an example (or better, an integration test as example).

There is https://github.com/mosuem/font_asset_test, but I can add that as an example here instead.

packages/flutter_tools/lib/src/asset.dart
mosuem: Yes, I just went with a simple setup first to test it.

packages/flutter_tools/pubspec.yaml
mosuem: Yeah I noticed that afterwards as well.. :upside_down_face:

Piinks:
> I would like to move forward with this, I still believe it is the right path. I do not have the time right now unfortunately...

Totally understand! This has made it back to stale triage. Is it something we need to leave open in the meantime?

mosuem:
> Totally understand! This has made it back to stale triage. Is it something we need to leave open in the meantime?

That would be great!

Piinks:
Welcome back to stale triage! Should we close this for now? It has landed in stale triage many times now over the last year. If it is not actively being working on, perhaps we should close it. It can always be re-opened later.

mosuem:
Yeah it has unfortunately accumulated many merge conflicts by now, which is why I try to not let PRs get stale :/ Closing it, maybe someone else wants to tackle this sometime.

PR #174880, opened 2025-09-03
Make sure that a DropdownButton doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the DropdownButton widget.


dkwingsmt (approved):
Checklist:
- [x] The target widget is not a overlay 
- [x] Expect the target widget has a size of exactly `Size.zero`

packages/flutter/test/material/dropdown_test.dart
```diff
@@ -4694,4 +4694,26 @@ void main() {
...
+                  DropdownMenuItem<String>(value: 'a', child: Text('a')),
+                  DropdownMenuItem<String>(value: 'b', child: Text('b')),
+                ],
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(DropdownButton<String>)), Size.zero);
```
ahmedsameha1: It would be convenient to test that the DropdownMenuItem has a zero size, but the test fails with a size of (16.1, 0.0).

dkwingsmt: It's fine and expected. We've encountered many cases where the children of zero-sized parents have non-zero sizes. We should only test the target widget alone. `DropdownMenuItem` can have its own test in a similar fashion.

dkwingsmt (approved):
LGTM, thank you!

victorsanni (approved):

packages/flutter/test/material/dropdown_test.dart
```diff
@@ -4699,6 +4699,42 @@ void main() {
...
+          body: Center(
+            child: SizedBox.shrink(
+              child: DropdownMenuItem<String>(value: 'a', child: Text('a')),
+            ),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(DropdownMenuItem<String>)), Size.zero);
+  );
```
dkwingsmt:
```suggestion
  });
```

dkwingsmt:
Analyze error:
```dart

  error • The named parameter 'items' is required, but there's no corresponding argument • packages/flutter/test/material/dropdown_test.dart:4707:43 • missing_required_argument
  error • The named parameter 'onChanged' is required, but there's no corresponding argument • packages/flutter/test/material/dropdown_test.dart:4707:43 • missing_required_argument
```

PR #169720, opened 2025-05-30
[iOS] Fix tapping on selection should toggle the toolbar with iOS System Context Menu

Renzo-Olivares:
Fixes an issue with tapping the selection to toggle the toolbar on iOS when using the system context menu. 

Cause: On the framework side we have implemented toggling the selection toolbar when tapping on the current selection. This does not work with the system context menu as seen in the video in the actual results of this issue, tapping the selection hides the toolbar but immediately shows it again after. This happens because first the system dismisses the platform context menu when the tap happens outside the context menu (tap on the selection), the framework then responds to this by hiding the context menu. The framework also then responds to the tap on the selection and tries to toggle the toolbar, this ends up bringing up the toolbar again. The expected behavior would have been for the tap on the selection to hide the toolbar and not reshow it until a subsequent tap.

Fix: Add a toggle debounce that prevents the toolbar from being shown when calling `EditableTextState.toggleToolbar` if it has only just been hidden (within 100ms).

Fixes #168743

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #174890, opened 2025-09-03
Add `showToolbar` and `toggleToolbar` to `TextSelectionDelegate`

Renzo-Olivares:
Currently `TextSelectionDelegate` only provides access to `hideToolbar` through its interface to hide the selection toolbar. `showToolbar` is a method on `TextInputClient`, this PR adds `showToolbar` to the `TextSelectionDelegate` as well as `toggleToolbar`.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/services/text_input.dart
+++ b/packages/flutter/lib/src/services/text_input.dart
@@ -1235,12 +1235,22 @@ mixin TextSelectionDelegate {
   ///   updating the text editing state.
   void userUpdateTextEditingValue(TextEditingValue value, SelectionChangedCause cause);
 
+  /// Shows the text selection toolbar.
+  void showToolbar();
+
   /// Hides the text selection toolbar.
   ///
   /// By default, hideHandles is true, and the toolbar is hidden along with its
   /// handles. If hideHandles is set to false, then the toolbar will be hidden
   /// but the handles will remain.
   void hideToolbar([bool hideHandles = true]);
+
+  /// Toggles the visibility of the toolbar.
+  ///
+  /// If the toolbar is hidden calling this method should show the toolbar.
+  ///
+  /// If the toolbar is visible calling this method should hide the toolbar.
+  void toggleToolbar();
 
   /// Brings the provided [TextPosition] into the visible area of the text
   /// input.
--- a/packages/flutter/lib/src/services/text_input.dart
+++ b/packages/flutter/lib/src/services/text_input.dart
@@ -1247,9 +1247,11 @@ mixin TextSelectionDelegate {
 
   /// Toggles the visibility of the toolbar.
   ///
-  /// If the toolbar is hidden calling this method should show the toolbar.
+  /// If the toolbar is hidden calling this method shows the toolbar.
   ///
-  /// If the toolbar is visible calling this method should hide the toolbar.
+  /// If the toolbar is visible calling this method hides the toolbar.
+  ///
+  /// The handles will remain visible if they were already visible.
   void toggleToolbar();
 
   /// Brings the provided [TextPosition] into the visible area of the text
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -895,11 +895,11 @@ class SelectableRegionState extends State<SelectableRegion>
       case TargetPlatform.fuchsia:
         if (shouldShowSelectionOverlayOnMobile) {
           _showHandles();
-          _showToolbar();
+          showToolbar();
         }
       case TargetPlatform.iOS:
         if (shouldShowSelectionOverlayOnMobile) {
-          _showToolbar();
+          showToolbar();
         }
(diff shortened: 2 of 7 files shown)


Renzo-Olivares:
I wasn't too sure if I should do anything else on the testing side, but happy to add more tests if needed.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1235,13 +1235,23 @@ mixin TextSelectionDelegate {
...
   /// By default, hideHandles is true, and the toolbar is hidden along with its
   /// handles. If hideHandles is set to false, then the toolbar will be hidden
   /// but the handles will remain.
   void hideToolbar([bool hideHandles = true]);
 
+  /// Toggles the visibility of the toolbar.
+  ///
+  /// If the toolbar is hidden calling this method should show the toolbar.
+  ///
+  /// If the toolbar is visible calling this method should hide the toolbar.
```
loic-sharma: Nit: are the "should"s necessary here? Are there common scenarios where it doesn't show/hide the toolbar? If yes, should we call this out? If no, consider simplifying:

```suggestion
  /// If the toolbar is hidden calling this method shows the toolbar.
  ///
  /// If the toolbar is visible calling this method hides the toolbar.
```

I don't feel strongly about this, feel free to skip.

loic-sharma: I'd consider mentioning that the handles will remain visible if they are visible.

loic-sharma (approved):
LGTM, thanks!

packages/flutter/lib/src/services/text_input.dart
Renzo-Olivares: Good call!

Renzo-Olivares: Will simplify, thanks!

justinmc (approved):
LGTM 👍 

Nit: I think I would add a test for SelectableRegionState.showToolbar/toggleToolbar. It's probably covered indirectly already but I feel like it would be valuable to have a test that calls them directly.

Renzo-Olivares:
Sounds good, i'll do that and also do a g3 fix for the google testing. Looks like there was a mock class implementing `TextSelectionDelegate` that needs to add the new methods.

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -1341,12 +1336,24 @@ class SelectableRegionState extends State<SelectableRegion>
 
   /// Shows the text selection toolbar.
   ///
-  /// If the parameter `location` is set, the toolbar will be shown at the
-  /// location. Otherwise, the toolbar location will be calculated based on the
-  /// handles' locations. The `location` is in the coordinates system of the
-  /// [Overlay].
+  /// The toolbar location will be calculated based on the
+  /// location of the selection endpoints.
```
justinmc, line 1340: We discussed offline already, but seeing the code reinforces my thoughts that this is a good approach. I think we're fine to avoid the `location` parameter. I don't want to spend any time supporting the deprecated TextSelectionControls approach.

justinmc (approved):
LGTM 👍

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1235,13 +1235,25 @@ mixin TextSelectionDelegate {
   ///   updating the text editing state.
   void userUpdateTextEditingValue(TextEditingValue value, SelectionChangedCause cause);
 
+  /// Shows the text selection toolbar.
+  void showToolbar();
```
justinmc, line 1239: Another option for avoiding a breaking change here (and for toggleToolbar) might be to add an empty implementation here. I think the question is: do users want to be broken, because if they're mixing in TextSelectionDelegate, then they definitely want to write an implementation for showToolbar/toggleToolbar? Or, do most users that are mixing in TextSelectionDelegate not care that these methods were added?

Renzo-Olivares: I feel like users implementing `TextSelectionDelegate` should already have some method that does the work `showToolbar` does because they already have to abide by the `hideToolbar` contract in `TextSelectionDelegate`. Similarly if they have `showToolbar` and `hideToolbar` implementing `toggleToolbar` should be straight forward.

I think `showToolbar` is reasonable to add and break users for, but `toggleToolbar` might be a little more niche...

The reason I'm exposing `toggleToolbar` is because the plan in this PR https://github.com/flutter/flutter/pull/169720/files#diff-701a97b414026a5c87a06b546bb38ca70be2aa0ce44cbf170f4b43e8defcc91a is to add `toggleDebounceDuration` to `hideToolbar` to fix the issue with toggling the toolbar with a tap on the selection with the iOS system context menu. This requires some documentation that would be difficult to explain without the `toggleToolbar` method as part of the interface.

loic-sharma (approved):
LGTM!

Renzo-Olivares:
Closing this, will re-open when actively working on it.

Issue #174400, opened 2025-08-25
[UIScene] AppLifecycleState


vashworth:
## Problem
The Flutter widgets library provides the [AppLifecycleListener](https://api.flutter.dev/flutter/widgets/AppLifecycleListener-class.html) so that applications can listen to changes in its state. These states are “resumed”, “inactive”, “hidden”, “paused”, and "detached". 
* Detached - The application is still hosted by a Flutter engine but is detached from any host views. When the application is in this state, the engine is running without a view.
* Resumed - this corresponds to the app running in the foreground active state.
* Inactive - At least one view of the application is visible, but none have input focus. The application is otherwise running normally. This state corresponds to the Flutter host view running in the foreground inactive state. Apps transition to this state when: in a phone call, responding to a TouchID request, entering the app switcher or the control center, or the UIViewController hosting the Flutter app is transitioning.
* Hidden - All views of an application are hidden, either because the application is about to be paused
* Paused - The application is not currently visible to the user, and not responding to user input.

Currently, Flutter uses lifecycle events/notifications from the FlutterAppDelegate and FlutterViewController to determine the state and send it to the framework.

## Proposal 1 - AppLifecycleState is per FlutterViewController/FlutterEngine
If the AppLifecycleState is per FlutterViewController/FlutterEngine, it should be determined by view events and scene events.

For example, let’s say an app has 2 scenes, each with a FlutterEngine/FlutterViewController. Under this proposal, if Scene 1 is backgrounded, you’d expect the state of the FlutterEngine in Scene 1 to be “paused” and the state of the FlutterEngine in Scene 2 to be “resumed”. With our current setup, both would be “resumed”.

The following events would need to be replaced:
* applicationBecameActive → sceneDidBecomeActive
* applicationWillResignActive → sceneWillResignActive
* applicationWillTerminate → sceneDidDisconnect
* applicationDidEnterBackground → sceneDidEnterBackground
* applicationWillEnterForeground → sceneWillEnterForeground

Pros:
* Essentially AppLifecycleState would act more as if it was “EngineLifecycleState” or “SceneLifecycleState”
* Less complex to implement

Cons:
* Not inline with the future of multi-view (see “Multi-View Consideration” below).
* May require changes in future if multi-view is adopted

## Proposal 2 - AppLifecycleState is per Application
[...]

PR #174910, opened 2025-09-03
Connect the FlutterEngine to the FlutterSceneDelegate

### LINKED ISSUE #174395: [UIScene] Associate the `FlutterEngine` with `FlutterSceneDelegate`
## Problem
Flutter does certain things in response to lifecycle events. After migrating to scenes, application lifecycle events may not work the same as before. Therefore, we need to migrate to scene lifecycle events. However, we need to consider that there could be multiple scene, which could have multiple engines. As such, we should track engines that are associated with the scene.

## Proposal

When a `FlutterViewController` is added to the scene, we should add its `FlutterEngine` to the `FlutterSceneDelegate`.

`FlutterSceneDelegate.mm`:

```
_engines = [NSPointerArray weakObjectsPointerArray];

...

- (void)addFlutterViewController:(FlutterViewController*)controller {
  NSLog(@"Engine added to scene");

  // NSPointerArray is clever and assumes that unless a mutation operation has occurred on it that
  // has set one of its values to nil, nothing could have changed and it can skip compaction.
  // That's reasonable behaviour on a regular NSPointerArray but not for a weakObjectPointerArray.
  // As a workaround, we mutate it first. See: http://www.openradar.me/15396578
  [self.engines addPointer:nil];
  [self.engines compact];

  // Check if the engine is already in the array to avoid duplicates.
  if (![self.engines.allObjects containsObject:controller.engine]) {
    [self.engines addPointer:(__bridge void*)controller.engine];
  }

  [controller.engine.sceneLifeCycleDelegate flutterViewController:controller
                                                didConnectToScene:(UIScene*)self
                                                          options:self.sceneConnectionOptions];
}
```

`FlutterViewController`:
```
- (void)viewIsAppearing:(BOOL)animated {
  FML_LOG(ERROR) << "viewIsAppearing";
  [self connectFlutterViewControllerWithScene];
  [super viewIsAppearing:animated];
}

- (void)connectFlutterViewControllerWithScene {
  // The scene is not available until viewIsAppearing
  UIWindowScene* scene = self.view.window.windowScene;

  // self.view.window.windowScene
  if ([scene.delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleProvider)]) {
    id<FlutterSceneLifeCycleProvider> lifeCycleProvider =
        (id<FlutterSceneLifeCycleProvider>)scene.delegate;
    [lifeCycleProvider addFlutterViewController:self];
  }
}
```

See [prototype](https://github.com/flutter/flutter/compare/master...vashworth:flutter:uiscene_lifecycle_wip).

vashworth:
Once the `FlutterViewController` is able to get its `UIWindowScene`, pass itself to the `FlutterSceneDelegate` through the `FlutterSceneLifeCycleProvider` protocol. The scene will store the `FlutterEngine` in a weak pointer array.

This is an incremental change towards https://github.com/flutter/flutter/issues/174398.

Fixes https://github.com/flutter/flutter/issues/174395.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneDelegate.m
```diff
@@ -6,12 +6,25 @@
 
 #import "flutter/shell/platform/darwin/common/framework/Headers/FlutterMacros.h"
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterAppDelegate_Internal.h"
+#import "flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.h"
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterSharedApplication.h"
 
 FLUTTER_ASSERT_ARC
 
+@interface FlutterSceneDelegate () <FlutterSceneLifeCycleProvider>
```
vashworth, line 14: `FlutterSceneLifeCycleProvider` is a private protocol for now, but will later become public. It's purpose is to allow add-to-app developers who can't adopt the FlutterSceneDelegate to still be able to use plugin scene lifecycle events. 

See app equivalent for example of future use case: https://docs.flutter.dev/add-to-app/ios/add-flutter-screen#if-you-cant-directly-make-flutterappdelegate-a-subclass

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneDelegate.m
```diff
@@ -6,12 +6,25 @@
...
+@end
+
 @implementation FlutterSceneDelegate
 
+@synthesize sceneLifeCycleDelegate = _sceneLifeCycleDelegate;
+
+- (instancetype)init {
+  if (self = [super init]) {
+    _sceneLifeCycleDelegate = [[FlutterPluginSceneLifeCycleDelegate alloc] init];
```
vashworth, line 23: Rather than storing the engines directly in the `FlutterSceneDelegate`, we store them in a `FlutterPluginSceneLifeCycleDelegate` object,

This is to handle the add-to-app case when they can't adopt the `FlutterSceneDelegate`. This way when the `FlutterViewController` connects, it can use the `FlutterSceneLifeCycleProvider` to access the `FlutterPluginSceneLifeCycleDelegate` to store the engines.

See app equivalent for example of future use case: https://docs.flutter.dev/add-to-app/ios/add-flutter-screen#if-you-cant-directly-make-flutterappdelegate-a-subclass

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
```diff
@@ -0,0 +1,38 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.h"
+#import "flutter/shell/platform/darwin/common/framework/Headers/FlutterMacros.h"
+
+FLUTTER_ASSERT_ARC
```
LongCatIsLooong, line 8: I thought there was no MRC code left in the ios embedder so we don't need this any more?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
```diff
@@ -0,0 +1,38 @@
...
+
+@interface FlutterPluginSceneLifeCycleDelegate () {
```
LongCatIsLooong: hmm my objective-c is rusty, what are the curly brackets for?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -906,6 +907,21 @@ - (void)viewWillAppear:(BOOL)animated {
   [super viewWillAppear:animated];
 }
 
+- (void)viewIsAppearing:(BOOL)animated {
+  TRACE_EVENT0("flutter", "viewIsAppearing");
+  if (self.engine.viewController == self) {
+    // The earliest the view can access the scene is in viewIsAppearing.
```
LongCatIsLooong, line 913: It looks like the `UIWindowScene` first becomes accessible to the view in `[UIView didMoveToWindow]` (when establishing the view hierarchy), and `viewIsAppearing` is called by the layout process so it's a bit later than that:
```
  * frame #0: 0x000000018545a010 UIKitCore`-[UIViewController viewIsAppearing:]
    frame #1: 0x0000000185449ab8 UIKitCore`-[UIViewController __viewIsAppearing:skipWindowCheck:] + 188
    frame #2: 0x0000000186142cd4 UIKitCore`-[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 1072
```
Not sure if that makes a difference tho.

LongCatIsLooong: It looks like the `UIWindowScene` first becomes accessible to the view in `[UIView didMoveToWindow]` (when establishing the view hierarchy. `UIWindow`'s initializer takes a `UIWindowScene`), and `viewIsAppearing` is called by the layout process so it's a bit later than that:
```
  * frame #0: 0x000000018545a010 UIKitCore`-[UIViewController viewIsAppearing:]
    frame #1: 0x0000000185449ab8 UIKitCore`-[UIViewController __viewIsAppearing:skipWindowCheck:] + 188
    frame #2: 0x0000000186142cd4 UIKitCore`-[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 1072
```
Not sure if that makes a difference tho.

LongCatIsLooong: The [windowScene](https://developer.apple.com/documentation/uikit/uiwindow/windowscene) property is not readonly I guess that means a `UIWindow` can be moved to a different scene while the app is running? I think that is something that can happen in an add-to-app scenario? Also the viewController associated with the engine can be changed too (https://api.flutter.dev/ios-embedder/interface_flutter_engine.html#a9247723288f07f7f7cc6d85fb701fbdc), so the engine pointers associated with a scene may also need to be updated on those events?

LongCatIsLooong: Given that the engine instance associated with a FlutterViewController can change, would it be better to change the addViewController API (currently private) to addEngine, also do we need a `deleteEngine` method if, say, two FlutterViewControllers with different `UIWindowScene`s can swap their engines?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
vashworth: It appears that any builds that use the `flutter_cflags_objc` config should have ARC enabled, but that's not very clear from a glance. If someone were to add a new target or edit the configs, they may not know they need to do that. Due to that, I think it's safer to keep using `FLUTTER_ASSERT_ARC`.

https://github.com/flutter/flutter/blob/e25651bfb938e7f1591a77174e914d0bd5222d55/engine/src/flutter/common/config.gni#L71-L75

https://github.com/flutter/flutter/blob/e25651bfb938e7f1591a77174e914d0bd5222d55/engine/src/flutter/shell/platform/darwin/ios/BUILD.gn#L62-L210

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
vashworth: Nothing, I just forgot to remove them!

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
vashworth: This is great feedback! Thanks for letting me know about `didMoveToWindow`, I didn't know about that!

In regards to the viewController moving engines - this will be handled by moving to `didMoveToWindow` actually! I also added a `removeFlutterEngine` method.

In regards to UIWindow moving scenes, there doesn't appear to be an API to detect when this happens. I added a comment with more of a explanation here: https://github.com/flutter/flutter/blob/bab8f251cc96f612b57aee1c9d8c18f5c1349ece/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm#L66-L74

LongCatIsLooong: Ah the `-[UIWindow setWindowScene:]` setter posts a `_UIWindow{Will, Did}MoveToSceneNotification` notification and calls `-[UIWindow _didMoveFromScene:toScene:]` :(

The `addFlutterEngine` and `removeFlutterEngine` API won't be visible to app developers so they can't call these APIs in `UIWindow.windowScene.didSet` to manage the scene->engine mapping themselves right?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -241,4 +244,28 @@ - (BOOL)isAccessibilityElement {
...
+  // When a FlutterView moves windows, it may also be moving scenes. Add/remove the FlutterEngine
+  // from the sceneLifeCycleDelegate.
+  UIWindowScene* scene = newWindow.windowScene;
+  if (scene) {
+    if ([scene.delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleProvider)]) {
+      id<FlutterSceneLifeCycleProvider> lifeCycleProvider =
+          (id<FlutterSceneLifeCycleProvider>)scene.delegate;
+      [lifeCycleProvider.sceneLifeCycleDelegate addFlutterEngine:(FlutterEngine*)self.delegate];
+    }
+    _previousScene = scene;
```
LongCatIsLooong: Is it possible to remove the additional `_previousScene` ivar? UIKit already keeps track of that I think (_previousScene === window?.windowScene)?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -241,4 +244,28 @@ - (BOOL)isAccessibilityElement {
...
+  } else if ([_previousScene.delegate
```
LongCatIsLooong: When you move from one window to another both `add` and `remove` should be called right? The `else if` will allow only one scene to be notified.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
vashworth: > The `addFlutterEngine` and `removeFlutterEngine` API won't be visible to app developers so they can't call these APIs in `UIWindow.windowScene.didSet` to manage the scene->engine mapping themselves right?

Correct, these are private APIs

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -241,4 +242,30 @@ - (BOOL)isAccessibilityElement {
   return nil;
 }
 
+- (void)willMoveToWindow:(UIWindow*)newWindow {
+  // When a FlutterView moves windows, it may also be moving scenes. Add/remove the FlutterEngine
+  // from the FlutterSceneLifeCycleProvider.sceneLifeCycleDelegate if it changes scenes.
+  UIWindowScene* newScene = newWindow.windowScene;
+  UIWindowScene* previousScene = self.window.windowScene;
+  if (newScene) {
```
LongCatIsLooong: nit: I think this is redundant? (if `newScene` is `nil` the expression in the next if condition evaluates to `nil` too?)

LongCatIsLooong: Also: maybe return early if `newScene == previousScene` since the scene -> engine mapping would remain the same?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
```diff
@@ -0,0 +1,89 @@
...
+}
+
+- (void)removeFlutterEngine:(FlutterEngine*)engine {
+  NSUInteger index = [self.engines.allObjects indexOfObject:engine];
+  if (index != NSNotFound) {
+    [self.engines removePointerAtIndex:index];
+  }
+}
+
+- (void)updateEnginesInScene:(UIScene*)scene {
```
LongCatIsLooong, line 52: I don't see this is being called anywhere. Is this for future-use only or this should be called in the engines getter?

vashworth: This is for future-use

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterView.mm
vashworth: Done!

LongCatIsLooong (commented):
The implementation LGTM. 

However given the potential caveat (it looks like we can't reliably maintain the scene -> engine mapping since currently the window did change scene notification is private), I think it would probably be better if we keep week refs to all engines in an application instead, and determine which scene each engine belongs to when dispatching lifecycle event callbacks? It will be slower since lifecycle event dispatching is much more frequent than window changing scenes but realistically people probably won't have 1000 engines in the same app such that the performance difference becomes noticeable.

vashworth:
> The implementation LGTM.
> 
> However given the potential caveat (it looks like we can't reliably maintain the scene -> engine mapping since currently the window did change scene notification is private), I think it would probably be better if we keep week refs to all engines in an application instead, and determine which scene each engine belongs to when dispatching lifecycle event callbacks? It will be slower since lifecycle event dispatching is much more frequent than window changing scenes but realistically people probably won't have 1000 engines in the same app such that the performance difference becomes noticeable.

Please see my comment:

https://github.com/flutter/flutter/blob/bd8c4a5b9eb70c46470a28f96bc5a23c34ae5cd9/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm#L66-L74

I think this is an acceptable limitation for now, given that we don't explicitly support multi-scene.

> I think it would probably be better if we keep week refs to all engines in an application instead

Unfortunately, we can't rely on the [application singleton](https://developer.apple.com/documentation/uikit/uiapplication/shared?language=objc) to be available. For example, in iOS app extensions, the application is not accessible.

LongCatIsLooong:
> Unfortunately, we can't rely on the [application singleton](https://developer.apple.com/documentation/uikit/uiapplication/shared?language=objc) to be available. For example, in iOS app extensions, the application is not accessible.

Ah interesting. But even without access to an UIApplication instance, we can still use a class/global variable I think? We might need to create a global list of engines for proposal 2 in https://github.com/flutter/flutter/issues/174400 so I guess we'll come back and visit this later since everything is implementation details for now.

EDIT: oh you did mention in that issue:

> To access other FlutterViewControllers and FlutterEngines, we can [iterate through the Scenes](https://developer.apple.com/documentation/uikit/uiapplication/connectedscenes?language=objc) and iterate through each FlutterEngine they have connected (see Proposal 1of “Plugin lifecycle events”).

LongCatIsLooong (approved):

PR #174957, opened 2025-09-04
[web] Reuse chrome instance to run all flutter tests

mdebbar:
Web test shards take a significant time to run in CI compared to their VM counterparts. One theory is that launching and shutting down Chrome repeatedly is a major part of the slowness.

This PR makes `flutter test --platform=chrome` reuse the same Chrome instance for all test files instead of launching a new one for each test file.

At first glance, it may seem like we aren't shutting down Chrome anymore, but we actually do in `closeEphemeral` and `close` methods.

Thanks @eyebrowsoffire for the idea!


eyebrowsoffire (approved):
LGTM

PR #174963, opened 2025-09-04
[web] Minor simplification in flutter.js loader

mdebbar:


eyebrowsoffire (approved):
Can you check if you can remove the `buildContainsRenderer` function altogether? I think it was only used for this code path, which was for the "auto" html+canvaskit renderer, which is no longer a thing.

mdebbar:
> Can you check if you can remove the `buildContainsRenderer` function altogether?

Yes, that's what the PR is doing.

eyebrowsoffire:
> > Can you check if you can remove the `buildContainsRenderer` function altogether?
> 
> 
> 
> Yes, that's what the PR is doing.
> 
> 
> 
> 

Lol, sorry, I thought it was being imported from another file. LGTM!

PR #174981, opened 2025-09-05
[a11y] Add `expanded` flag support to Android.

### LINKED ISSUE #92040: [a11y] Add expanded state in Semantics widget on Android
## Use case

I cannot set expanded or collapsed states for screen readers. This accessibility issue is WCAG's conformance level A, Criterion 4.1.2: Name, Role, Value
WCAG: https://www.w3.org/WAI/WCAG21/Understanding/name-role-value

I would like the `Semantics` widget to have an `expanded` parameter so that it announces the state for VoiceOver and TalkBack.

After creating this parameter in `Semantics`, I recommend putting it in the `Material's Accordion` and similar components.

## Proposal

Analyze the behavior of the `accessibilityState={{expanded: true}}` parameter of React Native and do the same.

https://github.com/facebook/react-native/search?q=expanded

## Extra

W3C Documentation about HTML's aria-expanded: https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element

React Native documentation about accessibilityState: https://reactnative.dev/docs/accessibility#accessibilitystate

Example of accordion accessible in react material design: https://mui.com/pt/components/accordion/

ksokolovskyi:
Closes https://github.com/flutter/flutter/issues/92040

- Adds `expanded` semantics flag support to Android
- Adds `onExpand` and `onCollapse` semantics actions
- Updates `robolectric` library
- Adds java and dart tests

#### Why were `onExpand` and `onCollapse` actions added?
It turned out that TalkBack doesn't announce the `expanded` state if `expand/collapse` action is not set for the accessibility node.

#### Why was the `robolectric` library updated?

The `expanded` state support in Android was introduced in API 36. The `roboelectric: 4.14.1` doesn't support API 36. To run tests for a newly added functionality `roboelectric` library was updated to `4.16`, which supports the latest Android version (https://github.com/robolectric/robolectric/releases/tag/robolectric-4.16).

In case you think it would be better to update the `roboelectric` in a separate PR, please let me know.

<br/>
<details>
<summary>Example Source Code</summary>

```dart

void main() {
  runApp(const App());
}

class App extends StatefulWidget {
  const App({super.key});

  @override
  State<App> createState() => _AppState();
}

class _AppState extends State<App> {
  final _controller = ExpansibleController();

  @override
  void dispose() {
    _controller.dispose();
    super.dispose();
  }

  @override
    return MaterialApp(
      home: Scaffold(
        body: Column(
          spacing: 24,
          children: [
            Text('Expansible Example'),
            ListenableBuilder(
              listenable: _controller,
              builder: (context, child) {
                return Semantics(
                  expanded: _controller.isExpanded,
                  onExpand: () {
                    print(' \n onExpand \n ');
                    _controller.expand();
                  },
                  onCollapse: () {
                    print(' \n onCollapse \n ');
                    _controller.collapse();
                  },
                  child: child,
                );
              },
              child: Expansible(
                headerBuilder: (context, _) => ListTile(
                  tileColor: Colors.blue.shade100,
                  leading: Text(
                    'Expansible',
                    style: TextStyle(fontSize: 20),
                  ),
                  trailing: Icon(
                    _controller.isExpanded
                        ? Icons.arrow_upward
                        : Icons.arrow_downward,
                    semanticLabel: _controller.isExpanded
                        ? 'Arrow Up icon'
                        : 'Arrow Down icon',
                  ),
                ),
                bodyBuilder: (context, _) {
                  return Container(
                    color: Colors.blue,
                    height: 200,
                    width: 200,
                  );
                },
                controller: _controller,
              ),
            ),
          ],
        ),
      ),
    );
  }
}
```

</details>

https://github.com/user-attachments/assets/256c4182-a1e3-44fc-b028-5e6c9ec05ad7

## Pre-launch Checklist

- [X] I signed the [CLA].


ash2moon (approved):

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -2116,6 +2116,139 @@ public void itAddsClickActionToSliderNodeInfo() {
...
+
+    node = new TestSemanticsNode();
+    node.addFlag(AccessibilityBridge.Flag.HAS_EXPANDED_STATE);
+    node.addAction(AccessibilityBridge.Action.EXPAND);
+    testSemanticsUpdate = node.toUpdate();
+    testSemanticsUpdate.sendUpdateToBridge(accessibilityBridge);
+
+    nodeInfo = accessibilityBridge.createAccessibilityNodeInfo(0);
+    actions = nodeInfo.getActionList();
+    assertTrue(actions.contains(AccessibilityNodeInfo.AccessibilityAction.ACTION_EXPAND));
```
chunhtai, line 2184: Can you mock the a11y channel and add 
```
accessibilityBridge.performAction(1, AccessibilityNodeInfo.ACTION_EXPAND, bundle);
    verify(mockChannel)
        .dispatchSemanticsAction(<id>, AccessibilityBridge.Action.ACTION_EXPAND);
```
to verify it hooks up the callback? there should be a bunch of example in this file

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -2116,6 +2116,139 @@ public void itAddsClickActionToSliderNodeInfo() {
...
+    node = new TestSemanticsNode();
+    node.addFlag(AccessibilityBridge.Flag.HAS_EXPANDED_STATE);
+    node.addFlag(AccessibilityBridge.Flag.IS_EXPANDED);
+    node.addAction(AccessibilityBridge.Action.COLLAPSE);
+    testSemanticsUpdate = node.toUpdate();
+    testSemanticsUpdate.sendUpdateToBridge(accessibilityBridge);
+
+    nodeInfo = accessibilityBridge.createAccessibilityNodeInfo(0);
+    actions = nodeInfo.getActionList();
+    assertTrue(actions.contains(AccessibilityNodeInfo.AccessibilityAction.ACTION_COLLAPSE));
```
chunhtai, line 2244: same here

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -2116,6 +2116,139 @@ public void itAddsClickActionToSliderNodeInfo() {
...
+  }
+
+  @Config(sdk = API_LEVELS.API_36)
+  @TargetApi(API_LEVELS.API_36)
+  @Test
+  public void itDoesNotAddExpandAndCollapseActionsAtTheSameTime() {
```
chunhtai: shouldn't this be consider a malformed semantics node that if framework passes both expand and collapse?

chunhtai (dismissed):
LGTM

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
ksokolovskyi: Thanks for your review!

I am not sure what the proper way of passing `onExpand` and `onCollapse` from framework to engine is. As far as I can say, there are two options:
1) user passes both `onExpand` and `onCollapse`, and the engine decides which action to add depending on the `expanded` property
2) user passes only one callback `onExpand` or `onCollapse` depending on `expanded` property, and the  engine just adds action without checks

This PR implements option 1. The user can pass both callbacks, and  `AccessibilityBridge` will add the appropriate action according to the `expanded` property state.

Could you please share your thoughts regarding which option is better aligned with the current a11y implementation?

ksokolovskyi:
`[Linux linux_unopt]` fails because there is error happening when running Android tests with updated `roboelectric` library: 
`java.lang.UnsupportedOperationException: Failed to create a Robolectric sandbox: Android SDK 36 requires Java 21 (have Java 17)`.

@ash2moon, could you please help here, as I'm unsure how to resolve this issue on CI/CD?

robert-ancell (approved):
Reviewing how this will work on Linux.

The a11y framework on Linux (AT-SPI) supports named actions, so this should be translatable there. The nearest widget I could find that was similar (GtkExpander) uses the button role and thus I didn't find any existing cases where this action was used. GTK4 uses AccessKit (Flutter is using GTK3 which uses ATK), which does support the expand/collapse actions. So this should be implementable in Linux.

gmackall:
From android triage: Thanks for the contribution, we will get the java updated to unblock this PR
https://github.com/flutter/flutter/issues/175454

mdebbar (approved):
web changes look good!

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
chunhtai: We should go with (2) that is what we have being doing for all other things like increase/decrease and scrolling left/right/up/down

chunhtai (commented):
I left a comment in https://github.com/flutter/flutter/pull/174981#discussion_r2363732184

and also we should add a new API can we add some runtime check if people misused the property?

A good place to add it is https://github.com/flutter/flutter/blob/f930c8105beb988c81cf12a80feb36674ca51341/packages/flutter/lib/src/semantics/semantics.dart#L123

Another question, can this be wire up to Expansible widget?

ksokolovskyi:
@chunhtai, thanks a lot for your comment!

> and also we should add a new API, can we add some runtime check if people misuse the property?
> A good place to add it is [flutter/packages/flutter/lib/src/semantics/semantics.dart](https://github.com/flutter/flutter/blob/f930c8105beb988c81cf12a80feb36674ca51341/packages/flutter/lib/src/semantics/semantics.dart#L123)

I agree that we have to add additional checks to make sure that users don't mess up with the new API. However, `_checkSemanticsData` is currently only focused on `role` specific checks. Would you like me to add additional checks for a new API after `role` specific checks?

> Another question, can this be wire up to Expansible widget?

We can definitely do this, check it [here](https://github.com/flutter/flutter/pull/174981#:~:text=let%20me%20know.-,Example%20Source%20Code).
But we need to additionally verify how widgets that already use `Expansible` under the hood and add their own semantics will behave on Android, to ensure that there are no duplicates in labels (e.g., "Expanded Expanded Dropdown Button", etc.)

chunhtai:
> Would you like me to add additional checks for a new API after role specific checks?

I was suggesting just add a one more step somewhere in this semantics check call loop for generic check like expand/collapse property and actions

chunhtai:
> But we need to additionally verify how widgets that already use Expansible under the hood and add their own semantics will behave on Android, to ensure that there are no duplicates in labels (e.g., "Expanded Expanded Dropdown Button", etc.)

Do we have an issue for this follow up? I imagine we would want to do this.

ksokolovskyi:
> > But we need to additionally verify how widgets that already use Expansible under the hood and add their own semantics will behave on Android, to ensure that there are no duplicates in labels (e.g., "Expanded Expanded Dropdown Button", etc.)
> 
> Do we have an issue for this follow up? I imagine we would want to do this.

I haven't seen such an issue. I will create one.

ash2moon:
> `[Linux linux_unopt]` fails because there is error happening when running Android tests with updated `roboelectric` library: `java.lang.UnsupportedOperationException: Failed to create a Robolectric sandbox: Android SDK 36 requires Java 21 (have Java 17)`.
> 
> @ash2moon, could you please help here, as I'm unsure how to resolve this issue on CI/CD?

Just pushed the update! I went ahead and pulled the changes into your branch to see if it resolves your issues.

ksokolovskyi:
@ash2moon, thanks a lot for your update, everything works well now!

ksokolovskyi:
@chunhtai, could you please take another look at this PR when you have time?

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -120,43 +120,57 @@ final int _kUnblockedUserActions =
...
+      return error;
+    }
+
+    final SemanticsData data = node.getSemanticsData();
+
+    if (data.flagsCollection.isExpanded != Tristate.none) {
+      error = _semanticsExpandable(node, data);
+    }
+
+    return error ?? _noCheckRequired(node);
```
chunhtai: _noCheckRequired returns null. no need to call it again
```suggestion
    return error;
```

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -120,43 +120,57 @@ final int _kUnblockedUserActions =
...
+      SemanticsRole.loadingSpinner => _unimplemented,
+      SemanticsRole.progressBar => _unimplemented,
+      SemanticsRole.hotKey => _unimplemented,
+    }(node);
+
+    if (error != null) {
+      return error;
+    }
+
+    final SemanticsData data = node.getSemanticsData();
```
chunhtai: Can you refactor this part to something like _semanticsGeneral

chunhtai (commented):
LGTM, just some code style nits

chunhtai (approved):
LGTM, thanks for working on this!

ksokolovskyi:
Thank you all for your reviews!

PR #175001, opened 2025-09-05
Run skwasm framework unit tests on a single shard.

eyebrowsoffire:
This PR puts all the skwasm framework unit tests onto a single shard. The time taken to run the actual tests on each shard was roughly 90-180 seconds, so a large portion of the work done on each shard was just setup/teardown. I'm trying this to see if they might all just fit on a single shard, like the VM unit tests do.


Piinks:

Piinks:
Welcome back to stale PR triage - do you have plans to move this change forward?

eyebrowsoffire:
I think Mouad already did some consolidation of these, so this can just be closed now.

PR #175025, opened 2025-09-06
Fix: use 'exit' instead of 'return' for curlExitCode

moluopro:
When updating Flutter using `flutter upgrade`, the following message appeared:
```
curl: (18) Transferred a partial file
/opt/flutter/bin/internal/update_dart_sdk.sh: 
158: return: can only `return' from a function or sourced script
```
An invalid syntax was used here(Line 158 of `flutter/bin/internal/update_dart_sdk.sh`):
```
if [ $curlExitCode != 33 ]; then
   return $curlExitCode
fi
```
I applied a simple fix: use `exit` instead of `return` for curlExitCode.

## Pre-launch Checklist

- [x] I signed the [CLA].


Piinks:
Hi @moluopro it looks like this PR fell through the review cracks, sorry about that. Would you still like to move forward with this change?

PR #175047, opened 2025-09-08
Feat: make tooltip position customizeable

### LINKED ISSUE #172758: Make tooltip positioning more customizable
### Use case

The current Tooltip implementation doesn't support positioning on the left or right of the child widget. I ran into this limitation while adding tooltips to destinations in my custom navigation rail.

Similar case:

https://stackoverflow.com/questions/72202530/positioning-tooltip-in-flutter

Margin is not useful because it overlays the child widget, as mentioned in the comments to the most voted answer.

### Proposal

Tooltip positioning should be considered in horizontal/vertical axis with additional horizontal/vertical offset.

Image

To achieve this result I've edited getPositionForChild in _TooltipPositionDelegate class.

`
  @override
  Offset getPositionForChild(Size size, Size childSize) {
    if (axis == Axis.vertical) {
      return positionDependentBoxVertically(
        size: size,
        childSize: childSize,
        target: target,
        horizontalOffset: horizontalOffset,
        verticalOffset: verticalOffset,
        preferBelow: preferBelow,
      );
    } else {
      return positionDependentBoxHorizontally(
        size: size,
        childSize: childSize,
        target: target,
        horizontalOffset: horizontalOffset,
        verticalOffset: verticalOffset,
        preferRight: preferRight,
      );
    }
  }
`

The "preferRight" parameter behaves similarly to "preferBelow". Its effect may depend on the current TextDirection, but it's unclear whether this could cause confusion.

rkishan516:
This MR adds positionDelegate argument to tooltip which allows to change tooltip position.
fixes: #172758

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/tooltip.dart
+++ b/packages/flutter/lib/src/material/tooltip.dart
@@ -25,6 +25,25 @@ import 'tooltip_visibility.dart';
 /// Signature for when a tooltip is triggered.
 typedef TooltipTriggeredCallback = void Function();
 
+/// Signature for computing the position of a tooltip.
+///
+/// The arguments are:
+/// * `target`: The center point of the target widget in the global coordinate system.
+/// * `targetSize`: The size of the target widget that triggers the tooltip.
+/// * `childSize`: The size of the tooltip itself.
+/// * `verticalOffset`: The configured vertical offset.
+/// * `preferBelow`: Whether the tooltip prefers to be positioned below the target.
+///
+/// Returns the offset from the top left of the overlay to the top left of the tooltip.
+typedef TooltipPositionDelegate =
+    Offset Function({
+      required Offset target,
+      required Size targetSize,
+      required Size childSize,
+      required double verticalOffset,
+      required bool preferBelow,
+    });
+
 /// A special [MouseRegion] that when nested, only the first [_ExclusiveMouseRegion]
 /// to be hit in hit-testing order will be added to the BoxHitTestResult (i.e.,
 /// child over parent, last sibling over first sibling).
@@ -192,6 +211,7 @@ class Tooltip extends StatefulWidget {
     this.onTriggered,
     this.mouseCursor,
     this.ignorePointer,
+    this.positionDelegate,
     this.child,
   }) : assert(
          (message == null) != (richMessage == null),
@@ -392,6 +412,17 @@ class Tooltip extends StatefulWidget {
   /// handled or ignored.
   final bool? ignorePointer;
 
+  /// A custom position delegate function for computing where the tooltip should be positioned.
+  ///
+  /// If provided, this function will be called instead of the default positioning logic
+  /// to determine where to place the tooltip relative to its target.
+  ///
+  /// This allows for custom positioning such as left/right positioning, or any other
+  /// arbitrary positioning logic.
+  ///
+  /// If null, the default positioning behavior is used (above or below the target).
+  final TooltipPositionDelegate? positionDelegate;
+
   static final List<TooltipState> _openedTooltips = <TooltipState>[];
 
   /// Dismiss all of the tooltips that are currently shown on the screen,
@@ -469,6 +500,13 @@ class Tooltip extends StatefulWidget {
       FlagProperty('enableFeedback', value: enableFeedback, ifTrue: 'true', showName: true),
     );
     properties.add(DiagnosticsProperty<TextAlign>('textAlign', textAlign, defaultValue: null));
+    properties.add(
+      DiagnosticsProperty<TooltipPositionDelegate>(
+        'positionDelegate',
+        positionDelegate,
+        defaultValue: null,
+      ),
+    );
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -25,6 +25,24 @@ import 'tooltip_visibility.dart';
 /// Signature for when a tooltip is triggered.
 typedef TooltipTriggeredCallback = void Function();
 
+/// Signature for computing the position of a tooltip.
+///
+/// The arguments are:
+/// * `target`: The offset of the target the tooltip is positioned near in the global coordinate system.
```
QuncCccccc, line 31: Is this offset the center of the target? If so, we might want to clarify it in documentation.

QuncCccccc: Is this offset the center of the target? If so, we might want to clarify it in documentation.

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -25,6 +25,24 @@ import 'tooltip_visibility.dart';
...
+/// * `size`: The size of the overlay in which the tooltip is being positioned.
```
QuncCccccc: Not sure if this would be helpful for users to calculate the offset. It's fine to me. Can we add one more parameter like "targetSize" to this callback? I think it would be helpful to calculate the tooltip position. For example, if we want the tooltip to show up on the upper right corner of the target, this parameter would be useful. WDYT:)? I think the target size info can be found in `OverlayChildLayoutInfo`.

QuncCccccc (commented):
Overall looks good! Thanks for your contribution:)!

QuncCccccc (commented):
The linux analyzer shows some format issue. To fix, run `dart format packages/flutter/lib/src/material/tooltip.dart`

packages/flutter/lib/src/material/tooltip.dart
rkishan516: Updated the documentation to show, target is center of triggering widget.

packages/flutter/lib/src/material/tooltip.dart
rkishan516: Yes, I don't think size will be useful, so replacing it with targetSize which as you said will help in aligning tooltip to different required position. Also updating test, such that it shows as an example for aligning to top right.

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -25,6 +25,25 @@ import 'tooltip_visibility.dart';
 /// Signature for when a tooltip is triggered.
 typedef TooltipTriggeredCallback = void Function();
 
+/// Signature for computing the position of a tooltip.
+///
+/// The arguments are:
+/// * `target`: The center point of the target widget in the global coordinate system.
+/// * `targetSize`: The size of the target widget that triggers the tooltip.
+/// * `childSize`: The size of the tooltip itself.
```
QuncCccccc: how about just `tooltipSize`?

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -823,6 +861,7 @@ class TooltipState extends State<Tooltip> with SingleTickerProviderStateMixin {
       layoutInfo.childPaintTransform,
       layoutInfo.childSize.center(Offset.zero),
     );
+    final Size targetSize = layoutInfo.childSize;
```
QuncCccccc: just nit:
```suggestion
    final Size tooltipSize = layoutInfo.childSize;
```

QuncCccccc (approved):
LGTM with nits:) Thank you for your contribution!

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -25,6 +25,25 @@ import 'tooltip_visibility.dart';
...
+///
+/// The arguments are:
+/// * `target`: The center point of the target widget in the global coordinate system.
+/// * `targetSize`: The size of the target widget that triggers the tooltip.
+/// * `tooltipSize`: The size of the tooltip itself.
+/// * `verticalOffset`: The configured vertical offset.
+/// * `preferBelow`: Whether the tooltip prefers to be positioned below the target.
+///
+/// Returns the offset from the top left of the overlay to the top left of the tooltip.
+typedef TooltipPositionDelegate =
```
dkwingsmt: Do you think we might encounter new properties for this delegate in the future? Extending this parameter list will be breaking. What do you think of grouping them into a data class for extensibility? We might call it `TooltipPositionContext` and the delegate will just take one parameter.

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -25,6 +25,25 @@ import 'tooltip_visibility.dart';
...
+    Offset Function({
+      required Offset target,
+      required Size targetSize,
+      required Size tooltipSize,
+      required double verticalOffset,
+      required bool preferBelow,
```
dkwingsmt: I wonder if we can remove the values from `Tooltip`'s parameter list. If the app provides a delegate they would know all information needed for this algorithm anyway.

dkwingsmt (commented):
Generally looking good.

packages/flutter/lib/src/material/tooltip.dart
rkishan516: Yup, makes sense. Let me update this.

packages/flutter/lib/src/material/tooltip.dart
rkishan516: I don't know both verticalOffset and preferBelow are used when delegate is not given. So, I don't think we can remove these.

dkwingsmt (approved):
LGTM

PR #175091, opened 2025-09-08
Fix `Select All` button behavior on iOS System Context Menu

### LINKED ISSUE #168857: [iOS] Tapping "Select All" when using system context menu does not show selection handles
### Steps to reproduce

.

### Expected results

.

### Actual results

.

### Code sample

```dart
[Paste your code here]
```.

</details>


### Screenshots or Video

<details open>
<summary>Screenshot.s / Video demonstration</summary>

[Upload media here]

</details>


### Logs

```console
[Paste your logs here]
```

</details>.

### Flutter Doctor output

```console
[Paste your output here]
```

</details>.

Renzo-Olivares:
https://github.com/user-attachments/assets/ba9292fb-595e-485a-af06-9b5b2b964293

Fixes #168857

Issue:

When using the system context menu on iOS and pressing the `Select All` button, the context menu should remain visible and should update its options based on the new selection, and the selection handles should remain visible. Instead when pressing `Select All` the context menu disappears along with the selection handles. 

Cause of disappearing menu: By default the system will close the menu when an item is pressed.
Cause of disappearing handles: Instead of going through `EditableTextState.selectAll`, the `Select All` from the system context menu goes directly through `EditableTextState.updateEditingValue` and an accurate `SelectionChangedCause` cannot be inferred as a result causing the handles to disappear when the update is processed.

Fix:

This change exposes `attributes` on `IOSSystemContextMenuItemCustom` and `IOSSystemContextMenuItemDataCustom`. One can set multiple `IOSSystemContextMenuItemCustomAttributes` on a context menu button, for example `attributes: IOSSystemContextMenuItemCustomAttributes.destructive | IOSSystemContextMenuItemCustomAttributes.keepsMenuPresented`, for a button that has red text and keeps the menu visible when it is pressed. These attributes are used by the underlying `UIAction` to build an action that has the given attributes set as its [`attributes`](https://developer.apple.com/documentation/uikit/uimenuelement/attributes?language=objc).

The default `Select All` item is replaced with a custom item that maps to the `EditableTextState` handling of `Select All`. This item has [`IOSSystemContextMenuItemCustomAttributes.keepsMenuPresented`](https://developer.apple.com/documentation/uikit/uimenuelement/attributes/keepsmenupresented?language=objc) as its attribute to keep the context menu visible when the `Select All` button is pressed.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/system_context_menu.dart
+++ b/packages/flutter/lib/src/widgets/system_context_menu.dart
@@ -163,7 +163,7 @@ class SystemContextMenu extends StatefulWidget {
       if (editableTextState.selectAllEnabled)
         IOSSystemContextMenuItemCustom(
           title: 'Select All',
-          keepsMenuPresented: true,
+          attributes: IOSSystemContextMenuItemCustomAttributes.keepsMenuPresented | IOSSystemContextMenuItemCustomAttributes.disabled,
           onPressed: () {
             editableTextState.selectAll(SelectionChangedCause.toolbar);
           },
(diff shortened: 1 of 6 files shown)


packages/flutter/lib/src/widgets/system_context_menu.dart
```diff
@@ -160,7 +160,14 @@ class SystemContextMenu extends StatefulWidget {
       if (editableTextState.copyEnabled) const IOSSystemContextMenuItemCopy(),
       if (editableTextState.cutEnabled) const IOSSystemContextMenuItemCut(),
       if (editableTextState.pasteEnabled) const IOSSystemContextMenuItemPaste(),
-      if (editableTextState.selectAllEnabled) const IOSSystemContextMenuItemSelectAll(),
+      if (editableTextState.selectAllEnabled)
+        IOSSystemContextMenuItemCustom(
```
Renzo-Olivares, line 164: This could also be made into its own class/ replace the existing select all class so it can be available to more users, something like:

```dart
final class IOSSystemContextMenuItemSelectAll extends IOSSystemContextMenuItem {
  /// Creates an instance of [IOSSystemContextMenuItemSelectAll].
  const IOSSystemContextMenuItemSelectAll({required this.editableTextState});

  final EditableTextState editableTextState;

  @override
  IOSSystemContextMenuItemData getData(WidgetsLocalizations localizations) {
    return IOSSystemContextMenuItemDataCustom(
      title: localizations.selectAllButtonLabel,
      attributes: IOSSystemContextMenuItemAttributes.keepsMenuPresented,
      onPressed: () => editableTextState.selectAll(SelectionChangedCause.toolbar),
    );
  }
}
```

Renzo-Olivares:
This is still on my radar, I plan to work on this soon.

PR #175092, opened 2025-09-08
Add subtextGap to InputDecorationThemeData

vasilich6107:
Hi.

This PR adds `subtextGap` property to `InputDecorationThemeData` which fixes the issue with inability to customize the value.

I intentionally skipped tests to save the time.
Could you check if this PR makes sense and if it does I will try to add tests.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/input_decorator.dart
+++ b/packages/flutter/lib/src/material/input_decorator.dart
@@ -589,6 +589,7 @@ class _Decoration {
     required this.visualDensity,
     required this.inputGap,
     required this.maintainHintSize,
+    required this.subtextGap,
     this.icon,
     this.input,
     this.label,
@@ -615,6 +616,7 @@ class _Decoration {
   final VisualDensity visualDensity;
   final double inputGap;
   final bool maintainHintSize;
+  final double subtextGap;
   final Widget? icon;
   final Widget? input;
   final Widget? label;
@@ -649,6 +651,7 @@ class _Decoration {
         other.visualDensity == visualDensity &&
         other.inputGap == inputGap &&
         other.maintainHintSize == maintainHintSize &&
+        other.subtextGap == subtextGap &&
         other.icon == icon &&
         other.input == input &&
         other.label == label &&
@@ -683,7 +686,7 @@ class _Decoration {
     prefix,
     suffix,
     prefixIcon,
-    Object.hash(suffixIcon, helperError, counter, container),
+    Object.hash(suffixIcon, helperError, counter, container, subtextGap),
   );
 }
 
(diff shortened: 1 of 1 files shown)


packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -5759,6 +5777,12 @@ class InputDecorationThemeData with Diagnosticable {
         defaultValue: defaultTheme.visualDensity,
       ),
     );
+    properties.add(
+          DoubleProperty(
+        'subtextGap     subtextGap,
+        defaultValue: defaultTheme.subtextGap
+      )
+          );
```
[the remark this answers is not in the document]
vasilich6107: Fixed

vasilich6107:
Hi @bleroux 
Could you help with review

bleroux:
We tend to be very conservative on adding parameter to input decoration as the code is already complex.
In that case there is already some logic to deal with this value being different in M3 and M2 so it seems ok to expose it.
cc @justinmc for a decision on whether we're ok to add this property.

If yes, this PR would need some changes (M2 and M3 default subtextGap values would have to come from _InputDecoratorDefaultsM2 and _InputDecoratorDefaultsM3, which requires updating the corresponding template files)

vasilich6107:
@justinmc @bleroux 
This is so much pain on non having this parameter exposed
As a Flutter dev I'm struggling with smth like `TextStyle(height: 0.8)` when I need to adjust this prop a bit.
I would appreciate if you guide me towards acceptable solution

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -725,9 +728,7 @@ class _RenderDecoration extends RenderBox
        _expands = expands,
        _material3 = material3;
 
-  // TODO(bleroux): consider defining this value as a Material token and making it
-  // configurable by InputDecorationThemeData.
```
justinmc, line 729: @bleroux Does this PR satisfy what you wanted from this TODO? Should we leave the part about defining it as a Material token?

justinmc (commented):
+1 to @bleroux's statement about being cautious about adding parameters here. @vasilich6107 can you post a screenshot of the problem that you're having with `TextStyle(height: 0.8)` and how this PR fixes it? I just want to be sure that the change is valuable and there's not another way to get what you want. Otherwise I'm on board with adding this parameter.

packages/flutter/lib/src/material/input_decorator.dart
justinmc: Oh I think you referred to this in your comment https://github.com/flutter/flutter/pull/175092#issuecomment-3280280915.

dkwingsmt:

PR #175291, opened 2025-09-12
chore: update content workflow to use itnernal script

jtmcdole:
Now that our internal scripts check for private github branches, this makes the itnernal scripts the source of truth.

towards #175265


jtmcdole:
Can wait till Monday. Tested locally with `act`.

matanlurey (approved):

ievdokdm (approved):

PR #170943, opened 2025-06-20
Correct position for an open CupertinoContextMenu

victorsanni:
### Flutter
https://github.com/user-attachments/assets/e120bdda-4236-480d-a85f-183b176bc2a9

### Native

https://github.com/user-attachments/assets/7577b0a0-9111-47e0-ad5d-f533a14310ea

### Flutter
https://github.com/user-attachments/assets/98ba9ffe-715f-4ec6-8d33-e10a8ae69d7d

### Native

https://github.com/user-attachments/assets/b6dbc891-5e01-4ac5-a362-d714328e46d3

### Flutter
https://github.com/user-attachments/assets/9191b8ba-e80a-4d88-b6c4-75876704c2aa

### Native

https://github.com/user-attachments/assets/bb7dfe21-a87b-4eab-a2d2-fab35a7a4c70

<details>
<summary>Flutter sample code</summary>

```dart

import 'package:flutter/cupertino.dart';

void main() => runApp(const ContextMenuApp());

class ContextMenuApp extends StatelessWidget {
  const ContextMenuApp({super.key});

  @override
  Widget build(BuildContext context) {
    return CupertinoApp(
      home: MyStatelessWidget(),
    );
  }
}

/// This is the stateless widget that the main application instantiates.
class MyStatelessWidget extends StatelessWidget {
  const MyStatelessWidget({super.key});

  List<String> getDataList() {
    List<String> list = [];
    for (int i = 0; i < 15; i++) {
      list.add(i.toString());
    }
    return list;
  }

  List<Widget> getWidgetList(BuildContext context) {
    return getDataList()
        .map((item) => getItemContainer(context, item))
        .toList();
  }

  Widget getItemContainer(BuildContext context, String item) {
    return CupertinoContextMenu(
      actions: List<Widget>.generate(
        4,
        (index) {
          return CupertinoContextMenuAction(
            child: Text('Action ${index + 1}'),
            onPressed: () {
              Navigator.pop(context);
            },
          );
        },
      ),
      child: SizedBox.square(
        dimension:
            350, // Change to increase the size of the context menu's child.
        child: ColoredBox(
          color: CupertinoColors.destructiveRed,
          child: Align(
            alignment: Alignment.center,
            child: Text(
              item,
              style: TextStyle(color: CupertinoColors.white, fontSize: 20),
            ),
          ),
        ),
      ),
    );
  }

  @override
  Widget build(BuildContext context) {
    return CupertinoPageScaffold(
[...]

PR #175300, opened 2025-09-13
CupertinoContextMenu child respects available screen width

victorsanni:
Missed this in [Correct position for an open CupertinoContextMenu](https://github.com/flutter/flutter/pull/170943)

Fixes [CupertinoContextMenu wrong position](https://github.com/flutter/flutter/issues/175225)


packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -1463,10 +1463,17 @@ class _ContextMenuAlignedChildrenDelegate extends MultiChildLayoutDelegate {
 
     final double availableHeightForChild =
         screenBounds.height - _ContextMenuRouteStaticState._kPadding;
+    final double availableWidth = screenBounds.width - _ContextMenuRouteStaticState._kPadding * 2;
```
MitchellGoodwin, line 1466: It looks like elsewhere in the file, _kPadding is used without doubling it, so is multiplying it by two necessary here?

packages/flutter/test/cupertino/context_menu_test.dart
```diff
@@ -1453,4 +1453,78 @@ void main() {
...
+
+    // Start a press on the child.
+    final TestGesture gesture = await tester.startGesture(childRect.center);
+    await tester.pump();
+    await tester.pump(const Duration(milliseconds: 500));
+    await gesture.up();
+    await tester.pumpAndSettle();
+
+    expect(tester.takeException(), null);
+  });
```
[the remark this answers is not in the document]
MitchellGoodwin: I don't know if this code from the bot would work, but you might be able to try getting the x an y coordinates of the four corners of the context menu and seeing if they are within the limits you expect and that might be more robust then just checking for exceptions.

MitchellGoodwin (commented):
Approach looks good, I left two comments

packages/flutter/lib/src/cupertino/context_menu.dart
victorsanni: Yes, the native context menu applies a vertical margin only to the bottom but not the top, but the horizontal margin is applied to the left and right sides. 

The other uses of the horizontal _kPadding (that aren't doubled) refer to the padding between the context menu child and the menu sheet.

MitchellGoodwin (approved):
LGTM!

PR #175321, opened 2025-09-14
Make sure that GridTile doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the GridTile widget.


packages/flutter/test/material/grid_title_test.dart
```diff
@@ -47,4 +47,21 @@ void main() {
...
+            child: GridTile(
+              header: GridTileBar(title: Text('X')),
+              child: Text('X'),
+            ),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(GridTile)), Size.zero);
+    expect(tester.getSize(find.byType(GridTileBar)).isEmpty, isTrue);
```
dkwingsmt: Can you split `GridTileBar` into a separate test case so that we can verify its layout in a constraint of exactly zero?

dkwingsmt (approved):
Checklist:

victorsanni (approved):

Issue #158050, opened 2024-11-02
Stop pinning dependencies


spydon:
### Use case

One of the biggest pain points for package authors (and probably app developers too) that have been brought up during the Dart & Flutter Package Ecosystem Summits is that Flutter has pinned versions of its dependencies.

Why do we not want pinned dependencies? Because it leads to a lot trickier dependency version resolution than it needs to be for the end user. I would say almost all Flutter developers have at some point ended up in dependency resolution hell, before they realize that some dependencies needs to be treated with special care.

Since all of the pinned dependencies are controlled by Google it should be fairly straight forward to set up a process so that they all have some tests towards Flutter before doing a new non-breaking release or doing deprecations.
The main ones are: intl, meta, collection path, vector_math and async.

The Flutter team’s rationale for using pinned versions is outlined in [this document](https://github.com/dart-lang/sdk/blob/main/docs/Flutter-Pinned-Packages.md) but **I have been told that removing the dependency pinning is on the radar, so I just want to open this so that we can track it somewhere.**

### Proposal

Things that needs to be solved:
 - [ ] Flutter builds should be hermetic.
 - [ ] When dependencies bump their versions they should not be able to break Flutter.

For the first one I'm thinking that it could be solved with a lock file that isn't shipped with Flutter, but used for the builds.

For the second one my suggestion is that we introduce CI checks that needs to run before releasing a new version of the dependencies to make sure that they don't break anything in Flutter. Deprecations have to be taken into consideration here too, since they are not counted as breaking in semver.

spydon:
> We could also flip this around a little bit, if the flutter repo migrated to use the "workspaces" feature from pub. Then, we would manually curate these packages and their pinned versions - and this would transitively enforce the versions in the other packages in the workspace.

That sounds like a great idea! Then it should be very easy to properly check that a change is not breaking anything in Flutter too.

stuartmorgan:
I'm not sure if there's an issue tracking it, but the other approach that's been floated for this in vendoring all of our dependencies. That doesn't work for anything we want to use in public APIs though.

jakemac53:
Many of the pinned deps come from `flutter_test` which depends on (and re-exports) `test_api`, which also definitely leaks various things from its own public APIs. So, I don't think vendoring is really an option for `flutter_test` at least.

devoncarew:
Re:

[...]

PR #175352, opened 2025-09-15
Have `flutter create` create a pubspec.lock to ensure pinned versions are being used.

sigurdm:
Part of https://flutter.dev/go/unpin-flutter-sdk-dependencies
Preparing for https://github.com/flutter/flutter/issues/158050


packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -1351,3 +1355,49 @@ List<String>? _getBuildGradleConfigurationFilePaths(
...
+  'matcher',
+  'fake_async',
+  'clock',
+  'stack_trace',
+  'vector_math',
+  'leak_tracker_flutter_testing',
+  'collection',
+  'meta',
+  'stream_channel',
+];
```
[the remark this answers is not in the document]
sigurdm: Done

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -1351,3 +1355,68 @@ List<String>? _getBuildGradleConfigurationFilePaths(
...
+    if (descriptor is YamlMap && descriptor['sdk'] == 'flutter') {
+      // a flutter dependency.
+      final name = dependency.key as String;
+      sdkPackages.add(name);
+    }
+  }
+
+  final result = <String>{};
+  for (final sdkPackage in sdkPackages) {
+    final String flutterRoot = Cache.flutterRoot!;
```
stuartmorgan-g: This could use a comment saying why force-unwrapping is safe.

Also, it doesn't need to be in the loop.

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -1351,3 +1355,68 @@ List<String>? _getBuildGradleConfigurationFilePaths(
...
+  };
+
+  /// Pub will rewrite this file as yaml pretty after resolution.
+  directory
+      .childFile('pubspec.lock')
+      .writeAsStringSync(const JsonEncoder.withIndent('  ').convert({'packages': packages}));
+}
+
+/// Find the external dependencies from the SDK packages that the package in
+/// [directory] depends on.
```
stuartmorgan-g: Maybe clarify that this is package names; `String` doesn't convey much detail about what exactly is returned.

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -1351,3 +1355,68 @@ List<String>? _getBuildGradleConfigurationFilePaths(
...
+
+  final packages = <String, Object?>{
+    for (final package in gatherSdkPackageDependencies(directory))
+      package: flutterPackages[package],
+  };
+
+  /// Pub will rewrite this file as yaml pretty after resolution.
+  directory
+      .childFile('pubspec.lock')
+      .writeAsStringSync(const JsonEncoder.withIndent('  ').convert({'packages': packages}));
```
stuartmorgan-g, line 1381: Is the format of this file formal API, or is it subject to change in the future?

packages/flutter_tools/test/commands.shard/permeable/create_test.dart
```diff
@@ -5125,6 +5125,30 @@ To keep the default AGP version $templateAndroidGradlePluginVersion, download a
     expect(logger.statusText, isNot(contains(r'  $ cd')));
   }, overrides: {Logger: () => logger});
 
+  testUsingContext(
+    'default project locks known dependencies to known versions with pubspec.lock',
+    () async {
+      await _createProject(projectDir, <String>['--no-pub'], <String>['pubspec.lock']);
```
stuartmorgan-g, line 5131: The `create` command will itself trigger the pub logic that will consume and rewrite this file, right? I.e., the check below is testing the post-processed version of the file, and would thus fail if `pub` stopped understanding the format we are writing?

packages/flutter_tools/lib/src/commands/create.dart
sigurdm: It can change in a backwards compatible manner. We always want pub to be able to read old pubspec.lock files.

packages/flutter_tools/test/commands.shard/permeable/create_test.dart
sigurdm: That's true. That should not happen

packages/flutter_tools/lib/src/commands/create.dart
sigurdm: Done

stuartmorgan-g (approved):
LGTM, sorry for the delay.

PR #161955, opened 2025-01-21
Expose Semantics.identifier as Windows UIA AutomationId

loic-peron-inetum-public:
Expose Semantics.identifier as Windows AutomationId to ease application automation.

see https://github.com/flutter/flutter/issues/148763

followup to https://github.com/flutter/engine/pull/53729 and https://github.com/flutter/engine/pull/53476

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

Issue #148763, opened 2024-05-21
[Windows] Missing automationId for testing purposes


nicerloop:
### Use case

The Windows platform does not implement the mapping of  `Semantics.identifier` property to `automationId`, which is usefull in automated test, specifically using Appium.

This is implemented for iOS and Android:
- https://github.com/flutter/flutter/issues/17988#issuecomment-1867097631
- https://github.com/flutter/flutter/pull/138331
- https://github.com/flutter/engine/pull/47961
- https://github.com/flutter/engine/pull/48858

Windows favors `automationId `for automated testing:
- https://learn.microsoft.com/en-us/dotnet/framework/ui-automation/using-ui-automation-for-automated-testing#automationid
- https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-usefortesting#key-properties-for-test-automation

1. 

### Proposal

The Flutter engine should map `Semantics.identifier` to `automationId`.

nicerloop:
Could I get some pointers to the code responsible for handling Windows accessiblity properties. I am willing to try and implement this feature, but as I am completely new to Flutter, I need some directions.

nicerloop:
I found https://github.com/flutter/flutter/wiki/Accessibility-on-Windows
Still searching for the corresponding code.

loic-sharma:
@nicerloop Hello, some pointers:

1. [`AccessibilityBridgeWindows`](https://github.com/flutter/engine/blob/8b094fbb94d83fa1dfdde4598c5ff76f6147f5cc/shell/platform/windows/accessibility_bridge_windows.h#L27) is the logic that translates from Flutter's accessibility object model to the object model used by our fork of Chromium's accessibility library
2. [`//third_party/accessibility/`](https://github.com/flutter/engine/tree/main/third_party/accessibility) is our fork of Chromium's accessibility library.
    1. [`//third_party/accessibility/ax/platform/`](https://github.com/flutter/engine/tree/main/third_party/accessibility/ax/platform) is the platform-specific code. Check the files with `_win` suffixes for Windows platform-specific code. This maps from Chromium's object model to [Microsoft Active Accessibility](https://learn.microsoft.com/en-us/windows/win32/winauto/microsoft-active-accessibility)
[...]

PR #175405, opened 2025-09-16
[Desktop] Propagate SemanticsNode::identifier to AXPlatformNodeDelegate::AuthorUniqueId

loic-peron-inetum-public:
Expose flutter Semantics.identifier as AXPlatformNodeDelegate::AuthorUniqueId.
To be used for UI test automation through Windows AutomationId.

see https://github.com/flutter/flutter/issues/148763

followup to https://github.com/flutter/flutter/pull/161955

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/common/accessibility_bridge.cc
+++ b/engine/src/flutter/shell/platform/common/accessibility_bridge.cc
@@ -624,6 +624,9 @@ AccessibilityBridge::FromFlutterSemanticsNode(
         flutter_node.custom_accessibility_actions +
             flutter_node.custom_accessibility_actions_count);
   }
+  if (flutter_node.identifier) {
+    result.identifier = std::string(flutter_node.identifier);
+  }
   return result;
 }
 
--- a/engine/src/flutter/shell/platform/common/accessibility_bridge.cc
+++ b/engine/src/flutter/shell/platform/common/accessibility_bridge.cc
@@ -291,6 +291,7 @@ void AccessibilityBridge::ConvertFlutterUpdate(const SemanticsNode& node,
   SetIntAttributesFromFlutterUpdate(node_data, node);
   SetIntListAttributesFromFlutterUpdate(node_data, node);
   SetStringListAttributesFromFlutterUpdate(node_data, node);
+  SetIdentifierFromFlutterUpdate(node_data, node);
   SetNameFromFlutterUpdate(node_data, node);
   SetValueFromFlutterUpdate(node_data, node);
   SetTooltipFromFlutterUpdate(node_data, node);
@@ -521,6 +522,13 @@ void AccessibilityBridge::SetStringListAttributesFromFlutterUpdate(
         ax::mojom::StringListAttribute::kCustomActionDescriptions,
         custom_action_description);
   }
+}
+
+void AccessibilityBridge::SetIdentifierFromFlutterUpdate(
+    ui::AXNodeData& node_data,
+    const SemanticsNode& node) {
+  node_data.AddStringAttribute(ax::mojom::StringAttribute::kIdentifier,
+                               node.identifier);
 }
 
 void AccessibilityBridge::SetNameFromFlutterUpdate(ui::AXNodeData& node_data,
--- a/engine/src/flutter/shell/platform/common/flutter_platform_node_delegate_unittests.cc
+++ b/engine/src/flutter/shell/platform/common/flutter_platform_node_delegate_unittests.cc
@@ -60,6 +60,7 @@ TEST(FlutterPlatformNodeDelegateTest, canPerfomActions) {
   root.tooltip = "";
   root.child_count = 0;
   root.custom_accessibility_actions_count = 0;
+  root.identifier = "";
   bridge->AddFlutterSemanticsNodeUpdate(root);
 
   bridge->CommitUpdates();
@@ -108,6 +109,7 @@ TEST(FlutterPlatformNodeDelegateTest, canGetAXNode) {
   root.tooltip = "";
   root.child_count = 0;
   root.custom_accessibility_actions_count = 0;
+  root.identifier = "";
   bridge->AddFlutterSemanticsNodeUpdate(root);
 
   bridge->CommitUpdates();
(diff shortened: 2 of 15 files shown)


loic-peron-inetum-public:
@loic-sharma could you review this updated PR ?

engine/src/flutter/shell/platform/embedder/embedder.h
```diff
@@ -1693,6 +1693,11 @@ typedef struct {
   // The set of semantics flags associated with this node. Prefer to use this
   // over `flags__deprecated__`.
   FlutterSemanticsFlags* flags2;
+  /// An identifier for the semantics node in native accessibility hierarchy.
+  /// This value should not be exposed to the users of the app.
+  /// This is usually used for UI testing with tools that work by querying the
+  /// native accessibility, like UIAutomator, XCUITest, or Appium.
```
loic-sharma: Looks like UI Automator should have a space: https://developer.android.com/training/testing/other-components/ui-automator

```suggestion
  /// native accessibility, like UI Automator, XCUITest, or Appium.
```

engine/src/flutter/shell/platform/common/accessibility_bridge.cc
```diff
@@ -623,6 +631,9 @@ AccessibilityBridge::FromFlutterSemanticsNode(
         flutter_node.custom_accessibility_actions +
             flutter_node.custom_accessibility_actions_count);
   }
+  if (flutter_node.identifier) {
+    result.identifier = std::string(flutter_node.identifier);
```
loic-sharma, line 636: This could cause the screen reader to announce the identifier to the user, right? Per this [comment](https://github.com/flutter/flutter/pull/175405/files#diff-a52c655cab63cbe1d9bfb9a43cef19c510c16d8c1aba5336b02b0a1685a7caa8R1697), that doesn't seem desirable. Should we remove this?

engine/src/flutter/third_party/accessibility/ax/ax_enums.h
```diff
@@ -532,6 +532,7 @@ enum class StringAttribute {
   kDisplay,
   // Only present when different from parent.
   kFontFamily,
+  kHtmlId,
```
loic-sharma: Just curious, how did you pick this name? Did you consider something like `kAuthorUniqueId`? This would better match `PlatformNodeDelegateBase::GetAuthorUniqueId`.

engine/src/flutter/shell/platform/common/flutter_platform_node_delegate_unittests.cc
```diff
@@ -60,6 +60,7 @@ TEST(FlutterPlatformNodeDelegateTest, canPerfomActions) {
   root.tooltip = "";
   root.child_count = 0;
   root.custom_accessibility_actions_count = 0;
+  root.identifier = "";
```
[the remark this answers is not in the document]
loic-sharma: This is a reasonable suggestion, but let's skip it for this PR.

loic-sharma:
@loic-peron-inetum-public Thanks for the contribution! This is looking good. I left some comments, but otherwise this is looking good!

cc @chunhtai as I'm less familiar with the `//third_party/accessibility/ax/` codebase.

engine/src/flutter/shell/platform/common/accessibility_bridge.cc
loic-peron-inetum-public: No, this alone will not make the identifier exposed to the reader, only propagated from the flutter engine to the desktop accessibility framework. It then needs to be exposed to the platform-specific API, for example https://github.com/flutter/flutter/pull/175406
In all cases, many attributes exposed to the screen reader are silently ignored, only the main ones are announced to the user.

engine/src/flutter/third_party/accessibility/ax/ax_enums.h
loic-peron-inetum-public: I went to the original source for the desktop accessibility framework in Chromium, where the technical id for a DOM element is `kHtmlId`. In an old version https://github.com/flutter/engine/pull/53729 with Windows UI Automation, I used  `kAuthorUniqueId`. When I went the alternate route of `IAccessibleEx`, I took the name from the Chromium sources.

Would you prefer `kAuthorUniqueId`, or `kIdentifier` in line with the previous Gemini proposal and the global Flutter naming ?

chunhtai: We since divert from chromium code base, we should pick a name that best suit our use case.

htmlId doesn't make sense in the context. I vote for kIdentifier.

chunhtai:
agree with @loic-sharma , we should not use the htmlId since the data doesn't come from a web dom but flutter app

engine/src/flutter/shell/platform/common/accessibility_bridge.cc
```diff
@@ -523,6 +524,13 @@ void AccessibilityBridge::SetStringListAttributesFromFlutterUpdate(
   }
 }
 
+void AccessibilityBridge::SetHtmlIdFromFlutterUpdate(
+    ui::AXNodeData& node_data,
+    const SemanticsNode& node) {
+  node_data.AddStringAttribute(ax::mojom::StringAttribute::kHtmlId,
+                               node.identifier);
+}
```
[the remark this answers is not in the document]
loic-peron-inetum-public: Done

engine/src/flutter/third_party/accessibility/ax/ax_enums.h
loic-peron-inetum-public: Done

loic-peron-inetum-public:
Rebased on master and replaced kHtmlId with kIdentifier.

chunhtai (approved):
LGTM

engine/src/flutter/shell/platform/common/accessibility_bridge.cc
loic-sharma: Thanks for confirming!

loic-sharma (approved):
LGTM! Thanks for the contribution :)

justgitstuff:
Hi these changes are not working for macos and linux flutter apps, can you please suggest a workaround?

PR #175448, opened 2025-09-16
Minor refactor to eliminate repetitive pattern

AbdeMohlbi:
to my understanding :
```dart
final int? x = matchGroups[0] == null ? null : int.tryParse(matchGroups[0]!);
```
means that if `matchGroups[0] ` is `null` then `x` is `null` , else try to parse the int from `matchGroups[0]` which is a `String` now (because `final List<String?> matchGroups`).
from the other side `int.tryParse` would parse `null.toString()` and return `null` which is what i need in the first place 
so :
```dart
 matchGroups[0] == null ? null : int.tryParse(matchGroups[0]!);
// is equivalent to : 
int.tryParse(matchGroups[0].toString());
```
here is an simple example for that :
```dart
void main() {
  String? g = null;
  final result = int.tryParse(g.toString());
  print('g.toString() -> "$g" -> result=$result');
}
// g.toString() -> "null" -> result=null
```

## Pre-launch Checklist

- [x] I signed the [CLA].


AbdeMohlbi:
TODO: check other instances ( 5 to my knowledge)

reidbaker (commented):
This is a delightful abuse of the language feature that null is a Null object with a toString implementation and that tryParse has a well defined path for ints it cant parse. 

That said I dont think this is more readable. Tagging some others on the team to see what they think.

@mboetger @stuartmorgan-g what do you think of this pattern?

stuartmorgan-g:
> That said I dont think this is more readable.

Strongly agree; this replaces code that is explicit about what it is doing to code that only works for non-obvious reasons. And while I don't think it will actually matter, this version is almost certainly less efficient at runtime.

(Also, I think `null.toString()` working is a foot-gun, so I'm inherently inclined not to like code that relies on it.)

AbdeMohlbi:
> > That said I dont think this is more readable.
> 
> Strongly agree; this replaces code that is explicit about what it is doing to code that only works for non-obvious reasons. And while I don't think it will actually matter, this version is almost certainly less efficient at runtime.
> 
> (Also, I think `null.toString()` working is a foot-gun, so I'm inherently inclined not to like code that relies on it.)

If i may ask why is it less efficient at runtime ?

stuartmorgan-g:
> If i may ask why is it less efficient at runtime ?

Because parsing a string requires a lot more operations than a single pointer comparison.

reidbaker:
Given that @stuartmorgan-g agrees that this change is less readable I am going to close this pr. @AbdeMohlbi thank you for your many contributions and keep up the good work.

Issue #164267, opened 2025-02-27
Cupertino blurry surfaces (such as CupertinoAlertDialog) should not source out-of-clip content


dkwingsmt:
The following screenshot from XCode displays an alert dialog over a black rectangle that's just as wide as the dialog. The dialog's blurry background is almost uniformly #c5c5c5, i.e. does not blur at all from the white content, where the dialog does not cover.
<img width="313" alt="Image" src="https://github.com/user-attachments/assets/88aa8141-456f-4a0f-a221-1bf7b756fc64" />

This is not the same in Flutter, as shown in the following screenshot, where the white obviously bleeds in.

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/ff30f961-6559-4910-8d20-b936abb9b822" />

This is expected, since Flutter's widgets blurs the background before clipping. We should clip it before blurring.

Code:
```dart
import 'package:flutter/cupertino.dart';
import 'package:flutter/material.dart';

Widget createAppWithButtonThatLaunchesDialog(WidgetBuilder dialogBuilder) {
  return CupertinoApp(
    theme: CupertinoThemeData(brightness: Brightness.light),
    home: CupertinoPageScaffold(
      child: Center(
        child: Column(
            mainAxisAlignment: MainAxisAlignment.center,
            crossAxisAlignment: CrossAxisAlignment.center,
            children: <Widget>[
              Builder(builder: (BuildContext context) {
                return CupertinoButton(
                  onPressed: () {
                    showCupertinoDialog<void>(
                      context: context,
                      builder: dialogBuilder,
                    );
                  },
                  child: const Text('Flutter'),
                );
              }),
              Container(
                width: 270,
                height: 500,
                color: Colors.black,
              )
            ]),
      ),
    ),
  );
}

void main() => runApp(const App());

class App extends StatelessWidget {
  const App({super.key});

  @override
  Widget build(BuildContext context) {
    return createAppWithButtonThatLaunchesDialog((BuildContext context) {
        return CupertinoAlertDialog(
          title: const Text('The title'),
          actions: <Widget>[
            CupertinoDialogAction(
              child: const Text('One'),
              onPressed: () {
                Navigator.pop(context);
              },
            ),
            CupertinoDialogAction(
              child: const Text('Two'),
              onPressed: () {
                Navigator.pop(context);
              },
            ),
          ],
        );
      },
    );
  }
}

```

delfme:
[...]

Issue #99691, opened 2022-03-07
Surrounding color leaks into widgets when BackdropFilter with ImageFilter.blur is applied


Ruzo:
I'm having an issue with a GridView of blurred tiles using `BackdropFilter with ImageFilter.blur` where any surrounding color leaks into them:

![imagefilter_blur_issue](https://user-images.githubusercontent.com/1992115/156901663-c02e981c-29fc-4b41-bc27-91ac3853f3f1.png)

![ImageFilter_blur_issue_white_color](https://user-images.githubusercontent.com/1992115/156901757-1f0f29d9-13d5-4dc8-8aa3-ffcd0847d861.png)

Here is a Dartpad example: https://dartpad.dev/?id=1196002918293fdd251557b037d2d786

<details>
<summary>Here is the code sample:</summary>

```dart
import 'dart:ui';
import 'package:flutter/material.dart';

void main() => runApp(MyApp());

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'ImageFilter Color Leak Issue',
      debugShowCheckedModeBanner: false,
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: Scaffold(
        backgroundColor: Colors.white,
        appBar: AppBar(
          title: const Text('ImageFilter.blur Color Leak Issue'),
        ),
        body: Column(
          mainAxisAlignment: MainAxisAlignment.spaceEvenly,
          children: [
            Expanded(
              child: Container(
                color: Colors.white,
                width: double.infinity,
                child: FittedBox(
                  fit: BoxFit.contain,
                  child: Padding(
                    padding: const EdgeInsets.all(40),
                    child: Container(
                      width: 400,
                      height: 400,
                      color: const Color(0xFF0000FF),
                      child: GridView.count(
                        padding: EdgeInsets.zero,
                        shrinkWrap: true,
                        physics: const NeverScrollableScrollPhysics(),
                        crossAxisCount: 3,
                        children: tiles(),
                      ),
                    ),
                  ),
                ),
              ),
            ),
            Expanded(
              child: Container(
                color: Colors.black,
                width: double.infinity,
                child: FittedBox(
                  fit: BoxFit.contain,
                  child: Padding(
                    padding: const EdgeInsets.all(40),
                    child: Container(
                      width: 400,
                      height: 400,
                      color: const Color(0xFF0000FF),
                      child: GridView.count(
[...]

PR #175458, opened 2025-09-16
[Engine] iOS style blurring

dkwingsmt:
This PR adds the engine support for a new iOS-style blur. It works by adding parameters to the blur filter that specify its _blurring bounds_.

This is the engine-side implementation. The corresponding framework changes that expose this to developers are in:
* Framework PR: https://github.com/flutter/flutter/pull/175473

Related issues:
* Main tracking issue: https://github.com/flutter/flutter/issues/99691
* Algorithm details: https://github.com/flutter/flutter/issues/164267#issuecomment-2731163225

Design doc & previous discussions: [flutter.dev/go/ios-style-blur-support](flutter.dev/go/ios-style-blur-support)

### The Visual (Before & After)

This new mode, which I'm calling "bounded blur," is different from the traditional (global) gaussian blur in that blurs would not sample transparent pixels from outside the provided area.

The demo below shows the old blur (left) and the new bounded blur (right). Both are blurring a black triangle.

image

image

Notice the new version on the right no longer has the bright "lining" at the top and left edges. This is because the blur algorithm now knows its own bounds and correctly stops sampling pixels from outside that area.

#### API Change

To pass the bounds information down, I've added new parameters to `_initBlur`:

```dart
  // painting.dart
  external void _initBlur(
    double sigmaX,
    double sigmaY,
    int tileMode,
    bool bounded,  // Start of new parameters
    double boundsLeft,
    double boundsTop,
    double boundsRight,
    double boundsBottom,
  );
```

#### How the Bounds Are Used
These bounds are passed all the way down to `GaussianBlurFilterContents` and affect two key parts of the process:

* Downsampling Pass: The shader is instructed not to sample any pixels outside the provided bounds.
* Blurring Passes: The final blurred result is divided by the resulting opacity. This normalizes the varying alpha (due to varying sum of weights) across the pixels near the edge.

#### Notable Engine Changes
To handle the downsampling logic, I created a new downsampling shader `texture_downsample_bounded`.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/display_list/effects/dl_image_filter.cc
+++ b/engine/src/flutter/display_list/effects/dl_image_filter.cc
@@ -12,8 +12,7 @@ namespace flutter {
 std::shared_ptr<DlImageFilter> DlImageFilter::MakeBlur(DlScalar sigma_x,
                                                        DlScalar sigma_y,
                                                        DlTileMode tile_mode) {
-  return DlBlurImageFilter::Make(sigma_x, sigma_y, /*bounds=*/std::nullopt,
-                                 tile_mode);
+  return DlBlurImageFilter::Make(sigma_x, sigma_y, tile_mode);
 }
 
 std::shared_ptr<DlImageFilter> DlImageFilter::MakeDilate(DlScalar radius_x,
--- a/engine/src/flutter/display_list/effects/dl_image_filter.cc
+++ b/engine/src/flutter/display_list/effects/dl_image_filter.cc
@@ -12,7 +12,8 @@ namespace flutter {
 std::shared_ptr<DlImageFilter> DlImageFilter::MakeBlur(DlScalar sigma_x,
                                                        DlScalar sigma_y,
                                                        DlTileMode tile_mode) {
-  return DlBlurImageFilter::Make(sigma_x, sigma_y, tile_mode);
+  return DlBlurImageFilter::Make(sigma_x, sigma_y, /*bounds=*/std::nullopt,
+                                 tile_mode);
 }
 
 std::shared_ptr<DlImageFilter> DlImageFilter::MakeDilate(DlScalar radius_x,
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -27,16 +27,19 @@ class DlBlurImageFilter final : public DlImageFilter {
    * The [bounds] parameter is optional and dictates the blur's sampling
    * behavior:
    *
-   * - If [bounds] is provided (i.e., not std::nullopt), the filter performs a
-   * "bounded blur." This means the blur kernel will only sample pixels
-   * from *within* this rectangle, treating all pixels outside of it as
-   * transparent. This mode is used to implement high-fidelity iOS-style blurs.
-   *
    * - If [bounds] is std::nullopt, a standard, unbounded blur is performed,
    * with edge behavior defined by [tile_mode].
    *
+   * - If [bounds] is provided (i.e., not std::nullopt), the filter performs a
+   * "bounded blur". This means the blur kernel will only sample pixels from
+   * *within* this rectangle, treating all pixels outside of it as transparent,
+   * and the resulting pixels are opaque. This mode is used to implement
+   * iOS-style blurs.
+   *
    * The [bounds] rectangle must be specified in the current coordinate space
    * of the canvas (i.e., it is subject to the canvas's current transform).
+   * In other words, in the canvas coordinate space, the bounds might not be an
+   * axis-aligned rectangle.
    */
   DlBlurImageFilter(DlScalar sigma_x,
                     DlScalar sigma_y,
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -21,21 +21,26 @@ class DlBlurImageFilter final : public DlImageFilter {
    * @param sigma_y The standard deviation of the Gaussian kernel in the Y
    * direction.
    * @param bounds An optional rectangle that enables "bounded blur" mode.
-   * @param tile_mode The tile mode used for sampling pixels at the edges
-   * when performing a standard, unbounded blur.
+   * @param tile_mode Defines how to sample from areas outside the bounds of the
+   * input texture.
    *
-   * If [bounds] is std::nullopt, a standard Gaussian blur is applied
-   * and edge sampling behavior is determined by [tileMode].
+   * If `bounds` is std::nullopt, a standard Gaussian blur is applied and to the
+   * entire surface.
    *
-   * If [bounds] is not std::nullopt, the filter performs a "bounded blur": the
-   * blur kernel only samples pixels from inside the rectangle, pixels outside
-   * the rectangle are treated as transparent, and the resulting pixels are
-   * unpremultiplied to produce opaque output. This mode is used to implement
-   * iOS-style bounded blurs.
+   * If `bounds` is not std::nullopt, the filter performs a "bounded blur": the
+   * image filter substitutes transparent black for any sample it reads from
+   * outside the defined bounding rectangle. The final weighted sum is then
+   * divided by the total weight of the non-transparent samples (the effective
+   * alpha), resulting in opaque output.
    *
-   * The [bounds] rectangle is specified in the canvas's current coordinate
-   * space (it is affected by the current transform). Consequently, the bounds
-   * may not be axis-aligned in canvas coordinates.
+   * The bounded mode prevents color bleeding from content adjacent to the
+   * bounds into the blurred area, and is typically used when the blur must be
+   * strictly contained within a clipped region, such as for iOS-style frosted
+   * glass effects.
+   *
+   * The `bounds` rectangle is specified in the canvas's current coordinate
+   * space and is affected by the current transform; consequently, the bounds
+   * may not be axis-aligned in the final canvas coordinates.
    */
   DlBlurImageFilter(DlScalar sigma_x,
                     DlScalar sigma_y,
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -13,17 +13,50 @@ namespace flutter {
 
 class DlBlurImageFilter final : public DlImageFilter {
  public:
-  DlBlurImageFilter(DlScalar sigma_x, DlScalar sigma_y, DlTileMode tile_mode)
-      : sigma_x_(sigma_x), sigma_y_(sigma_y), tile_mode_(tile_mode) {}
+  /**
+   * @brief Creates an ImageFilter that applies a Gaussian blur to its input.
+   *
+   * @param sigma_x The standard deviation of the Gaussian kernel in the X
+   * direction.
+   * @param sigma_y The standard deviation of the Gaussian kernel in the Y
+   * direction.
+   * @param bounds An optional rectangle that enables "bounded blur" mode.
+   * @param tile_mode The tile mode used for sampling pixels at the edges
+   * when performing a standard, unbounded blur.
+   *
+   * The [bounds] parameter is optional and dictates the blur's sampling
+   * behavior:
+   *
+   * - If [bounds] is provided (i.e., not std::nullopt), the filter performs a
+   * "bounded blur." This means the blur kernel will only sample pixels
+   * from *within* this rectangle, treating all pixels outside of it as
+   * transparent. This mode is used to implement high-fidelity iOS-style blurs.
+   *
+   * - If [bounds] is std::nullopt, a standard, unbounded blur is performed,
+   * with edge behavior defined by [tile_mode].
+   *
+   * The [bounds] rectangle must be specified in the current coordinate space
+   * of the canvas (i.e., it is subject to the canvas's current transform).
+   */
+  DlBlurImageFilter(DlScalar sigma_x,
+                    DlScalar sigma_y,
+                    std::optional<DlRect> bounds,
+                    DlTileMode tile_mode)
+      : sigma_x_(sigma_x),
+        sigma_y_(sigma_y),
+        bounds_(bounds),
+        tile_mode_(tile_mode) {}
   explicit DlBlurImageFilter(const DlBlurImageFilter* filter)
       : DlBlurImageFilter(filter->sigma_x_,
                           filter->sigma_y_,
+                          filter->bounds_,
                           filter->tile_mode_) {}
   DlBlurImageFilter(const DlBlurImageFilter& filter)
       : DlBlurImageFilter(&filter) {}
 
   static std::shared_ptr<DlImageFilter> Make(DlScalar sigma_x,
                                              DlScalar sigma_y,
+                                             std::optional<DlRect> bounds,
                                              DlTileMode tile_mode);
 
   std::shared_ptr<DlImageFilter> shared() const override {
@@ -50,6 +83,7 @@ class DlBlurImageFilter final : public DlImageFilter {
 
   DlScalar sigma_x() const { return sigma_x_; }
   DlScalar sigma_y() const { return sigma_y_; }
+  std::optional<DlRect> bounds() const { return bounds_; }
   DlTileMode tile_mode() const { return tile_mode_; }
 
  protected:
@@ -58,6 +92,7 @@ class DlBlurImageFilter final : public DlImageFilter {
  private:
   DlScalar sigma_x_;
   DlScalar sigma_y_;
+  std::optional<DlRect> bounds_;
   DlTileMode tile_mode_;
 };
 
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -30,11 +30,11 @@ class DlBlurImageFilter final : public DlImageFilter {
    * - If [bounds] is std::nullopt, a standard, unbounded blur is performed,
    * with edge behavior defined by [tile_mode].
    *
-   * - If [bounds] is provided (i.e., not std::nullopt), the filter performs a
-   * "bounded blur". This means the blur kernel will only sample pixels from
-   * *within* this rectangle, treating all pixels outside of it as transparent,
-   * and the resulting pixels are opaque. This mode is used to implement
-   * iOS-style blurs.
+   * - If [bounds] is not std::nullopt, the filter performs a "bounded blur".
+   * This means the blur kernel will only sample pixels from within this
+   * rectangle, treating all pixels outside of it as transparent, and the
+   * resulting pixels are opaque. This mode is used to implement iOS-style
+   * blurs.
    *
    * The [bounds] rectangle must be specified in the current coordinate space
    * of the canvas (i.e., it is subject to the canvas's current transform).
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -24,22 +24,18 @@ class DlBlurImageFilter final : public DlImageFilter {
    * @param tile_mode The tile mode used for sampling pixels at the edges
    * when performing a standard, unbounded blur.
    *
-   * The [bounds] parameter is optional and dictates the blur's sampling
-   * behavior:
+   * If [bounds] is std::nullopt, a standard Gaussian blur is applied
+   * and edge sampling behavior is determined by [tileMode].
    *
-   * - If [bounds] is std::nullopt, a standard, unbounded blur is performed,
-   * with edge behavior defined by [tile_mode].
+   * If [bounds] is not std::nullopt, the filter performs a "bounded blur": the
+   * blur kernel only samples pixels from inside the rectangle, pixels outside
+   * the rectangle are treated as transparent, and the resulting pixels are
+   * unpremultiplied to produce opaque output. This mode is used to implement
+   * iOS-style bounded blurs.
    *
-   * - If [bounds] is not std::nullopt, the filter performs a "bounded blur".
-   * This means the blur kernel will only sample pixels from within this
-   * rectangle, treating all pixels outside of it as transparent, and the
-   * resulting pixels are opaque. This mode is used to implement iOS-style
-   * blurs.
-   *
-   * The [bounds] rectangle must be specified in the current coordinate space
-   * of the canvas (i.e., it is subject to the canvas's current transform).
-   * In other words, in the canvas coordinate space, the bounds might not be an
-   * axis-aligned rectangle.
+   * The [bounds] rectangle is specified in the canvas's current coordinate
+   * space (it is affected by the current transform). Consequently, the bounds
+   * may not be axis-aligned in canvas coordinates.
    */
   DlBlurImageFilter(DlScalar sigma_x,
                     DlScalar sigma_y,
--- a/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
+++ b/engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
@@ -44,24 +44,25 @@ class DlBlurImageFilter final : public DlImageFilter {
    */
   DlBlurImageFilter(DlScalar sigma_x,
                     DlScalar sigma_y,
-                    std::optional<DlRect> bounds,
-                    DlTileMode tile_mode)
+                    DlTileMode tile_mode,
+                    std::optional<DlRect> bounds = std::nullopt)
       : sigma_x_(sigma_x),
         sigma_y_(sigma_y),
-        bounds_(bounds),
-        tile_mode_(tile_mode) {}
+        tile_mode_(tile_mode),
+        bounds_(bounds) {}
   explicit DlBlurImageFilter(const DlBlurImageFilter* filter)
       : DlBlurImageFilter(filter->sigma_x_,
                           filter->sigma_y_,
-                          filter->bounds_,
-                          filter->tile_mode_) {}
+                          filter->tile_mode_,
+                          filter->bounds_) {}
   DlBlurImageFilter(const DlBlurImageFilter& filter)
       : DlBlurImageFilter(&filter) {}
 
-  static std::shared_ptr<DlImageFilter> Make(DlScalar sigma_x,
-                                             DlScalar sigma_y,
-                                             std::optional<DlRect> bounds,
-                                             DlTileMode tile_mode);
+  static std::shared_ptr<DlImageFilter> Make(
+      DlScalar sigma_x,
+      DlScalar sigma_y,
+      DlTileMode tile_mode,
+      std::optional<DlRect> bounds = std::nullopt);
 
   std::shared_ptr<DlImageFilter> shared() const override {
     return std::make_shared<DlBlurImageFilter>(this);
@@ -87,8 +88,8 @@ class DlBlurImageFilter final : public DlImageFilter {
 
   DlScalar sigma_x() const { return sigma_x_; }
   DlScalar sigma_y() const { return sigma_y_; }
-  std::optional<DlRect> bounds() const { return bounds_; }
   DlTileMode tile_mode() const { return tile_mode_; }
+  std::optional<DlRect> bounds() const { return bounds_; }
 
  protected:
   bool equals_(const DlImageFilter& other) const override;
@@ -96,8 +97,8 @@ class DlBlurImageFilter final : public DlImageFilter {
  private:
   DlScalar sigma_x_;
   DlScalar sigma_y_;
-  std::optional<DlRect> bounds_;
   DlTileMode tile_mode_;
+  std::optional<DlRect> bounds_;
 };
 
 }  // namespace flutter
--- a/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
+++ b/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
@@ -13,6 +13,7 @@
 #include "flutter/display_list/effects/dl_color_source.h"
 #include "flutter/display_list/effects/dl_image_filter.h"
 #include "flutter/display_list/effects/dl_mask_filter.h"
+#include "flutter/display_list/effects/image_filters/dl_blur_image_filter.h"
 #include "flutter/display_list/geometry/dl_path_builder.h"
 #include "flutter/impeller/display_list/aiks_unittests.h"
 
@@ -361,6 +362,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
   auto backdrop_filter =
       DlImageFilter::MakeBlur(999999, 999999, DlTileMode::kClamp);
   builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter.get());
+  builder.Restore();
+
+  ASSERT_TRUE(OpenPlaygroundHere(builder.Build()));
+}
+
+TEST_P(AiksTest, CanRenderBoundedBlurFarFromEdge) {
+  // This test case corresponds to the 1st branch of downsampling, where
+  // the coverage hint is non-null but was ignored during snapshotting.
+
+  auto image = DlImageImpeller::Make(CreateTextureForFixture("kalimba.jpg"));
+
+  DisplayListBuilder builder;
+
+  DlPaint paint;
+  builder.DrawImage(image, DlPoint(0.0, 0.0), DlImageSampling::kNearestNeighbor,
+                    &paint);
+
+  DlPaint save_paint;
+  save_paint.setBlendMode(DlBlendMode::kSrcOver);
+
+  // Subcase 1: When the blur bounds has a non-zero offset.
+  // We still need a clip to trigger the 1st branch of downsampling, therefore
+  // the clip is chosen as an expanded rect of the blur bounds.
+  // The resulting blur is expected to extend a bit beyond rect1, which is
+  // caused by the gaussian blur not being clipped.
+  DlRect rect1 = DlRect::MakeLTRB(50, 80, 233, 180);
+  Point offset = Point(10, 10);
+  builder.Save();
+  builder.ClipRect(rect1.Expand(offset));
+  auto backdrop_filter1 = DlBlurImageFilter::Make(
+      20, 20, rect1.Shift(offset - rect1.GetLeftTop()), DlTileMode::kDecal);
+  builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter1.get());
+  builder.Restore();
+  builder.Restore();
+
+  // Stroke the rect so that we can see where the bounds are.
+  DlPaint stroke_paint(DlColor::kBlack());
+  stroke_paint.setStrokeWidth(1);
+  stroke_paint.setDrawStyle(DlDrawStyle::kStroke);
+  builder.DrawRect(rect1, stroke_paint);
+
+  // Subcase 2: When the blur bounds has a zero offset.
+  DlRect rect2 = DlRect::MakeLTRB(50, 200, 233, 300);
+  builder.ClipRect(rect2);
+  builder.Save();
+  auto backdrop_filter = DlBlurImageFilter::Make(
+      20, 20, rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
+  builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter.get());
+  builder.Restore();
+  builder.Restore();
+
+  ASSERT_TRUE(OpenPlaygroundHere(builder.Build()));
+}
+
+TEST_P(AiksTest, CanRenderBoundedBlurCloseToEdge) {
+  // This test case corresponds to the 2nd branch of downsampling, where
+  // the coverage hint is null or was already used during snapshotting.
+  auto image = DlImageImpeller::Make(CreateTextureForFixture("kalimba.jpg"));
+
+  DisplayListBuilder builder;
+
+  DlPaint paint;
+  builder.DrawImage(image, DlPoint(0.0, 0.0), DlImageSampling::kNearestNeighbor,
+                    &paint);
+
+  DlPaint save_paint;
+  save_paint.setBlendMode(DlBlendMode::kSrcOver);
+
+  // Subcase 1: When the blur bounds has a non-zero offset.
+  // The resulting blur is expected to extend a bit beyond rect1, which is
+  // caused by the gaussian blur not being clipped.
+  DlRect rect1 = DlRect::MakeLTRB(10, 80, 233, 180);
+  builder.Save();
+  auto backdrop_filter1 =
+      DlBlurImageFilter::Make(20, 20, rect1, DlTileMode::kDecal);
+  builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter1.get());
+  builder.Restore();
+  builder.Restore();
+
+  // Stroke the rect so that we can see where the bounds are.
+  DlPaint stroke_paint(DlColor::kBlack());
+  stroke_paint.setStrokeWidth(1);
+  stroke_paint.setDrawStyle(DlDrawStyle::kStroke);
+  builder.DrawRect(rect1, stroke_paint);
+
+  // Subcase 2: When the blur bounds has a zero offset.
+  DlRect rect2 = DlRect::MakeLTRB(10, 200, 233, 300);
+  builder.ClipRect(rect2);
+  builder.Save();
+  auto backdrop_filter = DlBlurImageFilter::Make(
+      20, 20, rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
+  builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter.get());
+  builder.Restore();
   builder.Restore();
 
   ASSERT_TRUE(OpenPlaygroundHere(builder.Build()));
--- a/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
+++ b/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
@@ -453,7 +453,7 @@ TEST_P(AiksTest, CanRenderBoundedBlurCloseToEdge) {
   builder.ClipRect(rect2);
   builder.Save();
   auto backdrop_filter = DlBlurImageFilter::Make(
-      20, 20, rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
+      20, 20, /*bounds=*/rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
   builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter.get());
   builder.Restore();
   builder.Restore();
--- a/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
+++ b/engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
@@ -392,7 +392,8 @@ TEST_P(AiksTest, CanRenderBoundedBlurFarFromEdge) {
   builder.Save();
   builder.ClipRect(rect1.Expand(offset));
   auto backdrop_filter1 = DlBlurImageFilter::Make(
-      20, 20, rect1.Shift(offset - rect1.GetLeftTop()), DlTileMode::kDecal);
+      20, 20, /*bounds=*/rect1.Shift(offset - rect1.GetLeftTop()),
+      DlTileMode::kDecal);
   builder.SaveLayer(std::nullopt, &save_paint, backdrop_filter1.get());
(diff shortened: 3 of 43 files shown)


engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
```diff
@@ -13,17 +13,25 @@ namespace flutter {
 
 class DlBlurImageFilter final : public DlImageFilter {
  public:
-  DlBlurImageFilter(DlScalar sigma_x, DlScalar sigma_y, DlTileMode tile_mode)
-      : sigma_x_(sigma_x), sigma_y_(sigma_y), tile_mode_(tile_mode) {}
+  DlBlurImageFilter(DlScalar sigma_x,
```
gaaclarke, line 45: Add docstring please.  Include the coordinate space of the bounds parameter.

engine/src/flutter/impeller/display_list/paint.cc
```diff
@@ -384,8 +384,8 @@ std::shared_ptr<FilterContents> Paint::MaskBlurDescriptor::CreateMaskBlur(
   texture_contents->SetSamplerDescriptor(descriptor);
   std::shared_ptr<FilterContents> blurred_mask =
       FilterContents::MakeGaussianBlur(FilterInput::Make(mask), sigma, sigma,
-                                       Entity::TileMode::kDecal, style,
-                                       geometry);
+                                       std::nullopt, Entity::TileMode::kDecal,
```
gaaclarke: Please add parameter labels to literals in parameters.

```suggestion
                                       /*bounds=*/std::nullopt, Entity::TileMode::kDecal,
```

engine/src/flutter/impeller/display_list/paint.cc
```diff
@@ -401,7 +401,7 @@ std::shared_ptr<FilterContents> Paint::MaskBlurDescriptor::CreateMaskBlur(
   // blur. The color filter will always be applied on the CPU.
   if (color_source_contents->IsSolidColor()) {
     return FilterContents::MakeGaussianBlur(
-        FilterInput::Make(color_source_contents), sigma, sigma,
+        FilterInput::Make(color_source_contents), sigma, sigma, std::nullopt,
```
gaaclarke: here too

engine/src/flutter/impeller/display_list/paint.cc
```diff
@@ -414,8 +414,8 @@ std::shared_ptr<FilterContents> Paint::MaskBlurDescriptor::CreateMaskBlur(
   /// 2. Blur the mask.
 
   auto blurred_mask = FilterContents::MakeGaussianBlur(
-      FilterInput::Make(mask), sigma, sigma, Entity::TileMode::kDecal, style,
-      color_source_contents->GetGeometry());
+      FilterInput::Make(mask), sigma, sigma, std::nullopt,
```
gaaclarke: here too

engine/src/flutter/impeller/display_list/paint.cc
```diff
@@ -459,7 +459,7 @@ std::shared_ptr<FilterContents> Paint::MaskBlurDescriptor::CreateMaskBlur(
   }
   if (is_solid_color) {
     return FilterContents::MakeGaussianBlur(input, Sigma(blur_sigma.x),
-                                            Sigma(blur_sigma.y),
+                                            Sigma(blur_sigma.y), std::nullopt,
```
gaaclarke: here too

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -147,6 +147,19 @@ Rect MakeReferenceUVs(const Rect& reference, const Rect& rect) {
...
+//   if (a.has_value()) {
+//     if (b.has_value()) {
+//       return a.value().Intersection(b.value());
+//     } else {
+//       return a;
+//     }
+//   } else {
+//     return b;
+//   }
+// }
```
gaaclarke: commented code

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -274,9 +297,17 @@ DownsamplePassArgs CalculateDownsamplePassArgs(
...
+    std::optional<Rect> bounds_uv;
+    if (source_bounds.has_value()) {
+      Rect aligned_bounds = ExpandToDivisible(source_bounds.value(), divisor);
+      bounds_uv = Rect::MakePointBounds(
+          CalculateSnapshotUVs(input_snapshot, aligned_bounds));
+    }
     return {
         .subpass_size = subpass_size,
         .uvs = uvs,
+        .bounds_uv = bounds_uv,  // TODO(dkwingsmt): Untested. Should I?
```
gaaclarke: Unresolved TODO, do you want an issue here?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -335,6 +373,8 @@ fml::StatusOr<RenderTarget> MakeDownsampleSubpass(
   if (pass_args.effective_scalar.x >= 0.5f ||
       (!input_texture->NeedsMipmapGeneration() &&
        input_texture->GetTextureDescriptor().mip_count > 1)) {
+    // TODO(dkwingsmt): For now this branch ignores bounds_uv because I haven't
+    // met such cases.
```
gaaclarke: I don't think this case is rare.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -391,62 +431,69 @@ fml::StatusOr<RenderTarget> MakeDownsampleSubpass(
...
+                                                               renderer,
+                                                           RenderPass& pass) {
+      HostBuffer& data_host_buffer = renderer.GetTransientsDataBuffer();
+
+      pass.SetCommandLabel("Gaussian blur downsample");
+      auto pipeline_options = OptionsFromPass(pass);
+      pipeline_options.primitive_type = PrimitiveType::kTriangleStrip;
+      // The GLES backend conditionally supports decal tile mode, while
+      // decal is always supported for Vulkan and Metal.
+      // Also, `bounds_uv` requires software decal whatsoever.
```
gaaclarke: "Whatsoever"?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -391,62 +431,69 @@ fml::StatusOr<RenderTarget> MakeDownsampleSubpass(
...
+      if ((renderer.GetDeviceCapabilities().SupportsDecalSamplerAddressMode() ||
+           tile_mode != Entity::TileMode::kDecal) &&
+          !pass_args.bounds_uv.has_value()) {
+        pass.SetPipeline(renderer.GetDownsamplePipeline(pipeline_options));
+      } else {
+        pass.SetPipeline(
+            renderer.GetDownsampleSoftwareDecalPipeline(pipeline_options));
+        Rect bounds_rect =
+            pass_args.bounds_uv.value_or(Rect::MakeLTRB(0, 0, 1, 1));
```
gaaclarke: Does this match old behavior when setting the bounds to (0,0,1,1)?  Seems like that is potentially doing clipping that wasn't happening before?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -725,8 +782,12 @@ std::optional<Entity> GaussianBlurFilterContents::RenderFilter(
   // be ignored so must be accounted for in the downsample pass by adding a
   // transparent gutter.
   std::optional<Rect> expanded_coverage_hint;
+  std::optional<Rect> shifted_bounds = bounds_;
   if (coverage_hint.has_value()) {
     expanded_coverage_hint = coverage_hint->Expand(blur_info.local_padding);
+    if (bounds_.has_value()) {
+      shifted_bounds = bounds_->Shift(blur_info.local_padding);
+    }
```
gaaclarke: Probably worth expanding the comment above this block now.

engine/src/flutter/impeller/entity/entity_unittests.cc
```diff
@@ -1061,14 +1061,14 @@ TEST_P(EntityTest, GaussianBlurFilter) {
     switch (selected_pass_variation) {
       case 0:
         blur = std::make_shared<GaussianBlurFilterContents>(
-            blur_sigma_x.sigma, blur_sigma_y.sigma,
+            blur_sigma_x.sigma, blur_sigma_y.sigma, std::nullopt,
```
gaaclarke: This would be a good place to expand the arguments to be able to control this.  It's not an automated test though so it's optional.  It would be helpful for you probably to mess around with it.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -19,7 +19,16 @@ uniform KernelSamples {
...
 kernel_samples;
 
+uniform FragInfo {
+  float bounded;
+  vec4 bounds_uv;
+}
+frag_info;
+
 f16vec4 Sample(f16sampler2D tex, vec2 coords) {
+  if (frag_info.bounded == 1.0) {
```
gaaclarke: Instead of introducing a conditional here, I think we should have a different shader.  It will be the same value across all fragments.  The downside is the cost of yet another pipeline.  Another option is a push constant.  I can't remember if we have an example of that elsewhere.  That would be ideal.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -19,7 +19,16 @@ uniform KernelSamples {
 }
 kernel_samples;
 
+uniform FragInfo {
+  float bounded;
```
gaaclarke: We should have int values in all our targets so you don't have to use a float as a boolean.

gaaclarke (changes requested):
Generally I'm good with the approach.  It's on a good trajectory.  You need to add a golden test to aiks_dl_blur_unittests.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: Rephrased. Can you see if it's clearer?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: It should match the old behavior.

The `if` branch works in the following way: (I've added this new commment above this statement)
```cpp
      // If the blur is unbounded, then the pipeline is selected based on
      // whether the platform supports decal tile mode: The GLES backend
      // conditionally supports decal tile mode, while decal is always supported
      // for Vulkan and Metal.
      //
      // If the blur is bounded, software decal is always needed.
```

So there are two ways to enter the `else` branch. The first way is that the blur is unbounded, and the `else` branch is chosen because of GLES, corresponding to `GetDownsampleTextureGlesPipeline` before the PR, whose behavior was defined by the `Sample` as follows:
```glsl
vec4 Sample(vec2 uv) {
#ifdef SUPPORTS_DECAL   // 0 for GLES
  return texture(texture_sampler, uv, float16_t(kDefaultMipBias));
#else
  if ((uv.x < 0 || uv.y < 0 || uv.x > 1 || uv.y > 1)) {
    return vec4(0);
  } else {
    return texture(texture_sampler, uv, float16_t(kDefaultMipBias));
  }
#endif
}
```
The `Sample` function always checks the coord against (0,0,1,1).

The second way is that the blur is bounded. In this case the coord is checked against the `bounds`.

engine/src/flutter/impeller/entity/entity_unittests.cc
dkwingsmt: I've attempted to mess around it but it's kind of complicated considering all the transform matrices and texture stuff. I think it'd be better done in a separate PR.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
dkwingsmt: I've moved it to a specialized constant as per suggested (if I understand correctly).

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
dkwingsmt: I've moved it to a specialized constant as per suggested (if I understand correctly). But the other existing const uses a float. Do you think I should still use a float?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: Do you think backdrop blurs can reach this branch at all? I've investigated a bit and it turns out that `input_snapshot.transform.Equals(snapshot_entity.GetTransform())` is not true, even though the comment above said that this branch should have been for backdrop blurs.

If I can somehow make it reach here, of course we wouldn't need a TODO. If we can't find a way to go this branch, I think we're better off not implement bounded blur for this branch at all and note it in the doc.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: I'm not familiar with this part. Can you suggest a way where I can trigger mipmap reuse and test this branch?

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -40,5 +49,10 @@ void main() {
                           v_texture_coords + kernel_samples.sample_data[i].xy);
   }
 
-  frag_color = total_color;
+  if (bounded_blur == 1.0) {
+    frag_color =
+        (total_color.w != 0) ? (total_color / total_color.w) : total_color;
```
gaaclarke: This is where the ci failure is, switch 0 to 0.0 and give that a shot.

gaaclarke:
I don't know if you are ready yet, but I'm going to hold off reviewing again until we get this PR green through CI so we can see the goldens.  Let me know if you need any help getting it through.

mdebbar:
web part looks good!

dkwingsmt:
gaaclarke : I'm aware that the golden test isn't representative. Actually some maths of the shaders aren't right, and I've been trying to figure them out for days. I'll let you know when I made progress.

gaaclarke:
> I'm aware that the golden test isn't representative.

Thanks for the heads up.  Let me know when you need me to look again.

dkwingsmt:
I've completed the PR and it should be ready to review. Here are notable changes:
1. `gaussian.frag` is now controlled by a specialized constant `bounded_blur`.
2. Both branches of `CalculateDownsamplePassArgs` have been implemented and each tested by a golden test. Each golden test verifies two subcases: One has a non-zero offset and unclipped, and the other has a zero offset and clipped. (The latter one is possibly the only case that matters in practice.)
3. Regarding the two branches of `MakeDownsampleSubpass`, I've decided that the first branch, which reuses existing mipmap renderers, does not suit our case, since a bounded blur requires special treatment during downsampling.
4. I've added an interactive playground for bounded blurs.

https://github.com/user-attachments/assets/91f805a5-3de2-47ec-a969-dfac67cd4923

5. I disabled the kernel lerp hack for bounded blurs. This was to reduce the wave-like artifacts close to edges (see video below).

https://github.com/user-attachments/assets/b8109f42-c449-468f-86ac-a2f2c114f8f3

  * Known issue: These artifacts are greatly reduced, although not completely gone, without the lerp hack. It becomes much more noticeable (but still better than with the lerp hack) if the texture is drawn with a non-zero offset. Fortunately in practice we should almost only draw texture with a zero-artifacts. I've investigated this for days in vain, and I've decided that the result is good enough for zero offset texture anyway.

gaaclarke:
It's been a second since I looked at this.  I think you may have answered this in the past but it wasn't captured.  Can you remind me why we couldn't implement this by tweaking the downsample pass to conditionally the crop the gutter and set the tilemode to clamp?

So for example, in the case where we are blurring something that is running out of content it looks like this:

input:
```
[####]
[####]
[####]
[####]
```
downsample output:
```
[.....]
[.###.]
[.###.]
[.###.]
[.....]
```

Then the blur is executed on it.  If instead the downsample output was similar to the input with a tilemode set to clamp, wouldn't that give you the same result?

gaaclarke:
Note: I also asked tong offline if the following golden image was as expected and he confirmed it was.  I was wondering about the blurred region not being in the drawn rect.

38afaf598978e996a5aebe3101c4d675

dkwingsmt:
@gaaclarke Regarding the first comment: why we couldn't implement this by tweaking the downsample pass to conditionally the crop the gutter and set the tilemode to clamp?

I'm not sure if I correctly get the question. Do you mean if we can tweek the texture boundary (such as the `uv` variable) so that we will only operate on a reduced surface anyway? 

I think this is probably the ultimate direction, but I decided to not go with this direction for the time being. It will modify much more variables that the current approach, which is already very hard to figure out maths for. Furthermore, there are potential risks of actually degrading the result, especially when there are some seemingly related comments in the code:
```
  // TODO(gaaclarke): The padding could be removed if we know it's not needed or
  //   resized to account for the expanded_clip_coverage. There doesn't appear
  //   to be the math to make those calculations though. The following
  //   optimization works, but causes a shimmer as a result of
  //   https://github.com/flutter/flutter/issues/140193 so it isn't applied.
  //
  //   !input_snapshot->GetCoverage()->Expand(-local_padding)
  //     .Contains(coverage_hint.value()))
```

In general, I think it's better to implement the feature with the minimal changes for now and we can explore optimization in the future.

dkwingsmt:
Regarding the 2nd question (the golden file), I'm putting the answer here for the record: As noted in the comment of the unit tests, the result is expected. Although the blur is expected to only source in-bound pixels, the gaussian blur will blur it out since the entire texture is larger than that. Typically we will clip it. This subcase is just to verify that the offset works correctly.

engine/src/flutter/display_list/effects/dl_image_filter.cc
```diff
@@ -12,7 +12,7 @@ namespace flutter {
 std::shared_ptr<DlImageFilter> DlImageFilter::MakeBlur(DlScalar sigma_x,
                                                        DlScalar sigma_y,
                                                        DlTileMode tile_mode) {
-  return DlBlurImageFilter::Make(sigma_x, sigma_y, tile_mode);
+  return DlBlurImageFilter::Make(sigma_x, sigma_y, std::nullopt, tile_mode);
```
gaaclarke:
```suggestion
  return DlBlurImageFilter::Make(sigma_x, sigma_y, /*bounds=*/std::nullopt, tile_mode);
```

engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
```diff
@@ -13,17 +13,50 @@ namespace flutter {
...
+   * - If [bounds] is provided (i.e., not std::nullopt), the filter performs a
+   * "bounded blur." This means the blur kernel will only sample pixels
+   * from *within* this rectangle, treating all pixels outside of it as
+   * transparent. This mode is used to implement high-fidelity iOS-style blurs.
+   *
+   * - If [bounds] is std::nullopt, a standard, unbounded blur is performed,
+   * with edge behavior defined by [tile_mode].
+   *
+   * The [bounds] rectangle must be specified in the current coordinate space
+   * of the canvas (i.e., it is subject to the canvas's current transform).
```
gaaclarke: Where is the origin in this case?  top-left of the object being rendered or top-left of the canvas?

engine/src/flutter/impeller/compiler/shader_lib/impeller/texture.glsl
```diff
@@ -138,6 +138,17 @@ f16vec4 IPHalfSampleDecal(f16sampler2D texture_sampler, vec2 coords) {
   return texture(texture_sampler, coords, float16_t(kDefaultMipBias));
 }
 
+/// Sample a texture with decal tile mode with custom bounds.
+f16vec4 IPHalfSampleDecalBounded(f16sampler2D texture_sampler,
```
gaaclarke: It's worth documenting how the bounds are packed into the vec4.

engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
```diff
@@ -366,6 +367,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
...
+  // We still need a clip to trigger the 1st branch of downsampling, therefore
+  // the clip is chosen as an expanded rect of the blur bounds.
+  // The resulting blur is expected to extend a bit beyond rect1, which is
+  // caused by the gaussian blur not being clipped.
+  DlRect rect1 = DlRect::MakeLTRB(50, 80, 233, 180);
+  Point offset = Point(10, 10);
+  builder.Save();
+  builder.ClipRect(rect1.Expand(offset));
+  auto backdrop_filter1 = DlBlurImageFilter::Make(
+      20, 20, rect1.Shift(offset - rect1.GetLeftTop()), DlTileMode::kDecal);
```
gaaclarke:
```suggestion
      20, 20, /*bounds=*/rect1.Shift(offset - rect1.GetLeftTop()), DlTileMode::kDecal);
```

engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
```diff
@@ -366,6 +367,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
...
+
+  // Subcase 1: When the blur bounds has a non-zero offset.
+  // We still need a clip to trigger the 1st branch of downsampling, therefore
+  // the clip is chosen as an expanded rect of the blur bounds.
+  // The resulting blur is expected to extend a bit beyond rect1, which is
+  // caused by the gaussian blur not being clipped.
+  DlRect rect1 = DlRect::MakeLTRB(50, 80, 233, 180);
+  Point offset = Point(10, 10);
+  builder.Save();
+  builder.ClipRect(rect1.Expand(offset));
```
gaaclarke: The golden confuses me a bit still, it looks like we are offsetting in x and y directions but the output just looks like it is expanded in the x direction.

engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
```diff
@@ -366,6 +367,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
...
+  stroke_paint.setStrokeWidth(1);
+  stroke_paint.setDrawStyle(DlDrawStyle::kStroke);
+  builder.DrawRect(rect1, stroke_paint);
+
+  // Subcase 2: When the blur bounds has a zero offset.
+  DlRect rect2 = DlRect::MakeLTRB(50, 200, 233, 300);
+  builder.ClipRect(rect2);
+  builder.Save();
+  auto backdrop_filter = DlBlurImageFilter::Make(
+      20, 20, rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
```
gaaclarke:
```suggestion
      20, 20, /*bounds=*/rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
```

engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
```diff
@@ -366,6 +367,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
...
+  DlPaint save_paint;
+  save_paint.setBlendMode(DlBlendMode::kSrcOver);
+
+  // Subcase 1: When the blur bounds has a non-zero offset.
+  // The resulting blur is expected to extend a bit beyond rect1, which is
+  // caused by the gaussian blur not being clipped.
+  DlRect rect1 = DlRect::MakeLTRB(10, 80, 233, 180);
+  builder.Save();
+  auto backdrop_filter1 =
+      DlBlurImageFilter::Make(20, 20, rect1, DlTileMode::kDecal);
```
gaaclarke:
```suggestion
      DlBlurImageFilter::Make(20, 20, /*bounds=*/rect1, DlTileMode::kDecal);
```

engine/src/flutter/impeller/display_list/aiks_dl_blur_unittests.cc
```diff
@@ -366,6 +367,99 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
...
+  stroke_paint.setStrokeWidth(1);
+  stroke_paint.setDrawStyle(DlDrawStyle::kStroke);
+  builder.DrawRect(rect1, stroke_paint);
+
+  // Subcase 2: When the blur bounds has a zero offset.
+  DlRect rect2 = DlRect::MakeLTRB(10, 200, 233, 300);
+  builder.ClipRect(rect2);
+  builder.Save();
+  auto backdrop_filter = DlBlurImageFilter::Make(
+      20, 20, rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
```
gaaclarke:
```suggestion
      20, 20, /*bounds=*/rect2.Shift(-rect2.GetOrigin()), DlTileMode::kDecal);
```

engine/src/flutter/impeller/display_list/dl_unittests.cc
```diff
@@ -721,6 +721,57 @@ TEST_P(DisplayListTest, CanDrawBackdropFilter) {
   ASSERT_TRUE(OpenPlaygroundHere(callback));
 }
 
+TEST_P(DisplayListTest, CanDrawBoundedBlur) {
```
gaaclarke, line 724: 👍 hoped this helped during development

engine/src/flutter/impeller/entity/contents/filters/filter_contents.h
```diff
@@ -44,6 +44,7 @@ class FilterContents : public Contents {
       const FilterInput::Ref& input,
       Sigma sigma_x,
       Sigma sigma_y,
+      std::optional<Rect> bounds,
```
gaaclarke: Worth adding a link to the other doc you wrote above for bounds.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.h
```diff
@@ -57,10 +58,12 @@ class GaussianBlurFilterContents final : public FilterContents {
  public:
   explicit GaussianBlurFilterContents(Scalar sigma_x,
                                       Scalar sigma_y,
+                                      std::optional<Rect> bounds,
```
gaaclarke: It's worth linking to your other documentation here.

engine/src/flutter/impeller/entity/entity_unittests.cc
gaaclarke: Did you get a chance to test how the bounded blur reacts to transforms?

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -19,7 +20,15 @@ uniform KernelSamples {
 }
 kernel_samples;
 
+uniform FragInfo {
+  vec4 bounds_uv;
+}
+frag_info;
+
 f16vec4 Sample(f16sampler2D tex, vec2 coords) {
+  if (bounded_blur == 1.0) {
```
gaaclarke: You are doing a conditional for each Sample call.  You should do the conditional once.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -40,5 +49,10 @@ void main() {
                           v_texture_coords + kernel_samples.sample_data[i].xy);
   }
 
-  frag_color = total_color;
+  if (bounded_blur == 1.0) {
+    frag_color =
+        (total_color.w != 0.0) ? (total_color / total_color.w) : total_color;
+  } else {
+    frag_color = total_color;
+  }
```
gaaclarke: Why is this unpremultiplying colors conditionally?  Also we have a function for this, `IPUnpremultiply`.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4195,8 +4195,18 @@ abstract class ImageFilter {
   ImageFilter._(); // ignore: unused_element
 
   /// Creates an image filter that applies a Gaussian blur.
-  factory ImageFilter.blur({double sigmaX = 0.0, double sigmaY = 0.0, TileMode? tileMode}) {
-    return _GaussianBlurImageFilter(sigmaX: sigmaX, sigmaY: sigmaY, tileMode: tileMode);
+  factory ImageFilter.blur({
+    double sigmaX = 0.0,
+    double sigmaY = 0.0,
+    TileMode? tileMode,
+    Rect? bounds,
```
gaaclarke, line 4290: Needs docstring, this is what users see.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4367,10 +4383,12 @@ class _GaussianBlurImageFilter implements ImageFilter {
...
 
   @override
-  String get _shortDescription => 'blur($sigmaX, $sigmaY, $_modeString)';
+  String get _shortDescription => 'blur($sigmaX, $sigmaY, $_modeString${_boundsString()})';
+
+  String _boundsString() => bounds == null ? '' : ', bounds: $bounds';
 
   @override
-  String toString() => 'ImageFilter.blur($sigmaX, $sigmaY, $_modeString)';
+  String toString() => 'ImageFilter.blur($sigmaX, $sigmaY, $_modeString${_boundsString()})';
```
gaaclarke, line 4479: did we lose modeString?

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/renderer.dart
```diff
@@ -156,7 +156,11 @@ class CanvasKitRenderer extends Renderer {
     double sigmaX = 0.0,
     double sigmaY = 0.0,
     ui.TileMode? tileMode,
-  }) => CkImageFilter.blur(sigmaX: sigmaX, sigmaY: sigmaY, tileMode: tileMode);
+    ui.Rect? bounds,
+  }) =>
+      // TODO(dkwingsmt): `bounds` is currently not implemented in CanvasKit.
```
gaaclarke, line 161: Should we throw an exception if it's used?

gaaclarke (commented):
Okay, I've done a review on everything but the gaussian blur contents file.  Can you expand your interactive test to include rotating and scaling the blur?  I want to make sure that the bounds work as expected in that case.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -505,9 +599,18 @@ fml::StatusOr<RenderTarget> MakeBlurSubpass(
                 linear_sampler_descriptor));
         GaussianBlurVertexShader::BindFrameInfo(
             pass, data_host_buffer.EmplaceUniform(frame_info));
+        KernelSamples full_kernel_samples = GenerateBlurInfo(blur_info);
+        // For bounded blurs, kernel samples must not be lerped, as this can
+        // introduce wave-like artifacts at the edges of the bounds. This is
+        // likely due to increased sensitivity to pixel accuracy caused by
+        // opacity normalization.
```
gaaclarke: I think we should try to address this.  Without the lerp hack the blurs are much more costly.  4x more costly since it doubles in both dimensions?  A lot of work has gone into making our blurs fast and apps are very sensitive to it.  What have you tried so far?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -197,12 +197,36 @@ Scalar FloorToDivisible(Scalar val, Scalar divisor) {
...
+  /// The bounds used for the downsampling pass of a bounded blur, in the same
+  /// UV space as the texture input of the downsampling pass.
+  ///
+  /// During downsampling, out-of-bound pixels are treated as transparent.
+  std::optional<Rect> downsample_uv_bounds;
+  /// The bounds used for the blurring pass of a bounded blur, in the same UV
+  /// space as the output of the downsampling pass.
+  ///
+  /// During downsampling, out-of-bound pixels are treated as transparent.
+  std::optional<Rect> blur_uv_bounds;
```
gaaclarke: Can you mention why we track these 2 separately since the input is just one bounds.

gaaclarke (commented):
Okay I gave everything a look now!

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -735,11 +841,20 @@ std::optional<Entity> GaussianBlurFilterContents::RenderFilter(
...
+          snapshot_entity.GetTransform() * entity.GetTransform().Invert();
+    }
+    if (expanded_coverage_hint.has_value()) {
+      source_expanded_coverage_hint =
+          expanded_coverage_hint->TransformBounds(source_transform.value());
+    }
+    if (bounds_.has_value()) {
+      source_bounds = bounds_->TransformBounds(source_transform.value());
+    }
   }
```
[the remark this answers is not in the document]
gaaclarke: This is an interesting observation, hopefully adding rotation and scaling to your test will help validate this.  Coordinate spaces get very confusing in this code, I've tried to make it better than how it was.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: oops it's there, just need a comma.

```suggestion
  String toString() => 'ImageFilter.blur($sigmaX, $sigmaY, $_modeString, ${_boundsString()})';
```

mdebbar (approved):
web part looks good

engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
dkwingsmt: If I understand correctly, it aligns with what you described as "top-left of the object being rendered". However, I think the current explanation about the canvas's CTM already clarifies the origin. Basically, if the bounds rectangle is defined from (a, b) to (c, d), the respective corners on canvas are subject to CTM in a similar way as all other drawing methods such as `builder.DrawLine()`. Do you still think the origin needs a special mention?

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: I've revised the algorithm and it only uses one bounds now.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
dkwingsmt: This has been addressed. The bounded blur now uses the kernel hack perfectly.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/renderer.dart
dkwingsmt: No. The bounded blur should fall back to the normal blur. Because:
* The unbounded blur is an acceptable result.
* It aligns with the behavior before this PR. 
* The framework will send a drawing op of bounded blur to the engine regardless of the platform, and the engine should be able to handle the ops at best effort instead of crashing.

engine/src/flutter/lib/ui/painting.dart
dkwingsmt: No, I think the current version is correct. I intentionally made it so so that "bounds: " is hidden when null. What do you think?

engine/src/flutter/impeller/compiler/shader_lib/impeller/texture.glsl
dkwingsmt: This function has been removed.

engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -748,6 +878,12 @@ std::optional<Entity> GaussianBlurFilterContents::RenderFilter(
     return std::nullopt;
   }
 
+  std::optional<Quad> source_bounds;
+  if (bounds_.has_value()) {
+    Quad global_bounds = bounds_->GetTransformedPoints(effect_transform);
+    source_bounds = snapshot_entity.GetTransform().Transform(global_bounds);
```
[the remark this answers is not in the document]
dkwingsmt: Reimplemented it in a way that's easier to understand.

engine/src/flutter/impeller/entity/shaders/texture_downsample_gles.frag
```diff
@@ -3,3 +3,11 @@
 // found in the LICENSE file.
 
 #include <downsample.glsl>
+
+vec4 Sample(vec2 uv) {
+  if ((uv.x < 0 || uv.y < 0 || uv.x > 1 || uv.y > 1)) {
```
dkwingsmt, line 8: Alternative option: combine `_gles.frag` and `_bounded.frag`, since the decal bound can also be implemented as `quad_line_params`, saving a pipeline. However, I'd guess that this is a performance regression for unbounded GLES.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
```diff
@@ -40,5 +46,9 @@ void main() {
                           v_texture_coords + kernel_samples.sample_data[i].xy);
   }
 
-  frag_color = total_color;
+  if (frag_info.unpremultiply > 0.5) {
```
dkwingsmt, line 49: Alternative option: Split the unpremultiply shader into another pipeline. However, I think the current implementation only adds a branching per pixel to `gaussian.frag`, which should be acceptable.

dkwingsmt (commented):
@gaaclarke I've overhauled the algorithm, and it should be correct now. The PR is ready for another review.

As shown in the following video, the blur bound correctly applies canvas transform, without any rippling artifacts at the edge.

https://github.com/user-attachments/assets/e8abcdee-7da1-4e4b-a763-654bc5734e0a

The implementation uses a simple and elegant algorithm that allows kernel lerp hack for all cases. The detail is documented at the `GaussianBlurFilterContents` class.

The golden unit test has also been updated, which should have addressed previous concerns.

image

engine/src/flutter/impeller/entity/entity_unittests.cc
dkwingsmt: Yes. As shown in the playground demo (in the latest comment) and the golden test, the blur bound respects the current transform matrix.

engine/src/flutter/impeller/entity/shaders/filters/gaussian.frag
dkwingsmt: Also, although an earlier comment of you discouraged using float as boolean, I encountered an error saying that only float is allowed (which led me to https://github.com/flutter/flutter/pull/175458/files#diff-4cedd9ef3102e6a651299999f6b97b46ab70363b2608d720b43d93410343df08R410 ). I'm not sure what it is. Let me know if there's something I can do to use non-float here.

engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
```diff
@@ -13,17 +13,49 @@ namespace flutter {
...
+   * iOS-style bounded blurs.
+   *
+   * The [bounds] rectangle is specified in the canvas's current coordinate
+   * space (it is affected by the current transform). Consequently, the bounds
+   * may not be axis-aligned in canvas coordinates.
+   */
+  DlBlurImageFilter(DlScalar sigma_x,
+                    DlScalar sigma_y,
+                    std::optional<DlRect> bounds,
+                    DlTileMode tile_mode)
```
gaaclarke: This would be better off as an `std::variant<DlRect, DlTileMode>`, no?  It's my understanding that tile_mode is ignored if the rect is set and decal is used?  This would get rid of a lot of this PR too.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: Got it, yea this is good.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/renderer.dart
gaaclarke: sgtm, one nit though, I wouldn't expect an exception to crash

gaaclarke (commented):
Here's a review of everything except the gaussian blur changes.  I'll look at that next.  I just had a question about the semantics of time_mode when bounds is set.

engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
gaaclarke: Can you respond to comments like this with "done" if you think they are addressed?  This one looks good to me.

engine/src/flutter/display_list/effects/image_filters/dl_blur_image_filter.h
```diff
@@ -13,17 +13,49 @@ namespace flutter {
...
+   * @param bounds An optional rectangle that enables "bounded blur" mode.
+   * @param tile_mode The tile mode used for sampling pixels at the edges
+   * when performing a standard, unbounded blur.
+   *
+   * If [bounds] is std::nullopt, a standard Gaussian blur is applied
+   * and edge sampling behavior is determined by [tileMode].
+   *
+   * If [bounds] is not std::nullopt, the filter performs a "bounded blur": the
+   * blur kernel only samples pixels from inside the rectangle, pixels outside
+   * the rectangle are treated as transparent, and the resulting pixels are
```
gaaclarke: (The github vscode plugin lost some of my comments from the last review =(

This documentation seems to conflict with the dart documentation.  Do we respect the tile mode when a bounds is supplied or do we treat it as decal?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4236,8 +4236,39 @@ abstract class ImageFilter {
...
+  /// iOS-style bounded blurs.
+  ///
+  /// The `bounds` rectangle is specified in the canvas's current coordinate
+  /// space (it is affected by the current transform). Consequently, the bounds
+  /// may not be axis-aligned in canvas coordinates.
+  factory ImageFilter.blur({
+    double sigmaX = 0.0,
+    double sigmaY = 0.0,
+    TileMode? tileMode,
+    Rect? bounds,
```
gaaclarke, line 4290: If I'm correct in understanding that tileMode is ignored when a bounds is supplied I think we should introduce a new parameter tileBehavior that will be a TileMode or a Rect and we deprecate the old parameter.  That will boost discoverability of the new feature.

gaaclarke (commented):
I checked out the guassian blur changes too.  Looks awesome.  This is much cleaner and easier to understand and doesn't have the artifacting.  Thanks tong!

engine/src/flutter/lib/ui/painting.dart
gaaclarke: The code could be something like this since the bounded blur seems to be an extension of the decal behavior.
```dart
sealed class TileMode {
  static TileMode get decal => _Decal(null);
  static TileMode boundedDecal(Rect rect) => _Decal(rect);
  static TileMode get mirror => _Mirror();
};
final class _Decal extends TileMode {
  const _Decal(this.bounds);
  final Rect? bounds;
}
...
```

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4269,20 +4269,20 @@ abstract class ImageFilter {
...
+  /// The bounded mode prevents color bleeding from content adjacent to the
+  /// bounds into the blurred area, and is typically used when the blur must be
+  /// strictly contained within a clipped region, such as for iOS-style frosted
+  /// glass effects.
   ///
   /// The `bounds` rectangle is specified in the canvas's current coordinate
-  /// space (it is affected by the current transform). Consequently, the bounds
-  /// may not be axis-aligned in canvas coordinates.
+  /// space and is affected by the current transform; consequently, the bounds
+  /// may not be axis-aligned in the final canvas coordinates.
```
gaaclarke, line 4285: No action required: I wish we had a canonical word for this coordinate space.  I typically call it local coordinate space.

gaaclarke (approved):
lgtm, thanks for all the following up and clarifying.  I'm happy with were we ended up.

engine/src/flutter/lib/ui/painting.dart
dkwingsmt: Yeah honestly I would call it "local coordinate" but I don't think it's clear enough for many people, which matters more.

Issue #166944, opened 2025-04-10
[Impeller] Provide an API for loading Shader at runtime (and tooling support for producing them)


Hellomik2002:
### Use case

We would like to be able to dynamically load shaders. Like in maps or something dynamically

### Proposal

Big thanks to @bdero and all Flutter team. Your job is one of the important and awesome tools. It gives new opportunities and opens a lot of doors.

It would be good to add a shader dynamically at runtime at the current moment I added some function to engine but it should be reviewed. I will continue working with impeller and package to generate shaders dynamically

Hellomik2002:
https://github.com/flutter/flutter/pull/166946

Hellomik2002:
I need to use impeller/compiler to generate shaders. In theory, this shouldn't be too difficult, but I don't fully understand how Flutter handles the build compilation process. Additionally, I'm unsure how to compile it correctly for target devices. :((

jonahwilliams:
We'd need to add version metadata to the shader bundles and expose the impellerc compiler for this to be practical.

Hellomik2002:
I have read how GN works and am planning to finish the code with impeller compiler in 7-14 days. That will be not ideal and temporary solution.
I will try to do some research. I don't quite understand how it should be, it takes time, I think for me to understand.

Thanks for your time. I will try my best

Hellomik2002:
I found out how shaders perform registration and how they detect if a shader has already been registered. However, if registration needs to happen at runtime, some changes are necessary. For now, I have added an `IsFirstTime` property to  shader class (engine/src/flutter/lib/gpu/shader.cc shader.h), so that the Impeller engine knows whether it needs to register or unregister the shader.

I would like to know how the Flutter team would ideally approach this situation. I will provide a link to a minimal working example here.  It is hard to duplicate, because I didn't documented but it works.
To build the shaders at runtime, you need to compile working dynamic library of Impeller compiler.  I gave raw solution that works (By the way, At the current moment size of file is 12MB, it can be reduced for each platform - iOS/Android/macOS from about 3-8 MB)  
https://github.com/Hellomik2002/flutter/tree/runtime_shader_not_ready_compiler

Hellomik2002:
Also I think shader generator have to be seperated and will work as seperate package

PR #175470, opened 2025-09-17
Adds Impellerc flatbuffer format versioning.

### LINKED ISSUE #172899: Add versioning and validation metadata to Impeller shader blobs to ensure forward compatibility
### Use case

Runtime Shaders

### Proposal

Currently, shader blobs emitted by `impellerc` do not include any versioning or validation metadata. This creates a risk when:

- Applications persist these blobs (e.g., to ship precompiled shaders),
- But `impellerc` or the Impeller runtime changes assumptions about layout, encoding, or expectations.

Such mismatches could lead to runtime crashes, undefined behavior, or difficult-to-diagnose bugs due to incompatible blobs being loaded.

This issue was raised during review of [flutter/flutter#168294](https://github.com/flutter/flutter/pull/168294), which introduces support for loading precompiled Impeller shaders from in-memory blobs.

The same problem exists with the fragment program API, which also accepts external shader inputs.

## Proposal

We should introduce in `impellerc`:

- A version tag (semantic or numeric) embedded in each shader blob
- A validation header or magic number for early failure detection
- Possibly a checksum or hash for corruption detection
- Runtime blob consumers (e.g., `RuntimeStage`, fragment program loader) should verify this metadata before accepting the blob

By doing this once in `impellerc`, we ensure both the mobile APIs and fragment program APIs avoid future forward-compatibility issues.

---

**Related PR:** https://github.com/flutter/flutter/pull/168294

timmaffett:
Fixes #172899

This PR adds a format_version number to the root table of the 3 flat buffer formats that impellerc creates; runtime stages, shader bundles and shader archives.  
It also adds a 'format_version' key to the root table of the shader bundle json format for completeness (Though it could be argued that the json format does not require this, I feel it may still be useful there).

This allows for format compatibility checks with future versions of the table as @chinmaygarde expressed concerns about in #168294 and @jonahwilliams mentions in #166944 .
The current code only checks for compatibility with the current version, but this mechanism allows for backwards compatibility in the future and not handling flat buffers with formats with new version number.
The big win here is that format_version numbers allow us to abort parsing anything that we may not understand that could possible cause problems, corruptions or crashes.  The changes are simple and straightforward and I feel they lay a reasonable foundation for version handling for any future changes.

The version numbers themselves are defined within the `runtime_stage_types.fbs`, `shader_bundle.fbs` and `shader_archive.fbs` where I feel they should be as they can directly be incremented if any changes to the flat buffer format are made.  Mechanisms for handling of backwards compatibility in the future should be straightforward to make in the corresponding c++ code allowing for the continued support of older (lower) version numbers if so desired.

This addresses #172899 and the other concerns mentioned in the issues/pr's mentioned above. 

I plan on including my tests for exclusion of non-matching versions - Existing test verify that the version checking confirms matching versions.  I wanted to get this out there for discussion before spending more time on this in case this approach is rejected outright. 
   
I have a companion PR which adds FragmentProgram.fromBytes() to the engine (c++ and canvaskit/wasm). I wanted to separate the impellerc output versioning into it's own (this) PR.

  
## Pre-launch Checklist

- [X] I signed the [CLA].
- [FORTHCOMING] I added new tests to check the change I am making, or this PR is [test-exempt].


engine/src/flutter/impeller/runtime_stage/runtime_stage.cc
```diff
@@ -69,6 +69,19 @@ RuntimeStage::Map RuntimeStage::DecodeRuntimeStages(
   }
 
   auto raw_stages = fb::GetRuntimeStages(payload->GetMapping());
+  if (!raw_stages) {
+    return {};
+  }
+
+  const auto version = raw_stages->format_version();
```
gaaclarke: please remove type deduction

https://google.github.io/styleguide/cppguide.html#Type_deduction

engine/src/flutter/impeller/runtime_stage/runtime_stage.cc
```diff
@@ -69,6 +69,19 @@ RuntimeStage::Map RuntimeStage::DecodeRuntimeStages(
...
+  const auto expected =
+      static_cast<uint32_t>(fb::RuntimeStagesFormatVersion::kVersion);
+  if (version != expected) {
+    VALIDATION_LOG << "Unsupported runtime stages format version: " << version
+                   << ". Expected: " << expected;
```
gaaclarke:
```suggestion
                   << ", expected: " << expected;
```

engine/src/flutter/impeller/shader_bundle/shader_bundle.fbs
```diff
@@ -101,6 +108,7 @@ table Shader {
 }
 
 table ShaderBundle {
+  format_version: uint32;  // This is intended to hold ShaderBundleFormatVersion.kVersion.
```
gaaclarke: Please add at the end of the table to avoid breaking compatibility with existing blobs.

engine/src/flutter/impeller/shader_archive/shader_archive.fbs
```diff
@@ -17,6 +24,7 @@ table ShaderBlob {
 }
 
 table ShaderArchive {
+  format_version: uint32; // This is intended to hold ShaderArchiveFormatVersion.kVersion.
```
gaaclarke: Same here, add at the end to void breaking existing blobs.

engine/src/flutter/testing/dart/fragment_shader_test.dart
```diff
@@ -25,7 +25,7 @@ void main() async {
     expect(rawData is Map<String, Object?>, true);
 
     final Map<String, Object?> data = rawData! as Map<String, Object?>;
-    expect(data.keys.toList(), <String>['sksl']);
+    expect(data.keys.toList(), <String>['format_version', 'sksl']);
```
gaaclarke, line 28: This is testing that the serialized blobs have the field, there are not tests that handle parsing it.

engine/src/flutter/impeller/shader_archive/shader_archive.cc
```diff
@@ -42,6 +42,15 @@ ShaderArchive::ShaderArchive(std::shared_ptr<fml::Mapping> payload)
     return;
   }
 
+  const auto version = shader_archive->format_version();
+  const auto expected =
+      static_cast<uint32_t>(fb::ShaderArchiveFormatVersion::kVersion);
+  if (version != expected) {
+    VALIDATION_LOG << "Unsupported shader archive format version: " << version
+                   << ". Expected: " << expected;
```
gaaclarke:
```suggestion
                   << ", expected: " << expected;
```

engine/src/flutter/lib/gpu/shader_library.cc
```diff
@@ -185,6 +185,15 @@ fml::RefPtr<ShaderLibrary> ShaderLibrary::MakeFromFlatbuffer(
     return nullptr;
   }
 
+  const auto version = bundle->format_version();
+  const auto expected = static_cast<uint32_t>(
+      impeller::fb::shaderbundle::ShaderBundleFormatVersion::kVersion);
+  if (version != expected) {
+    VALIDATION_LOG << "Unsupported shader bundle format version: " << version
+                   << ". Expected: " << expected;
```
gaaclarke:
```suggestion
                   << ", expected: " << expected;
```

gaaclarke (commented):
Looking good to me.  I'd like to see tests that cover parsing as well as serializing the version number.  I also wonder if relying on VALIDATION_LOG is the right thing.  Wouldn't we want to surface these errors to users in the form on dart exceptions?

gaaclarke:
> I also wonder if relying on VALIDATION_LOG is the right thing. Wouldn't we want to surface these errors to users in the form on dart exceptions?

I've filed an issue for this so we don't have to fix it here ( https://github.com/flutter/flutter/issues/176593 ).  This PR is just following the example set from the other errors.

gaaclarke:
@timmaffett I checked with the rest of the team and they are happy with this approach.  It looks like you addressed all the other feedback except the one that asks for a test that parses version, preferably in the negative case.  I didn't hear back from you after you addressed the other feedback; is that something you are planning to look into or could use some more guidance on?  Let me know where your head is at.

timmaffett:
Yes, I am planning on adding the negative test - I just got busy with work.  I was planning on adding a negative parsing case as all of the existing test confirm the positive (successful parsing).
~I researched 'surfacing' the parsing errors out to dart in addition to relying on the VALIDATION_LOG  log.  I some non-exhaustive searching I could not find any examples of this in the code that I looked through (of surfacing an error from C++ to dart as an exception) - do you have any suggestions here ?~ I see that you filed https://github.com/flutter/flutter/issues/176593 and these can all be addressed at the same time.
  I will try to get to this ASAP.

ps. I was also hoping to get some feedback on #175479 if possible ;)

gaaclarke (approved):
lgtm!

engine/src/flutter/impeller/renderer/backend/gles/shader_library_gles.cc
```diff
@@ -67,12 +67,12 @@ ShaderLibraryGLES::ShaderLibraryGLES(
     return true;
   };
   for (auto library : shader_libraries) {
-    auto blob_library = ShaderArchive{std::move(library)};
-    if (!blob_library.IsValid()) {
+    auto blob_library = ShaderArchive::Create(std::move(library));
+    if (!blob_library.ok()) {
```
gaaclarke, line 71: We should probably be reporting the status here, but we already have a PR open to go back and surface these to dart.

chinmaygarde (approved):

PR #175473, opened 2025-09-17
[Framework] iOS style blurring and `ImageFilterConfig`

### LINKED ISSUE #99691: Surrounding color leaks into widgets when BackdropFilter with ImageFilter.blur is applied

dkwingsmt:
This PR adds the framework support for a new iOS-style blur. The new style, which I call "bounded blur", works by adding parameters to the blur filter that specify the bounds for the region that the filter sources pixels from.

As discussed in design doc [flutter.dev/go/ios-style-blur-support](http://flutter.dev/go/ios-style-blur-support), it's impossible to pass layout information to filters with the current `ImageFilter` design. Therefore this PR creates a new class `ImageFilterConfig`.

This PR also applies bounded blur to `CupertinoPopupSurface`. The following images show the different looks of a dialog in front of background with abrupt color changes just outside of the border. Notice how the abrupt color changes no longer bleed in.

image

image

This feature continues to matter for iOS 26, since the liquid glass design also heavily features blurring.

Fixes https://github.com/flutter/flutter/issues/99691.

### API changes

* `BackdropFilter`: Add `filterConfig`
* `RenderBackdropFilter`: Add `filterConfig`. Deprecate `filter`.
* `ImageFilter`: Add `debugShortDescription` (previously private property `_shortDescription`)

### Demo

The following demo app, which is implemented on the widget layer, compares the effect of a bounded blur and an unbounded blur.

https://github.com/user-attachments/assets/f715db44-c0a0-4ac8-a163-6b859665b032

<details>
<summary>
Demo source
</summary>

```
// Add to pubspec.yaml:
//
//  assets:
//      - assets/kalimba.jpg
//
// and download the image from
// https://github.com/flutter/flutter/blob/ec6f55023760ea4f44d311b9c69c39910f6b8b0c/engine/src/flutter/impeller/fixtures/kalimba.jpg

void main() => runApp(const MaterialApp(home: BlurEditorApp()));

class ControlPoint extends StatefulWidget {
  const ControlPoint({
    super.key,
    required this.position,
    required this.onPanUpdate,
    this.radius = 20.0,
  });

  final Offset position;
  final GestureDragUpdateCallback onPanUpdate;
  final double radius;

  @override
  ControlPointState createState() => ControlPointState();
}

class ControlPointState extends State<ControlPoint> {
  bool isHovering = false;

  @override
    return Positioned(
      left: widget.position.dx - widget.radius,
      top: widget.position.dy - widget.radius,
      child: MouseRegion(
        onEnter: (_) { setState((){ isHovering = true; }); },
        onExit: (_) { setState((){ isHovering = false; }); },
        cursor: SystemMouseCursors.move,
        child: GestureDetector(
          onPanUpdate: widget.onPanUpdate,
          child: AnimatedContainer(
            duration: const Duration(milliseconds: 200),
            width: widget.radius * 2,
            height: widget.radius * 2,
            decoration: BoxDecoration(
              shape: BoxShape.circle,
              color: isHovering
                  ? Colors.white.withValues(alpha: 0.8)
                  : Colors.white.withValues(alpha: 0.4),
              border: Border.all(color: Colors.white, width: 2),
              boxShadow: [
                if (isHovering)
                  BoxShadow(
                    color: Colors.white.withValues(alpha: 0.5),
                    blurRadius: 10,
                    spreadRadius: 2,
                  )
              ],
            ),
            child: const Icon(Icons.drag_indicator, size: 16, color: Colors.black54),
          ),
        ),
      ),
    );
  }
}

class BlurPanel extends StatelessWidget {
  const BlurPanel({super.key, required this.blurRect, required this.bounded});

  final Rect blurRect;
  final bool bounded;

  @override
    return Scaffold(
      body: Column(
        children: [
          Expanded(
            child: Stack(
              children: [
                Positioned.fill(
                  child: Image.asset(
                    'assets/kalimba.jpg',
                    fit: BoxFit.cover,
                  ),
                ),
                Positioned.fromRect(
                  rect: blurRect,
                  child: ClipRect(
                      child: BackdropFilter(
                    filterConfig: ImageFilterConfig.blur(
                        sigmaX: 10, sigmaY: 10, bounded: bounded),
                    child: Container(),
                  )),
                ),
              ],
            ),
          ),
          Padding(
            child: Text(
              bounded ? 'Bounded Blur' : 'Unbounded Blur',
              style: const TextStyle(fontSize: 20, fontWeight: FontWeight.bold),
            ),
          ),
        ],
      ),
    );
  }
}

class BlurEditorApp extends StatefulWidget {
  const BlurEditorApp({super.key});

  @override
  State<BlurEditorApp> createState() => _BlurEditorAppState();
}

class _BlurEditorAppState extends State<BlurEditorApp> {
  Offset p1 = const Offset(100, 100);
  Offset p2 = const Offset(300, 300);

  @override
    final blurRect = Rect.fromPoints(p1, p2);

    return Scaffold(
      body: Stack(
        children: [
          Positioned.fill(
            child: Row(
              children: [
                Expanded(
                  child: BlurPanel(blurRect: blurRect, bounded: true),
                ),
                Expanded(
                  child: BlurPanel(blurRect: blurRect, bounded: false),
                ),
              ],
            ),
          ),

          ControlPoint(position: p1, onPanUpdate: (details) { setState(() => p1 = details.globalPosition); }),
          ControlPoint(position: p2, onPanUpdate: (details) { setState(() => p2 = details.globalPosition); }),
        ],
      ),
    );
  }
}

```

</details>

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,182 @@
...
+/// Most filters can be used via [ImageFilterConfig.filter]. The most notable
+/// filter that requires this class is [ImageFilterConfig.blur] with the
+/// `bounded` option set to true.
+///
+/// See also:
+///
+///  * [BackdropFilter.filterConfig], which uses this class to configure its filter.
+///  * [ui.ImageFilter], the engine-level class that this config resolves to.
+@immutable
+abstract class ImageFilterConfig {
```
chunhtai, line 59: For naming

This feels like a builder pattern, can consider go with that.

If the bound is the only resolving factor, consider call this ImageFilterResolver

if there may be more way to resolve in the future, consider ImageFilterBuilder, ImageFilterBuilder.addBound, ImageFilterBuilder.build

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,182 @@
...
+    required ImageFilterConfig inner,
+  }) {
+    return _ComposeImageFilterConfig(outer: outer, inner: inner);
+  }
+
+  /// Resolves this configuration into a [ui.ImageFilter], given the
+  /// `bounds` of the widget applying the filter.
+  ///
+  /// The `bounds` can be used to create layout-dependent filters, such as
+  /// a blur that is clipped to the widget's bounds.
```
chunhtai: This is not all it does, right? it also means the sampling is kept in the bounds

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,182 @@
...
+  final ui.ImageFilter filter;
+
+  @override
+  ui.ImageFilter resolve(ui.Rect bounds) {
+    return filter;
+  }
+
+  @override
+  bool operator ==(Object other) {
+    if (other.runtimeType != runtimeType) {
```
chunhtai: this should be defined in super class

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,182 @@
...
+      return false;
+    }
+    return other is _DirectImageFilterConfig &&
```
chunhtai: this is not needed if other.runtimeType == runtimeType

packages/flutter/lib/src/rendering/proxy_box.dart
```diff
@@ -1201,11 +1204,15 @@ class RenderBackdropFilter extends RenderProxyBox {
...
-    required ui.ImageFilter filter,
+    ui.ImageFilter? filter,
+    ImageFilterConfig? filterConfig,
     BlendMode blendMode = BlendMode.srcOver,
     bool enabled = true,
     BackdropKey? backdropKey,
-  }) : _filter = filter,
+  }) : assert(filter != null || filterConfig != null, 'Either filter or filterConfig must be provided.'),
+       assert(filter == null || filterConfig == null, 'Cannot provide both a filter and a filterConfig.'),
+       _filter = filter,
```
chunhtai: reimplement this with ImageFilterConfig.filter

packages/flutter/lib/src/rendering/proxy_box.dart
```diff
@@ -1201,11 +1204,15 @@ class RenderBackdropFilter extends RenderProxyBox {
   /// The [blendMode] argument defaults to [BlendMode.srcOver].
   RenderBackdropFilter({
     RenderBox? child,
-    required ui.ImageFilter filter,
+    ui.ImageFilter? filter,
```
chunhtai, line 1210: any plan to deprecate this?

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -653,18 +656,35 @@ class BackdropFilter extends SingleChildRenderObjectWidget {
   /// widgets.
   const BackdropFilter.grouped({
     super.key,
-    required this.filter,
+    this.filter,
```
chunhtai, line 674: same, consider deprecation and reimplementing with ImageFilterConfig.filter

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,182 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ui' as ui;
+
+import 'package:flutter/foundation.dart';
+
+/// A configuration for a [ui.ImageFilter].
+///
+/// This class provides a framework-level abstraction for image filters that can
```
chunhtai: Just curious have you thought about renaming existing `ImageFilter` to something else and reuse the name for this class?

dkwingsmt: I'm not a fan of this idea.
* Renaming existing `ImageFilter`s is too big of a breakage that I think requires at least a 3-step deprecation process. 
* `ImageFilter` is not only used for backdrop filters. For example, you can apply an image filter to a `Paint` so that the `Canvas` can draw an image with the image filter directly.
* Even for the current use case of backdrop filters, the name `ImageFilter` makes sense. It filters the image, just that the filtering algorithm needs layout data as a parameter.

packages/flutter/lib/src/rendering/proxy_box.dart
dkwingsmt: For now I don't have any plan to deprecate this. `ImageFilterConfig` is a bit more tedious to use than `ImageFilter` and its only benefit is the iOS-style bounded blur. Forcing everyone to migrate to that isn't worth it.

packages/flutter/lib/src/widgets/basic.dart
dkwingsmt: For now I don't have any plan to deprecate this. `ImageFilterConfig` is a bit more tedious to use than `ImageFilter` and its only benefit is the iOS-style bounded blur. Forcing everyone to migrate to that isn't worth it. The proposed API isn't that bad either since we can enforce it with assertion.

packages/flutter/lib/src/rendering/image_filter_config.dart
dkwingsmt: > Naming

I thought about calling it a builder. It's not bad. The only concern I have is that the name `builder` is typically associated with returning a widget. If this doesn't seem a concern to you I'm happy with this.

Resolver might be a choice as well, but providing the `BackdropFilter` widget with a `filterResolver` doesn't feel as intuitive as `filterBuilder`.

> Parameters

I'm not confident enough to believe that the bounds is the only parameter for resolving. That's why I even created a separate data class `ImageFilterContext` to store the parameters so that we can extend the parameter list without breaking existing `ImageFilterConfig` types.

dkwingsmt:
Regarding the API `filter` vs `filterConfig`: (https://github.com/flutter/flutter/pull/175473#discussion_r2376919150)
I attempted to implement `filter` with `filterConfig`, 
```dart 
  RenderBackdropFilter({
    RenderBox? child,
    ui.ImageFilter? filter,
    ImageFilterConfig? filterConfig,
  }) : assert(filter != null || filterConfig != null, 'Either filter or filterConfig must be provided.'),
       assert(filter == null || filterConfig == null, 'Cannot provide both a filter and a filterConfig.'),
       _filterConfig = filterConfig ?? ImageFilterConfig.filter(filter!),
```

And then we remove the `_filter` property.

But then comes the problem: I can't write a `filter` getter anymore. Doing so requires having a public way of checking whether a filter config is `_DirectImageFilterConfig`.

We can make `_DirectImageFilterConfig` public, which I'm not strongly against. But it'll be good to have your opinion. (Also then we probably will call it `FilterImageFilterConfig`, which is really ugly... but I just thought that `ImageFilterConfig.filter` is an intuitive name.)

packages/flutter/lib/src/rendering/image_filter_config.dart
chunhtai: > Renaming existing ImageFilters is too big of a breakage that I think requires at least a 3-step deprecation process.

How so? let's say we rename ImageFilters into EngineImageFilters. if we make the current ImageFilterConfig mirror the ImageFilters (which we kind of do for .compose) API and completely replace the ImageFilters. Why would this become a breaking change?

> For example, you can apply an image filter to a Paint so that the Canvas can draw an image with the image filter directly.

Inside these class, they will turn ImageFilters into EngineImageFilters and send to engine to draw. This should be fine right?

chunhtai:
> But then comes the problem: I can't write a filter getter anymore. Doing so requires having a public way of checking whether a filter config is _DirectImageFilterConfig.

I 'think' this is fine, or just resolve with
 ```
_filterConfig.resolve(bound: size ?? null /*or whatever default value for blur*/).
```

If we really can't have a reasonable getter, I would rather deprecate the getter then have to maintain both in these class

**Comment by dkwingsmt** [comment]
Just to be clear, the obstacle of the getter has nothing to do with resolving.

If we only store `_filterConfig` in the class, then the `get filter` will need to somehow extract the `ImageFilter` object that was assigned via either the constructor or the setter. But the `ImageFilterConfig` class does not have a way to query its direct filter (when applicable). Therefore we can't write a `get filter`, which is further required for the setter. In this way, the `filter` property can not be queried or set, breaking all code that uses it. I doubt anyone uses the render object though, so maybe we can just delete it as well.

File: packages/flutter/lib/src/rendering/image_filter_config.dart (hunk 8 above)
Comment (dkwingsmt) [reply packages/flutter/lib/src/rendering/image_filter_config.dart]: That's quite a clever duck typing. Well I can think of one way it can break: `ImageFilter` actually comes from `dart:ui` and people might write import `dart:ui` as `ui` and write `ui.ImageFilter` instead.

File: packages/flutter/lib/src/rendering/image_filter_config.dart (hunk 3 above)
Comment (dkwingsmt) [reply packages/flutter/lib/src/rendering/image_filter_config.dart]: We can't define `operator ==` entirely in super class and therefore I think super class should not define it at all. (I've removed the superclass definition now.)

File: packages/flutter/lib/src/rendering/image_filter_config.dart (hunk 2 above)
Comment (dkwingsmt) [reply packages/flutter/lib/src/rendering/image_filter_config.dart]: Fixed. Thank you!

**Comment by dkwingsmt** [comment]
> > But then comes the problem: I can't write a filter getter anymore. Doing so requires having a public way of checking whether a filter config is _DirectImageFilterConfig.
> 
> I 'think' this is fine, or just resolve with
> 
> ```
> _filterConfig.resolve(bound: size ?? null /*or whatever default value for blur*/).
> ```
> 
> If we really can't have a reasonable getter, I would rather deprecate the getter then have to maintain both in these class

That's a really good way! I'll have a try.

File: packages/flutter/lib/src/rendering/image_filter_config.dart (hunk 4 above)
Comment (dkwingsmt) [reply packages/flutter/lib/src/rendering/image_filter_config.dart]: As far as I see this is the convention we used for many, if not all, `bool operator ==` implementations. (The reason this is needed is probably for Dart type checking)

(During this check I found that our convention also includes an `identical` check, which I've added to the PR.)

**Comment by dkwingsmt** [comment]
I've removed the `filter` property from the render object while keeping it in the widget. The reasons are:
* The render object should not be used much and hence safe to break.
* The widget is commonly used and should not be broken.
* The widget is immutable and therefore having two properties there is much less of an issue.

Let me know what you think! (Also, do you prefer `ImageFilterBuilder` over `ImageFilterConfig`?)

File: packages/flutter/lib/src/rendering/image_filter_config.dart (hunk 8 above)
Comment (chunhtai) [reply packages/flutter/lib/src/rendering/image_filter_config.dart]: That won't be a problem if this is definded in dart:ui right? though that means we need to define both web and not web version of this

File: packages/flutter/lib/src/rendering/image_filter_config.dart [hunk 9]
```diff
@@ -0,0 +1,174 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ui' as ui;
+
+import 'package:flutter/foundation.dart';
```
chunhtai, line 10: what is this used for?

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,174 @@
...
+    required ImageFilterConfig outer,
+    required ImageFilterConfig inner,
+  }) {
+    return _ComposeImageFilterConfig(outer: outer, inner: inner);
+  }
+
+  /// Resolves this configuration into a [ui.ImageFilter], given the
+  /// `bounds` of the widget applying the filter.
+  ///
+  /// The `bounds` can be used to create layout-dependent filters, such as
```
chunhtai: somewhere in the doc should mention what coordinate this bound is in. based on the code, This is in global coordinate

packages/flutter/lib/src/rendering/proxy_box.dart
```diff
@@ -1271,17 +1273,37 @@ class RenderBackdropFilter extends RenderProxyBox {
...
+  // Computes the bounds of this render object in the coordinate space where the
+  // filter is applied.
+  //
+  // This method applies all ancestor transforms from this render object up to
+  // the root of the render tree, including the root transform of [RenderView],
+  // and then excludes the offset used for the pushed layer. The resulting
+  // transform is applied to the rectangle that represents this object's size.
+  Rect _sizeForFilter(Offset offset) {
+    final Matrix4 transform = Matrix4.translationValues(-offset.dx, -offset.dy, 0);
+    for (RenderObject current = this; current.parent != null; current = current.parent!) {
```
chunhtai: you can use getTransformTo to get the global transform

chunhtai (commented):
Approach overall looks good, just wonder if we can do anything about https://github.com/flutter/flutter/pull/175473/files#r2376931055

packages/flutter/lib/src/rendering/proxy_box.dart
dkwingsmt: If I remember correctly I've tried it, but it doesn't include the root transform of `RenderView`. Rather than combining `getTransformTo` with the root transform, I think it's easier to compute everything by myself this way.

dkwingsmt:
Continuing discussion in https://github.com/flutter/flutter/pull/175473/files#r2383499661 :

I think this will not be a good idea for the following reasons:
1. The reason why we needed a builder class for filters are completely within the framework, since layout is a framework concept and the engine does not care where your information comes from. Imagine an app that is built completely on dart:ui API. They will not need the builder class at all. Therefore, making the builder class in the engine is putting framework-only information into the engine level and does not make much sense.
2. Apps that are built completely on dart:ui API will be broken. They make engine calls with `ui.ImageFilter` objects and after your proposal they will need to resolve them to engine filters. Similarly, render objects that draws stuff with `ui.ImageFilter` will be broken. For example, apps can draw texture with paints that contain image filters. In general, the engine calls require resolved filter objects, and apps or libraries that directly make these engine calls will be broken if we use a new class to represent the resolved filter objects.

packages/flutter/lib/src/rendering/image_filter_config.dart
dkwingsmt: It's for `@immutable`.

packages/flutter/lib/src/rendering/image_filter_config.dart
dkwingsmt: Fixed.

chunhtai:
I don't feel strong against renaming or not, just try to see if this will make it easier to maintain then creating a new config class and APIs

> making the builder class in the engine is putting framework-only information into the engine level and does not make much sense.

The I think the additional infromation is the bound, this is not a framework-only information I think?

> Apps that are built completely on dart:ui API will be broken

This correct, but I think this is not something we prioritize after mono repo. So I don't think this is a concern.

dkwingsmt:
I added a `ImageFilterContext` class that groups the data to resolve filters, so that we can expand the parameter set without breakages. I also added documents. Let me know what you think. @chunhtai

dkwingsmt:
I've just seen your reply. :)  Just want to quickly reply

> The I think the additional infromation is the bound, this is not a framework-only information I think?

You're completely right. Let me think more about it.

packages/flutter/lib/src/rendering/image_filter_config.dart
chunhtai: This is not addressed yet?

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,193 @@
...
+    this.sigmaX = 0.0,
+    this.sigmaY = 0.0,
+    this.tileMode = ui.TileMode.clamp,
+    this.bounded = false,
+  });
+
+  final double sigmaX;
+  final double sigmaY;
+  final ui.TileMode tileMode;
+  final bool bounded;
```
chunhtai, line 194: is this only for backward compatibility? is there a good reason to preserver the old behavior?

chunhtai: also if this only be meant to be internal, consider making the renderobject to pass a null bound in context instead to reduce API surface

packages/flutter/lib/src/rendering/proxy_box.dart
chunhtai: how's that possible? unless the owner!.rootNode is not the RenderView. I think it is best if we can reuse getTransformTo in case we create more RenderObject tweak like overlayportal in the future, and we don't create additional place to that needs to be updated.

packages/flutter/lib/src/rendering/image_filter_config.dart
dkwingsmt: > is there a good reason to preserver the old behavior?

Yes. Not all blurs need to be bounded. The bounded blur is at least slightly more computationally heavy. The different style of the blur might not even be considered correct for non-iOS platforms. I think it's necessary to allow the traditional unbounded blur.

dkwingsmt: What do you mean by "consider making the renderobject to pass a null bound in context instead to reduce API surface"?

Whenever we need the `ImageFilterConfig` class or its subclasses, the layout information is not available yet. This class is used in the widget and object level. It can only be resolved when the object is painting.

chunhtai: so if the decision of using bound can be made internally instead of from developers, the render object can do something like

```dart
class MyRenderObject ....{

ImageFiluterContext _contextForFilter(
   if (needsBound) {
     return ImageFiluterContext(size: this.size);
   }  else {
     return ImageFiluterContext(size: null);
   }
) 
```

chunhtai: and ImagieFilterConfig.resolve will just be smart about handling null size means no bound assigned to the blur filter

dkwingsmt: But how do we know `needsBound`? This is in fact a real extra parameter that needs to be public. And there's no reason why we can't provide the size. The `size` parameter has to be there anyway.

chunhtai: > how do we know needsBound? 

Right so that my suggestion is only viable if we can determine this internally, for example if 2 widget that use blur effect, and one needs bounds and the other don't. then the can decide whether to pass bound into the context.

an alternative is to put the boolean in BackdropFilter widget instead, it is probably more straight forward then setting a boolean in config to ignore rect from renderobject. I don't feel strongly either way though

dkwingsmt:
I discussed with @chunhtai and here is the summary:
* The current API design, specifically `_BlurImageFilterConfig.bounded` and `ImageFilterContext.bounds`, is reasonable, resolving https://github.com/flutter/flutter/pull/175473#discussion_r2415167019.
* The way that `_sizeForFilter` is written is dubious.
  * Technically, the render object layer might not be aware of its relation shipwith the device (hence the dpr). For example, the current `RenderView` might not necessarily be active. Therefore getting its transform matrix might not be reasonable. I will explore whether it makes more sense to move the resolution down to the `Layer` layer.
  * Also, @chunhtai is curious how the current loop would work, considering that `renderView.parent` should be null.
  * Even if we will keep the current structure, it might be better to move `_sizeForFilter` to `RenderObject` as a public method. @chunhtai proposed to somehow merge its logic with `getTransformTo`, which I'm not hopeful of.

Piinks:

packages/flutter/lib/src/rendering/image_filter_config.dart
dkwingsmt: Update: Since the parameter here has been changed to a `ImageFilterContext`, it no longer need to define the bounds. Instead, `ImageFilterContext.bounds` contains a complete definition of the bounds, including the coordinate it uses.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4401,9 +4401,9 @@ abstract class ImageFilter {
   // subclasses for the exact type of DlImageFilter this method converts to.
   _ImageFilter _toNativeImageFilter();
 
-  // The description text to show when the filter is part of a composite
-  // [ImageFilter] created using [ImageFilter.compose].
-  String get _shortDescription;
+  /// The description text to show when the filter is part of a composite
+  /// [ImageFilter] created using [ImageFilter.compose].
+  String get shortDescription => toString();
```
chunhtai: consider change this to something like debugShortDescription since this is not needed in released mode

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,258 @@
...
+///
+///  * [ImageFilterContext], which provides the layout information used to
+///    resolve this configuration.
+///  * [ImageFilterConfig.blur], which can be used to create a "bounded blur"
+///    that limits sampling to the object's boundaries.
+///  * [ui.ImageFilter], the engine-level class that this config resolves to.
+///  * [BackdropFilter.filterConfig], which uses this class to configure its effect.
+@immutable
+abstract class ImageFilterConfig {
+  /// Creates a configuration that directly uses the given filter.
```
chunhtai: Maybe mention this won't resolve the given image filter with the rect if they want to use ImageFilterConfig(ImageFilter.blur), they should use ImageFilterConfig.blur

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,258 @@
...
+  /// wrapping an existing filter.
+  ///
+  /// This getter returns non-null only if this object was created using the
+  /// default [ImageFilterConfig.new] constructor.
+  ///
+  /// For all other constructors (such as [ImageFilterConfig.blur]), this getter
+  /// returns null, even if the filter's parameters do not currently depend on
+  /// layout information. For these configurations, you must use [resolve] to
+  /// obtain the actual [ui.ImageFilter].
+  ui.ImageFilter? get filter {
```
chunhtai: instead of implementing here. default to null and override this in _DirectImageFilterConfig

chunhtai (commented):
LGTM for the most part, just some minor comments

dkwingsmt:
@chunhtai I've addressed all your comments. PTAL. Thank you!

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -57,7 +57,21 @@ class ImageFilterContext {
...
+  /// Creates a configuration that directly wraps an existing [ui.ImageFilter].
+  ///
+  /// This constructor is used to adapt standard engine-level filters to APIs
+  /// that require an [ImageFilterConfig].
+  ///
+  /// Since the provided [ui.ImageFilter] is already instantiated,
+  /// it cannot automatically incorporate layout information (such as bounds) from the
+  /// [ImageFilterContext] during resolution. For example, wrapping a
+  /// [ui.ImageFilter.blur] with this constructor results in a static blur
+  /// with a fixed blur bounds. If you need a blur that adapts to its layout
```
chunhtai: avoid `you`, our style guide suggest passive term such as  `one can` or `consider`

packages/flutter/lib/src/rendering/image_filter_config.dart
```diff
@@ -0,0 +1,274 @@
...
+  /// wrapping an existing filter.
+  ///
+  /// This getter returns non-null only if this object was created using the
+  /// default [ImageFilterConfig.new] constructor.
+  ///
+  /// For all other constructors (such as [ImageFilterConfig.blur]), this getter
+  /// returns null, even if the filter's parameters do not currently depend on
+  /// layout information. For these configurations, you must use [resolve] to
+  /// obtain the actual [ui.ImageFilter].
+  ui.ImageFilter? get filter => null;
```
chunhtai, line 141: There is no subclass override this getter

dkwingsmt: _Direct is overriding this getter with the final variable.
(Also the description is tested in the unit tests.)

chunhtai: oh thanks for pointing out, I was searching for the wrong keyword.

chunhtai (approved):
LGTM, except for the a doc nit

Issue #167404, opened 2025-04-18
Allow loading fragment shaders from custom asset bundles


Stumblinbear:
### Use case

As it currently stands, fragment shaders appear to be loaded outside of the typical method of loading assets. They seem to sidestep the defined DefaultAssetBundle as well as the rootBundle, opting to fetch the shader directly from an asset manager in C++ instead. I have not found a way to modify this behavior.

We have built a custom embedding, and we make our assets available via our own custom asset bundle implementation, but if we want to use any custom shaders we'd have to ship those via the built-in asset bundling method instead. Shaders don't even seem to use the asset platform channel, so making it available through that's not doable, either.

Our asset bundle *does* include the compiled shader, the Flutter engine just cannot find it.

Correct me if I'm wrong, of course, but it seems there's no way to load shaders any other way.

### Proposal

Updating it to search the root asset bundle would likely be a breaking change, and it would require access to `BuildContext` in order to get the current `DefaultAssetBundle`. The fragment program load function could take an optional asset bundle argument and use that if supplied.

I'm unsure of the specifics of the implementation that may make this solution infeasible

jonahwilliams:
The reasons this API doesn't exist is similar to the reasons in https://github.com/flutter/flutter/issues/166944 , namely that the shader bundle format hasn't been stabilized (though it could be at this point) and there is no easy way to run impellerc except via the flutter asset bundling step. Loading directly from the C++ asset bundle side steps these problems as we avoid having to expose the format at all.

If you have a custom embedder though, part of the embedding API is specifying the asset base directory: https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/embedder/embedder.h#L2344 . I'm surprised that isn't working.

Changing the existing ui.FragmentProgram.fromAsset API is neither feasible nor desirable: its defined in dart:ui and exists before the asset bundle or any parts of the framework are defined. We could support ui.FragmentProgram.fromBytes and then default asset bundle hooks for loading shaders, but that relies on  https://github.com/flutter/flutter/issues/166944 being completed as well.

Stumblinbear:
> specifying the asset base directory: <snip>. I'm surprised that isn't working.

We distribute our assets in a single packed file instead of a folder structure, so even if we did that it wouldn't be able to find it.

What we ended up doing in the mean time is copying the shader data into a temp directory when it's loaded where Flutter is looking for it in the filesystem. It's janky, but seems for work for now.

PR #175479, opened 2025-09-17
add FragmentProgram.fromBytes to c++ and web engines. 

### LINKED ISSUE #166944: [Impeller] Provide an API for loading Shader at runtime (and tooling support for producing them)

### LINKED ISSUE #169442: Allow for shader loading at runtime
### Use case

I'm building a desktop app with Flutter, whose purpose is to let developers write, edit, and explore shaders for use in Flutter apps.

Flutter currently won't let me load shaders from anywhere but assets, which prevents any runtime shader work.

### Proposal

This app requires that shader source code be compiled and loaded at runtime. It doesn't necessarily require that Flutter provide full support for shader compilation at runtime, assuming I'm able to include some other package that does compilation. However, this use-case does require that I be able to somehow load a shader at runtime that comes from somewhere other than assets.

There was a related issue to this that was closed: https://github.com/flutter/flutter/issues/124759

timmaffett:
Fixes #169442
Fixes #166944
Accomodates #167404 and some of #166944 

This PR adds a FragmentProgram.fromBytes() method to allow loading of fragment shaders from outside the original asset bundle.  This greatly extends the capabilities of the current restrictions of only allowing fragment shaders to be bundled in the build time asset bundle. (allowing many new capabilities, including items like the protection of shader source code via encryption, downloads of new shaders without reinstallation of program, runtime shader editing, ad infinitum). The referenced issues only partially touch upon why this feature is desperately needed.

This PR goes hand in hand with #175470 which allows versioning of impellerc outputs and protection for the engine from attempting to load incompatible versions of any of the impellerc output flat buffers.  PR #175470 should be considered a required component of this PR to protect the engine from unexpected input.

I will incorporate unit tests for these capabilities which exercise the fromBytes() method - I wanted to initiate the conversation for now.  I personally feel that the bang for the buck from this PR is extraordinary.

Other issues that this PR resolves or at least provides capabilities for that allow work-arounds:
  resolves - #169442 - Allow for shader loading at runtime
  partial - #171284 - Add support for loading assets with non-package paths in dart:ui APIs
                     (for the FragmentProgram.fromAsset() limitations .fromBytes() can be used to load your assets from
                            any AssetBundle)
  

## Pre-launch Checklist

- [X] I signed the [CLA].
- [FORTHCOMING] I added new tests to check the change I am making, or this PR is [test-exempt].

### DIFF
--- a/engine/src/flutter/lib/ui/painting.dart
+++ b/engine/src/flutter/lib/ui/painting.dart
@@ -5253,7 +5253,7 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
       return Future<FragmentProgram>.value(program);
     }
     return Future<FragmentProgram>.microtask(() {
-      final FragmentProgram program = FragmentProgram._fromBytes(nameForShaderRegistry,bytes);
+      final FragmentProgram program = FragmentProgram._fromBytes(nameForShaderRegistry, bytes);
       _shaderRegistry[nameForShaderRegistry] = program;
       return program;
     });
--- a/engine/src/flutter/lib/ui/painting.dart
+++ b/engine/src/flutter/lib/ui/painting.dart
@@ -5203,6 +5203,19 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
     }());
   }
 
+  @pragma('vm:entry-point')
+  FragmentProgram._fromBytes(String nameForShaderRegistry, Uint8List bytes) {
+    _constructor();
+    final String result = _initFromBytes(nameForShaderRegistry, bytes);
+    if (result.isNotEmpty) {
+      throw Exception(result);
+    }
+    assert(() {
+      _debugName = nameForShaderRegistry;
+      return true;
+    }());
+  }
+
   String? _debugName;
 
   /// Creates a fragment program from the asset with key [assetKey].
@@ -5228,8 +5241,27 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
     });
   }
 
+  /// Creates a fragment program from the provided byte data.
+  ///
+  /// The byte data must be produced as the output of the `impellerc`
+  /// compiler. The constructed object should then be reused via the
+  /// [fragmentShader] method to create [Shader] objects that can be used by
+  /// [Paint.shader].
+  static Future<FragmentProgram> fromBytes(String nameForShaderRegistry, Uint8List bytes) {
+    final FragmentProgram? program = _shaderRegistry[nameForShaderRegistry];
+    if (program != null) {
+      return Future<FragmentProgram>.value(program);
+    }
+    return Future<FragmentProgram>.microtask(() {
+      final FragmentProgram program = FragmentProgram._fromBytes(nameForShaderRegistry,bytes);
+      _shaderRegistry[nameForShaderRegistry] = program;
+      return program;
+    });
+  }
+
   // This is a cache of shaders that have been loaded by
-  // FragmentProgram.fromAsset. It holds a strong reference to theFragmentPrograms
+  // FragmentProgram.fromAsset or FragmentProgram.fromBytes.
+  // It holds a strong reference to the FragmentPrograms.
   // The native engine will retain the resources associated with this shader
   // program (PSO variants) until shutdown, so maintaining a strong reference
   // here ensures we do not perform extra work if the dart object is continually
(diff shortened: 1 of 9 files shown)


engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5228,8 +5241,27 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
     });
   }
 
+  /// Creates a fragment program from the provided byte data.
+  ///
+  /// The byte data must be produced as the output of the `impellerc`
+  /// compiler. The constructed object should then be reused via the
+  /// [fragmentShader] method to create [Shader] objects that can be used by
+  /// [Paint.shader].
```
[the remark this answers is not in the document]
timmaffett: I stated that this PR goes hand in hand with #175470 and with such a requirement a doc comment would not be necessary because this PR should not be merged without #175470 being merged.   I separated them as #175470 has other general advantages outside of protecting FragmentProgram (ie protecting ShaderLibrary, et al.)

harryterkelsen (approved):
LGTM for the web bits. Don't have much familiarity with the native engine but the change looks pretty straightforward and correct to me there, too.

mdebbar:
Does this PR also fix https://github.com/flutter/flutter/issues/166944?

This still needs a review from the engine team. Maybe @gaaclarke or @flar?

gaaclarke:
This is still on the backburner.

gaaclarke:
Once we accept this we have a responsibility to maintain support for a file format.  I'm still hesitant to finally accept that file format.  I'm not sure what criteria I'd even need to comfortably make that designation.  We'd probably need some documentation that at least explains the file format formally.  I am interested in this however.

harryterkelsen:
What do we want to do for this one? There are merge conflicts now

gaaclarke:
I'm still of the opinion I stated a couple weeks ago here: https://github.com/flutter/flutter/pull/175479#issuecomment-4595395941

I wouldn't spend any time addressing conflicts until we give it the green light.

gaaclarke:
Okay, I think we're ready to accept this feature now.  We are doing it for flutter-gpu, so why not for flutter.  The big use-case there was for creating a scene preview / editor for a Flutter project.  The binary needed to be able to load shaders from outside its asset bundle in order to do that correctly.

In https://github.com/flutter/flutter/commit/2bd9ecbb73d938ef6537adc0e4992f4423ca7a77#diff-54b00397d8da03d8fc5e01b0c4f2c5e2bdb5a91ef78f1ebf75433e5bf8bf79dd we added c++ unit tests.  I would like a dart integration test to make sure that we get clear messaging from the dart side about version incompatibilities.

engine/src/flutter/lib/ui/painting/fragment_program.cc
```diff
@@ -43,18 +44,8 @@ static std::string RuntimeStageBackendToString(
...
-  UIDartState* ui_dart_state = UIDartState::Current();
-  std::shared_ptr<AssetManager> asset_manager =
-      ui_dart_state->platform_configuration()->client()->GetAssetManager();
-
-  std::unique_ptr<fml::Mapping> data = asset_manager->GetAsMapping(asset_name);
-  if (data == nullptr) {
-    return std::string("Asset '") + asset_name + std::string("' not found");
-  }
-
+std::string FragmentProgram::Init(const std::string& asset_name,
```
gaaclarke, line 47: This should use absl::Status.  I know we aren't using it elsewhere but we can make this do the right thing at least.

engine/src/flutter/lib/ui/painting/fragment_program.h
```diff
@@ -17,6 +17,63 @@
...
+
+    Dart_TypedDataAcquireData(handle, &type, &data, &length);
+
+    if (type != Dart_TypedData_kUint8) {
+      Dart_TypedDataReleaseData(handle);
+      return {};
+    }
+
+    const uint8_t* bytes = static_cast<const uint8_t*>(data);
+    NativeType result(bytes, bytes + length);
```
gaaclarke, line 48: Do we want to copy the data or do we want to just retain a hold on the dart data?

engine/src/flutter/lib/ui/painting.dart
gaaclarke, line 5249: The ramifications to hot-reload need to be spelled out in the docstring.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5228,8 +5241,27 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
...
+  static Future<FragmentProgram> fromBytes(String nameForShaderRegistry, Uint8List bytes) {
+    final FragmentProgram? program = _shaderRegistry[nameForShaderRegistry];
+    if (program != null) {
+      return Future<FragmentProgram>.value(program);
+    }
+    return Future<FragmentProgram>.microtask(() {
```
gaaclarke, line 5255: Comment on why this is a microtask.

gaaclarke (changes requested):
This PR needs integration tests, check out `//engine/src/flutter/testing/dart/fragment_shader_test.dart` for an example.

PR #175509, opened 2025-09-17
Refactor `KeyData.java` to use safer code 

AbdeMohlbi:
the reason why `try/catch` were removed is because `String()` is safe and does not throw see [docs](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html)
## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/android/io/flutter/embedding/android/KeyData.java
```diff
@@ -43,9 +43,9 @@ public enum Type {
     kUp(1),
     kRepeat(2);
 
-    private long value;
+    private final long value;
 
-    private Type(long value) {
+    Type(long value) {
```
chunhtai, line 48: why this change?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/KeyData.java
```diff
@@ -77,7 +77,7 @@ public enum DeviceType {
 
     private final long value;
 
-    private DeviceType(long value) {
+    DeviceType(long value) {
```
chunhtai, line 80: same

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/KeyData.java
```diff
@@ -130,11 +130,7 @@ public KeyData(@NonNull ByteBuffer buffer) {
     if (charSize != 0) {
       final byte[] strBytes = new byte[(int) charSize];
       buffer.get(strBytes, 0, (int) charSize);
-      try {
-        this.character = new String(strBytes, "UTF-8");
-      } catch (UnsupportedEncodingException e) {
-        throw new AssertionError("UTF-8 unsupported");
-      }
+      this.character = new String(strBytes, StandardCharsets.UTF_8);
```
chunhtai, line 133: based on the doc, this won't throw but replaced malformed character with some placeholders, previously this will throw, but it would continue now. Is this going to be a problem?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/KeyData.java
AbdeMohlbi: this was to fix the lint of `private modifier is redundant for enum constructor`

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/KeyData.java
AbdeMohlbi: i don' think so , the original intention was to  `UTF-8` into ` StandardCharsets.UTF_8` then i noticed the lint and after i checked the docs it seems that the usage is safe , but still i could be wrong

chunhtai: yeah the reason it has the lint is because using Charsets it no longer throw but instead replace unencodeable character with some placeholder. I am not sure if this is actually going to be a problem though

AbdeMohlbi: if i may ask what is the placeholder in this case ?

chunhtai: I have no idea. I read the doc here https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-byte:A-java.nio.charset.Charset-

It states

`This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string`

AbdeMohlbi: i think it is not worth changing then , thank u for you time

PR #175515, opened 2025-09-17
added onUserInteractionIfError for form

akashefrath:
This PR adds a new `AutovalidateMode` value:  
`onUserInteractionIfError`.

This mode allows `Form` and `FormField` widgets to auto-validate **only when a field already has an error and the user interacts with it again**.  

### Why
- Current modes (`disabled`, `always`, `onUserInteraction`, `onUnfocus`) do not cover the case where developers want validation **only when correcting an error**.  
- This improves UX by reducing unnecessary validation calls while still ensuring errors disappear as soon as the user fixes them.  

### How
- Added `onUserInteractionIfError` to `AutovalidateMode`.  
- Updated Dartdoc with detailed description and example snippet.  
- Added new unit tests to validate behavior.  
- Verified no regressions in other modes (`always`, `onUnfocus`, `onUserInteraction`).  

### Example

```dart
TextFormField(
  autovalidateMode: AutovalidateMode.onUserInteractionIfError,
  validator: (value) => value!.isEmpty ? 'Required field' : null,
)
````

---

## Related Issues

Fixes #\<INSERT\_RELATED\_ISSUE\_NUMBER\_IF\_ANY>

---

## Tests

* Added widget tests to confirm that:

  * No validation occurs until a field has an error.
  * Once an error exists, validation runs automatically on user interaction.
  * Resetting the form clears errors correctly.
* Regression tests ensure that `always`, `onUserInteraction`, and `onUnfocus` modes still behave correctly.

---

## Pre-launch / Reviewer Checklist

* [x] I have read the [[Contributor Guide](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md)](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md)
* [x] I have read the \[Tree Hygiene] wiki page and followed its guidance
* [x] I have read and followed the \[Flutter Style Guide]
* [x] I signed the \[CLA]
* [x] I added a new enum value with clear Dartdoc (`onUserInteractionIfError`)
* [x] I added example code in the Dartdoc
* [x] I added new widget/unit tests for the new mode
* [x] All existing and new tests pass
* [x] I ran `flutter analyze` and fixed all issues
* [x] No breaking changes introduced (or marked accordingly)

---

## Breaking Change

* [x] Yes
* [x] No

---

## Screenshots / Demo (if applicable)

*(N/A — logic-only change)*

---

## Reviewer Notes

* The new mode reduces unnecessary validation until the user corrects an error.
* All tests confirm correct behavior and no regressions in existing modes.

### CODE REVIEW

File: packages/flutter/lib/src/widgets/form.dart [hunk 1]
```diff
@@ -249,6 +250,7 @@ class FormState extends State<Form> {
     _hasInteractedByUser = _fields.any(
       (FormFieldState<dynamic> field) => field._hasInteractedByUser.value,
     );
+    _hasError = _fields.any((FormFieldState<dynamic> field) => field.hasError);
```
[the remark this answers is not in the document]
akashefrath: using same as old style to keep the structure same

justinmc: To avoid any possible performance hit, could you make this a getter that only calculates this when called?

packages/flutter/test/widgets/form_test.dart
```diff
@@ -839,6 +839,45 @@ void main() {
...
+              textDirection: TextDirection.ltr,
+              child: Center(
+                child: Material(
+                  child: FormField<String>(
+                    initialValue: 'foo',
+                    autovalidateMode: AutovalidateMode.onUserInteractionIfError,
+                    builder: (FormFieldState<String> state) {
+                      formFieldState = state;
+                      return Container();
+                    },
```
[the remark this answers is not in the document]
justinmc: I don't think the Flutter styleguide allows this currently.

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -827,4 +838,11 @@ enum AutovalidateMode {
   /// In order to validate all fields of a [Form] after the first time the user interacts
   /// with one, use [always] instead.
   onUnfocus,
+
+  /// Used to auto-validate [Form] and [FormField] only when a field
+  /// already has an error, and the user starts interacting with it again.
+  ///
+  /// This is useful for reducing unnecessary validation calls while
+  /// still ensuring errors are re-checked when the user attempts to fix them.
+  onUserInteractionIfError,
```
justinmc, line 852: Nit: Maybe onUserInteractionAndError? Either way is fine, though, up to you.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1037,61 @@ void main() {
...
+      // User types valid input → still no error.
+      await tester.enterText(find.byType(TextFormField), 'bar');
+      await tester.pump();
+      expect(find.text('Required'), findsNothing);
+
+      // Clear the input (invalid) → submit form manually → error should show.
+      await tester.enterText(find.byType(TextFormField), '');
+      formState.currentState!.validate();
+      await tester.pump();
+      expect(find.text('Required'), findsOneWidget);
```
justinmc: You should also add an expectation and a pump before calling `.validate`. I'm curious about the behavior at that point, too.

justinmc (commented):
I think that the real, long-term solution here is to give users more control over when errors are displayed. I have been meaning to work on this myself for years... But given that I haven't gotten around to it, I'm up for adding things like this for realistic use cases that aren't possible today. I think this one qualifies. Thanks for the PR.

akashefrath:
> I think that the real, long-term solution here is to give users more control over when errors are displayed. I have been meaning to work on this myself for years... But given that I haven't gotten around to it, I'm up for adding things like this for realistic use cases that aren't possible today. I think this one qualifies. Thanks for the PR.



Thanks so much for the thoughtful feedback! I completely agree — giving users more control over error display timing would definitely make validation more flexible and intuitive. I’m glad this use case aligns with that direction, and really appreciate your openness to the change.

And will fix the suggested changes,

packages/flutter/lib/src/widgets/form.dart
akashefrath: Thanks for the feedback! As per your suggestion, I’ve made the changes.

packages/flutter/lib/src/widgets/form.dart
akashefrath: We can go with `onUserInteractionIfError`; it provides better clarity.

packages/flutter/test/widgets/form_test.dart
akashefrath: changed

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -385,17 +388,15 @@ class FormState extends State<Form> {
         unawaited(
           Future<void>(() async {
             await Future<void>.delayed(_kIOSAnnouncementDelayDuration);
-            SemanticsService.sendAnnouncement(
-              view,
+            SemanticsService.announce(
```
justinmc: I think this method is deprecated, right?

justinmc: SemanticsService.announce is deprecated: https://main-api.flutter.dev/flutter/semantics/SemanticsService/announce.html

packages/flutter/test/widgets/form_test.dart
```diff
@@ -839,6 +839,45 @@ void main() {
...
+      Widget builder() {
+        return MaterialApp(
+          home: MediaQuery(
+            data: const MediaQueryData(),
+            child: Directionality(
+              textDirection: TextDirection.ltr,
+              child: Center(
+                child: Material(
+                  child: FormField<String>(
+                    initialValue: 'foo',
```
justinmc, line 859: Can you add another test like this one, but the initial value is empty?

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1037,76 @@ void main() {
...
+      expect(find.text('Required'), findsNothing);
+
+      // User types valid input 'bar' → autovalidate is disabled → still no error.
+      await tester.enterText(find.byType(TextFormField), 'bar');
+      await tester.pump();
+      expect(find.text('Required'), findsNothing);
+
+      // -----------------------------------------------------------------------
+      // 1. GENERATE THE ERROR (This is where the mode is activated)
+      // -----------------------------------------------------------------------
```
justinmc: Nit: I see a number `1` here but there's no corresponding `2` down below. I would just make this a normal comment.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1037,76 @@ void main() {
...
+
+      // Clear the input (invalid)
```
justinmc: Periods at the ends of these comments.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1037,76 @@ void main() {
...
+      await tester.enterText(find.byType(TextFormField), '');
+      await tester.pump(); // Pump 1: Process the text change
+
+      // Manually submit form to show the initial error (AutovalidateMode is now active)
+      formState.currentState!.validate();
```
justinmc: Can you add an expectation here between the `pump` and the `validate`? I think that's an interesting situation for this autovalidate mode.

```dart
expect(find.text('Required'), findsNothing);
```

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1037,76 @@ void main() {
...
+      // Check the behavior of a manual validate when the text is already valid.
+      // This should *confirm* the error is cleared, not re-introduce it.
+      formState.currentState!.validate();
+      await tester.pump();
+      expect(find.text('Required'), findsNothing);
+
+      // Resetting should clear form (already cleared, but a safety check).
+      formState.currentState!.reset();
+      await tester.pump();
+      expect(find.text('Required'), findsNothing);
```
justinmc, line 1158: Before resetting the form, can you do another `await tester.enterText(find.byType(TextFormField), '');` and expect that the error is shown again?

justinmc (commented):
Thanks for the fixes! Two main things here: a use of a deprecated member, and some minor suggestions for improving the tests.

packages/flutter/lib/src/widgets/form.dart
akashefrath: Reverted accidental use of deprecated method during merge, Thank you

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -385,17 +390,15 @@ class FormState extends State<Form> {
...
```
justinmc: This is still using the deprecated method: https://main-api.flutter.dev/flutter/semantics/SemanticsService/announce.html

packages/flutter/test/widgets/form_test.dart
```diff
@@ -839,6 +839,97 @@ void main() {
...
     },
   );
+  testWidgets(
+    'Does not auto-validate before value changes when autovalidateMode is set to onUserInteractionIfError',
+    (WidgetTester tester) async {
+      late FormFieldState<String> formFieldState;
+
+      String? errorText(String? value) => (value == null || value.isEmpty) ? 'Required' : null;
+
+      Widget builder() {
```
justinmc, line 850: `builder` is not reused, I think you can just pass it directly to pumpWidget.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -839,6 +839,97 @@ void main() {
...
+
+      await tester.pumpWidget(builder());
+
+      // The form field has no error.
+      expect(formFieldState.hasError, isFalse);
+      // No "Required" error is visible.
+      expect(find.text('Required'), findsNothing);
+    },
+  );
+  testWidgets(
```
justinmc, line 883: There should be an empty line between these.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -998,6 +1089,74 @@ void main() {
...
+
+      // No error text is visible yet. (Initial valid state)
+      expect(find.text('Required'), findsNothing);
+
+      // User types valid input 'bar' → autovalidate is disabled → still no error.
+      await tester.enterText(find.byType(TextFormField), 'bar');
+      await tester.pump();
+      expect(find.text('Required'), findsNothing);
+
+      // Clear the input (invalid)
```
justinmc: These comments are still missing periods at the ends.

justinmc (approved):
I think your revert didn't work for some reason, I still see the deprecated method being used.

packages/flutter/test/widgets/form_test.dart
akashefrath: Other tests also use builder — should we keep it that way or change it as you suggested?

packages/flutter/test/widgets/form_test.dart
akashefrath: I had the space between tests

packages/flutter/test/widgets/form_test.dart
akashefrath: added

packages/flutter/lib/src/widgets/form.dart
akashefrath: Also, I've noted your comment on the form.dart file regarding the deprecated SemanticsService.announce. I'll update that now to use the new SemanticsService.sendAnnouncement method as per the docs you linked.

I really appreciate the guidance! I'm learning a lot from your code reviews, and it's definitely helping me improve my contributions.   ❤️❤️

packages/flutter/test/widgets/form_test.dart
justinmc: Ah ok, either way is fine then.

justinmc (approved):
LGTM 👍 Thanks for the fixes!

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -251,6 +251,7 @@ class FormState extends State<Form> {
     _hasInteractedByUser = _fields.any(
       (FormFieldState<dynamic> field) => field._hasInteractedByUser.value,
     );
+
```
Renzo-Olivares, line 254: nit: unnecessary new line I think?

packages/flutter/test/widgets/tap_region_test.dart
```diff
@@ -1201,18 +1201,17 @@ void main() {
                   Navigator.push(
                     tester.element(find.byType(GestureDetector)),
                     MaterialPageRoute<void>(
-                      builder:
```
Renzo-Olivares: Are the changes in this file necessary?

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -242,7 +242,7 @@ class FormState extends State<Form> {
   int _generation = 0;
   bool _hasInteractedByUser = false;
   final Set<FormFieldState<dynamic>> _fields = <FormFieldState<dynamic>>{};
-
+  bool get _hasError => _fields.any((FormFieldState<dynamic> field) => field.hasError);
```
loic-sharma: nit: Instead of making this a property, could you move this logic to the `build` method? The property makes it look like this is an `O(1)` operation when it is actually `O(n)`. Moving this logic to the `build` method would make this more obvious.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -839,6 +839,98 @@ void main() {
       expect(find.text(errorText('foo')!), findsNothing);
     },
   );
+  testWidgets(
```
loic-sharma, line 843:
```suggestion

  testWidgets(
```

packages/flutter/test/widgets/tap_region_test.dart
loic-sharma: (My guess is this file wasn't formatted properly and running `dart format` caused this diff. If that's the case, I'm OK with this change)

akashefrath: Yes, these changes happened because I ran dart format locally. The file wasn’t previously formatted, so the formatter applied the standard Dart style rules automatically.

packages/flutter/test/widgets/form_test.dart
akashefrath: commit suggested changes , Thank you,,

packages/flutter/lib/src/widgets/form.dart
akashefrath: Thanks for the suggestion! 👍
I removed the _hasError getter and moved the check directly into the build method so the O(n) cost is explicit in the control flow. Let me know if you'd like it structured differently.

Right now the code looks like this:

```
final bool hasError =
    _fields.any((FormFieldState<dynamic> field) => field.hasError);

switch (widget.autovalidateMode) {
  case AutovalidateMode.always:
    _validate(View.of(context));
  case AutovalidateMode.onUserInteraction:
    if (_hasInteractedByUser) {
      _validate(View.of(context));
    }
  case AutovalidateMode.onUserInteractionIfError:
    if (_hasInteractedByUser && hasError) {
      _validate(View.of(context));
    }
  case AutovalidateMode.onUnfocus:
  case AutovalidateMode.disabled:
    break;
}

```

I also tried a slightly more explicit version, where the error scan only happens if _hasInteractedByUser is true:

```
case AutovalidateMode.onUserInteractionIfError:
  if (_hasInteractedByUser) {
    final bool hasError =
        _fields.any((FormFieldState<dynamic> field) => field.hasError);
    if (hasError) {
      _validate(View.of(context));
    }
  }
  break;

```

Do you think this version is cleaner, or should I keep the current implementation?
Happy to adjust whichever way you prefer.

packages/flutter/lib/src/widgets/form.dart
akashefrath: Thanks! I removed the unnecessary blank line in that section.

packages/flutter/test/widgets/form_test.dart
akashefrath: leave as same  :)

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -831,4 +842,11 @@ enum AutovalidateMode {
   /// In order to validate all fields of a [Form] after the first time the user interacts
   /// with one, use [always] instead.
   onUnfocus,
+
+  /// Used to auto-validate [Form] and [FormField] only when a field
+  /// already has an error, and the user starts interacting with it again.
```
loic-sharma: What do you think of this?

```suggestion
  /// Used to auto-validate [Form] and [FormField] after each user
  /// interaction, only if the the field already has an error.
```

loic-sharma (approved):
LGTM

packages/flutter/lib/src/widgets/form.dart
akashefrath: Much better committing this

loic-sharma:
@akashefrath You'll likely need to merge the latest changes from the master branch to pass the ci.yaml check.

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -242,7 +242,6 @@ class FormState extends State<Form> {
   int _generation = 0;
   bool _hasInteractedByUser = false;
   final Set<FormFieldState<dynamic>> _fields = <FormFieldState<dynamic>>{};
-
```
Renzo-Olivares, line 245: nit: unnecessary deletion? Or is this from dart format?

packages/flutter/lib/src/widgets/form.dart
Renzo-Olivares: Hello, it looks like the new line is still here. Is it a result of formatting?

Renzo-Olivares (approved):
LGTM w/ small nits. Thank you for the contribution!

packages/flutter/lib/src/widgets/form.dart
akashefrath: Noted, thanks! I only added back an empty line.

Piinks:
Oh! It looks like I cannot rebase, can you take a look @akashefrath?

akashefrath:
@Piinks The autosubmit label was removed because the Google testing check failed earlier. I’ve now rebased the PR and re-tested, and currently only Tree Status is red; all other required checks are passing.

The remaining tree status failure is unrelated to this change. Could you please re-apply the autosubmit label to flutter/flutter#175515?  

Thank you

PR #175535, opened 2025-09-17
[ Widget Preview ] Allow for custom `Preview` annotations, add support for runtime transformations

### LINKED ISSUE #175534: [ Widget Preview ] Add support for creating custom `Preview` annotations
Developers should be able to create custom `Preview` annotations for common preview patterns within their projects. It should also be possible to apply runtime transformations to `Preview` instances to remove some of the restrictions associated to working with annotations and constant expressions.

bkonyi:
This change updates `Preview` to be a `base` class instead of `final`, allowing for developers to extend `Preview` and create their own custom preview annotations.

To reduce friction associated with working with `const` constructors, two new methods have been added: `MultiPreview.transform()` and `Preview.transform()`. These methods are invoked at runtime by the preview environment, allowing for developers to create or modify `Preview` instances with non-constant values. This also relaxes restrictions around `Preview`'s callback parameters (e.g., `wrapper`), which do not need to be static or public when provided to a `Preview` in the context of a `*.transform()` method. This makes it possible to create custom preview classes that completely encapsulate their callbacks, removing the need for separate public, static function declarations.

To make transforming `Preview`s simpler, a `PreviewBuilder` utility class has also been added.

Fixes https://github.com/flutter/flutter/issues/175534


packages/flutter/lib/src/widget_previews/widget_previews.dart
```diff
@@ -183,17 +219,122 @@ final class Preview {
...
+      wrapper = preview.wrapper,
+      theme = preview.theme,
+      brightness = preview.brightness,
+      localizations = preview.localizations;
+
+  /// The group the preview belongs to.
+  ///
+  /// Previews with the same group name will be displayed together in the
+  /// previewer. If not provided, 'Default' is used.
+  String? group;
```
elliette, line 286: I'm not sure if we have dartdoc best practices/suggestions around this, but I wonder if instead all these members could have comments like `/// See [Preview.group]` so that we don't have to keep the documentation updated in two places.

elliette: Ah hah! https://pub.dev/packages/dartdoc#macros

packages/flutter/lib/src/widget_previews/widget_previews.dart
```diff
@@ -183,17 +219,122 @@ final class Preview {
...
+  List<Preview> transform() => previews.map((Preview e) => e.transform()).toList();
+}
+
+/// A utility class used to build a [Preview] instance.
+final class PreviewBuilder {
+  /// Creates a [PreviewBuilder] with no initial state.
+  PreviewBuilder();
+
+  /// Creates a [PreviewBuilder] initialized with the values set in [preview].
+  PreviewBuilder.fromPreview(Preview preview)
```
elliette: [Optional] add a corresponding `toBuilder` method on `Preview`

packages/flutter_tools/lib/src/widget_preview/preview_code_generator.dart
```diff
@@ -29,16 +29,23 @@ class PreviewCodeGenerator {
...
-  static const _kBuilderLibraryUri = 'package:flutter/widgets.dart';
-  static const _kBuilderProperty = 'builder';
   static const _kListType = 'List';
   static const _kPreviewsFunctionName = 'previews';
-  static const _kGroupNameProperty = 'name';
-  static const _kGroupPreviewsProperty = 'previews';
   static const _kWidgetPreviewClass = 'WidgetPreview';
-  static const _kWidgetPreviewGroupClass = 'WidgetPreviewGroup';
   static const _kWidgetPreviewLibraryUri = 'widget_preview.dart';
+  static const _kBuildWidgetPreviewError = 'buildWidgetPreviewError';
```
elliette: What is the logic behind the ordering here? Why not alphabetized?

packages/flutter_tools/test/commands.shard/hermetic/widget_preview/utils/preview_details_matcher.dart
```diff
@@ -24,9 +37,16 @@ class PreviewDetailsMatcher extends Matcher {
     this.localizations,
   });
 
-  final String functionName;
-  final bool isBuilder;
-  final String packageName;
+  static const kScriptUri = 'scriptUri';
```
elliette: Same question about ordering logic here

elliette (approved):
A few suggestions but otherwise LGTM!

packages/flutter/lib/src/widget_previews/widget_previews.dart
bkonyi: I totally forgot about that! Thanks for reminding me :)

packages/flutter_tools/lib/src/widget_preview/preview_code_generator.dart
bkonyi: Apparently there wasn't any. I was messing around with various constants and probably meant to come back and sort these. Sorted now!

packages/flutter/lib/src/widget_previews/widget_previews.dart
bkonyi: Good call, I think that's actually a cleaner way to initialize a `PreviewBuilder` from a `Preview`. I've made this constructor private and added `PreviewBuilder toBuilder() => PreviewBuilder._fromPreview(this)` to `Preview`.

PR #175551, opened 2025-09-17
Add blockAccessibilityFocus flag

hannah-hyj:
Add a new flag for a11y focusable

- Accessibility focus, which is the focus used by screen readers like TalkBack and VoiceOver, is different from input focus.
-  Our current logic use some existing flags to decide if a node is accessibilty focusable. like "if it's a slider / has a check state / has keyboard focus/..., then it's a11y focusable" https://github.com/flutter/flutter/blob/ecbb115ae3a8cba2977ecab9f52086df860cfb1a/engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java#L98
- but we lack the ability to explicitly set a node to be unfocusable in a11y! 
- This flag can be used to explicitly set some semantics nodes to be unfocusable in a11y mode. if it is false, we fall back to the logic "if it's a slider / has a check state / has keyboard focus/..., then it's a11y focusable" 

future use case 1:
     user can set live region to be not focusable, so when content changes, it will still announce, but the content can't be focused by swiping. 
future use case 2:
    when pushing a new route like a dialog, setting the semantics nodes in old pages to be un focusable. 

## Pre-launch Checklist

- [ ] I signed the [CLA].


hannah-hyj:
still a draft

justinmc:
@hannah-hyj FYI it still says "draft" in the title of this PR, if it is indeed no longer a draft.

engine/src/flutter/lib/ui/semantics/semantics_flags.cc
```diff
@@ -25,6 +25,7 @@ void NativeSemanticsFlags::initSemanticsFlags(
     int isExpanded,
     int isRequired,
     int isFocused,
+    int isA11yFocusable,
```
chunhtai: our style guide suggest avoid abbreviation. we should use the full name isAccessibilityFocusable

hannah-hyj:
cc @ash2moon

engine/src/flutter/shell/platform/darwin/ios/framework/Source/SemanticsObject.mm
```diff
@@ -466,6 +466,11 @@ - (NSString*)accessibilityLanguage {
 }
 
 - (bool)isFocusable {
```
chunhtai, line 468: keep in mind this in turn sets the isAccessibilityElement to NO, which will completely shut down everything a11y related in this node. A more appropriate one may be 
https://developer.apple.com/documentation/objectivec/nsobject-swift.class/accessibilityrespondstouserinteraction?language=objc

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -1715,6 +1716,16 @@ class SemanticsProperties extends DiagnosticableTree {
   /// element it is reading, and is separate from input focus.
   final bool? focused;
 
+  /// If non-null, whether this node can be focused by accessibility services.
```
chunhtai:
```suggestion
  /// If non-null, whether the subtree can be focused by accessibility services.
```

chunhtai (commented):
Should the focusability applies the subtree? The current code seems to only disable the focusability of the node but not the sub tree.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/SemanticsObject.mm
hannah-hyj: done

hannah-hyj:
Yeah it should be for the subtree, added it to _SemanticsParentData and added a test to check the flag pass from parent to subtree.

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -2768,6 +2771,23 @@ private boolean isFocusable() {
...
     }
+    // From the Android doc
+    // https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo#isScreenReaderFocusable()
+    // `true` indicates the node is specifically marked as a focusable unit for screen readers.
+    // `false` indicates that it is not explicitly marked, not that the node is not
+    // a focusable unit. Screen readers should generally use other signals, such as
+    // isFocusable(), or the presence of text in a node, to determine what should
+    // receive focus.
+    //
+    // TODO(hangyujin): To explictly make a node a11y non-focusable,
```
chunhtai: We can enforce this in the framework side. Since we just added this feature, we should do it now. probably in _semanticsGeneral check

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -4859,6 +4859,9 @@ mixin SemanticsAnnotationsMixin on RenderObject {
     if (_properties.focused != null) {
       config.isFocused = _properties.focused;
     }
+    if (_properties.accessibilityFocusable != null) {
+      config.isAccessibilityFocusable = _properties.accessibilityFocusable;
```
chunhtai: we should gives this a more appropriate name like isSubTreeAccessibilityFocusable. we need to differenatiate this and the SemanticsNode.isAccessibilityFocusable where it only applies to SemanticsNode but not its children and will be handled as such in Engine.

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -5718,6 +5729,13 @@ class SemanticsConfiguration {
     _hasBeenAnnotated = true;
   }
 
+  /// Whether the owning [RenderObject] can hold the a11y focus (different from input focus).
```
chunhtai: and its subtree

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
hannah-hyj: added

packages/flutter/lib/src/rendering/object.dart
hannah-hyj: renamed

ash2moon (approved):

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -489,6 +489,10 @@ sealed class _DebugSemanticsRoleChecks {
         return FlutterError('A collapsed node cannot have a collapse action.');
       }
     }
+    if (data.flagsCollection.isFocused != Tristate.none &&
```
chunhtai: This will also throw if data.flagsCollection.isFocused == Tristate.isFalse and   data.flagsCollection.isAccessibilityFocusable == Tristate.isFalse

but it is totally reasonable they are both false.

chunhtai: we should add a test in flutter/packages/flutter/test/widgets/semantics_role_checks_test.dart

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -1715,6 +1720,16 @@ class SemanticsProperties extends DiagnosticableTree {
   /// element it is reading, and is separate from input focus.
   final bool? focused;
 
+  /// If non-null, whether the subtree can be focused by accessibility services.
```
chunhtai: some where in the doc should mention the behavior when this is false but somewhere in the subtree set this to true. Will that part of the subtree because focusable or it will get override by the parent.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5553,6 +5560,8 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
 
     final bool blocksUserAction =
         (parentData?.blocksUserActions ?? false) || configProvider.effective.isBlockingUserActions;
+    final bool? subTreeAccessibilityFocusable =
+        configProvider.effective.isSubTreeAccessibilityFocusable ?? parentData?.subTreeAccessibilityFocusable;
```
chunhtai: This means the child can override parent's behavior. 

for example

```dart
Semantics(
   isSubTreeAccessibilityFocusable: false,
   child: Semantics(
      isSubTreeAccessibilityFocusable: true,
      child: Text('text'),
   )

)
```

the text will still be focusable. I IMO parent false should take priority similar to ignore pointer (which is the opposite that true take priority)

any reason we don't want this behavior?

hannah-hyj: If we make unfocusable always taking priority, not parent or child take priority,  just like for ignore pointer `true` always takes priority. i think the flag should be named to something like `blockSubTreeAccessibilityFocus`?  and it should be a bool, not a tristate flag. 

In my opinion 
```
Semantics(
   isSubTreeAccessibilityFocusable: false,
   child: Semantics(
      isSubTreeAccessibilityFocusable: true,
      child: Text('text'),
   )
)
```
reads like something child should take the priority, why else would the child explicitly set isSubTreeAccessibilityFocusable to `true`? 
 
In the end it's about what we want,  do we just want to explicitly set unfocusable or do we also want to explicitly set a subtree to focusable?

packages/flutter/lib/src/semantics/semantics.dart
hannah-hyj: i think if data.flagsCollection.isFocused == Tristate.isFalse, it implies that this node *is* keyboard focusable, but is just not focused right now. 

in this case it shouldn't be isAccessibilityFocusable == Tristate.isFalse

this should throw

packages/flutter/lib/src/rendering/object.dart
chunhtai: One of a use case I hope can be solved is to fix the live region in a route that gets covered by a dialog route. We will have to change ModalRoute to not use BlockSemantics, but use this to hide the accessibility focus to the previous route instead.

If we go with the approach that child take priority we may need to come up with a different solution for that use case.

packages/flutter/lib/src/semantics/semantics.dart
chunhtai: oh ok, yeah I forgot the false on these two means different things. LGTM with the logic.

chunhtai (commented):
Still missing a test for the _semanticsGeneral check in 
flutter/packages/flutter/test/widgets/semantics_role_checks_test.dart

packages/flutter/lib/src/rendering/object.dart
hannah-hyj: Renamed the flag to `blockSubTreeAccessibilityFocus `  after the offline discussion

hannah-hyj:
discussed offline and will add options to block a subtree or block a current node

packages/flutter/lib/src/rendering/object.dart
ash2moon: After another offline discussion we discovered that it would not be possible to maintain just a boolean on `blockSubTreeAccessibilityFocus` because there will be cases where the underlying components that the liveRegion speaks aloud for is a parent. Here is an over-simplified tree of a date picker that speaks the full month and year:

```
├─SemanticsNode#1
│ │ Rect.fromLTRB(0.0, 106.3, 411.4, 494.3)
│ │ flags: isLiveRegion
│ │ label: "April 2025"
│ │ textDirection: ltr
│ │
│ ├─SemanticsNode#2
│ │   Rect.fromLTRB(311.4, 2.0, 359.4, 50.0)
│ │   actions: focus, tap
│ │   flags: isButton, hasEnabledState, isEnabled, isFocusable
│ │   tooltip: "April"
│ │   textDirection: ltr
│ │
│ ├─SemanticsNode#3
│ │   Rect.fro≠mLTRB(359.4, 2.0, 407.4, 50.0)
│ │   actions: focus, tap
│ │   flags: isButton, hasEnabledState, isEnabled, isFocusable
│ │   tooltip: "2025"
│ │   textDirection: ltr
```

In this case, we would want the month and year to be separately focusable, but not the parent that speaks for changes in both sub components. We came to the conclusion that an enum that contains the flags to either: do nothing, block just the current node, or block all child nodes would allow us to keep both use cases.

engine/src/flutter/lib/ui/semantics/semantics_flags.cc
```diff
@@ -40,7 +40,8 @@ void NativeSemanticsFlags::initSemanticsFlags(
     bool isReadOnly,
     bool isLink,
     bool isSlider,
-    bool isKeyboardKey) {
+    bool isKeyboardKey,
+    bool blockAccessibilityFocus) {
```
chunhtai:
```suggestion
    bool isAccessibilityFocusBlocked) {
```

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -2263,7 +2263,8 @@ public enum Action {
     IS_EXPANDED(1 << 27),
     HAS_SELECTED_STATE(1 << 28),
     HAS_REQUIRED_STATE(1 << 29),
-    IS_REQUIRED(1 << 30);
+    IS_REQUIRED(1 << 30),
+    BLOCK_ACCESSIBILITY_FOCUS(1 << 31);
```
chunhtai: same for the naming here

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5133,6 +5137,9 @@ final class _SemanticsParentData {
   /// [AbsorbPointer]s.
   final bool blocksUserActions;
 
+  /// Whether to block this subtree in accessibility focus.
```
chunhtai: I think this can use some more description since we have different mode now

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -114,6 +114,22 @@ typedef SemanticsUpdateCallback = void Function(SemanticsUpdate update);
 typedef ChildSemanticsConfigurationsDelegate =
     ChildSemanticsConfigurationsResult Function(List<SemanticsConfiguration>);
 
+/// Controls how accessibility focus is blocked.
+///
+/// This is typically used to prevent screen readers
+/// from focusing on parts of the UI.
+enum BlockAccessibilityFocus {
```
chunhtai: This can use a better name, maybe AccessiblityFocusBlockType

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5554,6 +5561,13 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
     final bool blocksUserAction =
         (parentData?.blocksUserActions ?? false) || configProvider.effective.isBlockingUserActions;
 
+    BlockAccessibilityFocus blockAccessibilityFocus;
+    if (parentData?.blockAccessibilityFocus == BlockAccessibilityFocus.blockSubtree) {
+      blockAccessibilityFocus = BlockAccessibilityFocus.blockSubtree;
+    } else {
+      blockAccessibilityFocus = configProvider.effective.blockAccessibilityFocus;
```
chunhtai: won't this propagate blockNode down?

chunhtai (commented):
We also need to consider in the case of implicit merge, whether carry blockaccessibilityFocus up to the parent the right thing to do.

packages/flutter/lib/src/rendering/object.dart
hannah-hyj: it will not, it only checks if the parentdata has `blockSubtree` and will propagate `blockSubtree`  down. 
`blockNode ` only affect the node itself, 

also add a test in semantics_test to check `blockSubtree` will change child nodes but `blockNode ` won't.

packages/flutter/lib/src/semantics/semantics.dart
hannah-hyj: thanks, the new name is way more clear

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1575,6 +1596,9 @@ class SemanticsFlags extends NativeFieldWrapperClass1 {
   /// Checks if any of the boolean semantic flags are set to true
   /// in both this instance and the [other] instance.
   bool hasRepeatedFlags(SemanticsFlags other) {
+    // isAccessibilityFocusBlocked is not checked in `hasRepeatedFlags` because
+    // two semantics nodes both with `isAccessibilityFocusBlocked` set to true is
+    // still compatible and not conflicting.
```
chunhtai: In this case, the method name no longer correct. Consider rename to something more descriptive

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -6421,6 +6462,10 @@ class SemanticsConfiguration {
     if (_flags.hasRepeatedFlags(other._flags)) {
       return false;
     }
+    if (_flags.isAccessibilityFocusBlocked != other._flags.isAccessibilityFocusBlocked) {
```
chunhtai: if one is BlockSubTree another one is BlockCurrent, I think they can merge to be BlockSubTree

chunhtai: and in this case we would need to implement absorb

engine/src/flutter/lib/ui/semantics.dart
hannah-hyj: This should be renamed to hasConflictingFlags, but it will be a breaking change.

chunhtai: we should create a new method and deprecate this one.

packages/flutter/lib/src/semantics/semantics.dart
hannah-hyj: added to the absorb

engine/src/flutter/lib/ui/semantics.dart
hannah-hyj: i will add a breaking change doc

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -114,6 +114,40 @@ typedef SemanticsUpdateCallback = void Function(SemanticsUpdate update);
...
+  none,
+
+  /// Blocks accessibility focus for the entire subtree.
+  blockSubtree,
+
+  /// Blocks accessibility focus for the **current node only**. Its descendants
+  /// may still be focusable.
+  blockNode;
+
+  /// The AccessiblityFocusBlockType when two node get merged.
```
chunhtai:
```suggestion
  /// The AccessiblityFocusBlockType when two nodes get merged.
```

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -114,6 +114,40 @@ typedef SemanticsUpdateCallback = void Function(SemanticsUpdate update);
...
+  AccessiblityFocusBlockType merge(AccessiblityFocusBlockType other) {
```
chunhtai: should this be private?

chunhtai (approved):
LGTM

PR #175552, opened 2025-09-17
Add an assertion for the relationship between `Visibility.maintainState` and `Visibility.maintainFocusability`

Renzo-Olivares:
This change adds an assertion for the relationship between `Visibility.maintainState` and `Visibility.maintainFocusability`. Setting `maintainFocusability` to `true` is a no-op if `maintainState` is set to `false`, so we should add a helpful assertion here.

## Pre-launch Checklist

- [x] I signed the [CLA].


justinmc (approved):
LGTM 👍  Thanks for following up with a separate PR for this!

PR #175560, opened 2025-09-18
Refactor: migrate fade upwards page transition builder to widgets

rkishan516:
Changes:

* Move FadeUpwardsPageTransitionsBuilder from `material/page_transitions_theme.dart` to `widget/page_transitions_builder.dart`

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -86,3 +86,69 @@ abstract class PageTransitionsBuilder {
...
+  Widget build(BuildContext context) {
+    return SlideTransition(
+      position: _positionAnimation,
+      // TODO(ianh): tell the transform to be un-transformed for hit testing
+      child: FadeTransition(opacity: _opacityAnimation, child: child),
+    );
+  }
+}
+
+/// Used by [PageTransitionsTheme] to define a vertically fading
```
victorsanni: Like in the previous PR, the documentation should be more generic, specifically defining the use of the class. Then the Material-specific logic can be added in a subsequent paragraph as an example of how this class is used.

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -175,6 +175,76 @@ void main() {
     expect(find.text('Navigate'), findsNothing);
     expect(find.text('Second Page'), findsOneWidget);
   });
+
+  testWidgets(
```
victorsanni, line 256: like in the previous PR, is it possible to add a test that shows how to use this builder without any Material artifacts?

victorsanni (commented):
Looking good so far.

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -175,6 +175,158 @@ void main() {
...
+
+      await tester.pump(const Duration(milliseconds: 300));
+
+      expect(find.text('Page 1'), isOnstage);
+      expect(find.text('Page 2'), findsNothing);
+    },
+    variant: TargetPlatformVariant.only(TargetPlatform.android),
+  );
+
+  testWidgets('FadeUpwardsPageTransitionsBuilder without Material artifacts', (
```
victorsanni: This should have the default name ('FadeUpwardsPageTransitionsBuilder test') and the other test should have the name + 'with PageTransitionsTheme'. And this test should come before the other one.

victorsanni: Basically '$builder test' + '$builder test with Material PageTransitionsTheme' instead of '$builder test vs $builder test without Material'. Material should be an example of how to use the widgets layer, not the other way round.

rkishan516: Sure, make sense.

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -175,6 +175,156 @@ void main() {
     expect(find.text('Navigate'), findsNothing);
     expect(find.text('Second Page'), findsOneWidget);
   });
+
+  testWidgets('FadeUpwardsPageTransitionsBuilder test', (WidgetTester tester) async {
+    const FadeUpwardsPageTransitionsBuilder transitionsBuilder =
```
victorsanni: why does this variable exist? Instead of just using `FadeUpwardsPageTransitionsBuilder()` directly.

packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -175,6 +175,156 @@ void main() {
     expect(find.text('Navigate'), findsNothing);
     expect(find.text('Second Page'), findsOneWidget);
   });
+
+  testWidgets('FadeUpwardsPageTransitionsBuilder test', (WidgetTester tester) async {
```
victorsanni, line 179: Looking at the body of this test (everything after `tester.pumpWidget`), I'm wondering why it can't be just a duplicate of the second test (with `PageTransitionsTheme`).

victorsanni: Just to make sure both tests capture the full behavior of the transitions builder.

victorsanni (approved):
LGTM with some more comments.

packages/flutter/test/widgets/page_transitions_builder_test.dart
rkishan516: I was planning to get some information from it, but I guess changed the plan. So, using it directly.

packages/flutter/test/widgets/page_transitions_builder_test.dart
rkishan516: Sure, makes sense.

MitchellGoodwin (approved):
LGTM as well!

PR #175659, opened 2025-09-19
Fix Xcode cache errors

### LINKED ISSUE #174444: Multiple iOS test failing with Xcode cache errors
```
Error (Xcode): no such file or directory: '/Users/swarming/Library/Developer/Xcode/DerivedData/ModuleCache.noindex/Session.modulevalidation'
Error (Xcode): stat cache file '/Users/swarming/Library/Developer/Xcode/DerivedData/SDKStatCaches.noindex/iphoneos18.2-22C146-16b195fa110ddf6a83f99c533fc1e7a6.sdkstatcache' not found
```

It seems to be isolated to these tests:
* Mac_arm64_ios integration_test_test_ios
* Mac_ios wide_gamut_ios
* Mac_ios native_assets_ios
* Mac_ios spell_check_test
* Mac_ios channels_integration_test_ios

But doesn't seem to be isolated to specific bots.

Some examples:
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_arm64_ios%20integration_test_test_ios/8450/overview (flutter-devicelab-mac-66) https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_arm64_ios%20integration_test_test_ios/8451/overview (flutter-devicelab-mac-58)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20wide_gamut_ios/14504/overview (flutter-devicelab-mac-71)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20wide_gamut_ios/14505/overview (flutter-devicelab-mac-61)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20native_assets_ios/10790/overview (flutter-devicelab-mac-71)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20native_assets_ios/10785/overview (flutter-devicelab-mac-42)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20spell_check_test/15126/overview (flutter-devicelab-mac-56)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20spell_check_test/15127/overview (flutter-devicelab-mac-66)
* https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20channels_integration_test_ios/23960/overview (flutter-devicelab-mac-63)

okorohelijah:
This PR fixes a race condition that causes intermittent build failures on iOS with errors like no such file or directory. These failures were frequently observed in CI tests running on macOS.

The root cause is that the background wireless device discovery process, which uses `devicectl`, was not being terminated after a target device was selected. This background process could then delete Xcode's module caches while `xcodebuild` was actively using them, leading to a build failure.

The fix ensures that the wireless device discovery process is explicitly stopped at all logical exit points of the device discovery flow. This is achieved by calling stopExtendedWirelessDeviceDiscovery() as soon as:

- A single specified device is found and validated.

- An ephemeral device is chosen.

- A user interactively selects a device from a list.

- The discovery process concludes in a non-interactive environment (like CI).

This change prevents the devicectl process from running concurrently with the app build, thus resolving the race condition and improving the reliability of iOS builds.

### Testing: 

- This is tested by checking the bringup to make sure the following tests are passing:
* [Mac_arm64_ios integration_test_test_ios](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_arm64_ios%20integration_test_test_ios?limit=200)
* [Mac_ios wide_gamut_ios](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20wide_gamut_ios?limit=200)
* [Mac_ios native_assets_ios](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20native_assets_ios?limit=200)
* [Mac_ios spell_check_test](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20spell_check_test?limit=200)
* [Mac_ios channels_integration_test_ios](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_ios%20channels_integration_test_ios?limit=200)

fixes #174444 

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/flutter_tools/lib/src/runner/target_devices.dart
```diff
@@ -494,12 +501,14 @@ class TargetDevicesWithExtendedWirelessDeviceDiscovery extends TargetDevices {
           // If the only matching device is not paired, print a warning
           if (!matchedDevice.isPaired) {
             _logger.printStatus(flutterSpecifiedDeviceUnpaired(matchedDevice.displayName));
+            stopExtendedWirelessDeviceDiscovery();
```
vashworth: Rather than calling stopExtendedWirelessDeviceDiscovery() during every return, perhaps put it in a `try` with a `finally`:

```
try {
  // do stuff
  return x;
} finally {
  stopExtendedWirelessDeviceDiscovery();
}

File: packages/flutter_tools/lib/src/runner/flutter_command.dart [hunk 2]
```diff
@@ -1950,15 +1950,18 @@ abstract class FlutterCommand extends Command<void> {
       includeDevicesUnsupportedByProject: includeDevicesUnsupportedByProject,
     );
     if (deviceList == null) {
+      _targetDevices.stopExtendedWirelessDeviceDiscovery();
```
vashworth: Do you need to call this here if `_targetDevices.findAllTargetDevices` is calling it?

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -106,6 +106,7 @@ class XCDevice {
     _stopObservingTetheredIOSDevices();
     _usbDeviceWaitProcess?.kill();
     _wifiDeviceWaitProcess?.kill();
+    _coreDeviceControl.stopListDevices();
```
vashworth: So this will kill the `devicectl` process, but it won't prevent it from starting. Let's say hypothetically that `getAvailableIOSDevices` is still running the `_getAllDevices` function when `dispose` is called. It will then proceed to start `_coreDeviceControl.getCoreDevices`

https://github.com/flutter/flutter/blob/9c101b75a77000b95bc603ecd2372b16b6dd5639/packages/flutter_tools/lib/src/macos/xcdevice.dart#L446-L463

vashworth: Maybe could do something like this?
 ```dart
  Completer<void>? _cancelCompleter;
  void cancelListDevices() {
    _cancelCompleter?.complete();
  }

  Future<List<IOSDevice>> getAvailableIOSDevices({Duration? timeout}) async {
    _cancelCompleter = Completer<void>();
    final Completer<void> cancelCompleter = _cancelCompleter!;

    final List<Object>? allAvailableDevices = await _getAllDevices(
      timeout: timeout ?? const Duration(seconds: 2),
    );

    if (allAvailableDevices == null) {
      return const <IOSDevice>[];
    }

    if (cancelCompleter.isCompleted) {
      return const <IOSDevice>[];
    }

    final coreDeviceMap = <String, IOSCoreDevice>{};
    if (_xcode.isDevicectlInstalled) {
      final Process coreDeviceProcess = await _coreDeviceControl.startGetCoreDevices();
      unawaited(
        cancelCompleter.future.whenComplete(() {
          coreDeviceProcess.kill(); // I'm not sure what happens if you kill a process that is completed, so may need to wrap this in a try catch or something
        }),
      );
      await coreDeviceProcess.exitCode;
      final List<IOSCoreDevice> coreDevices = _coreDeviceControl.getCoreDevicesFromProcess(coreDeviceProcess);
```

File: packages/flutter_tools/lib/src/ios/devices.dart [hunk 4]
```diff
@@ -239,6 +239,8 @@ class IOSDevices extends PollingDeviceDiscovery {
   @override
   Future<void> stopPolling() async {
     await _observedDeviceEventsSubscription?.cancel();
+    _observedDeviceEventsSubscription = null;
+    xcdevice.dispose();
```
vashworth: So `_observedDeviceEventsSubscription` is used for listening for devices to attach when using an IDE like VSCode. We don't want to stop that process in our use case. Rather than using `stopPolling`, you'll probably need to introduce a new method

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -371,7 +381,17 @@ class IOSCoreDeviceControl {
     ];
 
     try {
-      final RunResult result = await _processUtils.run(command, throwOnError: true);
+      _devicectlListProcess = await _processUtils.start(command);
```
vashworth: What would happen if this function is called twice?

If I remember correctly, the way device discovery works is that we kick off the device discovery process early using `startExtendedWirelessDeviceDiscovery` because wireless device discovery takes longer. Then later when we do actual device discovery for attached devices, we do it again but with a lower timeout. So it's possible that this function is being called multiple times.

vashworth (changes requested):

packages/flutter_tools/lib/src/ios/core_devices.dart
vashworth: See comment below for my recommendation: https://github.com/flutter/flutter/pull/175659/files#r2372644434

vashworth:
We synced offline. Here's a summary:

Okay so here's the current flow for wireless device discovery:

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/runner/target_devices.dart#L388-L395

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/device.dart#L215-L224

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/device.dart#L516-L519

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/device.dart#L526-L545

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/ios/devices.dart#L245-L251

https://github.com/flutter/flutter/blob/4a042046a0f182b2b0729ce7a7271276b08d4f19/packages/flutter_tools/lib/src/macos/xcdevice.dart#L444-L462

We need a solution that cancels "getAvailableIOSDevices", but only for the wireless device discovery. IDE device polling should not be impacted.

One potential solution would be to introduce a new method just for wireless device discovery.

In `xcdevice.dart`, add a `getAvailableIOSDevicesForWirelessDiscovery`, which does essentially the same thing as `getAvailableIOSDevices`, but also uses a Completer that will kill the `devicectl` process or will return early:

```dart
Completer<void>? _cancelCompleter;
 void cancelListDevices() {
   _cancelCompleter?.complete();
 }

 Future<List<IOSDevice>> getAvailableIOSDevicesForWirelessDiscovery({Duration? timeout}) async {
   _cancelCompleter = Completer<void>();
   final Completer<void> cancelCompleter = _cancelCompleter!;

   final List<Object>? allAvailableDevices = await _getAllDevices(
     timeout: timeout ?? const Duration(seconds: 2),
   );

   if (allAvailableDevices == null) {
     return const <IOSDevice>[];
   }

   if (cancelCompleter.isCompleted) {
     return const <IOSDevice>[];
   }

    final coreDeviceMap = <String, IOSCoreDevice>{};
   if (_xcode.isDevicectlInstalled) {
     final Process coreDeviceProcess = await _coreDeviceControl.startGetCoreDevices();
     unawaited(
       cancelCompleter.future.whenComplete(() {
         coreDeviceProcess.kill(); // I'm not sure what happens if you kill a process that is completed, so may need to wrap this in a try catch or something
       }),
     );
     await coreDeviceProcess.exitCode;
     final List<IOSCoreDevice> coreDevices = _coreDeviceControl.getCoreDevicesFromProcess(coreDeviceProcess);
     }

    return parseDevices(allAvailableDevices, coreDeviceMap);
  }
  ```
  
  Then in `ios/devices.dart`, add a flag to `pollingGetDevices` that indicates it's for the wireless discovery call and call `getAvailableIOSDevicesForWirelessDiscovery`:
  
  ```dart
  @override
  Future<List<Device>> pollingGetDevices({Duration? timeout, bool forWirelessDiscovery = false}) async {
    if (!_platform.isMacOS) {
      throw UnsupportedError('Control of iOS devices or simulators only supported on macOS.');
    }
    if (forWirelessDiscovery) {
      return xcdevice.getAvailableIOSDevicesForWirelessDiscovery(timeout: timeout);
    }
    return xcdevice.getAvailableIOSDevices(timeout: timeout);
  }
```

The `forWirelessDiscovery` flag would need to be piped through `pollingGetDevices`, `_populateDevices`, `discoverDevices` from `refreshExtendedWirelessDeviceDiscoverers` in `devices.dart`:
```dart
  Future<void> refreshExtendedWirelessDeviceDiscoverers({
    Duration? timeout,
    DeviceDiscoveryFilter? filter,
  }) async {
    await Future.wait<List<Device>>(<Future<List<Device>>>[
      for (final DeviceDiscovery discoverer in _platformDiscoverers)
        if (discoverer.requiresExtendedWirelessDeviceDiscovery)
          discoverer.discoverDevices(timeout: timeout, forWirelessDiscovery: true),
    ]);
  }
```

Then could do something like
```dart
  void stopExtendedWirelessDeviceDiscoverers() {
    for (final DeviceDiscovery discoverer in _platformDiscoverers) {
      if (discoverer is IOSDevices) {
        discoverer.cancelListDevices();
      }
    }
  }
```

packages/flutter_tools/lib/src/ios/core_devices.dart
okorohelijah: done.

packages/flutter_tools/lib/src/ios/devices.dart
okorohelijah: done.

packages/flutter_tools/lib/src/macos/xcdevice.dart
okorohelijah: done.

packages/flutter_tools/lib/src/runner/target_devices.dart
okorohelijah: done

camsim99 (commented):
Android change LGTM

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -452,7 +462,9 @@ class XCDevice {
 
     final coreDeviceMap = <String, IOSCoreDevice>{};
     if (_xcode.isDevicectlInstalled) {
-      final List<IOSCoreDevice> coreDevices = await _coreDeviceControl.getCoreDevices();
+      final List<IOSCoreDevice> coreDevices = await _coreDeviceControl.getCoreDevices(
+        timeout: timeout ?? const Duration(seconds: 2),
```
vashworth: Why is this timeout added here? If I remember correctly, this will get overriden since it's too low

https://github.com/flutter/flutter/blob/999fd04e16246c7694b9caf61e242ba46e6759ee/packages/flutter_tools/lib/src/ios/core_devices.dart#L348-L356

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -498,8 +510,51 @@ class XCDevice {
     //  },
     // ...
 
+    return _parseXcdeviceList(
+      allAvailableDevices: allAvailableDevices,
+      coreDeviceMap: coreDeviceMap,
```
vashworth: Why not pass in `List<IOSCoreDevice> coreDevices` here and do the `if (device.udid == null)` check there? Both `getAvailableIOSDevices` and `getAvailableIOSDevicesForWirelessDiscovery` do that logic, so makes sense to move it into `_parseXcdeviceList`

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -498,8 +510,51 @@ class XCDevice {
     //  },
     // ...
```
vashworth, line 539: The comment here can be moved to _parseXcdeviceList

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -498,8 +510,51 @@ class XCDevice {
...
+      timeout: timeout ?? const Duration(seconds: 2),
+    );
+    if (allAvailableDevices == null || cancelCompleter.isCompleted) {
+      return const <IOSDevice>[];
+    }
+
+    final coreDeviceMap = <String, IOSCoreDevice>{};
+    if (_xcode.isDevicectlInstalled) {
+      final List<IOSCoreDevice> coreDevices = await _coreDeviceControl.getCoreDevices(
+        timeout: timeout ?? const Duration(seconds: 30),
```
vashworth: Why is this timeout added here?

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -304,6 +304,8 @@ class IOSCoreDeviceControl {
   final Xcode _xcode;
   final FileSystem _fileSystem;
 
+  Process? _devicectlListProcess;
```
vashworth: You shouldn't need to track the process in a global variable, it should all be done within the `_listCoreDevices` function. For example, something like this should work:

```dart
Future<List<IOSCoreDevice>> getCoreDevices({
    Duration timeout = const Duration(seconds: _minimumTimeoutInSeconds),
    Completer<void>? cancelCompleter,
  }) async {
    final List<Object?> devicesSection = await _listCoreDevices(timeout: timeout, cancelCompleter:cancelCompleter);
    return <IOSCoreDevice>[
      for (final Object? deviceObject in devicesSection)
        if (deviceObject is Map<String, Object?>)
          IOSCoreDevice.fromBetaJson(deviceObject, logger: _logger),
    ];
  }
  
  ...
  
  Future<List<Object?>> _listCoreDevices({
    Duration timeout = const Duration(seconds: _minimumTimeoutInSeconds),
     Completer<void>? cancelCompleter,
  }) async {
    ...
    try {
      final Process process = await _processUtils.start(command);
      unawaited(
        cancelCompleter?.future.whenComplete(() {
          process.kill(); // I'm not sure what happens if you kill a process that is completed, so may need to wrap this in a try catch or something
        }),
      );

      await process.exitCode;
      final String stdout = await utf8.decodeStream(process.stdout);
      final String stderr = await utf8.decodeStream(process.stderr);
      
      var isToolPossiblyShutdown = false;
      if (_fileSystem is ErrorHandlingFileSystem) {
      ...
```

packages/flutter_tools/lib/src/ios/devices.dart
```diff
@@ -239,14 +239,21 @@ class IOSDevices extends PollingDeviceDiscovery {
   @override
   Future<void> stopPolling() async {
     await _observedDeviceEventsSubscription?.cancel();
+    _observedDeviceEventsSubscription = null;
+    xcdevice.cancelWirelessDiscovery();
```
vashworth, line 243: You shouldn't use `stopPolling` because this will also kill IDE device discovery. I recommend making a new in src/device.dart that does something like

```dart
void stopExtendedWirelessDeviceDiscoverers() {
    final IOSDevices? iosDiscoverer = _platformDiscoverers.whereType<IOSDevices>().firstOrNull;
    await iosDiscoverer?.cancelWirelessDiscovery();
  }
```

packages/flutter_tools/lib/src/runner/target_devices.dart
```diff
@@ -605,13 +616,15 @@ class TargetDevicesWithExtendedWirelessDeviceDiscovery extends TargetDevices {
   ) async {
     final Device? ephemeralDevice = _deviceManager.getSingleEphemeralDevice(attachedDevices);
     if (ephemeralDevice != null) {
+      stopExtendedWirelessDeviceDiscovery();
```
vashworth: Is this needed? `_handleRemainingDevices` is called within `findAllTargetDevices`, which calls `stopExtendedWirelessDeviceDiscovery`

packages/flutter_tools/lib/src/runner/target_devices.dart
```diff
@@ -703,6 +718,7 @@ class TargetDevicesWithExtendedWirelessDeviceDiscovery extends TargetDevices {
     // be prompted again.
     _deviceManager.specifiedDeviceId = chosenDevice.id;
 
+    stopExtendedWirelessDeviceDiscovery();
```
vashworth: Same, is this needed?

packages/flutter_tools/lib/src/macos/xcdevice.dart
okorohelijah: I added the 30 seconds timeout because of the wirelessDiscovery which usually takes more time

packages/flutter_tools/lib/src/macos/xcdevice.dart
okorohelijah: done

packages/flutter_tools/lib/src/macos/xcdevice.dart
okorohelijah: done

packages/flutter_tools/lib/src/macos/xcdevice.dart
okorohelijah: done

packages/flutter_tools/lib/src/ios/core_devices.dart
okorohelijah: done

packages/flutter_tools/lib/src/ios/devices.dart
okorohelijah: done

packages/flutter_tools/lib/src/runner/target_devices.dart
okorohelijah: done

packages/flutter_tools/lib/src/runner/target_devices.dart
okorohelijah: done

packages/flutter_tools/lib/src/device.dart
```diff
@@ -454,6 +468,9 @@ abstract class DeviceDiscovery {
   ///
   /// For example, 'windows' or 'linux'.
   List<String> get wellKnownIds;
+
+  /// Stop the discoverer from polling for devices.
+  void stopPolling() {}
```
vashworth: Is this still needed?

packages/flutter_tools/lib/src/device.dart
```diff
@@ -493,11 +510,15 @@ abstract class PollingDeviceDiscovery extends DeviceDiscovery {
     });
   }
 
+  @override
```
vashworth: Is this still needed?

packages/flutter_tools/lib/src/device.dart
```diff
@@ -219,10 +219,20 @@ abstract class DeviceManager {
...
       for (final DeviceDiscovery discoverer in _platformDiscoverers)
         if (discoverer.requiresExtendedWirelessDeviceDiscovery)
-          discoverer.discoverDevices(timeout: timeout),
+          discoverer.discoverDevices(timeout: timeout, forWirelessDiscovery: true),
     ]);
   }
 
+  /// Stop any running extended wireless device discoverers.
+  void stopExtendedWirelessDeviceDiscoverers() {
+    final DeviceDiscovery? iosDiscoverer = _platformDiscoverers
```
vashworth: It's possible there could be more discoverers that use `requiresExtendedWirelessDeviceDiscovery`, so I wouldn't give it an iOS specific name.

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -370,65 +369,73 @@ class IOSCoreDeviceControl {
...
 
+    Process? process;
     try {
-      final RunResult result = await _processUtils.run(command, throwOnError: true);
-      var isToolPossiblyShutdown = false;
-      if (_fileSystem is ErrorHandlingFileSystem) {
-        final FileSystem delegate = _fileSystem.fileSystem;
-        if (delegate is LocalFileSystem) {
-          isToolPossiblyShutdown = delegate.disposed;
-        }
```
vashworth, line 380: Why is the `isToolPossiblyShutdown` stuff removed?

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -370,65 +369,73 @@ class IOSCoreDeviceControl {
...
+      _logger.printError('devicectl returned unexpected JSON response: $stringOutput');
+      return const <Object?>[];
+    } on ProcessException catch (e) {
+      _logger.printError('Error executing devicectl: $e');
+      return const <Object?>[];
+    } on FileSystemException catch (e) {
+      _logger.printTrace('Error reading devicectl output: $e');
+      return const <Object?>[];
+    } on FormatException {
+      _logger.printError('devicectl returned non-JSON response.');
```
vashworth: Why is this moved out? Now it won't print the `stringOutput`

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -370,65 +369,73 @@ class IOSCoreDeviceControl {
...
-          }
+      if (exitCode != 0) {
+        final String stderr = await utf8.decodeStream(process.stderr);
+        _logger.printError('devicectl exited with a non-zero exit code: $exitCode');
+        _logger.printTrace('devicectl stdout:\n$stdout');
+        _logger.printTrace('devicectl stderr:\n$stderr');
+        return const <Object?>[];
+      }
+
+      _logger.printTrace(stdout);
```
vashworth: Is this needed?

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -370,65 +369,73 @@ class IOSCoreDeviceControl {
...
-      try {
-        final Object? decodeResult = (json.decode(stringOutput) as Map<String, Object?>)['result'];
-        if (decodeResult is Map<String, Object?>) {
-          final Object? decodeDevices = decodeResult['devices'];
-          if (decodeDevices is List<Object?>) {
-            return decodeDevices;
-          }
+      if (exitCode != 0) {
+        final String stderr = await utf8.decodeStream(process.stderr);
+        _logger.printError('devicectl exited with a non-zero exit code: $exitCode');
```
vashworth: I know this existed before your PR, but can you change all of the `_logger.printError` to `_logger.printTrace`? We only want to use `printError` if the error message is something actionable for the developer to do.

packages/flutter_tools/lib/src/ios/core_devices.dart
```diff
@@ -370,65 +369,73 @@ class IOSCoreDeviceControl {
...
+      _logger.printError('Error executing devicectl: $e');
+      return const <Object?>[];
+    } on FileSystemException catch (e) {
+      _logger.printTrace('Error reading devicectl output: $e');
+      return const <Object?>[];
+    } on FormatException {
+      _logger.printError('devicectl returned non-JSON response.');
+      return const <Object?>[];
     } finally {
-      ErrorHandlingFileSystem.deleteIfExists(tempDirectory, recursive: true);
```
vashworth, line 422: Why is this replaced?

packages/flutter_tools/lib/src/macos/xcdevice.dart
```diff
@@ -450,13 +459,46 @@ class XCDevice {
...
+      timeout: timeout ?? const Duration(seconds: 2),
+    );
+    if (allAvailableDevices == null || cancelCompleter.isCompleted) {
+      return const <IOSDevice>[];
+    }
+
+    var coreDevices = <IOSCoreDevice>[];
+    if (_xcode.isDevicectlInstalled) {
+      coreDevices = await _coreDeviceControl.getCoreDevices(
+        timeout: timeout ?? const Duration(seconds: 30),
```
vashworth: Why does this have a timeout duration of 30 seconds?

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
```diff
@@ -2753,7 +2753,8 @@ invalid JSON
           ),
         );
 
-        await deviceControl.getCoreDevices();
+        await expectLater(deviceControl.getCoreDevices(), completion(isEmpty));
+        //  await deviceControl.getCoreDevices();
```
vashworth: Remove?

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
```diff
@@ -2786,24 +2787,8 @@ invalid JSON
...
-          ),
-        );
-        expect(
-          logger.errorText,
-          contains(
-            'After running the command xcrun devicectl list devices '
-            '--timeout 5 --json-output ${tempFile.path} the file\n'
-            '${tempFile.path} was expected to exist, but it did not',
-          ),
-        );
```
vashworth, line 2806: Why is this removed?

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
```diff
@@ -3174,16 +3154,10 @@ invalid JSON
           '00001234-0001234A3C03401E.coredevice.local',
           '123456BB5-AEDE-7A22-B890-1234567890DD.coredevice.local',
         ]);
-        expect(devices[0].connectionProperties?.pairingState, 'paired');
-        expect(devices[0].connectionProperties?.potentialHostnames, <String>[
-          '00001234-0001234A3C03401E.coredevice.local',
-          '123456BB5-AEDE-7A22-B890-1234567890DD.coredevice.local',
-        ]);
```
vashworth, line 3181: Why is this removed?

packages/flutter_tools/test/general.shard/ios/devices_test.dart
```diff
@@ -864,10 +864,12 @@ void main() {
       xcdevice.isInstalled = true;
       xcdevice.devices.add(<IOSDevice>[device1]);
 
-      final List<Device> devices = await iosDevices.pollingGetDevices();
+      final List<Device> devices = await iosDevices.pollingGetDevices(forWirelessDiscovery: true);
```
vashworth, line 885: This test is specifically title for "attached devices". Perhaps make a new test specific to wireless devices?

packages/flutter_tools/test/src/fake_process_manager.dart
```diff
@@ -339,6 +344,10 @@ abstract class FakeProcessManager implements ProcessManager {
       completer: fakeCommand.completer,
       outputFollowsExit: fakeCommand.outputFollowsExit,
     );
+    if (fakeCommand.onProcess != null) {
+      fakeCommand.onProcess!(process);
+    }
+    return process;
```
vashworth: Why is this changed?

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
```diff
@@ -3027,22 +3010,20 @@ invalid JSON
               tempFile.path,
             ],
             onRun: (_) {
-              expect(tempFile, exists);
+              tempFile.createSync(recursive: true);
```
vashworth: Why is this changed?

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
vashworth, line 1: You should have a test here that tests the cancel completer

packages/flutter_tools/lib/src/device.dart
```diff
@@ -219,10 +219,20 @@ abstract class DeviceManager {
...
 
+  /// Stop any running extended wireless device discoverers.
+  void stopExtendedWirelessDeviceDiscoverers() {
+    final DeviceDiscovery? extendedWirelessDiscoverer = _platformDiscoverers
+        .where((DeviceDiscovery d) => d.requiresExtendedWirelessDeviceDiscovery)
+        .firstOrNull;
+    if (extendedWirelessDiscoverer is PollingDeviceDiscovery) {
+      extendedWirelessDiscoverer.cancelWirelessDiscovery();
+    }
+  }
```
vashworth, line 234: nit: The name of the function make it sound like it's stopping any that qualify rather than just the first. Perhaps loop instead of using `firstOrNull`?

```dart
    for (final PollingDeviceDiscovery deviceDiscoverer in _platformDiscoverers.whereType<PollingDeviceDiscovery>()) {
      if (deviceDiscoverer.requiresExtendedWirelessDeviceDiscovery) {
        deviceDiscoverer.cancelWirelessDiscovery();
      }
    }
```

packages/flutter_tools/test/general.shard/ios/core_devices_test.dart
```diff
@@ -3439,7 +3430,45 @@ invalid JSON
...
+                expect(tempFile, exists);
+              },
+            ),
+          );
+
+          final Future<List<IOSCoreDevice>> devicesFuture = deviceControl.getCoreDevices(
+            cancelCompleter: cancelCompleter,
+          );
+
+          cancelCompleter.complete();
```
vashworth, line 3470: I don't think this actually tests the process is killed since the fake process will complete basically immediately. Perhaps try add a completer to `FakeCommand` to prevent it from finishing and see if it's possible for a `FakeCommand` to get killed

packages/flutter_tools/test/general.shard/ios/devices_test.dart
```diff
@@ -1095,7 +1122,14 @@ class FakeXcdevice extends Fake implements XCDevice {
 
   @override
   Future<List<IOSDevice>> getAvailableIOSDevices({Duration? timeout}) async {
-    return devices[getAvailableIOSDevicesCount++];
+    getAvailableIOSDevicesCount++;
+    return devices.removeAt(0);
```
vashworth: Why does it remove the device?

packages/flutter_tools/test/general.shard/macos/xcode_test.dart
```diff
@@ -1518,6 +1518,46 @@ void main() {
...
+
+            await pumpEventQueue();
+
+            xcdevice.cancelWirelessDiscovery();
+            getCoreDevicesCompleter.complete();
+
+            final List<IOSDevice> devices = await futureDevices;
+            expect(devices, isEmpty);
+
+            expect(fakeProcessManager, hasNoRemainingExpectations);
```
vashworth, line 1582: This is test doesn't really validate anything. Since no devices are set in the `FakeIOSCoreDeviceControl` and `stdout` of the `xcdevice` FakeCommand is empty, it'll always return empty devices. Also, since it's using a fake for `FakeIOSCoreDeviceControl`, it'll never actually do another process.

To validate that it's really cancelling, it needs to have devices set in the fake `FakeIOSCoreDeviceControl` and `FakeCommand` to validate that it's retuning no devices when it previously was returning devices.

packages/flutter_tools/test/general.shard/ios/devices_test.dart
okorohelijah: it was added to get a different device to be used each time it is called but they do the same thing

vashworth (approved):

PR #175710, opened 2025-09-21
feat: add infinite carousel support

### LINKED ISSUE #161369: CarouselView infinite scroll
### Use case

Currently there is a start and an end to the carousel. However a common use case is to have a carousel where the items repeat themselves.

The idea is to not have a hole on the side when the scroll extent is at the edge, so instead of this:

image

The hole on the left side of the first item, could be replaced by the last item.

### Proposal

```
CarouselView(
  isInfinite: true,
  // ...
)
```

rkishan516:
Add infinite carousel

fixes: #161369 

## Pre-launch Checklist

- [x] I signed the [CLA].


cedvdb:
Would it be fair to make this the default behavior ?

packages/flutter/test/material/carousel_test.dart
```diff
@@ -1975,6 +1975,84 @@ void main() {
...
+            children: List<Widget>.generate(3, (int index) {
+              return Center(child: Text('Item $index'));
+            }),
+          ),
+        ),
+      ),
+    );
+
+    // Initial state - should show items
+    expect(find.textContaining('Item'), findsAtLeastNWidgets(1));
```
QuncCccccc: Can we try to expect all item 0, 1 and 2 are on the screen?

QuncCccccc (commented):
Sorry for the late reply! I tried the change in carousel.0.dart but seems when we set `infinite: true`, `CarouselView.weighted()` will throw error and the carousel items cannot show up correctly. Could you double check whether the behavior is working?

rkishan516:
> Sorry for the late reply! I tried the change in carousel.0.dart but seems when we set infinite: true, CarouselView.weighted() will throw error and the carousel items cannot show up correctly. Could you double check whether the behavior is working?

Let me take a look at it, I will also update test to capture this case.

QuncCccccc:
> Let me take a look at it, I will also update test to capture this case.

Sounds good! Let me know whenever it's ready to review:)

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -531,6 +546,13 @@ class _CarouselViewState extends State<CarouselView> {
   }
 
   Widget _buildCarouselItem(int index) {
+    // For infinite scrolling, wrap the index to the actual children range
+    if (widget.infinite && widget.children.isNotEmpty) {
+      index = index % widget.children.length;
+      if (index < 0) {
+        index += widget.children.length;
+      }
```
QuncCccccc: Is this possible at all? If not we can remove this check.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -594,7 +616,11 @@ class _CarouselViewState extends State<CarouselView> {
 
   Widget _buildSliverCarousel(ThemeData theme) {
     // Determine the child count and builder based on whether we're using lazy loading
-    final int? childCount = widget.itemBuilder != null ? widget.itemCount : widget.children.length;
+    final int? childCount = widget.infinite
+        ? null
```
QuncCccccc, line 670: In the API doc, we can also mention itemBuilder and itemCount will be ignored when infinite is not true:)

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -833,6 +877,10 @@ class _RenderSliverFixedExtentCarousel extends RenderSliverFixedExtentBoxAdaptor
       return 0;
     }
 
+    if (infinite) {
+      return (scrollOffset / maxExtent).floor();
```
QuncCccccc: This calculation seems the same as the non-infinite version.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -833,6 +877,10 @@ class _RenderSliverFixedExtentCarousel extends RenderSliverFixedExtentBoxAdaptor
...
+    }
+
     final int firstVisibleIndex = (constraints.scrollOffset / maxExtent).floor();
```
QuncCccccc: I think we can just use scrollOffset here instead of constraints.scrollOffset.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -848,6 +896,13 @@ class _RenderSliverFixedExtentCarousel extends RenderSliverFixedExtentBoxAdaptor
     double itemExtent,
   ) {
     if (maxExtent > 0.0) {
+      if (infinite) {
+        // For infinite scrolling, calculate the max visible index
+        // based on the viewport size and allow negative indices
+        final double viewportItems = constraints.viewportMainAxisExtent / maxExtent;
+        return (scrollOffset / maxExtent + viewportItems).ceil();
+      }
```
QuncCccccc: This seems not accurate as well. My understanding is the max index should just be scrollOffset / maxExtent - 1. `scrollOffset / maxExtent` is how many items we can include and `- 1` is to indicate the item index.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1209,23 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
   ) {
     final int? childCount = childManager.estimatedChildCount;
+
+    // For infinite scrolling, calculate how many items fit in the viewport
+    if (infinite && childCount == null) {
+      double visibleItemsTotalExtent = _distanceToLeadingEdge;
+      int index = _firstVisibleItemIndex + 1;
+      // Estimate a reasonable number of items to check
+      while (visibleItemsTotalExtent < constraints.viewportMainAxisExtent &&
+          index < _firstVisibleItemIndex + 100) {
```
QuncCccccc: Putting a hard-code number might be not safe here. We can get the upper bound by using the firstVisibleItemIndex and weights.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1298,16 +1391,22 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      if (lastIndex + 1 == childCount) {
+        trailingScrollOffset = indexToLayoutOffset(deprecatedExtraItemExtent, lastIndex);
+
+        trailingScrollOffset += math.max(
+          weights.last * extentUnit,
+          _buildItemExtent(lastIndex, _currentLayoutDimensions),
+        );
+        trailingScrollOffset += extraLayoutOffset;
+      } else {
+        trailingScrollOffset = indexToLayoutOffset(deprecatedExtraItemExtent, lastIndex + 1);
```
QuncCccccc: When infinite, we can just go here, instead of create another if statement.

packages/flutter/lib/src/material/carousel.dart
rkishan516: Have removed this, was adding this for bidirectional infinite carousel(droped that idea).

packages/flutter/lib/src/material/carousel.dart
rkishan516: Update API doc

packages/flutter/lib/src/material/carousel.dart
rkishan516: Merged both condition.

packages/flutter/lib/src/material/carousel.dart
rkishan516: Again, this was for bidirectional carousel. Have removed it.

packages/flutter/lib/src/material/carousel.dart
rkishan516: Updated this

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -460,8 +467,18 @@ class CarouselView extends StatefulWidget {
   /// The number of items in the carousel.
   ///
   /// If null, the carousel will continue to build items until [itemBuilder] returns null.
+  ///
+  /// When [infinite] is true, this parameter is ignored and the carousel will loop indefinitely.
```
QuncCccccc:
```suggestion
  /// When [infinite] is true, this parameter is ignored and the carousel will loop infinitely.
```

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1197,34 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      double visibleItemsTotalExtent = _distanceToLeadingEdge;
+      int index = _firstVisibleItemIndex + 1;
+      // Calculate upper bound based on viewport extent and minimum possible item extent
+      // In worst case, all items would be at minimum extent (minChildExtent)
+      // Using weights.length as a multiplier provides a reasonable buffer
+      final double safeMinExtent = math.max(minChildExtent, 1.0);
+      final int estimatedUpperBound =
+          _firstVisibleItemIndex +
+          clampDouble(
+            ((constraints.viewportMainAxisExtent / safeMinExtent).ceil() * weights.length)
```
QuncCccccc: `(constraints.viewportMainAxisExtent / safeMinExtent).ceil()` is how many items with minimum extent in the viewport, but why do we need to multiply it by weights length? Feels the result doesn't make sense.

seems `(constraints.viewportMainAxisExtent / safeMinExtent).ceil()` is safe enough?

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1745,11 +1837,13 @@ class CarouselController extends ScrollController {
     }
 
     final bool hasFlexWeights = _carouselState!._flexWeights?.isNotEmpty ?? false;
-    if (_carouselState!.widget.itemBuilder != null) {
-      final int? itemCount = _carouselState!.widget.itemCount;
-      index = itemCount != null ? index.clamp(0, itemCount - 1) : 0;
-    } else {
-      index = index.clamp(0, _carouselState!.widget.children.length - 1);
+    if (!_carouselState!.widget.infinite) {
+      if (_carouselState!.widget.itemBuilder != null) {
```
QuncCccccc: The original `if` `else` is to prevent the index to be out of range. Don't we need to constrain the index when infinite is true?

packages/flutter/test/material/carousel_test.dart
```diff
@@ -1975,6 +1975,149 @@ void main() {
...
+
+    // Initial state - should show Item 0
+    expect(find.text('Item 0'), findsAtLeastNWidgets(1));
+
+    // Scroll forward - should be able to scroll infinitely
+    await tester.drag(find.byType(CarouselView), const Offset(-400, 0));
+    await tester.pumpAndSettle();
+
+    // Should see one of the items
+    expect(find.textContaining('Item'), findsAtLeastNWidgets(1));
```
QuncCccccc: We should make the tests more accurate, like checking the layout rects for each item are correct. If scrolling forward by 500, each item extent is 200, then the visible item should be "item 2", "item 0", "item 1", and "item 2", right?

QuncCccccc: Here and everywhere else.

packages/flutter/lib/src/material/carousel.dart
rkishan516: Logically, it should be good, I just gave extra buffer, but i will change.

packages/flutter/lib/src/material/carousel.dart
rkishan516: for infinite, its already coming in constrain, it will be unnecessary operation, but I can remove the condition.

packages/flutter/test/material/carousel_test.dart
rkishan516: I have pushed the change.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1197,33 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
     double itemExtent,
   ) {
     final int? childCount = childManager.estimatedChildCount;
+
+    // For infinite scrolling, calculate how many items fit in the viewport
+    if (infinite && childCount == null) {
+      double visibleItemsTotalExtent = _distanceToLeadingEdge;
+      int index = _firstVisibleItemIndex + 1;
+      // Calculate upper bound based on viewport extent and minimum possible item extent
```
QuncCccccc:
```suggestion
      // Calculate upper bound based on viewport extent and minimum possible item extent.
```

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1197,33 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      // In worst case, all items would be at minimum extent (minChildExtent)
```
QuncCccccc: Remove this line if we don't use the multiplier.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1197,33 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      // Using weights.length as a multiplier provides a reasonable buffer
```
QuncCccccc: Remove this line if we don't use the multiplier.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1136,6 +1197,33 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      final double safeMinExtent = math.max(minChildExtent, 1.0);
+      final int estimatedUpperBound =
+          _firstVisibleItemIndex +
+          clampDouble(
+            (constraints.viewportMainAxisExtent / safeMinExtent).ceil().toDouble(),
```
QuncCccccc: I think no need to use clampDouble. `_firstVisibleItemIndex + constraints.viewportMainAxisExtent / safeMinExtent).ceil()` is safe enough.

QuncCccccc (commented):
Just left a few more comments. 

Do we have any reference for the behavior of infinite Carousel? Such as any examples from real app. I'm just thinking if we can see the infinite scrolling when we scroll forward, should we also see infinite scrolling when we scroll back? For example, if we layout item [0, 1, 2] initially, then we scroll back, should we see item [2, 0, 1] instead of seeing the boundary? WDYT?

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1666,6 +1787,41 @@ class _CarouselPosition extends ScrollPositionWithSingleContext implements _Caro
     _itemExtent = other._itemExtent;
   }
 
+  /// Returns the length of one complete cycle in pixels.
+  /// A cycle is the scroll distance needed to return to the same visual state.
```
justinmc, line 1843: Nit: Empty line of `///` here between these two sentences.

justinmc (commented):
This looks good to me but I'll defer to @QuncCccccc. I'm excited to have infinite carousel support.

I tried this out locally and it worked great, but I noticed one thing. This is probably working as intended or unrelated to this PR, but the layout of items near the edge of the screen is affected. See the "Item 2" text is wrapping:

image

QuncCccccc (commented):
The CarouselView works great to me! But in CarouselView.weighted, when `consumeMaxWeight` is false, and for example if we have a flexWeight to be [1, 7, 1], ideally the leading item should be item 0, but when infinite is true, the leading item becomes the last item of the carousel (item 9 in my example.)

Here's the demo:

https://github.com/user-attachments/assets/777b8499-e54b-4913-81bf-e693f8a8504e

QuncCccccc (commented):
When we scroll forward and the original first item appears again, will the index be 0 or $n$? The same question applies to backward scrolling. I manually tested the behavior using the example app, but the item indexing feels a bit problematic. Ideally, `controller.leadingIndex` should only range from 0 to (n-1), where n is the number of children in the carousel. Thanks a lot for continuing to work on this; I think we’re getting close!

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1645,7 +1766,11 @@ class _CarouselPosition extends ScrollPositionWithSingleContext implements _Caro
     // The subtraction may cause negative number for leading item. In this case,
     // constrain the leading item to 0.
     if (consumeMaxWeight && flexWeights != null) {
-      return math.max(leadingItem - flexWeights!.indexOf(flexWeights!.max), 0);
+      leadingItem = math.max(leadingItem - flexWeights!.indexOf(flexWeights!.max), 0);
+    }
+    // For infinite scrolling, wrap the index to the range [0, itemCount - 1].
+    if (infinite && itemCount != null && itemCount! > 0) {
```
QuncCccccc, line 1733: Can we add unit tests to verify the leadingItem shows up correctly for both scrolling forwards and backwards, when infinite is true:)

packages/flutter/test/material/carousel_test.dart
```diff
@@ -1975,6 +1975,270 @@ void main() {
...
+              children: List<Widget>.generate(3, (int index) {
+                return Center(child: Text('Item $index'));
+              }),
+            ),
+          ),
+        ),
+      );
+
+      // Verify animating to an index beyond the array length.
+      controller.animateToItem(5);
```
QuncCccccc, line 1999: Another issue I can think of is that how do we control animateToItem(x)? Because if infinite is true, then the xth item can be reached both backwards and forwards. How should the animation behave?

packages/flutter/lib/src/material/carousel.dart
rkishan516: Sure, will add.

packages/flutter/test/material/carousel_test.dart
rkishan516: I think current behaviour is always scroll forward, but we can go for shortest path (because I think thats how other carousel widgets work). WDYT?

QuncCccccc: I see! I think scrolling forward sounds good to me. I'm checking the current behavior.

QuncCccccc: Could you double check the scrolling behavior? When I want the carousel to animate to the next item, it seems scrolling back to get the "next" one.

https://github.com/user-attachments/assets/8ded14eb-49c8-4ec1-b52b-ef1e04eab5bb

rkishan516: Yaa, this was happening because we were always targeting first cycle. I have updated behaviour to have move forward no mattter in which cycle you are in.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -657,15 +678,31 @@ class _CarouselViewState extends State<CarouselView> {
...
+    final int? childCount = widget.infinite
+        ? null
+        : widget.itemBuilder != null
+        ? widget.itemCount
+        : widget.children.length;
+
+    NullableIndexedWidgetBuilder effectiveBuilder;
+    if (widget.itemBuilder != null) {
+      if (widget.infinite && widget.itemCount != null && widget.itemCount! > 0) {
+        final int itemCount = widget.itemCount!;
```
QuncCccccc, line 678: It seems we should remove the doc that we added for itemCount because itemCount is actively used here.

QuncCccccc (commented):
Sorry for the late response. Overall looks good but I'm still looking at some of the logic in this PR. I'll make sure we make progress this week!

packages/flutter/lib/src/material/carousel.dart
rkishan516: Updated

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -492,8 +499,18 @@ class CarouselView extends StatefulWidget {
...
   ///
   /// If null, the carousel will continue to build items until [itemBuilder] returns null.
+  ///
+  /// When [infinite] is true, the carousel will loop infinitely.
   final int? itemCount;
 
+  /// Whether the carousel should loop infinitely.
+  ///
+  /// If true, the carousel will create an infinite loop of items,
+  /// allowing continuous scrolling in both directions.
```
QuncCccccc, line 509: Worth to mention that when animateToItem is called, the infinite carousel will always scrolling forward to find the item.

QuncCccccc (dismissed):
LGTM. Thanks a lot for adding this valuable feature!

QuncCccccc (dismissed):
LGTM.

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -1165,6 +1237,27 @@ class _RenderSliverWeightedCarousel extends RenderSliverFixedExtentBoxAdaptor {
...
+      double visibleItemsTotalExtent = _distanceToLeadingEdge;
+      int index = _firstVisibleItemIndex + 1;
+      // Calculate upper bound based on viewport extent and minimum possible item extent.
+      // In worst case, all items would be at minimum extent i.e. minChildExtent.
+      final double safeMinExtent = math.max(minChildExtent, 1.0);
+      final int estimatedUpperBound =
+          _firstVisibleItemIndex + (constraints.viewportMainAxisExtent / safeMinExtent).ceil();
+      while (visibleItemsTotalExtent < constraints.viewportMainAxisExtent &&
+          index < estimatedUpperBound) {
+        visibleItemsTotalExtent += _buildItemExtent(index, _currentLayoutDimensions);
```
QuncCccccc:
```suggestion
        visibleItemsTotalExtent += _buildItemExtent(index, layoutDimensions);
```

QuncCccccc: Sorry this should be `layoutDimensions`. I think I messed it up when I was handling the conflict.

QuncCccccc (dismissed):

QuncCccccc (approved):

hannah-hyj (approved):
LGTM

Issue #148026, opened 2024-05-08
FontWeight should subsume `wght` in FontVariation


gnprice:
### Problem

Flutter has two ways to specify the font weight in a TextStyle: you can pass either a `FontWeight` value or a `FontVariation` with variation `wght`. For example these look like `TextStyle(fontWeight: FontWeight.w500)` or `TextStyle(fontVariations: [FontVariation('wght', 500)])`.

The trouble is that, at present, neither of these subsumes the job of the other, and nor do they interact well with each other.

* Setting only `fontWeight` doesn't work if the font that gets used turns out to offer variable weight. This is a pitfall that even Flutter's own Material implementation falls into: https://github.com/flutter/flutter/issues/109308#issuecomment-2067918226

* Setting only a `wght` variation doesn't work, I believe, if the font that gets used turns out *not* to offer variable weight.

* Setting both a `fontWeight` and a `wght` variation can have interacting effects:
  https://github.com/flutter/flutter/issues/136779
  https://github.com/zulip/zulip-flutter/issues/500
  These effects don't follow any documented pattern, and don't seem desirable.

In general it's impossible for an app to know in advance whether the relevant font will offer variable weight: if it's displaying any kind of user-controlled text, the text may be in a variety of scripts, which may not be covered by the app's primary font and may fall back to a variety of other fonts.

As a result, there doesn't seem to be any existing way for a Flutter app to reliably control the font weight of its text, if any of the fonts it uses offer variable weight.

### Proposed solution

When `fontWeight` is specified, interpret it as the exact font weight to use, regardless of whether the font offers variable weight and regardless of whether a `wght` variation is also present. Consult any `wght` variation only as a fallback, for compatibility, when `fontWeight` is absent.

In order to support the same fine-grained range of possible weights currently expressed with a `wght` FontVariation, add a constructor `const FontWeight(this.value)` that accepts an int. (The `FontWeight` type is already a class with a field `int value`.)

This is the same API that [CSS offers](https://github.com/flutter/flutter/issues/136779#issuecomment-1947599504): there's just one attribute `font-weight`, it accepts any number from 1 to 1000, and it works regardless of whether the font ultimately used offers variable weight. This design seems to have worked out well for CSS.

### Implementation

[...]

PR #175771, opened 2025-09-22
Set the font weight variation axis based on the text style's FontWeight

jason-simmons:
This makes it possible for applications to set a FontWeight and get the expected result for both variable fonts and fonts that provide separate assets for each weight.

See https://github.com/flutter/flutter/issues/148026

### DIFF
--- a/engine/src/flutter/lib/ui/text.dart
+++ b/engine/src/flutter/lib/ui/text.dart
@@ -56,15 +56,18 @@ enum FontStyle {
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
 /// When using these fonts, applications can specify [FontWeight] instances
-/// constructed using values other than the predefined values.
+/// constructed using values other than the predefined values. For these fonts,
+/// [FontWeight] will set the value of the `wght` axis (producing the same
+/// results as explicitly setting that attribute using [FontVariation.weight]).
 class FontWeight {
-  /// Create a [FontWeight] object, which can be added to a [TextStyle] to
+  /// Creates a [FontWeight] object, which can be added to a [TextStyle] to
   /// select the thickness of a font's glyphs.
   const FontWeight(this.value)
     : assert(value >= 1, 'Font weight must be between 1 and 1000'),
       assert(value <= 1000, 'Font weight must be between 1 and 1000');
 
   /// The encoded integer value of this font weight.
+  @Deprecated('Use value, which is more precise.')
   int get index => (value ~/ 100 - 1).clamp(0, 8);
 
   /// The thickness value of this font weight.
--- a/engine/src/flutter/lib/ui/text.dart
+++ b/engine/src/flutter/lib/ui/text.dart
@@ -58,6 +58,8 @@ enum FontStyle {
 /// When using these fonts, applications can specify [FontWeight] instances
 /// constructed using values other than the predefined values.
 class FontWeight {
+  /// Create a [FontWeight] object, which can be added to a [TextStyle] to
+  /// select the thickness of a font's glyphs.
   const FontWeight(this.value)
     : assert(value >= 1, 'Font weight must be between 1 and 1000'),
       assert(value <= 1000, 'Font weight must be between 1 and 1000');
@@ -121,6 +123,9 @@ class FontWeight {
     }
     return other is FontWeight && other.value == value;
   }
+
+  @override
+  int get hashCode => value;
 
   /// Linearly interpolates between two font weights.
   ///
--- a/engine/src/flutter/lib/ui/text.dart
+++ b/engine/src/flutter/lib/ui/text.dart
@@ -2232,6 +2232,7 @@ ByteData _encodeStrut(
   data.setInt32(0, bitmask, _kFakeHostEndian);
 
   assert(byteCount <= 24);
+  assert(bitmask >> 32 == 0, 'strut bitmask overflow: $bitmask');
   return ByteData.view(data.buffer, 0, byteCount);
 }
 
@@ -3560,9 +3561,9 @@ base class _NativeParagraphBuilder extends NativeFieldWrapperClass1 implements P
       final TextLeadingDistribution leadingDistribution =
           strutStyle._leadingDistribution ?? style._leadingDistribution;
       encodedStrutStyle = strutStyle._encoded;
-      int bitmask = encodedStrutStyle.getInt8(0);
+      int bitmask = encodedStrutStyle.getInt32(0, _kFakeHostEndian);
       bitmask |= (leadingDistribution.index) << 3;
-      encodedStrutStyle.setInt8(0, bitmask);
+      encodedStrutStyle.setInt32(0, bitmask, _kFakeHostEndian);
     } else {
       encodedStrutStyle = null;
     }
--- a/engine/src/flutter/lib/ui/text.dart
+++ b/engine/src/flutter/lib/ui/text.dart
@@ -53,42 +53,43 @@ enum FontStyle {
 /// with the name "Roboto" and the weight [FontWeight.w500].
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
-/// See [FontVariation.weight] for details.
+/// When using these fonts, applications can specify [FontWeight] instances
+/// constructed using values other than the predefined values.
 class FontWeight {
-  const FontWeight._(this.index, this.value);
+  const FontWeight(this.value);
 
   /// The encoded integer value of this font weight.
-  final int index;
+  int get index => (value ~/ 100 - 1).clamp(0, 8);
 
   /// The thickness value of this font weight.
   final int value;
 
   /// Thin, the least thick.
-  static const FontWeight w100 = FontWeight._(0, 100);
+  static const FontWeight w100 = FontWeight(100);
 
   /// Extra-light.
-  static const FontWeight w200 = FontWeight._(1, 200);
+  static const FontWeight w200 = FontWeight(200);
 
   /// Light.
-  static const FontWeight w300 = FontWeight._(2, 300);
+  static const FontWeight w300 = FontWeight(300);
 
   /// Normal / regular / plain.
-  static const FontWeight w400 = FontWeight._(3, 400);
+  static const FontWeight w400 = FontWeight(400);
 
   /// Medium.
-  static const FontWeight w500 = FontWeight._(4, 500);
+  static const FontWeight w500 = FontWeight(500);
 
   /// Semi-bold.
-  static const FontWeight w600 = FontWeight._(5, 600);
+  static const FontWeight w600 = FontWeight(600);
 
   /// Bold.
-  static const FontWeight w700 = FontWeight._(6, 700);
+  static const FontWeight w700 = FontWeight(700);
 
   /// Extra-bold.
-  static const FontWeight w800 = FontWeight._(7, 800);
+  static const FontWeight w800 = FontWeight(800);
 
   /// Black, the most thick.
-  static const FontWeight w900 = FontWeight._(8, 900);
+  static const FontWeight w900 = FontWeight(900);
 
   /// The default font weight.
   static const FontWeight normal = w400;
@@ -109,13 +110,15 @@ class FontWeight {
     w900,
   ];
 
+  @override
+  bool operator ==(Object other) {
+    if (other.runtimeType != runtimeType) {
+      return false;
+    }
+    return other is FontWeight && other.value == value;
+  }
+
   /// Linearly interpolates between two font weights.
-  ///
-  /// Rather than using fractional weights, the interpolation rounds to the
-  /// nearest weight.
-  ///
-  /// For a smoother animation of font weight, consider using
-  /// [FontVariation.weight] if the font in question supports it.
   ///
   /// If both `a` and `b` are null, then this method will return null. Otherwise,
   /// any null values for `a` or `b` are interpreted as equivalent to [normal]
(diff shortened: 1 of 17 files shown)


engine/src/flutter/lib/ui/text.dart
```diff
@@ -53,42 +53,43 @@ enum FontStyle {
 /// with the name "Roboto" and the weight [FontWeight.w500].
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
-/// See [FontVariation.weight] for details.
+/// When using these fonts, applications can specify [FontWeight] instances
+/// constructed using values other than the predefined values.
 class FontWeight {
-  const FontWeight._(this.index, this.value);
+  const FontWeight(this.value);
```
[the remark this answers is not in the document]
jason-simmons: Added the assert

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/paragraph.dart
```diff
@@ -433,26 +440,30 @@ class SkwasmTextStyle implements ui.TextStyle {
...
+      ui.FontVariation(_kWeightAxisTag, weightValue.toDouble()),
+    ];
+    Iterable<ui.FontVariation> allFontVariations;
+    if (fontVariations == null) {
+      allFontVariations = weightVariation;
+    } else if (fontVariations!.any((ui.FontVariation v) => v.axis == _kWeightAxisTag)) {
+      allFontVariations = fontVariations!;
+    } else {
+      allFontVariations = fontVariations!.followedBy(weightVariation);
     }
```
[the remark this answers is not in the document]
jason-simmons: Simplified this while also avoiding unnecessary copies of the fontVariations list.

engine/src/flutter/lib/web_ui/lib/text.dart
```diff
@@ -14,18 +14,20 @@ enum FontStyle { normal, italic }
 enum PlaceholderAlignment { baseline, aboveBaseline, belowBaseline, top, bottom, middle }
 
 class FontWeight {
-  const FontWeight._(this.index, this.value);
-  final int index;
+  const FontWeight(this.value);
```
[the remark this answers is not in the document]
jason-simmons: Added the assert

engine/src/flutter/lib/ui/text.dart
```diff
@@ -2200,10 +2221,9 @@ ByteData _encodeStrut(
     bitmask |= 1 << 7;
   }
 
-  data.setInt8(0, bitmask);
+  data.setInt32(0, bitmask, _kFakeHostEndian);
 
-  assert(byteCount <= 16);
-  assert(bitmask >> 8 == 0, 'strut bitmask overflow: $bitmask');
```
harryterkelsen, line 2214: Why remove this assert? Seems just as useful as the preceding one.

engine/src/flutter/lib/ui/text.dart
```diff
@@ -2161,18 +2182,18 @@ ByteData _encodeStrut(
     return ByteData(0);
   }
 
-  final ByteData data = ByteData(16); // Max size is 16 bytes
-  int bitmask = 0; // 8 bit mask
-  int byteCount = 1;
+  final ByteData data = ByteData(24); // Max size is 24 bytes
```
harryterkelsen, line 2196: nit: Some of these changes seem unnecessary. Why make the bitmask 4 bytes when it's still just 1 byte? Why make the fontStyle 4 bytes?

eyebrowsoffire (approved):
Skwasm changes look good!

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/text.dart
```diff
@@ -87,6 +89,12 @@ class CkParagraphStyle implements ui.ParagraphStyle {
       skTextStyle.fontStyle = toSkFontStyle(fontWeight, fontStyle);
     }
 
+    final int weightValue = fontWeight?.value ?? ui.FontWeight.normal.value;
```
harryterkelsen, line 92: We want to avoid adding `fontVariations` to the text style when we can because it requires us to malloc an array on the WASM side.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/text.dart
```diff
@@ -579,16 +587,27 @@ class CkTextStyle implements ui.TextStyle {
...
         skFontVariation.value = fontVariation.value;
         skFontVariations.add(skFontVariation);
+        if (fontVariation.axis == _kWeightAxisTag) {
+          weightAxisSet = true;
+        }
       }
-      properties.fontVariations = skFontVariations;
     }
+    if (!weightAxisSet) {
+      final int weightValue = fontWeight?.value ?? ui.FontWeight.normal.value;
```
harryterkelsen, line 604: ditto above

harryterkelsen (changes requested):

engine/src/flutter/lib/ui/text.dart
jason-simmons: Added an assert that checks if the strut bitmask fits in 32 bits

engine/src/flutter/lib/ui/text.dart
jason-simmons: The strut encoding had to be changed because FontWeight is now represented as a value that does not fit in a single byte.

The goal of the new strut encoding was to simplify it by storing all fields as 32-bit values at 32-bit aligned offsets.  This is similar to how text styles and paragraph styles are encoded.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/text.dart
jason-simmons: Unfortunately I don't know of a way to work around this.

Some variable fonts may not have the default value of the `wght` attribute configured to match the value of `FontWeight.normal`.  If the engine does not explicitly set the `wght` value in `fontVariations`, then these fonts will not render as intended at the normal weight.

engine/src/flutter/lib/ui/text.dart
```diff
@@ -1110,8 +1127,12 @@ class FontVariation {
 
   /// Variable font weight. (`wght`)
   ///
-  /// Varies the stroke thickness of the font, similar to [FontWeight] but on a
-  /// continuous axis.
+  /// Applications should avoid using this and should instead declare font
```
LongCatIsLooong, line 1133: Should this be marked as deprecated then?

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/text.dart
harryterkelsen: That makes sense. I guess we can live with it, given that it seems that browsers apparently always set axes according to CSS, it makes sense for us to set them according to Flutter text style properties (source: https://groups.google.com/g/skia-discuss/c/eFFDObvJyQ8). But let's keep an eye on text benchmarks after this lands.

engine/src/flutter/lib/ui/text.dart
harryterkelsen: Makes sense.

engine/src/flutter/lib/ui/text.dart
harryterkelsen: Thanks!

harryterkelsen (approved):
LGTM for web_ui parts

engine/src/flutter/lib/ui/text.dart
```diff
@@ -53,42 +55,47 @@ enum FontStyle {
 /// with the name "Roboto" and the weight [FontWeight.w500].
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
-/// See [FontVariation.weight] for details.
+/// When using these fonts, applications can specify [FontWeight] instances
+/// constructed using values other than the predefined values.
 class FontWeight {
-  const FontWeight._(this.index, this.value);
+  /// Create a [FontWeight] object, which can be added to a [TextStyle] to
```
LongCatIsLooong:
```suggestion
  /// Creates a [FontWeight] object, which can be added to a [TextStyle] to
```

engine/src/flutter/lib/ui/text.dart
```diff
@@ -53,42 +55,47 @@ enum FontStyle {
...
+  /// select the thickness of a font's glyphs.
+  const FontWeight(this.value)
+    : assert(value >= 1, 'Font weight must be between 1 and 1000'),
+      assert(value <= 1000, 'Font weight must be between 1 and 1000');
 
   /// The encoded integer value of this font weight.
-  final int index;
+  int get index => (value ~/ 100 - 1).clamp(0, 8);
```
LongCatIsLooong, line 71: Do we still need this? I think this should be deprecated / discouraged in the docs?

engine/src/flutter/lib/ui/text.dart
```diff
@@ -137,11 +149,16 @@ class FontWeight {
     if (a == null && b == null) {
       return null;
     }
-    return values[_lerpInt((a ?? normal).index, (b ?? normal).index, t).round().clamp(0, 8)];
+    return FontWeight(
+      _lerpInt((a ?? normal).value, (b ?? normal).value, t).round().clamp(100, 900),
```
LongCatIsLooong, line 156: Is there a reason for still clamping the weight value to the `[100, 900]` range?

engine/src/flutter/lib/ui/text.dart
LongCatIsLooong: Or do we want to keep this in case people want to use `weight` for selecting a specific type face they have in mind and `wght` for the actual weight used for rendering?

engine/src/flutter/lib/ui/text.dart
```diff
@@ -53,42 +55,47 @@ enum FontStyle {
 /// with the name "Roboto" and the weight [FontWeight.w500].
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
-/// See [FontVariation.weight] for details.
+/// When using these fonts, applications can specify [FontWeight] instances
+/// constructed using values other than the predefined values.
 class FontWeight {
```
LongCatIsLooong, line 62: I wonder if this can be an extension type now that it is just an `int` under the hood?

engine/src/flutter/lib/ui/text.dart
```diff
@@ -53,42 +55,47 @@ enum FontStyle {
 /// with the name "Roboto" and the weight [FontWeight.w500].
 ///
 /// Some modern fonts allow the weight to be adjusted in arbitrary increments.
-/// See [FontVariation.weight] for details.
+/// When using these fonts, applications can specify [FontWeight] instances
+/// constructed using values other than the predefined values.
```
LongCatIsLooong: nit: maybe cross ref `FontVariation.weight` and how it interacts with `FontWeight`?

engine/src/flutter/lib/ui/text.dart
LongCatIsLooong: Ah there are FontWeight subclasses in the wild but since we were using `index` to communicate the weight I don't think subclassing does anything.

LongCatIsLooong (approved):
LGTM with nits.

engine/src/flutter/lib/ui/text.dart
jason-simmons: `FontVariation.weight` is being kept for backward compatibility and for scenarios like the one you described where the app wants direct control over `wght` (see https://github.com/flutter/flutter/issues/148026#issuecomment-2108886092)

engine/src/flutter/lib/ui/text.dart
jason-simmons: done

engine/src/flutter/lib/ui/text.dart
jason-simmons: `index` is a public API that is being retained for backward compatibility.  Marked it as deprecated.

engine/src/flutter/lib/ui/text.dart
jason-simmons: This is also for backward compatibility.  The docs for `FontWeight.lerp` say that "The result is clamped to the range w100–w900" (https://api.flutter.dev/flutter/dart-ui/FontWeight/lerp.html)

The `wght` attribute can potentially range from 1 to 1000.  But 100-900 covers most typical use cases.

engine/src/flutter/lib/ui/text.dart
jason-simmons: Done

engine/src/flutter/lib/ui/text.dart
loic-sharma: ~~Consider adding a quick comment as it's not immediately obvious why the clamp is necessary.~~

Edit: On second thought, that's not necessary as this is commented on the API already.

ZhuJHua:
Thanks for your great work! I noticed that some variable fonts don’t strictly follow the standard weight scale. For instance, MiSans VF defines its regular weight as 330.
Here is the fvar table for MiSansVF.
截屏2026-01-21 09 32 09
As you can download MiSansVF for free [here](https://hyperos.mi.com/font/en/download/)

ZhuJHua:
Could we provide a map to preset variable axis values, instead of simply mapping fontWeight to wght?

limoyancn:
> 感谢您的出色工作！我注意到一些可变字体并没有严格遵循标准字重比例。例如，MiSans VF 将其常规字重定义为 330。 这是 MiSans VF 的 fvar 表。您可以点击[此处](https://hyperos.mi.com/font/en/download/) 免费下载 MiSans VF 。 截屏2026-01-21 09 32 09[](https://hyperos.mi.com/font/en/download/)

请问一下 xiaomi misans 这个问题解决了，一直困扰了很久

ZhuJHua:
> 请问一下 xiaomi misans 这个问题解决了，一直困扰了很久

可以写一个ttf解析，或者引用一个现成的库，解析出可变字体内的fvar表然后配置全局的theme

Lamda303:
@ZhuJHua 配置theme只对没有设置fontweight的文本生效吧，如果显示设置TextStyle fontWeight，theme不起作用的

ZhuJHua:
> 显示设置TextStyle fontWeight

用textStyle的时候最好有统一的变量，而不是每个地方单独创建TextStyle

PR #175783, opened 2025-09-22
[ Widget Preview ] Fix filter by file on Windows

### LINKED ISSUE #175524: Widget preview "filter to current file" doesn't work (Windows?)
When the "Filter by file" option is enabled on the previewer, when I click inside my file (which is where the widgets and previews are defined), the previews vanish and I get a message that there aren't any:

Image

Image

@bkonyi not sure if this might be windows-specific (because paths/URIs 😄)

bkonyi:
Windows file URIs reported by IDEs can be percent encoded. This change decodes all URIs for `TextDocument` instances to ensure they're consistent with the URIs provided by the tool.

Fixes https://github.com/flutter/flutter/issues/175524


packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/filter_by_selected_file_test.dart
```diff
@@ -11,112 +11,143 @@ import 'package:widget_preview_scaffold/src/widget_preview_rendering.dart';
...
+    );
+  });
+  testWidgets('Filter previews based on currently selected file (Windows)', (
+    tester,
+  ) async {
+    // Regression test for https://github.com/flutter/flutter/issues/175524
+    const kScript1 = 'file:///C:/script1';
+    const kScript2 = 'file:///C:/script2';
+    const kTextEditorScript1 = 'file:///C%3A/script1';
+    const kTextEditorScript2 = 'file:///C%3A/script2';
```
DanTup: This reminds me that we can also get inconsistent drive letter casing on Windows, so I think we should ensure uppercase and lowercase drive letters are treated the same (actually the entire path could be different case and - depending on the filesystem - might be case-insensitive - but in practice it's only the drive letter that causes us issues and that's all we usually try to handle elsewhere).

I don't know if the decoding could cause any issues, but if so, maybe calling `toFilePath()` (ensuring it's a file scheme) and comparing the path would be better (although it still won't handle the drive letter casing).

DanTup (approved):
lgtm, though one comment about drive letter casing which could be another potential bug that's worth fixing.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/filter_by_selected_file_test.dart
bkonyi: `toFilePath()` won't handle the decoding either, so I'm not sure doing that would make any difference here.

Do we have Dart code anywhere that you're aware of where we perform this sort of comparison?

DanTup: > `toFilePath()` won't handle the decoding either

It should do, because the file path can't have the encoded colon:

```dart
  var a = Uri.parse('file:///c%3a/foo');
  print(a); // file:///c%3A/foo
  print(a.toFilePath(windows: true)); // c:\foo
```

On Windows, you wouldn't need the `windows: true` but I'm currently on macOS and Dartpad testing this.


> Do we have Dart code anywhere that you're aware of where we perform this sort of comparison?

Not for the colon, because in both LSP and DAP we just use `toFilePath()` because both things tend to use file paths instead of URIs internally.

For drive letter casing, we have this in DAP: https://github.com/dart-lang/sdk/blob/d3281b44e9dff1167c90dde8f7a23995969d261d/pkg/dds/lib/src/dap/adapters/mixins.dart#L224

But I'm failing to find where we handle this in LSP right now (I'm pretty sure we do, but it'll likely look the same as the above, just uppercasing the drive letter after we've gone from the URI to the path).

bkonyi: Right, I misread the `toFilePath` code 🤦

I've reworked this PR to go one step further and actually use `package:path`'s `Context.equals` to compare paths returned by `Uri.toFilePath`. This included adding a new DTD service method to determine if the current platform is Windows from the web application without pulling in any external dependencies.

PTAL and let me know what you think!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/editor_service.dart
```diff
@@ -158,12 +163,20 @@ class ActiveLocationChangedEvent extends EditorEvent {
 /// The [version] is an integer corresponding to LSP's
 /// [VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedTextDocumentIdentifier)
 class TextDocument {
-  TextDocument({required this.uriAsString, required this.version});
-
-  TextDocument.fromJson(Map<String, Object?> map)
+  TextDocument({
+    required bool isWindows,
```
DanTup: I wonder if it would be simpler to just infer this from the URI? The Windows URIs will always start `file:///x:/` where x is a drive letter. It feels a little awkward that `TextDocument.fromJson()` needs to know if the platform is Windows.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
```diff
@@ -93,14 +101,21 @@ class WidgetPreviewScaffoldController {
...
               previews: group.previews
                   .where(
-                    (preview) =>
-                        preview.scriptUri == selectedSourceFile.uriAsString,
+                    (preview) => context.equals(
+                      // TODO(bkonyi): we can probably save some cycles by caching the file URI
+                      // rather than computing it on each filter.
+                      Uri.parse(
+                        preview.scriptUri,
+                      ).toFilePath(windows: isWindows),
```
DanTup: Since you have `PathContext` here, I think this could be simplified to something like: `context.fromUri(preview.scriptUri)`

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/editor_service.dart
```diff
@@ -158,12 +163,20 @@ class ActiveLocationChangedEvent extends EditorEvent {
...
+    required String uriAsString,
+    required this.version,
+  })
+    // Resolve any percent encoding
+    // See https://github.com/flutter/flutter/issues/175524.
+    : uriAsString = Uri.parse(uriAsString).toFilePath(windows: isWindows);
```
DanTup: Do we still need this if we're converting to paths for comparison later now?

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/editor_service.dart
bkonyi: I'd rather avoid inferring the platform from the URI when we have a way to get the exact platform from the tool. I'll just defer processing the URI until we actually use it.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
bkonyi: That's much nicer! Done!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/editor_service.dart
bkonyi: Removed.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
```diff
@@ -107,14 +113,21 @@ class WidgetPreviewScaffoldController {
     // ignore these updates.
     final selectedSourceFile = dtdServices.selectedSourceFile.value;
     if (selectedSourceFile != null) {
+      // Resolve any percent encoding
+      // See https://github.com/flutter/flutter/issues/175524.
+      final selectedSourceUri = context.fromUri(selectedSourceFile.uriAsString);
```
DanTup: I think this comment and variable name are a bit misleading - this is just converting the URI to a file path. I would name the variable something like `selectedSourcePath` and maybe change the comment to something like:

> Convert to a file path for comparing to avoid issues with optional encoding in URIs (https://github.com/flutter/flutter/issues/175524)

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
```diff
@@ -107,14 +113,21 @@ class WidgetPreviewScaffoldController {
...
           .map(
             (group) => WidgetPreviewGroup(
               name: group.name,
               previews: group.previews
                   .where(
-                    (preview) =>
-                        preview.scriptUri == selectedSourceFile.uriAsString,
+                    (preview) => context.equals(
+                      // TODO(bkonyi): we can probably save some cycles by caching the file URI
+                      // rather than computing it on each filter.
```
DanTup: Same here - if we were to cache anything here, it's a file path now.

DanTup (approved):
Some minor nits, but otherwise lgtm!

PR #175808, opened 2025-09-22
Update AGP/Java/Gradle comparison when using analyze --suggestions

### LINKED ISSUE #175669: Flutter analyze --suggestions gives false negative
When working on a different issue I was updating agp/gradle/java to see if an update fixed my reproduced issue and I got this message. 

```
│ Incompatible Java/Gradle versions.                                │
│ Java Version: 17.0.15, Gradle Version: 8.13
```

When the documentation indicates that those versions should be compatible https://docs.gradle.org/current/userguide/compatibility.html#java and the app compiles and runs successfully. 

found with flutter 3.35.4.

reidbaker:
Update the maxKnownAndSupported versions of Java (now 26), AGP (from 8.9.1 to 8.13), and Gradle (now 9.1). 

Add tests for new codepaths and expand existing test coverage. Some typos and changes to aid debugging included as well. 

AGP 9 was made available for preview last week https://developer.android.com/build/releases/agp-preview. 

Fixes: https://github.com/flutter/flutter/issues/175669

Commits: 
- **update max known and supported java, gradle and agp**
- **make tests in a loop and add testing for get java version for and validate java and agp**
- **move some tests to gradle_utils, make debugging easier**
- **Migrated gradle_utils tests from gradle_test to gradle_utils_test**


packages/flutter_tools/test/general.shard/android/gradle_test.dart
```diff
@@ -571,69 +570,6 @@ flutter:
     });
   });
 
-  group('gradle version', () {
```
reidbaker, line 574: These tests were moved to gradle_utils_test

reidbaker:
@gmackall While authoring this pr I believe the decision we made in https://github.com/flutter/flutter/pull/175808/files#diff-afa4fef23f35d6a99b734a7c1ec54b98750e48714a3026e2f2eccb5468ca488cR969-R977 was incorrect and that if AGP is higher than maxKnownAndSupportedAgpVersion we should assume java is valid if it is equal to or higher than the last known supported java version. 

Users with very new agp versions and older flutter_tool versions would false negative pretty regularly. Especially since java 17 is supported for many years worth of agp versions. 

If you agree I will make that change in a different pr since I think changing default behavior is riskier.

gmackall:
> @gmackall While authoring this pr I believe the decision we made in https://github.com/flutter/flutter/pull/175808/files#diff-afa4fef23f35d6a99b734a7c1ec54b98750e48714a3026e2f2eccb5468ca488cR969-R977 was incorrect and that if AGP is higher than maxKnownAndSupportedAgpVersion we should assume java is valid if it is equal to or higher than the last known supported java version.
> 
> Users with very new agp versions and older flutter_tool versions would false negative pretty regularly. Especially since java 17 is supported for many years worth of agp versions.
> 
> If you agree I will make that change in a different pr since I think changing default behavior is riskier.

I agree we should default to assuming the Java version is fine if we know nothing about the AGP version

packages/flutter_tools/lib/src/android/gradle_utils.dart
```diff
@@ -1135,19 +1173,20 @@ void exitWithNoSdkMessage() {
 //
 // The [javaMax] is exclusive in terms of supporting the noted [gradleMin],
 // whereas [javaMin] is inclusive.
+// Gradle Max is not required and when set should mean unbound.
```
camsim99:
```suggestion
// Gradle Max is not required and when unset should mean unbound.
```

camsim99 (approved):

jtmcdole:
PR Was stuck in the merge queue; removed and re-added.

PR #175810, opened 2025-09-22
Fix Voiceover traversal for OutlinedButton.icon

### LINKED ISSUE #172489: [Google3 Bug][A11Y]: Voiceover focus traversal breaks if a button's state     changes to include an icon
###Help us understand the severity of this issue

- [ ] causing severe production issues e.g. malfunctions or data loss
- [ ] blocking next binary release
- [X] blocking a client feature launch within a quarter
- [ ] nice-to-have but does not block a launch within the next quarter

### Steps to reproduce

1. Using a screenreader such as Voiceover, focus on the first button
2. Click the first button using the screenreader
3. Observe that the focus jumps forward to the next button
4. Click the second button using the screenreader
5. Observe that the focus jumps backwards to the first button

### Expected results

When clicking a button using a screenreader, the focus should remain on that button.

### Actual results

The focus jumps forward or backwards if a button's state changes to include an icon.

### Code sample

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return const MaterialApp(
      home: Scaffold(body: Center(child: ButtonIconExample())),
    );
  }
}

class ButtonIconExample extends StatefulWidget {
  const ButtonIconExample({super.key});

  @override
  State<StatefulWidget> createState() => ButtonIconExampleState();
}

class ButtonIconExampleState extends State<ButtonIconExample> {
  bool toggled1 = false;
  bool toggled2 = false;

  @override
    return Container(
      child: Card(
        color: Colors.blue,
        child: SizedBox(
          width: 500,
          height: 100,
          child: Row(
            children: [
              OutlinedButton.icon(
                label: Text('test button 1'),
                onPressed: () => setState(() => toggled1 = !toggled1),
                icon: toggled1 ? Icon(Icons.favorite) : null,
              ),
              OutlinedButton.icon(
                label: Text('test button 2'),
                onPressed: () => setState(() => toggled2 = !toggled2),
                icon: toggled2 ? Icon(Icons.favorite) : null,
              ),
            ],
          ),
        ),
      ),
    );
  }
}

```

</details>

### Screenshots or Video

<details open>

https://github.com/user-attachments/assets/26d063a3-a1a5-4b59-a900-5c9cafe5dc64

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[Paste your output here]
```

</details>

LouiseHsu:
Fixes #172489 

When an OutlinedButton.icon is initially built with a null icon and later updated to display an icon, the underlying widget implementation changes. The button switches from a standard OutlinedButton to a private _OutlinedButtonWithIcon widget.

This switch causes the button's entire semantic node to be destroyed and a new one created in its place. For screen readers like VoiceOver, this behavior is disruptive. If an accessibility service is focused on the button when its icon appears, that focus is lost because the original semantic node is discarded, leading to strange behaviour as it makes an best effort to focus on an existing node.

This PR resolves the issue by ensuring the same widget is used regardless of whether the icon is present. The logic for _OutlinedButtonWithIcon has been merged into the base class OutlinedButton, ensuring that no matter if you call OutlinedButton.icon or OutlinedButton you recieve a widget of type OutlinedButton.

Demo:

https://github.com/user-attachments/assets/e012bac9-823e-46f1-8eba-ec70e6b260a1


packages/flutter/lib/src/material/outlined_button.dart
```diff
@@ -491,12 +491,19 @@ class _OutlinedButtonWithIconChild extends StatelessWidget {
         outlinedButtonTheme.style?.iconAlignment ??
         buttonStyle?.iconAlignment ??
         IconAlignment.start;
+    final Widget? icon = this.icon;
     return Row(
```
chunhtai, line 451: Do you still need a row if there is no icon?

chunhtai (commented):
Is the intention of this pr to preserve the renderobject so that it can have stable id when switching? If so I think this approach is fine.

packages/flutter/test/material/outlined_button_test.dart
```diff
@@ -1260,6 +1260,60 @@ void main() {
     semantics.dispose();
   });
 
+  testWidgets('When an OutlinedButton gains an icon, preserves the same Semantic Node ID', (
```
chunhtai:
```suggestion
  testWidgets('When an OutlinedButton gains an icon, preserves the same SemanticsNode id', (
```

chunhtai (approved):
LGTM

dev/integration_tests/flutter_gallery/test/demo/material/buttons_demo_test.dart
```diff
@@ -35,9 +35,9 @@ void main() {
     {
       await tester.tap(find.text('OUTLINED'));
       await tester.pumpAndSettle();
-      expect(find.byType(OutlinedButton).evaluate().length, 2);
+      expect(find.byType(OutlinedButton).evaluate().length, 4);
```
LouiseHsu, line 38: Changed this test as originally OutlinedButton.icon was returning _OutlinedButtonWithIcon, and  didnt count as an OutlinedButton.

packages/flutter/lib/src/material/outlined_button.dart
```diff
@@ -396,6 +398,25 @@ class OutlinedButton extends ButtonStyleButton {
             alignment: Alignment.center,
             splashFactory: InkRipple.splashFactory,
           );
+
+    // Only apply paddings when OutlinedButton has an Icon
+    if (child is _OutlinedButtonWithIconChild && theme.useMaterial3) {
```
chunhtai: We should avoid runtimeType check like this. This compromised the composition nature of Flutter. Can the padding be applied in the _OutlinedButtonWithIconChild instead?

chunhtai (commented):
If you can't figure out a good way to fix this, you can use a MergeSemantics on top of the button subtree to force everything into a single node, and as long as the MergeSemantics doesn't change, it should persist the semanticsnode

packages/flutter/lib/src/material/outlined_button.dart
LouiseHsu: I don't think so? _OutlinedButtonWithIconChild doesnt have a method defaultStyleOf. The styling here ^ also requires context so I can't easily move it to the child

LouiseHsu:
> If you can't figure out a good way to fix this, you can use a MergeSemantics on top of the button subtree to force everything into a single node, and as long as the MergeSemantics doesn't change, it should persist the semanticsnode

Sorry, can you elaborate what you mean? My guess was that I apply MergeSemantics around the entire check for OutlinedButton.icon , like
```
return MergeSemantics(
    child: icon ? OutlinedButton(
         ...
      ) : _OutlinedButtonWithIcon(
        icon : icon
        ...
      ),
    );
```

But OutlinedButton.icon requires a type of OutlinedButton to be returned

chunhtai:
> But OutlinedButton.icon requires a type of OutlinedButton to be returned

yeah you are right, you could change the factory constructor to a static method to get around it though.

How bad is the g3 migration for the previous fix?

Another suggestion.

Convert the factory into a regular constructor, and add a private property `_getPadding` to the class(so that it is not exposed), and then the .icon constructor can set this callback to get padding if icon is not null, default constructor will just set this to a static method that always return null so that it can still be const

packages/flutter/lib/src/material/outlined_button.dart
```diff
@@ -599,32 +569,32 @@ class _OutlinedButtonDefaultsM3 extends ButtonStyle {
...
+        if (states.contains(WidgetState.focused)) {
+          return BorderSide(color: _colors.primary);
+        }
+        return BorderSide(color: _colors.outline);
+      });
 
   @override
   WidgetStateProperty<OutlinedBorder>? get shape =>
-    const MaterialStatePropertyAll<OutlinedBorder>(StadiumBorder());
+      const MaterialStatePropertyAll<OutlinedBorder>(StadiumBorder());
```
LouiseHsu: i ran dart format and now theres a bunch of these format changes 😿

packages/flutter/lib/src/material/outlined_button.dart
```diff
@@ -83,8 +83,9 @@ class OutlinedButton extends ButtonStyleButton {
     super.autofocus = false,
     super.clipBehavior,
     super.statesController,
+    Widget? icon,
```
chunhtai: looking at the code this `_icon` is just to check whether a padding is needed. The developer are still expect to build an icon in the child by themselves.

might as well let this be `bool addPaddding`.

packages/flutter/lib/src/material/outlined_button.dart
```diff
@@ -111,22 +112,7 @@ class OutlinedButton extends ButtonStyleButton {
     required Widget label,
     IconAlignment? iconAlignment,
```
chunhtai, line 112: We can consider make this a regular constructor so you don't need to pass "addPaddding" information through the default constructor.
```dart

const OutlinedButton({
 ...
}) : _addPadding = false;


const OutlinedButton.icon({
 ...
}) : child = icon != null
          ? _OutlinedButtonWithIconChild(
              iconAlignment: iconAlignment,
              label: label,
              buttonStyle: style,
              icon: icon,
            )
          : label,
      _addPadding = true;

bool _addPadding

```

packages/flutter/lib/src/material/outlined_button.dart
chunhtai: you may need to update `flutter update-packages`

chunhtai (approved):
LGTM

PR #175815, opened 2025-09-22
In Gradle Flutter task, correctly replace '\ ' with ' '.

### LINKED ISSUE #174043: Running the project does not compile the latest code version and instead, it shows an older version.
### Steps to reproduce

1. Create a new Flutter project (counter app; it doesn't matter though).
2. Run the project for the first time (Android Studio).
3. Make some changes, like adding new widgets or editing some texts.
4. Save the code so that it hot reloads. The UI is updated accordingly.
5. Close the app either by stopping the process from Android Studio or through the OS.
6. Run the project again.
7. You'll notice that the UI no longer matches the code. If it does match the code, then repeat steps 3-6.

This issue was introduced in version 3.32 and wasn't resolved since. I followed 2 issues, but the OPs didn't come back with more info. They most likely gave up and started running `flutter clean && flutter pub get` every time they re-run the app.
https://github.com/flutter/flutter/issues/169914
https://github.com/flutter/flutter/issues/171507

### Expected results

The app should always displays the latest version of the code irrespective of whether the changes were made during a previous session using hot reload or not.

### Actual results

The app shows a UI resembling an older version of the code.

### Code sample

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        //
        // TRY THIS: Try running your application with "flutter run". You'll see
        // the application has a purple toolbar. Then, without quitting the app,
        // try changing the seedColor in the colorScheme below to Colors.green
        // and then invoke "hot reload" (save your changes or press the "hot
        // reload" button in a Flutter-supported IDE, or press "r" if you used
        // the command line to start the app).
        //
        // state is not lost during the reload. To reset the state, use hot
        // restart instead.
        //
        // This works for code too, not just values: Most codae changes can be
        // tested with just a hot reload.
      ),
      home: const MyHomePage(title: 'Flutter Test page 2'),
    );
  }
}

  // how it looks.

  // always marked "final".

  final String title;

  @override
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });
  }

  @override
    //
    return Scaffold(
      appBar: AppBar(
        // TRY THIS: Try changing the color here to a specific color (to
        // Colors.amber, perhaps?) and trigger a hot reload to see the AppBar
        // change color while the other colors stay the same.
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          //
          // horizontal).
          //
          // TRY THIS: Invoke "debug painting" (choose the "Toggle Debug Paint"
          // action in the IDE, or press "p" in the console), to see the
          // wireframe for each widget.
          children: <Widget>[
            const Text('You have pushed the button this many times:'),
            Text(
              '$_counter',
            ),
          ],
        ),
      ),
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>

https://drive.google.com/file/d/1Xdq6ww54OfsrlWAANA1hhZjCNeK35q1H/view?usp=drive_link

</details>

### Logs

```console
Launching lib/main.dart on sdk gphone64 arm64 in debug mode...
Running Gradle task 'assembleDebug'...
✓ Built build/app/outputs/flutter-apk/app-debug.apk
Installing build/app/outputs/flutter-apk/app-debug.apk...
D/FlutterJNI( 6454): Beginning load of flutter...
D/FlutterJNI( 6454): flutter (null) was loaded normally!
I/flutter ( 6454): [IMPORTANT:flutter/shell/platform/android/android_context_gl_impeller.cc(104)] Using the Impeller rendering backend (OpenGLES).
Debug service listening on ws://127.0.0.1:51111/AYH7KOyLRmg=/ws
Syncing files to device sdk gphone64 arm64...
D/ProfileInstaller( 6454): Installing profile for com.example.testing_app
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.1, on macOS 14.7.1 23H222 darwin-arm64, locale en-RO) [2.1s]
    • Flutter version 3.35.1 on channel stable at /Users/mihaiborcan/flutter
    • Framework revision 20f8274939 (5 days ago), 2025-08-14 10:53:09 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0
    • Feature flags: enable-web, no-enable-linux-desktop, no-enable-macos-desktop, no-enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [3.4s]
    • Android SDK at /Users/mihaiborcan/Library/Android/sdk

[✓] Xcode - develop for iOS and macOS (Xcode 16.1) [2.9s]
    • Build 16B40
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [24ms]

[✓] Android Studio (version 2025.1) [24ms]

[✓] IntelliJ IDEA Ultimate Edition (version 2025.1.3) [22ms]
    • IntelliJ at /Applications/IntelliJ IDEA.app

[✓] Connected device (2 available) [6.3s]
    • sdk gphone64 arm64 (mobile) • emulator-5554 • android-arm64  • Android 15 (API 35) (emulator)
    • Chrome (web)                • chrome        • web-javascript • Google Chrome 139.0.7258.67
    ! Error: Browsing on the local area network for Maricica’s iPhone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.

[✓] Network resources [268ms]

• No issues found!
```

</details>

mboetger:
Gradle does not like paths with '\ ' in them. 

Fixes: https://github.com/flutter/flutter/issues/174043

## Pre-launch Checklist

- [X] I signed the [CLA].


gmackall (approved):
LGTM, I think this should be easy to test by writing another test for `getSourceFiles` in https://github.com/flutter/flutter/blob/master/packages/flutter_tools/gradle/src/test/kotlin/tasks/FlutterTaskHelperTest.kt#L106

Also discussed offline that this is a regression because the old groovy logic was using replaceAll which takes a regex
https://github.com/flutter/flutter/blob/ea121f8859e4b13e47a8f845e4586164519588bc/packages/flutter_tools/gradle/src/main/groovy/flutter.groovy#L1816
and this was mistakenly replaced with a replacement method taking a string, so the escaping was no longer needed, which is the explanation of why this was a regression

mboetger:
> his was mistakenly replaced with a replacement method taking a string, so the escaping was no longer needed, which is the explanation o

Added a test.  PTAL.

gmackall (approved):
LGTM!

jmagman (commented):
Can we also update an existing integration test that create, builds, and validates a project to instead be created in a directory with a space in it?

packages/flutter_tools/gradle/src/test/kotlin/tasks/FlutterTaskHelperTest.kt
```diff
@@ -140,6 +140,42 @@ class FlutterTaskHelperTest {
...
+        val mockFlutterTask = mockk<FlutterTask>()
+
+        every { project.files() } returns mockProjectFileCollection
+        every { project.files(any()) } returns mockProjectFileCollection
+
+        every { mockFlutterTask.intermediateDir } returns tempDir.toFile()
+        every { mockFlutterTask.getDependenciesFiles() } returns mockDependenciesFileCollection
+        val dependenciesFile =
+            tempDir
+                .resolve("${mockFlutterTask.intermediateDir}/flutter_build.d")
```
ash2moon: Should we use File.seperator instead for these?

reidbaker: +1

packages/flutter_tools/gradle/src/test/kotlin/tasks/FlutterTaskHelperTest.kt
```diff
@@ -140,6 +140,42 @@ class FlutterTaskHelperTest {
         verify { project.files("pubspec.yaml") }
     }
 
+    @Test
+    fun `getSourceFiles correctly replaces spaces`(
```
reidbaker:
```suggestion
    fun `getSourceFiles correctly replaces escaped spaces`(
```

reidbaker (approved):

gmackall (commented):
Looks like there is a test actually failing on line 33 of `FlutterTaskHelperTest`, presumably because the method it is testing doesn't actually handle windows slashes properly:
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/gradle/src/main/kotlin/tasks/FlutterTaskHelper.kt#L22

mboetger:
> y because the method it is testing doesn't

Updated to use File.separator there too.

packages/flutter_tools/gradle/src/main/kotlin/tasks/FlutterTaskHelper.kt
```diff
@@ -38,7 +38,7 @@ object FlutterTaskHelper {
             from("${flutterTask.intermediateDir}")
             if (flutterTask.buildMode == "release" || flutterTask.buildMode == "profile") {
                 flutterTask.targetPlatformValues!!.forEach { targetArch ->
-                    include("${FlutterPluginConstants.PLATFORM_ARCH_MAP[targetArch]}/app.so")
+                    include("${FlutterPluginConstants.PLATFORM_ARCH_MAP[targetArch]}${File.separator}app.so")
```
mboetger, line 41: @gmackall - wondering if this should be File.sparator or not.

gmackall: Sorry I didn't see this, I would assume all slashes should be `File.separator` unless it breaks stuff (so yes probably)

PR #175866, opened 2025-09-23
Add scene plugin lifecycle events

vashworth:
This PR forwards scene lifecycle events from the SceneDelegate to registered plugins. 

This PR introduced 3 new public APIs:
* `FlutterSceneLifeCycleDelegate` (protocol) - This is the protocol that plugins conform to
* `FlutterSceneLifeCycleProvider` (protocol) - This is the protocol that a SceneDelegate can conform to instead of `FlutterSceneDelegate`
* `FlutterPluginSceneLifeCycleDelegate` (interface) - This is the class object that a SceneDelegate using `FlutterSceneLifeCycleProvider` can use to forward scene events to

The flow of events are as followed when SceneDelegate subclasses the `FlutterSceneDelegate`:
* `FlutterSceneDelegate` receives `sceneDidBecomeActive` event from UIKit
* `FlutterSceneDelegate` forwards `sceneDidBecomeActive` to `FlutterPluginSceneLifeCycleDelegate`
* `FlutterPluginSceneLifeCycleDelegate` loops through each `FlutterEngine` it has and forwards `sceneDidBecomeActive` to the engine's `FlutterEnginePluginSceneLifeCycleDelegate`
  * `FlutterEnginePluginSceneLifeCycleDelegate` loops through each plugin that conforms to `FlutterSceneLifeCycleDelegate` it has and forwards `sceneDidBecomeActive` to the plugin
* `FlutterPluginSceneLifeCycleDelegate` forwards `sceneDidBecomeActive` to `FlutterPluginAppLifeCycleDelegate` using `sceneDidBecomeActiveFallback` 
  * `FlutterPluginAppLifeCycleDelegate` loops through all plugins and calls `applicationDidBecomeActive` if the plugin does not conform to `FlutterSceneLifeCycleDelegate`
  
When using `FlutterSceneLifeCycleProvider` instead of `FlutterSceneDelegate`, scene events are manually forwarded to `FlutterPluginSceneLifeCycleDelegate` by the iOS developer. For example:

```swift
class SceneDelegate: UIResponder, UIWindowSceneDelegate, FlutterSceneLifeCycleProvider {

  var sceneLifeCycleDelegate: FlutterPluginSceneLifeCycleDelegate = FlutterPluginSceneLifeCycleDelegate()

  func sceneDidBecomeActive(_ scene: UIScene) {
    sceneLifeCycleDelegate.sceneDidBecomeActive(scene)
  }
```

Fixes https://github.com/flutter/flutter/issues/174398.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/BUILD.gn
```diff
@@ -35,6 +35,7 @@ _flutter_framework_headers = [
   "framework/Headers/FlutterPlugin.h",
```
vashworth, line 35: All changes above this point are for the integration test

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -0,0 +1,108 @@
...
+#define FLUTTER_SHELL_PLATFORM_DARWIN_IOS_FRAMEWORK_HEADERS_FLUTTERSCENELIFECYCLE_H_
+
+#import <UIKit/UIKit.h>
+#import "FlutterMacros.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Protocol for listener of events from the UIWindowSceneDelegate, typically a FlutterPlugin.
+ */
```
LongCatIsLooong, line 27: is `FLUTTER_DARWIN_EXPORT` needed here?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -0,0 +1,108 @@
...
+@protocol FlutterSceneLifeCycleDelegate
+
+@optional
+
+#pragma mark - Connecting and disconnecting the scene
+
+- (void)flutterViewDidConnectTo:(UIScene*)scene
```
LongCatIsLooong: These methods are public so they probably need proper documentation, especially this one since it has no obvious counterpart in `UIWindowSceneDelegate`, so it could be difficult for the implementer of the callback to figure out if this is any different from https://developer.apple.com/documentation/uikit/uiscenedelegate/scene(_:willconnectto:options:), ~and the app developer needs to know when to call this, when they can't use `FlutterSceneDelegate`~ ah the `FlutterPluginSceneLifeCycleDelegate` implementation is supposed to handle this.

dev/devicelab/bin/tasks/module_uiscene_test_ios.dart
```diff
@@ -0,0 +1,447 @@
...
+Future<Directory> _createFlutterModuleApp({
+  required Directory templatesDir,
+  required Directory destinationDir,
+}) async {
+  section('Create Flutter Module');
+
+  const String moduleName = 'my_module';
+  await flutter(
+    'create',
+    options: <String>['--org', 'io.flutter.devicelab', '--template=module', moduleName],
```
stuartmorgan-g: Do flutter/flutter tests still use `io.flutter` rather than `dev.flutter`? For flutter/plugins we phased out as much use of `io.flutter` as possible.

(Doesn't really matter, just curious.)

dev/devicelab/bin/tasks/module_uiscene_test_ios.dart
```diff
@@ -0,0 +1,447 @@
...
+  recursiveCopy(xcodeProjectTemplate, xcodeProjectDir);
+
+  return xcodeProjectDir;
+}
+
+Future<void> _installPlugins({
+  required Directory appDir,
+  required Directory xcodeProjectDir,
+}) async {
+  // Poke the pubspec to make sure the tooling generates an updated Podfile.
```
stuartmorgan-g: Maybe future-proof this comment (for a SwiftPM world) by saying "generates updated native plugin build steps"?

dev/devicelab/bin/tasks/module_uiscene_test_ios.dart
```diff
@@ -0,0 +1,447 @@
...
+    required this.appDir,
+  });
+
+  final Directory templatesDir;
+  final Directory xcodeProjectDir;
+  final Directory pluginDir;
+  final Directory appDir;
+
+  late Map<String, List<FileReplacements>> scenarios = <String, List<FileReplacements>>{
+    // When both the app and the plugin has migrated to scenes,
```
stuartmorgan-g: Nit: have

dev/devicelab/bin/tasks/module_uiscene_test_ios.dart
```diff
@@ -0,0 +1,447 @@
...
+    required this.pluginDir,
+    required this.appDir,
+  });
+
+  final Directory templatesDir;
+  final Directory xcodeProjectDir;
+  final Directory pluginDir;
+  final Directory appDir;
+
+  late Map<String, List<FileReplacements>> scenarios = <String, List<FileReplacements>>{
```
stuartmorgan-g: Maybe add a doc comment here? There's a lot going on here, so a brief overview of what this is for could be helpful.

dev/integration_tests/ios_add2app_uiscene/flutterapp/pubspec-LifeCycleTest.yaml
```diff
@@ -0,0 +1,90 @@
+name: my_module
+description: "A new Flutter module project."
+publish_to: none
+
+# The following defines the version and build number for your application.
```
stuartmorgan-g: Consider stripping out all the boilerplate comments to make it clear what in this file is actually important.

dev/integration_tests/ios_add2app_uiscene/native/UITests-ApplicationEvents-AppMigrated.swift
```diff
@@ -0,0 +1,60 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import XCTest
+
+final class xcode_uikit_swiftUITests: XCTestCase {
+
+  override func setUpWithError() throws {
+    // Put setup code here. This method is called before the invocation of each test method in the class.
```
stuartmorgan-g: Remove the boilerplate comments? (Applies to all these UI test files.)

dev/integration_tests/ios_add2app_uiscene/xcode_uikit_swift/xcode_uikit_swiftUITests/xcode_uikit_swiftUITests.swift
stuartmorgan-g, line 1: This file is just unused boilerplate, right? If so, it should probably be removed to minimize potential confusion about what test code is actually real.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifecycle.h
```diff
@@ -0,0 +1,108 @@
...
+#ifndef FLUTTER_SHELL_PLATFORM_DARWIN_IOS_FRAMEWORK_HEADERS_FLUTTERSCENELIFECYCLE_H_
+#define FLUTTER_SHELL_PLATFORM_DARWIN_IOS_FRAMEWORK_HEADERS_FLUTTERSCENELIFECYCLE_H_
+
+#import <UIKit/UIKit.h>
+#import "FlutterMacros.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Protocol for listener of events from the UIWindowSceneDelegate, typically a FlutterPlugin.
```
stuartmorgan-g: It would be good to explain that almost all of these correspond exactly to UISceneDelegete callbacks, and link to https://developer.apple.com/documentation/uikit/uiscenedelegate for more information.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
stuartmorgan-g: Since this one is different, it should have its own doc comment explaining it.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifecycle.h
```diff
@@ -0,0 +1,108 @@
...
+
+- (void)windowScene:(UIWindowScene*)windowScene
+    performActionForShortcutItem:(UIApplicationShortcutItem*)shortcutItem
+               completionHandler:(void (^)(BOOL succeeded))completionHandler;
+
+@end
+
+/**
+ * Forwards `UIWindowSceneDelegate` callbacks to the `FlutterEngine`'s
+ * `FlutterPluginSceneLifeCycleDelegate` to then be forwarded to registered plugins.
```
stuartmorgan-g: Shouldn't this be `FlutterEnginePluginSceneLifeCycleDelegate`? Although explaining how exactly it relays things to plugins doesn't seem like information clients would need, so if we want to explain the implementation I would move that to the implementation file since this is a client-facing header.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1537,9 +1544,20 @@ - (void)addMethodCallDelegate:(NSObject<FlutterPlugin>*)delegate
...
-  id<UIApplicationDelegate> appDelegate = [[UIApplication sharedApplication] delegate];
+- (void)addApplicationDelegate:(NSObject<FlutterPlugin>*)delegate {
+  // If the plugin conforms to FlutterSceneLifeCycleDelegate, add it to the engine.
+  if ([delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleDelegate)]) {
+    NSObject<FlutterSceneLifeCycleDelegate>* lifecycleDelegate =
+        (NSObject<FlutterSceneLifeCycleDelegate>*)delegate;
+    [_flutterEngine addSceneLifeCycleDelegate:lifecycleDelegate];
+  } else {
+    // TODO(vashworth): If the plugin doesn't conform to the FlutterSceneLifeCycleDelegate,
+    // print a warning pointing to documentation.
```
stuartmorgan-g: If this isn't going to be in this PR, it should have a tracking issue.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -89,6 +90,30 @@ - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector {
...
+- (BOOL)shouldHandleNotification {
+  // When UIScene lifecycle is being used, some application lifecycle events are not call by UIKit.
+  // However, the notifications are still sent. When a Flutter app has been migrated to UIScene,
+  // Flutter should not use the notifications to forward application events to plugins since they
+  // are not expected to be called.
+  // See https://flutter.dev/go/ios-ui-scene-lifecycle-migration?tab=t.0#heading=h.eq8gyd4ds50u
+  if (FlutterSharedApplication.hasSceneDelegate) {
+    return NO;
+  }
+  return YES;
```
stuartmorgan-g: `return !FlutterSharedApplication.hasSceneDelegate`?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -89,6 +90,30 @@ - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector {
...
+- (BOOL)unnecessaryFallbackFor:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
+  // The fallback is unnecessary if the plugin conforms to FlutterSceneLifeCycleDelegate.
+  // This means that the plugin has migrated to scene lifecycle events and shouldn't require
+  // application events. However, the plugin may still have the application event implemented to
+  // maintain compatibility with un-migrated apps, which is why the fallback should be checked
+  // before checking that the delegate responds to the selector.
+  if ([delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleDelegate)]) {
+    return YES;
+  }
+  return NO;
```
stuartmorgan-g: As above, consider returning the condition directly instead of using an if.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -331,6 +450,32 @@ - (BOOL)application:(UIApplication*)application
   return NO;
 }
 
+static NSDictionary<UIApplicationOpenURLOptionsKey, id>* ConvertOptions(
```
stuartmorgan-g, line 444: This could use a comment explaining what it converts them *to*, since the `id` part of the return value doesn't give any context clues.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -331,6 +450,32 @@ - (BOOL)application:(UIApplication*)application
...
+    UISceneOpenURLOptions* options) {
+  if (@available(iOS 14.5, *)) {
+    return @{
+      UIApplicationOpenURLOptionsSourceApplicationKey : options.sourceApplication
+          ? options.sourceApplication
+          : [NSNull null],
+      UIApplicationOpenURLOptionsAnnotationKey : options.annotation ? options.annotation
+                                                                    : [NSNull null],
```
stuartmorgan-g: Consider using the Elvis operator for all of these instead of duplicating the value.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -89,6 +90,30 @@ - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector {
...
+  // Flutter should not use the notifications to forward application events to plugins since they
+  // are not expected to be called.
+  // See https://flutter.dev/go/ios-ui-scene-lifecycle-migration?tab=t.0#heading=h.eq8gyd4ds50u
+  if (FlutterSharedApplication.hasSceneDelegate) {
+    return NO;
+  }
+  return YES;
+}
+
+- (BOOL)unnecessaryFallbackFor:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
```
stuartmorgan-g: I found this naming very unintuitive at the call sites; what about reversing it to `shouldCallFallbackForDelegate:`, and using a `!` at the call sites? (Similar to the argument in https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-double-negatives-in-apis)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -162,17 +206,36 @@ - (void)handleDidEnterBackground:(NSNotification*)notification
...
+  [self applicationWillEnterForeground:application isFallbackForScene:YES];
+}
+
+- (void)applicationWillEnterForeground:(UIApplication*)application
+                    isFallbackForScene:(BOOL)isFallback {
   for (NSObject<FlutterApplicationLifeCycleDelegate>* delegate in _delegates) {
     if (!delegate) {
       continue;
     }
+    if (isFallback && [self unnecessaryFallbackFor:delegate]) {
```
stuartmorgan-g: For all of these, consider folding the two `continue` conditions together instead of having the statement twice (e.g., `if (!delegate || (isFallback && [self unnecessaryFallbackFor:delegate])) { continue; }`).

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -0,0 +1,108 @@
...
+               completionHandler:(void (^)(BOOL succeeded))completionHandler;
+
+@end
+
+/**
+ * Forwards `UIWindowSceneDelegate` callbacks to the `FlutterEngine`'s
+ * `FlutterPluginSceneLifeCycleDelegate` to then be forwarded to registered plugins.
+ */
+FLUTTER_DARWIN_EXPORT
+@interface FlutterPluginSceneLifeCycleDelegate : NSObject
```
LongCatIsLooong, line 108: Question: looks like this is a strict subset of `UIWindowSceneDelegate`. So what are the reasons we have to introduce methods to this interface explicitly, instead of conforming to `UIWindowSceneDelegate`?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
```diff
@@ -72,18 +81,255 @@ - (void)updateEnginesInScene:(UIScene*)scene {
...
+
+- (void)scene:(UIScene*)scene
+    willConnectToSession:(UISceneSession*)session
+                 options:(UISceneConnectionOptions*)connectionOptions {
+  self.connectionOptions = connectionOptions;
+}
+
+- (void)sceneDidDisconnect:(UIScene*)scene {
+  [self updateEnginesInScene:scene];
+  for (FlutterEngine* engine in [_engines allObjects]) {
```
stuartmorgan-g: Nit: `allObjects` is a property, so `_engines.allObjects` is the preferred syntax.

(Applies everywhere below.)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifecycle.mm
```diff
@@ -72,18 +81,255 @@ - (void)updateEnginesInScene:(UIScene*)scene {
...
+    if (!engine) {
+      continue;
+    }
```
stuartmorgan-g: Here and everywhere below: this isn't doing anything useful, because `(nil).sceneLifeCycleDelegate` is `nil`, and `[nil anything]` is a no-op.

stuartmorgan-g (approved):
All my comments are minor, and almost all optional, so I don't think this needs another review from me, but let me know if you want me to take another look at anything you change.

The integration testing here is really, really nice!

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1537,9 +1544,20 @@ - (void)addMethodCallDelegate:(NSObject<FlutterPlugin>*)delegate
   }];
 }
 
-- (void)addApplicationDelegate:(NSObject<FlutterPlugin>*)delegate
-    NS_EXTENSION_UNAVAILABLE_IOS("Disallowed in plugins used in app extensions") {
-  id<UIApplicationDelegate> appDelegate = [[UIApplication sharedApplication] delegate];
+- (void)addApplicationDelegate:(NSObject<FlutterPlugin>*)delegate {
+  // If the plugin conforms to FlutterSceneLifeCycleDelegate, add it to the engine.
+  if ([delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleDelegate)]) {
```
LongCatIsLooong: We're expecting `FlutterPlugin` to conform to `FlutterSceneLifeCycleDelegate`, but it doesn't seem to be documented or implied in the interface (also I think even with documentation that says `FlutterPlugin` also needs to conform to `FlutterSceneLifecycleDelegate`, it will be a bit difficult to discover).

dev/devicelab/bin/tasks/module_uiscene_test_ios.dart
vashworth: Yes, we do. I'll swap it. Do you know why we replaced it?

stuartmorgan-g: No clue; I assume it was a marketing decision. (We still have flutter.io, but it redirects to flutter.dev).

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -0,0 +1,108 @@
...
+
+#import <UIKit/UIKit.h>
+#import "FlutterMacros.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Protocol for listener of events from the UIWindowSceneDelegate, typically a FlutterPlugin.
+ */
+@protocol FlutterSceneLifeCycleDelegate
```
LongCatIsLooong: nit: looks like we use "lifecycle" as a single word in most places? e.g., 

1: https://main-api.flutter.dev/flutter/dart-ui/AppLifecycleState.html,
2: https://api.flutter.dev/macos-embedder/category_flutter_app_delegate_07_08.html#a926aab76a90dbc4e9fba5033ac694183

The only place with this spelling I found is here: https://api.flutter.dev/ios-embedder/category_flutter_app_delegate_07_08.html#ab86a53588ed62c71760a9c5d52da18ff

It's already inconsistent but I suppose we should stick with the most prevalent spelling?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
vashworth: TIL!

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -89,6 +90,30 @@ - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector {
   return NO;
 }
 
+- (BOOL)shouldHandleNotification {
+  // When UIScene lifecycle is being used, some application lifecycle events are not call by UIKit.
+  // However, the notifications are still sent. When a Flutter app has been migrated to UIScene,
+  // Flutter should not use the notifications to forward application events to plugins since they
+  // are not expected to be called.
+  // See https://flutter.dev/go/ios-ui-scene-lifecycle-migration?tab=t.0#heading=h.eq8gyd4ds50u
+  if (FlutterSharedApplication.hasSceneDelegate) {
```
LongCatIsLooong: uber nit: `return !FlutterSharedApplication.hasSceneDelegate`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: nit: `isFallbackNeccessaryFor:` / `shouldUseFallbackFor:` for readability maybe?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
LongCatIsLooong: Would it be possible to add a separate method to the registrar for UISceneDelegate registration instead of reusing the one for UIAppDelegate?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -89,6 +90,30 @@ - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector {
...
+    return NO;
+  }
+  return YES;
+}
+
+- (BOOL)unnecessaryFallbackFor:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
+  // The fallback is unnecessary if the plugin conforms to FlutterSceneLifeCycleDelegate.
+  // This means that the plugin has migrated to scene lifecycle events and shouldn't require
+  // application events. However, the plugin may still have the application event implemented to
+  // maintain compatibility with un-migrated apps, which is why the fallback should be checked
```
LongCatIsLooong, line 106: It's pretty sad that now plugin authors will have to maintain two different implementations for migrated and unmigrated apps. I'm curious whether UIKit also posts a `UISceneDidEnterBackgroundNotification` for unmigrated apps so plugin authors don't have worry about unmigrated apps?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: For my own understanding: so the method above indicates whether the app is migrated and this one tells us whether a specific plugin is migrated (if that's true maybe `isPluginMigratedToScene` or some such would be a bit easer to understand for future readers especially at the call sites, and same for `shouldHandleNotification`)?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth, line 1: @stuartmorgan-g Can you re-review just the comments I (with Gemini 😝) added to this file?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -72,18 +81,255 @@ - (void)updateEnginesInScene:(UIScene*)scene {
...
 }
+
+/**
+ * Makes a best effort to get the FlutterPluginAppLifeCycleDelegate from the AppDelegate if
+ * available. It may not be available if embedded in an iOS app extension or the AppDelegate doesn't
+ * subclass FlutterAppDelegate.
+ */
+- (FlutterPluginAppLifeCycleDelegate*)applicationLifeCycleDelegate {
+  id appDelegate = FlutterSharedApplication.application.delegate;
+  if ([appDelegate respondsToSelector:@selector(lifeCycleDelegate)]) {
```
LongCatIsLooong, line 104: Is it going to crash if the `delegate` responds to `lifeCycleDelegate` but the method actually returns a different type (by sheer coincidence)? IIRC `@selector(lifeCycleDelegate)` is just a string, it doesn't encode any type information? So the static return type of `[appDelegate lifeCycleDelegate]` below is actually `id` instead of `FlutterPluginAppLifeCycleDelegate*`?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneDelegateTest.m
```diff
@@ -72,45 +73,216 @@ - (void)testMoveRootViewControllerWhenNoWindow {
...
-  OCMVerify([(FlutterPluginAppLifeCycleDelegate*)mockLifecycleDelegate application:[OCMArg any]
-                                                      performActionForShortcutItem:[OCMArg any]
-                                                                 completionHandler:[OCMArg any]]);
+  OCMVerify(times(1), [mockLifecycleDelegate windowScene:scene
+                            performActionForShortcutItem:shortcutItem
+                                       completionHandler:[OCMArg any]]);
 }
 
-- (void)testOpenURL {
+// TODO
```
LongCatIsLooong: Remove the commented-out tests if this is meant to be implemented in a different PR?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: Hmm I tried that in a Flutter app and `UIScene.willEnterForegroundNotification` is posted even when the app is unmigrated. And the notification gives you the `UIWindowScene`:
image

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Good catch! Fixed.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth: I added a comment!

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth: So there are already three public APIs with the 2-word spelling: `FlutterApplicationLifeCycleDelegate`, `FlutterAppLifeCycleProvider`, `FlutterPluginAppLifeCycleDelegate`. And the APIs added here are like the scene-equivalents to those, which is why I went this this spelling in the public headers. It also bums me out that it's different all over the place 🙈

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth: We don't implement all of the properties/methods of `UIWindowSceneDelegate`. 

For example, we don't currently implement [- windowScene:didUpdateEffectiveGeometry:](https://developer.apple.com/documentation/uikit/uiwindowscenedelegate/windowscene(_:didupdateeffectivegeometry:)?language=objc)

So if `FlutterPluginSceneLifeCycleDelegate` conformed to `UIWindowSceneDelegate` and a developer used `- windowScene:didUpdateEffectiveGeometry:`, it would crash

```swift
class SceneDelegate: UIResponder, UIWindowSceneDelegate, FlutterSceneLifeCycleProvider {

  var sceneLifeCycleDelegate: FlutterPluginSceneLifeCycleDelegate = FlutterPluginSceneLifeCycleDelegate()

  func windowScene(_ windowScene: UIWindowScene, didUpdateEffectiveGeometry previousEffectiveGeometry: UIWindowScene.Geometry) {
    sceneLifeCycleDelegate.windowScene(windowScene, didUpdateEffectiveGeometry: previousEffectiveGeometry) // this would compile but crash at runtime
  }
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: I updated the doc comment for `FlutterSceneLifeCycleDelegate`. I also plan to write website documentation. We could definitely separate this into 2 methods. Do you think that would be better?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
vashworth: There are scene notifications for like the [basic events](https://developer.apple.com/documentation/uikit/uiscene/didenterbackgroundnotification?language=objc) like active, disconnect, foreground, background, etc. However, there are not notifications for events like [scene: openURLContexts](https://developer.apple.com/documentation/uikit/uiscenedelegate/scene(_:openurlcontexts:)?language=objc), [scene:continueUserActivity](https://developer.apple.com/documentation/uikit/uiscenedelegate/scene(_:continue:)?language=objc), etc. I personally think it's be more confusing if plugin authors only need to support both for specific ones rather than all of them.

Perhaps this is something we can come back to if we have time?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
vashworth: Oh good idea, I like that much better!

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth: tbh, I don't really understand what does and doesn't need `FLUTTER_DARWIN_EXPORT`. My integration tests worked without it.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifecycle.h
```diff
@@ -0,0 +1,137 @@
...
+#import "FlutterMacros.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * A protocol for delegates that handle `UISceneDelegate` and `UIWindowSceneDelegate` life-cycle
+ * events.
+ *
+ * This protocol provides a way for Flutter plugins to observe and react to scene-based life-cycle
+ * events. To use this, a `FlutterPlugin` can conform to this protocol and implement the desired
```
stuartmorgan-g: Let's get rid of "To use this, a `FlutterPlugin` can conform to this protocol and implement the desired methods." (I have more faith than Gemini in plugin developers' understanding of the basic concept of what a `@protocol` is.)

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
stuartmorgan-g: LGTM with one change.

stuartmorgan-g (approved):

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
LongCatIsLooong: Ah ok the optional methods on the protocol were only supposed to be called by UIKit now plugin authors will have to call them and handle the optional-ness.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
LongCatIsLooong: I'm leaning towards having a separate method for scene event registration since people don't always read docs (and the existing method is called `addApplicationDelegate:`).

LongCatIsLooong (approved):

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: All unmigrated apps are single-scene apps (I think?) so I suppose we can get the scene from `UIApplication.connectedScenes` (not sure what's supposed to happen in shared extensions tho). Yeah we can come back to this later.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: Sounds good to me, I was on the fence about it. I added it and updated the integration test

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -59,15 +59,30 @@ API_AVAILABLE(ios(13.0))
 
 #pragma mark - Opening URLs
 
-- (void)scene:(UIScene*)scene openURLContexts:(NSSet<UIOpenURLContext*>*)URLContexts;
+/**
+ * Called if this has been registered for `UIWindowScene` callbacks.
+ *
+ * @return `YES` if this handles the request.
```
LongCatIsLooong: Hmm what if my plugin only handles a subset of the incoming `URLContexts`? I suppose the BOOL indicates any URL instead of all URLs?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1537,16 +1544,29 @@ - (void)addMethodCallDelegate:(NSObject<FlutterPlugin>*)delegate
...
+- (void)addApplicationDelegate:(NSObject<FlutterPlugin>*)delegate {
+  id<UIApplicationDelegate> appDelegate = FlutterSharedApplication.application.delegate;
   if ([appDelegate conformsToProtocol:@protocol(FlutterAppLifeCycleProvider)]) {
     id<FlutterAppLifeCycleProvider> lifeCycleProvider =
         (id<FlutterAppLifeCycleProvider>)appDelegate;
     [lifeCycleProvider addApplicationLifeCycleDelegate:delegate];
   }
 }
 
+- (void)addSceneDelegate:(NSObject<FlutterPlugin>*)delegate {
```
LongCatIsLooong: nit: `NSObject<FlutterSceneLifeCycleDelegate>*` or `NSObject<FlutterPlugin,  FlutterSceneLifeCycleDelegate>*` to avoid the runtime conformance check?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -318,11 +400,28 @@ - (void)userNotificationCenter:(UNUserNotificationCenter*)center
 - (BOOL)application:(UIApplication*)application
             openURL:(NSURL*)url
             options:(NSDictionary<UIApplicationOpenURLOptionsKey, id>*)options {
+  return [self application:application openURL:url options:options isFallbackForScene:NO];
+}
+
+- (BOOL)sceneFallbackOpenURLContexts:(NSSet<UIOpenURLContext*>*)URLContexts {
+  for (UIOpenURLContext* context in URLContexts) {
+    return [self application:FlutterSharedApplication.application
```
LongCatIsLooong: I think we can't return early here? Or some URLs are going to be left unprocessed.

LongCatIsLooong: But I don't understand what we're supposed to return here, does the boolean flag mean all incoming URLs are processed or any URL?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -166,37 +166,70 @@ - (void)sceneDidEnterBackground:(UIScene*)scene {
...
   [self updateEnginesInScene:scene];
+  BOOL consumedByPlugin = NO;
   for (FlutterEngine* engine in _engines.allObjects) {
-    [engine.sceneLifeCycleDelegate scene:scene openURLContexts:URLContexts];
+    BOOL result = [engine.sceneLifeCycleDelegate scene:scene openURLContexts:URLContexts];
+    if (result) {
+      consumedByPlugin = YES;
+    }
+  }
+  if (!consumedByPlugin) {
```
LongCatIsLooong, line 199: If I'm understanding correctly this is an example of converting a scene event to an app event, for unmigrated plugins. But it looks like the implementation is biased towards migrated plugins, as in if any of the migrated plugins returned YES the fallback won't happen (but all migrated plugins will get a chance to handle the event no matter what other plugins responded to the openURL request?

LongCatIsLooong (commented):
LGTM in general modulo comments regarding skipping incoming URLs. 

Additionally the `applicationLifeCycleDelegate` implementation where we have to guess to get access to the `FlutterPluginAppDelegateLifeCycle` instance to send the fallback events seems to be making too much assumptions but we can address this later.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
vashworth: Updated comment to say any. It's not clear in what situation UIKit would send more than one url, but I looked at other usages in g3 and Github and people are just using the first that succeeds.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: Oh good idea! Done.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
vashworth: Oops this was my bad, thank you for being such a great reviewer and catching 🙏 Yeah the bool means any url was processed.

I updated it. It now returns whichever succeeds first.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Basically only one plugin per engine should get to consume the event.

> if any of the migrated plugins returned YES the fallback won't happen

Correct

> but all migrated plugins will get a chance to handle the event no matter what other plugins responded to the openURL request

Nope, each *FlutterEngine* will get a change to handle the event with its plugins, but only one plugin per engine can handle the event. See below:

https://github.com/flutter/flutter/blob/a640ddb6c18f2fcb395ad6d4c43f07b42e247443/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm#L316-L326

vashworth: This is how it's handled for applications, which is what I'm mimicking:

https://github.com/flutter/flutter/blob/f3ea7a1d201034c7975d621afb1f9e4d7b397c78/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm#L318-L332

vashworth:
@stuartmorgan-g Can you review my latest [change](https://github.com/flutter/flutter/pull/175866/commits/5667fbd9701ad80a51a7748f2349c2bcdaa1dc62)? As I was writing integration tests for some upcoming PRs I realized a flaw in the `flutterViewDidConnectTo:options:` approach (which passed the scene's connection options from when the scene launched to the plugin once the FlutterView connected) - when a FlutterView is created later, such as on button click in a native app, it should not use the connection options; however, plugins don't know if the FlutterView was created on launch or not so it'll always use the connection options.

To fix this, I implemented a solution which allows us to connect the FlutterEngine to the scene when the scene connects by using the `UISceneWillConnectNotification` within the FlutterEngine and then sending the actual `scene:willConnectToSession:options` event to plugins. This solution only works for single-scene apps, though. Multi-scene apps will still connect later when the FlutterView is added to the view hierarchy and therefore will still get life cycle events except for the `scene:willConnectToSession:options` event. I think supporting this for multi-scene is something we can come back to later on.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -114,10 +112,33 @@ - (FlutterPluginAppLifeCycleDelegate*)applicationLifeCycleDelegate {
 
 #pragma mark - Connecting and disconnecting the scene
 
-- (void)scene:(UIScene*)scene
+- (void)engine:(FlutterEngine*)engine receivedConnectNotificationFor:(UIScene*)scene {
+  // Connection options may be nil if the notification was received before the
+  // `scene:willConnectToSession:options:` event. In which case, we can wait for the actual event.
+  [self addFlutterEngine:engine];
+  if (self.connectionOptions != nil) {
+    [self scene:scene willConnectToSession:scene.session options:self.connectionOptions];
```
stuartmorgan-g, line 120: Do we need to nil the options back out in case we get another scene connection later, so we don't accidentally re-use them?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -258,13 +279,18 @@ - (void)addDelegate:(NSObject<FlutterSceneLifeCycleDelegate>*)delegate {
...
-- (void)flutterViewDidConnectTo:(UIScene*)scene
-                        options:(UISceneConnectionOptions*)connectionOptions {
+- (BOOL)scene:(UIScene*)scene
+    willConnectToSession:(UISceneSession*)session
+                 options:(UISceneConnectionOptions*)connectionOptions {
   for (NSObject<FlutterSceneLifeCycleDelegate>* delegate in _delegates.allObjects) {
     if ([delegate respondsToSelector:_cmd]) {
-      [delegate flutterViewDidConnectTo:scene options:connectionOptions];
+      if ([delegate scene:scene willConnectToSession:session options:connectionOptions]) {
+        // Only allow one plugin to process this event.
```
stuartmorgan-g, line 293: Don't we need a secondary event for any remaining plugins (or maybe nullable connection options)? Even if plugins can't handle the connection options, it seems like they could need to know a scene has connected for state tracking.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Each scene will have it own delegate and therefore it's own `FlutterPluginSceneLifeCycleDelegate`, therefore we can expect the connectionOptions to only be used in that specific scene.

Plus, if there are multiple FlutterEngine's in a scene, if we null it out, they can't use the options

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Done, I went with nullable connection options and updated header doc comment

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -0,0 +1,390 @@
...
+- (BOOL)scene:(UIScene*)scene
+    willConnectToSession:(UISceneSession*)session
+                 options:(UISceneConnectionOptions*)connectionOptions {
+  BOOL consumedByPlugin = NO;
+  for (NSObject<FlutterSceneLifeCycleDelegate>* delegate in _delegates.allObjects) {
+    if ([delegate respondsToSelector:_cmd]) {
+      // If this event has already been consumed by a plugin, send the event with nil options.
+      if (consumedByPlugin) {
+        [delegate scene:scene willConnectToSession:session options:nil];
+        continue;
```
stuartmorgan-g, line 291: Nit: the `continue` isn't doing anything here.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -0,0 +1,390 @@
...
+      } else if ([delegate scene:scene willConnectToSession:session options:connectionOptions]) {
+        // Only allow one plugin to process this event.
+        consumedByPlugin = YES;
+      }
+    }
```
stuartmorgan-g, line 296: Optional: it might be easier to understand this as just:

```objc
// Only allow one plugin to process the connection options.
if ([delegate scene:scene willConnectToSession:session options:(consumedByPlugin ? nil : connectionOptions)]) {
  consumedByPlugin = YES;
}
```

stuartmorgan-g (approved):
Minor nits, but otherwise this seems like a reasonable approach (with the usual caveat that I'm definitely not a UIScene expert).

vashworth:
> Minor nits

I'll fix in my next PR so I don't have to wait for testing again lol

PR #175900, opened 2025-09-24
Exit with code 1 when calling `flutter build` without arguments

### LINKED ISSUE #5797: `flutter build` is returning zero, should return not zero ?
`flutter build` doesn't do anything other than print the help. We might not want to return zero there. Instead, I'd expect some non-zero code.

For example, this will run the second command, which I didn't expect:

`flutter build && flutter install`

The second command should not run because the first was not successful, however the first command returned zero.

Mr-Pepe:
Resolves #5797

## Pre-launch Checklist

- [x] I signed the [CLA].

A whole bunch of tool tests failed locally even before the change, so I'll have to see CI results to know if something actually broke due to the change.

### DIFF
--- a/packages/flutter_tools/lib/src/runner/flutter_command_runner.dart
+++ b/packages/flutter_tools/lib/src/runner/flutter_command_runner.dart
@@ -12,6 +12,7 @@ import '../artifacts.dart';
 import '../base/common.dart';
 import '../base/context.dart';
 import '../base/file_system.dart';
+import '../base/process.dart';
 import '../base/terminal.dart';
 import '../base/utils.dart';
 import '../cache.dart';
@@ -286,12 +287,15 @@ class FlutterCommandRunner extends CommandRunner<void> {
 
   @override
   Future<void> run(Iterable<String> args) {
+    var exitWithCodeOne = false;
+
     // Have invocations of 'build', 'custom-devices', and 'pub' print out
     // their sub-commands.
     // TODO(ianh): Move this to the Build command itself somehow.
     if (args.length == 1) {
       if (args.first == 'build') {
         args = <String>['build', '-h'];
+        exitWithCodeOne = true;
       } else if (args.first == 'custom-devices') {
         args = <String>['custom-devices', '-h'];
       } else if (args.first == 'pub') {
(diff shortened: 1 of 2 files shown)


Mr-Pepe:
No idea why Google testing is hanging but otherwise looking good from my side

Mr-Pepe:
@bkonyi You labeled the referenced issue as still valid. Are you interested in reviewing this PR?

packages/flutter_tools/lib/src/runner/flutter_command_runner.dart
```diff
@@ -300,7 +303,12 @@ class FlutterCommandRunner extends CommandRunner<void> {
     }
 
     _machineFlagPresentInAnyCliArg = args.contains('--${FlutterGlobalOptions.kMachineFlag}');
-    return super.run(args);
+    return super.run(args).then((_) {
+      if (exitWithCodeOne) {
+        // No need to print anything because the help was already printed.
+        throwToolExit('', exitCode: 1);
```
bkonyi: Can we just call `exitWithHooks(1, shutdownHooks: ...)` instead? I'm pretty sure this will result in `Error: ` being printed.

Mr-Pepe: I don't see `Error: ` being printed, only the help:

```shell
❯ ../../bin/flutter build
Building flutter tool...
Resolving dependencies... 
Downloading packages... 
Got dependencies.
Build an executable app or install bundle.

Global options:
-h, --help                  Print this usage information.
-v, --verbose               Noisy logging, including all shell commands executed.
                            If used with "--help", shows hidden options. If used with "flutter doctor", shows
                            additional diagnostic information. (Use "-vv" to force verbose logging in those
                            cases.)
-d, --device-id             Target device id or name (prefixes allowed).
    --version               Reports the version of this tool.
    --enable-analytics      Enable telemetry reporting each time a flutter or dart command runs.
    --disable-analytics     Disable telemetry reporting each time a flutter or dart command runs, until it is
                            re-enabled.
    --suppress-analytics    Suppress analytics reporting for the current CLI invocation.

Usage: flutter build <subcommand> [arguments]
-h, --help    Print this usage information.

Available subcommands:
  aar         Build a repository containing an AAR and a POM file.
  apk         Build an Android APK file from your app.
  appbundle   Build an Android App Bundle file from your app.
  bundle      Build the Flutter assets directory from your app.
  linux       Build a Linux desktop application.
  web         Build a web application bundle.

Run "flutter help" to see global options.

❯ echo $status
1
```

bkonyi: Yeah, I still think this would be better if we just call `exitWithHooks(1, shutdownHooks: ...)` instead. `throwToolExit` is meant for situations where we provide actionable feedback to users, but we're just using it to get to the root exception handler here to call `exitWithHooks(...)`. We may as well just do it here.

Mr-Pepe: Done :+1:

bkonyi (approved):
Looks good to me! We'll just need one more reviewer to land this (@chingjun can you please take a look? 😄)

chingjun (approved):

PR #175970, opened 2025-09-24
Use module names that allow for import cycles

nshahan:
Makes breakpoints work when set in files that are part of a library cycle and compiled into a single strongly connected component, even when the original source files are not in the same directory.

Issue: https://github.com/dart-lang/webdev/issues/1692

### DIFF
--- a/packages/flutter_tools/lib/src/isolated/web_asset_server.dart
+++ b/packages/flutter_tools/lib/src/isolated/web_asset_server.dart
@@ -329,7 +329,7 @@ class WebAssetServer implements AssetReader {
             ? FrontendServerDdcLibraryBundleStrategyProvider(
                 ReloadConfiguration.none,
                 server,
-                PackageUriMapper(packageConfig),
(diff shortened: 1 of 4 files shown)


packages/flutter_tools/lib/src/isolated/web_asset_server.dart
```diff
@@ -676,9 +676,13 @@ _flutter.buildConfig = ${jsonEncode(buildConfig)};
     // The file might have been a package file which is signaled by a
     // `/packages/<package>/<path>` request.
     if (segments.first == 'packages') {
-      final Uri? filePath = _packages.resolve(
-        Uri(scheme: 'package', pathSegments: segments.skip(1)),
-      );
+      Iterable<String> pathSegments = segments.skip(1);
+      final String packageName = segments[1];
+      if (segments[2] == 'lib' &&
+          _packages.packages.any((package) => package.name == packageName)) {
```
bkonyi, line 682: Can we do `_packages[packageName] != null` instead?

PR #176014, opened 2025-09-25
feat: Add isTextModified getter to TextEditingController

Musaddiq625:
- Introduced `isTextModified` getter to check if the text has been modified from its original value
- Added comprehensive test coverage for the new functionality
- Updated related test cases to use the new getter

## Use Cases
This feature could be useful in many cases:
- Enable/disable "Save" or "Submit" buttons based on whether the content has been modified
- Track if there are changes to undo/redo
- Optimize network calls by only syncing modified content

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -219,6 +219,17 @@ class _RenderCompositionCallback extends RenderProxyBox {
 ///    controlled with a [TextEditingController].
 ///  * Learn how to use a [TextEditingController] in one of our [cookbook recipes](https://docs.flutter.dev/cookbook/forms/text-field-changes#2-use-a-texteditingcontroller).
 class TextEditingController extends ValueNotifier<TextEditingValue> {
+  /// The original text value when this controller was created.
+  ///
+  /// This field is used to track if the text has been modified from its original
+  /// value. It remains unchanged when:
+  /// - [clear] is called
+  /// - [value] is updated directly
+  /// - [text] is updated
+  ///
+  /// This field is automatically cleared when the controller is disposed.
+  String? _originalText;
+
   /// Creates a controller for an editable text field, with no initial selection.
   ///
   /// This constructor treats a null [text] argument as if it were the empty
@@ -242,7 +253,10 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
   /// ** See code in examples/api/lib/widgets/editable_text/text_editing_controller.1.dart **
   /// {@end-tool}
   TextEditingController({String? text})
-    : super(text == null ? TextEditingValue.empty : TextEditingValue(text: text));
+    : _originalText = text ?? '',
+      super(
+        text == null ? TextEditingValue.empty : TextEditingValue(text: text),
+      );
 
   /// Creates a controller for an editable text field from an initial [TextEditingValue].
   ///
@@ -250,7 +264,9 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
   /// [TextEditingValue.empty].
   TextEditingController.fromValue(TextEditingValue? value)
     : assert(
-        value == null || !value.composing.isValid || value.isComposingRangeValid,
+        value == null ||
+            !value.composing.isValid ||
+            value.isComposingRangeValid,
         'New TextEditingValue $value has an invalid non-empty composing range '
         '${value.composing}. It is recommended to use a valid composing range, '
         'even for readonly text fields.',
@@ -259,6 +275,9 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
 
   /// The current string the user is editing.
   String get text => value.text;
+
+  /// Returns true if the current [text] is different from the original text.
+  bool get isTextModified => _originalText != value.text;
 
   /// Updates the current [text] to the given `newText`, and removes existing
   /// selection and composing range held by the controller.
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -219,17 +219,6 @@ class _RenderCompositionCallback extends RenderProxyBox {
 ///    controlled with a [TextEditingController].
 ///  * Learn how to use a [TextEditingController] in one of our [cookbook recipes](https://docs.flutter.dev/cookbook/forms/text-field-changes#2-use-a-texteditingcontroller).
 class TextEditingController extends ValueNotifier<TextEditingValue> {
-  /// The original text value when this controller was created.
-  ///
-  /// This field is used to track if the text has been modified from its original
-  /// value. It remains unchanged when:
-  /// - [clear] is called
-  /// - [value] is updated directly
-  /// - [text] is updated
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -219,6 +219,17 @@ class _RenderCompositionCallback extends RenderProxyBox {
...
 class TextEditingController extends ValueNotifier<TextEditingValue> {
+  /// The original text value when this controller was created.
+  ///
+  /// This field is used to track if the text has been modified from its original
+  /// value. It remains unchanged when:
+  /// - [clear] is called
+  /// - [value] is updated directly
+  /// - [text] is updated
+  ///
+  /// This field is automatically cleared when the controller is disposed.
```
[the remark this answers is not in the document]
Musaddiq625: Documentation updated

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -10207,6 +10207,75 @@ void main() {
...
+      
+      // Verify initial state
+      expect(controller.isTextModified, false);
+      
+      // Dispose the controller
+      controller.dispose();
+      
+      // Verify _originalText is cleared
+      expect(controller.isTextModified, false);
+    });
```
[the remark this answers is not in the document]
Musaddiq625: Test cases updated, duplicate/incorrect test cases removed

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -242,24 +242,45 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
...
+  ///
+  /// This field is used to track if the text has been modified from its original
+  /// value. It remains unchanged when:
+  /// - [clear] is called
+  /// - [value] is updated directly
+  /// - [text] is updated
+  ///
+  /// The field will be garbage collected when the controller is no longer
+  /// referenced, along with the rest of the controller's state.
+  String? _originalText;
```
justinmc, line 276: This could be a performance problem if there is a lot of text in here. And I feel like you might be able to implement this feature without this...

justinmc (commented):
Can you implement this kind of thing on top of TextEditingController without modifying the code in Flutter? I see what you're trying to do but I want to avoid adding bloat to a core class like this if possible.

Let me know if you have a use case that's not possible without this change.

packages/flutter/lib/src/widgets/editable_text.dart
Musaddiq625: I understand your concern about overhead, especially with large text. However, I think a lot of text shouldn't be an issue since it's just holding the original value in a String.

It’s possible to do this outside the class, but that would mean manually handling the original text each time, which could lead to more duplicate code.

One use case for me is with multi-page forms where I need to track if any field has been updated. Last year, I did this manually by creating *duplicate variables*, and it became a.. bit of a hassle.

cedvdb:
You could extend text editing controller instead.

Musaddiq625:
> You could extend text editing controller instead.

@cedvdb 
Yes it can be done by extension, but it's my 3rd or 4th project in which I'm adding the same functionality. So now I have a feeling this should be a part of it.

I feel like this problem could be faced by many other developers as well, and to avoid extensions or custom implementations, it can be integrated directly into the controller itself.
I think if this were part of it, it would be much easier to implement.

@justinmc 
If overhead is a concern, I can implement it with something like:
`enableCache: true`
to only store the original text when needed

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -242,24 +242,45 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
...
+        value == null ||
+            !value.composing.isValid ||
+            value.isComposingRangeValid,
         'New TextEditingValue $value has an invalid non-empty composing range '
         '${value.composing}. It is recommended to use a valid composing range, '
         'even for readonly text fields.',
       ),
       super(value ?? TextEditingValue.empty);
 
+
```
justinmc, line 265: The extra line of whitespace isn't needed here.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -242,24 +242,45 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
...
+  /// The original text value when this controller was created.
+  ///
+  /// This field is used to track if the text has been modified from its original
+  /// value. It remains unchanged when:
+  /// - [clear] is called
+  /// - [value] is updated directly
+  /// - [text] is updated
```
justinmc, line 272: This is somewhat opinionated, pushing me towards thinking that users should implement this themselves.

justinmc (commented):
I think my recommendation is that we should not adopt this because of API bloat and performance concerns for something that can be done at the app developer level already. Maybe it would make a good Pub package.

However I could be wrong if lots of people are having this problem and the benefits outweigh the costs. @Musaddiq625 is there an issue for this somewhere? Or if not, consider creating one to gather input from anyone else experiencing this problem.

I'll leave this PR open for now and defer the final decision to @loic-sharma and @Renzo-Olivares.

Either way @Musaddiq625 thank you for taking the time to submit this PR and getting us thinking about this.

Musaddiq625:
> I think my recommendation is that we should not adopt this because of API bloat and performance concerns for something that can be done at the app developer level already. Maybe it would make a good Pub package.
> 
> However I could be wrong if lots of people are having this problem and the benefits outweigh the costs. @Musaddiq625 is there an issue for this somewhere? Or if not, consider creating one to gather input from anyone else experiencing this problem.
> 
> I'll leave this PR open for now and defer the final decision to @loic-sharma and @Renzo-Olivares.
> 
> Either way @Musaddiq625 thank you for taking the time to submit this PR and getting us thinking about this.

Thanks, I totally understand.

I’ll wait to hear back from @loic-sharma and @Renzo-Olivares before proceeding further.

btw.. if this is going to be a performance bottleneck, separating it into another package probably won't help much 😄
so in that case, we can discard the idea altogether 🤷🏻‍♂️

Appreciate the feedback!

Renzo-Olivares:
Hi @Musaddiq625, thank you for your contribution. I'm leaning towards recommending we don't add this to the framework because of the reasons listed previously in API bloat and performance concerns. Unless there is a use-case that cannot be accomplished by extending the `TextEditingController` and adding this feature to it.

Musaddiq625:
> Hi @Musaddiq625, thank you for your contribution. I'm leaning towards recommending we don't add this to the framework because of the reasons listed previously in API bloat and performance concerns. Unless there is a use-case that cannot be accomplished by extending the `TextEditingController` and adding this feature to it

Thank you all for taking the time to review this PR, really appreciate ❤️
I understand the points about API bloat and performance trade-offs

Also thanks a lot for the guidance, @justinmc, based on your suggestion, I’ve implemented this externally and released it as a package: [cached_text_editing_controller](https://pub.dev/packages/cached_text_editing_controller)

Hopefully, it can help others with a similar use case without adding extra weight to the framework ✨

PR #176029, opened 2025-09-25
fix: path was url decoded but it should not for browser

ArthurHlt:
Web engine on `routeInformationUpdated` navigation message was handling uri parameter wrongly by decoding url encoded path leading to browser setting incorrect url.

Fix Issue:
- https://github.com/flutter/flutter/issues/171757

## Pre-launch Checklist

- [x] I signed the [CLA].


mdebbar (changes requested):
Thanks for the contribution!

Have you tried it in a Flutter app to confirm that this fix works?

We also need some tests before merging the PR.

PR #176063, opened 2025-09-25
Android implementation of content sizing

### LINKED ISSUE #149033: Allow an embedded Flutter View on Android to size itself

mboetger:
Android Embedder implementation of content sizing. 

Fixes: #149033 

- [X] I signed the [CLA].


engine/src/flutter/shell/platform/android/android_shell_holder_unittests.cc
```diff
@@ -210,7 +214,7 @@ TEST(AndroidShellHolder, CreateWithUnMergedPlatformAndUIThread) {
       Settings::MergedPlatformUIThread::kDisabled;
   auto jni = std::make_shared<MockPlatformViewAndroidJNI>();
   auto holder = std::make_unique<AndroidShellHolder>(
-      settings, jni, AndroidRenderingAPI::kImpellerOpenGLES);
+      settings, jni, AndroidRenderingAPI::kImpellerOpenGGLES);
```
gmackall: Is this intentional?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -110,12 +111,17 @@ public class FlutterView extends FrameLayout
   private static final String TAG = "FlutterView";
   private static final String GBOARD_PACKAGE_NAME = "com.google.android.inputmethod.latin";
 
+  @VisibleForTesting static final int CONTENT_SIZING_MAX = 2 << 12; // 2 << 13
```
gmackall: Is this commented out code from another implementation, or does it mean something I don't understand here?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -1496,10 +1562,24 @@ private void sendViewportMetricsToFlutter() {
...
+      viewportMetrics.minHeight = 0;
+      viewportMetrics.maxHeight = CONTENT_SIZING_MAX;
+    } else {
+      viewportMetrics.minHeight = viewportMetrics.height;
+      viewportMetrics.maxHeight = viewportMetrics.height;
+    }
+    if (widthMode == MeasureSpec.UNSPECIFIED) {
+      Log.d(TAG, "FlutterView width is set to wrap content - updating viewport metrics to max");
+      viewportMetrics.minHeight = 0;
+      viewportMetrics.maxHeight = CONTENT_SIZING_MAX;
```
gmackall: Should these be width?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -387,6 +394,18 @@ public boolean isCodePointRegionalIndicator(int codePoint) {
...
 
+  @NonNull
+  private final Set<FlutterUiResizeListener> flutterUiResizeListeners = new CopyOnWriteArraySet<>();
+
+  @NonNull
+  private final FlutterUiResizeListener flutterUiResizeListener =
+      new FlutterUiResizeListener() {
+        @Override
+        public void resizeEngineView(int width, int height) {
+          shouldSendViewportMetrics.set(false);
```
gmackall: This seems like an odd implementation of a method called `resizeEngineView` (it doesn't do any resizing, and is behaving more like an `onEngineViewResized`. Maybe it could be called `onEngineViewResizeRequested` or something?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -1496,10 +1562,24 @@ private void sendViewportMetricsToFlutter() {
               + "FlutterView was not attached to a FlutterEngine.");
       return;
     }
-
     viewportMetrics.devicePixelRatio = getResources().getDisplayMetrics().density;
     viewportMetrics.physicalTouchSlop = ViewConfiguration.get(getContext()).getScaledTouchSlop();
-
+    if (heightMode == MeasureSpec.UNSPECIFIED) {
```
gmackall: It seems a little odd to me to do this manipulation of the actual values of the viewport metrics here in a method called `sendViewportMetricsToFlutter` (if I were debugging I think I would be surprised to find that the method which purports to send values is also setting them, and also now maybe not sending them).

Do all of the call sites of this method need this manipulation? If so should we maybe move it to its own method, and if not can we maybe lift the manipulation into the call site that needs it? Also what do you think about renaming to `maybeSendViewportMetricsToFlutter`

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -113,6 +115,11 @@ public class FlutterJNI {
   // platform thread and doesn't require locking.
   private ReentrantReadWriteLock shellHolderLock = new ReentrantReadWriteLock();
 
+  // Boolean that gates if view port metrics should be sent.  When the
+  // engine informs the embedder of a resize, it is not necessary send
```
gmackall:
```suggestion
  // engine informs the embedder of a resize, it is not necessary to send
```

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
gmackall: Also, just for my understanding - I believe you mentioned this but I want to make sure I understand. The case we want to prevent is the engine telling the embedder to resize, and then `FlutterView.onSizeChanged` triggering and sending those values which were just requested right back to the engine?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -175,6 +181,35 @@ public boolean deliverSelfNotifications() {
...
+                changed = true;
+                surfaceParams.height = height;
+              }
+            }
+            if (widthMode == MeasureSpec.UNSPECIFIED) {
+              if (flutterEngineView.getWidth() != width) {
+                changed = true;
+                surfaceParams.width = width;
+              }
+            }
```
gmackall, line 206: Instead of having a `FlutterJNI` level `AtomicBoolean`, can we not have `FlutterView` level one `shouldSendViewportMetrics` which we set to false here, before modifying any layout params.

And then in `onSizeChanged`, can we not skip entirely the `sendViewportMetricsToFlutter` in the case that we don't want to send? I think that flow would be easier for readers to understand personally (though we'd again need to move the work that we are doing to modify the viewport metrics that is currently added to `sendViewportMetricsToFlutter`.

Or is this point in the flow not early enough for some reason, or am I misunderstanding the flow?

gmackall (commented):
Some comments

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
gmackall: See also the question below about having a `FlutterView` level boolean which would also resolve this comment, if what I'm suggesting makes sense.

engine/src/flutter/shell/platform/android/android_shell_holder_unittests.cc
mboetger: Ugh, that would be gemini cli's doing....

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
mboetger: No, sorry.  2 << 13 is the largets I can pass to the engine without it throwing errors, but decided to dial it back for performance reasons.  Had added documentation, but it got reverted - adding back.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
mboetger: Yeah, I originally had the boolean at the FlutterView level but switched it to the JNI level to insure some other site didn't start the feedback loop off by calling sendViewportMetrics.  I can move it back.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
mboetger: Yeah, answered elsewhere.  I had this originally but decided to move it closer to the JNI layer.  Move it back.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -175,6 +188,36 @@ public boolean deliverSelfNotifications() {
...
 
+  @VisibleForTesting
+  final FlutterUiResizeListener flutterUiResizeListener =
+      new FlutterUiResizeListener() {
+        @Override
+        public void resizeEngineView(int width, int height) {
+          boolean changed = false;
+          if (flutterEngineView != null) {
+            ViewGroup.LayoutParams surfaceParams = flutterEngineView.getLayoutParams();
+            if (heightMode == MeasureSpec.UNSPECIFIED) {
```
gmackall: I'm curious - if we aren't in unspecified height mode (i.e. aren't using a content sized view), does this still not mean we resized here? 

I would sort of think, given the logic below, that a resize is happening either way - would we really expect this method to be called in non-content-sized view world, and get here and short circuit out?

gmackall (approved):
one question, but LGTM

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
mboetger: I guess no - technically we do not expect this method to be called in a non-content sized world.  It's possible since the engine does not check before "maybeResizing".  This check is being performed on the initial kick-off of the code path so we can remove it.

I guess I'm just doing this a safety/sanity check.  Removing this could open us up to bugs in the future, but it's reasonable to say this is unnecessary.  I will remove.

gmackall: > I guess I'm just doing this a safety/sanity check. Removing this could open us up to bugs in the future

Yeah I suspected that was the case. Perhaps we could use an `assert` here? I'm not actually sure how those function in the Java code here, but I see we've used them in a couple other places (though not very many)

https://github.com/search?q=repo%3Aflutter%2Fflutter+assert+language%3AJava+&type=code

Not sure if these asserts are actually configured to work in debug but not release, anyways lgtm either way just a suggestion

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -477,7 +525,31 @@ protected void onSizeChanged(int width, int height, int oldWidth, int oldHeight)
...
+    if (widthMode == MeasureSpec.UNSPECIFIED) {
+      Log.d(TAG, "FlutterView width is set to wrap content - updating viewport metrics to max");
+      viewportMetrics.minWidth = 0;
+      viewportMetrics.maxWidth = CONTENT_SIZING_MAX;
+    } else {
+      viewportMetrics.minWidth = viewportMetrics.width;
+      viewportMetrics.maxWidth = viewportMetrics.width;
+    }
+
+    if (shouldSendViewportMetrics.compareAndSet(false, true)) {
```
loic-sharma, line 546: Can an Android Flutter view be both content sized and parent sized at the same time?

If yes, I fear this might result in data races. For example, imagine that the engine resizes the view concurrently while the parent also tries to resize the view. Could this result in a scenario where the wrong viewport metric is suppressed here?

Assuming I haven't missed anything, I'd consider a model where an Android Flutter view is either in "content sizing" mode, or "parent sizing" mode, but never both at the same time. This would help us not need to worry about these kinds of finicky concurrency problems.

Please let me know if I misunderstood how this works!

mboetger: The layout parameters can either be wrap_content OR match_parent - can't be both.

loic-sharma: I chatted offline with Matt, this sounds good to me 👍

Issue #171194, opened 2025-06-25
Mutating Paint breaks paints


justinmc:
### Steps to reproduce

Mutating an instance of [Paint](https://main-api.flutter.dev/flutter/dart-ui/Paint-class.html) will affect the saved calls that are used when matching with [paints](https://main-api.flutter.dev/flutter/flutter_test/paints.html) in a test.

Run the test case given below, which does this:

 1. Draws something to the canvas with a Paint instance.
 2. Later, modifies that Paint instance.
 3. Test that the drawing happened with the unmodified Paint.

### Expected results

The test passes. `paints` sees the correct Paint that was used at the time of the draw.

### Actual results

The test fails. `paints` sees the modified Paint.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';
import 'package:flutter_test/flutter_test.dart';

void main() {
  testWidgets('paints.circle is not affected by mutated colors', (
    WidgetTester tester,
  ) async {
    final Key customPaintKey = UniqueKey();
    Widget buildApp({required bool enabled}) {
      return MaterialApp(
        home: Scaffold(
          body: Center(child: CustomPaint(key: customPaintKey, painter: _MutantPainter())),
        ),
      );
    }

    await tester.pumpWidget(buildApp(enabled: true));

    // Selected and enabled.
    await tester.pumpAndSettle();
    expect(
      Material.of(tester.element(find.byKey(customPaintKey))),
      paints..circle(color: Colors.red),
    );
  });
}

class _MutantPainter extends ChangeNotifier implements CustomPainter {
  @override
  void paint(Canvas canvas, Size size) {
    final Paint paint = Paint()..color = Colors.red;

    canvas.drawCircle(Offset.zero, 10.0, paint);

    // Mutate paint after drawing.
    paint.color = Colors.blue;
  }

  @override
  bool shouldRepaint(covariant CustomPainter oldDelegate) => false;

  @override
  bool? hitTest(Offset position) => null;

  @override
  SemanticsBuilderCallback? get semanticsBuilder => null;

  @override
  bool shouldRebuildSemantics(covariant CustomPainter oldDelegate) => true;
}
```

</details>

### Logs

<details><summary>Test output</summary>

```console
✗ flutter test /usr/local/google/home/jmccandless/Projects/flutter/packages/flutter_test/test/recording_canvas.dart
Resolving dependencies... 
Downloading packages... 
[...]

PR #176073, opened 2025-09-26
Colored box optimization (#176028)

### LINKED ISSUE #176028: ColoredBox create unwanted strip when placed close together due to the framework assuming isAntiAlias ​​= true
### Steps to reproduce

I encountered this issue when placing multiple widgets together with the same background. It creates an unwanted visible strip in the background.

Image

I tried the same code in other frameworks, such as SwiftUI, and it rendered as expected.

| Flutter | SwiftUI | 
| :---: | :---: | 
| Image | Image | 
| Image | Image | 

### Expected results

No unwanted strip when render multiple widgets with the same background color close together.

Image

### Actual results

I dug into the implementation of ColoredBox and noticed that the paint function always assumes that Paint() is anti-aliased by default. This is generally good, but as we can see when drawing a rectangle, there’s no need to enable anti-aliasing to ensure smoothness. In fact, it introduces a visual bug like the one I described above.

```dart

class _RenderColoredBox extends RenderProxyBoxWithHitTestBehavior {
   ...

  @override
    // It's tempting to want to optimize out this `drawRect()` call if the
    // color is transparent (alpha==0), but doing so would be incorrect. See
    // https://github.com/flutter/flutter/pull/72526#issuecomment-749185938 for
    // a good description of why.
    if (size > Size.zero) {
      context.canvas.drawRect(offset & size, Paint()..color = color);
    }
    if (child != null) {
      context.paintChild(child!, offset);
    }
  }
}
```

The fix for this problem is simple:

```dart
@override
    // It's tempting to want to optimize out this `drawRect()` call if the
    // color is transparent (alpha==0), but doing so would be incorrect. See
    // https://github.com/flutter/flutter/pull/72526#issuecomment-749185938 for
    // a good description of why.
    if (size > Size.zero) {
      context.canvas.drawRect(
        offset & size,
        Paint()
          ..color = color
          ..isAntiAlias = false,
      );
    }
    if (child != null) {
      context.paintChild(child!, offset);
    }
  }
```

### Code sample

Dartpad link: https://dartpad.dev/?id=72d0881637a25d884d7b922cab8bb0ff

### Screenshots or Video

_No response_

### Logs

_No response_

### Flutter Doctor output

```console
[✓] Flutter (Channel master, 3.37.0-1.0.pre-232, on macOS 26.0 25A354 darwin-arm64, locale en-VN)
[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 26.0.1)
    ! Error: Browsing on the local area network for fhihung.iphone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for Bui’s iPad. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for Duong phone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for zennn.phone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for iPad của Chi bé. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for iPhone 12 Chi. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
[✓] Network resources

• No issues found!
```

</details>

definev:
This PR fixed visual bug when placing multiple `ColoredBox`s together. (Fixes #176028)

## Pre-launch Checklist

- [x] I signed the [CLA].


victorsanni (commented):
Hi @definev, thanks for the PR. Could you add a test to make sure this change does not get regressed?

definev:
Hi @victorsanni, I created the test. Feel free to let me know if you’d like me to write more tests for it. This is my first time contributing to Flutter, thank you so much ☺️

victorsanni (commented):
This [golden test](https://github.com/flutter/packages/blob/c7a25fa3463b6e9033a1c6db7acff42e1a3a8799/packages/rfw/test/argument_decoders_test.dart#L495) in the rfw package is failing with the changes in this PR.

definev:
I think that’s the expected behavior since it changes the way ColoredBox is rendered. The next step I should take is to either regenerate all the golden tests or declare this as a breaking change that may affect existing golden tests? cc: @victorsanni 
Screenshot 2025-09-30 at 01 56 11

definev:
@victorsanni @dkwingsmt Are there any new updates for me to do?

victorsanni (approved):
LGTM awaiting review from someone more knowledgeable about the subject such as @dkwingsmt.

For the failing customer test, the process is to disable it in a separate PR which will land before this one, then re-enable it after this PR lands while approving the new goldens. I can open a PR to do that.

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -8430,14 +8430,17 @@ class _RenderColoredBox extends RenderProxyBoxWithHitTestBehavior {
...
   @override
   void paint(PaintingContext context, Offset offset) {
     // It's tempting to want to optimize out this `drawRect()` call if the
     // color is transparent (alpha==0), but doing so would be incorrect. See
     // https://github.com/flutter/flutter/pull/72526#issuecomment-749185938 for
     // a good description of why.
     if (size > Size.zero) {
-      context.canvas.drawRect(offset & size, Paint()..color = color);
+      _paint.color = color;
+      context.canvas.drawRect(offset & size, _paint);
```
justinmc: Can you instantiate the Paint here instead of reusing it each time? I avoid try to avoid this pattern after seeing https://github.com/flutter/flutter/issues/171194, even though that should be fixed by now.

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -8429,7 +8429,12 @@ class _RenderColoredBox extends RenderProxyBoxWithHitTestBehavior {
     // https://github.com/flutter/flutter/pull/72526#issuecomment-749185938 for
     // a good description of why.
     if (size > Size.zero) {
-      context.canvas.drawRect(offset & size, Paint()..color = color);
+      context.canvas.drawRect(
+        offset & size,
+        Paint()
+          ..color = color
+          ..isAntiAlias = false,
```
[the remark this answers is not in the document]
justinmc: I really wonder if there is a performance benefit here in real life or not. After seeing https://github.com/flutter/flutter/issues/171194 it always makes me nervous. Fine as-is though.

justinmc (approved):
LGTM 👍 . I would guess that we were notting getting much benefit from setting isAntiAlias to true in this case, but I'm interested to see if the Google tests have any failures. We should watch out for any failures that appear to have benefitted from antialiasing.

And +1 to @dkwingsmt reviewing this as well.

packages/flutter/lib/src/widgets/basic.dart
definev: Sure, that my initial approach to. Seeing Gemini suggest reuse paint object so i change to fit it comment :-) Let reverse that

definev: Updated

dkwingsmt:
Thanks for the nomination, although I wouldn't call myself an expert at all. :)

I played with this change locally and observed that there are visual changes to all `ColoredBox`es, as expected. The edges are no longer as clear.
image


Even if we think this change is too small to notice or even healthy for having a clearer border, it's bound to break millions of golden tests. The impact is far too great. I don't think we can justify this breakage.

It might be better to expose the `isAntiAlias` as a parameter and keep the current behavior default. I'm not sure how viable it is, since the parameter might need to be propagated to some other widgets, such as `Container`. But it's definitely better than breakage like this.

justinmc:
Good find @dkwingsmt, I was worried something like that might happen. +1 to your recommendation of exploring exposing isAntiAlias.

definev:
I think the effect should apply to all ColoredBox widgets since they depend on each other. For example, placing one anti-aliased ColoredBox between two non-anti-aliased ones still causes the visual bug to appear.

We could introduce a new static variable in ColoredBox, called defaultIsAntiAlias, which would be set to true by default. However, users could set it to false if they prefer crisp borders.

If we still want to expose it, let add a 'bool? isAntiAlias' and set it to defaultIsAntiAlias if null.

dkwingsmt:
> I think the effect should apply to all ColoredBox widgets since they depend on each other. For example, placing one anti-aliased ColoredBox between two non-anti-aliased ones still causes the visual bug to appear.

Turning off anti-alias has many down sides, such as artifacts during rotation or zooming animation. Notice the borders in the following examples:

image

https://github.com/user-attachments/assets/4af30f8c-2e50-42c3-aa3a-52d54c78c918

It should be up to the developers to decide if they really want to use, or need, the non-antialias version. In general, the antialias version should be more widely applicable.

definev:
Agree! I’ve updated the code based on your comment, exposing isAntiAlias in ColoredBox.
One small detail I observed in the SwiftUI framework is that they also use antiAlias, but they don’t have the same visual bug we encountered. Apparently, Flutter on mobile doesn’t have that issue either, so I think it’s related to the engine layer. The current solution is to expose isAntiAlias in ColoredBox.

Note: The stripe is still visible in SwiftUI, but it’s much thinner than in Flutter and only appears when rotating the child.

https://github.com/user-attachments/assets/5b4eaace-6808-45a6-9b54-4f750955c9f8 

https://github.com/user-attachments/assets/bd2b34e1-ea59-4254-bf48-f73edb49bb11

| SwiftUI | Flutter |
|---|---|
| Screenshot 2025-10-09 at 18 32 33 | Screenshot 2025-10-09 at 18 38 24 |

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -8392,31 +8392,38 @@ class _StatefulBuilderState extends State<StatefulBuilder> {
 /// child on top of that color.
 class ColoredBox extends SingleChildRenderObjectWidget {
   /// Creates a widget that paints its area with the specified [Color].
-  const ColoredBox({required this.color, super.child, super.key});
+  const ColoredBox({required this.color, this.isAntiAlias = true, super.child, super.key});
 
   /// The color to paint the background area with.
   final Color color;
 
+  /// Whether to apply anti-aliasing to the color. Defaults to true.
```
dkwingsmt: I think the developers probably would like more explanation on this property. Here's my proposal (suggestions welcome!):

```suggestion
  /// Whether to apply anti-aliasing when painting the box.
  ///
  /// Defaults to `true`.
  ///
  /// When `true`, the painted box will have smooth edges. This is crucial for
  /// animations and transformations (such as rotation or scaling) where the
  /// widget's edges may not align perfectly with the physical pixel grid.
  /// Anti-aliasing allows for sub-pixel rendering, which prevents a 'jagged'
  /// appearance during motion and ensures visually smooth transitions.
  ///
  /// Set this to `false` for specific use cases where multiple `ColoredBox`
  /// widgets are positioned adjacent to each other to form a larger, seamless
  /// area of solid color. With anti-aliasing enabled (`true`), faint seams or
  /// gaps might appear between the boxes due to the semi-transparent pixels at
  /// their edges. Disabling anti-aliasing ensures that the boxes align perfectly
  /// without such visual artifacts.
  ///
  /// See also:
  ///
  ///  * [Paint.isAntiAlias], the underlying property that this controls.
```

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -326,6 +327,9 @@ class Container extends StatelessWidget {
   /// null.
   final Color? color;
 
+  /// Whether to apply anti-aliasing to the color. Defaults to true.
```
dkwingsmt: After we've decided on `ColoredBox.isAntiAlias`'s doc, can you clone it to here with the `@template` and `@macro` tags?

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -8392,31 +8392,38 @@ class _StatefulBuilderState extends State<StatefulBuilder> {
...
+    (renderObject as _RenderColoredBox)
+      ..color = color
+      ..isAntiAlias = isAntiAlias;
   }
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
     properties.add(DiagnosticsProperty<Color>('color', color));
+    properties.add(DiagnosticsProperty<bool?>('isAntiAlias', isAntiAlias));
```
dkwingsmt: Maybe `defaultValue: true`?

packages/flutter/test/widgets/basic_test.dart
```diff
@@ -1220,6 +1220,24 @@ void main() {
...
+    });
+
+    testWidgets('ColoredBox - passing isAntiAlias = false', (WidgetTester tester) async {
+      await tester.pumpWidget(const ColoredBox(color: colorToPaint, isAntiAlias: false));
+      expect(find.byType(ColoredBox), findsOneWidget);
+      final RenderObject renderColoredBox = tester.renderObject(find.byType(ColoredBox));
+
+      renderColoredBox.paint(mockContext, Offset.zero);
+      expect(mockCanvas.paints.single.isAntiAlias, isFalse);
+    });
```
dkwingsmt, line 1240: Can you create a golden test to display the difference from antialising? Something just like our screenshots during the discussion. I'm not sure if the antialiasing shows well on golden though and we'll have to see.

packages/flutter/lib/src/widgets/basic.dart
definev: I like the detailed explanation for each use case 👍

definev:
@dkwingsmt i added the test but i don't know how to make it generate new golden test image and the CI is failing i don't know why. I follow the Writing-a-golden-file-test-for-package-flutter.md but seem i don't need to edit ci.yaml file

dkwingsmt:
The `ci.yaml` validation failure is simply due to the PR base being too old. I've updated it with the latest master. The golden file will be generated once all checks are green. Let's see how it goes. Thank you.

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -326,6 +327,11 @@ class Container extends StatelessWidget {
   /// null.
   final Color? color;
 
+  /// The [ColoredBox.isAntiAlias] property.
+  ///
```
dkwingsmt: The macro already has the first paragraph.
```suggestion
```

packages/flutter/lib/src/widgets/basic.dart
dkwingsmt: This isn't fixed?

definev: I already fixed it

packages/flutter/lib/src/widgets/container.dart
definev: Done ✅

packages/flutter/lib/src/widgets/basic.dart
definev: I set default to true on constructor

justinmc:
@definev Looks like there are still analyzer failures FYI.

definev:
I will check on that

definev:
I’ve fixed the format, but I haven’t found a way to resolve the Mac framework_tests_impeller and tree-status issues yet.
One more question - I saw the [flutter/goldens](https://github.com/flutter/goldens) repository. Do I need to upload the golden test images before merging the PR into master?

victorsanni:
> I’ve fixed the format, but I haven’t found a way to resolve the Mac framework_tests_impeller and tree-status issues yet.

Those are infra issues unrelated to this PR.

> One more question - I saw the [flutter/goldens](https://github.com/flutter/goldens) repository. Do I need to upload the golden test images before merging the PR into master?

No, adding a golden test using `matchesGoldenFile` is sufficient. A contributor with commit access will then go ahead to triage the golden images in the flutter gold tool.

packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -8392,31 +8392,59 @@ class _StatefulBuilderState extends State<StatefulBuilder> {
...
+    (renderObject as _RenderColoredBox)
+      ..color = color
+      ..isAntiAlias = isAntiAlias;
   }
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
     properties.add(DiagnosticsProperty<Color>('color', color));
+    properties.add(DiagnosticsProperty<bool>('isAntiAlias', isAntiAlias));
```
dkwingsmt: `defaultValue: true`?

dkwingsmt:
The golden tests don't seem to demonstrate the non-antialias grids well. I think it's because the dimensions aren't fractional, which would be more noticeable when the dimensions are decided by text.

image

definev:
Got it, I'll create a more concrete example with similar colored squares placed close together to demonstrate the original problem

dkwingsmt:
Yeah. Instead of fixed widths maybe try wrapping them with text.

definev:
I intentionally set `scale: 1.04` to demonstrate the anti-aliasing process. It doesn't show the golden image as expected if I set it to 1 on my mac.
image

packages/flutter/test/widgets/basic_test.dart
```diff
@@ -1220,6 +1220,210 @@ void main() {
...
+
+    testWidgets('ColoredBox golden test - anti-aliasing and rotation variations', (
+      WidgetTester tester,
+    ) async {
```
dkwingsmt, line 1248: nit: Document how to interpret the golden results. My proposal:
```dart
  // This test verifies how `ColoredBox.isAntiAlias` affects rendering.
  // The first row uses `isAntiAlias: true`, showing gaps between the white backgrounds.
  // The second row uses `isAntiAlias: false`, demonstrating no gaps between the white backgrounds.
  // The third row contains three tilted boxes with `isAntiAlias` set to true, false, and false, respectively.
```

dkwingsmt:
The 1.04x scale is brilliant! The golden file looks great as well. I only have two comments left.

justinmc (approved):
LGTM as long as @dkwingsmt's comments are addressed. Thanks for pivoting on this PR!

packages/flutter/lib/src/widgets/basic.dart
definev: This is already true i set `isAntiAlias: true` in the constructor `const ColoredBox({required this.color, this.isAntiAlias = true, super.child, super.key});`

packages/flutter/test/widgets/basic_test.dart
definev: Done ✅

packages/flutter/lib/src/widgets/basic.dart
dkwingsmt: Nope, this `defaultValue: true` is used to tell `DiagnosticsProperty` not to display this property when the value is the default value. For example, the reason why `framework_test.dart` needs changing in this PR is because this property. Although this change is relatively subjective, I think it's a good move because it avoids breaking existing logging test and also avoids printing relatively nuance info.

definev:
Thanks @dkwingsmt, @justinmc, and @victorsanni for supporting me in resolving issues and guiding me through my very first PRs in the Flutter project :-) Because of the time zone difference, I couldn’t reply to you guys right away, but you were patient and helped me finish it!

packages/flutter/lib/src/widgets/basic.dart
definev: Ah got it, i thought you was mention the `bool?` i mistakenly put in very first commit

definev: Done ✅ Thanks for the clear explanation!

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -257,6 +257,7 @@ class Container extends StatelessWidget {
     this.alignment,
     this.padding,
     this.color,
+    this.isAntiAlias = true,
```
victorsanni, line 260: Should `Container` test for this property as well?

definev: Could be, but I think we don’t need to test it since under the hood Container still uses `ColoredBox`, so their behavior should be the same. Maybe we should add `isAntiAlias` to existing `Decoration`(`BoxDecoration`, `ShapeDecoration`, etc) to ensure consistent behavior when users want to configure the color? After that, we can add a test for it.

victorsanni: > I think we don’t need to test it since under the hood Container still uses ColoredBox

I don't think that's enough of a reason not to test it. Imagine a future where someone removes the property from `Container`. Without a test such a change won't raise any alarms.

definev: I think it’s reasonable. Currently, we don’t expose isAntiAlias on `Decoration` yet, so if we test it in `Container`, it should only use `ColoredBox`, meaning the behavior will be the same.
On the other hand, Container is usually used with a `decoration` setup that includes things like `color`, `borderRadius`, etc. So if we want to write a test for `Container`, we should first implement `isAntiAlias` in the `BoxDecoration` and `ShapeDecoration` classes, and then write tests for them.

definev: If you think it’s too big for this PR, I can create a new one to implement isAntiAlias on Decoration for consistent behavior 👍

victorsanni: > If you think it’s too big for this PR, I can create a new one to implement isAntiAlias on Decoration for consistent behavior 👍

That would be great. What about for `Container` in a separate PR also?

definev: Sure, I'd love to do that 💙 I’ll write tests for `Container` to verify `isAntiAlias` in both cases: when using `Decoration`, and when using the `isAntiAlias` + `color` properties in that PR too ✅

dkwingsmt (approved):

PR #176105, opened 2025-09-26
Engine version for 3.35.5

justinmc:
Sets the engine version to the the most recent engine change (https://github.com/flutter/flutter/pull/176030).


reidbaker (approved):
RSGLTM

PR #176114, opened 2025-09-26
[ Widget Preview ] Improve IDE integration support

### LINKED ISSUE #176113: Support navigating to source locations from stack frames displayed by `WidgetPreviewErrorWidget`
`WidgetPreviewErrorWidget` displays a list of stack frames if a widget throws an exception during initialization. The script URIs for each stack frame should be links that cause connected IDEs to navigate to the specified source location.

bkonyi:
This change adds support for two DTD Editor service RPCs and makes functionality that's dependent on the availability of the Editor service reactive to the service availability changing.

The newly added `getActiveLocation` Editor RPC now allows for us to properly determine which source file is selected at startup, removing the need for special casing around initializing the filtered preview set.

The `navigateToCode` Editor RPC is now supported, and URIs in stack traces can now be clicked to navigate to the source location in the IDE.

Both filtering previews by selected file and stack frame navigation links are only enabled when the Editor service is available.

Fixes https://github.com/flutter/flutter/issues/176113


bkonyi:
@DanTup would you be okay to review this? No pressure if not!

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/editor_service.dart.tmpl
```diff
@@ -41,14 +65,59 @@ mixin DtdEditorService {
...
+          // Manually retrieve the currently selected source file.
+          unawaited(_updateSelectedSourceFile());
+          _editorServiceAvailable.value = true;
+        case DTDEvent(
+          kind: kServiceUnregistered,
+          data: {DtdParameters.service: kEditorService},
+        ):
+          // TODO(bkonyi): this needs to be made more robust if we're going to use multiple service
+          // methods. It's unlikely that individual methods on the Editor service will be added or
+          // removed, but we should at least log an error if this happens.
```
DanTup: Aren't we already using multiple methods? (`getActiveLocation` and `navigateToCode`)?

Shipped versions of IntelliJ probably have `navigateToCode` but will not have `getActiveLocation`, so I think these should be tracked separately.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
```diff
@@ -89,15 +105,31 @@ class _WidgetPreviewErrorWidget extends StatelessWidget {
     // Print out the stack trace nicely formatted.
     return frames.map<TextSpan>((frame) {
       if (frame is UnparsedFrame) return TextSpan(text: '$frame\n');
+      // The Editor service can't handle Dart core library paths, so don't
+      // allow for navigation to them. Also disable navigation if the Editor
+      // service isn't available.
```
DanTup: Maybe refer to "Editor.navigateToCode service method" (here and in other places) instead of "Editor service" - it might be clearer to any readers unfamiliar with these.

packages/flutter_tools/test/commands.shard/permeable/widget_preview/dtd_services_test.dart
```diff
@@ -5,11 +5,13 @@
 import 'dart:async';
 
 import 'package:dtd/dtd.dart';
+import 'package:file/memory.dart';
 import 'package:flutter_tools/src/base/logger.dart';
 import 'package:flutter_tools/src/base/process.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 import 'package:flutter_tools/src/widget_preview/dtd_services.dart';
 
+import '../../../general.shard/migrations/swift_package_manager_gitignore_migration_test.dart';
```
DanTup: This seems weird... I think it's to get `FakeFlutterProject` which is only a few lines.. maybe it should be moved, or just copied here?

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/editor_service.dart.tmpl
bkonyi: Yes, I meant to get rid of this TODO. Ideally Intellij will also have an implementation of `getActiveLocation` before we support previews, so we hopefully won't need to track them separately.

packages/flutter_tools/test/commands.shard/permeable/widget_preview/dtd_services_test.dart
bkonyi: I apparently just auto-imported the first available `FakeFlutterProject` thinking it would be in a common utility file 🤦

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_rendering.dart.tmpl
bkonyi: Done here. Elsewhere, we're referring to the Editor service in general, not any particular RPC.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/editor_service.dart.tmpl
DanTup: > Ideally Intellij will also have an implementation of `getActiveLocation` before we support previews

Ah yes, I forgot it wasn't already done today. Perhaps we should note that assumption in a comment then (and maybe note on https://github.com/flutter/flutter-intellij/issues/8543 so it's done together?).

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/editor_service.dart.tmpl
```diff
@@ -41,14 +65,61 @@ mixin DtdEditorService {
...
+          },
+        ):
+          // Manually retrieve the currently selected source file.
+          unawaited(_updateSelectedSourceFile());
+          _editorServiceAvailable.value = true;
+        case DTDEvent(
+          kind: kServiceRegistered,
+          data: {DtdParameters.service: kEditorService},
+        ):
+          _editorServiceAvailable.value = true;
```
DanTup, line 86: This additional check added in the last commit might affect already-shipped versions of Dart-Code (where people have enabled the preview), since those versions register many Editor services, but neither `getActiveLocation` or `navigateToCode` (they are only currently in pre-releases).

I don't know if that's a big deal though (it's an opt-in experiment, the update will ship next week, and you might have error handling that handles them not existing), but I'll let you decide :-)

DanTup (approved):

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/editor_service.dart.tmpl
bkonyi: Yeah, I'm not too concerned here, especially with VSCode auto-updating the Dart-Code plugin. I don't think this would cause any noticeable breakages :)

PR #173715, opened 2025-08-13
Implement Regular Windows for the win32 framework + add an example application for regular windows

mattkae:
## What's new?
- Implement and test `_window_win32.dart`, the Win32 implementation of `_window.dart` :tada: 
- Wrote a bunch of tests in `window_win32_test.dart` which uses a mock API barrier to mock the native Win32 layer
- Update `BaseWindowController` to extend `ChangeNotifier` and use it to inform the `WindowScope` of when things change
- Made `RegularWindowController.empty()` a non-internal constructor, as it is needed by implementations
- Added a _large_ example application for windowing where users can test out different windowing concepts and see how they work!

To test, simply run the tests or try out the example app!

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #176187, opened 2025-09-29
Implement Regular Windows for Linux

robert-ancell:


robert-ancell:
Putting this up for review now. I'm still working on the integration tests, but I figure it's better to review now before they're done.

examples/multiple_windows/linux/runner/main.cc
```diff
@@ -0,0 +1,6 @@
+#include "my_application.h"
```
mattkae, line 5: CI is complaining that this file (along iwth `my_application.c` and `my_application.h`) lack license comment headers

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+
+String _nativeToString(ffi.Pointer<ffi.Uint8> value) {
+  var length = 0;
+  while (value[length] != 0) {
+    length++;
+  }
+  return utf8.decode(value.asTypedList(length));
+}
+
+typedef GCallback = ffi.Void Function();
```
mattkae:
```suggestion
typedef _GCallback = ffi.Void Function();
```

We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+
+class GObject {
```
mattkae:
```suggestion
class _GObject {
```


We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+  @ffi.Native<ffi.Void Function(ffi.Pointer, ffi.UnsignedLong)>(
+    symbol: 'g_signal_handler_disconnect',
+  )
+  external static void _gSignalHandlerDisconnect(ffi.Pointer instance, int handlerId);
+  void signalHandlerDisconnect(int handlerId) {
+    _gSignalHandlerDisconnect(instance, handlerId);
+  }
+}
+
+class GtkWidget extends GObject {
```
mattkae:
```suggestion
class _GtkWidget extends _GObject {
```



We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+  const GdkWindow(ffi.Pointer instance) : super(instance);
+
+  @ffi.Native<ffi.Int Function(ffi.Pointer)>(symbol: 'gdk_window_get_state')
+  external static int _gdkWindowGetState(ffi.Pointer window);
+  int getState() {
+    return _gdkWindowGetState(instance);
+  }
+}
+
+final class GdkGeometry extends ffi.Struct {
```
mattkae:
```suggestion
final class _GdkGeometry extends ffi.Struct {
```



We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+  external int win_gravity;
+
+  factory GdkGeometry() {
+    return ffi.Struct.create();
+  }
+}
+
+const int GDK_WINDOW_STATE_ICONIFIED = 1 << 1;
+const int GDK_WINDOW_STATE_MAXIMIZED = 1 << 2;
+const int GDK_WINDOW_STATE_FULLSCREEN = 1 << 4;
```
mattkae:
```suggestion
const int _GDK_WINDOW_STATE_ICONIFIED = 1 << 1;
const int _GDK_WINDOW_STATE_MAXIMIZED = 1 << 2;
const int _GDK_WINDOW_STATE_FULLSCREEN = 1 << 4;
```



We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+
+class GtkWindow extends GtkWidget {
```
mattkae:
```suggestion
class _GtkWindow extends _GtkWidget {
```



We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+    final height = _gMalloc0(ffi.sizeOf<ffi.Int>()).cast<ffi.Int>();
+    _gtkWindowGetSize(instance, width, height);
+    final result = Size(width.value.toDouble(), height.value.toDouble());
+    _gFree(width);
+    _gFree(height);
+    return result;
+  }
+}
+
+class FlView extends GtkWidget {
```
mattkae:
```suggestion
class _FlView extends GtkWidget {
```



We can decide whether or not we wan to expose this part of the Linux API later on, but for V1, we should hide these internals.

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  @internal
+  void setConstraints(BoxConstraints constraints) {
+    _window.setGeometryHints(
+      minWidth: constraints.minWidth.toInt(),
+      minHeight: constraints.minHeight.toInt(),
+      maxWidth: constraints.maxWidth.isInfinite ? 0x7fffffff : constraints.maxWidth.toInt(),
+      maxHeight: constraints.maxHeight.isInfinite ? 0x7fffffff : constraints.maxHeight.toInt(),
```
mattkae, line 680: Is `0x7fffffff` max int or something? Maybe break it out into a constant?

mattkae (changes requested):
Code looks mostly good, but we need to hide some things just to simplify the initial PR :)

Also - could you look into integration with the integration tests that I wrote for windows: https://github.com/flutter/flutter/pull/173715/files#diff-0330858a9e31a7103b8864b06c48602346c682a94a8532fb086d22a89fae7c78

I wonder if they'll behave on Linux as well. They run in CI and are a bit complex, so let me know if you need any help!

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Would this be simpler moving into a `_gtk_ffi.dart` file?

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Yeah, this is INT_MAX which is not available from Dart afaict. Annoyingly GTK doesn't have a method of setting one of the dimensions to unlimited. Setting it to -1 causes the limit to be the requested size, i.e. unresizable (https://docs.gtk.org/gdk3/struct.Geometry.html).

robert-ancell:
> I wonder if they'll behave on Linux as well. They run in CI and are a bit complex, so let me know if you need any help!

Almost. I've enabled the tests but disabled resize and fullscreen/maximize/minimize on Linux for now as setting these doesn't immediately return the same value when getting - we need to wait until the window has resized before this is the case. I assume that on win32 these are synchronous? Do we update the test to wait? Do we need more windowing API to do this consistently across platforms?

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,546 @@
...
+       _window = GtkWindow(),
+       super.empty() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    _deleteEventId = _window.signalConnect("delete-event", () {
+      // FIXME(robert-ancell): Can't return true to stop this running default destroy handler. Might need a helper functions to make a callback that always returns true. Can use `gtk_true` in GTK3, but this doesn't exist in GTK4.
+      _delegate.onWindowCloseRequested(this);
+    });
```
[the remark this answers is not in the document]
robert-ancell: Damn, I did not think Gemini would notice this... I think this is safe, as I believe C returns 0 by default but perhaps Gemini is smarter than me. I tried but wasn't able to make signalConnect more flexible - the conclusion was I'd need to write a new method for each callback type. Also it appears we can't return values from callbacks as putting them in an isolate disallows this. I think this is almost certainly going to need to be done with a helper function.

examples/multiple_windows/linux/runner/my_application.cc
```diff
@@ -0,0 +1,146 @@
...
+  char** dart_entrypoint_arguments;
+};
+
+G_DEFINE_TYPE(MyApplication, my_application, GTK_TYPE_APPLICATION)
+
+// Called when first Flutter frame received.
+static void first_frame_cb(MyApplication* self, FlView *view)
+{
+  gtk_widget_show(gtk_widget_get_toplevel(GTK_WIDGET(view)));
+}
```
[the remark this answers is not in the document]
robert-ancell: Fixed here and fixed in template in https://github.com/flutter/flutter/pull/176256

packages/flutter/lib/src/widgets/_window_linux.dart
mattkae: What benefit do we get from moving things to a new file vs keeping it contained here?

mattkae:
> > I wonder if they'll behave on Linux as well. They run in CI and are a bit complex, so let me know if you need any help!
> 
> Almost. I've enabled the tests but disabled resize and fullscreen/maximize/minimize on Linux for now as setting these doesn't immediately return the same value when getting - we need to wait until the window has resized before this is the case. I assume that on win32 these are synchronous? Do we update the test to wait? Do we need more windowing API to do this consistently across platforms?

We could update the tests to wait, if possible. It probably wouldn't be too hard, but we could explore that in a follow-up PR if we don't want to muddy the waters of this one. It is preferable to share the integration tests as our north star though :star:

robert-ancell:
I've fixed all the known issues, and this is now ready for a full review.

We have a couple of quirks that I'm not sure if/how we can resolve.

1. When running on Wayland we have no way to know when we are iconified. We can enter and leave this state, but the `isMinimized` property will aways be false. I have disabled these tests for Linux. See https://gitlab.gnome.org/GNOME/gtk/-/issues/67. I figure the best we can do is just document this, anyone have suggestions the best place to do this?
2. Setting the window constraints sets the size of the surface that GTK is using for the window. This includes the window decorations and shadows. Therefore the amount allocated to the Flutter view is much less, in the case of a 500x500 constraint the Flutter area is 448x412. I adjusted the tests to check that the size was less than the constraint rather than the limit. I don't think there's any other methods in GTK3 to allow setting the maximum size of a widget that also stops the window from resizing. We could just limit the Flutter view and put padding around it. There doesn't seem to be any method of getting the size of the decorations for us to pad out the request. There also doesn't seem to be a method of making `contentSize` return the size of the surface so this is symmetrical. Not sure how to proceed here other than to document the limitations.

engine/src/flutter/shell/platform/linux/fl_window_monitor.h
```diff
@@ -0,0 +1,45 @@
...
+ * stacking.
+ * @on_state_changed: the function to call when the window state changes.
+ * @on_is_active_notify: the function to call when the is-active property
+ * changes.
+ * @on_close: the function to call when the user requests the window to be
+ * closed.
+ * @on_destroy: the function to call when the window is destroyed.
+ *
+ * Helper class to allow Dart to monitor a GtkWindow.
+ * Callbacks are called in the isolate this class was created with.
```
mattkae: Should you explicitly mention that this is only used in Dart over FFI?

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -0,0 +1,694 @@
...
+class WindowingOwnerLinux extends WindowingOwner {
+  /// Number of windows being managed by Flutter.
+  int _windowCount = 0;
+
+  /// Creates a new [WindowingOwnerLinux] instance.
+  ///
+  /// If [Platform.isLinux] is false, then this constructor will throw an
+  /// [UnsupportedError]
+  ///
+  ///
```
mattkae:
```suggestion
```

mattkae (changes requested):
Pretty clean! We should just add the integration tests to CI (if possible) and see if they pass.

To run the integration tests in CI, please add a devicelab runner similar to the one found at `devicelab/bin/tasks/windowing_test_windows.dart`.

Also, you will need to add it to `.ci.yaml`, e.g. https://github.com/flutter/flutter/blob/cd189a18da69490006591808cdbdaf6c2a8a7982/.ci.yaml#L7009

mattkae (approved):
Working well for me :rocket: You just have some analyze business to attend to, but I will let you look at that

PR #176202, opened 2025-09-29
Implement framework interface for the dialog window archetype

mattkae:
## What's new?
- :green_apple: Added the concept of dialogs to the windowing API. New symbols:
   - `DialogWindowControllerDelegate`
   - `DialogWindowController`
   - `WindowingOwner.createDialogWindowController`
   - `DialogWindow`
   - `WindowControllerCommon`
- :shallow_pan_of_food: Introduced the `WindowControllerCommon` mixin to share common windowing functionality between controller types
- I have NOT yet implemented dialogs. This pull request ONLY adds the idea to the API. Win32 implementation to follow :)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
+  ///
+  /// The [delegate] argument can be used to listen to the window's
+  /// lifecycle. For example, it can be used to save state before
+  /// a window is closed.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  factory DialogWindowController({
+    Size? preferredSize,
+    BoxConstraints? preferredConstraints,
+    FlutterView? parent,
```
mattkae: Should this be a `FlutterView`? Or perhaps it would be better off as a `WindowController` that they can get from the `WindowScope`?

mattkae: `BaseWindowController` is a much better idea here!

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -98,6 +98,87 @@ sealed class BaseWindowController extends ChangeNotifier {
...
 }
 
+/// A mixin that defines common behavior for window controllers.
+///
+/// See also:
+/// * [RegularWindowController], the controller that creates and manages regular windows.
+/// * [DialogWindowController], the controller that creates and manages dialog windows.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+mixin WindowControllerCommon {
```
justinmc: What's the distinction between this and BaseWindowController? I guess it would be possible to put all of this on there.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
 }
 
+/// Delegate class for dialog window controller.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [DialogWindowController], the controller that creates and manages dialog windows.
+///  * [DialogWindow], the widget for a dialog window.
```
justinmc, line 386: Also add RegularWindowControllerDelegate.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
+///
+/// The user of this class is responsible for managing the lifecycle of the window.
+/// When the window is no longer needed, the user should call [destroy] on this
+/// controller to release the resources associated with the window.
+///
+/// {@tool snippet}
+/// An example usage might look like:
+///
+/// ```dart
+/// // ignore_for_file: invalid_use_of_internal_member
```
justinmc, line 446: Don't forget to remove this when the feature flag is removed. Maybe put a TODO above this code block or something?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
+/// void main() {
+///   runWidget(
+///     RegularWindow(
+///       controller: RegularWindowController(
+///         preferredSize: const Size(800, 600),
+///         preferredConstraints: const BoxConstraints(minWidth: 640, minHeight: 480),
+///         title: 'Example Window',
+///       ),
+///       child: MaterialApp(home: DialogWindow(
+///           controller: DialogWindowController(
```
justinmc: It looks like the indentation is messed up here and below.

justinmc: Oh, or the `home` part should be on its own line.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
+  /// {@macro flutter.widgets.windowing.constraints}
+  ///
+  /// The [parent] argument specifies the parent window of this dialog.
+  ///
+  /// If the [parent] is null, then the dialog is modeless. Such dialogs can
+  /// be minimized but not maximized. They also have a disabled close button.
+  ///
+  /// If the [parent] is non-null, then the dialog is modal to the parent.
+  /// Such dialogs do not have a system menu. They are also not selectable
+  /// from the window switcher.
```
justinmc: Are they closed when their parent window is closed?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -372,6 +383,166 @@ abstract class RegularWindowController extends BaseWindowController {
...
+    return owner.createDialogWindowController(
+      delegate: delegate ?? DialogWindowControllerDelegate(),
+      preferredSize: preferredSize,
+      preferredConstraints: preferredConstraints,
+      title: title,
+      parent: parent,
+    );
+  }
+
+  @protected
```
justinmc: Extra `protected` here.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -148,6 +148,18 @@ class WindowingOwnerWin32 extends WindowingOwner {
...
+  @internal
+  @override
+  DialogWindowController createDialogWindowController({
+    required DialogWindowControllerDelegate delegate,
+    Size? preferredSize,
+    BoxConstraints? preferredConstraints,
+    FlutterView? parent,
+    String? title,
+  }) {
+    throw UnimplementedError('Dialog windows are not yet implemented on Windows.');
```
justinmc, line 160: Assuming it makes sense to you, could you add this kind of error message to the other UnimplementedErrors in this PR?

justinmc (commented):
Mostly small stuff in the comments, the only real thing is tests. We should probably test this even though it's incomplete. I guess we can't test anything related to the engine, but just writing unit tests for the framework code that's here should be fine.

packages/flutter/lib/src/widgets/_window.dart
mattkae: Actually - it shouldn't be possible to do that, so maybe `WindowControllerCommon` is a mistake. For example, tooltip-style windows will not have a title and cannot be minimized. This is me trying to share code where that is unnecessary :)

packages/flutter/lib/src/widgets/_window.dart
mattkae: Yes :+1:

packages/flutter/lib/src/widgets/_window_win32.dart
mattkae: This is the only `UnimplementedError` in this work right now

mattkae:
@justinmc Added tests and responded to your feedback :tada:

packages/flutter/lib/src/widgets/_window_win32.dart
justinmc: Ah nailed it then :dart:

packages/flutter/lib/src/widgets/_window.dart
justinmc: Sounds good! We can always revisit this kind of thing when everything else is added.

justinmc (approved):
LGTM 👍 

Thanks for doing all the work to add those monotonous tests 🙏 .

Issue #166423, opened 2025-04-02
Add first-class support for inspecting previews through the Widget Inspector


bkonyi:

bkonyi:
FYI @elliette @kenzieschmoll

elliette:
I believe this support was added in https://github.com/flutter/devtools/pull/8808. Let me know if there is anything missing here!

bkonyi:
While the widget inspector does work with the previewer, I think there's probably some polish items we'd want to take on to avoid showing details about the widget preview application itself (e.g., only widgets that are part of the preview, not the generated scaffolding).

elliette:
Are you referring to these widgets (which are included when implementation widgets are hidden):

<img width="422" alt="Image" src="https://github.com/user-attachments/assets/17a186ff-07a6-4788-93f4-0f9540992c88" />

Or the widget preview implementation widgets (which are only included when implementation widgets are shown):

<img width="399" alt="Image" src="https://github.com/user-attachments/assets/fe441e27-1648-4587-913e-43dda84efd46" />

PR #176218, opened 2025-09-29
[ Widget Preview ] Forward Widget Inspector navigation events via DTD

bkonyi:
With the assumption that IDEs will not create a debug session for the widget previewer, the widget inspector's source navigation functionality won't function as there's no IDE to listen to `navigate` events sent via `postEvent`.

This change overrides some widget inspector behavior to allow for navigating to source locations via the DTD Editor service instead of relying on the VM service's `postEvent`.

This change also makes some minor changes to the diagnostic properties and descriptions displayed within the inspector to display group and custom preview annotation details.

Towards https://github.com/flutter/flutter/issues/166423

### DIFF
--- a/packages/flutter_tools/lib/src/widget_preview/preview_details.dart
+++ b/packages/flutter_tools/lib/src/widget_preview/preview_details.dart
@@ -22,10 +22,10 @@ final class PreviewDetails {
   /// The file:// URI pointing to the script in which the preview is defined.
   final Uri scriptUri;
 
-  /// The line at which the Preview annotation was applied.
+  /// The 1-based line at which the Preview annotation was applied.
   final int line;
 
-  /// The column at which the Preview annotation was applied.
+  /// The 1-based column at which the Preview annotation was applied.
   final int column;
 
   /// The name of the package in which the preview was defined.
--- a/packages/flutter_tools/lib/src/widget_preview/preview_details.dart
+++ b/packages/flutter_tools/lib/src/widget_preview/preview_details.dart
@@ -10,6 +10,8 @@ typedef PreviewProperty = ({String key, DartObject object, bool isCallback});
 final class PreviewDetails {
   PreviewDetails({
     required this.scriptUri,
+    required this.line,
+    required this.column,
     required this.packageName,
     required this.functionName,
     required this.isBuilder,
@@ -19,6 +21,12 @@ final class PreviewDetails {
 
   /// The file:// URI pointing to the script in which the preview is defined.
   final Uri scriptUri;
+
+  /// The line at which the Preview annotation was applied.
+  final int line;
+
+  /// The column at which the Preview annotation was applied.
+  final int column;
 
   /// The name of the package in which the preview was defined.
   ///
--- a/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview.dart.tmpl
+++ b/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview.dart.tmpl
@@ -31,14 +31,32 @@ class WidgetPreview {
   const WidgetPreview({
     required this.builder,
     required this.scriptUri,
+    required this.line,
+    required this.column,
     required this.previewData,
     required this.packageName,
+  });
+
+  @visibleForTesting
+  const WidgetPreview.test({
+    required this.builder,
+    required this.previewData,
+    this.scriptUri = '',
+    this.line = -1,
+    this.column = -1,
+    this.packageName = '',
   });
 
   /// The absolute file:// URI pointing to the script containing this preview.
   ///
   /// This matches the URI format sent by IDEs for active location change events.
   final String scriptUri;
+
+  /// The line at which the Preview annotation was applied.
+  final int line;
+
+  /// The column at which the Preview annotation was applied.
+  final int column;
 
   /// The name of the package in which a preview was defined.
   ///
@@ -104,6 +122,7 @@ class WidgetPreview {
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     properties
       ..add(DiagnosticsProperty<String>('name', name, ifNull: 'not set'))
+      ..add(DiagnosticsProperty<String>('group', previewData.group))
       ..add(DiagnosticsProperty<Size>('size', size))
       ..add(DiagnosticsProperty<double>('textScaleFactor', textScaleFactor))
       ..add(DiagnosticsProperty<PreviewThemeData>('theme', theme))
--- a/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
+++ b/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
@@ -228,7 +228,9 @@ class _WidgetPreviewScaffoldInspectorService with WidgetInspectorService {
           kColumn: location.column,
         });
       }
-      dtdServices.navigateToCode(location!);
+      if (location != null) {
+        dtdServices.navigateToCode(location);
+      }
     }
     super.postEvent(eventKind, eventData, stream: stream);
   }
--- a/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
+++ b/packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
@@ -222,10 +222,10 @@ class _WidgetPreviewScaffoldInspectorService with WidgetInspectorService {
         // event data, just in case an IDE is attached and listening for
         // navigation events through the VM service.
         // TODO(bkonyi): determine if this is necessary
-        eventData.addAll({
+        eventData.addAll(<String, Object>{
           kFile: location.uri,
-          kLine: location.line,
-          kColumn: location.column,
+          kLine: location.line!,
+          kColumn: location.column!,
         });
       }
       if (location != null) {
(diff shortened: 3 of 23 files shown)


DanTup:
@bkonyi 

> With the assumption that IDEs will not create a debug session for the widget previewer

How sound is this assumption? (I asked because of the question below, and that we are already auto-attaching to the preview on Firebase Studio, so I don't think the idea of attaching to it is completely crazy - but I don't know what's already been thought about/discussed).


> the widget inspector's source navigation functionality won't function

Which widget inspector do we mean here? The one that's embedded in the sidebar in Dart-Code is already quite tightly tied to a debug session, so if we're planning to connect that to the preview (and not have the preview be a debug session) then we might need to change some of these assumptions anyway.

(that's not to say it's not a good idea to switch this to DTD anyway, but it would be useful for me to better understand the end goals here)

**Edit:** To capture some offline discussion:

> With the assumption that IDEs will not create a debug session for the widget previewer

We can't have a full debug session against the widget preview, because there's no ability to use the Chrome debug port or Dart debug extension, so actually it probably does not make sense to try and connect up a normal IDE debug session here.

packages/flutter_tools/lib/src/widget_preview/preview_details.dart
```diff
@@ -20,6 +22,12 @@ final class PreviewDetails {
   /// The file:// URI pointing to the script in which the preview is defined.
   final Uri scriptUri;
 
+  /// The line at which the Preview annotation was applied.
+  final int line;
```
DanTup: nit: in docs for line/char fields, can we include whether these are 1-based or 0-based. All of the tools and specs are not consistent so it's useful for the intentions to be documented if we need to track down off-by-one bugs (which have definitely not happened before 😄)

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview.dart.tmpl
```diff
@@ -31,15 +31,33 @@ class WidgetPreview {
...
     required this.packageName,
   });
 
+  @visibleForTesting
+  const WidgetPreview.test({
+    required this.builder,
+    required this.previewData,
+    this.scriptUri = '',
+    this.line = -1,
+    this.column = -1,
```
DanTup, line 46: Any reason we can't use `null` instead? We've had bugs in server/LSP before because it wasn't obvious from the types that we might get a -1 and they were used in bad maths - whereas `null`s are harder to handle accidentally.

(if these are strictly only for tests and wouldn't show up elsewhere, then it might not be worth it?)

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
```diff
@@ -21,7 +24,11 @@ class WidgetPreviewScaffoldController {
     required PreviewsCallback previews,
     @visibleForTesting WidgetPreviewScaffoldDtdServices? dtdServicesOverride,
   }) : _previews = previews,
-       dtdServices = dtdServicesOverride ?? WidgetPreviewScaffoldDtdServices();
+       dtdServices = dtdServicesOverride ?? WidgetPreviewScaffoldDtdServices() {
+    // Overrides the default WidgetInspectorService instance to handle selection
+    // events.
+    _WidgetPreviewScaffoldInspectorService(dtdServices: dtdServices);
```
DanTup, line 30: Not using the instance returned here looks odd - would it be better to extract the body, so it's this instead?:

```dart
WidgetInspectorService.instance = _WidgetPreviewScaffoldInspectorService(dtdServices: dtdServices);
```

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview.dart
```diff
@@ -31,15 +31,33 @@ class WidgetPreview {
   const WidgetPreview({
     required this.builder,
     required this.scriptUri,
+    required this.line,
+    required this.column,
     required this.previewData,
     required this.packageName,
   });
 
+  @visibleForTesting
```
DanTup, line 40: The two copies of these files confuse me - is there any way the test could cause this to be generated so a) they can't be out of sync and b) simplify reviews?

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/filter_by_selected_file_test.dart
```diff
@@ -192,12 +190,16 @@ void main() {
       WidgetPreview(
         builder: () => Text('widget1'),
         scriptUri: '',
+        line: -1,
+        column: -1,
```
DanTup: Why don't these ones use `WidgetPreview.test`?

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
```diff
@@ -154,3 +161,77 @@ class WidgetPreviewScaffoldController {
...
+    // the navigation target in `postEvent`.
+    if (object is PreviewWidgetElement) {
+      final previewData = (object.widget as PreviewWidget).preview;
+      _nextNavigationLocation = CodeLocation(
+        uri: previewData.scriptUri,
+        line: previewData.line,
+        column: previewData.column,
+      );
+    }
+    final result = super.setSelection(object, groupName);
```
DanTup, line 197: This seems a little awkward - would it be simpler to skip this call and just directly call `super.postEvent()` with this location instead? Or if we still need the other base code (setting the selection), what about making `setSelection` take an optional `Location`?

DanTup (approved):
I'm not terribly familiar with the widget inspector service code so I'll let you choose whether to wait for Elliot's review, but LGTM. I did add some minor nits, but nothing I think is necessary.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview.dart.tmpl
bkonyi: Yup, this is only for tests and should never show up anywhere else!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview.dart
bkonyi: Yeah, this is an unfortunate but necessary duplication since the templates need to exist in the `flutter_tools/templates` directory but we also need to be able to test the template contents. If we don't have a committed hydrated template, we can't write tests against it unless those tests are also generated at runtime, which means we lose static analysis, making writing these tests harder.

We do have a check in `flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/template_change_detection_smoke_test.dart` that verifies these copies don't diverge.

In general, feel free to ignore the `*.tmpl` files during review as their changes are identical to those made in the `widget_previews_scaffold` test project.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/filter_by_selected_file_test.dart
bkonyi: Because I missed them 😅 Done.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview.dart
DanTup: > We do have a check in `flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/template_change_detection_smoke_test.dart`

Ah, I missed that - thanks! :)

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart
bkonyi: It's definitely awkward, but `super.setSelection` needs to be called as there's a lot of other state that's updated as the result of a selection.

This code is called via a service extension, and DevTools doesn't know that there's anything special about the `PreviewWidget`, so we need to keep this logic contained within the previewer.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_scaffold_controller.dart.tmpl
bkonyi: That's technically not allowed as `WidgetInspectorService.instance` is protected:

`The member 'instance' can only be used within instance members of subclasses of 'WidgetInspectorService'`

packages/flutter_tools/lib/src/widget_preview/preview_details.dart
bkonyi: Done!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/custom_previews_test.dart
```diff
@@ -71,6 +71,8 @@ WidgetPreviewerWidgetScaffolding previewsForCustomMultiPreview() {
     scriptUri: '',
     preview: BrightnessPreview(name: 'MyPreview'),
     previewFunction: () => Text('Foo'),
+    line: -1,
+    column: -1,
```
DanTup, line 75: @bkonyi nit: there are a couple more of these 😄

PR #176230, opened 2025-09-29
Add SwiftUI support for UIScene migration

vashworth:
When embedding Flutter in a SwiftUI app, it appears that SwiftUI wraps the scene delegate given by the developer in an internal class called `SwiftUI.AppSceneDelegate`. This causes the scene delegate to fail to check if it conforms to the `FlutterSceneLifeCycleProvider` protocol - even though it does. However, after force casting it, selectors respond and can be used.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -252,6 +252,32 @@ - (BOOL)windowScene:(UIWindowScene*)windowScene
...
+
++ (FlutterPluginSceneLifeCycleDelegate*)fromScene:(UIScene*)scene {
+  if ([scene.delegate conformsToProtocol:@protocol(FlutterSceneLifeCycleProvider)]) {
+    NSObject<FlutterSceneLifeCycleProvider>* sceneProvider =
+        (NSObject<FlutterSceneLifeCycleProvider>*)scene.delegate;
+    return sceneProvider.sceneLifeCycleDelegate;
+  }
+
+  // When embedded in a SwiftUI app, the scene delegate does not conform to
+  // FlutterSceneLifeCycleProvider even if it does. However, after force casting it,
```
LongCatIsLooong, line 264: It looks like `AppSceneDelegate` forwards selectors to the wrapped scene delegate:

```
    frame #0: 0x0000000104f1ddd4 SwiftUIPlayground.debug.dylib`SwiftUIPlayground.MySceneDelegate.responds(aSelector="wowie") -> Swift.Bool at MySceneDelegate.swift:14:12
  * frame #2: 0x00000001d892e25c SwiftUI`AppSceneDelegate.responds(to:) + 128
    frame #3: 0x00000001d892e2cc SwiftUI`@objc AppSceneDelegate.responds(to:) + 32
    frame #4: 0x0000000104f1d7f8
```

In that case the class name check doesn't seem to be necessary? I think all we have to do is to do the `respondsToSelector` check and dynamic return type check.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -286,11 +312,10 @@ - (BOOL)scene:(UIScene*)scene
   for (NSObject<FlutterSceneLifeCycleDelegate>* delegate in _delegates.allObjects) {
     if ([delegate respondsToSelector:_cmd]) {
       // If this event has already been consumed by a plugin, send the event with nil options.
-      if (consumedByPlugin) {
-        [delegate scene:scene willConnectToSession:session options:nil];
-        continue;
-      } else if ([delegate scene:scene willConnectToSession:session options:connectionOptions]) {
-        // Only allow one plugin to process this event.
+      // Only allow one plugin to process the connection options.
```
LongCatIsLooong, line 311: Ah I missed this part in the previous PR. But why do we have this limitation that only one plugin gets to handle this event?

LongCatIsLooong (approved):
The engine changes LGTM.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Oops I was going to add a comment in the PR and then forgot. This was some last minute [nits from Stuart on the other PR](https://github.com/flutter/flutter/pull/175866#pullrequestreview-3280895529), but I wanted to land it because all my other PRs build off it

Basically, `connectionOptions` are use to restore state or open to another page (like deeplinking). We only want one plugin to take action. For example, if I have two plugins that implement deeplinking and both were to handle the event, it would be a bad user experience because one plugin would open to a page and then another plugin would open to a page causing two page openings. Does that make sense?

Stuart thought we should still send the `scene:willConnectToSession:options` event to plugins, though, with `nil` options so they can use it to track state.

PR #176235, opened 2025-09-29
Add structured warning event for slow wireless debugging on iOS 26+ devices

### LINKED ISSUE #176206: Wish for the flutter tool to help the user switch to USB instead of WiFi when running/debugging against iOS 26
### Use case

Due to architectural changes from iOS 26, running/debugging a Flutter app during development against an iOS 26 iPhone connected to the host development machine via WiFi can be quite slow. We've noticed that the latency is felt most when the target device is connected over WiFi.

### Proposal

We propose to detect when people are using iOS 26 devices connected via WiFi, and suggest to them to instead use a wired (USB) connection.

We propose to deliver this feedback via the flutter tool, and our IDE plugins. But, this issue specifically is for the flutter tool.

okorohelijah:
This PR improves the developer experience when running or debugging Flutter apps on iOS 26+ devices connected via WiFi. The tool now emits a structured `app.warning` event (in addition to a console warning) when it detects wireless debugging on iOS 26 or later. This allows IDEs and other tools using `--machine` mode to display a prominent warning to users, encouraging them to use a wired (USB) connection for better performance.

## Details

- Detects when an iOS device is:
  - Running iOS 26 or later,
  - Connected wirelessly,
  - Debugging is enabled.
- Emits a human-readable warning to the console.
- Emits a structured `app.warning` event with relevant metadata (category, deviceId, iosVersion, suggestion, etc.) for IDEs and tools using `--machine` mode.
- No changes to existing workflows for users not matching these conditions.

Fixes #176206 

## Pre-launch Checklist

- [X] I signed the [CLA].
- [] I updated/added relevant documentation (doc comments with `///`).
- [] I added new tests to check the change I am making, or this PR is [test-exempt].
- [] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.


packages/flutter_tools/lib/src/ios/devices.dart
```diff
@@ -494,6 +494,26 @@ class IOSDevice extends Device {
...
+
+      _logger.sendEvent('app.warning', <String, Object?>{
+        'warning': warningMessage,
+        'category': 'ios-wireless-performance',
+        'deviceId': id,
+        'iosVersion': sdkVersion!.major,
+        'actionable': true,
+        'suggestion': 'Use a wired (USB) connection for better performance',
+      });
+    }
```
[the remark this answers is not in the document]
okorohelijah: done.

packages/flutter_tools/lib/src/ios/devices.dart
```diff
@@ -494,6 +494,28 @@ class IOSDevice extends Device {
...
+      _logger.printWarning(warningMessage);
+
+      _logger.sendEvent('app.warning', <String, Object?>{
+        'warning': warningMessage,
+        'category': 'ios-wireless-performance',
+        'deviceId': id,
+        'iosVersion': sdkVersion!.major,
+        'actionable': true,
+        'suggestion': 'Use a wired (USB) connection for better performance',
+      });
```
DanTup: I think we should document this in https://github.com/flutter/flutter/blob/master/packages/flutter_tools/doc/daemon.md since editors will need to handle it specifically and we should have something more formal than just what's in this code.

Some additional thoughts/qs:

- What does `actionable` mean?
- How should `warning` and `suggestion` be combined if an editor can only show one block of text with no formatting (from the current values above, I think we can join them with a space, but we should ensure this is clear, since the `suggestion` doesn't end with a full stop like `warning` does)
- Should we add an (optional) `url` field that could link to somewhere with more info? Even if we don't have it for this, if we're going to add some new code to the editors, adding this now would allow it to be used for other warnings in future

okorohelijah: @DanTup I've updated `daemon.md` to document the new app.warning event and explain what each of its parameters, including actionable, is for.

Regarding combining the warning and suggestion, I agree we should make it as simple as possible for editors. I've updated the implementation to merge the suggestion directly into the main warning message. This makes the warning self-explanatory and removes the need for a separate suggestion field

packages/flutter_tools/doc/daemon.md
```diff
@@ -158,6 +158,17 @@ This is sent once the application launch process is complete and the app is eith
 
 This is sent when output is logged for a running application. The `params` field will be a map with the fields `appId` and `log`. The `log` field is a string with the output text. If the output indicates an error, an `error` boolean field will be present, and set to `true`.
 
+#### app.warning
+
+This is sent when a warning is issued for a running application that an editor may wish to handle specifically. The `params` field will be a map with the following fields:
+
+- `warning`: A string containing the warning message to be displayed to the user.
+- `category`: A string used to categorize the warning (for example, `ios-wireless-performance`).
```
DanTup, line 167: What might an editor do with this string?

I was thinking it might be useful for the editor to use if they want to show a "Don't warn me again" prompt and then store this as something to skip showing in future. However, for this I think a unique identifier per-message would be better (and the name `category` doesn't make it clear if this is unique).

packages/flutter_tools/doc/daemon.md
```diff
@@ -158,6 +158,17 @@ This is sent once the application launch process is complete and the app is eith
...
+- `deviceId`: The ID of the device this warning is relevant to.
+- `iosVersion`: The iOS versionn of the device
```
DanTup: Although the current warning is for iOS, presumably in future we might use this for something else - so should we change this to something more generic (`deviceOsVersion`?) or just drop it? I'm not sure how useful it is to a client that's using this event generically.

(also nit: missing full stop at the end that the other points have)

packages/flutter_tools/doc/daemon.md
```diff
@@ -158,6 +158,17 @@ This is sent once the application launch process is complete and the app is eith
...
+- `actionable`: A boolean indicating if the warning includes a suggested action for the user.'
```
DanTup:
```suggestion
- `actionable`: A boolean indicating if the warning includes a suggested action for the user.
```

packages/flutter_tools/lib/src/ios/devices.dart
```diff
@@ -494,6 +494,27 @@ class IOSDevice extends Device {
...
+          'Wireless debugging on iOS ${sdkVersion!.major} may be slower than expected. '
+          'For better performance, consider using a wired (USB) connection.';
+
+      _logger.printWarning(warningMessage);
+
+      _logger.sendEvent('app.warning', <String, Object?>{
+        'warning': warningMessage,
+        'category': 'ios-wireless-performance',
+        'deviceId': id,
+        'iosVersion': sdkVersion!.major,
```
DanTup: nit: there are quite a few `!`s in the code above, maybe assign `sdkVersion` to a local variable to avoid them all?

DanTup (approved):
Added some minor comments/nits, but otherwise lgtm - thanks! 🙂

packages/flutter_tools/doc/daemon.md
okorohelijah: I think using `deviceOsVersion` is better than removing because it gives the IDE more flexibility to provide a structured data. and in a scenario where lets say the warning is more critical in some versions than some others, then the ide could maybe display a red icon instead of yellow to show severity

packages/flutter_tools/lib/src/ios/devices.dart
okorohelijah: done

packages/flutter_tools/doc/daemon.md
okorohelijah: having an id for a specific warning which the IDE can ignore based on users' choice and category which could include different warningIds make sense!

packages/flutter_tools/doc/daemon.md
okorohelijah: done

packages/flutter_tools/doc/daemon.md
DanTup: > in a scenario where lets say the warning is more critical in some versions than some others, then the ide could maybe display a red icon instead of yellow to show severity

I don't think the IDE should need to know about these versions though - if we want something to be higher priority, it should be encoded into the JSON so it can be handled generically (eg. a severity field).

I don't see any downside to having the OS version too though - if nothing else it could be useful when reviewing logs if this warning is being reported.

packages/flutter_tools/lib/src/ios/devices.dart
```diff
@@ -494,6 +494,29 @@ class IOSDevice extends Device {
...
+
+      _logger.sendEvent('app.warning', <String, Object?>{
+        'warningId': 'ios-wireless-slow',
+        'warning': warningMessage,
+        'category': 'ios-wireless-performance',
+        'deviceId': id,
+        'deviceOsVersion': sdkVersion.major,
+        'actionable': true,
+      });
+    }
```
vashworth: This needs unit tests. See packages/flutter_tools/test/general.shard/ios/ios_device_start_prebuilt_test.dart

okorohelijah: done

PR #176240, opened 2025-09-29
Add an AppDelegate callback for implicit FlutterEngines

### LINKED ISSUE #173357: [UIScene] Create a "plugin" registrar that can be used for creating app-level method channels and platform views
## Problem 

Due to changes in initialization order, accessing the FlutterViewController from the application(_:didFinishLaunchingWithOptions:) method no longer works and will result in a crash (after migrating to UIScene). 

Flutter’s official documentation was to [set up app-level method channels in application:didFinishLaunching:](https://docs.flutter.dev/platform-integration/platform-channels#step-4-add-an-ios-platform-specific-implementation).

```
- (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
  FlutterViewController* controller = (FlutterViewController*)self.window.rootViewController;

  FlutterMethodChannel* batteryChannel = [FlutterMethodChannel
                                    methodChannelWithName:@"samples.flutter.dev/battery"
                                          binaryMessenger:controller.binaryMessenger];
```

To mitigate this, https://github.com/flutter/flutter/pull/169399 introduced a callback called `registerWithRegistry` that can be used in a iOS project's AppDelegate. This callback returns an object that conforms to `FlutterPluginRegistry`. The `FlutterAppDelegate`, `FlutterEngine`, and `FlutterViewController` all conform to this protocol. 

```
- (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
  self.pluginRegistrant = self;
  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}

- (void)registerWithRegistry:(NSObject<FlutterPluginRegistry>*)registry {
  [GeneratedPluginRegistrant registerWithRegistry:registry];
}
```

However, to create a method channel you need a `binaryMessenger`.

## Proposal

Introduce a new base protocol, called `FlutterBaseRegistrar`, make [`FlutterPluginRegistrar`](https://github.com/flutter/flutter/blob/96933bf400e4a927facab702a74129ce4ae17065/engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPlugin.h#L283-L380) inherit from it, and move most methods up to the base.

Plugin specific methods can stay in `FlutterPluginRegistrar`, such as:
* addApplicationDelegate:, as application authors can do this directly instead.
* lookupKeyForAsset:*, as it’s not clear that this has a use case in the app context. (It could easily be moved up later if there is a use case).
* publish:, which similarly does not seem likely to be useful in the app context.

Introduce a new sibling protocol to `FlutterPluginRegistrar` called `FlutterEngineRegistrar`, which inherits from the same base. For now this would have no new methods, but by adding this intermediate layer we can trivially add app-specific methods later if needed, without breaking changes or new delegate methods.

Then add a property to `FlutterPluginRegistry`:

```
- (nonnull NSObject<FlutterEngineRegistrar>*)registrarForEngine;
```

The implementation of `registrarForEngine` would create a `FlutterEngineRegistrar` with a unique identifier and register it to the engine: 
https://github.com/flutter/flutter/blob/96933bf400e4a927facab702a74129ce4ae17065/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm#L1310-L1317

Then from a developer's `AppDelegate`, they can do the following:

```
- (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
  self.pluginRegistrant = self;
  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}

- (void)registerWithRegistry:(NSObject<FlutterPluginRegistry>*)registry {
  NSObject<FlutterEngineRegistrar>* registrar = registry.registrarForEngine;
  FlutterMethodChannel* batteryChannel = [FlutterMethodChannel
                                    methodChannelWithName:@"samples.flutter.dev/battery"
                                          binaryMessenger:registrar.messenger];
  
  [GeneratedPluginRegistrant registerWithRegistry:registry];
}
```

vashworth:
This PR introduces a new protocol `FlutterImplicitEngineDelegate`, which adds a callback called `didInitializeImplicitFlutterEngine`, which returns a new interface called `FlutterImplicitEngineBridge`.

This was added to expose parts of an implicit `FlutterEngine`, such as when created from a `FlutterViewController` in a storyboard.

``` swift diff
@objc class AppDelegate: FlutterAppDelegate, FlutterImplicitEngineDelegate {
  override func application(
    _ application: UIApplication,
    didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
  ) -> Bool {
    return super.application(application, didFinishLaunchingWithOptions: launchOptions)
  }
}

func didInitializeImplicitFlutterEngine(_ engineBridge: FlutterImplicitEngineBridge) {
    // Register plugins with `engineBridge.pluginRegistry`
    GeneratedPluginRegistrant.register(with: engineBridge.pluginRegistry)

    // Create method channels with `engineBridge.applicationRegistrar.messenger()`
    let batteryChannel = FlutterMethodChannel(
      name: "samples.flutter.dev/battery",
      binaryMessenger: engineBridge.applicationRegistrar.messenger()
    )
    ...

    // Create platform views with `engineBridge.applicationRegistrar.messenger()`
    let factory = FLNativeViewFactory(messenger: engineBridge.applicationRegistrar.messenger())
    ...
}
```

This PR also refactors `FlutterPluginRegistrar` into 3 separate protocols: 
* `FlutterBaseRegistrar`
* `FlutterApplicationRegistrar`
* `FlutterPluginRegistrar`

Most methods are moved from `FlutterPluginRegistrar` to `FlutterBaseRegistrar` and then `FlutterPluginRegistrar` and `FlutterApplicationRegistrar` inherit from `FlutterBaseRegistrar`. `FlutterPluginRegistrar` also has additional methods specific to it - there are no meaningful/breaking changes to `FlutterPluginRegistrar`.

Fixes https://github.com/flutter/flutter/issues/173357.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine_Internal.h
```diff
@@ -107,6 +107,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)addSceneLifeCycleDelegate:(NSObject<FlutterSceneLifeCycleDelegate>*)delegate;
 
+- (BOOL)notifyAppDelegateOfEngineInitialization;
```
stuartmorgan-g: Could you add a declaration comment here, documenting what this does and what the return value means?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -115,8 +118,36 @@ - (void)addDelegate:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
...
+    return;
+  }
+  // Send nil launchOptions since UIKit sends nil when UIScene is enabled.
+  [self application:application willFinishLaunchingWithOptions:nil];
+}
+
 - (BOOL)application:(UIApplication*)application
     didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
+  if (_delegates.count > 0) {
+    self.didReceiveApplicationDidLaunch = YES;
```
stuartmorgan-g: Should the properties be called `didSend...` (or maybe `didForward...`) instead of `didReceive...` since they can be NO even when the notification has been received?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate_internal.h
```diff
@@ -19,6 +19,22 @@
  */
 - (BOOL)hasPluginThatRespondsToSelector:(SEL)selector;
 
+/**
+ * Forwards the `application:didFinishLaunchingWithOptions:` lifecycle event to plugins if they have
+ * not received it yet. This compensates for the UIScene migration, which causes storyboards (and
+ * thus the plugin registration via `FlutterImplicitEngineDelegate`) to be instantiated after the
+ * application launch events.
+ */
+- (void)applicationDidFinishLaunchingSceneFallback:(UIApplication*)application;
```
stuartmorgan-g: It's kind of gross, but technically this should be called something like `sceneFallbackDidFinishLaunchingApplication:`, since the "Application" part should be right before the `:`. (Same below.)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -78,6 +79,7 @@ @interface FlutterViewController () <FlutterBinaryMessenger, UIScrollViewDelegat
 @property(nonatomic, assign) UIStatusBarStyle statusBarStyle;
 @property(nonatomic, assign) BOOL initialized;
 @property(nonatomic, assign) BOOL engineNeedsLaunch;
+@property(nonatomic, assign) BOOL wokenFromNib;
```
stuartmorgan-g: Pedantry nit: since it's awakeFromNib rather than wakeFromNib, this should be `awokenFromNib`.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,30 @@ FLUTTER_DARWIN_EXPORT
...
 
+/**
+ * Exposes parts of a `FlutterEngine` for registration purposes.
+ *
+ * This is used when the engine is created implicitly to allow registering
+ * plugins, application-level method channels, platform views, etc.
+ */
+@interface FlutterImplicitEngineBridge : NSObject
+
+@property(nonatomic, strong) NSObject<FlutterPluginRegistry>* pluginRegistry;
```
stuartmorgan-g: These should have declaration comments.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,30 @@ FLUTTER_DARWIN_EXPORT
...
+
+/**
+ * Protocol for receiving a callback when an implicit engine is initialized.
+ *
+ * This provides the engine bridge to the listener.
+ */
+@protocol FlutterImplicitEngineDelegate
+@required
+
+- (void)didInitializeImplicitFlutterEngine:(FlutterImplicitEngineBridge*)engineBridge;
```
stuartmorgan-g: This should have a declaration comment explaining when it's called/what it's for.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,30 @@ FLUTTER_DARWIN_EXPORT
...
+ */
+@interface FlutterImplicitEngineBridge : NSObject
+
+@property(nonatomic, strong) NSObject<FlutterPluginRegistry>* pluginRegistry;
+@property(nonatomic, strong) NSObject<FlutterApplicationRegistrar>* applicationRegistrar;
+
+@end
+
+/**
+ * Protocol for receiving a callback when an implicit engine is initialized.
```
stuartmorgan-g: Maybe expand more on what an implicit engine is here? I'm not sure the average Flutter developer (even restricting to the subset who are interacting with the native app code) will understand what an implicit engine is.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
stuartmorgan-g, line 156: [This protocol ](https://github.com/vashworth/flutter/blob/842cec336c19a9b587a716a674f87f37bb205b8f/engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPlugin.h#L34-L43) (incorrectly?) defines these as non-nullable, and changing that would be a breaking change, but your PR allows `launchOptions` to be nil here, so this is a nullability violation. (Yay for terrible nullability warnings in Obj-C 😐).

I think the best option we have here is to pass `(launchOptions ?: @{})`.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -153,6 +166,16 @@ @interface FlutterEngine () <FlutterIndirectScribbleDelegate,
 
 @end
 
+@implementation FlutterImplicitEngineBridge
+- (instancetype)initWithEngine:(FlutterEngine*)engine {
+  self = [super init];
+  _pluginRegistry = engine;
+  _applicationRegistrar =
+      [engine registrarForApplication:@"io.flutter.flutter.application_registrar"];
```
stuartmorgan-g: Why the double-`flutter`?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1549,12 +1594,42 @@ - (instancetype)initWithPlugin:(NSString*)pluginKey flutterEngine:(FlutterEngine
...
+
+@implementation FlutterEnginePluginRegistrar {
+  NSString* _pluginKey;
+}
+
+- (instancetype)initWithPlugin:(NSString*)pluginKey flutterEngine:(FlutterEngine*)flutterEngine {
+  self = [super init];
+  NSAssert(self, @"Super init cannot be nil");
+  _pluginKey = [pluginKey copy];
+  self.flutterEngine = flutterEngine;
```
stuartmorgan-g: `_flutterEngine`, since this is an `init` method.

stuartmorgan-g (changes requested):

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: I'm assuming I copied from here:
https://github.com/flutter/flutter/blob/17e2120be8a9a5edf110c5c9d426c57a5f18fd14/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterDartProject.mm#L415

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: So `flutterEngine` is inside the parent class so when I try to access `_flutterEngine`, it errors

```
./engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm:1623:3: error: instance variable '_flutterEngine' is private
```

stuartmorgan-g: Oh, I missed that this is a subclass. It looks like you should be using `[super initWithKey:pluginKey flutterEngine:flutterEngine]` then.

Also, it looks like both layers are keeping a copy of the key, which could be avoided by making the key accessible to subclasses.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
vashworth: Why would making them nullable be a breaking change?

stuartmorgan-g: `FlutterApplicationLifeCycleDelegate` is for plugins to implement, and the contract is that the parameter is non-nullable. If they are relying on our annotations being correct and assumed that the value was—as we claimed—non-nullable, then code (already published, in the wild in the package ecosystem) using it could fail in a variety of ways if we start passing nil.

And if the plugin implementing it is written in Swift, presumably the sudden mismatch in nullability would be a compile-time failure.

vashworth: So the interface that would need to be made nullable is actually `FlutterPluginAppLifeCycleDelegate`, which is used for add to app:

https://docs.flutter.dev/add-to-app/ios/add-flutter-screen#if-you-cant-directly-make-flutterappdelegate-a-subclass

It's an object they create and call, so allowing them to now pass in `nil` doesn't seem like it would break?

```swift
private let lifecycleDelegate = FlutterPluginAppLifeCycleDelegate()

  let flutterEngine = FlutterEngine(name: "my flutter engine")

  func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
    flutterEngine.run()
    return lifecycleDelegate.application(application, didFinishLaunchingWithOptions: launchOptions ?? [:])
  }
```

https://github.com/flutter/flutter/pull/176240/files#diff-74aad43841c846d7733e55a3a036423605a4797e4818481649ff2fd98fd66e42R32

vashworth: Oh wait, it then goes to `FlutterApplicationLifeCycleDelegate`, got it. Nvm

stuartmorgan-g: Yeah, I had to check the names very carefully several times to make sure this was doing what I thought it was doing.

It's annoying, because in Swift or Dart it just wouldn't have compiled, making it much easier to reason about even with the confusing names.

vashworth: Swapped out nil for `@{}`

https://github.com/flutter/flutter/pull/176240/files#diff-41c25fbdc646eededa4c404895bf4325923e1985ffef00ebe739af0b3f071af3R128

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1578,12 +1623,38 @@ - (instancetype)initWithPlugin:(NSString*)pluginKey flutterEngine:(FlutterEngine
...
+}
+
+@end
+
+@implementation FlutterEnginePluginRegistrar
+
+- (instancetype)initWithPlugin:(NSString*)pluginKey flutterEngine:(FlutterEngine*)flutterEngine {
+  self = [super init];
+  NSAssert(self, @"Super init cannot be nil");
+  return [super initWithKey:pluginKey flutterEngine:flutterEngine];
```
stuartmorgan-g: This is calling init twice now, which is bad news.

It looks like this no longer needs to do anything the superclass doesn't already do, so you can remove the whole override and just inherit the superclass initializer.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -89,10 +89,25 @@ static void IOSPlatformThreadConfigSetter(const fml::Thread::ThreadConfig& confi
 
 NSString* const kFlutterKeyDataChannel = @"flutter/keydata";
 static constexpr int kNumProfilerSamplesPerSec = 5;
+NSString* const kFlutterApplicationRegistrarKey = @"io.flutter.flutter.application_registrar";
+
+@interface FlutterEngineBaseRegistrar : NSObject <FlutterBaseRegistrar>
 
-@interface FlutterEngineRegistrar : NSObject <FlutterPluginRegistrar>
 @property(nonatomic, weak) FlutterEngine* flutterEngine;
+@property(nonatomic, strong) NSString* key;
```
stuartmorgan-g: `readonly`?

Also, just for consistency `copy` would be better than `strong` (even though it doesn't actually matter if you make it `readonly`).

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,49 @@ FLUTTER_DARWIN_EXPORT
...
+ * plugins, application-level method channels, platform views, etc.
+ */
+@interface FlutterImplicitEngineBridge : NSObject
+
+/**
+ * The `FlutterPluginRegistry` for the created `FlutterEngine`.
+ *
+ * This can be used to vend `FlutterPluginRegistrar`s for plugins.
+ */
+@property(nonatomic, strong) NSObject<FlutterPluginRegistry>* pluginRegistry;
```
stuartmorgan-g: I missed this before; both of these should be `readonly`.

stuartmorgan-g: You may want to make this `weak` instead? Clients may not realize that keeping the bridge around would keep the whole engine alive.

It doesn't seem like it should be surprising that it would, so it's not critical, but if we can easily make it harder for people to "leak" something as big as the engine it wouldn't hurt. (Alternately if it becomes a problem later we could make this a weak bridge object under the hood like we've done for things like the binary messenger.)

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,49 @@ FLUTTER_DARWIN_EXPORT
...
+
+/**
+ * The `FlutterApplicationRegistrar` for the created `FlutterEngine`.
+ *
+ * This registrar provides access to application-level services, such as the engine's
+ * `FlutterBinaryMessenger` or `FlutterTextureRegistry`.
+ */
+@property(nonatomic, strong) NSObject<FlutterApplicationRegistrar>* applicationRegistrar;
```
stuartmorgan-g: `readonly`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -294,19 +297,44 @@ - (void)sharedSetupWithProject:(nullable FlutterDartProject*)project
...
+  // is only needed when the FlutterEngine is implicit. If it's not implicit there's no need for
+  // them to have a callback to expose the engine since they created the FlutterEngine directly.
+  BOOL notified = [_engine performAppDelegateEngineInitializationCallback];
 
+  // TODO(vashworth): Deprecate, see https://github.com/flutter/flutter/issues/176424
   if ([FlutterSharedApplication.application.delegate
           respondsToSelector:@selector(pluginRegistrant)]) {
     NSObject<FlutterPluginRegistrant>* pluginRegistrant =
         [FlutterSharedApplication.application.delegate performSelector:@selector(pluginRegistrant)];
     [pluginRegistrant registerWithRegistry:self];
```
stuartmorgan-g, line 312: I know this is existing code, but the context and my comment above highlighted it for me, and it's *really* dangerous since it's giving out external `self` reference before initialization is complete. Can we just change this to `registerWithRegistry:engine`?

(If you want to punt on this that's fine, but since it's related and I think it's probably an easy fix I wanted to flag it.)

stuartmorgan-g (changes requested):

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
vashworth: If we make it weak, don't you have to force unwrap it in swift?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: 🤦‍♀️

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
stuartmorgan-g: Oh, good point. Let's leave it as-is then.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
stuartmorgan-g: Having thought about this more, we shouldn't change this here given the timing; if it causes some subtle problem we don't want that to impact getting this stuff landed. I'll make a note to look at this soon myself.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -447,6 +447,49 @@ FLUTTER_DARWIN_EXPORT
 
 @end
 
+/**
+ * Exposes parts of a `FlutterEngine` for registration purposes.
+ *
+ * This is used when the engine is created implicitly to allow registering
+ * plugins, application-level method channels, platform views, etc.
+ */
+@interface FlutterImplicitEngineBridge : NSObject
```
stuartmorgan-g: Sorry, one final small change request, since I've just been in the world of trying to replace OCMock with fakes: let's make this a protocol instead of a class, and then have a private class (or just the engine itself) implement it. That makes calling it from tests much easier.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -294,19 +297,44 @@ - (void)sharedSetupWithProject:(nullable FlutterDartProject*)project
                                                 opaque:_viewOpaque
                                        enableWideGamut:engine.project.isWideGamutEnabled];
   [_engine createShell:nil libraryURI:nil initialRoute:initialRoute];
+  [self performImplicitEngineCallbacks];  // This is the earliest this can be called because it
```
stuartmorgan-g: I missed the context here before; since this is an `init`, it's dangerous to call `self`. Flutter engine code is terrible about this (see the TODO a few lines below), but we shouldn't make it worse. It would be better to either inline all that code, or make it a static C function, to avoid accidental obfuscated calls to `self`.

stuartmorgan-g: Argh, I just discovered that I left this comment in draft so I didn't go out yesterday.

stuartmorgan-g (approved):
LGTM with some minor final suggestions. They should be straightforward changes, so approving now to avoid an extra review cycle.

Issue #171231, opened 2025-06-26
[Google3 Bug]: MenuItemButton voice over/semantics doesn't pronounce disabled buttons


rezgoogle:
b/426553069

### Help us understand the severity of this issue

- [ ] causing severe production issues e.g. malfunctions or data loss
- [ ] blocking next binary release
- [x] blocking a client feature launch within a quarter
- [ ] nice-to-have but does not block a launch within the next quarter

### Steps to reproduce

Google Earth uses MenuItemButton which calls private member _MenuItemLabel, however when the menu item is disabled the voice over semantics doesn’t announce that it’s disabled.
https://source.corp.google.com/piper///depot/google3/third_party/dart/flutter/lib/src/material/menu_anchor.dart;l=2766

what I see happening is if 
a) there’s no semantic label provided, on line 2875, it labels it using _LocalizedShortcutLabel which doesn’t take in account the enabled state, or
b) on line 2889 it creates a Semantics(…) object if we provide a label, but again doesn’t have “enabled” field used.

this is where we use MenuItemButton: https://source.corp.google.com/piper///depot/google3/geo/earth/app/flutter/components/menu_items/lib/menu_items.dart;l=111

### Expected results

voice over pronounces that the menu item is disabled when that menu item is disabled

### Actual results

voice over calls a disabled menu item button as its regular semantic name

### Code sample

<details open><summary>Code sample</summary>

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[Paste your output here]
```

</details>

darshankawar:
Thanks for the report. I used [this](https://api.flutter.dev/flutter/material/MenuItemButton-class.html#material.MenuBar.1) code sample which has enabled and disabled menu entries and verified by running on iOS voiceover. 
The enabled entry is read as: `Show Message control S`. 
The disabled entry is `Reset Message Esc` which is read as: `Reset message esc dim`.

Based on the findings as mentioned in issue description, I'll keep the issue open and label for team's tracking.

```
stable : 3.32.5
master : 3.33.0-1.0.pre.677
```

PR #127620, opened 2023-05-25
Remove button announcement for `MenuItemButton` and `SubmenuButton`

QuncCccccc:
Related to #127617.
This is just to remove "button" announcement when we traverse menu items. 

Todo:
For `SubmenuButton`, we should be able to announce its status: collapsed or expanded after we announce its text label and shortcut. But currently `Semantics.hint` doesn't work. It might be because hint is not wired up. Will take a look on the engine side.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #176255, opened 2025-09-30
Improve menu item accessibility semantics

flutter-zl:
This PR addresses issue #171231 by enhancing the accessibility semantics for `MenuItemButton` and `SubmenuButton`. Previously, screen readers did not correctly announce the disabled state of these menu items.

This ensures that menu items are correctly identified as buttons and their enabled/disabled state is communicated to assistive technologies.

**Before change**
https://button-0923-before.web.app/

**After change**
https://button-0923-after.web.app/


mdebbar:
It looks like the button flag was removed on purpose: https://github.com/flutter/flutter/pull/127620. I'll defer to @chunhtai to advise on the right path to take here.

chunhtai:
@QuncCccccc Do you remember why we want to remove the button from the menu before?

QuncCccccc:
IIRC, menu items don't announce "button" on windows, so we removed it in that PR.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -2893,10 +2945,13 @@ class _MenuItemLabel extends StatelessWidget {
           ),
       ],
     );
-    if (semanticsLabel != null) {
+    if (semanticsLabel != null || isEnabled != null || isSemanticButton != null) {
       menuItemLabel = Semantics(
         label: semanticsLabel,
-        excludeSemantics: true,
+        enabled: isEnabled,
+        button: isSemanticButton,
```
mdebbar: I'm not sure that this is the right place to set the semantics `button` property. I think we should be replacing the `isSemanticButton: null` passed to `TextButton` above (line 1886).

Same for `enabled`.

mdebbar (commented):
@chunhtai @QuncCccccc do you think `isSemanticButton` should be configurable by the user, or just default to the right value for each platform (or both)?

Default value per platform would be:
- Windows - `isSemanticButton: null`
- Web - `isSemanticButton: true`
- Others - ??

chunhtai:
I agree we may need to be platform adaptive on this, I think for now Web - `isSemanticButton: true` and null otherwise should be fine

packages/flutter/lib/src/material/menu_anchor.dart
chunhtai: we should programatically assign the button: true based on kIsWeb, this is not something developer can easily pick and know about the consequence, and it seems  like there may not be a good reason to provide the flexibility

flutter-zl: Great suggestion. I updated the PR.

flutter-zl: Great suggestion. I updated the PR.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -935,7 +935,7 @@ class _MenuItemButtonState extends State<MenuItemButton> {
       autofocus: widget.enabled && widget.autofocus,
       statesController: widget.statesController,
       clipBehavior: widget.clipBehavior,
-      isSemanticButton: null,
+      isSemanticButton: kIsWeb,
```
mdebbar: Is `null` equivalent to `false`?

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -1846,7 +1846,7 @@ class _SubmenuButtonState extends State<SubmenuButton> {
                 focusNode: _buttonFocusNode,
                 onFocusChange: _enabled ? widget.onFocusChange : null,
                 onPressed: _enabled ? toggleShowMenu : null,
-                isSemanticButton: null,
+                isSemanticButton: kIsWeb,
```
mdebbar: Same question here.

mdebbar (commented):
We changed the value on other platform from `null` to `false`. If those are equivalent, then the PR looks good to me!

packages/flutter/lib/src/material/menu_anchor.dart
flutter-zl: good catch. updated.

packages/flutter/lib/src/material/menu_anchor.dart
flutter-zl: good catch. updated.

mdebbar (approved):

PR #176256, opened 2025-09-30
Fix code style in Linux embedder template

robert-ancell:
As noticed by Gemini in another PR


stuartmorgan-g:

packages/flutter_tools/templates/app/linux.tmpl/runner/my_application.cc.tmpl
```diff
@@ -15,8 +15,7 @@ struct _MyApplication {
 G_DEFINE_TYPE(MyApplication, my_application, GTK_TYPE_APPLICATION)
 
 // Called when first Flutter frame received.
-static void first_frame_cb(MyApplication* self, FlView *view)
-{
+static void first_frame_cb(MyApplication* self, FlView *view) {
```
stuartmorgan-g: Do you want to fix the pointer alignment while you are fixing style on this line? `FLView *view` appears to be the only thing in the file using a right-aligned pointer.

robert-ancell: I ran it though clang-format and they should have all been left aligned. A bigger change but probably a better long term fix.

stuartmorgan-g: What rule set did you use? It appears to have right-aligned everything, which is [the opposite of Google C++ style](https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions).

robert-ancell: I ran clang-format from the project root, which didn't pick up the config. Re-run.

stuartmorgan-g (approved):
LGTM

mattkae (approved):
Cool :beans:

PR #175558, opened 2025-09-18
Add new properties to DropdownMenu that allows padding customization for its leadingIcon and trailingIcon

LauYewHang:
This PR adds new properties to DropdownMenu class that allows padding customization for its leadingIcon and trailingIcon, as well as customization on how the trailing icon button in a DropdownMenu should be constructed. This changes hope to address the issue where a [DropdownMenu's minimum size (specifically the height) is affected by the trailing IconButton](https://github.com/flutter/flutter/issues/135903#issuecomment-1748014507) within it. In which due to the constraints of the IconButton having a minimum size of 48x48 under M3 design guidelines, the DropdownMenu's height is unable to go lower than this value.

<details open>
<summary>Comparison</summary>

<details>
<summary>Before changes:</summary>

Code:
```dart
DropdownMenu<IconLabel>(
  dropdownMenuEntries: IconLabel.entries,
  requestFocusOnTap: true,
  inputDecorationTheme: InputDecorationTheme(
    border: OutlineInputBorder(),
    suffixIconConstraints: BoxConstraints(),
    suffixIconColor: Color(0xFF00FFFF),
    isDense: true,
    contentPadding: EdgeInsets.zero,
  ),
  trailingIcon: Icon(Icons.favorite),
)
```
</details>

<details open>
<summary>After changes:</summary>

Code:
```dart
DropdownMenu<IconLabel>(
  dropdownMenuEntries: IconLabel.entries,
  requestFocusOnTap: true,
  inputDecorationTheme: InputDecorationTheme(
    border: OutlineInputBorder(),
    suffixIconConstraints: BoxConstraints(),
    suffixIconColor: Color(0xFF00FFFF),
    isDense: true,
    contentPadding: EdgeInsets.zero,
  ),
  trailingIcon: Icon(Icons.favorite),
  trailingIconPadding: EdgeInsets.zero,
  trailingIconButtonStyle: DropDownMenuTrailingIconButtonStyle(
    iconSize: 12,
    padding: EdgeInsets.zero,
    style: ButtonStyle(
      tapTargetSize: MaterialTapTargetSize.shrinkWrap
    )
  ),
)
```
</details>
</details>

Fixes #135903 

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
[...]

Issue #174096, opened 2025-08-20
Make DropdownMenu's trailing icon not focusable by default


bleroux:
### Use case

The current DropdownMenu implementation wrapped the trailing icon in an IconButton. Because both the TextField and the IconButton are focusable by default, this led to several focus issues.
it would be better if the icon was not a button by default.

Quoting @davidhicks980 from https://github.com/flutter/flutter/pull/172753#discussion_r2255412067.

"most comboboxes in the wild default to an unfocusable "button" that usually only opens the listbox when clicked"

Some examples:

* https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/
* https://react-spectrum.adobe.com/react-aria/ComboBox.html#menu-trigger-behavior

On MacOS:
https://github.com/user-attachments/assets/7738df66-c144-42aa-b8ee-d4d229915d43

<img width="1010" height="518" alt="image" src="https://github.com/user-attachments/assets/3ea096b1-f457-44d3-8daa-16ba8d55f616" />

From WAI: 
"The button has been removed from the tab sequence of the page, but is still important to assistive technologies for mobile devices that use touch events to open the list of options."

### Proposal

Make the DropdownMenu trailing icon not focusable by default.

dkwingsmt:
@bleroux Has this issue been resolved by #172753 ? I saw the mention but don't see where.

bleroux:
> Has this issue been resolved by #172753 ? I saw the mention but don't see where.

No, it has not been resolved by #172753. 
This issue was created after some discussions on #172753. More discussions, experimentations and work are needed to fix this issue.

PR #167782, opened 2025-04-24
Add option to hide trailing icon in DropdownMenu

mohamedamara:
This PR adds support for hiding the trailing icon in `DropdownMenu` widget. Currently, there's no built-in option to remove it. 

The change is non-breaking, as the trailing icon remains visible by default unless `showTrailingIcon` is explicitly set to `false`. The `showTrailingIcon` parameter follows a similar pattern to its use in the `ExpansionTile` widget. 

Fixes #164908

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #176264, opened 2025-09-30
Add DropdownMenu.decorationBuilder

bleroux:
## Description

This PR adds `DropdownMenu.decorationBuilder`.
The goal is to make `DropdownMenu` more flexible.

Before this PR, several fields are used by `DropdownMenu` to create an inner `InputDecoration`. This approach has several limitations:
- `InputDecoration` has more fields that the ones that are exposed
- `DropdownMenu` makes some choices that can't be change. Especially, it creates an IconButton (with hardcoded padding)  which is passed to `InputDecoration.suffixIcon`. This inner `IconButton` introduces some difficulty related to focus management and UI customization.

The new `DropdownMenu.decorationBuilder` property offers users a way to take control on the inner `InputDecoration` in a non-breaking way.

In a future PR, this property will help replacing the default `IconButton`.
Currently users can replace the `IconButton` using this code sample:

<details><summary>DropdownMenu without IconButton</summary>

```dart

void main() {
  runApp(MyApp());
}

  MyApp({super.key});

  final List<DropdownMenuEntry<String>> menuEntries = [
    "Red",
    "Green",
    "Blue",
  ].map((t) => DropdownMenuEntry<String>(label: t, value: t)).toList();

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: SizedBox(
            width: 220,
            child: DropdownMenu<String>(
              expandedInsets: EdgeInsets.zero,
              requestFocusOnTap: true,
              dropdownMenuEntries: menuEntries,
              decorationBuilder: (context, controller) {
                return InputDecoration(
                  labelText: 'Label text',
                  helperText: 'Select a color or enter one',
                  suffixIcon: controller.isOpen
                      ? const Icon(Icons.arrow_drop_up)
                      : const Icon(Icons.arrow_drop_down),
                );
              },
            ),
          ),
        ),
      ),
    );
  }
}
``` 

</details> 

## Related Issue

Fixes [DropDownMenu secondary trailing widget](https://github.com/flutter/flutter/issues/175847)
Will help for [Make DropdownMenu's trailing icon not focusable by default](https://github.com/flutter/flutter/issues/174096)

## Related discussions

https://github.com/flutter/flutter/issues/175847#issuecomment-3330098375
https://github.com/flutter/flutter/pull/175558#discussion_r2380227394

## Tests

- Adds 7 tests.

### DIFF
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -1176,7 +1176,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
         final InputDecoration effectiveDecoration = decoration.applyDefaults(
           effectiveInputDecorationTheme,
         );
-        final InputDecoration keyedDecoration = effectiveDecoration.prefixIcon == null
+        final InputDecoration textFieldDecoration = effectiveDecoration.prefixIcon == null
             ? effectiveDecoration
             : effectiveDecoration.copyWith(
                 prefixIcon: SizedBox(
@@ -1243,7 +1243,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
                 });
               },
               inputFormatters: widget.inputFormatters,
-              decoration: keyedDecoration,
+              decoration: textFieldDecoration,
               restorationId: widget.restorationId,
             ),
           ),
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -46,6 +46,16 @@ typedef FilterCallback<T> =
 /// Used by [DropdownMenu.searchCallback].
 typedef SearchCallback<T> = int? Function(List<DropdownMenuEntry<T>> entries, String query);
 
+/// The type of builder function used by [DropdownMenu.decorationBuilder] to
+/// build the [InputDecoration] passed to the inner text field.
+///
+/// The `context` is the context that the decoration is being built in.
+///
+/// The `controller` is the [MenuController] that can be used to open and close
+/// the menu with and query the current state.
+typedef DropdownMenuDecorationBuilder =
+    InputDecoration Function(BuildContext context, MenuController controller);
+
 const double _kMinimumWidth = 112.0;
 
 const double _kDefaultHorizontalPadding = 12.0;
@@ -182,6 +192,7 @@ class DropdownMenu<T extends Object> extends StatefulWidget {
     this.textAlign = TextAlign.start,
     // TODO(bleroux): Clean this up once `InputDecorationTheme` is fully normalized.
     Object? inputDecorationTheme,
+    this.decorationBuilder,
     this.menuStyle,
     this.controller,
     this.initialSelection,
@@ -207,6 +218,10 @@ class DropdownMenu<T extends Object> extends StatefulWidget {
                  inputDecorationTheme is InputDecorationThemeData),
        ),
        assert(trailingIconFocusNode == null || showTrailingIcon),
+       assert(
+         decorationBuilder == null ||
+             (label == null && hintText == null && helperText == null && errorText == null),
+       ),
        _inputDecorationTheme = inputDecorationTheme;
 
   /// Determine if the [DropdownMenu] is enabled.
@@ -368,6 +383,23 @@ class DropdownMenu<T extends Object> extends StatefulWidget {
   }
 
   final Object? _inputDecorationTheme;
+
+  /// The builder function used to create the [InputDecoration] passed to the text field.
+  ///
+  /// If a value is provided for this property and the resulting [InputDecoration.suffixIcon]
+  /// is null, a default [IconButton] is assigned as the suffix icon. This button's icon will
+  /// use [trailingIcon] and [selectedTrailingIcon] if those are explicitly defined; otherwise,
+  /// it defaults to [Icons.arrow_drop_down] for the collapsed state and [Icons.arrow_drop_up]
+  /// for the expanded state.
+  ///
+  /// If null, the default builder creates a decoration where:
+  /// - [InputDecoration.label] is set to [label].
+  /// - [InputDecoration.hintText] is set to [hintText].
+  /// - [InputDecoration.helperText] is set to [helperText].
+  /// - [InputDecoration.errorText] is set to [errorText].
+  /// - [InputDecoration.prefixIcon] is set to [leadingIcon].
+  /// - [InputDecoration.suffixIcon] is set to an [IconButton] which uses [trailingIcon] and [selectedTrailingIcon] if defined, or [Icons.arrow_drop_down] and [Icons.arrow_drop_up] otherwise.
+  final DropdownMenuDecorationBuilder? decorationBuilder;
 
   /// The [MenuStyle] that defines the visual attributes of the menu.
   ///
@@ -1132,38 +1164,26 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
       crossAxisUnconstrained: false,
       builder: (BuildContext context, MenuController controller, Widget? child) {
         assert(_initialMenu != null);
-        final bool isCollapsed = widget.inputDecorationTheme?.isCollapsed ?? false;
-        final Widget trailingButton = widget.showTrailingIcon
-            ? Padding(
-                padding: isCollapsed ? EdgeInsets.zero : const EdgeInsets.all(4.0),
-                child: ExcludeSemantics(
-                  // When the text field is treated as a button (i.e., it can
-                  // not be focused), the trailing button should become part of
-                  // the text field button by excluding semantics. Otherwise,
-                  // it will inappropriately announce whether this icon button
-                  // is selected or not.
-                  excluding: !canRequestFocus(),
-                  child: IconButton(
-                    focusNode: _trailingIconButtonFocusNode,
-                    isSelected: controller.isOpen,
-                    constraints: widget.inputDecorationTheme?.suffixIconConstraints,
-                    padding: isCollapsed ? EdgeInsets.zero : null,
-                    icon: widget.trailingIcon ?? const Icon(Icons.arrow_drop_down),
-                    selectedIcon: widget.selectedTrailingIcon ?? const Icon(Icons.arrow_drop_up),
-                    onPressed: !widget.enabled
-                        ? null
-                        : () {
-                            handlePressed(controller);
-                          },
-                  ),
-                ),
-              )
-            : const SizedBox.shrink();
-
-        final Widget leadingButton = Padding(
-          padding: const EdgeInsets.all(8.0),
-          child: widget.leadingIcon ?? const SizedBox.shrink(),
+        final DropdownMenuDecorationBuilder decorationBuilder =
+            widget.decorationBuilder ?? _buildDefaultDecoration;
+        InputDecoration decoration = decorationBuilder(context, controller);
+        // If no suffixIcon is provided, the default IconButton is used for convenience.
+        if (decoration.suffixIcon == null) {
+          decoration = decoration.copyWith(
+            suffixIcon: _buildDefaultSuffixIcon(context, controller),
+          );
+        }
+        final InputDecoration effectiveDecoration = decoration.applyDefaults(
+          effectiveInputDecorationTheme,
         );
+        final InputDecoration keyedDecoration = effectiveDecoration.prefixIcon == null
+            ? effectiveDecoration
+            : effectiveDecoration.copyWith(
+                prefixIcon: SizedBox(
+                  key: _leadingKey, // Used to query the width in refreshLeadingPadding.
+                  child: effectiveDecoration.prefixIcon,
+                ),
+              );
 
         final MaterialLocalizations localizations = MaterialLocalizations.of(context);
         final bool isButton = !canRequestFocus();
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -44,6 +44,17 @@ typedef FilterCallback<T> =
...
 typedef SearchCallback<T> = int? Function(List<DropdownMenuEntry<T>> entries, String query);
 
+/// The type of builder function used by [DropdownMenu.decorationBuilder] to
+/// build the [InputDecoration] passed to the inner text field.
+///
+/// The `context` is the context that the decoration is being built in.
+///
+/// The `controller` is the [MenuController] that can be used to open and close
+/// the menu with and query the current state.
+///
```
QuncCccccc:
```suggestion
```

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -367,6 +383,22 @@ class DropdownMenu<T> extends StatefulWidget {
 
   final Object? _inputDecorationTheme;
 
+  /// The builder function used to create the [InputDecoration] passed to the text field.
+  ///
+  /// If provided and the resulting [InputDecoration.suffixIcon] is null,
+  /// the [InputDecoration.suffixIcon] is set to an [IconButton] which uses [trailingIcon]
+  /// and [selectedTrailingIcon] if defined, or [Icons.arrow_drop_down] and
+  /// [Icons.arrow_drop_up] otherwise.
```
QuncCccccc: nit: Maybe this.
```suggestion
/// If a value is provided for this property and the resulting [InputDecoration.suffixIcon] is null, a default [IconButton] is assigned as the suffix icon. This button's icon will use [trailingIcon] and [selectedTrailingIcon] if those are explicitly defined; otherwise, it defaults to [Icons.arrow_drop_down] for the collapsed state and [Icons.arrow_drop_up] for the expanded state.
```

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1125,30 +1157,24 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
-                        },
-                ),
-              )
-            : const SizedBox.shrink();
-
-        final Widget leadingButton = Padding(
-          padding: const EdgeInsets.all(8.0),
-          child: widget.leadingIcon ?? const SizedBox.shrink(),
+        final DropdownMenuDecorationBuilder decorationBuilder =
+            widget.decorationBuilder ?? _buildLegacyDecoration;
```
QuncCccccc: nit: do you think if _buildDefaultDecoration will be better? And other places.
```suggestion
            widget.decorationBuilder ?? _buildDefaultDecoration;
```

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1125,30 +1157,24 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
+        InputDecoration decoration = decorationBuilder(context, controller);
+        // If no suffixIcon is provided, the legacy IconButton is used for convenience.
```
QuncCccccc: also nit here:
```suggestion
        // If no suffixIcon is provided, the default IconButton is used for convenience.
```

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1125,30 +1157,24 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
+        if (decoration.suffixIcon == null) {
+          decoration = decoration.copyWith(suffixIcon: _buildLegacySuffixIcon(context, controller));
+        }
+        final InputDecoration effectiveDecoration = decoration.applyDefaults(
+          effectiveInputDecorationTheme,
         );
+        final InputDecoration keyedDecoration = effectiveDecoration.prefixIcon == null
```
QuncCccccc: Can we just make changes on the `effectiveDecoration`? Feels a bit confused for the name of this variable😄

QuncCccccc (commented):
I like this approach! Thank you for the improvement. Overall this looks good to me:) Will take another look tomorrow.

packages/flutter/lib/src/material/dropdown_menu.dart
bleroux: > Can we just make changes on the effectiveDecoration?

Thats not possible because both fields are needed below.
The 'effectiveDecoration' is used in DropdownMenuBody and 'keyedDecoration' is passed to the TextField.
Maybe I can use 'textFieldDecoration' instead of 'keyedDecoration'? ("Keyed" was because the purpose of this decoration copy is to add a key to the prefixIcon, key which is used later for measurement).
I can also get rid of of the local variable and move the logic to line 1210 where this variable is used.

packages/flutter/lib/src/material/dropdown_menu.dart
bleroux: The reason I used 'Legacy' is because my end goal is to not have this logic to be the default in the future.
The default logic I have in mind would replace the IconButton with a simple icon. Doing so we would have to find a way to keep the current behavior which would become optional and marked as "Legacy".

Maybe I should just let this thinking for a future PR :smile: . Let me apply your suggestion, it just mean that `_buildDefaultDecoration` implementation will change in the future. Seems reasonable.

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: `textFieldDecoration` sounds good to me!

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1132,38 +1164,26 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
...
-              )
-            : const SizedBox.shrink();
-
-        final Widget leadingButton = Padding(
-          padding: const EdgeInsets.all(8.0),
-          child: widget.leadingIcon ?? const SizedBox.shrink(),
+        final DropdownMenuDecorationBuilder decorationBuilder =
+            widget.decorationBuilder ?? _buildDefaultDecoration;
+        InputDecoration decoration = decorationBuilder(context, controller);
+        // If no suffixIcon is provided, the default IconButton is used for convenience.
```
QuncCccccc, line 1170: Do you think if there will be a case where the user doesn't want the `sufficIcon` at all? I'm thinking whether we should still give it a default value when suffixIcon is null.

bleroux: Interesting.
The property `DropdownMenu.showTrailingIcon` was added some time ago in https://github.com/flutter/flutter/pull/167782 for that purpose.

Because `DropdownMenu.decorationBuilder` is new we can decide to use a different solution. Nonetheless, If we don't add a default suffix when `suffixIcon` is null, most of the users will have to provide their implementation or we have to offer them an API to get the default Icon. It seems to add complexity for a use case which is somewhat specific.

Gustl22: I'd propose to always add the default icon, also with a custom decoration (when the user adds an additinal trailing icon, then keep it in a `Row`). And the user can still disable it via the property `showTrailingIcon`. That way the dev does not need to implement it, e.g. also with providing more additional trailing actions, but still has the opportunity to remove it.
That way it has a purpose in all cases, while providing maximum flexibility and minimal effort for the dev.

The other way around: in the current implementation, if the user defines its own trailing icon, the `showTrailingIcon` does not have any effect, right?

bleroux: > I'd propose to always add the default icon, also with a custom decoration (when the user adds an additinal trailing icon, then keep it in a Row)

Adding a Row fits specific use cases and for these use cases it might be better for us to expose the default button construction if you think this would be interesting and let users add a Row (or something else) if needed (that way they can control padding for instance or anything matching their design).
My reasonning is to remove as much as possible custom rendering logic outside of `DropdownMenu`. In that particular case, `InputDecorator` already implements APIs that comply with M3 spec (padding, defaults colors, borders, etc) and offer flexibility (InputDecoration.suffix, InputDecoration.suffixIcon, etc).

> The other way around: in the current implementation, if the user defines its own trailing icon, the showTrailingIcon does not have any effect, right?

Yes. It surely should be documented (in fact the best move would probably to rename `showTrailingIcon` to `showDefaultTrailingIcon` in that case). We can also decide to ignore `InputDecoration.suffixIcon` if `showTrailingIcon` is false, it might be easier to document/understand.

Gustl22: > Adding a Row fits specific use cases and for these use cases it might be better for us to expose the default button construction if you think this would be interesting and let users add a Row

Good point.
Of course the `Row` was just meant as an internal helper to still be able to access the default trailing icon. But yes, it would then only serve in that (in my opinion most common) use case, and shrinks flexibility how (e.g. in which order) to use it.

I personally don't need the default trailing button, as the fuctionality can easily be recreated with the controller. Not sure if folks would notice / search for the constructor of the default icon.

> Yes. It surely should be documented (in fact the best move would probably to rename showTrailingIcon to showDefaultTrailingIcon in that case). We can also decide to ignore InputDecoration.suffixIcon if showTrailingIcon is false, it might be easier to document/understand.

Sounds reasonable. Happy with both variants :)

Btw thank you for your efforts!

QuncCccccc: >  when suffixIcon is null, most of the users will have to provide their implementation or we have to offer them an API to get the default Icon. It seems to add complexity for a use case which is somewhat specific.

I see, it makes sense to me.

packages/flutter/test/material/dropdown_menu_test.dart
```diff
@@ -4765,6 +4765,162 @@ void main() {
...
+            ? const Icon(Icons.arrow_drop_up)
+            : const Icon(Icons.arrow_drop_down),
+      );
+    }
+
+    InputDecoration buildDecoration(BuildContext context, MenuController controller) {
+      return const InputDecoration(labelText: labelText);
+    }
+
+    testWidgets('Custom decorationBuilder can replace default suffixIcon', (
```
QuncCccccc, line 4844: nit: Can we add one more unit test to create a custom InputDecoration using decorationBuilder (a few custom arguments would be fine) and check the results match the customization?

QuncCccccc (approved):
LGTM. Thank you for adding all the tests:)

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -46,6 +46,16 @@ typedef FilterCallback<T> =
...
 
+/// The type of builder function used by [DropdownMenu.decorationBuilder] to
+/// build the [InputDecoration] passed to the inner text field.
+///
+/// The `context` is the context that the decoration is being built in.
+///
+/// The `controller` is the [MenuController] that can be used to open and close
+/// the menu with and query the current state.
+typedef DropdownMenuDecorationBuilder =
+    InputDecoration Function(BuildContext context, MenuController controller);
```
Gustl22, line 57: @bleroux I noticed a little naming improvement: Maybe it makes sense to call the parameter `menuController`, as in the widget itself `controller` is always referenced as the `TextEditingController`. Maybe it makes sense to rename the default param name to reduce confusion (?)

bleroux: @Gustl22 I usually try to use more precise names. In the context of DropdownMenu, there are multiple controllers so it could help, especially for instance variables.

About this particular case, because the name is just in a typedef and the explicit type is just before, I probably preferred `controller` as the type of controller was explicit.
Is this variable visible anywhere else? Maybe when using completion?

Gustl22: Yes this is only a problem of code completion for now.

I just stumbled across this scenario, where I wanted to add a clear button:

```dart
class SimpleDropdown<T extends Object> extends StatefulWidget {
  final Iterable<DropdownMenuEntry<T>> options;
  final T? selected;
  final void Function(T? value)? onChange;
  final FormFieldSetter<T>? onSaved;

  const SimpleDropdown({required this.options, required this.selected, this.onChange, this.onSaved});

  @override
  State<SimpleDropdown<T>> createState() => _SimpleDropdownState<T>();
}

class _SimpleDropdownState<T extends Object> extends State<SimpleDropdown<T>> {
  final _textEditingController = TextEditingController();

  @override
    final items = widget.options.toList();
    return DropdownMenuFormField<T>(
      initialSelection: widget.selected,
      controller: _textEditingController,
      onSelected: widget.onChange,
      onSaved: widget.onSaved,
      dropdownMenuEntries: items,
      decorationBuilder:
          (context, menuController) => InputDecoration(
            suffixIcon: IconButton(onPressed: () => _textEditingController.clear(), icon: Icon(Icons.clear)),
          ),
    );
  }
}
```

I needed to use a StatefulWidget, because the `TextEditingController` was not provided by the `decorationBuilder`.
For me it would make sense to have both controllers available in the callback, or none (so I should pass both the `MenuController` and the `TextEditingController` to the `DropdownMenu`).
Its not per se bad, how it is right now. Just wanted to hear your opinion, what makes `MenuController` more valuable to exist in the callback than the `TextEditingController`, which might be the option to mimic the current behavior of opening and closing.

bleroux: > Just wanted to hear your opinion, what makes MenuController more valuable to exist in the callback than the TextEditingController, which might be the option to mimic the current behavior of opening and closing.

Yes, the reasoning was to give an easy way to reproduce the default behavior,
Another reason might be that I started to work on this before the DropdownMenu.menuController property was added (some weeks before my PR, see https://github.com/flutter/flutter/pull/175039).

PR #176289, opened 2025-09-30
[ Widget Preview ] Persist "Filter by Selected File" toggle

bkonyi:
This change introduces `PersistentPreferences`, which allows for the widget previewer to save settings to disk.

`PersistentPreferences` makes use of the existing `~/.flutter-devtools` directory used by DevTools for the same purpose, writing preferences in JSON format to `~/.flutter-devtools/.widget-preview`.


packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
```diff
@@ -0,0 +1,74 @@
...
+// found in the LICENSE file.
+import 'package:devtools_shared/devtools_server.dart' as devtools;
+import 'package:meta/meta.dart';
+
+import '../base/file_system.dart';
+import '../convert.dart';
+
+/// A persistent store used to save settings across sessions.
+///
+/// The store is written to `~/.flutter-devtools/.widget-preview` in JSOn format.
```
DanTup: nit: `JSOn`

packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    if (value == null) {
+      _map.remove(key);
+    } else {
+      _map[key] = value;
+    }
+
+    try {
+      const jsonEncoder = JsonEncoder.withIndent('  ');
+      file.writeAsStringSync('${jsonEncoder.convert(_map)}\n');
```
DanTup: We spawn a widget preview for each "window" in VS Code. If you have multiple windows open (which I don't think is uncommon), they could overwrite each others files here. It's a bit of an edge case, but it might be confusing. I wonder if we should re-read them on-change and then persist with just the one value updated?

packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
```diff
@@ -0,0 +1,74 @@
...
+  static const _kPreferencesFileName = '.widget-preview';
+
+  final FileSystem fs;
+
+  @visibleForTesting
+  late final File file = fs.file(
+    fs.path.join(devtools.LocalFileSystem.devToolsDir(), _kPreferencesFileName),
+  );
+
+  late Map<String, Object?> _map;
```
DanTup: _map here is typed with `Object?` as keys, but the other code seems to avoid persisting `null`s and assumes `Object`. Should we drop the `?` here for consistency? (or add it elsewhere)

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/dtd_services.dart.tmpl
```diff
@@ -81,6 +83,33 @@ class WidgetPreviewScaffoldDtdServices with DtdEditorService {
     return result == null ? null : Uri.parse(result);
   }
 
+  /// Retrieves an arbitrary value associated with [key] from the persistent
+  /// preferences map.
+  ///
+  /// Returns null if [key] is not in the map.
+  Future<String?> getPreference(String key) async {
```
DanTup, line 95: How come we're using `String`s here and then parsing them, if the underlying json/map can store `Object`?

packages/flutter_tools/test/commands.shard/permeable/widget_preview/dtd_services_test.dart
```diff
@@ -70,5 +76,100 @@ void main() {
...
+          WidgetPreviewDtdServices.kWidgetPreviewService,
+          WidgetPreviewDtdServices.kSetPreference,
+          params: {'key': key, 'value': value},
+        );
+      }
+
+      const kTestKey = 'myKey';
+
+      // The preferences file should be empty.
+      expect(await getPreference(kTestKey), 'null');
```
DanTup: Related to above - having the _string_ `null` here seems weird (particularly as the return type is `Future<String?>`).

packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
bkonyi: I think the only correct way to handle this is to register a file watcher that's responsive to changes in the preferences file, which then sends DTD events to the widget preview scaffold so it can update its state immediately.

I pulled most of this implementation from DevTools, so it'll have the same issue. Maybe this is something we revisit?

packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
bkonyi: Updated the `PersistentPreferences` APIs to not allow `null` values, except for signaling an entry should be removed.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/dtd/dtd_services.dart.tmpl
bkonyi: Good point. Moved to a more generic interface.

packages/flutter_tools/lib/src/widget_preview/dtd_services.dart
```diff
@@ -115,6 +127,18 @@ class WidgetPreviewDtdServices {
...
+    final String key = params['key'].asString;
+    final String value = params['value'].asString;
+    preferences[key] = value;
+    return const Success().toJson();
+  }
+
+  Future<Map<String, Object?>> _getPreference(Parameters params) async {
+    final String key = params['key'].asString;
+    return StringResponse(preferences[key].toString()).toJson();
+  }
```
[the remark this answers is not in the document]
bkonyi: This isn't a valid change as `StringResponse` takes a `String` as a parameter to the constructor, not `String?`.

packages/flutter_tools/lib/src/widget_preview/persistent_properties.dart
DanTup: I agree watching the file is probably overkill, but what about re-reading the file just before modifying, so if another process had saved a value, it's preserved rather than wipe? (just an idea, not suggesting it should be done to land this)

DanTup (approved):

PR #176292, opened 2025-09-30
[web] Updates to the README

mdebbar:
The main addition here is the instructions for using a locally built dart sdk.

Along the way, I also did some other minor updates to this file.


engine/src/flutter/lib/web_ui/README.md
```diff
@@ -327,6 +311,12 @@ roll the dependency on `third_party/emsdk` in DEPS to the same version as in
 Once you know the version for the Emscripten SDK, change the line in
 `tools/activate_emsdk.py` which defines `EMSDK_VERSION` to match Skia.
 
+## Using a locally built Dart SDK
+
+In order to use a locally built Dart SDK, you simply set the `DART_SDK_DIR` environment variable like so:
+```
+DART_SDK_DIR=path/to/dart-sdk/ felt ...
+```
```
eyebrowsoffire, line 319: You may want to mention here that the `DART_SDK_DIR` is mostly for `felt test` but not `felt build`. The felt tool itself uses the specified dart sdk, and unit tests are built with the specified dart SDK, but the actual `gn` build (that `felt build` invokes) still uses its normal deps system to resolve the dart SDK, which will be whatever is in `flutter/third_party/dart`.

mdebbar: Done

eyebrowsoffire (approved):
LGTM

PR #168981, opened 2025-05-16
Normalize input decoration theme

bleroux:
This PR is to make `InputDecorationTheme` conform to Flutter Material's conventions for component themes:

- Added a `InputDecorationThemeData` class which defines overrides for the defaults for `InputDecorator` properties.
- Added `InputDecorationTheme` constructor parameters: `InputDecorationThemeData? data` and `Widget? child`. This is now the preferred way to configure a `InputDecorationTheme`:

```dart
InputDecorationTheme(
	data: InputDecorationThemeData(
	    filled: true,
	    fillColor: Colors.amber,
	    ...
	  ),
  	child: const TextField()
)
```
These two properties are made nullable to not break existing apps which has customized `ThemeData.inputDecorationTheme`.

- Update `InputDecorationTheme` to be an `InheritedTheme` subclass.
- Changed the type of component theme defaults from `InputDecorationTheme` to `InputDecorationThemeData`.
- Changed the `InputDecorationTheme bottomAppBarTheme` property to `Object? bottomAppBarTheme` in `ThemeData` and `ThemeData.copyWith()` (Object? is used for the moment to minimize Google tests failure. A follow-up PR will replace `Object?` with `InputDecorationThemeData`.
- Addresses the "theme normalization" sub-project within #91772.

A migration guide will be created on website repo.

PR #176300, opened 2025-09-30
Fix TextField does not inherit local InputDecorationTheme

bleroux:
## Description

This PR replaces global `ThemeData.inputDecorationTheme` usage in `TextField` with `InputDecorationTheme.of ` which returns the ambient `InputDecorationTheme`.
It is a follow up to https://github.com/flutter/flutter/pull/168981 which introduces `InputDecorationTheme.of `.

## Related Issue

Fixes [[SearchBar] Inconsistent inheritance of InputDecorationTheme](https://github.com/flutter/flutter/issues/176198)

## Tests

- Adds 1 test
- Updates 1 test

### DIFF
--- a/packages/flutter/lib/src/material/text_field.dart
+++ b/packages/flutter/lib/src/material/text_field.dart
@@ -1193,13 +1193,13 @@ class _TextFieldState extends State<TextField>
   InputDecoration _getEffectiveDecoration() {
     final MaterialLocalizations localizations = MaterialLocalizations.of(context);
     final ThemeData themeData = Theme.of(context);
-    final InputDecorationThemeData ambientDecoration = InputDecorationTheme.of(context);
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1193,14 +1193,13 @@ class _TextFieldState extends State<TextField>
   InputDecoration _getEffectiveDecoration() {
     final MaterialLocalizations localizations = MaterialLocalizations.of(context);
     final ThemeData themeData = Theme.of(context);
+    final InputDecorationThemeData ambientDecoration = InputDecorationTheme.of(context);
```
QuncCccccc: nit: maybe just `decorationTheme` because seems we don't use ambient a lot.

QuncCccccc (approved):
Nice improvement. LGTM! Thank you!

PR #176302, opened 2025-09-30
Make sure that a MenuAnchor doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the MenuAnchor widget.


dkwingsmt:
Since `MenuAnchor` contains an overlay, can you use the same approach as `DropdownMenu` and set the view size?

dkwingsmt (approved):
LGTM.

(I don't know why the unit tests keep failing. They don't fail on my computer. Let me retry them again.)

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -5247,6 +5247,24 @@ void main() {
     );
     expect(tester.getSize(find.byType(CheckboxMenuButton)), Size.zero);
   });
+
+  testWidgets('MenuAnchor does not crash at zero area', (WidgetTester tester) async {
+    tester.view.physicalSize = Size.zero;
```
victorsanni, line 5540: Can you ensure the view resets to it's normal default value after this test? Otherwise other tests will be run with a zero-sized view which can cause unrelated test failures.

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -5247,6 +5247,26 @@ void main() {
     );
     expect(tester.getSize(find.byType(CheckboxMenuButton)), Size.zero);
   });
+
+  testWidgets('MenuAnchor does not crash at zero area', (WidgetTester tester) async {
+    final Size temp = tester.view.physicalSize;
```
victorsanni: Instead of creating a new local variable, you can just add this one-liner instead:

```
addTearDown(tester.view.reset);
```

dkwingsmt:
Ohhhh right, we haven't reverted the view size after setting it.

victorsanni (approved):

PR #176303, opened 2025-09-30
Add deeplinking for UIScene migration

### LINKED ISSUE #174403: [UIScene] Migrate Deep Linking / Universal Links for scenes
## Problem

Deep linking on iOS relies on two application lifecycle events. After migrating to UIScene, these events are no longer called by UIKit. Instead, the UIScene lifecycle event is called.
* [application:continueUserActivity:restorationHandler:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:continue:restorationhandler:)?language=objc) → [scene:continueUserActivity:](https://developer.apple.com/documentation/UIKit/UISceneDelegate/scene(_:continue:))
* [application:openURL:options:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:open:options:)?language=objc) → [scene:openURLContexts:](https://developer.apple.com/documentation/uikit/uiscenedelegate/scene(_:openurlcontexts:))

In the engine, when these events are called, it forwards them to the FlutterViewController, which then uses a method channel from the FlutterEngine to send the information to the framework. It will then return with either a success or failure. ([Source](https://github.com/flutter/flutter/blob/f63b38f275f63bd2776d7f107e2f7bc660b837ec/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterAppDelegate.mm#L165-L192), [Source](https://github.com/flutter/flutter/blob/f63b38f275f63bd2776d7f107e2f7bc660b837ec/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm#L1903-L1931))

https://docs.flutter.dev/cookbook/navigation/set-up-universal-links#test-the-universal-link

## Prerequisites 

* https://github.com/flutter/flutter/issues/174395

## Proposal
Flutter will need to add the UIScene lifecycle equivalents outlined above. Then when the FlutterSceneDelegate receives the event, instead of forwarding to the FlutterViewController, it will forward it directly to each FlutterEngine associated with it.

```objc
- (void)scene:(UIScene*)scene continueUserActivity:(NSUserActivity*)userActivity {
 if (![self isFlutterDeepLinkingEnabled]) {
   return NO;
 }
 for (FlutterEngine* engine in [_engines allObjects]) {
   if (!engine) {
     continue;
   }
   [engine sendDeepLinkToFramework:url ...];
 }
}
```

See [prototype](https://github.com/flutter/flutter/compare/master...vashworth:flutter:uiscene_lifecycle_wip).

vashworth:
When an iOS app migrates to UIScene, the application events for handling deeplinks are no longer called by UIKit. Instead, scene events in the SceneDelegate must be used. This PR adds support for handling deeplinks in UIScene while still maintaining support for unmigrated apps, as well.

Fixes https://github.com/flutter/flutter/issues/174403.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -252,6 +292,29 @@ - (BOOL)windowScene:(UIWindowScene*)windowScene
...
+
+- (BOOL)handleOpenURL:(NSURL*)url relayToSystemIfUnhandled:(BOOL)throwBack {
+  if (!url) {
+    return NO;
+  }
+  // Don't process the link if deep linking is disabled.
+  if (!FlutterSharedApplication.isFlutterDeepLinkingEnabled) {
+    return NO;
+  }
+  for (FlutterEngine* engine in _engines.allObjects) {
```
LongCatIsLooong: IIRC there's method you need to call before accessing the weak pointer array?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -137,8 +137,33 @@ - (BOOL)scene:(UIScene*)scene
...
+    return consumedByPlugin;
+  }
+
+  //  If your app has opted into Scenes, and your app is not running, the system delivers the
+  //  universal link to the scene(_:willConnectTo:options:) delegate method after launch, and to
+  //  scene(_:continue:) when the universal link is tapped while your app is running or suspended in
+  //  memory.
+  for (NSUserActivity* userActivity in connectionOptions.userActivities) {
+    if (userActivity.webpageURL != nil) {
+      if ([self handleOpenURL:userActivity.webpageURL relayToSystemIfUnhandled:YES]) {
```
LongCatIsLooong: nit: looks like `handleOpenURL` has logic for the url being nil so the nil check isn't necessary?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -137,8 +137,33 @@ - (BOOL)scene:(UIScene*)scene
...
-  // There is no application equivalent for this event and therefore no fallback.
+  if (consumedByPlugin) {
+    return consumedByPlugin;
+  }
+
+  //  If your app has opted into Scenes, and your app is not running, the system delivers the
+  //  universal link to the scene(_:willConnectTo:options:) delegate method after launch, and to
+  //  scene(_:continue:) when the universal link is tapped while your app is running or suspended in
+  //  memory.
+  for (NSUserActivity* userActivity in connectionOptions.userActivities) {
```
LongCatIsLooong: Can user activities and URLContexts be both present in a connection option?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
LongCatIsLooong: Ah `[self updateEnginesInScene:scene];`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
LongCatIsLooong: uber nit: here and below: slightly more concise if replace the if blocks with
```obj-c
consumedByPlugin = consumedByPlugin || [[self applicationLifeCycleDelegate]  sceneFallbackOpenURLContexts:URLContexts];

```

LongCatIsLooong (approved):
LGTM. It looks like the reason that deeplinking implementation is called at the end is that it's always going to return YES if deeplinking is enabled?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
LongCatIsLooong, line 134: Not directly related to this PR, but I thought at this point `_engines` should be an empty array since the scene has not connected yet so we don't know which views will be in this scene?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: Not that I'm aware of. `userActivities` handled universal links (when the link prefixed with `http`/`https`). `URLContexts` is when it's a non-standard scheme, such as `my-custom-scheme://my-url.com`. Both are a form of deeplinking

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
vashworth: I found away around that, see this comment: https://github.com/flutter/flutter/pull/175866#issuecomment-3347373087

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -112,10 +112,8 @@ API_AVAILABLE(ios(13.0))
 /**
  * Calls all plugins registered for `UIWindowScene` callbacks in order of registration until
  * a plugin handles the request.
- *
- * @return `YES` if any plugin handles the request.
  */
-- (BOOL)scene:(UIScene*)scene
+- (void)scene:(UIScene*)scene
```
vashworth, line 116: @LongCatIsLooong I changed these APIs in `FlutterSceneLifeCycleProvider` (protocol add-to-app devs can use instead of using FlutterSceneDelegate) to void because in UIKit they are void. I thought our application-equivalent returned BOOLs, however, I was mistaken in that it only retuned BOOLs for the ones UIKit did.

For example, `application:continueUserActivity:restorationHandler:` returns a BOOL because it also does in UIKit.

https://github.com/flutter/flutter/blob/72eec03e860accebd7adca221c8568e9802cd7d6/engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPluginAppLifeCycleDelegate.h#L141-L143

https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:continue:restorationhandler:)?language=objc

But `application:performActionForShortcutItem:completionHandler:` returns void because UIKit does
https://github.com/flutter/flutter/blob/72eec03e860accebd7adca221c8568e9802cd7d6/engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPluginAppLifeCycleDelegate.h#L111-L114

https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:performactionfor:completionhandler:)?language=objc

UIKit returns void for all of the scene events

vashworth: `FlutterPluginAppLifeCycleDelegate` (the protocol plugins conform to) and `FlutterEnginePluginSceneLifeCycleDelegate` (internal API that forwards events to plugin) still uses `BOOL`s so we can track that we only allow one plugin to process

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1528,6 +1528,35 @@ - (FlutterDartProject*)project {
   return self.dartProject;
 }
 
+- (void)sendDeepLinkToFramework:(NSURL*)url completionHandler:(void (^)(BOOL success))completion {
+  __weak FlutterEngine* weakSelf = self;
+  [self waitForFirstFrame:3.0
```
chunhtai, line 1533: We need this due to the existing behavior that iOS deeplink will first launch the app and then a random delay later it will sends a [UIApplication openUrl]. that is why we need a delay to wait for first frame. Is this still the case for UIScene?

vashworth: Yes, we still need the delay

PR #176309, opened 2025-09-30
Implement dialog windows for the win32 platform

mattkae:
## What's new?
- Implemented `DialogWindowControllerWin32` for win32 dialogs :rocket:
- Refactored and updated the Win32 embedder to support dialogs
- Updated the `multiple_windows` example to demonstrate both modal and modeless dialogs
- Added integration tests for the dialog windows
- Added tests for dialogs in the embedder

## How to test?
1. Run the `multiple_windows` example application with a local engine built from this pull request
2. Click the `Modeless Dialog` creation button on the main window
3. Open a regular window and click the `Modal Dialog` creation button
4. Note the behavior of modal and modeless dialogs

## Pre-launch Checklist

- [x] I signed the [CLA].


dev/integration_tests/windowing_test/.vscode/settings.json
justinmc, line 1: I guess the two .vscode files were accidentally added?

engine/src/flutter/shell/platform/windows/host_window.cc
```diff
@@ -761,4 +701,140 @@ ActualWindowSize HostWindow::GetWindowContentSize(HWND hwnd) {
...
+                                      extended_window_style, dpi)) {
+    FML_LOG(ERROR) << "Failed to run AdjustWindowRectExForDpi: "
+                   << GetLastErrorAsString();
+    return std::nullopt;
+  }
+
+  double width = static_cast<double>(rect.right - rect.left);
+  double height = static_cast<double>(rect.bottom - rect.top);
+
+  // Apply size constraints
```
justinmc: Period at the end here.

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
```diff
@@ -0,0 +1,138 @@
...
+      if (HostWindow* const owner_window = GetOwnerWindow()) {
+        UpdateModalState();
+        FocusRootViewOf(owner_window);
+      }
+      return 0;
+
+    case WM_ACTIVATE:
+      if (LOWORD(wparam) != WA_INACTIVE) {
+        // Prevent disabled window from being activated using the task
+        // switcher
```
justinmc: Missing period here and 2 lines below.

engine/src/flutter/shell/platform/windows/host_window_regular.cc
```diff
@@ -0,0 +1,40 @@
...
+                         preferred_size.preferred_view_height),
+                    constraints.smallest(), constraints.biggest(),
+                    WS_OVERLAPPEDWINDOW, 0, nullptr);
+            return {{CW_USEDEFAULT, CW_USEDEFAULT},
+                    window_size ? *window_size
+                                : Size{CW_USEDEFAULT, CW_USEDEFAULT}};
+          }(),
+          title,
+          nullptr) {
+  // TODO(knopp): What about windows sized to content?
```
justinmc: I know this was already here, but do we have an issue we could link to for this?

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [DialogWindowController], the base class for dialog windows.
+class DialogWindowControllerWin32 extends DialogWindowController {
+  /// Creates a new dialog window controller for Win32.
+  ///
+  /// When this constructor completes the native window has been created and
```
justinmc, line 488: Nit: I think it's more readable with a comma after "completes".

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+  /// has a view associated with it.
```
justinmc, line 489: Is this done synchronously on the platform?

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+  late final _DialogWindowMesageHandler _handler;
+  bool _destroyed = false;
+
+  @override
+  @internal
+  Size get contentSize {
+    _ensureNotDestroyed();
+    final _ActualContentSize size = _Win32PlatformInterface.getWindowContentSize(getWindowHandle());
+    final Size result = Size(size.width, size.height);
+    return result;
```
justinmc, line 549: Nit: I would just return this without assigning it first.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+    _ensureNotDestroyed();
+    return _Win32PlatformInterface.isIconic(getWindowHandle()) != 0;
+  }
+
+  @override
+  @internal
+  void setSize(Size? size) {
+    _ensureNotDestroyed();
+    _Win32PlatformInterface.setWindowContentSize(_owner.allocator, getWindowHandle(), size);
+  }
```
justinmc, line 581: Why no notifyListeners here?

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+
+  @override
+  @internal
+  void setMinimized(bool minimized) {
+    if (parent != null) {
+      return;
+    }
+
+    _ensureNotDestroyed();
```
justinmc, line 613: Nit: I think I would put this as the first line in the method.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,211 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+    if (minimized) {
+      _Win32PlatformInterface.showWindow(getWindowHandle(), _SW_MINIMIZE);
+    } else {
+      _Win32PlatformInterface.showWindow(getWindowHandle(), _SW_RESTORE);
+    }
```
justinmc, line 618: I'm confused but probably reading this wrong... If it's minimized then minimize it, if it's not then restore it?

examples/multiple_windows/lib/app/dialog_window_content.dart
```diff
@@ -0,0 +1,83 @@
...
+
+class DialogWindowContent extends StatelessWidget {
+  const DialogWindowContent({super.key, required this.window});
+
+  final DialogWindowController window;
+
+  @override
+  Widget build(BuildContext context) {
+    final dpr = MediaQuery.of(context).devicePixelRatio;
+    final windowSize = WindowScope.contentSizeOf(context);
```
justinmc: Would it be possible to get these two inside of the ListenableBuilder instead of here? Just trying to limit the scope of rebuilds, but probably not a major performance concern in reality.

examples/multiple_windows/lib/app/dialog_window_edit_dialog.dart
```diff
@@ -0,0 +1,157 @@
...
+        nextIsMinimized = null;
+      });
+    }
+  }
+
+  @override
+  void dispose() {
+    widget.controller.removeListener(_onNotification);
+    super.dispose();
+  }
```
justinmc, line 101: You'll need to dispose all of the TextEditingControllers that are instantiated in this class as well to avoid leaks.

examples/multiple_windows/lib/app/dialog_window_edit_dialog.dart
```diff
@@ -0,0 +1,157 @@
...
+              if (value != null) {
+                setState(() => nextIsMinimized = value);
+              }
+            },
+          ),
+        ],
+      ),
+      actions: [
+        TextButton(onPressed: () => widget.onClose(), child: Text('Cancel')),
+        TextButton(onPressed: () => _onSave(), child: Text('Save')),
```
justinmc: Nit: You could drop the lambdas here and just pass e.g. `widget.onClose`.

examples/multiple_windows/lib/app/regular_window_edit_dialog.dart
```diff
@@ -47,7 +47,7 @@ class _RegularWindowEditDialogState extends State<_RegularWindowEditDialog> {
 
   bool? nextIsFullscreen;
   bool? nextIsMaximized;
-  bool? nextIsMinized;
+  bool? nextIsMinimized;
```
justinmc, line 50: I pride myself on being a spelling nitpicker but apparently I missed this one.

examples/multiple_windows/lib/app/dialog_window_edit_dialog.dart
justinmc, line 1: I'm definitely interested to try this form out live, but looks like the code makes sense.

justinmc (commented):
Looks as expected to me! Loic should probably review the engine code when he's back.

Can you add simple unit tests for the changes in _window_win32.dart? Assuming you can avoid bumping up against the FFI calls. I think you should be able to test that createDialogWindowController now returns a controller instead of throwing at least, and hopefully test the methods on DialogWindowControllerWin32 itself.

I'm going to try to get the example running on my Windows machine tonight, will post back.

engine/src/flutter/shell/platform/windows/host_window_regular.cc
mattkae: We have no issue, but we can link to https://github.com/flutter/flutter/pull/173610

packages/flutter/lib/src/widgets/_window_win32.dart
mattkae: I believe so, but only @robert-ancell can confirm for me if that's also the case on the Linux implementation.

This comment is only for win32 anyway, so it's okay in this context.

packages/flutter/lib/src/widgets/_window_win32.dart
mattkae: The notification will come in `_handleWindowsMessage` if the size was actually changed.

packages/flutter/lib/src/widgets/_window_win32.dart
mattkae: This is actually: "If we're setting `minimized` to `true`, set the window as minimized. If not, then restore the window from minimization".

mattkae:
> Can you add simple unit tests for the changes in _window_win32.dart? Assuming you can avoid bumping up against the FFI calls. I think you should be able to test that createDialogWindowController now returns a controller instead of throwing at least, and hopefully test the methods on DialogWindowControllerWin32 itself.

I had tried this previously when I did the regular windows bits, but to no avail. There are too many things that we'd have to mock to have a meaningful unit test there. The integration tests should serve as the test of this API instead, as they are truer to reality in this regard.

engine/src/flutter/shell/platform/windows/host_window.cc
```diff
@@ -260,40 +207,46 @@ std::unique_ptr<HostWindow> HostWindow::CreateRegularWindow(
...
       initial_window_rect.width(), initial_window_rect.height(),
       engine->display_manager(), engine->windows_proc_table());
 
   std::unique_ptr<FlutterWindowsView> view =
       engine->CreateView(std::move(view_window));
-  if (view == nullptr) {
-    FML_LOG(ERROR) << "Failed to create view";
-    return nullptr;
-  }
+  FML_CHECK(view != nullptr);
```
loic-sharma, line 248: I slightly prefer the previous model where we would try to not crash if the window could not be created (I assume this resulted in an exception on the Dart side). Should we stick to that approach? (Not a blocker for landing this PR).

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
 
-  HostWindow(WindowManager* controller,
+  HostWindow(WindowManager* window_manager,
              FlutterWindowsEngine* engine,
              WindowArchetype archetype,
-             std::unique_ptr<FlutterWindowsViewController> view_controller,
-             const BoxConstraints& constraints,
-             HWND hwnd);
+             DWORD window_style,
+             DWORD extended_window_style,
```
loic-sharma, line 129: These `DWORD`s are a bit opaque. Would it make sense to switch to high-level enums and then map to the window style / extended window style `DWORD`s as late as possible? Alternatively, consider linking to the docs (https://learn.microsoft.com/windows/win32/winmsg/window-styles and https://learn.microsoft.com/windows/win32/winmsg/extended-window-styles), but I don't feel strongly about that.

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
+  virtual LRESULT HandleMessage(HWND hwnd,
+                                UINT message,
+                                WPARAM wparam,
+                                LPARAM lparam);
 
   // Sets the focus to the child view window of |window|.
-  static void FocusViewOf(HostWindow* window);
+  static void FocusRootViewOf(HostWindow* window);
+
+  // Enables or disables this window and all its descendants.
```
loic-sharma: Consider:

```suggestion
  // Enables or disables mouse and keyboard input to this window and all its descendants.
```

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
+  void EnableWindowAndDescendants(bool enable);
+
+  // Returns the first enabled descendant window. If the current window itself
+  // is enabled, returns the current window.
+  HostWindow* FindFirstEnabledDescendant() const;
+
+  // Returns windows owned by this window.
+  std::vector<HostWindow*> GetOwnedWindows() const;
+
+  // Disables the window and all its descendants.
```
loic-sharma: Consider:

```suggestion
  // Disables mouse and keyboard input to the window and all its descendants.
```

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
+                                UINT message,
+                                WPARAM wparam,
+                                LPARAM lparam);
 
   // Sets the focus to the child view window of |window|.
-  static void FocusViewOf(HostWindow* window);
+  static void FocusRootViewOf(HostWindow* window);
+
+  // Enables or disables this window and all its descendants.
+  void EnableWindowAndDescendants(bool enable);
```
loic-sharma: Ideally this method would align with its disable counterpart. What do you think of naming the methods `EnableRecursively`? I don't feel strongly, feel free to skip if you prefer the current name.

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
+
+  // Returns the first enabled descendant window. If the current window itself
+  // is enabled, returns the current window.
```
loic-sharma: Should we add a comment that this returns null if the current window and all its children are disabled?

engine/src/flutter/shell/platform/windows/host_window_dialog.h
```diff
@@ -0,0 +1,40 @@
...
+#define FLUTTER_SHELL_PLATFORM_WINDOWS_HOST_WINDOW_DIALOG_H_
+
+#include "host_window.h"
+
+namespace flutter {
+class HostWindowDialog : public HostWindow {
+ public:
+  // Creates a dialog window.
+  //
+  // If |owner_window| is not null, the dialog will be modal to the owner.
```
loic-sharma, line 15: TODO: What does it mean to be modal to the owner? I assume you are destroyed if your owner is destroyed. Anything else?

engine/src/flutter/shell/platform/windows/window_manager.cc
```diff
@@ -45,6 +45,20 @@ FlutterViewId WindowManager::CreateRegularWindow(
...
 }
 
+FlutterViewId WindowManager::CreateDialogWindow(
+    const DialogWindowCreationRequest* request) {
+  auto window = HostWindow::CreateDialogWindow(
+      this, engine_, request->preferred_size, request->preferred_constraints,
+      request->title, request->parent);
+  if (!window || !window->GetWindowHandle()) {
+    FML_LOG(ERROR) << "Failed to create host window";
+    return -1;
```
loic-sharma, line 55: What happens if we return -1?

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -443,6 +450,210 @@ class RegularWindowControllerWin32 extends RegularWindowController {
...
+       _delegate = delegate,
+       _parent = parent,
+       super.empty() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    _handler = _DialogWindowMesageHandler(controller: this);
+    owner._addMessageHandler(_handler);
+    final int viewId = _Win32PlatformInterface.createDialogWindow(
```
loic-sharma, line 514: Should we throw an exception if `viewId` is `-1` here?

engine/src/flutter/shell/platform/windows/window_manager_unittests.cc
```diff
@@ -313,5 +315,80 @@ TEST_F(WindowManagerTest, CanSetWindowConstraintsWhileFullscreen) {
   EXPECT_EQ(actual_size.height, 500);
 }
 
+TEST_F(WindowManagerTest, CreateModelessDialogWindow) {
```
loic-sharma, line 318: Consider:

```suggestion
TEST_F(WindowManagerTest, CreateNonModalDialogWindow) {
```

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
```diff
@@ -0,0 +1,138 @@
...
+}
+
+LRESULT HostWindowDialog::HandleMessage(HWND hwnd,
+                                        UINT message,
+                                        WPARAM wparam,
+                                        LPARAM lparam) {
+  switch (message) {
+    case WM_DESTROY:
+      is_being_destroyed_ = true;
+      if (HostWindow* const owner_window = GetOwnerWindow()) {
```
loic-sharma, line 98: Let's say I have this window hierarchy:

```
1, regular window
└── 2, regular window
    └── 3, dialog window
```

And let's say I delete window 2. What will happen here? 

I expect that window 2 and 3 should both be destroyed, and that window 1 should be enabled for input. However, I suspect the current logic won't do that.

`WM_DESTROY`'s [docs](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-destroy) says:

> This message is sent first to the window being destroyed and then to the child windows (if any) as they are destroyed. During the processing of the message, it can be assumed that all child windows still exist.

In other words, I expect that when the dialog window receives `WM_DESTROY`, its `GetOwnerWindow` will return `nullptr`. If so, I expect that window 1 won't be enabled.

Let me know if I misunderstood how this works!

engine/src/flutter/shell/platform/windows/host_window_dialog.h
loic-sharma, line 15: Consider:

```suggestion
  // If |owner_window| is not null, the dialog will be modal to the owner. This also affects the dialog window's styling.
```

I don't feel strongly, feel free to skip.

engine/src/flutter/shell/platform/windows/host_window.cc
```diff
@@ -761,4 +701,140 @@ ActualWindowSize HostWindow::GetWindowContentSize(HWND hwnd) {
...
+    // Non-leaf window in the active path, disable it and process children.
+    EnableWindow(window_handle_, false);
+
+    // On same level of window hierarchy the most recently created window
+    // will remain enabled.
+    auto latest_child = *std::max_element(
+        children.begin(), children.end(), [](HostWindow* a, HostWindow* b) {
+          return a->view_controller_->view()->view_id() <
+                 b->view_controller_->view()->view_id();
+        });
```
loic-sharma, line 830: Does this work if the modal window's owner is not the most recently created window on its level of the window hierarchy? For example imagine I have this window hierarchy:

```
1, regular window
├── 2, regular window
└── 3, regular window
```

I then create a modal window that is owned by window 2.

```
1, regular window
├── 2, regular window
│   └── 4, newly created modal window
└── 3, regular window
```

Will this algorithm enable window 3 instead of window 4 (since view ID 3 is greater than view ID 2)?

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -60,17 +84,23 @@ class HostWindow {
...
 
   // Given a window identifier, returns the window content size of the
   // window.
   static ActualWindowSize GetWindowContentSize(HWND hwnd);
 
- private:
+  // Returns the owner window, or nullptr if none.
+  HostWindow* GetOwnerWindow() const;
+
+  // Processes modal state update for single layer of window hierarchy.
```
loic-sharma: Could you expand what this method does?

From my understanding, this is called recursively when a child dialog window is created or destroyed. This starts from the root and works its way down to the child dialog window (or its parent) to ensure the windows on this path have the correct enabled/disabled state. Is that correct?

If so, could you add more docs explaining this?

packages/flutter/lib/src/widgets/_window_win32.dart
loic-sharma: Could we add a quick comment to explain this? That's not immediately obvious :)

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -629,11 +873,19 @@ class _Win32PlatformInterface {
...
+  external _WindowConstraintsRequest preferredConstraints;
+  external ffi.Pointer<_Utf16> title;
+}
+
+/// Payload for the creation method used by [_Win32PlatformInterface.createDialogWindow].
+final class _DialogWindowCreationRequest extends ffi.Struct {
   external _WindowSizeRequest preferredSize;
   external _WindowConstraintsRequest preferredConstraints;
   external ffi.Pointer<_Utf16> title;
+  external HWND parent;
```
loic-sharma: Consider naming this `parentOrNull` (and `parent_or_null` on the C++ side) to make it extra clear that it is expected that we'll pass `nullptr`s down here. I don't feel strongly about this, feel free to skip.

engine/src/flutter/shell/platform/windows/window_manager.h
```diff
@@ -39,12 +39,19 @@ struct WindowConstraints {
...
   WindowSizeRequest preferred_size;
   WindowConstraints preferred_constraints;
   LPCWSTR title;
 };
 
+struct DialogWindowCreationRequest {
+  WindowSizeRequest preferred_size;
+  WindowConstraints preferred_constraints;
+  LPCWSTR title;
+  HWND parent;
```
loic-sharma: If you don't rename this to `parent_or_null`, please add a comment that this can be `nullptr` if the dialog doesn't have a parent.

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +115,63 @@ class HostWindow {
...
              WindowArchetype archetype,
-             std::unique_ptr<FlutterWindowsViewController> view_controller,
-             const BoxConstraints& constraints,
-             HWND hwnd);
+             DWORD window_style,
+             DWORD extended_window_style,
+             const BoxConstraints& box_constraints,
+             Rect const initial_window_rect,
+             LPCWSTR title,
+             HWND owner_window);
```
loic-sharma: Do you think we should make this a `std::optional<HWND>`?

I know HWND is a pointer itself that can be `nullptr`, but wrapping it in a `std::optional` would force people to remember to check for null. This might help reduce the likelihood of segfaults. What do you think?

engine/src/flutter/shell/platform/windows/host_window.cc
mattkae: @knopp and I spoke about this earlier, but a failure to make a view is most likely a fatal error and a precondition to the engine working properly. We decided that it should be asserted rather than handled, since it would be breaking your app anyway

engine/src/flutter/shell/platform/windows/host_window.h
mattkae: This API is protected, so I don't feel strongly about having a high-level enum here. The subclasses serve this purpose in my opinion as they "know" which styles are required to construct themselves

engine/src/flutter/shell/platform/windows/host_window.h
mattkae: I think so too!

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
mattkae: `GetOwnerWindow` will not return `nullptr` there because the handle is still valid until _after_ all owned windows are destroyed.

There is some other bug afoot here though that needs fixing in the UI, but the window does get destroyed properly

engine/src/flutter/shell/platform/windows/host_window.cc
mattkae: In this case, regular windows 2 and 3 would not be actual children of regular window 1, since regular window can't take a parent right now.

However, if we imagine that they are dialogs, then the latest dialog would be focused, which would not be a problem

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
```diff
@@ -0,0 +1,138 @@
...
+                                   LPCWSTR title,
+                                   std::optional<HWND> const& owner_window)
+    : HostWindow(
+          window_manager,
+          engine,
+          WindowArchetype::kDialog,
+          GetWindowStyleForDialog(owner_window),
+          GetExtendedWindowStyleForDialog(owner_window),
+          constraints,
+          [&]() -> Rect {
```
loic-sharma: This guy is a bit chonky. I'd consider moving this to a helper, like `GetWindowStyleForDialog` or `GetExtendedWindowStyleForDialog`

engine/src/flutter/shell/platform/windows/host_window.cc
loic-sharma: > if we imagine that they are dialogs, then the latest dialog would be focused, which would not be a problem

Sorry, I don't follow. If we imagine they are dialogs and 4 was newly created, would 4 be enabled instead of 3? (From my understanding, this algorithm would start at window 1, then recurse to 3 since 3 > 2, and thus never reach 4).

engine/src/flutter/shell/platform/windows/host_window.h
loic-sharma: Gotcha, thanks for the thoughts!

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
loic-sharma: Same comment for the similar callback pattern in regular window.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -474,28 +700,61 @@ class _Win32PlatformInterface {
...
+    BoxConstraints? preferredConstraints,
+    String? title,
+    HWND? parent,
+  ) {
+    final ffi.Pointer<_DialogWindowCreationRequest> request =
+        allocator<_DialogWindowCreationRequest>();
+    try {
+      request.ref.preferredSize.from(preferredSize);
+      request.ref.preferredConstraints.from(preferredConstraints);
+      request.ref.title = (title ?? 'Regular window').toNativeUtf16(allocator: allocator);
```
loic-sharma: `Regular window` might be confusing, maybe `Dialog window` instead?

```suggestion
      request.ref.title = (title ?? 'Dialog window').toNativeUtf16(allocator: allocator);
```

engine/src/flutter/shell/platform/windows/host_window.cc
mattkae: Yes it would, and I would argue that this is by design. If the main window opens up two dialogs (dialog 2 and dialog 3), and then dialog 2 opens up dialog 4, then I would still expect dialog 3 to have focus. After dialog 3 closes, dialog 4 should get focus.

In reality, 99% of users will not use this behavior, but this at least keeps it predictable.

loic-sharma: Ok gotcha, thanks for the explanation!

loic-sharma:
@mattkae FYI, I think this comment got missed due to GitHub's comment hiding: https://github.com/flutter/flutter/pull/176309#discussion_r2409040265 :)

loic-sharma (approved):
LGTM, nice work!

scnufan:
I wanna know when will the multiple-window will publish in release instead of show in engine ? 
In recent project, It's now a decision time to choose desktop_multi_window or waitting for the new flutter multiple-window。

So it's important to know the schedule of the multiple-window plan.

packages/flutter/lib/src/widgets/_window_win32.dart
mattkae: Oh yup, on it!

mattkae:
> I wanna know when will the multiple-window will publish in release instead of show in engine ? In recent project, It's now a decision time to choose desktop_multi_window or waitting for the new flutter multiple-window。
> 
> So it's important to know the schedule of the multiple-window plan.

Hiya. If you're asking about the timeline, we're working as fast as we can to land the initial window types. On win32, this is estimated to be some time in November, with macOS and Linux to follow.

Please keep in mind that this work will remain behind a feature flag until we have confidence in the interface.

kjxbyz:
Many important features in Flutter, including multi-window, do not have a timeline. In fact, our hope is not to require contributors to strictly follow the timeline, but to let us developers have a clear understanding when making technology choices.

packages/flutter/lib/src/widgets/_window_win32.dart
justinmc: Friendly nag for this comment (unless I missed it somewhere).

justinmc (approved):
Tentative LGTM because I want to take another shot at running the example locally when I get home tonight. No need to wait for me if the PR is ready to merge, though. I don't want to block this. I'll open issues if necessary.

justinmc:
I failed to get this running again. Last time `et build` was working for me but I couldn't get it to run. This time I tried to rebuild the engine and even `et build` wasn't working (something about a file in the perfetto directory was missing). I need to spend more time on it but don't let me hold up this PR.

mattkae:
> I failed to get this running again. Last time `et build` was working for me but I couldn't get it to run. This time I tried to rebuild the engine and even `et build` wasn't working (something about a file in the perfetto directory was missing). I need to spend more time on it but don't let me hold up this PR.

I'm currently busy getting CI to stop failing, so you may have time to try :laughing: But I will land beforehand regardless. We can make changes to the example app after if need be

dev/integration_tests/windowing_test/test_driver/main_test.dart
```diff
@@ -168,5 +169,11 @@ void main() {
       timeout: Timeout.none,
       onPlatform: {'linux': Skip('isMinimized is not supported on Wayland')},
     );
+
+    test('Can open dialog', () async {
+      await driver.requestData(jsonEncode({'type': 'open_dialog'}));
+      await driver.waitFor(find.byValueKey('close_dialog'));
+      await driver.requestData(jsonEncode({'type': 'close_dialog'}));
+    }, timeout: Timeout.none);
```
robert-ancell: This will need a `onPlatform: {'linux': Skip('Dialogs not supported on Linux')}` right?

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -27,19 +28,42 @@ class HostWindow {
...
       const WindowConstraints& preferred_constraints,
       LPCWSTR title);
 
+  // Creates a dialog Win32 window with a child view confined to its client
+  // area. |window_manager| is a pointer to the window manager that manages the
+  // |HostWindow|. |engine| is a pointer to the engine that manages
+  // the window manager. |preferred_size| is the preferred size of the window.
+  // |preferred_constraints| are the constraints set on the window's size.
+  // |title| is the title of the window. |parent| is the parent of this dialog,
+  // which can be `null`.
```
robert-ancell: `null` -> `nullptr` or null (not a valid constant)

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +117,70 @@ class HostWindow {
...
+                                UINT message,
+                                WPARAM wparam,
+                                LPARAM lparam);
 
   // Sets the focus to the child view window of |window|.
-  static void FocusViewOf(HostWindow* window);
+  static void FocusRootViewOf(HostWindow* window);
+
+  // Enables or disables mouse and keyboard input to this window and all its
+  // descendants
```
robert-ancell: Trailing '.'

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +117,70 @@ class HostWindow {
...
+
+  // Returns the first enabled descendant window. If the current window itself
+  // is enabled, returns the current window. If no window is enabled, returns
+  // |nullptr|.
+  HostWindow* FindFirstEnabledDescendant() const;
+
+  // Returns windows owned by this window.
+  std::vector<HostWindow*> GetOwnedWindows() const;
+
+  // Disables mouse and keyboard input to the window and all its descendants
```
robert-ancell: Trailing '.'

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -85,27 +117,70 @@ class HostWindow {
...
-  static void FocusViewOf(HostWindow* window);
+  static void FocusRootViewOf(HostWindow* window);
+
+  // Enables or disables mouse and keyboard input to this window and all its
+  // descendants
+  void EnableRecursively(bool enable);
+
+  // Returns the first enabled descendant window. If the current window itself
+  // is enabled, returns the current window. If no window is enabled, returns
+  // |nullptr|.
```
robert-ancell: |nullptr| -> `nullptr` (not a constant)

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -27,19 +28,42 @@ class HostWindow {
...
+  // Creates a dialog Win32 window with a child view confined to its client
+  // area. |window_manager| is a pointer to the window manager that manages the
+  // |HostWindow|. |engine| is a pointer to the engine that manages
+  // the window manager. |preferred_size| is the preferred size of the window.
+  // |preferred_constraints| are the constraints set on the window's size.
+  // |title| is the title of the window. |parent| is the parent of this dialog,
+  // which can be `null`.
+  //
+  // On success, a valid window handle can be retrieved
+  // via |HostWindow::GetWindowHandle|. |nullptr| will be returned
```
robert-ancell: |nullptr| -> `nullptr` (not an argument)

robert-ancell (approved):
I think a fix needed to stop the test failing on Linux, but otherwise just pedantic stuff.

jtmcdole:
This change looks to be super breaking the tree. `DialogCanNeverBeFullscreen` is the last message I see in several of the failing runs.

Can you please take over #177172 and find out why this test or the underlying code is having issues?

mattkae:
> This change looks to be super breaking the tree. `DialogCanNeverBeFullscreen` is the last message I see in several of the failing runs.
> 
> Can you please take over #177172 and find out why this test or the underlying code is having issues?

Yup! I'll look today

PR #176321, opened 2025-09-30
fix: support older git (ubuntu 22.04) in content hash

jtmcdole:
content aware hash was using --format which isn't present in older, but still supported operating systems. this pr removes the format string which basically reduces the output before hasing to:

```shell
100644 blob 198d80926b6e873c327f71350a0cdefee6a8402f	DEPS
040000 tree 139c1f10f92e4b9d4ac3ec7d4d27b2aa9775c5cd	engine
```

this format is still stable across all platforms and passed into `git hash-object` - which produces the actual hash fingerprint of the engine.

safety: this is the only scripts that produce this hash, so all downstream consumers keep consuming a sha1 output. Since this changes the sha, an engine version shouldn't exist for it and cocoon will build the artifacts for it.

fixes: #175265

### DIFF
--- a/bin/internal/content_aware_hash.ps1
+++ b/bin/internal/content_aware_hash.ps1
@@ -79,6 +79,6 @@ if (($currentBranch -ne "main") -and
 # 3. Out-File -NoNewline -Encoding ascii outputs 8bit ascii
 # 4. git hash-object with stdin from a pipeline consumes UTF-16, so consume
 #.   the contents of hash.txt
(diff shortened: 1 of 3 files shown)


jtmcdole:
Ah crud; cocoon probably won't produce the artifacts v2 for presubmit testing.

jtmcdole:
flutter/cocoon#4858 has to land first.

jtmcdole:
> [flutter/cocoon#4858](https://github.com/flutter/cocoon/pull/4858) has to land first.

will test again tomorrow assuming cocoon builds and deploys.

chingjun (approved):

PR #176323, opened 2025-09-30
Remove references to dart:js_util

fishythefish:
Flutter has been migrated from dart:js_util (which is unsupported by dart2wasm) to dart:js_interop, so we can remove/migrate all the remaining references to js_util.


fishythefish:
Hmm, the failing test (`js_access_test`) is probably worth looking into more deeply. It looks like the intent was to prevent un-audited libraries from accessing potentially unsafe APIs in `dart:js_util`. However, since Flutter has been migrated off `dart:js_util`, that check trivially passes. The equivalent APIs are found in `dart:js_interop_unsafe`, and it looks like several un-audited libraries use it.

cc @srujzs

srujzs:
Yeah, this is a good question. The list of audited libraries have been stale even before we migrated to `dart:js_interop_unsafe` (probably due to the use of the `package:js/js_util.dart` import instead). @mdebbar, do we find this test still useful, and if so, should we update the audited list to include all usages of `dart:js_interop_unsafe` (which contain many of the same members with "unsafe" access potential as `js_util`)?

mdebbar:
Thanks for catching this, @fishythefish!

Yes, `js_access_test` is still important and it should've been updated a while ago with `dart:js_interop_unsafe`.

Feel free to add `skip: true` for now. I'll file an issue to get it back in working order.

mdebbar (approved):
LGTM

srujzs:
Up to you @fishythefish, but it may be simple enough to update the allowlist in this PR. LGTM otherwise!

srujzs (approved):

fishythefish:
I'll go ahead and update the test's allowlist in this PR with a `TODO(176365)` to clean up those uses.

PR #176361, opened 2025-10-01
[macOS] Implement regular window

knopp:
*Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots.*

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -33,22 +33,22 @@ - (instancetype)initWithWindow:(NSWindow*)window
...
+  [self setContentSize:NSMakeSize(contentSize.width, contentSize.height)];
+}
+
+- (void)flutterSetConstraints:(FlutterWindowConstraints)constraints {
+  [self setContentMinSize:NSMakeSize(constraints.min_width, constraints.min_height)];
+  if (constraints.max_width > 0 && constraints.max_height > 0) {
+    [self setContentMaxSize:NSMakeSize(constraints.max_width, constraints.max_height)];
+  } else {
+    [self setContentMaxSize:NSMakeSize(CGFLOAT_MAX, CGFLOAT_MAX)];
   }
```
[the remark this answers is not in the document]
knopp: Disagree. I don't think there's much value in being able to specify maximum window size 0 in either dimension given that NSWindow has normally minimum height of 1 and minimum width at least the size of traffic lights.

examples/multiple_windows/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
+// This is a basic Flutter widget test.
```
mattkae: Was this added accidentally? We have no counter in this demo app :)

dev/integration_tests/windowing_test/lib/main.dart
```diff
@@ -59,20 +59,26 @@ void main() {
         controller.setConstraints(constraints);
       } else if (jsonMap['type'] == 'set_fullscreen') {
         controller.setFullscreen(true);
+        await Future.delayed(Duration(seconds: 1));
```
mattkae: I would try merging up and see if the async-ification makes this work for you without the delays :)

dev/integration_tests/windowing_test/macos/Runner/AppDelegate.swift
```diff
@@ -0,0 +1,20 @@
+import Cocoa
+import FlutterMacOS
+
+@main
+class AppDelegate: FlutterAppDelegate {
+  override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
+    return true
+  }
```
mattkae: Should this not be set in dart nowadays?

examples/multiple_windows/macos/Runner/AppDelegate.swift
```diff
@@ -0,0 +1,20 @@
+import Cocoa
+import FlutterMacOS
+
+@main
+class AppDelegate: FlutterAppDelegate {
+  override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
+    return true
+  }
```
mattkae: Is this needed too?

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,548 @@
...
+  ///  * [WindowingOwner], the abstract class that manages native windows.
+  @internal
+  WindowingOwnerMacOS() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    if (!Platform.isWindows) {
+      throw UnsupportedError('Only available on the Win32 platform');
+    }
```
mattkae:
```suggestion
    if (!Platform.isMacOS) {
      throw UnsupportedError('Only available on the macOS platform');
    }
```

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,548 @@
...
+production applications or plugins published to pub.dev.
+
+To try experimental windowing APIs:
+1. Switch to Flutter's main release channel.
+2. Turn on the windowing feature flag.
+
+See: https://github.com/flutter/flutter/issues/30701.
+''';
+
+/// The macOS implementation of the windowing API.
```
mattkae:
```suggestion
/// [WindowingOwner] implementation for macOS.
///
/// If [Platform.isMacOS] is false, then the constructor will throw an
/// [UnsupportedError].
///
/// {@macro flutter.widgets.windowing.experimental}
///
/// See also:
///
///  * [WindowingOwner], the abstract class that manages native windows.
```

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,548 @@
...
+  bool hasTopLevelWindows() {
+    return _activeControllers.isNotEmpty;
+  }
+
+  final List<BaseWindowController> _activeControllers = <BaseWindowController>[];
+
+  /// Returns the window handle for the given [view], or null is the window
+  /// handle is not available.
+  /// The window handle is a pointer to NSWindow instance.
+  static Pointer<Void> getWindowHandle(FlutterView view) {
```
mattkae, line 104: Does this need to be public? If not, let's hide it. If it should be, then we need to improve the docstring here

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,548 @@
...
+    BoxConstraints? preferredConstraints,
+    BaseWindowController? parent,
+    String? title,
+  }) {
+    throw UnimplementedError();
+  }
+}
+
+/// The macOS implementation of the regular window controller.
+class RegularWindowControllerMacOS extends RegularWindowController {
```
mattkae, line 130: Could this be private to this file? If it can, then let's make it private. Otherwise, we will need to improve the docstrings. See the win32 implementation for inspiration on the docstrings :)

mattkae (commented):
A few notes, but looks solid to me. I would get a macOS expert in here for a review of that side though

examples/multiple_windows/test/widget_test.dart
knopp: I think it was added by flutter create. Removing.

dev/integration_tests/windowing_test/macos/Runner/AppDelegate.swift
knopp: Good catch. We indeed control that logic from dart code now.

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: Fixed already, just forgot to push.

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: I'm not sure really. Is there good reason to return arbitrary window handle for situation where we don't have the controller?

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: It needs to be public. The idea is to let user cast to it in case we have platform specific functionality that we want to expose.

examples/multiple_windows/macos/Runner/AppDelegate.swift
knopp: Nope.

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,563 @@
...
+    );
+    rootView = flutterView;
+    if (title != null) {
+      setTitle(title);
+    }
+  }
+
+  /// Returns window handle for the current window.
+  /// The handle is a pointer to NSWindow instance.
+  Pointer<Void> getWindowHandle() {
```
mattkae, line 166:
```suggestion
  /// Returns window handle for the current window.
  ///
  /// The handle is a pointer to an NSWindow instance.
  Pointer<Void> getWindowHandle() {
```

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,563 @@
...
+/// Implementation of [RegularWindowController] for the macOS platform.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [RegularWindowController], the base class for regular windows.
+class RegularWindowControllerMacOS extends RegularWindowController {
+  /// Creates a new regular window controller for macOS. When this constructor
+  /// completes the FlutterView is created and framework is aware of it.
```
mattkae, line 132:
```suggestion
  /// Creates a new regular window controller for macOS.
  ///
  /// When this constructor completes the FlutterView is created and framework is aware of it.
```

mattkae: This is the general format of flutter constructors

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,563 @@
...
+  }) : _owner = owner,
+       _delegate = delegate,
+       super.empty() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    _onClose = NativeCallable<Void Function()>.isolateLocal(_handleOnClose);
+    _onResize = NativeCallable<Void Function()>.isolateLocal(_handleOnResize);
+    final int viewId = _MacOSPlatformInterface.createWindow(
```
mattkae, line 148: Can we add the `title` here instead of calling `setTitle` after the window has been created?

packages/flutter/lib/src/widgets/_window_macos.dart
mattkae: I can't think of one for now, but if you can, we can leave it public.

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -0,0 +1,563 @@
...
+  @override
+  bool hasTopLevelWindows() {
+    return _activeControllers.isNotEmpty;
+  }
+
+  final List<BaseWindowController> _activeControllers = <BaseWindowController>[];
+
+  /// Returns the window handle for the given [view], or null is the window
+  /// handle is not available.
+  /// The window handle is a pointer to NSWindow instance.
```
mattkae:
```suggestion
  /// Returns the window handle for the given [view], or null is the window
  /// handle is not available.
  /// 
  /// The window handle is a pointer to the NSWindow instance.
```

mattkae (commented):
Some further nits

wanjm:
Exception has occurred.
ArgumentError (Invalid argument(s): Couldn't resolve native function 'InternalFlutter_Window_GetTitle' in 'package:flutter/src/widgets/_window_macos.dart' : No asset with id 'package:flutter/src/widgets/_window_macos.dart' found. No available native assets. Attempted to fallback to process lookup. dlsym(RTLD_DEFAULT, InternalFlutter_Window_GetTitle): symbol not found.
)

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: Not sure why?

mattkae: It's a nit, but the title is technically an initialization parameter that can be sent up with the initial values. I'm not attached either way, but we made this particular change for Win32

dev/integration_tests/windowing_test/lib/main.dart
```diff
@@ -38,29 +38,6 @@ void main() {
...
-        controller.addListener(notificationHandler);
-
-        act();
-        await for (final _ in streamController.stream) {
-          if (predicate()) {
-            break;
-          }
-        }
-        controller.removeListener(notificationHandler);
-      }
```
mattkae, line 65: Instead of removing this, can we make it not wait on the MacOS platform only? This awaiting is very useful to prove the Win32 and Linux platforms. I would rather just do something different on MacOS in this method then have the other platforms change to accommodate MacOS

mattkae (changes requested):
We should add the integration tests to CI (if possible) and see if they pass.

To run the integration tests in CI, please add a devicelab runner similar to the one found at devicelab/bin/tasks/windowing_test_windows.dart.

Also, you will need to add it to `.ci.yaml`, e.g. https://github.com/flutter/flutter/blob/cd189a18da69490006591808cdbdaf6c2a8a7982/.ci.yaml#L7009

mattkae (approved):
All good by me! Looks like you have some CI things to address though

TESTOWNERS
```diff
@@ -306,6 +306,7 @@
 /dev/devicelab/bin/tasks/windows_home_scroll_perf__timeline_summary.dart @jonahwilliams @flutter/engine
 /dev/devicelab/bin/tasks/windows_startup_test.dart @loic-sharma @flutter/desktop
 /dev/devicelab/bin/tasks/windowing_test_linux.dart @robert-ancell @flutter/desktop
+/dev/devicelab/bin/tasks/windowing_test_mnacos.dart @knopp @flutter/desktop
```
Saviq:
```suggestion
/dev/devicelab/bin/tasks/windowing_test_macos.dart @knopp @flutter/desktop
```

Issue #30112, opened 2019-03-28
Return data with popUntil 


hgl:
Navigator.pop supports returning data, I wonder why popUntil doesn't?

Ideally it should also accept a result argument, and pass it to the pop method it calls.

Would flutter team accept a such proposal?

hungnt-kayac:
i have a workaround solution here. Hope that's useful for you
https://medium.com/@hungregistermail/flutter-popuntil-with-return-data-ece825aedbd0

Nico04:
> i have a workaround solution here. Hope that's useful for you
> https://medium.com/@hungregistermail/flutter-popuntil-with-return-data-ece825aedbd0

But this solution doesn't skip the animations right ?

afi-mobile:
This thread was opened nearly a year ago. Is there any feedback on whether support is going to be added for adding a result to popUntil?

topnax:
I would really appreciate this feature as well. I have no clue why is this not implemented yet.

Nico04:
It will be probably implemented with the new [Navigator 2.0](https://github.com/flutter/flutter/issues/45938) ?

chunhtai:
With the page API it is possible, you will be able to decide which return value you would like to pass in to each removed route by using transition delegate, see https://github.com/flutter/flutter/pull/50362/files#diff-c798125dc01f08ce6bd3a0931482078bR810. You will get to choose how route exit the route stack, and you can specify the result when you do that. This api will be surface with Router Widget where you should be able to specify transition delegate for the MaterialApp/WidgetApp/CupterinoApp. I hope this solve your use case, and there is currently no plan to update the existing popuntil api. Let me know if you have further question.

For more info place check out flutter.dev/go/navigator-with-router

namchuai:
It's seems like an overkill but I'm using provider for this.

moneer-muntazah:
I used `pushNamedAndRemoveUntil` Go back to the screen behind the one in question, and use `arguments`  to pass your values. Better then trying to keep track of state.

tudor07:
> I fix it using shared_preferences, you can check the example and explanation here: https://stackoverflow.com/a/71426281/8291686

Very ugly for something that should've been provided out of the box by Flutter.

rrifafauzikomara:
> > I fix it using shared_preferences, you can check the example and explanation here: https://stackoverflow.com/a/71426281/8291686
> 
> Very ugly for something that should've been provided out of the box by Flutter.

[...]

PR #169341, opened 2025-05-23
Add `Navigator.popUntilWithResult`

alex-medinsh:
This PR adds a new `Navigator.popUntilWithResult` method that allows to pass a value to the last `pop` call.

Fixes https://github.com/flutter/flutter/issues/30112

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #176380, opened 2025-10-02
fix: Return data with popUntil

shang1219178163:
fix problem: https://github.com/flutter/flutter/issues/30112


shang1219178163:
Return data with popUntil  #[30112](https://github.com/flutter/flutter/issues/30112)

justinmc (commented):
This is potentially a breaking change, but you're probably right that we should allow result data with popUntil. What if we added a new popUntilWithResult and deprecated popUntil? This is the path we've taken with other methods like [onPopWithResult](https://main-api.flutter.dev/flutter/widgets/NavigatorPopHandler/onPopWithResult.html).

Also, this will need tests.

justinmc:
Actually, there is already an older PR open that solves this problem. We'll prefer that PR since it was opened first, but if not then I'll come back to this one. Thank you for looking at the problem anyway! https://github.com/flutter/flutter/pull/169341

PR #176384, opened 2025-10-02
feat: add interfaces for overlay window

rkishan516:
## What's new?
* Added the concept of overlay to the windowing API. New symbols:
  * OverlayWindowControllerDelegate
  * OverlayWindowController
  * WindowingOwner.createOverlayWindowController
  * OverlayWindow

*  This pull request ONLY adds the idea to the API. Win32 implementation to follow :)

## Design doc

You can read more about overlay window [here](https://flutter.dev/go/multi-window-overlay).

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/multiple_windows/lib/app/main_window.dart
+++ b/examples/multiple_windows/lib/app/main_window.dart
@@ -141,6 +141,7 @@ class _WindowsTable extends StatelessWidget {
         controller: tooltip,
       ),
       PopupWindowController() => null,
+      OverlayWindowController() => null,
     };
   }
 
@@ -150,6 +151,7 @@ class _WindowsTable extends StatelessWidget {
       DialogWindowController() => 'Dialog',
       TooltipWindowController() => 'Tooltip',
       PopupWindowController() => 'Popup',
+      OverlayWindowController() => 'Overlay',
     };
   }
 
--- a/examples/multiple_windows/lib/app/window_content.dart
+++ b/examples/multiple_windows/lib/app/window_content.dart
@@ -49,6 +49,7 @@ class WindowContent extends StatelessWidget {
         ),
       ),
       PopupWindowController() => throw UnimplementedError(),
+      OverlayWindowController() => throw UnimplementedError(),
     };
   }
 }
--- a/packages/flutter/lib/src/widgets/_window.dart
+++ b/packages/flutter/lib/src/widgets/_window.dart
@@ -422,6 +422,50 @@ mixin class DialogWindowControllerDelegate {
   }
 }
 
+/// Delegate class for overlay window controller.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [OverlayWindowController], the controller that creates and manages overlay windows.
+///  * [OverlayWindow], the widget for an overlay window.
+@internal
+mixin class OverlayWindowControllerDelegate {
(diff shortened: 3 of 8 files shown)


mattkae (changes requested):
Hiya! I am hesitant to land this before popups and tooltips. The positioner API is subject to change and requires the context of the other window types to be meaningful. In the meantime, could we hold off on implementing this type until those other types land? That should be happening over the course of this month. This type would also be best supported as a plugin in our opinion, but we can discuss this further once we land the other types

rkishan516:
> Hiya! I am hesitant to land this before popups and tooltips. The positioner API is subject to change and requires the context of the other window types to be meaningful. In the meantime, could we hold off on implementing this type until those other types land? That should be happening over the course of this month. This type would also be best supported as a plugin in our opinion, but we can discuss this further once we land the other types

Hey @mattkae, I agree that's why keeping this as draft. Once all 4 main window types are merged, then we can discuss and update PR accordingly.

wanjm:
will multi-view be supported? 
1. can i use one view to display flutter in the raw Application ( such as developed by QT or MFC) or something like  show flutter in one DIV while the while applicaiton is still using javascript;

2. can i create a native view in flutter to display native things such as webview;

rkishan516:
> will multi-view be supported?
> 
> 1. can i use one view to display flutter in the raw Application ( such as developed by QT or MFC) or something like  show flutter in one DIV while the while applicaiton is still using javascript;
> 2. can i create a native view in flutter to display native things such as webview;

Hey @wanjm, its currently just another window type for multi window APIs, just like newly introduced RegularWindow in `master`.

wanjm:
> > will multi-view be supported?
> > 
> > 1. can i use one view to display flutter in the raw Application ( such as developed by QT or MFC) or something like  show flutter in one DIV while the while applicaiton is still using javascript;
> > 2. can i create a native view in flutter to display native things such as webview;
> 
> Hey @wanjm, its currently just another window type for multi window APIs, just like newly introduced RegularWindow in `master`.

yes, i know , so  i just ask is there any plan for this ?

rkishan516:
> yes, i know , so i just ask is there any plan for this ?

No, currently not. Currently we are not even sure that this should be part of framework or just external plugin.

Piinks:
Hey @rkishan516! Is this ready to move forward now? Is there a blocking issue we can link to?

rkishan516:
> Hey @rkishan516! Is this ready to move forward now? Is there a blocking issue we can link to?

Hey @Piinks, I think we should be good to go, @mattkae do you think it will be a good idea to expose some kind of APIs which will allow devs to create custom windows in packages/apps? (Given the fact that getWindowHandle is already there to modify window)? If we do that than we can keep core 4 window type in framework and other types like overlay window can be 3rd party or 1st party packages. WDYT?

Saviq:
> do you think it will be a good idea to expose some kind of APIs which will allow devs to create custom windows in packages/apps?

That's exactly the idea :)

Check Matej's approach here: https://github.com/knopp/custom_window/

rkishan516:
> > do you think it will be a good idea to expose some kind of APIs which will allow devs to create custom windows in packages/apps?
> 
> That's exactly the idea :)
> 
> Check Matej's approach here: https://github.com/knopp/custom_window/

Oh perfect, Then I will close this MR and try to create a package for overlay window.

PR #176393, opened 2025-10-02
feat(assets): add platform-specific asset filtering in pubspec.yaml

### LINKED ISSUE #65065: Bundling assets only on a specific platform (and remove assets on another platform)
**NOTE: This is *not* a duplicate!** My previous issue https://github.com/flutter/flutter/issues/64993 is closed because of wrongly marked as a duplicate. Thus I have no choice but to create it again :/

(Why not duplicate of #8230 : I want to bundle those 1MB photos only in Android. Thus, in the final .ipa file of iOS, there are no such photos taking space.)

## Steps to Reproduce

Hi thanks for the wonderful Flutter! I have some images (1.1MB) that should only be used in Android, not iOS. Therefore, I want to *remove* those images from the IPA file in iOS. I have read the official doc, but it did not mention things like this :/

**Expected results:** 

Give some options, such as (pubspec.yaml):
```
flutter:
  assets:
    android:
      - some_images_only_in_android/
    ios:
      - some_images_only_in_ios/
   all:
      - some_images_for_both/
```

**Actual results:** No such feature

<details>
  <summary>Logs</summary>

```
```

```
```

```
```

</details>

hm21:
## Description

This PR introduces platform-specific asset support in `pubspec.yaml`.

Currently, Flutter does not allow specifying which platforms an asset should be included for.  
This results in all declared assets being bundled for every target platform, even if some are irrelevant (e.g. desktop-only or mobile-only images).  

### What this PR changes
- Adds a new optional `platforms` field under each asset in `pubspec.yaml`.
- The field accepts a list of strings (platform identifiers, e.g. `["android", "ios", "web", "windows", "macos", "linux"]`).
- Assets with a `platforms` restriction are only included in the bundle when building for a matching platform.
- Invalid values (non-strings or unknown platform names) log an error.

### Example
```yaml
flutter:
  assets:
    - path: assets/logo.png
   
    - path: assets/web_worker.js
      platforms: [web]

    - path: assets/desktop_icon.png
      platforms: [windows, linux, macos]
```

#### Before
All assets (`logo.png`, `web_worker.js`, `desktop_icon.png`) are bundled into **every build**, regardless of platform.

#### After
- `logo.png` is included on all platforms.
- `web_worker.js` is included only on web builds.
- `desktop_icon.png` is included only on desktop builds.

### Why this is useful
This significantly improves bundle size, prevents unused resources from being shipped, and gives developers better control over asset management.  

## Issues
Fixes #65065

## Reviewer note
Would a design document be helpful for this change, or is the current explanation sufficient?

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/lib/src/asset.dart
+++ b/packages/flutter_tools/lib/src/asset.dart
@@ -195,7 +195,7 @@ abstract class AssetBundle {
     String manifestPath = defaultManifestPath,
     required String packageConfigPath,
     bool deferredComponentsEnabled = false,
-    TargetPlatform? targetPlatform,
+    required TargetPlatform targetPlatform,
     String? flavor,
     bool includeAssetsFromDevDependencies = false,
   });
@@ -326,7 +326,7 @@ class ManifestAssetBundle implements AssetBundle {
     FlutterProject? flutterProject,
     required String packageConfigPath,
     bool deferredComponentsEnabled = false,
-    TargetPlatform? targetPlatform,
+    required TargetPlatform targetPlatform,
     String? flavor,
     bool includeAssetsFromDevDependencies = false,
   }) async {
@@ -402,6 +402,7 @@ class ManifestAssetBundle implements AssetBundle {
           assetBasePath,
           wildcardDirectories,
           flutterProject.directory,
+          targetPlatform: targetPlatform,
           flavor: flavor,
         );
     if (!_splitDeferredAssets || !deferredComponentsEnabled) {
@@ -716,7 +717,7 @@ class ManifestAssetBundle implements AssetBundle {
     }
   }
 
-  void _setLicenseIfChanged(String combinedLicenses, TargetPlatform? targetPlatform) {
+  void _setLicenseIfChanged(String combinedLicenses, TargetPlatform targetPlatform) {
     // On the web, don't compress the NOTICES file since the client doesn't have
     // dart:io to decompress it. So use the standard _setIfChanged to check if
     // the strings still match.
@@ -838,6 +839,7 @@ class ManifestAssetBundle implements AssetBundle {
     String assetBasePath,
     List<Uri> wildcardDirectories,
     Directory projectDirectory, {
+    required TargetPlatform targetPlatform,
     String? flavor,
   }) {
     final List<DeferredComponent>? components = flutterManifest.deferredComponents;
@@ -859,6 +861,7 @@ class ManifestAssetBundle implements AssetBundle {
             componentAssets,
             assetsEntry.uri,
             flavors: assetsEntry.flavors,
+            platforms: assetsEntry.platforms,
             transformers: assetsEntry.transformers,
           );
         } else {
@@ -870,13 +873,15 @@ class ManifestAssetBundle implements AssetBundle {
             componentAssets,
             assetsEntry.uri,
             flavors: assetsEntry.flavors,
+            platforms: assetsEntry.platforms,
             transformers: assetsEntry.transformers,
           );
         }
       }
 
       componentAssets.removeWhere(
-        (_Asset asset, List<_Asset> variants) => !asset.matchesFlavor(flavor),
+        (_Asset asset, List<_Asset> variants) =>
+            !asset.matchesFlavor(flavor) || !asset.matchesPlatform(targetPlatform),
       );
       deferredComponentsAssetVariants[component.name] = componentAssets;
     }
@@ -1018,7 +1023,7 @@ class ManifestAssetBundle implements AssetBundle {
     FlutterManifest flutterManifest,
     List<Uri> wildcardDirectories,
     String assetBase,
-    TargetPlatform? targetPlatform, {
+    TargetPlatform targetPlatform, {
     String? packageName,
     Package? attributedPackage,
     required String? flavor,
@@ -1039,6 +1044,7 @@ class ManifestAssetBundle implements AssetBundle {
           packageName: packageName,
           attributedPackage: attributedPackage,
           flavors: assetsEntry.flavors,
+          platforms: assetsEntry.platforms,
           transformers: assetsEntry.transformers,
         );
       } else {
@@ -1052,6 +1058,7 @@ class ManifestAssetBundle implements AssetBundle {
           packageName: packageName,
           attributedPackage: attributedPackage,
           flavors: assetsEntry.flavors,
+          platforms: assetsEntry.platforms,
           transformers: assetsEntry.transformers,
         );
       }
@@ -1063,6 +1070,15 @@ class ManifestAssetBundle implements AssetBundle {
           'Skipping assets entry "${asset.entryUri.path}" since '
           'its configured flavor(s) did not match the provided flavor (if any).\n'
           'Configured flavors: ${asset.flavors.join(', ')}\n',
+        );
+        return true;
+      }
+      if (!asset.matchesPlatform(targetPlatform)) {
+        _logger.printTrace(
+          'Skipping assets entry "${asset.entryUri.path}" since '
+          'its configured platform(s) did not match the target platform.\n'
+          'Configured platforms: ${asset.platforms.join(', ')}\n'
+          'Target platform: ${targetPlatform.osName}\n',
         );
         return true;
       }
@@ -1101,6 +1117,7 @@ class ManifestAssetBundle implements AssetBundle {
         attributedPackage: attributedPackage,
         assetKind: AssetKind.shader,
         flavors: <String>{},
+        platforms: <String>{},
         transformers: <AssetTransformerEntry>[],
       );
     }
@@ -1116,6 +1133,7 @@ class ManifestAssetBundle implements AssetBundle {
           attributedPackage,
           assetKind: AssetKind.font,
           flavors: <String>{},
+          platforms: <String>{},
           transformers: <AssetTransformerEntry>[],
         );
         final File baseAssetFile = baseAsset.lookupAssetFile(_fileSystem);
@@ -1142,6 +1160,7 @@ class ManifestAssetBundle implements AssetBundle {
     String? packageName,
     Package? attributedPackage,
     required Set<String> flavors,
+    required Set<String> platforms,
     required List<AssetTransformerEntry> transformers,
   }) {
     final String directoryPath;
@@ -1174,6 +1193,7 @@ class ManifestAssetBundle implements AssetBundle {
         attributedPackage: attributedPackage,
         originUri: assetUri,
         flavors: flavors,
+        platforms: platforms,
         transformers: transformers,
       );
     }
@@ -1191,6 +1211,7 @@ class ManifestAssetBundle implements AssetBundle {
     Package? attributedPackage,
     AssetKind assetKind = AssetKind.regular,
     required Set<String> flavors,
+    required Set<String> platforms,
     required List<AssetTransformerEntry> transformers,
   }) {
     final _Asset asset = _resolveAsset(
@@ -1202,6 +1223,7 @@ class ManifestAssetBundle implements AssetBundle {
       assetKind: assetKind,
       originUri: originUri,
       flavors: flavors,
+      platforms: platforms,
       transformers: transformers,
     );
 
@@ -1225,6 +1247,7 @@ class ManifestAssetBundle implements AssetBundle {
             package: attributedPackage,
             kind: assetKind,
             flavors: flavors,
+            platforms: platforms,
             transformers: transformers,
           ),
         );
@@ -1335,6 +1358,7 @@ class ManifestAssetBundle implements AssetBundle {
     Uri? originUri,
     AssetKind assetKind = AssetKind.regular,
     required Set<String> flavors,
+    required Set<String> platforms,
     required List<AssetTransformerEntry> transformers,
   }) {
     _ensureAssetPathIsValid(assetsBaseDir: assetsBaseDir, assetUri: assetUri);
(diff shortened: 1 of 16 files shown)


packages/flutter_tools/lib/src/flutter_manifest.dart
```diff
@@ -894,6 +903,38 @@ class AssetsEntry {
     return _parseList<String>(yaml, _flavorKey, 'String');
   }
 
+  static (List<String>? platforms, List<String> errors) _parsePlatformsSection(Object? yaml) {
+    if (yaml == null) {
+      return (null, <String>[]);
+    }
+
+    final (List<String>?, List<String>) result = _parseList<String>(yaml, _platformsKey, 'String');
```
bkonyi: Nit: this can just be

```dart
final (List<String>? platforms, List<String> errors) = _parseList<String>(...);
```

It's better to avoid positional accesses (e.g., `$1`) whenever possible.

packages/flutter_tools/lib/src/flutter_manifest.dart
```diff
@@ -894,6 +903,38 @@ class AssetsEntry {
...
+    final List<String>? platforms = result.$1;
+    final List<String> errors = result.$2;
+
+    if (platforms != null) {
+      final List<String> validPlatforms = [];
```
bkonyi: Nit: prefer initializing collections like this

```dart
final validPlatforms = <String>[];
```

packages/flutter_tools/test/general.shard/build_system/targets/assets_test.dart
```diff
@@ -461,6 +461,162 @@ flutter:
...
+      final File file = fileSystem.file('${environment.buildDir.path}/flutter_assets/$filePath');
+      final bool exists = file.existsSync();
+      if (exists) {
+        file.deleteSync();
+      }
+
+      return exists;
+    }
+
+    testUsingContext(
```
bkonyi: Prefer making this a `group` which uses the loop to define tests for each individual platform. That way we don't need to include the delete logic above as the file system will be destroyed between tests and won't leak state across tests for individual platforms.

packages/flutter_tools/test/general.shard/build_system/targets/assets_test.dart
```diff
@@ -461,6 +461,162 @@ flutter:
...
+        }
+      },
+      overrides: <Type, Generator>{
+        FileSystem: () => fileSystem,
+        ProcessManager: () => FakeProcessManager.any(),
+      },
+    );
+
+    testUsingContext(
+      'skips assets for non-matching platform',
```
bkonyi: Same comment here and below.

bkonyi (approved):
Thanks for your contribution!

Overall, this LGTM with some minor comments that should be addressed.

packages/flutter_tools/lib/src/flutter_manifest.dart
chingjun: The section that finds the valid and invalid platforms can be simplified to:

```dart
final validPlatforms = platforms.toSet().intersection(_kValidPluginPlatforms);
final invalidPlatforms = platforms.toSet().difference(_kValidPluginPlatforms);
```

Also, if `errors.isNotEmpty`, there isn't a reason to return the first value, right? (Based on how other functions in this file behave)

This can be further simplified to:
```dart
final invalidPlatforms = platforms.toSet().difference(_kValidPluginPlatforms);
if (invalidPlatforms.isNotEmpty) {
  return (null, <String>[
    ...
  ]);
}
return (platforms, errors);
```

packages/flutter_tools/lib/src/flutter_manifest.dart
```diff
@@ -894,6 +903,38 @@ class AssetsEntry {
     return _parseList<String>(yaml, _flavorKey, 'String');
   }
 
+  static (List<String>? platforms, List<String> errors) _parsePlatformsSection(Object? yaml) {
```
chingjun, line 910: I know this is a private function, and all the other similarly named functions in this file do not have a comment, but I think it would be really helpful for future readers to add a comment in these functions, since this function is returning a tuple, and this is not a well-established pattern in Dart.

Something like:
```dart
/// Parses the "platforms" section in the pubspec.yaml file. If it encounters an error while
/// parsing the file, the returned error will be non-empty, and the platforms will be null.
```

packages/flutter_tools/lib/src/flutter_manifest.dart
```diff
@@ -894,6 +903,38 @@ class AssetsEntry {
...
+    if (yaml == null) {
+      return (null, <String>[]);
+    }
+
+    final (List<String>?, List<String>) result = _parseList<String>(yaml, _platformsKey, 'String');
+    final List<String>? platforms = result.$1;
+    final List<String> errors = result.$2;
```
chingjun: Consider adding an early return here.

```dart
if (errors.isNotEmpty) {
  return (null, errors);
}
```

packages/flutter_tools/lib/src/build_system/targets/assets.dart
```diff
@@ -110,6 +111,17 @@ Future<Depfile> copyAssets(
...
   await Future.wait<void>(
     assetEntries.entries.map<Future<void>>((MapEntry<String, AssetBundleEntry> entry) async {
+      // Determine the current build target (e.g. "windows", "android", "ios")
+      // and compare it against the asset's declared `platforms` set from
+      // pubspec.yaml. If the asset specifies one or more platforms and the
+      // current target is not included, skip processing this asset to ensure
+      // it is not bundled for the wrong platform.
+      final String target = targetPlatform.osName;
+      final Set<String> targetPlatforms = entry.value.platforms;
+      if (targetPlatforms.isNotEmpty && !targetPlatforms.contains(target)) {
```
chingjun: Is there any reason why we're doing the filtering here, instead of in the `AssetBundle.build()` function like how `flavors` is implemented?

If it's done there, we don't need to add the `platforms` entry in the `AssetBundleEntry` class.

packages/flutter_tools/lib/src/flutter_manifest.dart
hm21: Fixed✅

packages/flutter_tools/test/general.shard/build_system/targets/assets_test.dart
hm21: Fixed✅

packages/flutter_tools/test/general.shard/build_system/targets/assets_test.dart
hm21: Fixed✅

packages/flutter_tools/lib/src/build_system/targets/assets.dart
hm21: You’re right it makes more sense to handle the filtering in AssetBundle.build() just like flavors.
I’ve moved the logic there✅

packages/flutter_tools/lib/src/flutter_manifest.dart
hm21: Fixed✅

packages/flutter_tools/lib/src/flutter_manifest.dart
hm21: Code changed, this comment became obsolete ✅

packages/flutter_tools/lib/src/flutter_manifest.dart
hm21: Fixed✅

packages/flutter_tools/lib/src/flutter_manifest.dart
hm21: Good catch, you're right this makes the code cleaner✅

packages/flutter_tools/lib/src/asset.dart
```diff
@@ -1472,11 +1504,24 @@ class _Asset {
     return flavors.contains(flavor);
   }
 
+  bool matchesPlatform(TargetPlatform? targetPlatform) {
+    if (platforms.isEmpty) {
+      return true;
+    }
+
+    if (targetPlatform == null) {
```
chingjun: In [run_hot.dart](https://github.com/flutter/flutter/blob/3fd81edbf1e015221e143c92b2664f4371bdc04a/packages/flutter_tools/lib/src/run_hot.dart#L494), the `targetPlatform` is not passed to `assetBundle.build()`. The way we handle `targetPlatform == null` here would make platform-specific assets to not be updated correctly on hot reload and hot restart.

I think we should pass the `targetPlatform` in run_hot.dart.


There is another one in [commands/test.dart](https://github.com/flutter/flutter/blob/3fd81edbf1e015221e143c92b2664f4371bdc04a/packages/flutter_tools/lib/src/commands/test.dart#L768). I am less sure what to do with this, should we include all platform specific assets in tests, or none?

chingjun (approved):
Thank you for the fixes, this now looks a lot cleaner!

There is still an open question about what assets should be included in tests, but even with that unanswered I would still be ok landing this PR, and figure that out in the future.

----

The following are fully optional. It is out of scope for this PR, and is is fully optional whether you want to fix it in this PR, in a separate PR, or leave it for now.

The PR led me to think whether `targetPlatform` should still be an optional parameter in `assetBundle.build()`. Perhaps we should make it a required parameter.

Again, fully optional.

packages/flutter_tools/lib/src/asset.dart
hm21: Good point. I think it's best to implement it directly in this PR as well so we don't have to revisit it later.

I checked [run_hot.dart](https://github.com/flutter/flutter/blob/3fd81edbf1e015221e143c92b2664f4371bdc04a/packages/flutter_tools/lib/src/run_hot.dart#L494) and it’s straightforward, so I added the available `TargetPlatform`.

I also checked [commands/test.dart](https://github.com/flutter/flutter/blob/3fd81edbf1e015221e143c92b2664f4371bdc04a/packages/flutter_tools/lib/src/commands/test.dart#L768) and saw that `TargetPlatform.tester` is used [here](https://github.com/flutter/flutter/blob/3fd81edbf1e015221e143c92b2664f4371bdc04a/packages/flutter_tools/lib/src/commands/test.dart#L780), so I used it as well and updated the logic from `matchesPlatform`. It now returns `true` when the `targetPlatform` is `null` or  `TargetPlatform.tester`. I could revert that, but at the moment I don’t see a reason why we shouldn't always include all assets if `targetPlatform` is `null`.
Initially, I was a bit concerned about integration tests, but as far as I can tell, `_buildTestAsset` is only used optional for unit and widget tests, so using `TargetPlatform.tester` seems like the best option here, right? One thing we could consider is adding a `stringArg` that lets developers specify a specific `targetPlatform`, but personally, I think that might be overkill since devs can already cover that through integration tests. However, if you guys think it’s useful, I could add it.

> The PR led me to think whether targetPlatform should still be an optional parameter in assetBundle.build(). Perhaps we should make it a required parameter.

We could make it required now or leave it as is, since if no `targetPlatform` is set, it will include all assets, which I assume is the desired behavior. However, I just noticed that if we make it required, we’ll also need to update the following test files: `asset_bundle_flavors_test.dart`, `asset_bundle_package_fonts_test.dart`, `asset_test.dart`, `asset_bundle_package_test.dart`, `asset_bundle_variant_test.dart`, and `asset_bundle_test.dart`.

Since these are all test files, I assume we’d just need to add `TargetPlatform.tester` in each of them. However I'm not as deep into the code as you guys, so I’ll leave the decision up to you, just let me know if I should go ahead and make it required.


@bkonyi What’s your opinion on this, should we make `targetPlatform` a required parameter? And what do you think about the part in `commands/test.dart`? Is it fine for you that I just add `TargetPlatform.tester` there, or would you prefer the extended version with the `stringArg` I mentioned above?

bkonyi: I don't think it's possible to have two assets with the same name specified for different platforms, so I think we should be safe to return all assets for tests.

Let's go ahead and make `targetPlatform` a required parameter.

hm21: Done✅

hm21:
Okay `targetPlatform` is now required, so it’s ready for review. I think the PR is complete now, but if you have any other suggestions or ideas to extend it before merging, let me know. Thanks again for your reviews.

packages/flutter_tools/lib/src/asset.dart
```diff
@@ -1472,11 +1504,20 @@ class _Asset {
     return flavors.contains(flavor);
   }
 
+  bool matchesPlatform(TargetPlatform? targetPlatform) {
```
chingjun: I think this can be `TargetPlatform targetPlatform` now, right?

chingjun (approved):
LGTM!

One more minor nit, and you probably want to rebase your branch as it's now 200+ commits behind HEAD.

Thank you again for your contribution, and specifically for adding the changes to make the code base more robust. Thank you!

packages/flutter_tools/lib/src/asset.dart
hm21: Fixed✅

chingjun:
I just fixed the Google Testing for this PR. Good to be submitted now. @bkonyi do you want to take a final look at it before adding the autosubmit tag?

jmagman:
Thank you so much for the contribution! This needs documentation, I just filed https://github.com/flutter/website/issues/12966.

hm21:
@jmagman wow I didn't imagine you guys would mention me, but I really appreciate it, that's really cool :)

PR #176428, opened 2025-10-02
Announce text and button together when DropdownMenu is treated as a button

QuncCccccc:
Fixes https://github.com/flutter/flutter/issues/175950.

This PR is to make screen reader announce the text field content and the trailing button together on `DropdownMenu`. When the `DropdownMenu` cannot be focused (`canRequestFocus()` is set to false), the "text field" should be treated as if it is a button, and the trailing button should **not** be announced separately. When the text field is focusable, it's reasonable to separate text field and the trailing button, so I leave it as is.

The native app doesn't announce "expanded"/"collapsed", so I use `Semantics.hint` to achieve the goal.

This demo is to show the screen reader announcement.

https://github.com/user-attachments/assets/e7a9da8f-acf8-4018-a778-1ded1b07103c

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -1129,18 +1129,26 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
         final Widget trailingButton = widget.showTrailingIcon
             ? Padding(
                 padding: isCollapsed ? EdgeInsets.zero : const EdgeInsets.all(4.0),
-                child: IconButton(
-                  focusNode: _trailingIconButtonFocusNode,
-                  isSelected: controller.isOpen,
-                  constraints: widget.inputDecorationTheme?.suffixIconConstraints,
-                  padding: isCollapsed ? EdgeInsets.zero : null,
-                  icon: widget.trailingIcon ?? const Icon(Icons.arrow_drop_down),
-                  selectedIcon: widget.selectedTrailingIcon ?? const Icon(Icons.arrow_drop_up),
-                  onPressed: !widget.enabled
-                      ? null
-                      : () {
-                          handlePressed(controller);
-                        },
+                child: ExcludeSemantics(
+                  // When the text field is treated as a button (i.e., it can
+                  // not be focused), the trailing button should become part of
+                  // the text field button by excluding semantics. Otherwise,
+                  // it will inappropriately announce whether this icon button
+                  // is selected or not.
+                  excluding: !canRequestFocus(),
+                  child: IconButton(
+                    focusNode: _trailingIconButtonFocusNode,
+                    isSelected: controller.isOpen,
+                    constraints: widget.inputDecorationTheme?.suffixIconConstraints,
+                    padding: isCollapsed ? EdgeInsets.zero : null,
+                    icon: widget.trailingIcon ?? const Icon(Icons.arrow_drop_down),
+                    selectedIcon: widget.selectedTrailingIcon ?? const Icon(Icons.arrow_drop_up),
+                    onPressed: !widget.enabled
+                        ? null
+                        : () {
+                            handlePressed(controller);
+                          },
+                  ),
                 ),
               )
             : const SizedBox.shrink();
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1150,48 +1158,52 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
-            helperText: widget.helperText,
-            errorText: widget.errorText,
-            prefixIcon: widget.leadingIcon != null
-                ? SizedBox(key: _leadingKey, child: widget.leadingIcon)
-                : null,
-            suffixIcon: widget.showTrailingIcon ? trailingButton : null,
-          ).applyDefaults(effectiveInputDecorationTheme),
-          restorationId: widget.restorationId,
+        final Widget textField = Semantics(
+          button: !canRequestFocus(),
```
chunhtai: some platform may still treat this as textfield if both istextField and isButton are true.

chunhtai: Can this be focusable but not enabled?

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1150,48 +1158,52 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
+          hint: _controller.isOpen ? 'Expanded' : 'Collapsed',
```
chunhtai: also set expandable? or is this already set by menuAnchor?

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: > some platform may still treat this as textfield if both istextField and isButton are true.

I see. I added `textField` and assigned it the opposite value of `button`.

> Can this be focusable but not enabled?

When it's disabled, the text field is not focusable.

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: Done.

packages/flutter/lib/src/material/dropdown_menu.dart
chunhtai: This setting textfield to false causes child textfield to create a separate node as the child of this semantics node. I assume this is ok, just want to point it out.

I am not to sure about the logic here when it can be focus, this will become a button, otherwise this becomes a text field? Why would we turn it into a textfield if it is not focusable

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1150,48 +1158,73 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
+          button: !canRequestFocus(),
+          // Some platforms may still treat this as text field if both `textField`
+          // and `button` are true. So manually set `textField` to false when
+          // `button` is true.
+          textField: canRequestFocus(),
+          // This is set specificly for iOS because iOS does not have any native
+          // APIs to show whether the menu is expanded or collapsed.
+          hint: Theme.of(context).platform == TargetPlatform.iOS
+              ? _controller.isOpen
+                    ? 'Expanded'
```
chunhtai: use the string from Widget localization

QuncCccccc: Done. I used the strings from MaterialLocalizations.

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: > This setting textfield to false causes child textfield to create a separate node as the child of this semantics node. I assume this is ok, just want to point it out.

I actually wasn't aware of that. Thanks for pointing this out! I noticed Web will still treat it as text field when both are true. Other platforms seems fine. Also I noticed an error shows up on web when we run the app (on master), so instead of setting `textField` to false, I exclude the semantics on web.

Image

> when it can be focus, this will become a button, otherwise this becomes a text field? 

No I think it should be the opposite. When it can be focused, it is text field, when it's not, it's a button. Does this make sense?

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: I responded the first comment [here](https://github.com/flutter/flutter/pull/176428#discussion_r2415341128), just in case it's missed:)!

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1150,48 +1160,74 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
...
-            suffixIcon: widget.showTrailingIcon ? trailingButton : null,
-          ).applyDefaults(effectiveInputDecorationTheme),
-          restorationId: widget.restorationId,
+          onCollapse: !_controller.isOpen
+              ? null
+              : () {
+                  _controller.close();
+                },
+          child: ExcludeSemantics(
+            excluding: isButton && kIsWeb,
```
chunhtai, line 1189: Can you add a bit comment on why additional check on kIsWeb?

chunhtai (approved):
LGTM

packages/flutter/lib/src/material/dropdown_menu.dart
QuncCccccc: Done:)

PR #176432, opened 2025-10-02
[web] Fix offscreen canvas bitmap size mismatch

mdebbar:


Piinks:

PR #176475, opened 2025-10-03
Align flutter dependencies with ones coming from dart.

aam:
Manual workaround for https://github.com/flutter/flutter/issues/176472


gaaclarke (approved):
Don't we want to add running the script, otherwise this will get out of date quick.

aam:
> Don't we want to add running the script, otherwise this will get out of date quick.

Of course we do want to run the create_updated_flutter_deps script automatically. This is just a quick catch-up as I'm not sure how far we got out of sync.

PR #176481, opened 2025-10-03
Remove the C99 flag from the compiler configuration for C sources on Mac/iOS

jason-simmons:
This was imported from the original Chromium buildroot and appears to be obsolete.

### DIFF
--- a/engine/src/build/config/compiler/BUILD.gn
+++ b/engine/src/build/config/compiler/BUILD.gn
@@ -98,6 +98,8 @@ config("compiler") {
       "/guard:cf",  # Enable control flow guard security checks.
     ]
   } else {
+    cflags_c += [ "-std=c11" ]
+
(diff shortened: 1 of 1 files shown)


chinmaygarde (approved):
I'd be more comfortable with explicitly specifying c11 instead of letting the compiler pick. We allow for custom toolchains to plug into the system and want to be more resilient to varying opinions from the compilers on their defaults.

jason-simmons:
Added a flag to use C11 mode on all platforms

jason-simmons:
I will merge this into the `vulkan-deps` roll (https://github.com/flutter/flutter/pull/176322)

This change was needed because the `vulkan-loader` in https://github.com/flutter/flutter/pull/176322 was not building on macOS with `-std=c99`.

However, the currently used `vulkan-loader` will not build on Linux with `-std=c11`.

PR #176490, opened 2025-10-03
Handle FlutterEngine registration when embedded in Multi-Scene apps

vashworth:
When Multi-Scene is enabled, Flutter cannot automatically associate the engine with a scene during the scene connection phase in most cases. However, the `scene:willConnectToSession:options:` event is critical for some Flutter plugins that rely on scene connection options (the scene-equivalent of application launch options). 

To workaround this, this PR adds two things:
* Adds a way to manually register a `FlutterEngine` with a `FlutterPluginSceneLifeCycleDelegate`. Manually registered engines must also be manually unregistered.
  * This use case is for add to app
* Automatically registers a `FlutterEngine` if the `rootViewController` of the scene is a `FlutterViewController`
  * This use case is for if a Flutter app (not add-to-app) were to enable multi-scene since the `rootViewController` of the storyboard is a `FlutterViewController`
  
This PR also makes a fix for when a `FlutterView` is removed from the scene:
* If a `FlutterView` is removed from a scene, keep the `FlutterEngine` associated with the scene until it is assigned to a new scene

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,48 @@ API_AVAILABLE(ios(13.0))
...
+/**
+ * A protocol for manually registering a `FlutterEngine` to receive scene life cycle events.
+ */
+@protocol FlutterSceneLifeCycleEngineRegistration
+/**
+ * Registers a `FlutterEngine` to receive scene life cycle events.
+ *
+ * This method is required **only** when both of the following conditions are true:
+ * 1. Multiple Scenes (UIApplicationSupportsMultipleScenes) is enabled.
+ * 2. The `rootViewController` of the scene is **not** a `FlutterViewController`.
```
LongCatIsLooong: It appears that a scene can have multiple windows thus multiple root view controllers: https://developer.apple.com/documentation/uikit/uiwindowscene/windows, 

so "the root view controller of a scene" isn't clear which view controller if multiple are connected. (also maybe we shouldn't be using https://developer.apple.com/documentation/uikit/uiwindowscenedelegate/window 
since it's only referring to the main window).

LongCatIsLooong: nit: when the first condition is true, decision whether an engine needs to managed manually may have to be made on an engine-by-engine basis. I think the instruction would easier for the reader to follow if it's put from the engine's perspective? Like an engine would need to be manually managed if it's not associated with the rootViewController of a UIWindow.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,48 @@ API_AVAILABLE(ios(13.0))
...
+ * @param engine The `FlutterEngine` to register for scene life cycle events.
+ * @return `NO` if already manually registered.
+ */
+- (BOOL)registerSceneLifeCycleWithFlutterEngine:(FlutterEngine*)engine;
+
+/**
+ * Use this method to unregister a `FlutterEngine` from the scene's life cycle events.
+ *
+ * If a `FlutterEngine`'s view changes scenes,
+ *
```
LongCatIsLooong: nit: extraneous empty line

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,48 @@ API_AVAILABLE(ios(13.0))
...
+ * Only manually registered engines can be unregistered. See
+ * `registerSceneLifeCycleWithFlutterEngine`.
+ *
+ * @param engine The `FlutterEngine` to unregister for scene life cycle events.
```
LongCatIsLooong, line 133: If I'm understanding correctly, the change in `willMovetoWindow:` made it so that a Flutter managed engine never truly unregisters (it only moves to one scene from another). So why do we still need this method? The `registerSceneLifeCycleWithFlutterEngine` method seems sufficient for the moving case?

LongCatIsLooong: Also this method will open up the possibility that flutter tries to manage an engine after the developer manually unregisters it. Do we want to enable that?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,48 @@ API_AVAILABLE(ios(13.0))
...
+ * @return `NO` if the engine was not found among the manually registered engines and could not be
+ * deregistered.
```
LongCatIsLooong: uber nit: unregister? since almost everywhere else uses that?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,48 @@ API_AVAILABLE(ios(13.0))
...
+ * In this specific configuration, Flutter cannot automatically associate the engine with a scene
+ * during the scene connection phase. Manually registering the engine ensures it receives the
+ * critical `scene:willConnectToSession:options:` event, which is necessary for some Flutter plugins
+ * that rely on scene connection options (the scene-equivalent of application launch options).
+ *
+ * For all other configurations, Flutter automatically handles the registration.
+ *
+ * Manually registered engines must also be manually deregistered if they
+ * switch scenes or when they are no longer needed. Use
+ * `unregisterSceneLifeCycleWithFlutterEngine:`.
```
LongCatIsLooong: nit: I think we should tell app developers to do this at their earliest convenience to avoid missing any lifecycle events.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -115,7 +167,12 @@ - (FlutterPluginAppLifeCycleDelegate*)applicationLifeCycleDelegate {
 - (void)engine:(FlutterEngine*)engine receivedConnectNotificationFor:(UIScene*)scene {
   // Connection options may be nil if the notification was received before the
   // `scene:willConnectToSession:options:` event. In which case, we can wait for the actual event.
-  [self addFlutterEngine:engine];
+  BOOL added = [self addFlutterManagedEngine:engine];
```
LongCatIsLooong, line 172: uber nit: could you also add a comment that says this method is only reachable in a single scene app so there is no manually managed engines (or can we make that assumption? Should we assert in the manual registration method that the app has to support multi-scene to be able to use manual registration?)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterSceneLifeCycle.mm
```diff
@@ -15,57 +15,104 @@
...
   if (index != NSNotFound) {
-    [self.engines removePointerAtIndex:index];
+    [self.flutterManagedEngines removePointerAtIndex:index];
+    return YES;
   }
+  return NO;
 }
 
-- (void)updateEnginesInScene:(UIScene*)scene {
+- (void)updateFlutterManagedEnginesInScene:(UIScene*)scene {
```
LongCatIsLooong, line 108: I wonder why we only have to do this for flutter managed engines? Are we expecting the developer to manually unregister before the engine is deallocated? Or the "allEngines" implementation already does the updating?

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterSceneLifeCycle.h
```diff
@@ -96,6 +98,45 @@ API_AVAILABLE(ios(13.0))
...
+ * 2. The `UIWindowSceneDelegate` `window.rootViewController` is not a `FlutterViewController`
+ *    initialized with the target `FlutterEngine`.
+ *
+ * When multiple scenes is enabled (UIApplicationSupportsMultipleScenes), Flutter cannot
+ * automatically associate a `FlutterEngine` with a scene during the scene connection phase. In
+ * order for plugins to receive launch connection information, the `FlutterEngine` must be manually
+ * registered with either the `FlutterSceneDelegate` or `FlutterPluginSceneLifeCycleDelegate` during
+ * `scene:willConnectToSession:options:`.
+ *
+ * In all other cases, or once the `FlutterViewController.view` associated with the `FlutterEngine`
```
LongCatIsLooong, line 119: nit: the `or` case seems a bit misleading as it sounded like Flutter automatically handle registration once `view` is added to the hierarchy even for manually managed engines?

LongCatIsLooong (approved):

Issue #130976, opened 2023-07-20
AccessibilityFeatures.disableAnimations flag is not set on app startup and animations are not disabled in app running on Android 12+ device.


sanderpensa:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)
- [X] I have read the [guide to filing a bug](https://flutter.dev/docs/resources/bug-reports)

### Steps to reproduce

1. On Android 12+ device disable animations from Settings > Accessibility > Remove animations.
2. Run the app **flutter run --release** (does not matter if in release, profile or debug mode, we have this bug in live app also).
3. Observe AccessibilityFeatures.disableAnimations flag is not set, animations are not removed.
4. Put the app to backgound (do not close it), and change the "Remove animations" setting off and on again from Settings > Accessibility > Remove animations.
5. Open app from recents, observe AccessibilityFeatures.disableAnimations flag is updated and set after few seconds and animations are disabled.

### Expected results

App animations are disabled on Android 12+. 

### Actual results

Animations are not disabled.

### Code sample

<details><summary>Code sample</summary>

```dart
import 'dart:ui';

import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Accessibility disableAnimations example',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const SlideAnimationDemo(),
    );
  }
}

class SlideAnimationDemo extends StatefulWidget {
  const SlideAnimationDemo({super.key});

  @override
  State<SlideAnimationDemo> createState() => _SlideAnimationDemoState();
}

class _SlideAnimationDemoState extends State<SlideAnimationDemo>
    with SingleTickerProviderStateMixin, WidgetsBindingObserver {
  late AnimationController _animationController;
  late Animation<Offset> _slideAnimation;
  String features = '';

  @override
  void initState() {
    super.initState();

    WidgetsBinding.instance.addObserver(this);

    _animationController = AnimationController(
      vsync: this,
      duration: const Duration(milliseconds: 1000),
    );

    _slideAnimation = Tween<Offset>(
      begin: const Offset(-1.0, 0.0),
      end: Offset.zero,
    ).animate(CurvedAnimation(
      parent: _animationController,
      curve: Curves.easeInOut,
    ));

[...]

PR #176492, opened 2025-10-03
Image GIF pausing

### LINKED ISSUE #175516: [Google3 Bug]: CachedNetworkImage, ImageStream, & ImageProvider provide no way to pause GIF animations
### Help us understand the severity of this issue

- [ ] causing severe production issues e.g. malfunctions or data loss
- [ ] blocking next binary release
- [x] blocking a client feature launch within a quarter
- [ ] nice-to-have but does not block a launch within the next quarter

### Steps to reproduce

1. Load a GIF asset into `ImageProvider`/`ImageStream` via `CachedNetworkImage`.

### Expected results

Either one of the following:

1. The `ImageProvider`/`ImageStream` library provides a way to disable any animation of the provided image.
2. `MediaQuery.of(context).disableAnimations` is respected by the `ImageProvider`/`ImageStream` library by default.

### Actual results

1.  Neither `ImageProvider`/`ImageStream` provide a way to disable animation of an image
2. `MediaQuery.of(context).disableAnimations` is not respected by the `ImageProvider`/`ImageStream` library.

### Code sample

yaml:
```xml
dependencies:
  flutter:
    sdk: flutter
  cached_network_image: ^3.3.1 # Use the latest version
```

Dart File:
```dart
import 'package:cached_network_image/cached_network_image.dart';

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'GIF Loading with CachedNetworkImage',
      home: Scaffold(
        appBar: AppBar(title: const Text('Cached GIF')),
        body: Center(
          child: CachedNetworkImage(
            imageUrl:
                'https://media1.tenor.com/m/Ps_NnXFTyRYAAAAC/send-link-send.gif',
            placeholder: (context, url) => const CircularProgressIndicator(),
            errorWidget: (context, url, error) => const Icon(Icons.error),
          ),
        ),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[Paste your output here]
```

</details>

justinmc:
Flutter needs a strong story around pausing animated GIFs because it affects accessibility via [MediaQueryData.disableAnimations](https://main-api.flutter.dev/flutter/widgets/MediaQueryData/disableAnimations.html). See https://github.com/flutter/flutter/issues/175516.

### Today

Currently, it's possible to pause a gif using TickerProvider. However, if paused on initial load, the first frame will never display, and I believe that's a bug.

```dart
TickerMode(
  // Pauses the gif, but if set to false on first load, the first frame never loads.
  enabled: false,
  child: Image(image: myAnimatedGif),
),
```

Currently `disableAnimations` has no effect:

```dart
MediaQuery(
  // Currently has no effect on the Image, until this PR.
  data: MediaQueryData(disableAnimations: true),
  child: Image(image: myAnimatedGif),
),
```

### With this PR

  * MediaQueryData.disableAnimations can be used to pause a gif in the same way that TickerMode can be used today.
  * If either MediaQuery.disableAnimations or TickerMode cause a gif to be paused on first load, it will show its first frame before pausing. This is potentially a breaking change.
  * The docs clearly explain this behavior and how to pause a gif.

Fixes https://github.com/flutter/flutter/issues/175516
Partial fix for https://github.com/flutter/flutter/issues/130976
Fixes Google b/419605327


justinmc:
@evantea Let me know if this won't solve your problem!

packages/flutter/lib/src/widgets/image.dart
```diff
@@ -1100,6 +1119,15 @@ class _ImageState extends State<Image> with WidgetsBindingObserver {
   StackTrace? _lastStack;
   ImageStreamCompleterHandle? _completerHandle;
 
+  /// True when animations are disabled and the image should not update, such as
+  /// when [TickerMode] is disabled or [MediaQueryData.disableAnimations] is
+  /// true.
+  bool _isPaused = false;
+
+  /// False when the class first is instantiated and true forever after the
+  /// first frame of the image is received.
```
chunhtai: What if image source changes?

packages/flutter/lib/src/widgets/image.dart
```diff
@@ -1211,6 +1241,11 @@ class _ImageState extends State<Image> with WidgetsBindingObserver {
   }
 
   void _handleImageFrame(ImageInfo imageInfo, bool synchronousCall) {
+    if (_hasReceivedFirstFrame && _isPaused) {
```
chunhtai: do we need to worry about first frame throws error?

packages/flutter/lib/src/widgets/image.dart
```diff
@@ -1123,7 +1151,9 @@ class _ImageState extends State<Image> with WidgetsBindingObserver {
     _updateInvertColors();
     _resolveImage();
 
-    if (TickerMode.of(context)) {
+    _isPaused = !TickerMode.of(context) || (MediaQuery.maybeDisableAnimationsOf(context) ?? false);
```
chunhtai, line 1151: not something needs to be addressed in this PR, but I am surprised `maybeDisableAnimationsOf` is not hooked up to the TickerMode directly

justinmc: I've been thinking about them as two different things: MediaQueryData.disableAnimations is related to the system accessibility setting while TickerMode is about definitively disabling all animations. @evantea was unable to use TickerMode because it also disabled the fade animation that displayed the image, but that is not a problem for accessibility (see https://github.com/flutter/flutter/issues/175516#issuecomment-3363396318).

packages/flutter/lib/src/widgets/image.dart
justinmc: I think you're right, if the image source changes we  want to load the first frame of the new image. I'll do that.

packages/flutter/lib/src/widgets/image.dart
chunhtai: > MediaQueryData.disableAnimations is related to the system accessibility setting

Yes, but what if a user toggle this setting and see all the animation is still animating. That feels bad. Right now we have to rely on developer or per widget to handle them seems not too scalable.

In the same time I wonder for the fade case, whether turning on disable animation should only turn off certain animation. Or maybe ticker may be used outside the just visual animation. Then in that case I agree hook up directly to ticker mode may not be a good idea. Maybe what we want is to hardcode all animation duration to zero if disableAnimations is on.

evantea:
@justinmc copied this commit into my local workspace, using only the "Remove Animations" system setting images, GIFs paused automatically (though only after rebuilding the view). Thanks for putting this together!

packages/flutter/lib/src/widgets/image.dart
justinmc: I guess we need to think hard about exactly which animations should be disabled and which (if any) shouldn't when disableAnimations is true. It is reasonable that maybe some Fade animations *should* be disabled.

The [docs for disableAnimations](https://main-api.flutter.dev/flutter/widgets/MediaQueryData/disableAnimations.html) do give me the impression that some animations will not be fully disabled:

> Whether the platform is requesting that animations be disabled or reduced as much as possible.
>
> See also:
>
> [dart:ui.PlatformDispatcher.accessibilityFeatures](https://main-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/accessibilityFeatures.html), where the setting originates.

packages/flutter/lib/src/widgets/image.dart
justinmc: I've rewritten this part, let me know if it's still a concern.

packages/flutter/lib/src/widgets/image.dart
justinmc: This works now but it was a journey. I put everything related to source swapping into this into one commit: https://github.com/flutter/flutter/pull/176492/commits/3665a79e64087e9a62f20816042ba30e1a0455eb. Summarizing:

 * When I first started looking at this, the behavior was that when I paused the image and then changed the image source, the Image would disappear (the first frame was never loaded).
 * I fixed that and it was pretty straightforward, but I noticed another problem: swapping the original source back in would advance it by one frame, not show the originally paused frame. If I switched back and forth between sources, they would each advance by one frame every time. Depending on the gif, this bug could be pretty noticeable.
 * To solve this, I implemented a cache of the most recent frame from each unique source.

Let me know if my cache solution is overkill and we should just accept the frame advancing bug. I'm happy with how the caching solution turned out but in retrospect I probably should have saved it for a separate PR if I had known it would take a day plus of hacking.

justinmc:
@chunhtai This is ready for re-review, but it seems to be failing a skwasm test in a way that I can't reproduce locally. See my comment at https://github.com/flutter/flutter/pull/176492#discussion_r2418059865. Since this is a P1, maybe I should split out the image source swapping fix into a follow-up PR. Interested in what you think.

packages/flutter/lib/src/widgets/image.dart
```diff
@@ -1232,11 +1282,21 @@ class _ImageState extends State<Image> with WidgetsBindingObserver {
...
     final ImageInfo? oldImageInfo = _imageInfo;
-    SchedulerBinding.instance.addPostFrameCallback(
-      (_) => oldImageInfo?.dispose(),
-      debugLabel: 'Image.disposeOldInfo',
-    );
+    final ImageProvider? oldImageProvider = _imageProvider;
+    if (oldImageInfo != null) {
+      SchedulerBinding.instance.addPostFrameCallback((Duration duration) {
+        // If oldImageInfo is still in the cache, then it will be disposed when
+        // the widget is disposed. Otherwise it has been evicted, so dispose it
```
chunhtai: any reason we don't dispose them now? also the current structure seems to make the _imageInfosCache to grow infinitely when Image source keep changing during the life time of this widget.

chunhtai (commented):
Just one comment on the cache management, the rest LGTM

chunhtai:
Hi @justinmc does this pr fixes https://github.com/flutter/flutter/issues/130976? I saw the pr is marked so in the description, but not sure whether that is the case

justinmc:
@chunhtai I guess this only helps with Image, but that issue seems more broad. I will edit the description.

packages/flutter/lib/src/widgets/image.dart
justinmc: I refactored this to no longer need a cache.

As to why not dispose the image info immediately, I don't know! It seems to still work if I change it to dispose immediately, and all the tests even continue to pass.

I'll open a separate PR to try that out.

mdebbar (dismissed):
Is https://github.com/flutter/flutter/pull/176492/commits/ebb7ee964302c52931df379d4632097743bcc8c4 still needed or can it be reverted?

mdebbar (commented):
Thanks @justinmc :)

justinmc:
I have a local commit to revert https://github.com/flutter/flutter/pull/176492/commits/a28b26a1ac35f6eab2903f55aa026fec262b6921 that I will push as soon as my fix for the Google tests is merged.

justinmc:
Google tests have been fixed in Google internal cl/829551684. Now triggering a re-run to confirm that they pass with this PR applied.

packages/flutter/lib/src/widgets/image.dart
```diff
@@ -1124,10 +1148,12 @@ class _ImageState extends State<Image> with WidgetsBindingObserver {
     _updateInvertColors();
     _resolveImage();
 
-    if (TickerMode.of(context)) {
-      _listenToStream();
-    } else {
+    _isPaused = !TickerMode.of(context) || (MediaQuery.maybeDisableAnimationsOf(context) ?? false);
+
+    if (_isPaused && _frameNumber != null) {
```
chunhtai, line 1153: do you still need to keep track of number of frames? or can this be a boolean field?

justinmc: I was about to change it but I realized it is used to give the frame number to frameBuilder: https://main-api.flutter.dev/flutter/widgets/ImageFrameBuilder.html

chunhtai (approved):
LGTM

PR #143416, opened 2024-02-14
Add animation for menus

QuncCccccc:
Fixes: https://github.com/flutter/flutter/issues/143316 & https://github.com/flutter/flutter/issues/143781
This PR is to 
 * add expand/collapse animation to `MenuAnchor` and `DropdownMenu`. On desktop, menus don't show any animations; On mobile, menu size shows animation when it is expanded and collapsed.
 * add `sizeAnimationStyle` to `MenuStyle` so we can customize the menu animation.

Default animation:

https://github.com/flutter/flutter/assets/36861262/e86d9fa1-96a4-47ef-a08a-791b4c1658ab

Custom animation:

https://github.com/flutter/flutter/assets/36861262/1a92d4b4-eadc-43c3-9293-296170c9d1fa

Default animation in slow mode:

https://github.com/flutter/flutter/assets/36861262/fc0ab2a4-6e63-4929-aea7-7429de91854b

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #174695, opened 2025-08-29
[cupertino.dart] Implement CupertinoMenuAnchor and CupertinoMenuItem using RawMenuAnchor

davidhicks980:
This PR implements `CupertinoMenuAnchor`, `CupertinoMenuItem`, and `CupertinoMenuDivider` using `RawMenuAnchor`. 

Resolves https://github.com/flutter/flutter/issues/60298, https://github.com/notDmDrl/pull_down_button/issues/26, https://github.com/flutter/flutter/pull/137936.

https://github.com/user-attachments/assets/551f2169-cac4-4078-890b-6728ab9c2ae5

Dartpad: https://dartpad.dev/8c6bba779b6a00e95582b61b132292bc

@dkwingsmt 

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[...]

PR #176494, opened 2025-10-03
[material/menu_anchor.dart] Add animations to MenuAnchor.

### LINKED ISSUE #135025: Add open and close animations for MenuAnchor

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)

### Use case

So I noticed this after replacing PopupMenuButtons and DropdownButtons in our app with MenuAnchors and DropdownMenus. 

According to the [material design guidelines](https://m3.material.io/components/menus/guidelines#:~:text=Menus%20use%20a%20fade%20transition%20pattern%20to%20enter%20and%20exit%20the%20screen.%20The%20animation%20creates%20a%20relationship%20between%20the%20menu%20and%20the%20interactive%20element%20that%20generates%20the%20menu.), there should be a fade transition when menu enters and exits on screen, but currently MenuAnchor just opens and close sthe menu without any animations.

### Proposal

It's not a big deal, and would actually be preferable if on desktop and web, but on Android it kinds of look weird for the menus to just suddenly appears.

davidhicks980:
* Adds MD3 animations to MenuAnchor. 
* Adds a "hoverOpenDelay" parameter on `SubmenuButton`
* Moves the layout algorithm from SingleChildLayoutDelegate to a custom render object. <s>This was done because I needed to access the dry layout of the menu panel for the best effect. I'll demonstrate below.</s> Turns out using dry layout significantly limits what children you can use with a widget. As a result, I'm calculating the final height using the inverse height factor.

Fixes https://github.com/flutter/flutter/issues/135025

I don't have access to the internal Google documentation, so I did my best to deduce the spec from https://github.com/material-components/material-web/blob/main/menu/internal/menu.ts.

This change is currently opt-in via an `animated` parameter on `MenuAnchor` and `SubmenuButton`. As a result, this is not a breaking change. Because the PR is already quite large, I chose to not add animation customization to this PR. I also didn't change any theming files or `DropdownMenu`. 

The only other API addition is a `hoverOpenDelay` parameter on `SubmenuButton`. This parameter delays the start of a submenu opening by the specified duration. It is independent of the `animated` parameter. 

When `animated == false`, the menu runs its forward and reverse animations with a duration of 0. I originally disposed of all animation assets when `animated` was false, but found that the code/testing complexity to be unwieldy.

<s>Blocked by https://github.com/flutter/flutter/pull/176503</s>

## Examples

* Basic example adapted from material-web

https://github.com/user-attachments/assets/e6d02fab-e11c-4dce-ab82-2fb39c1111e0

* Example showing scrollbar 

https://github.com/user-attachments/assets/847054ba-7218-4e73-9835-019bfa7b5521

* MenuBar usage

https://github.com/user-attachments/assets/8231b1cd-c40c-4f18-af87-1b77b43ecb7b

Thanks to @QuncCccccc  and @dkwingsmt for [their work on this issue.](https://github.com/flutter/flutter/pull/143416) Sorry for throwing a wrench into the original PR by introducing `RawMenuAnchor`...

## Pre-launch Checklist

- [x] I signed the [CLA].


guidezpl:
Looks good and seems consistent with https://m3.material.io/components/menus/guidelines#691d98ba-0375-4681-8c47-a5c3e7b58798. Only suggestion I would have is to not have a scrollbar appear temporarily during the animation, if there's not going to be one in the completed animated state.

davidhicks980:
@guidezpl I was thinking the same thing. Fix is pushed.

davidhicks980:
Looking at the expressive specs (https://m3.material.io/styles/motion/overview/specs), it may be worthwhile to use springs rather than easing curves, since springs can be interrupted without disrupting the animation's appearance. Let me know if I should switch over! I'm already using springs for the CupertinoMenuAnchor, so it'd be an easy change.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -84,6 +87,53 @@ const double _kMenuViewPadding = 8;
...
+
+// The default curve used to animate the height of the menu panel when opening.
+const Curve _kMenuPanelHeightForwardCurve = Cubic(.3, 0, 0, 1);
+
+// The default curve used to animate the height of the menu panel when closing.
+const Curve _kMenuPanelHeightReverseCurve = _TweenCurve(
+  0.35,
+  1,
+  curve: FlippedCurve(Easing.emphasizedAccelerate),
+);
```
dkwingsmt: I think these two (`_kMenuPanelHeightReverseCurve` and `_kMenuPanelHeightForwardCurve`) are not right, because they use the full duration from 0.0 to 1.0, compared to how `_kMenuPanelOpacityForwardCurve` and `_kMenuPanelOpacityReverseCurve` uses the time unit of fraction of 500ms.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -99,17 +149,44 @@ typedef MenuAnchorChildBuilder =
...
+  final double end;
+  final Curve curve;
+
+  @override
+  double transformInternal(double t) {
+    assert(begin >= 0.0);
+    assert(begin <= 1.0);
+    assert(end >= 0.0);
+    assert(end <= 1.0);
+    assert(end >= begin);
```
dkwingsmt: Move the asserts to the constructor?

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -99,17 +149,44 @@ typedef MenuAnchorChildBuilder =
...
 
   @override
   bool updateShouldNotify(_MenuAnchorScope oldWidget) {
     assert(oldWidget.state == state, 'The state of a MenuAnchor should not change.');
-    return false;
+    return oldWidget.animationStatus != animationStatus;
   }
 }
 
+class _TweenCurve extends Curve {
```
dkwingsmt, line 176: Add a brief doc. (What's its difference from `Interval`?)

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -1551,6 +1803,23 @@ class SubmenuButton extends StatefulWidget {
...
+  ///
+  /// Because [MenuBar] children can only be opened by hover if a sibling
+  /// [SubmenuButton] is already open, providing a [hoverOpenDelay] to direct
+  /// children of a [MenuBar] will throw an assertion error. This is to avoid
+  /// the case where the [hoverOpenDelay] for a [SubmenuButton] is longer than
+  /// the duration of a sibling's closing animation, which leads to that menu
+  /// never opening.
+  ///
+  /// Defaults to [Duration.zero].
+  final Duration hoverOpenDelay;
```
dkwingsmt, line 1886: Where is `hoverOpenDelay` used? I can't find it used in this PR and I don't think it's common to have hover-open-delay in common apps, is it?

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -1551,6 +1803,23 @@ class SubmenuButton extends StatefulWidget {
...
+
+  /// Whether the menu should open and close with an animation.
```
dkwingsmt, line 1888:
```suggestion
  /// Whether this widget should open or close a submenu with an animation.
```

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -1928,15 +2221,71 @@ class _SubmenuButtonState extends State<SubmenuButton> {
...
-      if (!_menuController.isOpen && _isOpenOnFocusEnabled) {
-        _menuController.open();
-      }
-    } else {
-      if (!_anchorState!._menuScopeNode.hasFocus && _menuController.isOpen) {
+    _handleTogglingInteraction();
+  }
+
+  // Handles interactions that could open or close the menu.
+  void _handleTogglingInteraction() {
```
dkwingsmt: I think it's easier to understand as `_handleFocusChange` instead of `_handleTogglingInteraction`, because it made me wonder why any event would trigger either opening or closing depending on some kind of condition, while in fact it's a function that has to deal with "onGainFocus" and "onLoseFocus" together due to API design and split at the first line of the body.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -2014,8 +2363,8 @@ class _SubmenuDirectionalFocusAction extends DirectionalFocusAction {
               // the menu.
               return;
             }
-            _parent?._focusButton();
             _parent?._menuController.close();
+            _parent?._focusButton();
```
dkwingsmt, line 2436: Any particular reason for this? Will this cause the focus to be briefly assigned to some default one due to `menuController.close()` unassigning focus from itself?

dkwingsmt (commented):
Generally looking good.

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Yeah, I agree. I added _handleTogglingInteraction because hovering can now trigger _handleFocusChange, even if the text button is already focused. 
```dart
 if (_buttonFocusNode.hasPrimaryFocus) {
    _handleFocusChange();
  } else {
    _buttonFocusNode.requestFocus();
  }
```
But, I can just call handleFocusChange()

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: I think I just accidentally left it in from the focus PR I'm working on. Whoops! Thanks for catching.

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: It's used in _SubmenuButtonState._handleFocusChange and it's incredibly common. The macOS and Chromium menus both have a hoverOpenDelay. 

```dart
  void _handleFocusChange() {
    _clearHoverOpenTimer();
    if (!_buttonFocusNode.hasPrimaryFocus) {
      if (!_anchorState!._menuScopeNode.hasFocus && _animationStatus.isForwardOrCompleted) {
        _menuController.close();
      }
      return;
    }

    if (!_isOpenOnFocusEnabled) {
      return;
    }

    if (_menuController.isOpen) {
      if (_animationStatus != AnimationStatus.reverse) {
        // If the menu isn't closing, there's no reason to reopen it.
        return;
      }

      if (_isHovered) {
        // If the button is hovered, avoid reopening the menu that the user deliberately
        // closed.
        return;
      }

      if (_parent?._orientation == Axis.horizontal) {
        // Top-level (horizontal) buttons in a menubar will stay open if we
        // don't prevent it.
        return;
      }
    }

    if (widget.hoverOpenDelay == Duration.zero) {
      _menuController.open();
      return;
    }

    _hoverOpenTimer = Timer(widget.hoverOpenDelay, () {
      // The menu controller could change, so we can't use a tearoff.
      _menuController.open();
    });
  }
```

https://github.com/user-attachments/assets/4434f9ec-2abb-49cf-8521-b80c6342def2

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Good catch! I think this needs to be done with the Interval() constructor as well. Done!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Elaborate a bit more on that...  I think the height curve is supposed to be applied across the whole duration (based off of: https://github.com/material-components/material-web/blob/fb6798bc02409eb79b0f8f988825efe570f803bc/menu/internal/menu.ts#L712 and https://github.com/material-components/material-web/blob/fb6798bc02409eb79b0f8f988825efe570f803bc/menu/internal/menu.ts#L828), but the opacity curve is only supposed to be applied over a fraction of the total duration.

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Done!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Fixed!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Done!

packages/flutter/lib/src/material/menu_anchor.dart
dkwingsmt: Oh, my bad. I somehow thought they only used part of the duration. Sorry!

packages/flutter/lib/src/material/menu_anchor.dart
dkwingsmt: I suggest either of the following way:
* Split `_handleTogglingInteraction` into `_openSubmenu` and `_closeSubmenu`, and `handlePointerHover` calls the opening one, and `_handleFocusChange` calls either based on status, which is clearer as well.
* Rename `_handleTogglingInteraction` to something like `_openCloseMenuBasedOnFocus`.

davidhicks980: I kept the closing behavior in the _handleFocusChange function and split off the opening behavior into a function called _maybeOpenMenuOnHoverOrFocus

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -84,6 +87,53 @@ const double _kMenuViewPadding = 8;
...
+// https://github.com/material-components/material-web/blob/e15f47e1a27d879d2cf46a6d6820cc81603621d3/menu/internal/menu.ts
+
+// The default opening animation duration for menus.
+const Duration _kMenuOpeningDuration = Duration(milliseconds: 500);
+
+// The default closing animation duration for menus.
+const Duration _kMenuClosingDuration = Duration(milliseconds: 150);
+
+// The default curve used to animate the height of the menu panel when opening.
+const Curve _kMenuPanelHeightForwardCurve = Cubic(.3, 0, 0, 1);
```
QuncCccccc, line 107: I checked the specs, the enter animation for menu should use `md.sys.motion.easing.emphasized` which should be this [curve](https://api.flutter.dev/flutter/animation/Curves/easeInOutCubicEmphasized-constant.html) ([reference](https://m3.material.io/styles/motion/easing-and-duration/tokens-specs#ed7ab8eb-9b9a-40a3-806f-a4485558f3df))

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -84,6 +87,53 @@ const double _kMenuViewPadding = 8;
...
+// duration.
+//
+// 50 ms of the 150 ms animation.
+const double _kMenuItemRelativeFadeOutDuration = 1 / 3;
+
+// The default delay between the start of each menu item's fade-in as a
+// fraction of the opening duration.
+//
+// 50 ms of the 150 ms animation
+const double _kMenuItemRelativeFadeOutDelay = 1 / 3;
```
QuncCccccc, line 133: Seems there is some inconsistency between the variable name and the doc. The doc mentioned this is for fade-in delay.

QuncCccccc: I'm not sure whether we can share the internal specs here but I checked the specs, everything here is pretty accurate (except this one and `_kMenuPanelHeightForwardCurve`)!

packages/flutter/lib/src/material/menu_anchor.dart
QuncCccccc: I have the same question and noticed it's answered here! Nice Observation!

QuncCccccc (commented):
Overall looks good to me! Thanks for adding this feature:)

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -174,6 +261,8 @@ class MenuAnchor extends StatefulWidget {
     this.onClose,
     this.crossAxisUnconstrained = true,
     this.useRootOverlay = false,
+    this.animated = false,
+    this.onAnimationStatusChanged,
```
QuncCccccc, line 264: Also curious the use case for this callback.

QuncCccccc:
Can we also update the demo in the PR description to show the correct Scrollbar behavior:)?

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Oh good catch! Docs should be fixed. I also removed some extraneous docs (e.g. spec being obtained from typescript version).

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Good question. We actually had this discussion during the RawMenuAnchor animation implementation. If you don't have access to the animation status, you can't open a menu while that menu is in the process of closing (i.e. during AnimationStatus.reverse)

Since MenuController.isOpen represents whether the overlay is visible, MenuController.isOpen is true while the menu is opening, open, and closing (AnimationStatus.forward, .completed, and .reverse, respectively). That means that, if you press the anchor button while the menu is closing, MenuController.isOpen will be true, so the button will call MenuController.close() despite the menu already closing. If you want to make a toggleable menu, you need to have access to the AnimationStatus so that you can test if AnimationStatus.reverse is true.

I didn't use an onAnimationStatusChanged listener on the example because I didn't want users to think that migration to an animated menu requires extra steps. But, do you think it'd be helpful to add it to the example? I did so on the CupertinoMenuAnchor implementation to allow toggleability: https://github.com/flutter/flutter/pull/174695/files#diff-330ffad0cc9875f50a0466e01e8a25e1612e30d5331ff40a6834c30ae7da7b3eR98-R104.

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Hmm, so I pushed the change, but the animation now looks noticeably worse:

https://github.com/user-attachments/assets/5369bc58-431a-4dd5-b99c-77111b01ea61

https://github.com/user-attachments/assets/6b295e67-ea5c-4ee1-bc51-044a259eeee3

It's more noticeable on screen with a higher refresh rate, but the opacity animation of the items doesn't line up with the motion of the menu: 

https://github.com/user-attachments/assets/5a780466-9a69-44eb-9165-c7be03671c11

https://github.com/user-attachments/assets/87d36198-3275-4930-a07e-a2358cce55fc

I imagine this may be because the Material spec was made with the Android implementation in mind, which doesn't use a cubic bezier for md.sys.motion.easing.emphasized. 

All of that to say, how would the material team feel if we went with the curve from the web implementation?  Or, I could try implementing the path-curve Android is using (FYI - I have no idea how much of a rabbit hole this could be).

That said, feel free to override my opinion here!

QuncCccccc: I see. I think if the original change fits more, then I agree we should keep it.

packages/flutter/lib/src/material/menu_anchor.dart
QuncCccccc: I see! Thanks a lot for your explanation! I think it would be great to create a separate PR to add the example or add a code sample in the API docs so developers can better understand how this property works!

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -1551,6 +1794,23 @@ class SubmenuButton extends StatefulWidget {
...
+  /// the duration of a sibling's closing animation, which leads to that menu
+  /// never opening.
+  ///
+  /// Defaults to [Duration.zero].
+  final Duration hoverOpenDelay;
+
+  /// Whether the menu should open and close with an animation.
+  ///
+  /// Defaults to false.
+  final bool animated;
```
QuncCccccc, line 1891: Do we need to add `onAnimationStatusChanged` to `SubmenuButton` since it can also be treated like a "anchor"? But it's fine to add it when we need it in the future😄

QuncCccccc (approved):
LGTM. Thanks a lot for implementing this feature!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Done and added tests

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: I added a snippet on the onAnimationStatusChanged prop and demonstrated how to use animation status on menu_anchor.0.dart. Good catch!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Reverted! Also, added a comment explaining that we chose to stick with the web curve because it was more consistent with the fade in effect of the menu items.

packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -293,6 +380,55 @@ class MenuAnchor extends StatefulWidget {
...
+  ///
+  /// If [animated] is false, this callback will only be invoked with
+  /// [AnimationStatus.completed] when the menu is opened, and
+  /// [AnimationStatus.dismissed] when the menu is closed.
+  ///
+  /// Because the [MenuController.isOpen] property is true while the menu is
+  /// opening, opened, and closing, it cannot be used to determine whether the
+  /// menu is closing or opening. As such, this callback provides a way to
+  /// determine when the menu is opening or closing, and allows anchors to
+  /// toggle between the opening and closing states of the menu.
```
dkwingsmt: I find the first sentence taking a bit effort to understand due to the change of subject and the "it". I propose the following rewrite (with the help of AI).
```suggestion
  /// This callback provides a way to determine when the menu is opening or
  /// closing. This is necessary because the [MenuController.isOpen] property
  /// remains true throughout the opening, opened, and closing phases, and
  /// therefore cannot be used on its own to determine the current animation
  /// direction.
```

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -2078,6 +2079,136 @@ void main() {
       await tester.sendKeyEvent(LogicalKeyboardKey.arrowDown);
       expect(focusedMenu, equals('MenuItemButton(Text("Sub Menu 12"))'));
     });
+    testWidgets('hoverOpenDelay delays when a SubmenuButton opens', (WidgetTester tester) async {
```
dkwingsmt, line 2083:
```suggestion

    testWidgets('hoverOpenDelay delays when a SubmenuButton opens', (WidgetTester tester) async {
```

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -2078,6 +2079,136 @@ void main() {
...
+          isA<FlutterError>().having(
+            (FlutterError e) => e.message,
+            'message',
+            contains(
+              'A non-zero hoverOpenDelay was used in a top-level SubmenuButton situated in a MenuBar',
+            ),
+          ),
+        );
+      },
+      experimentalLeakTesting: LeakTesting.settings
```
dkwingsmt: I think the convention, according to other tests, is to place this property before the test body.

dkwingsmt (approved):
LGTM. Thank you so much for the fantastic work! (I've asked Qun for final review.)

packages/flutter/lib/src/material/menu_anchor.dart
QuncCccccc: nit: We can also mention that this example also shows how to use [onAnimationStatusChanged] to track animation status and toggle the menu.

QuncCccccc (commented):
LGTM. Thanks for your contribution:)!

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Done -- copied verbatim

packages/flutter/test/material/menu_anchor_test.dart
davidhicks980: Done

packages/flutter/test/material/menu_anchor_test.dart
davidhicks980: Done

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Done. Copied verbatim

PR #176495, opened 2025-10-03
Add shared Darwin implementation for plugins

### LINKED ISSUE #161019: Allow creating "darwin" (shared codebase for ios/macos) plugins directly in `flutter create`
### Use case

Right now when you create darwin plugin, you have to start with an ios or macos template. Then follow the [docs](https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations). There are clearly some differences between macos and ios templates. Which one is preferred to start with? Why not just add a darwin platform?

### Proposal

Simply make 'darwin' a valid platform option and create a darwin project directly from CLI.

```sh
flutter create --template=plugin --platforms="darwin" my_plugin
```

okorohelijah:
This PR introduces support for creating Flutter plugins with a single, shared implementation for iOS and macOS. This is enabled by a new darwin platform option in the flutter create command, which simplifies code maintenance and reduces duplication for plugin authors.

Fixes #161019 

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -984,6 +1009,9 @@ List<String> _getPlatformWarningList(List<String> requestedPlatforms) {
     if (requestedPlatforms.contains('macos') && !featureFlags.isMacOSEnabled) 'macos',
     if (requestedPlatforms.contains('windows') && !featureFlags.isWindowsEnabled) 'windows',
     if (requestedPlatforms.contains('linux') && !featureFlags.isLinuxEnabled) 'linux',
+    if (requestedPlatforms.contains('darwin') &&
+        !(featureFlags.isMacOSEnabled && !featureFlags.isIOSEnabled))
+      'darwin',
```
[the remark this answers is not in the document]
okorohelijah: done

packages/flutter_tools/test/commands.shard/permeable/create_test.dart
```diff
@@ -3422,6 +3422,168 @@ void main() {
...
+
+  // testUsingContext('does not return darwin when only iOS is enabled', () {
+  //   final CreateCommand command = CreateCommand();
+  //   final List<String> requested = <String>['darwin'];
+  //   final List<String> warnings = command.getPlatformWarningList(
+  //     requestedPlatforms: requested,
+  //     featureFlags: FakeFeatureFlags(isMacOSEnabled: false, isIOSEnabled: true),
+  //   );
+  //   expect(warnings, isNot(contains('darwin')));
+  // });
```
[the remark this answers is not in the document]
okorohelijah: done

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -1015,6 +1046,12 @@ For more details, see: https://flutter.dev/to/add-desktop-support
     globals.printStatus('''
 The web is currently not supported on your local environment.
 For more details, see: https://flutter.dev/to/add-web-support
+''');
+  }
+  if (darwin.isNotEmpty) {
+    globals.printStatus('''
+The darwin is currently not supported on your local environment.
```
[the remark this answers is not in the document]
okorohelijah: sounds good

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -424,6 +424,12 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
   /// This parent folder of `Runner.xcodeproj`.
   @override
   Directory get hostAppRoot {
+    if (parent.manifest.isModule || (parent.manifest.ios?.sharedDarwinSource ?? false)) {
+      final Directory darwinDir = parent.directory.childDirectory('darwin');
```
jmagman: Why was this change needed?

I don't think `hostAppRoot` is the right spot to change, if it is needed for some reason. That's used for apps (or plugin example apps), not for plugins.
I think instead you may want to move the `ios` /`darwin` logic into the  `_editableDirectory` getter, since that's used in more that would otherwise need editing.  For example `_flutterLibRoot` isn't going to point the correct place with this PR.

packages/flutter_tools/templates/plugin_darwin_cocoapods/darwin.tmpl/Classes/pluginClass.swift.tmpl
```diff
@@ -0,0 +1,36 @@
...
+      #elseif os(macOS)
+        result("macOS " + ProcessInfo.processInfo.operatingSystemVersionString)
+      #else
+        result(FlutterMethodNotImplemented)
+      #endif
+    default:
+      result(FlutterMethodNotImplemented)
+    }
+  }
+}
```
jmagman: Missing endline

packages/flutter_tools/templates/plugin_darwin_spm/darwin.tmpl/projectName.tmpl/Sources/projectName.tmpl/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jmagman, line 1: You shouldn't need to copy files that exist in the other applied templates, only the template that has different content (`pluginClass.swift.tmpl`)

packages/flutter_tools/templates/plugin_darwin_spm/darwin.tmpl/projectName.tmpl/Package.swift.tmpl
```diff
@@ -0,0 +1,33 @@
...
+                // https://developer.apple.com/documentation/bundleresources/privacy_manifest_files
+                // .process("PrivacyInfo.xcprivacy"),
+
+                // If you have other resources that need to be bundled with your plugin, refer to
+                // the following instructions to add them:
+                // https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package
+            ]
+        )
+    ]
+)
```
jmagman: Add endline

jmagman (commented):
Can you also add an integration test?

After this line:
https://github.com/flutter/flutter/blob/f94a9421865f753f7beba56ede6d39c25a0ff38b/dev/devicelab/bin/tasks/plugin_test_ios.dart#L19
```dart
PluginTest('ios', <String>['--platforms=darwin']).call,
```

And here:
https://github.com/flutter/flutter/blob/f94a9421865f753f7beba56ede6d39c25a0ff38b/dev/devicelab/bin/tasks/plugin_test_macos.dart#L15
```dart
 PluginTest('macos', <String>['--platforms=darwin']).call, 
```

jmagman:
Some test failures: 
```
06:03 +739 ~1 -1: test/commands.shard/permeable/create_test.dart: app does not include ios if disabled in config [E]
  Expected: not a file system entity that exists
    Actual: ErrorHandlingDirectory:<LocalDirectory: '/b/s/w/ir/x/t/flutter_tools.EBRPEX/flutter_tools_create_test.UMYZJU/flutter_project/ios'>
```
```
09:15 +874 ~2 -2: test/commands.shard/permeable/create_test.dart: app does not include desktop or web by default [E]
  Expected: not a file system entity that exists
    Actual: ErrorHandlingDirectory:<LocalDirectory: '/b/s/w/ir/x/t/flutter_tools.MRWEKF/flutter_tools_create_test.XRILCZ/flutter_project/macos'>
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8701985648264674337/+/u/run_test.dart_for_tool_tests_shard_and_subshard_commands/stdout

jmagman (commented):
This LGTM, I'd like @vashworth to also check it out though, particularly the template changes, and SwiftPM specifically.
I'm hoping there's an easy way to not have a separate darwin gitignore as well?

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -377,12 +379,23 @@ class CreateCommand extends FlutterCommand with CreateBase {
...
+      includeDarwin = false;
     } else {
-      includeIos = featureFlags.isIOSEnabled && platforms.contains('ios');
+      final bool darwinRequested = platforms.contains('darwin');
+      final bool darwinSupported = featureFlags.isIOSEnabled || featureFlags.isMacOSEnabled;
+
+      if (darwinRequested && darwinSupported) {
+        includeDarwin = true;
+        includeIos = featureFlags.isIOSEnabled;
+        includeMacos = featureFlags.isMacOSEnabled;
```
vashworth: What happens if they select darwin, but macOS is not enabled?

packages/flutter_tools/templates/plugin_darwin_spm/darwin.tmpl/projectName.tmpl/Sources/projectName.tmpl/pluginClass.swift.tmpl
```diff
@@ -0,0 +1,40 @@
...
+        result("macOS " + ProcessInfo.processInfo.operatingSystemVersionString)
+      #else
+        // A fallback for any other Apple platform that might be supported in the future.
+        result(FlutterMethodNotImplemented)
+      #endif
+    default:
+      result(FlutterMethodNotImplemented)
+    }
+  }
+}
```
vashworth: Add trailing end line

packages/flutter_tools/templates/plugin_darwin_cocoapods/darwin.tmpl/projectName.podspec.tmpl
```diff
@@ -0,0 +1,47 @@
...
+  # required reason APIs, update the PrivacyInfo.xcprivacy file to describe your
+  # plugin's privacy impact, and then uncomment this line. For more information,
+  # see https://developer.apple.com/documentation/bundleresources/privacy_manifest_files
+  {{#withSwiftPackageManager}}
+  # s.resource_bundles = {'{{projectName}}_privacy' => ['{{projectName}}/Sources/{{projectName}}/PrivacyInfo.xcprivacy']}
+  {{/withSwiftPackageManager}}
+  {{^withSwiftPackageManager}}
+  # s.resource_bundles = {'{{projectName}}_privacy' => ['Resources/PrivacyInfo.xcprivacy']}
+  {{/withSwiftPackageManager}}
+end
```
vashworth: Add trailing end line

packages/flutter_tools/lib/src/commands/create_base.dart
```diff
@@ -18,11 +18,27 @@ import '../project.dart';
...
-const _kAvailablePlatforms = <String>['ios', 'android', 'windows', 'linux', 'macos', 'web'];
+const _kAvailablePlatforms = <String>[
+  'ios',
+  'android',
+  'windows',
+  'linux',
+  'macos',
+  'web',
+  'darwin',
+];
```
vashworth: `_kAvailablePlatforms` is used to show the `--platform` options when you do `flutter create --help`. 

`flutter create` can be used to create different types of projects (apps, modules, packages, plugins, etc). The `darwin` platform is only supported for creating plugins, though. I think maybe you can set `allowedHelp` in `addPlatformsOptions` to give it a custom description. Maybe something like "only support for plugins"

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -377,12 +379,23 @@ class CreateCommand extends FlutterCommand with CreateBase {
       includeLinux = false;
       includeMacos = false;
       includeWindows = false;
+      includeDarwin = false;
     } else {
-      includeIos = featureFlags.isIOSEnabled && platforms.contains('ios');
+      final bool darwinRequested = platforms.contains('darwin');
+      final bool darwinSupported = featureFlags.isIOSEnabled || featureFlags.isMacOSEnabled;
```
vashworth: Darwin is only supported for plugins

```suggestion
      final bool darwinSupported = (featureFlags.isIOSEnabled || featureFlags.isMacOSEnabled) && template == FlutterTemplateType.plugin;
```

packages/flutter_tools/lib/src/commands/create.dart
okorohelijah: Based on the new changes, there would be a warning but the fall back would be the default value of `featureFlags.isMacOSEnabled && platforms.contains('macos');` in this case. `printError` or setting to False might not be the best user experience and could cause other things to break

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -104,7 +104,12 @@ class CreateCommand extends FlutterCommand with CreateBase {
           'internally and should generally not be used manually.',
       hide: !verboseHelp,
     );
-    addPlatformsOptions(customHelp: kPlatformHelp);
+    addPlatformsOptions(
+      customHelp: kPlatformHelp,
+      allowedHelp: <String, String>{
+        'darwin': 'A shared platform for iOS and macOS. (only supported for plugins)',
```
vashworth: So before you added this `allowedHelp`, when you did `flutter create --help`, it looked like this:

```console
    --platforms            The platforms supported by this project. Platform folders (e.g. android/) will be generated in the target
                           project. This argument only works when "--template" is set to app or plugin. When adding platforms to a
                           plugin project, the pubspec.yaml will be updated with the requested platform. Adding desktop platforms
                           requires the corresponding desktop config setting to be enabled.
                           [ios (default), android (default), windows (default), linux (default), macos (default), web (default),
                           darwin (default)]
```

Now it looks like this:
```console
    --platforms                 The platforms supported by this project. Platform folders (e.g. android/) will be generated in the
                                target project. This argument only works when "--template" is set to app or plugin. When adding
                                platforms to a plugin project, the pubspec.yaml will be updated with the requested platform. Adding
                                desktop platforms requires the corresponding desktop config setting to be enabled.

          [darwin] (default)    A shared platform for iOS and macOS. (only supported for plugins)
```

It no longer lists the available platforms

Also, I don't think `darwin` should be a default

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -377,12 +384,34 @@ class CreateCommand extends FlutterCommand with CreateBase {
...
+      if (darwinRequested && darwinSupported) {
+        includeDarwin = true;
+        includeIos = featureFlags.isIOSEnabled;
+        includeMacos = featureFlags.isMacOSEnabled;
+      } else if (darwinRequested && !darwinSupported) {
+        includeDarwin = false;
+        globals.printWarning(
+          'Warning: To use the "darwin" platform, you must have both iOS and macOS enabled.\n'
+          'Run "flutter config --enable-ios --enable-macos-desktop" and try again.',
+        );
```
vashworth: Current `darwin` is included in `platforms` by default I think. So when you do `flutter create my_app`, you'll get this warning, which doesn't make sense when creating an app

packages/flutter_tools/lib/src/commands/create_base.dart
```diff
@@ -484,6 +498,10 @@ mixin CreateBase on FlutterCommand {
     if (macOSPlatform) {
       platformsForMigrateConfig.add(SupportedPlatform.macos);
     }
+    if (darwinPlatform) {
+      platformsForMigrateConfig.add(SupportedPlatform.ios);
+      platformsForMigrateConfig.add(SupportedPlatform.macos);
```
vashworth: Won't `SupportedPlatform.ios` and `SupportedPlatform.macos` be added twice?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -420,7 +420,12 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
 ''';
 
   Directory get ephemeralModuleDirectory => parent.directory.childDirectory('.ios');
-  Directory get _editableDirectory => parent.directory.childDirectory('ios');
+  Directory get _editableDirectory {
+    if (parent.manifest.ios?.sharedDarwinSource ?? false) {
+      return parent.directory.childDirectory('darwin');
+    }
+    return parent.directory.childDirectory('ios');
+  }
```
vashworth: Why would the `_editableDirectory` change? That's for the app, not the plugin if I recall correctly?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -988,7 +993,15 @@ class MacOSProject extends XcodeBasedProject {
...
   @override
-  Directory get hostAppRoot => parent.directory.childDirectory('macos');
+  Directory get hostAppRoot {
+    if (parent.manifest.macos?.sharedDarwinSource ?? false) {
+      final Directory darwinDir = parent.directory.childDirectory('darwin');
+      if (darwinDir.existsSync()) {
+        return darwinDir;
+      }
+    }
+    return parent.directory.childDirectory('macos');
```
vashworth: Same, why would `hostAppRoot` change?

vashworth (changes requested):
Okay so I checkout out your branch and tried it myself and somethings not right.

When I did `flutter create --template plugin --platform darwin darwin_plugin`, the structure I got was:

```
darwin_plugin/
  darwin/
    darwin_plugin/
      Package.swift
      Sources/
  ios/
    darwin_plugin.podspec
    darwin_plugin/
      Package.swift
      Sources/
  macos/
    darwin_plugin.podspec
    darwin_plugin/
      Package.swift
      Sources/
  example/
    ios/
    macos/
  ...
```

However, if you look at an [example plugin](https://github.com/flutter/packages/tree/main/packages/shared_preferences/shared_preferences_foundation) that uses the darwin setup, the structure should be like this:

```diff
darwin_plugin/
  darwin/
+   darwin_plugin.podspec
    darwin_plugin/
      Package.swift
      Sources/
-  ios/
-    darwin_plugin.podspec
-    darwin_plugin/
-      Package.swift
-      Sources/
-  macos/
-    darwin_plugin.podspec
-    darwin_plugin/
-      Package.swift
-      Sources/
  example/
    ios/
    macos/
  ...
```

Also, we should have an integration test that would have caught this.

Let me know if you need more help figuring this out.

packages/flutter_tools/lib/src/commands/create_base.dart
okorohelijah: The ios and macos dir was created because of the flag ios and mac(which determines if darwin can be created is set to true). The example app also needs both the flags to be true. I thought it would be best to use those flags. In the new approach, i created a temp flag to store the values so it could be false when creating the plugins and reset after

vashworth:
I checked it out again and the podspec isn't being generated at all this time. It should be located at `darwin_plugin/darwin/darwin_plugin.podspec`. This is likely why tests are failing.

See [comment above](https://github.com/flutter/flutter/pull/176495#pullrequestreview-3483822541) for expected format

packages/flutter_tools/lib/src/commands/create.dart
```diff
@@ -721,6 +771,9 @@ Your $application code is in $relativeAppMain.
       printStatusWhenWriting: printStatusWhenWriting,
     );
 
+    templateContext['ios'] = originalIos;
```
vashworth, line 776: Add a comment explaining to re-add these so the platforms are generated for the example app

okorohelijah: the podspec is generating when i checked - please can you clarify ? and the tests seem to be failing because of the deployment versioning

vashworth: > the podspec is generating when i checked - please can you clarify ? and the tests seem to be failing because of the deployment versioning

1. Run `flutter create --template plugin --platform darwin darwin_plugin`
2. `ls darwin_plugin/darwin`
3. There is no `darwin_plugin.podspec` in the list, but there should be

okorohelijah: ah i see myb- it shows for cocoapods( `--no-enable-swift-package-manager`)  but doesnt show if you have spm enabled `( --enable-swift-package-manager)`

vashworth (approved):
LGTM!

PR #176546, opened 2025-10-05
[ Widget Preview ] Fix type error when retrieving flags from persistent preferences

bkonyi:
Also moves `dtd_services_test.dart` to actually use the `WidgetPreviewScaffoldDtdServices` implementation, which would have caught the typecast issue in the implementation.


packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/dtd_services.dart
```diff
@@ -92,12 +92,14 @@ class WidgetPreviewScaffoldDtdServices with DtdEditorService {
...
   /// Returns null if [key] is not in the map.
-  Future<String?> getPreference(String key) async {
+  Future<Object?> getPreference(String key) async {
     try {
-      final response = StringResponse.fromDTDResponse(
-        (await _call(kGetPreference, params: {'key': key}))!,
-      );
-      return response.value;
+      final response = await _call(kGetPreference, params: {'key': key});
+      return switch (response!.type) {
```
DanTup: What about using `?` instead of `!` and have this fall into the `StateError` case so that if it happens the error message is a bit clearer?

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/dtd_services_test.dart
```diff
@@ -0,0 +1,133 @@
...
+        // Set a preference and ensure it's read back.
+        const kFirstValue = 'foo';
+        await dtd.setPreference(kTestKey, kFirstValue);
+        expect(await dtd.getPreference(kTestKey), kFirstValue);
+        expect(json.decode(preferencesFile.readAsStringSync()), {
+          kTestKey: kFirstValue,
+        });
+
+        // Overwrite kTestKey and ensure it's read back.
+        const kSecondValue = 'bar';
```
DanTup, line 115: Is it worth having a test that removes the key (since I think there was is special handling for nulls/missing keys?)

DanTup (approved):
Some minor comments/questions, but LGTM!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/dtd/dtd_services.dart
bkonyi: SGTM!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/test/dtd_services_test.dart
bkonyi: We definitely should. I've added one now!

PR #176550, opened 2025-10-05
[ Widget Preview ] Fix `WidgetInspectorService` override

bkonyi:
When registering the `WidgetPreviewScaffoldInspectorService`, we were originally setting it after the bindings were initialized. This meant that the widget inspector service extensions were registered with the original `WidgetInspectorService` and were not taking the custom codepaths in the override.

This change moves the `WidgetPreviewScaffoldInspectorService` initialization to before the bindings are initialized.


packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/widget_preview_inspector_service.dart
```diff
@@ -0,0 +1,86 @@
...
+      final previewData = (object.widget as PreviewWidget).preview;
+      _nextNavigationLocation = CodeLocation(
+        uri: previewData.scriptUri,
+        line: previewData.line,
+        column: previewData.column,
+      );
+    }
+    final result = super.setSelection(object, groupName);
+    _nextNavigationLocation = null;
+    return result;
```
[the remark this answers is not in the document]
bkonyi: This isn't true as the `postEvent` call will be made synchronously as the result of a service extension invocation or local selection. This is safe as is.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_inspector_service.dart.tmpl
```diff
@@ -0,0 +1,87 @@
...
+    // Save the location of preview annotation applications so we can override
+    // the navigation target in `postEvent`.
+    if (object is PreviewWidgetElement) {
+      final previewData = (object.widget as PreviewWidget).preview;
+      _nextNavigationLocation = CodeLocation(
+        uri: previewData.scriptUri,
+        line: previewData.line,
+        column: previewData.column,
+      );
+    }
```
DanTup, line 46: Do we have two copies of this implementation now? Is the original version from a previous PR still required?

DanTup (approved):
lgtm with one question about possible duplication

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/widget_preview_inspector_service.dart.tmpl
bkonyi: Good catch! I swear I had deleted the original copy, but clearly not... 🤦

PR #176579, opened 2025-10-06
Fix TextButton.icon breaks focus traversal and ink effect when toggling icon

### LINKED ISSUE #173944: TextButton.icon breaks focus traversal and ink effect when toggling icon
### Steps to reproduce

1. Run the code sample
2. Press tab two times to move the focus on the second TextButton which label is "TextButton.icon"
3. Click on the button or press enter

### Expected results

The icon appears and the focus stays on the button

### Actual results

The icon appears and the focus move to the first button
(Also the Ink animation is interrupted)

### Code sample

```dart

/// Flutter code sample for [SegmentedButton].

void main() {
  runApp(const SegmentedButtonApp());
}

class SegmentedButtonApp extends StatefulWidget {
  const SegmentedButtonApp({super.key});

  @override
  State<SegmentedButtonApp> createState() => _SegmentedButtonAppState();
}

class _SegmentedButtonAppState extends State<SegmentedButtonApp> {
  bool showIcon = false;

  @override
    return MaterialApp(
      home: Scaffold(
        body: TextButtonTheme(
          data: const TextButtonThemeData(
            style: ButtonStyle(
              backgroundColor: WidgetStateProperty.fromMap(<WidgetStatesConstraint, Color>{
                WidgetState.focused: Colors.greenAccent,
                WidgetState.any: Color(0xFFEEEEFF),
              }),
            ),
          ),
          child: Center(
            child: Row(
              spacing: 16,
              children: [
                TextButton(onPressed: onPressed, child: const Text('TextButton')),
                TextButton.icon(
                  onPressed: onPressed,
                  icon: showIcon ? const Icon(Icons.check) : null,
                  label: const Text('TextButton.icon'),
                ),
              ],
            ),
          ),
        ),
      ),
    );
  }

  void onPressed() {
    setState(() {
      showIcon = !showIcon;
    });
  }
}

```

</details>

### Screenshots or Video

<details open>

https://github.com/user-attachments/assets/bc16e480-4d48-4f25-ac03-3e8c30b0f971

</details>

### Flutter Doctor output

```console
Flutter (Channel [user-branch], 3.33.0-1.0.pre.401, on macOS 15.5 24F74 darwin-arm64, locale en-FR)
```

</details>

bleroux:
## Description

This PR changes `TextButton.icon` to avoid building a different widget. When a different widget is created the whole subtree is recreated which leads to various issues (Focus and A11y issues for instance).
The change is similar to https://github.com/flutter/flutter/pull/175810 which fixed the exact same problem for `OutlinedButton.icon`.

## Related Issue

Fixes [TextButton.icon breaks focus traversal and ink effect when toggling icon](https://github.com/flutter/flutter/issues/173944)

## Tests

- Adds 1 test


bleroux:
cc @chunhtai Assigning you for the review because the change is similar to https://github.com/flutter/flutter/pull/175810 that you reviewed recently.

chunhtai (approved):
LGTM, looks like it is the same for ElevatedButton and FilledButton as well. I wish we have some way to share code without exposing additional API

bleroux:
@chunhtai Are there any visual diffs in the Google testing failures? Or are those failures related to tests looking up widgets by type?
My bet it is the second case. Those tests will have to be updated similarly to the flutter gallery test updated in this PR where `find.byType(TextButton)` returns more widgets now as widgets created by `TextButton.icon` are now included in the result and were not before this PR.

chunhtai:
I saw a button that was previous disabled but becomes enabled, maybe some of the enabling or onpress callback is not wire up correctly?

chunhtai:
before 
image
after
image

chunhtai:
This is their code, are we overriding the disable style somehow?

```dart
final style =
        buttonStyle ??
        TextButton.styleFrom(
          foregroundColor: theme.colors.primary,
          disabledForegroundColor: theme.colors.onSurfaceVariant,
          iconColor: theme.colors.primary,
          disabledIconColor: theme.colors.onSurfaceVariant,
          textStyle: buttonSize == ButtonSize.medium
              ? theme.textStyles.labelLarge
              : theme.textStyles.bodyMedium,
        );

return TextButton(
              onPressed: onTap != null
                  ? () {
                      onTap!();
                    }
                  : null,
              style: style,
              child: Text(label),
            );
```

bleroux:
> This is their code, are we overriding the disable style somehow?

Thanks for providing this code extract :pray: 
I can't reproduce the problem on my side and I don't see what could lead to this change : this PR does not change the default constructor, it mainly changes the `TextButton.icon` implementation and the code extract does not use `TextButton.icon`.
It would be interesting to know how the 'buttonStyle' variable in this code snippet is initialized

bleroux:
All checks are green now but this is because the Google testing check did not run automatically after I pushed a new commit.
This new commit sets the default clipBehavior similarly to what was done previously (my goal is to have as little differences as possible compare to before to narrow down the Google testing failure, I don't think this change will help for the Google test failure which was reported).

@Piinks This is a case where the Google testing does not run automatically after a new commit (maybe it will in an indeterminate delay?). The problem here is that someone might jump in and merge this PR, then we will have to revert it later.

Piinks:
Filed https://github.com/flutter/flutter/issues/176721 and reached out to the infra team. I will rebase here to try once more.

bleroux:
@chunhtai We looked again at the change, the screenshot and the code extract and we have no clue. If you can share more of the code extract we might understand how it can happen.

chunhtai:
cl up cl/823197890, this won't fixed the google test failure, but will at least make scuba triaging less confusing

chunhtai (approved):
LGTM

PR #176581, opened 2025-10-06
[ Widget Preview ] Rework UI and theming

bkonyi:
This change consists of several UI changes:

 - Added theming support for light and dark mode based on system preferences
 - Pulled in a subset of the DevTools theming constants
 - Reduced size of widget preview buttons and controls
 - Added initial support for setting theme based on IDE parameters passed as query parameters. This does not include responding to theme changes made after the widget previewer is loaded.
 - Fixed issue where the `WidgetPreviewErrorWidget` would still show controls, even though they aren't relevant

**IDE Theming Example:**
image


packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/theme/_ide_theme_web.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// NOTE: originally from package:devtools_app_shared
```
DanTup, line 5: Looks like this is in devtools_app_shared - is that usable directly here?

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/theme/ide_theme.dart
```diff
@@ -0,0 +1,95 @@
...
+/// screen will be embedded in this view. This will result in the top level tab
+/// bar being hidden, and only the screen specified by the URI path will be
+/// shown.
+enum EmbedMode {
+  embedOne,
+  embedMany,
+  none;
+
+  static EmbedMode fromArgs(Map<String, String?> args) {
+    final embedMode = args[_embedModeKey];
```
DanTup: Are we going to use this here? Currently in Dart-Code I think I'm only passing foreground/background and not setting the embed mode, but I can set it if it would be useful here.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/controls.dart
```diff
@@ -195,21 +174,19 @@ class FilterBySelectedFileToggle extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
-    return _WidgetPreviewIconButton(
-      tooltip: 'Hot restart',
-      onPressed: enabled ? _onRestart : null,
-      icon: Icons.refresh,
+    return _ControlDecorator(
+      child: IconButton(
+        tooltip: 'Hot restart',
+        onPressed: _onRestart,
```
DanTup, line 186: Is the `enabled` condition gone deliberately? (I presume so because the field is also gone, though it's not clear to me why)

DanTup (commented):
Some minor comments/questions, but LGTM!

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/theme/_ide_theme_web.dart
bkonyi: I thought about it, but decided it's better to avoid pulling in that dependency as it could cause conflicts for DevTools extension authors. I figure it's better just to duplicate the code I need here.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/theme/ide_theme.dart
bkonyi: That's a good point. It can probably be removed, we're not using it for anything right now.

packages/flutter_tools/test/widget_preview_scaffold.shard/widget_preview_scaffold/lib/src/controls.dart
bkonyi: Yes, it's gone deliberately. Before, we were just supposed** to be disabling the buttons for the error widget, but still displaying them. This change just removed the controls altogether when displaying the error widget, so there's no need for a flag to disable them anymore.

** This apparently never worked 😅

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/theme/_ide_theme_web.dart.tmpl
```diff
@@ -0,0 +1,44 @@
...
+
+/// Converts a dart:ui Color into #RRGGBBAA format for use in CSS.
+String toCssHexColor(Color color) {
+  // In CSS Hex, Alpha comes last, but in Flutter's `value` field, alpha is
+  // in the high bytes, so just using `value.toRadixString(16)` will put alpha
+  // in the wrong position.
+  String hex(double channelValue) =>
+      (channelValue * 255).round().toRadixString(16).padLeft(2, '0');
+  return '#${hex(color.r)}${hex(color.g)}${hex(color.b)}${hex(color.a)}';
+}
```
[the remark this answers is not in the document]
bkonyi: This is incorrect. `Color` does have `r`, `g`, `b`, and `a` properties as doubles.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/utils/color_utils.dart.tmpl
```diff
@@ -0,0 +1,70 @@
...
+
+    final c = this;
+    return Color.from(
+      alpha: c.a,
+      red: c.r + ((1.0 - c.r) * percent),
+      green: c.g + ((1.0 - c.g) * percent),
+      blue: c.b + ((1.0 - c.b) * percent),
+    );
+  }
+}
```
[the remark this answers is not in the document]
bkonyi: This is incorrect.

packages/flutter_tools/templates/widget_preview_scaffold/lib/src/controls.dart.tmpl
```diff
@@ -3,72 +3,47 @@
...
-          icon: Icons.zoom_out,
-        ),
-        const SizedBox(width: 10),
-        _WidgetPreviewIconButton(
-          tooltip: 'Reset zoom',
-          onPressed: enabled ? _reset : null,
-          icon: Icons.zoom_out_map,
-        ),
-      ],
+    const iconColor = Colors.black;
```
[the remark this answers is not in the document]
bkonyi: This is intentional for now.

DanTup (approved):

PR #176621, opened 2025-10-07
Configure FfiNative resolver on dart:io

mraleph:
This is needed by file watching implementation after dart-lang/sdk@ed6bab847ba2a94a295a3c959157040f6c917fde

Ideally we should actually move this whole code into `io_natives.{h,cc}` on the Dart runtime side to avoid duplication.

### DIFF
--- a/engine/src/flutter/lib/io/dart_io.cc
+++ b/engine/src/flutter/lib/io/dart_io.cc
@@ -17,6 +17,7 @@ namespace flutter {
 
 void DartIO::InitForIsolate(bool may_insecurely_connect_to_all_domains,
                             const std::string& domain_network_policy) {
+  // TODO(https://dartbug.com/61694): move this code into dart_io_api.h
   Dart_Handle io_lib = Dart_LookupLibrary(ToDart("dart:io"));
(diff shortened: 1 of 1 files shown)


aam (approved):

mraleph:
@zanderso @jtmcdole I have put this PR to fix the engine embedding layer (which is currently not initializing `dart:io` correctly). 

I am going to fix duplication (move this code into Dart VM itself), so that we don't hit this again, but in the mean time I would have preferred to fix the engine (so that it is not broken). 

However instead of just going ahead and landing this change I am forced to waste my time in completely pointless back and forth with @stuartmorgan-g about tests. I unfortunately have neither time nor patience for this bureaucratic overhead. You are free to decide what to do with this PR.

stuartmorgan-g:
> However instead of just going ahead and landing this change I am forced to waste my time in completely pointless back and forth with @stuartmorgan-g about tests.

[The referenced discussion](https://discord.com/channels/608014603317936148/1425043527938736310), for context.

zanderso:
test-exempt: Tested in the Dart repo.

zanderso (approved):

PR #176723, opened 2025-10-08
fix: build windows_arm artifacts for cah

jtmcdole:
Adds target to merge queue, the only place that the CAH is used to produce artifacts. Today this is built in post submit and only uploads to the githash artifact.

This should not add these artifacts to release builders as the `enabled_branches` is set to "master"

fixes #176603


zanderso (approved):

jtmcdole:
Autosubmit is drunk; google3 testing wasn't failing.

PR #176728, opened 2025-10-08
impeller: allows access of float uniforms by name

### LINKED ISSUE #176417: Allow uniform bindings by name for FragmentShader
## Description of problem

When using the `FragmentShader` API today one has to carefully count uniform offsets by hand in order to set uniforms.  There should be a way to instead to set uniforms by name to make changing uniforms easier and for easier debugging.

### shader
```glsl
uniform float uScale;
uniform sampler2D uTexture;
uniform vec2 uMagnitude;
uniform vec4 uColor;
```

### today
Generate dart code for shaders that has the uniform indices as constants.

```dart
shader.setFloat(0, 23);  // uScale
shader.setFloat(1, 114); // uMagnitude x
shader.setFloat(2, 83);  // uMagnitude y
shader.setFloat(3, color.red / 255 * color.opacity);   // uColor r
shader.setFloat(4, color.green / 255 * color.opacity); // uColor g
shader.setFloat(5, color.blue / 255 * color.opacity);  // uColor b
shader.setFloat(6, color.opacity);                     // uColor a
shader.setImageSampler(0, image);
```

### proposal
```dart
shader.setFloat(SHADER_USCALE, 23);
shader.setFloat(SHADER_UMAGNITUDE_X, 114);
shader.setFloat(SHADER_UMAGNITUDE_Y, 83);
shader.setFloat(SHADER_UCOLOR_RED, color.red / 255 * color.opacity);
shader.setFloat(SHADER_UCOLOR_GREEN, color.green / 255 * color.opacity);
shader.setFloat(SHADER_UCOLOR_BLUE, color.blue / 255 * color.opacity);
shader.setFloat(SHADER_UCOLOR_ALPHA, color.opacity);
shader.setImageSampler(SHADER_UTEXTURE, image);
```

### runtime access

```dart
shader.setFloat('uScale', 23);
shader.setFloat('uMagnitude', 114, 0);
shader.setFloat('uMagnitude', 83, 1);
```

Having compile time checks for the validity of uniform names is a better user experience and better performance.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/176417

This fixes the linked issue in so far as it is possible today.  This opts for the runtime lookup of bindings since it avoids all sort of messy issues when dealing with codegen.  This doesn't stop us from doing codegen in the future.

This also adds tests for hot reload cases that would fail in the past, like inserting new uniforms.

## Limitations
1) It doesn't handle component names like ".r" or ".x".  This is possible if we want.
1) It doesn't handle nested structs.  This isn't possible with the current shader metadata and may be a limitation with glsl?

## Features
1) Slots can be cached to amortize the cost of looking up the bindings
1) Works with hot reload (tested inserting a uniform before one accessed by name)

## Followup work
I plan on filing the following issues:
1) Map postfixes like ".r" ".z" to offsets so a uniform can be "foo.z"
1) Support nested structs so a uniform name can be "foo.bar"
1) Support codegen for uniforms to get compile-time uniform name support

## Example
```dart
ShaderBuilder(
  assetKey: 'shaders/grayscale.frag',
  (context, shader, _) {
    Color color = Colors.red;
    shader.getUniformFloat('uColor', 0).set(color.r);
    shader.getUniformFloat('uColor', 1).set(color.g);
    shader.getUniformFloat('uColor', 2).set(color.b);
    return BackdropFilter(
      filter: ImageFilter.shader(shader),
      child: Container(
        color: Colors.transparent,
      ),
    );
  },
),
```

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke:
I forgot to stub this in for the web backend, I'll just throw if this is called there for now and file an issue for it.

flar:
A general question - it looks like this only solves the issue for floats. What about ImageSampler uniforms?

Also, Colors and Vec2 are handled by having the developer manually index all of the components. It would be fairly trivial to add `getUniformColor(name)` and `getUniformVec2(name)` just based on the existing support already being added for floats which would also alleviate the need to know how the color components and x,y coordinate values are laid out in the uniforms.

(assuming I got the relative offsets of the color components right in the example below)
```
base class UniformColorSlot {
  UniformColorSlot._(this._shader, this.name, this._index);

  /// Set the Color value of the bound uniform.
  void set(Color c) {
    _shader.setFloat(_index, c.a);
    _shader.setFloat(_index + 1, c.r);
    _shader.setFloat(_index + 2, c.g);
    _shader.setFloat(_index + 3, c.b);
  }

  /// VisibleForTesting: This is the index one would use with
  /// [FragmentShader.setFloat] for the base alpha component of this uniform.
  int get index {
    return _index;
  }

  final FragmentShader _shader;
  final int _index;

  /// The name of the bound uniform.
  final String name;
}

...

class FragmentShader {
  UniformColorSlot getUniformColor(String name) {
    return UniformColorSlot(this, name, _getBaseIndexForName(name));
  }
}
```

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5340,6 +5381,57 @@ base class FragmentShader extends Shader {
...
+  /// ```dart
+  /// shader.getUniformFloat('uScale');
+  /// shader.getUniformFloat('uMagnitude', 0);
+  /// shader.getUniformFloat('uMagnitude', 1);
+  /// shader.getUniformFloat('uColor', 0);
+  /// shader.getUniformFloat('uColor', 1);
+  /// shader.getUniformFloat('uColor', 2);
+  /// shader.getUniformFloat('uColor', 3);
+  /// ```
+  UniformFloatSlot getUniformFloat(String name, [int? index]) {
```
flar, line 5487: The name sounds like it will return a float. How about `getUniformFloatSlot`?

Is there an easy way to have it mimic a map for `shader.uniformFloats['foo']`?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5267,6 +5270,44 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
...
+/// ```
+///
+/// See also:
+///   [FragmentShader.getUniformFloat] - How [UniformFloatSlot] instances are acquired.
+///
+base class UniformFloatSlot {
+  UniformFloatSlot._(this._shader, this.name, this.offset, this._index);
+
+  /// Set the float value of the bound uniform.
+  void set(double val) {
```
flar, line 5366: Rather than a method that acts like a setter and calling it `set`, how about an actual setter for the name `value`?
```
  void set value(double val) ...
```

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5340,6 +5381,57 @@ base class FragmentShader extends Shader {
...
+    index ??= 0;
+
+    if (index < 0) {
+      throw ArgumentError('Index `$index` out of bounds for `$name`.');
+    }
+
+    int offset = 0;
+    bool found = false;
+    const int sizeOfFloat = 4;
+    for (final dynamic entryDynamic in _program._uniformInfo) {
```
flar: Perhaps turn this into a map in the constructor rather than looking it up dynamically for each get of a slot? Or, a self-initializing getter for said map.

Also, move the construction of that map into the program object so it only needs to be done once per program rather than once per shader or once per getFooSlot()?

flar (commented):
Some API and performance notes.

engine/src/flutter/lib/ui/painting.dart
flar: Do uniforms have getters?

gaaclarke:
> A general question - it looks like this only solves the issue for floats. What about ImageSampler uniforms?

Correct, that will be handled in a different PR.

> Also, Colors and Vec2 are handled by having the developer manually index all of the components. It would be fairly trivial to add `getUniformColor(name)` and `getUniformVec2(name)` just based on the existing support already being added for floats which would also alleviate the need to know how the color components and x,y coordinate values are laid out in the uniforms.
> 
> (assuming I got the relative offsets of the color components right in the example below)
> 
> ```
> base class UniformColorSlot {
>   UniformColorSlot._(this._shader, this.name, this._index);
> 
>   /// Set the Color value of the bound uniform.
>   void set(Color c) {
>     _shader.setFloat(_index, c.a);
>     _shader.setFloat(_index + 1, c.r);
>     _shader.setFloat(_index + 2, c.g);
>     _shader.setFloat(_index + 3, c.b);
>   }

Yep, I thought about that.  I think that's a good idea but I don't think it should live in `dart:ui`.  This provides the fundamental tools for someone to make that library.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: > The name sounds like it will return a float. How about getUniformFloatSlot?

I had that at first but it seemed superfluous considering the return type of the function.  It looked much more verbose when you are grabbing many of them.  Let's see what chinmay thinks.

> Is there an easy way to have it mimic a map for shader.uniformFloats['foo']?

There is, but in order for us to make that work we'd need to be able to do things like `shader.uniformFloats['foo.bar[3]']` which may not be possible considering some of the limitations chinmay was mentioning.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: > Rather than a method that acts like a setter and calling it set, how about an actual setter for the name value?

I had that at first, but it's actually more characters and not a clear win on clarity.  It get's weird when the point is to be one directional too.
```
foo.set(1.0);
foo.value = 1.0;
```

> Do uniforms have getters?

They have not had them.  I decided to preserve those semantics, ie FragmentShader has setFloat, not getFloat.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: It can't be in the constructor because it depends on _uniformInfo which is set in the c++ side.  It's dynamic because of the limitations with the dart C api.  I was a bit concerned about creating derivative data because I wanted to make sure this would work with hot reload.

flar:
> > A general question - it looks like this only solves the issue for floats. What about ImageSampler uniforms?
> 
> Correct, that will be handled in a different PR.
> 
> > add `getUniformColor(name)` and `getUniformVec2(name)`
> > ```
> > base class UniformColorSlot {
> > }
> > ```
> 
> Yep, I thought about that. I think that's a good idea but I don't think it should live in `dart:ui`. This provides the fundamental tools for someone to make that library.

Someone could do any of this as a separate library, though. Why even have these new classes when all you need is "getOffsetForName(String)" and let them do their library.

One of the advantages of doing the Color and Vec2 (and Matrix) slot objects is that in returning them we could do some type checking (might require some extra info from native for explicit types, or at least it could check that a Color name is associated with 4 floats). Or, just expose the type/count info via methods and let the external library do that checking as well:
- (optional) name-to-type
- name-to-count
- name-to-offset

are all that is needed to enable an external library. The (optional) type would be restricted to only the standard types and need to have a "custom" type for when someone makes up their own struct in the program.

gaaclarke:
> Someone could do any of this as a separate library, though. Why even have these new classes when all you need is "getOffsetForName(String)" and let them do their library.

I would be open to something like that.  I was thinking I'd have to do more to make sure that hot reload would work, but happily it seems to work fine like this.  I was worried I'd have to keep weak references around to slots so that when we hot reload we could invalidate the slots to get them to rebind.  There are symbols that seem related to hot reloading that we definitely wouldn't want to make public.  We'll see if chinmay sees something I didn't with regard to hot reloading.

There's definitely an art to drawing the line and there are tradeoffs for every place we could decide to draw it.  Having the most fundamental entity being useful and guaranteed to work with hot reload seems like a good place to draw it though.

engine/src/flutter/lib/ui/painting.dart
flar: That's where the self-initialization (lazy) that I mentioned comes into play. Create a map on first access - presumably all of the info is available by the first time they ask for a Slot...?

Ah, hot reload. Do the Program objects get recompiled and relinked on a hot reload?

flar: The same thing happens if they save a Slot object - it could become stale on a hot reload. It's basically an encapsulation of the current layout of the uniforms.

gaaclarke: > Ah, hot reload. Do the Program objects get recompiled and relinked on a hot reload?

Yep, they do.

> The same thing happens if they save a Slot object - it could become stale on a hot reload.

Yea, that's why I wanted to have access to the objects, so I could invalidate them after a hot reload.  I've attempted inserting a new uniform in the middle and everything seems to work as is without having to invalidate them.  I'm not sure how it works though.  The new indices must be getting added at the end so the old indices still are valid?  I haven't played around with every possible reload.

flar: I'm not clear. What works? I can see this working on a reload:

```
  shader.getUniformFloat('foo').set(1.0);
```

That doesn't save the object so it gets recomputed every time this code is executed.

But:

```
  // in setup code
  myfield = myshader.getUniformFloat('foo');
  ...
  ...
  // in different code
  myfield.set(2.0); // who knows what this will set as `myfield` could be stale
```

If we are worried about the offsets getting stale, then we should not be creating the Slot objects at all, instead we should provide:

```
class FragmentShader {
  void setFloat(String name, double value, [int offset]);
}
```

flar: > The new indices must be getting added at the end so the old indices still are valid?

I would be extremely surprised if that happened. How would it know?

flar: Did you dump the uniform data before and after the shader change and the hot reload?

gaaclarke: > That doesn't save the object so it gets recomputed every time this code is executed.

> I would be extremely surprised if that happened. How would it know?

Go ahead and give it a try.  I've noted what I've tested in the pr description =)

> instead we should provide:

This approach allows us to amortize the cost of calculating the binding while also keeping bindings stable across hot reloads.  That approach doesn't allow that.

gaaclarke:
draft posted for image samplers: https://github.com/flutter/flutter/pull/176749

gaaclarke:
I just tried hot reload again to show @flar and it doesn't work.  I may have accidentally tested this with hot restart.  I'll update update this.  I figured I'd have to account for this, that's why there is a slot class.  I'm dropping back to draft until it's fixed.  I don't think there is a mechanism to do automated tests for this unfortunately.

gaaclarke:
I've update the logic to handle hot reloading.  I've also added hot reloading tests.  Now it should be apparent why we have slot objects which I think addresses the majority of the feedback above.  PTAL.

engine/src/flutter/lib/ui/fixtures/shaders/general_shaders/uniforms.frag
```diff
@@ -4,6 +4,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// Note: Don't update this without updating `uniforms_reordered.frag`.
```
flar: Grammar nit: ~"this" is not a noun.~ My grade school teacher would always say "this what?"

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5222,6 +5222,7 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
   }
 
   String? _debugName;
+  final List<WeakReference<FragmentShader>> _shaders = <WeakReference<FragmentShader>>[];
```
flar, line 5225: Since we dynamically look them up every time, wouldn't it be enough to just add a version and to reload the data when the version changes?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5275,14 +5325,61 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
...
+  }
+
+  /// VisibleForTesting: This is the index one would use with
+  /// [FragmentShader.setFloat] for this uniform.
+  int get shaderIndex {
+    return _shaderIndex;
+  }
+
+  final FragmentShader _shader;
+  int _shaderIndex;
```
flar, line 5377: Rather than a bare int, how about an int holder, like `List<int>` or something more specific. Then the shader just has to update the holder with the new index rather than having to register all of the objects that have a bare copy of the (potentially) stale data?

`Shader::List<IntHolder> _slots;`
`UniformFloatSlot::IntHolder _index;`
`UniformFloatSlot::int get index => _index.value;`

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5267,6 +5268,55 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
     if (result.isNotEmpty) {
       throw result; // ignore: only_throw_errors
     }
+
+    // Update all the named bindings.
+    program._shaders.removeWhere((WeakReference<FragmentShader> shaderReference) {
```
flar, line 5273: If the code had a map of the old names to the old slots and the slot indices were held in a shared "holder" object, then it wouldn't have to keep weak references to anything, it could create a new list of "holder" objects, copy over the references to the uniforms that still exist. Mark the old holder objects that stopped existing as invalid, and then create new holder objects for the brand new uniforms that weren't there before. No outstanding shaders or slots would need to be updated and no weak lists would be needed.

flar (commented):
I think there is a far simpler way to manage the slots dynamically with just 1 level of indirection instead of registered weak lists of objects that contain stale data.

engine/src/flutter/lib/ui/painting.dart
flar:
```
class UniformIndexHolder {
  int get value => _index;
  int _index;
}
```

Then the shader code could update the _index field directly and all outstanding instances would immediately have the new value, but outside of the shader code, the value should be "read only, but sometimes it might change". (volatile?)

flar: The main thing, though, is that you need to remember which `UniformIndexHolder` is associated with which field so that you update it in place when it changes, and can remove stale uniforms and add new ones. Maybe:

```
class UniformIndexHolder {
  int get value {
    if (_index < 0) throw 'stale uniform $uniformName';
    return _index;
  }
  int _index;
  final String uniformName;

  // ? - final String shaderName; // useful?
}
```

Having the name in there means fewer fields in the objects used to set the data. They just need the Holder object to have access to both the index and the name of the field with a forwarding getter `String get name => _holder.name;` instead of storing the name.

engine/src/flutter/lib/ui/painting.dart
flar: (Note with the indirect slot holder concept below, this is probably moot since we would no longer look them up dynamically if they save the wrapper?)

engine/src/flutter/lib/ui/painting.dart
gaaclarke: > Then the shader just has to update the holder with the new index rather than having to register all of the objects that have a bare copy of the (potentially) stale data?

There's never any stale data.  The weak reference gets rid of any stale data.  By doing it this way we are offloading the cost of hot load support to when hot load is really used.  I want the amortized cost of setting uniforms by name to be the same as using the `setFloat` api.  This proposed solution does 2 index per uniform set and has added indirection with an IntHolder object.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: I responded to this above but this approach with the WeakReferences allows setting uniforms by name to be as fast using setFloat.  The approach in the PR puts all the overhead at hot reload time, not when it is used which is what we want since setting uniforms can happen multiple times per frame where hot reload is much less frequent.

engine/src/flutter/lib/ui/fixtures/shaders/general_shaders/uniforms.frag
gaaclarke: Your grade school teacher was wrong, it's a demonstrative pronoun =)

engine/src/flutter/lib/ui/painting.dart
gaaclarke: We could start only storing weakreferences in conditions where hot reload is available (debug builds?).  That would be an optimization.

engine/src/flutter/lib/ui/painting.dart
flar: Just because they are quickly updated doesn't mean that they don't - in the instant that a shader changes - contain stale data that needs to be updated. If there was no stale data then you wouldn't need the update method to track them down to update their stale data to the new data. Sure, the stale indices don't last long and if this happens synchronously then they don't get used, but they still have to be modified - that is stale data.

With your solution there is stale data located in every object that uses indices. It gets cleaned up aggressively when the notification comes in, but it is stale data and it needs to be updated. And the number of objects with the stale data that must be updated grows.

With my suggestion there is stale data located in a fixed number of objects that are easy to update within a finite predictable amount of time.

Also, what is special about `List<WeakRef>`? Sure, the GC will clear the target when there are no more strong references, but the WeakRef wrappers will live on and take up space in the list until they are removed. If the shader never changes, then the list grows without bound on every new field fetch.

engine/src/flutter/lib/ui/painting.dart
flar: You leak all of the WeakRef objects, or to be more precise, you accumulate them all with no cleanup until a shader change comes in. They may not continue to contain references if the using object went away, but only the target is nulled, not the WeakRef object itself, so the list will just keep accumulating ref objects. If the developer is not actively changing their shaders to trigger the cleanup, they will grow unbounded if the app keeps fetching them.

And there is overhead for maintaining weak refs. It isn't high if you are dealing with long-lived objects that don't tend to accumulate, but in this case we have no control over how often the app fetches a new slot object.

engine/src/flutter/lib/ui/fixtures/shaders/general_shaders/uniforms.frag
flar: It can be used in a conversation that already referred to something, but it is dangling here. There is no conversation that introduced an antecedent to which you could be referring.

This file?
This comment?
The following comment?
The next declaration in the file?
This what?

This comment is ambiguous.

flar (changes requested):

engine/src/flutter/lib/ui/painting.dart
gaaclarke: I'll respond to this below since I think these are all the same conversation.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: Yea, fair point.  It's only a leak if people don't trash their FragmentShader objects, but it's common for people to do that across frames and someone could be making multiple uniforms per frame.  That isn't optimal usage but it also isn't forbidden.  WeakReferences are tiny but it could add up for long running processes.  Here's options:

1) Only do this bookkeeping for debug builds since you can't hot reload in release - This may make testing a bit more complicated and the code is a little less clear.
2) Clear out the list when new slots are created - This adds more overhead to creating slots, but not caching slots is already suboptimal so that's not that big of a deal.
3) Clear out the list once per frame when a new one is created - I don't think there is a good latch in dart:ui for this.  It's more complicated than options 2.

Done.  I've opted for the 2nd option.

engine/src/flutter/lib/ui/fixtures/shaders/general_shaders/uniforms.frag
gaaclarke: I've added it.  I don't think it's ambiguous considering it is a file level comment, but if you were confused by it it's worth changing.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5275,14 +5325,61 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
...
   external void _constructor();
 
   @Native<Handle Function(Pointer<Void>, Handle)>(symbol: 'FragmentProgram::initFromAsset')
   external String _initFromAsset(String assetKey);
 
   /// Returns a fresh instance of [FragmentShader].
-  FragmentShader fragmentShader() => FragmentShader._(this, debugName: _debugName);
+  FragmentShader fragmentShader() {
+    final FragmentShader result = FragmentShader._(this, debugName: _debugName);
+    _shaders.add(WeakReference<FragmentShader>(result));
```
flar, line 5341: You should clean out this list on add as well.

flar (changes requested):
One last potential leak/infinite accumulation in the fragmentShader method that needs to be fixed.

Otherwise, I feel like this is overly complicated for the solution.

- We now have to accumulate shader references which we didn't use to require.
- Accumulating them also means cleaning them up every time we make one - more overhead.
- The total plumbing just to deal with the weak refs seems far more than worth it for the comparative cost of using a single indirect reference of the slot numbers - one extra field fetch per uniform update vs keeping lists and cleanup and maintenance?

Also, looking up slots should be faster with a map for those apps that choose not to keep the slot objects around. The map would have to be reinitialized when the shader changes, but looking up an index would be a simple map lookup instead of a list scan.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: > We now have to accumulate shader references which we didn't use to require.

We didn't use to require it, but we also didn't support hot reloading.  That bookmarking addition was added to fix the existing bug in the code.  Remember adding new uniforms didn't expand the float array.

> Accumulating them also means cleaning them up every time we make one - more overhead.

This is overhead which is associated with suboptimal usage.  Users can cache fragmentshaders and slots to avoid the overhead, which they should do naturally to avoid allocations.

> The total plumbing just to deal with the weak refs seems far more than worth it for the comparative cost of using a single indirect reference of the slot numbers - one extra field fetch per uniform update vs keeping lists and cleanup and maintenance?

Yes, anything that happens more than once per frame should have a higher scrutiny for performance.  Remember that people will be driving animations in some cases with uniforms.

> Also, looking up slots should be faster with a map for those apps that choose not to keep the slot objects around. The map would have to be reinitialized when the shader changes, but looking up an index would be a simple map lookup instead of a list scan.

For small sizes linear search can be faster than hash maps.  This performance is associated with hot reload which isn't as big of a deal.

> You should clean out this list on add as well.

Done.

Hopefully my concern for performance here doesn't seem capricious and I'm communicating it clearly.  I've mentioned it a few times but just to be clear I'll lay it out again: I want accessing uniforms by name to have an amortized cost that is equal to using `setFloat`.  I don't want people to choose a worse user experience to get better performance.  Performance matters here because updating uniforms happens multiple times per frame when driving animations.

Thanks for pushing on this Jim.

flar: I understand the concern about performance and I actually share it, but I just don't think a single indirection to set the uniform costs enough to worry about, even in an animation. Dart should be really quick about that, and it has so much work to do on each frame outside of setting uniforms that is way more complicated.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5299,16 +5397,21 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
...
+  final FragmentProgram _program;
+
   final String? _debugName;
 
   static final Float32List _kEmptyFloat32List = Float32List(0);
   Float32List _floats = _kEmptyFloat32List;
+  final List<WeakReference<UniformFloatSlot>> _slots = <WeakReference<UniformFloatSlot>>[];
+
+  void _reinitialize() {
+    _floats = _constructor(_program, _program._uniformFloatCount, _program._samplerCount);
```
flar, line 5413: I just noticed that this dumps all of the values set on a shader on hot reload. Could someone be depending on the uniforms being carried over? If we clear them every time we use the shader that might be one thing, but they could have a program that creates a shader once, sets the needed values on it for a given use case (that will remain stable on every frame), and then never set the uniforms ever again.

Does this complicate the hot reload support?

flar (approved):
A question about reload resiliency, but otherwise looks good to go/ better than we had before.

flar:
A question about the importance of hot reload - how often does one change the uniforms on a shader but then not need to rewrite the code anyway to use the new uniforms (or not use old ones). The only case that really helps here is rearranging them, but keeping all of their functions the same. Is that common compared to adding a new uniform that will now not have any values assigned to it without editing the shader-using code?

engine/src/flutter/lib/ui/painting.dart
gaaclarke: I think it's technically possible but unlikely given Flutter is reactive.  See the example for flutter_shaders: https://pub.dev/packages/flutter_shaders/example and the example code https://github.com/flutter/samples/blob/main/simple_shader/lib/main.dart

gaaclarke:
> A question about the importance of hot reload - how often does one change the uniforms on a shader but then not need to rewrite the code anyway to use the new uniforms (or not use old ones). The only case that really helps here is rearranging them, but keeping all of their functions the same. Is that common compared to adding a new uniform that will now not have any values assigned to it without editing the shader-using code?

Are you saying the PR here doesn't support (edit shader to have a new uniform -> edit dart code to set the uniform -> hot reload)?  That works with my testing.  What are you thinking?

(edit shader to have a new uniform -> hot reload -> edit dart code to set the uniform -> hot reload) should work too.

flar:
> > A question about the importance of hot reload - how often does one change the uniforms on a shader but then not need to rewrite the code anyway to use the new uniforms (or not use old ones). The only case that really helps here is rearranging them, but keeping all of their functions the same. Is that common compared to adding a new uniform that will now not have any values assigned to it without editing the shader-using code?
> 
> Are you saying the PR here doesn't support (edit shader to have a new uniform -> edit dart code to set the uniform -> hot reload)? That works with my testing. What are you thinking?
> 
> (edit shader to have a new uniform -> hot reload -> edit dart code to set the uniform -> hot reload) should work too.

Adding a new uniform to a shader won't magically edit the Dart code to make use of the new uniform, and removing that uniform won't stop the code from trying to set it. So, most uniform-changing shader modifications require editing the Dart code anyway. The uniform indices may get automatically updated, but its likely that they'll need to add some setFloat/getUniformFloat calls anyway.

PR #176742, opened 2025-10-08
[HCPP] Properly remove hcpp views that are no longer visible

gmackall:
Fixes https://github.com/flutter/flutter/issues/175882.

This ended up being a very similar bug to https://github.com/flutter/flutter/pull/173881, and in my understanding comes from the same incorrect assumption. In old PV land these views would stop displaying if we stopped sending updates to them. But in HCPP the underlying implementation ends up using an Android SurfaceFlinger which continues displaying what it was last told to. So again we have a bug here were we aren't taking the new responsibility of clearing that we need to. See https://github.com/flutter/flutter/pull/173881#discussion_r2283287104

Other solutions would be to just always call hide on the set of (keys of view_params which are not in composition order) on every frame (which would avoid this mirroring of state), or to modify the jni here to pass over one big bundle with all the information required to do all the existing calls as well as determine which views need to be hidden. But we are already doing this same mirroring of visibility state for the overlay layer (see Hide/ShowOverlayLayerIfNeeded()), so I think it is reasonable to take the same approach here.

This is sort of a follow up to https://github.com/flutter/flutter/pull/162908 as well, this PR properly removes the view when it is the last platform view, but does not fix the case that this PR attempts to more generally fix where we go from 2->1 (or n->n-1 > 0) pvs.

## Pre-launch Checklist

- [x] I signed the [CLA].


gmackall:
As usual, golden images for your viewing.
Without change:
hybrid_composition_pp_platform_view two_boxes hybrid_composition_pp_platform_view only_one_box



With change:
hybrid_composition_pp_platform_view two_boxes hybrid_composition_pp_platform_view only_one_box

gmackall:
@gaaclarke Reid suggested I request your review on hcpp stuff, but let me know if you don't think you're the right person to review here.

dev/integration_tests/android_engine_test/test_driver/hcpp/platform_view_clear_hidden_test.dart
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  tearDownAll(() async {
+    await nativeDriver.close();
+    await flutterDriver.close();
+  });
+
+  test('verify that HCPP is supported and enabled', () async {
+    final Map<String, Object?> response =
+    json.decode(await flutterDriver.requestData('')) as Map<String, Object?>;
```
gaaclarke: This is odd, requesting data with an empty string.  I would have expected some sort of payload to the request.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -179,9 +179,21 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
+          jni_facade->hidePlatformView2(view_id);
         }
+
         jni_facade_->onEndFrame2();
       }));
+
+  views_visible_last_frame_.clear();
+  for (int64_t view_id : composition_order_) {
+    views_visible_last_frame_.insert(view_id);
+  }
```
gaaclarke: nit: 
```suggestion
  views_visible_last_frame_.insert(composition_order_.begin(), composition_order_.end());
```

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.h
```diff
@@ -135,6 +136,9 @@ class AndroidExternalViewEmbedder2 final : public ExternalViewEmbedder {
   // mutation stack.
   std::unordered_map<int64_t, EmbeddedViewParams> view_params_;
 
+  // The set of platform views that were visible in the last frame.
+  std::set<int64_t> views_visible_last_frame_;
```
gaaclarke: Use absl::flat_hash_set (https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/refs/heads/lts_2018_12_18/absl/container/flat_hash_set.h)

engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h
```diff
@@ -244,6 +244,8 @@ class PlatformViewAndroidJNI {
                                       int32_t viewHeight,
                                       MutatorsStack mutators_stack) = 0;
 
+  virtual void hidePlatformView2(int32_t view_id) = 0;
```
gaaclarke, line 247: Inconsistent bit depth, sometimes int32_t is used sometimes int64_t.  Are the view_id's incremented per frame rendered?  If so we should make sure we use int64_t.

gaaclarke (commented):
looking good, just a few questions and a swap of container is needed.

engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h
gmackall: No, the numbers are consistent for a given view on screen across frames. I'm not actually sure what happens if you scroll past one, then to none, then down to a second view (if it is 0 or 1), but either way the ids won't get close to the limit here.

I copied this from the `onDisplayPlatformView2` above which also has this discrepancy, but can change if you prefer.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gmackall: changed

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.h
gmackall: changed

dev/integration_tests/android_engine_test/test_driver/hcpp/platform_view_clear_hidden_test.dart
gmackall: I've copied this as an initial smoke test from Jonah's original HCPP integration test
https://github.com/flutter/flutter/blob/master/dev/integration_tests/android_engine_test/lib/hcpp/platform_view_main.dart

just to verify we are using hcpp for the specific integration test. Agreed it's a little weird though - it looks like there isn't a `requestData` signature that doesn't take a string so I'm not sure what the immediate replacement would be. Can look into it if you prefer.

engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h
gaaclarke: That's a pretty nasty bug to try track down because users will say something like "leave app running overnight, then it blows up".  They can't provide reasonable reproductions.  I'd prefer we used int64 across the board.  If you don't feel comfortable changing it here we can do it in a different PR.  The gotcha is you have to change the "I" in the JNI signature to "J". ( https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/types.html )

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -72,9 +72,17 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
 
   if (!FrameHasPlatformLayers()) {
     frame->Submit();
-    // If the previous frame had platform views, hide the overlay surface.
-    HideOverlayLayerIfNeeded();
-    jni_facade_->applyTransaction();
+    task_runners_.GetPlatformTaskRunner()->PostTask(fml::MakeCopyable(
```
gmackall, line 75: Small additional change here, to clear the pvs also in the other branch of flow when there is no pv on screen but there was last frame, and then also to clear out the set.

I had thought this wasn't needed at first, but running the scrolling test again I found that without making these hide calls we still have a sliver of PV left on the screen, so it seems it is necessary both in the n->n-1 case as well as the 1->0 case.

I also moved this work to the same thread as the one we use for the onDisplayPlatformView2 work below. This is a follow up to the long discussion here https://github.com/flutter/flutter/pull/173881#discussion_r2287452036, though I don't think its necessary to fix this bug so I can revert it if you think it should be done in a follow up PR.

engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h
gmackall: Yeah, thats fair. Do you mind if I make it a follow up? I can do it right away after this but I would find it easier to do that to all the signatures on clean branch.

dev/integration_tests/android_engine_test/test_driver/hcpp/platform_view_clear_hidden_test.dart
gaaclarke: Okay, no, that's fine.  If that's how it's being done elsewhere we can grandfather it in.  I would have expected something like `flutterDriver.requestData()` or `flutterDriver.requestData('hccp_support')`.  Getting the info we want without any sort of meaningful request is just weird.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -72,9 +72,17 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
+        [&, jni_facade = jni_facade_,
```
gaaclarke:
```suggestion
        [jni_facade = jni_facade_,
```

Please capture the variables explicitly since we are jumping across threads.

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
```diff
@@ -72,9 +72,17 @@ void AndroidExternalViewEmbedder2::SubmitFlutterView(
...
+         views_visible_last_frame = views_visible_last_frame_]() {
+          HideOverlayLayerIfNeeded();
```
gaaclarke: This is potentially a dangling pointer with implicitly captured `this`.

gaaclarke (changes requested):

engine/src/flutter/shell/platform/android/jni/platform_view_android_jni.h
gaaclarke: sgtm

engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc
gaaclarke: it's fine, there is a bug in this change though that might be a pain to address.  it might be worth putting it in a different pr, your call.

gmackall: Hmm it looks like moving this to the other thread is actually required, as it crashes without it (very interesting, because on the last PR we didn't understand how it wasn't crashing as is).

The way to fix this would normally be to make a shared/weak ptr and pass that in, right? But the `this` in question here is declared as unique so I can't do that
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_wrapper.cc#L37

I don't really understand the consequences of changing that here - would the fix here normally be to just make that `make_shared` instead of `make_unique`? Or is there another pattern that makes more sense here?

gmackall: I made the instance of this class shared, and then I used a weak ptr and explicitly copied. Let me know if any of that looks wrong.

I assume this is also a problem with the lines below, and if the fix here is right I should probably fix that in a follow up pr too?
https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/platform/android/external_view_embedder/external_view_embedder_2.cc#L163-L167

gaaclarke: Okay, I looked into this a bit more.  What you've done is safe.  It may not be what we want though.  The safety we are adding here stops us from problems when we delete the AndroidExternalViewEmbedder2 without calling DestroySurfaces or calling SubmitFlutterView after having called DestroySurfaces.  This is because DestroySurfaces synchronizes against the platform task runner.

nit:  It would be nice to have some assertions in the code about what methods are intended to be called on what task runner.  that can be done with T`askRunner::RunsTasksOnCurrentThread()`

I think what I'd prefer is we go back to the unique_ptr.  Capture the pointer explicitly and add a comment that the pointer is guaranteed to not be dangling as long as DestroySurfaces is called before the embedder is deleted.  If you think there is a situation where that isn't the case we can keep the shared_ptr approach.

We are also modifying and reading overlay_layer_is_shown_ from different threads.  That instance variable should be converted to an std::atomic<bool> at the least (make sure the .load() and .store() methods with the atomic).

gmackall: Changed back to using a unique pointer and added the comment

gaaclarke: change and comment looks good, thanks

gmackall: Also updated overlay_layer_is_shown_ to be atomic

gaaclarke (approved):
LGTM modulo using int32 at jni bridge which we discussed as doing in a different PR.  Thanks @gmackall, this code is pretty gnarly.

gmackall:
Google testing seems to be a flake, overriding

PR #176781, opened 2025-10-09
Fix Dialog Semantics label and flags for mismatched platforms

huycozy:
- Address a partial issue within https://github.com/flutter/flutter/issues/176566
- Fix  https://github.com/flutter/flutter/issues/177001
- In this PR:
    - proposes using `defaultTargetPlatform` instead of platform from theme for Semantics title's `label` and `namesRoute` flag in AlertDialog and SimpleDialog widgets (see use case at https://github.com/flutter/flutter/issues/176566#issue-3486244630); 
    - improve documentation on each widget for more accessible to users
    -  add some new tests for this change, and also update existing tests.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/dialog.dart
+++ b/packages/flutter/lib/src/material/dialog.dart
@@ -406,14 +406,6 @@ class Dialog extends StatelessWidget {
 /// For finer-grained control over the sizing of a dialog, consider using
 /// [Dialog] directly.
 ///
-/// ## Accessibility
-///
-/// The accessibility behavior of [AlertDialog] is platform adaptive, based on
-/// the device's actual platform rather than the theme's platform setting. This
-/// ensures that assistive technologies like VoiceOver on iOS and macOS receive
-/// the correct platform-specific semantics label and appropriate flags, even
-/// when the app's theme is configured to mimic a different platform's appearance.
-///
 /// See also:
 ///
 ///  * [SimpleDialog], which handles the scrolling of the contents but has no [actions].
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/dialog.dart
```diff
@@ -405,6 +406,14 @@ class Dialog extends StatelessWidget {
 /// For finer-grained control over the sizing of a dialog, consider using
 /// [Dialog] directly.
 ///
+/// ## Accessibility
+///
+/// The accessibility behavior of [AlertDialog] is platform adaptive, based on
+/// the device's actual platform rather than the theme's platform setting. This
+/// ensures that assistive technologies like VoiceOver on iOS and macOS receive
+/// the correct platform-specific semantics label and appropriate flags, even
```
chunhtai: If we want to call out the platform adaptiveness, we should be more specific and how they affect the screen reader in each platform. In this case, I don't think we should mention anything about this, because setting name route to false on iOS and macos seems to make the title not be announced twice. instead of causing an actual different between platform

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1154,6 +1163,14 @@ class SimpleDialogOption extends StatelessWidget {
 /// ```
 /// {@end-tool}
 ///
+/// ## Accessibility
+///
+/// The accessibility behavior of [SimpleDialog] is platform adaptive, based on
+/// the device's actual platform rather than the theme's platform setting. This
+/// ensures that assistive technologies like VoiceOver on iOS and macOS receive
+/// the correct platform-specific semantics label and appropriate flags, even
+/// when the app's theme is configured to mimic a different platform's appearance.
```
chunhtai: same

chunhtai (commented):
LGTM, just comment on doc

packages/flutter/lib/src/material/dialog.dart
huycozy: Thank you for reviewing this! It makes sense as the additional docs seem to be overstated for a small different behavior. Thus, I removed these documentation changes.

chunhtai (approved):
LGTM

PR #176811, opened 2025-10-09
Set up a version of build_ios_framework_module_test that only runs on x64 machines and extend its timeout

jason-simmons:
The "Mac build_ios_framework_module_test" builder was being scheduled on either x64 or arm64 machines.  The test takes longer when run on an x64 machine and would sometimes exceed the default 30 minute timeout.

This PR changes that builder to only run on x64 and use a 45 minute timeout.  (There is already a separate "Mac_arm64 build_ios_framework_module_test" that runs only on arm64)

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -3717,6 +3717,7 @@ targets:
 
   - name: Mac_x64 build_ios_framework_module_test
     recipe: devicelab/devicelab_drone
+    bringup: true
     timeout: 60
     properties:
       add_recipes_cq: "true"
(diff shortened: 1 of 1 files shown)


jtmcdole (approved):

zanderso:
@jmagman @vashworth Do we still need this test to run on both intel and arm?

vashworth:
> @jmagman @vashworth Do we still need this test to run on both intel and arm?

I don't see any particular reason that this test specifically needs to run on both.

PR #171916, opened 2025-07-09
Properly lay out and position RenderWebImage

harryterkelsen:
Fixes https://github.com/flutter/flutter/issues/163288
Fixes https://github.com/flutter/flutter/issues/164405

The previous behavior of `RenderWebImage` for layout and positioning did not work for all values of `BoxFit` and `AlignmentGeometry`. This PR gets `RenderWebImage` to properly lay out and offset its child (the `` element).

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #176831, opened 2025-10-10
Fix Image.network not using cache when headers are specified

rajveermalviya:
Fix a bug introduced by 97aae2ad3 (#171916), where `hashCode` is giving different results for values which `==` says are equal. Which broke caching for uses of `Image.network` widget when `headers` map was passed.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/painting/image_provider_network_image_test.dart
```diff
@@ -288,6 +288,51 @@ void main() {
...
+    result1.addListener(
+      ImageStreamListener((ImageInfo image, bool synchronousCall) {
+        imageAvailable1.complete();
+      }),
+    );
+    await imageAvailable1.future;
+    expect(imageCache.liveImageCount, 1);
+
+    // NetworkImage should not make another request.
+    mockHttpClient.thrownError = Error();
```
gnprice: This line doesn't totally reassure me that the test will notice if the NetworkImage does make another request. With the current implementation of _FakeHttpClient, that holds because `thrownError` takes precedence over `request`; but there's no reason it has to be that way, and someone could well refactor that in the future without noticing that it causes this test to be ineffective.

Let's instead put this in the same terms as the `.request.response` setup above, so that it clearly overrides it:

```suggestion
    // NetworkImage should not make another request.
    mockHttpClient.request.response.statusCode = HttpStatus.badRequest;
```

Alternatively, could set `debugNetworkImageHttpClientProvider` to a fresh _FakeHttpClient instance.

packages/flutter/test/painting/image_provider_network_image_test.dart
```diff
@@ -349,6 +397,9 @@ class _FakeHttpClientRequest extends Fake implements HttpClientRequest {
 class _FakeHttpClientResponse extends Fake implements HttpClientResponse {
   bool drained = false;
 
+  @override
+  final HttpHeaders headers = _FakeHttpHeaders();
```
gnprice: This change isn't needed, is it? The new test adds headers only in the request, not the response.

gnprice (commented):
Thanks @rajveermalviya for the diagnosis and the fix! These changes look good; just nits below on the test.

I also tweaked the description (which becomes the merged commit message) to point to the previous PR as well as the commit, for ease of cross-reference.

rajveermalviya:
Thanks for the review @gnprice! Addressed those comments.

gnprice (approved):
Thanks! These changes LGTM.

BTW for future PRs in this repo, the most convenient thing is to push revisions as added commits on top of the existing ones. (The exception is when there's a need to rebase atop a newer main.)

That's helpful for later reviewers or other readers to be able to follow the thread in the GitHub web interface, because it effectively keeps the full history of your changes visible here. On merge, the changes will get squashed into one commit anyway, so there's no harm to the clarity of the final history in main.

victorsanni (approved):

dkwingsmt (approved):
LGTM. Thank you!

PR #172436, opened 2025-07-20
Fix #160622: Restore watchOS Companion App Build Compatibility

lukemmtt:
This PR is an attempt to fix a regression where Flutter apps with Apple Watch extensions fail to build via `flutter run` since Flutter 3.27.0.

## Background

Flutter projects containing Apple Watch companion apps have been failing with various errors since Flutter 3.27.0:
- `'StateObject' is only available in iOS 14.0 or newer` (watchOS SwiftUI code being compiled with iOS deployment targets)
- `'HKWorkoutSession' is unavailable in iOS` (watchOS-specific APIs being compiled for iOS)
- `Unsupported Swift architecture` (iOS architectures being forced on watchOS targets)
- Missing "Watch companion app found" log during `flutter run`

## Root Cause

The `containsWatchCompanion` method was incorrectly passing an iOS device identifier when querying build settings for watchOS schemes, causing xcodebuild to fail with "Unable to find a destination matching the provided destination specifier". This prevented proper detection of watch companion apps, leading to iOS-specific build settings being incorrectly applied to watchOS targets.

## Solution

This fix removes the `deviceId` parameter when querying build settings for watch schemes, allowing xcodebuild to properly return the watch target's configuration. This solution was identified by @vashworth in https://github.com/flutter/flutter/pull/172436#issuecomment-3140433428—thank you for pinpointing the exact issue!

## Previous Attempt

My initial PR #172436 attempted to restore the conditional `ARCHS` logic that was removed in PR #154645, but that addressed a symptom rather than the root cause. The real issue was that `containsWatchCompanion` was failing to detect the watch app due to the incorrect deviceId parameter.

Fixes #160622

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[...]

PR #176832, opened 2025-10-10
Fix #160622: change containsWatchConpanion function to detect companion watch apps defined by only the project info file.

### LINKED ISSUE #160622: Unable To Run Flutter App With Apple Watch Extension After Upgrading 3.27.0
### Steps to reproduce

My app has an Apple Watch extension and working without any problem before upgrading Flutter 3.27
But after upgrade I cannot run my app from VSCode. But I can run either app itself or Apple Watch extension from XCode without any problem.

I know you will ask a reproducable example but I can only provide what I have for now. I will try to create a new project asap. 

### Expected results

I expect to run without any problem

### Actual results

Build fails on VSCode

### Code sample

```dart
// no code available
```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
Launching lib/main.dart on iPhone 16 Pro in debug mode...
Xcode build done.                                           40.4s
Failed to build iOS app
Swift Compiler Error (Xcode): 'StateObject' is only available in iOS 14.0 or newer
/Users/bahadirarslan/Development/Flutter/myproject/ios/MyProject%20Watch%20App/MainView.swift:10:5

Swift Compiler Error (Xcode): 'main()' is only available in iOS 14.0 or newer
/Users/bahadirarslan/Development/Flutter/myproject/ios/MyProject%20Watch%20App/MyProjectApp.swift:9:0

Swift Compiler Error (Xcode): 'Scene' is only available in iOS 14.0 or newer
/Users/bahadirarslan/Development/Flutter/myproject/ios/MyProject%20Watch%20App/MyProjectApp.swift:12:19

Swift Compiler Error (Xcode): Type 'WatchViewModel' does not conform to protocol 'WCSessionDelegate'
/Users/bahadirarslan/Development/Flutter/myproject/ios/MyProject%20Watch%20App/ViewModel/WatchViewModel.swift:79:0

Uncategorized (Xcode): Command SwiftCompile failed with a nonzero exit code

Could not build the application for the simulator.
Error launching application on iPhone 16 Pro.

Exited (1).
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.27.0, on macOS 15.1 24B83 darwin-arm64, locale en-TR)
    • Flutter version 3.27.0 on channel stable at /Users/bahadirarslan/Development/SDKS/flutter
    • Framework revision 8495dee1fd (9 days ago), 2024-12-10 14:23:39 -0800
    • Engine revision 83bacfc525
    • Dart version 3.6.0
    • DevTools version 2.40.2

    • Android SDK at /Users/bahadirarslan/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • Java version OpenJDK Runtime Environment (build 21.0.3+-79915917-b509.11)

[✓] Xcode - develop for iOS and macOS (Xcode 16.2)
    • Build 16C5032a
    • CocoaPods version 1.15.2

[✓] Android Studio (version 2024.2)
    • Java version OpenJDK Runtime Environment (build 21.0.3+-79915917-b509.11)

[✓] VS Code (version 1.96.0)
    • Flutter extension version 3.102.0

    • iPhone 16 Pro (mobile)          • B5611E87-5F6A-4178-ABAD-7B837493DE55 • ios            • com.apple.CoreSimulator.SimRuntime.iOS-18-2 (simulator)
    • macOS (desktop)                 • macos                                • darwin-arm64   • macOS 15.1 24B83 darwin-arm64
    • Mac Designed for iPad (desktop) • mac-designed-for-ipad                • darwin         • macOS 15.1 24B83 darwin-arm64
    • Chrome (web)                    • chrome                               • web-javascript • Google Chrome 131.0.6778.205
    ! Error: Browsing on the local area network for Bahadır’s iPhone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for Bahadir Arslan’s iPad. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
    ! Error: Browsing on the local area network for Bahadır's iPad. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.

[✓] Network resources

• No issues found!
```

</details>

garrettjavalia:
(Please be generous for some typos and grammar error in sentences below. English is not my mother tongue)

TRDR : This patch fixes flutter run(and might be flutter build too) failure on iOS flutter apps with companion watchOS app in case the WKCompanionAppBundleIdentifier value is defined by xcode project configuration and uses different app bundle id by scheme(which happens when you do something like com.myapp.app1.dev things for scheme bundle id)

This change checks default scheme (the scheme with corresponding name for the debug/release mode and selected build flavor.) in the build settings check step of containsWatchCompanion function.

current code of containsWatchCompanion function works like below

1. check all default Info.plist file's content of all targets to determine if the project has watchOS companion apps.
(this doesn't work well in mordern xcode settings when they use multi plist files which are selected on build time by build configuration. Sometimes, the default Info.plist file doesn't even exist in the project.)

2. check if "WKCompanionAppBundleIdentifier" is included in the xcode project info file(the ios/Runner.xcodeproj/project.pbxproj file in case of iOS project generated by flutter)

3. If "WKCompanionAppBundleIdentifier" has found in the project info file,
check build configuration variables of every single scheme but the current target scheme if they have config value with key "INFOPLIST_KEY_WKCompanionAppBundleIdentifier" identical to current build configuration's bundle identifier returned by productBundleIdentifier@xcode_project.dart function.

I believe The third step causes many problems reported in some issue and pr, saying iOS app with companion watchOS app fails on build or run.
https://github.com/flutter/flutter/issues/160622
https://github.com/flutter/flutter/pull/172436

In my case, My iOS project had multiple schemes which are linked to different build configurations, but using single Runner setting per app.
because we exclude current scheme for the check, we get all the INFOPLIST_KEY_WKCompanionAppBundleIdentifier values except the one which should be same with the identifier returned by productBundleIdentifier@xcode_project.dart function.
that is because companion watch apps doesn't have to have same WKCompanionAppBundleIdentifier with other scheme's main app. We have to check the default scheme to see watchOS companion app`s WKCompanionAppBundleIdentifier, not other scheme. (other schemes can have slightly different or totally different bundle id. that was to support different environment systems like dev, prod in my case)
as a result, flutter run gave me WatchOS app built for device target iOS/iPad, not Watch. which caused problem during installation of iOS app to a simulator because the embeded Watch App was in invalid format. I found this by inspecting Info.plist of generated watchOS app.

below is part of the flutter run verbose log with failure caused by old code.
[+1154 ms] An error was encountered processing the command (domain=IXUserPresentableErrorDomain, code=1):
           App installation failed: ‘Bora Debug-dev’을(를) 설치할 수 없음 (which says "cannot install" in korean)
           나중에 다시 시도하십시오. (which says "try again later" in korean)
           Found WatchKit 2.0 app at /Users/javalia/Library/Developer/CoreSimulator/Devices/071691B3-7901-47E5-9B38-4D5B799F3530/data/Library/Caches/com.apple.mobile.installd.staging/temp.8HZIEG/extracted/Payload/Runner.app/Watch/bora Watch App.app but it does not have a WKWatchKitApp or WKApplication key set to true in its Info.plist
           Underlying error (domain=IXUserPresentableErrorDomain, code=1):
           	‘Bora Debug-dev’을(를) 설치할 수 없음 (which says "cannot install" in korean)
           	나중에 다시 시도하십시오. (which says "try it later" in korean)

I suggest this change with two reasons : 

1. I think forcing watchOS app makers split their runner configurations to support watch companion app is something not recommendable/not good to force as a convention.
(Which is implied in the comment which is removed by this pr, and could be one of the valid detouring.
some people in this issue https://github.com/flutter/flutter/issues/160622 said that a detouring is adding of dummy WKCompanionAppBundleIdentifier configuration in iOS app, which should be in only watchOS apps in normal.)

2. The function should work as it's name implies, so I guess it should not omit default scheme during check. Matching function's behavior with it's name will be likely to reduce potential errors caused by this change while fixing problems.

## Pre-launch Checklist

- [ X ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ X ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ X ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ X ] I signed the [CLA].
- [ X ] I listed at least one issue that this PR fixes in the description above.


garrettjavalia:
I changed the code to check only the scheme of current flavor, so that the slow build config generation is only executed for current flavor. the build config generation is slow as much as real build, so we must remove useless generation.

garrettjavalia:
Removed unrequired loop

LongCatIsLooong:
from triage: @vashworth could you take a look since you also reviewed https://github.com/flutter/flutter/pull/172436?

vashworth:
It appears that this PR checks for `WKCompanionAppBundleIdentifier` within the build settings of the default scheme. However, this does not makes sense as `WKCompanionAppBundleIdentifier` should be set in the companion Watch app's settings, not the iOS app target settings: https://developer.apple.com/documentation/bundleresources/information-property-list/wkcompanionappbundleidentifier?language=objc#Discussion

@garrettjavalia Did this solution not work for you? https://github.com/flutter/flutter/pull/172436

garrettjavalia:
I did not test solution in #172436. My solution was made separately from that, and I see that solution is currently closed.

please let me explain about this more.

first of all, checking default(currently targeting) scheme is a valid operation in my opinion.
schemes can include multiple build targets(the apps) and build configurations to use on each operation(debug run, build, etc.)
Imagine a valid Xcode scheme which includes two targets(a host iOS app and a companion watchOS app.) for archive.(which flutter build command should handle, in real world senarios.)
(we surely can omit watchOS app in the scheme we want to use in flutter run command, but not for the flutter build command.)

the second part, which you might feel this code is wrong is, How it can work if it just processes xcode build configuration without specifying correct watchOS app as target. (buildSettingsForBuildInfo function and it`s call chain can specify a target to test, but current code doesn't specify one.)
by running flutter run -v, we can see that if no target is given, it generates config for all targets in the project at once.
and it surely includes INFOPLIST_KEY_WKCompanionAppBundleIdentifier if the scheme has watchOS app which is correctly configured by xcode project's target build settings.

now we can think that the valid way to handle this problem might be by making it check only watchOS App's build configuration, and yes that is a valid point. but that is for some cases when a scheme has two watchOS app(like one for companion app and one for independent app, or just two companion watchOS apps in wrong configuration.)

with above inspections, I suggest current PR a pretty good way to solve current situation. well, maybe I can improve the code to handle more complicated situations I mentioned above, but it would be optional for now

garrettjavalia:
Following screenshot shows how a scheme can have multiple apps in it's scheme configuration of xcode.
스크린샷 2025-10-25 오후 6 45 32

And this shows how a host iOS app handles it's companion watchOS app during the build phases.
스크린샷 2025-10-25 오후 6 47 15

So in short, Xcode project scheme != Xcode build target(app)

garrettjavalia:
Updated existing tests and added new tests as a separated test group.

garrettjavalia:
found out that 'watch companion' test group's test codes are getting host iOS app bundle name configuration with XcodeProjectBuildContext(scheme: 'Runner')
while it requires XcodeProjectBuildContext(scheme: 'Runner',  deviceId: '123' ) for substituteXcodeVariables calls used to parse PList file.

refactored test codes in following ways.
1. Removed ineffective configurations related to substituteXcodeVariables in test cases those are not interacting with plist xcode variable.
2. unified code that sets bundle id into the 'watch companion' test group, which were divided into sub testgroups in my previous commit.

garrettjavalia:
I found that multi target schemes are Ok in xcode, but not in flutter code.
productBundleIdentifier function in xcode_project.dart is weak to multiple targets so that it can detect wrong bundle identifier.
It was pure luck that my xcode project was working fine in flutter run and flutter build command.
I will think of improvement to this. until then, I guess this PR should be draft

vashworth (approved):

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -796,13 +796,13 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
       projectInfo.reportFlavorNotFoundAndExit();
     }
     for (final String scheme in projectInfo.schemes) {
-      // the default scheme should not be a watch scheme, so skip it
+      // Flutter assumes single build target per scheme, so skip defalt scheme.
```
LouiseHsu: nit: defalt -> default

LouiseHsu (approved):
tiny typo nit, lgtm!

packages/flutter_tools/lib/src/xcode_project.dart
garrettjavalia: Thank you, it's fixed now.

PR #173609, opened 2025-08-12
service worker removal

SydneyBao:
Create a service worker that removes the old cached service worker.  

Related issue: [156910](https://github.com/flutter/flutter/issues/156910#issue-2589764282)
## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #170918, opened 2025-06-20
Service worker removal

salemiranloye:
Creating a service worker to remove the old cached service worker then unregister itself since flutters service worker is not being maintained so now it is being deprecated 

Related Issue: https://github.com/flutter/flutter/issues/156910#issue-2589764282

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

Issue #156910, opened 2024-10-15
[web] The future of `flutter_service_worker.js`


ditman:
### Document Link

https://flutter.dev/go/web-cleanup-service-worker

### What problem are you solving?

The current default service worker implementation is one of (several) approaches that can be used by Flutter web, however it's not a good fit for all use-cases.

This confuses users when it doesn't behave as they expect.

Having a default service worker implementation sends the message that it is *necessary* or *recommended* by Flutter web to function correctly, but that is not the case.

This design doc discusses how the current implementation of `flutter_service_worker.js` is going to be phased out (TL;DR: with a cleanup service worker for those currently using the default one), and how the flutter tools and flutter.js will eventually stop generating (or loading) a default service worker, allowing web app developers to do what's best for their specific apps.

slavap:
@ditman 
>> and flutter.js will eventually stop generating (or loading) a default service worker, allowing web app developers to do what's best for their specific apps.

the best for any web app is to have PWA "just working". If someone for some reason wants "custom PWA", let them do it, but for others 90% of "lazy" developers - the default implementation of "one of (several) approaches" is absolutely fine, just make it "dumb" and stable as AK-47.

uldall:
@slavap I do not believe a PWA requires a service worker now a days. See [https://web.dev/articles/install-criteria#criteria](https://web.dev/articles/install-criteria#criteria).

slavap:
@uldall if that is correct and PWA could work without service worker - then that has to be clearly explained in the document.

ditman:
@slavap thanks for your message, feedback is always appreciated! (you should also be able to comment on the google doc itself if you want to!)

To me, the biggest problem that the current default service worker has is that **people think that it is _mandatory_**, when in reality, it isn't.

For most of Flutter web developers, the standard browser cache will work as expected, and they'll have a better development and deployment experience. (The rest 1% is providing their own service worker implementation anyway to achieve whatever extra goals they have.)

I'm not opposed to generating service workers with Flutter web, but I think we should:

* have a very clear idea of *what the service worker is supposed to do*: What happens when it goes offline? How does it react to updates? How does it cache assets? etc... I don't think we have any of this right now.
* make it an opt-in feature (maybe make users manually register the `flutter_service_worker.js`, instead of doing it automatically)
* (Ideally) be able to use battle-tested technology that already exists (like [Workbox](https://developer.chrome.com/docs/workbox) or any other), and for that...
[...]

PR #108776, opened 2022-08-02
[web] Add onEntrypointLoaded to FlutterLoader.

ditman:
The current promise-based API of flutter.js to inject the entrypoint and report back with an "engine initializer" object does not work well with Flutter web's hot-restart feature, and in fact, reverts to "auto" initialization upon hot restart as a "fix". See: https://github.com/flutter/flutter/issues/108403. This is a problem for incoming changes on flutter, where we might want to pass custom parameters to the `initializeEngine`/`runApp` JS methods.

**`onEntrypointLoaded` event**

This PR adds a new `onEntrypointLoaded` method to the FlutterLoader provided by flutter.js that triggers a new "flutter:entrypoint-loaded" event every time the app (hot)restarts, so the rest of the custom initialization code can run.

The new init sequence of a flutter app that survives hot-restart looks like:

```html
<script>
  window.addEventListener('load', function(ev) {
      // Download main.dart.js
      _flutter.loader.loadEntrypoint({
        onEntrypointLoaded: onEntrypointLoaded,
        serviceWorker: {
          serviceWorkerVersion: serviceWorkerVersion,
        }
      });

      // Once the entrypoint is ready, do things!
      async function onEntrypointLoaded(engineInitializer) {
        const appRunner = await engineInitializer.initializeEngine();
        appRunner.runApp();
      }
  });
</script>
``` 

_This separation also makes quite obvious what things will run once (`loadEntrypoint`) vs what will run every time the engine hot-restarts (`onEntrypointLoaded`)._

**Detangle flutter.js ServiceWorkerLoader from EntrypointLoader**

In a first (small) step towards modularization of the library, this change detangles the old "loadWithServiceWorker" logic into: load service worker, then load the entrypoint.

The old logic basically attempted very hard to load a service worker, but if for any case it'd fail, it'd always "fall back" to the _loadEntrypoint method.

The new logic separates the Service Worker install logic from the Entrypoint loader, so it can be loaded in two steps:

```
loadServiceWorker --(warn of errors, but always continue)--> loadEntrypoint
```

The new structure makes it easier to ES-Modularize the file in the future, or inject custom behavior for any/all of the steps of the bootstrap. (Some extra modularization is still required from the FlutterLoader bits, but this looks like a nice start).

### Issues

* Fixes https://github.com/flutter/flutter/issues/108403

### Tests

* Tested manually on MacOS/Linux/iPhone, here: https://dit-bootstrap-tests.web.app
* Added an integration test to CI, similar to the others used to test flutter.js

[...]

PR #176834, opened 2025-10-10
[web] Self-cleaning service worker

### LINKED ISSUE #106225: [web] Improve dev/bots/service_worker_test
`dev/bots/service_worker_test.dart` used to be a small file that focused on testing if the service worker pre-cache and cache behavior worked as expected.

However in the last few months, more features have been tested using the same infrastructure, and the file is getting a little bit too big and hard to maintain.

A couple of bits that stand out:

* The app that is under test uses different HTML and Dart entrypoints. This is handled by copying around different index.html files, and using -t lib/some_other_main.dart with different logic. Copying the files is ad-hoc code, and would benefit from #78254
* There's a lot of copy-pasta maintaining the "server that logs how many hits each requested file has" to do assertions later, there's at least 3 copies of very similar code in the file (I just had to add the 3rd!), where most of it could be reused.

Code:

* https://github.com/flutter/flutter/blob/master/dev/bots/service_worker_test.dart
* https://github.com/flutter/flutter/tree/master/dev/integration_tests/web (service_worker_* apps under test)

## Proposal

Audit all the different usages of dev/integration_tests/web/lib/*.dart, and see if there are more integration tests that build those apps... and how they do it. Maybe service_worker_test.dart is not the only file that could benefit from this refactor (?)

Tweak dev/bots/service_worker_test.dart to reuse code, start web apps more properly (after #78254), and in general, show it some TLC.

mdebbar:
Introduce a self-cleaning service worker to replace the old one.

Previous attempts (https://github.com/flutter/flutter/pull/170918, https://github.com/flutter/flutter/pull/173609) had (accidentally?) disabled the loading of the service worker in `flutter.js`. This PR preserves the logic for loading the service worker, and it prints a deprecation warning.

Towards https://github.com/flutter/flutter/issues/156910
Closes https://github.com/flutter/flutter/issues/106225


dev/bots/service_worker_test.dart
```diff
@@ -17,820 +17,192 @@ final String _bat = Platform.isWindows ? '.bat' : '';
...
+    expect(
+      _requestedPaths,
+      isNot(contains('main.dart.js')),
+      reason:
+          'On a simple reload, main.dart.js should have been served from the cache, so no network request was expected.',
+    );
+    print('${green}Verification successful: Old caching worker is active.$reset');
+    await server.stop();
+
+    print('\n${yellow}Phase 2: Deploying cleanup worker and verifying cache is removed...$reset');
```
ditman, line 101: Haha, fancy colors in the output, nice :P

dev/bots/service_worker_test.dart
```diff
@@ -17,820 +17,192 @@ final String _bat = Platform.isWindows ? '.bat' : '';
...
     () async {
       int tries = 1;
-      while (!waitForCounts.entries.every(
-        (MapEntry<String, int> entry) => (requestedPathCounts[entry.key] ?? 0) >= entry.value,
-      )) {
-        if (tries++ % 20 == 0) {
-          print('Still waiting. Requested so far: $requestedPathCounts');
+      while (!_requestedPaths.contains(file)) {
+        if (tries++ % 40 == 0) {
+          print('-- Still waiting for app to request "$file". Requested so far: $_requestedPaths');
```
ditman, line 156: We're going to eventually remove this test, so it doesn't matter much, but this smells like `_requestedPaths` could be a `Stream` that we listen to for changes, instead of having this homebrew polling mechanism :/

dev/bots/service_worker_test.dart
```diff
@@ -17,820 +17,192 @@ final String _bat = Platform.isWindows ? '.bat' : '';
...
-
-  // Write out all but the first and last lines with [prefix].
   for (final String line in lines.skip(1).take(lines.length - 2)) {
     buffer.writeln('$prefix$line');
   }
   buffer.write('$prefix${lines.last}');
   return buffer.toString();
 }
-
-Future<void> runWebServiceWorkerTest({
```
ditman, line 280: 🎉

dev/bots/service_worker_test.dart
```diff
@@ -17,820 +17,192 @@ final String _bat = Platform.isWindows ? '.bat' : '';
...
-      )) {
-        if (tries++ % 20 == 0) {
-          print('Still waiting. Requested so far: $requestedPathCounts');
+      while (!_requestedPaths.contains(file)) {
+        if (tries++ % 40 == 0) {
+          print('-- Still waiting for app to request "$file". Requested so far: $_requestedPaths');
         }
         await Future<void>.delayed(const Duration(milliseconds: 100));
       }
+      print('++ App has requested "$file" (requested so far: $_requestedPaths)');
```
ditman: Printing an ever-growing list of `_requestedPaths` may make the test output super verbose. I think logging the `file` is fine, but `_requestedPaths` may be too much?

dev/integration_tests/flutter_gallery/web/index.html
```diff
@@ -19,72 +19,6 @@
...
-            loadMainDartJs();
-          }
-        }, 4000);
-      });
-    } else {
-      // Service workers not supported. Just drop the <script> tag.
-      loadMainDartJs();
-    }
-  </script>
+  <script src="main.dart.js" type="application/javascript"></script>
```
ditman, line 22: Good change! I hope this doesn't affect some random benchmark somewhere else :/

engine/src/flutter/lib/web_ui/flutter_js/src/service_worker_loader.js
```diff
@@ -62,40 +62,55 @@ export class FlutterServiceWorkerLoader {
...
-      if (!window.isSecureContext) {
-        errorMessage += "\nThe current context is NOT secure."
-        errorMessage += "\nRead more: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts";
+      return Promise.resolve();
+    }
+
+    const registerAndActivate = () => {
+      console.warn(
+        'Loading the service worker using Flutter bootstrap is deprecated and will stop working in ' +
+        'a future release. If a service worker is still needed, it can be loaded directly via the ' +
```
ditman: Weren't we going to log this only if `serviceWorkerUrl` is different than the default one (line 82 below). Also, I think this probably should link to the flutter issue (and from there, we can provide the link to the MDN)

People using the default service worker without any configuration probably don't care about this warning.

packages/flutter_tools/lib/src/web/file_generators/js/flutter_service_worker.js
```diff
@@ -1,172 +1,32 @@
...
-        return response || fetch(event.request).then((response) => {
-          if (response && Boolean(response.ok)) {
-            cache.put(event.request, response.clone());
-          }
-          return response;
+
+      try {
+        const clients = await self.clients.matchAll({
+          type: 'window',
+          includeUncontrolled: true,
```
ditman: I think adding `includeUncontrolled` will refresh [more tabs than needed](https://developer.mozilla.org/en-US/docs/Web/API/Clients/matchAll#includeuncontrolled).

I think we should use this cleanup service worker:

https://developer.chrome.com/docs/workbox/remove-buggy-service-workers

(After all, we'd tell people to use `workbox` to maybe generate their custom service worker instead of doing it by hand)

((PS: I don't remember finding that workbox link originally, it's very good!))

packages/flutter_tools/test/web.shard/test_data/hot_reload_index_html_samples.dart
```diff
@@ -30,11 +27,7 @@ String indexHtmlFlutterJsCallback = _generateFlutterJsIndexHtml('''
 String indexHtmlFlutterJsPromisesShort = _generateFlutterJsIndexHtml('''
```
ditman, line 27: IIRC the promise-based initialization has been warning a ["use callbacks" message](https://github.com/flutter/flutter/blob/master/engine/src/flutter/lib/web_ui/flutter_js/src/entrypoint_loader.js#L135) for a few years now, consider removing it!

ditman (approved):
This is great; I only have a small comment in the code of the cleanup service worker about refreshing "too many" tabs and to follow the workbox example. But all my comments are nits, or unrelated to this specific change. LGTM!

dev/bots/service_worker_test.dart
mdebbar: This is from the great work of @SydneyBao and @salemiranloye 😄

dev/bots/service_worker_test.dart
mdebbar: Agreed 😞 The polling mechanism predates the PR and I kinda want to limit the scope of this long overdue PR :)

dev/bots/service_worker_test.dart
mdebbar: These were added mainly for debugging CI failures. We still want _some_ logs but I'll clean it up a bit.

dev/integration_tests/flutter_gallery/web/index.html
mdebbar: 🤞

engine/src/flutter/lib/web_ui/flutter_js/src/service_worker_loader.js
mdebbar: Good catch! Done.

packages/flutter_tools/lib/src/web/file_generators/js/flutter_service_worker.js
mdebbar: Good catch! Done.

packages/flutter_tools/test/web.shard/test_data/hot_reload_index_html_samples.dart
mdebbar: The warning was added because the promise-based API has issues with hot restart (context: https://github.com/flutter/flutter/pull/108776).

I imagine the same issue still exists today, so the warning is still valid.

PR #176843, opened 2025-10-10
Tapping outside of `SelectableRegion` should dismiss the selection

Renzo-Olivares:
This PR updates selectable region so it dismisses the selection when a tap happens outside of it. This is done through the use of `TapRegion` and it's `onTapOutside` callback.

https://github.com/user-attachments/assets/d5678c85-9d04-4c9c-af16-9a09a001c228

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -1940,8 +1940,10 @@ class SelectableRegionState extends State<SelectableRegion>
     }
     return TapRegion(
       groupId: SelectableRegion,
-      onTapOutside: (PointerEvent event) {
+      onTapOutside: (PointerDownEvent event) {
         clearSelection();
+        _selectionStatusNotifier.value = SelectableRegionSelectionStatus.changing;
+        _finalizeSelectableRegionStatus();
       },
       child: CompositedTransformTarget(
         link: _toolbarLayerLink,
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -1941,11 +1941,12 @@ class SelectableRegionState extends State<SelectableRegion>
     return TapRegion(
       groupId: SelectableRegion,
       onTapOutside: (PointerDownEvent event) {
-        // To match native platforms, the selection is dismissed when tapping outside
-        // of the selectable region.
-        clearSelection();
-        _selectionStatusNotifier.value = SelectableRegionSelectionStatus.changing;
-        _finalizeSelectableRegionStatus();
+        // To match the native web behavior, this selectable region is
+        // unfocused when tapping outside of it causing the selection to
+        // be dismissed.
+        if (kIsWeb) {
+          _focusNode.unfocus();
+        }
       },
       child: CompositedTransformTarget(
         link: _toolbarLayerLink,
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -1944,6 +1944,9 @@ class SelectableRegionState extends State<SelectableRegion>
         // To match the native web behavior, this selectable region is
         // unfocused when tapping outside of it causing the selection to
         // be dismissed.
+        //
+        // Tapping outside the selectable region does not unfocus
+        // the region on non-web platforms.
         if (kIsWeb) {
           _focusNode.unfocus();
         }
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -1941,6 +1941,8 @@ class SelectableRegionState extends State<SelectableRegion>
     return TapRegion(
       groupId: SelectableRegion,
       onTapOutside: (PointerDownEvent event) {
+        // To match native platforms, the selection is dismissed when tapping outside
+        // of the selectable region.
         clearSelection();
         _selectionStatusNotifier.value = SelectableRegionSelectionStatus.changing;
         _finalizeSelectableRegionStatus();
--- a/packages/flutter/lib/src/widgets/selectable_region.dart
+++ b/packages/flutter/lib/src/widgets/selectable_region.dart
@@ -32,6 +32,7 @@ import 'media_query.dart';
 import 'overlay.dart';
 import 'platform_selectable_region_context_menu.dart';
 import 'selection_container.dart';
+import 'tap_region.dart';
 import 'text_editing_intents.dart';
 import 'text_selection.dart';
 import 'text_selection_toolbar_anchors.dart';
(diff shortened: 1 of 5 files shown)


packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1777,7 +1778,10 @@ class SelectionOverlay {
         dragStartBehavior: dragStartBehavior,
       );
     }
-    return TextFieldTapRegion(child: ExcludeSemantics(child: handle));
+    return TapRegion(
```
Renzo-Olivares, line 1781: This could be extracted into `SelectableRegionTapRegion`, but not sure if we wanted to do that in this PR. Also instead of wrapping two `TapRegion`s here we could just use one and have the `groupId` be provided.

Renzo-Olivares:
This PR adds the basic default functionality. I plan to do a follow up PR making this overridable similar to `EditableText` though since it introduces more APIs https://github.com/Renzo-Olivares/flutter/commit/0625dfc3d2432112e8d73a2b9175f01e9a2872bb i'd like to do that separately.

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -1937,18 +1938,29 @@ class SelectableRegionState extends State<SelectableRegion>
...
-        child: Actions(
-          actions: _actions,
-          child: Focus.withExternalFocusNode(
-            includeSemantics: false,
-            focusNode: _focusNode,
-            child: result,
+    return TapRegion(
+      groupId: SelectableRegion,
+      onTapOutside: (PointerDownEvent event) {
+        // To match the native web behavior, this selectable region is
```
justinmc, line 1944: Is this behavior the same on mobile and desktop?

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -1937,18 +1938,29 @@ class SelectableRegionState extends State<SelectableRegion>
...
-        excludeFromSemantics: true,
-        child: Actions(
-          actions: _actions,
-          child: Focus.withExternalFocusNode(
-            includeSemantics: false,
-            focusNode: _focusNode,
-            child: result,
+    return TapRegion(
+      groupId: SelectableRegion,
+      onTapOutside: (PointerDownEvent event) {
```
justinmc, line 1943: Is the correct behavior that the text should be unselected when the user taps *anywhere*, including on the selection or on unselected text? I guess those parts are already handled elsewhere in the code?

packages/flutter/test/widgets/selectable_region_test.dart
```diff
@@ -45,6 +45,15 @@ void main() {
     );
   });
 
+  Future<void> setAppLifecycleState(AppLifecycleState state) async {
+    final ByteData? message = const StringCodec().encodeMessage(state.toString());
+    await TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger.handlePlatformMessage(
+      'flutter/lifecycle',
+      message,
+      (_) {},
```
justinmc: Nit: Not important, but maybe replace the `_` with the type and variable name if you get a chance.

packages/flutter/test/widgets/selectable_region_test.dart
```diff
@@ -237,6 +246,56 @@ void main() {
...
+                child: SelectableRegion(
+                  selectionControls: materialTextSelectionControls,
+                  child: const Text(text),
+                ),
+              ),
+            ),
+          ),
+        );
+        // The selection only dismisses when unfocused if the app
+        // was currently active.
```
justinmc, line 266: Good catch.

packages/flutter/test/widgets/selectable_region_test.dart
```diff
@@ -237,6 +246,56 @@ void main() {
...
+        await tester.pump();
+
+        expect(paragraph.selections, isNotEmpty);
+
+        // Tap just outside the top-left corner of the selectable region
+        // to dismiss the selection.
+        final Rect selectableRegionRect = tester.getRect(find.byType(SelectableRegion));
+        await tester.tapAt(selectableRegionRect.topLeft - const Offset(10.0, 10.0));
+        await tester.pump();
+        expect(paragraph.selections, isEmpty);
```
justinmc, line 293: Nit: Reading this test, I'm also curious what would happen if you tapped inside the selectableRegionRect. I think it also dismissed? Is that tested anywhere?

justinmc (approved):
LGTM 👍. I'm glad you ended up finding this clean solution, I was worried it would be a lot more complicated!

Is this one of the things that SelectableRegion needs for parity with SelectableText by the way?

packages/flutter/lib/src/widgets/selectable_region.dart
Renzo-Olivares: Kind of, the expectation is when tapping inside the selection area the gesture recognizers handle this. So for example for a tap inside a selection area the selection would be set to the tapped position.

packages/flutter/lib/src/widgets/selectable_region.dart
Renzo-Olivares: From my testing they behave the same. I also tried native frameworks like jetpack compose and macOS swift ui and this seems like it also matches the behavior. I tried iOS swift UI but it technically doesn't have a widget that let's you granularly select static text, it's Text().textSelection(.enabled) only works for macOS, and on iOS it selects everything (context menu pops up with "copy") and doesn't show selection handles or a highlight. I tried using a `UITextView` but sometimes it dismissed the selection and sometimes it did not, i'm not sure if this is due to the interaction between SwiftUI and UIViews so I haven't changed the existing behavior for native iOS.

packages/flutter/test/widgets/selectable_region_test.dart
Renzo-Olivares: Taps inside the selectable region rect will be handled by the selectable regions gesture recognizers, a basic tap inside the selectable region will move the selection to the precise/nearest position. This should already be covered by the gesture tests in `selectable_region_test.dart`.

packages/flutter/lib/src/widgets/text_selection.dart
chunhtai: Can this share the same group id as textfield so we don't create 2?

chunhtai: talked offline, this will cause textfield and selection area to share the same tap region and not unfocus one when the other receives tap

chunhtai (approved):
LGTM

packages/flutter/lib/src/widgets/text_selection.dart
Renzo-Olivares: Talked offline: I think if we used the same groupId as TextField, wouldn't we run into a situation where if we have a focused TextField and a SelectionArea in a column and we press the SelectionArea, the TextField will not lose focus because it is the same TapRegion as SelectionArea.

packages/flutter/lib/src/widgets/selectable_region.dart
loic-sharma: I checked WinUI 3 on Windows, it also does not dismiss the text selection if you click outside of it (unless your click moves the focus).

In my initial review, I also wondered if this was correct. It might be worth updating the comment to clarify that this is indeed intentional. Maybe something like:

```
Tapping outside of the text selection does not dismiss it on Android, macOS, and Windows.
```

loic-sharma (approved):
LGTM!

PR #176862, opened 2025-10-10
[web] Match the behavior of other platforms in Web Locale.toString if the country code is an empty string

### LINKED ISSUE #176666: [web] Locale is returning "en_" if country code is empty string
### Steps to reproduce

1. Create two locales `one = Locale('en')` and `two = Locale('en', '')`
2. Log both

This is an exact duplicate of https://github.com/flutter/flutter/issues/43607. That issue was fixed for native platforms but not for web platforms. The same fix should be applied to https://github.com/flutter/flutter/blob/master/engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart and https://github.com/flutter/flutter/blob/master/engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart

### Expected results

Both should print as `en`.

### Actual results

See that one is `en` and two is `en_`

### Code sample

```dart

void main() {
  print(Locale('en'));
  print(Locale('en', ''));
}
```

Can be run in dartpad

</details>

### Screenshots or Video

N/A

### Logs

N/A

### Flutter Doctor output

N/A

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/176666

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart
@@ -530,10 +530,10 @@ class Locale {
 
   String _rawToString(String separator) {
     final StringBuffer out = StringBuffer(languageCode);
-    if (scriptCode != null) {
+    if (scriptCode != null && scriptCode!.isNotEmpty) {
       out.write('$separator$scriptCode');
     }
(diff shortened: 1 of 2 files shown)


mdebbar (approved):
Thank you!

PR #176942, opened 2025-10-13
Removes built-in jpeg codec from iOS

b-luk:
Partially addresses #144438

- Removes the RegisterCodecsWithSkia() function in shell.cc. This function is redundant because SkCodec.cpp already automatically registers all codecs that are included based on build rules.
- Updates the build rules so that the jpeg encoder and decoder is not included on iOS.

### DIFF
--- a/engine/src/flutter/shell/common/shell.cc
+++ b/engine/src/flutter/shell/common/shell.cc
@@ -85,20 +85,6 @@ std::unique_ptr<Engine> CreateEngine(
                                   runtime_stage_backend);
 }
 
-void RegisterCodecsWithSkia() {
-  // These are in the order they will be attempted to be decoded from.
-  // If we have data to back it up, we can order these by "frequency used in
-  // the wild" for a very small performance bump, but for now we mirror the
-  // order Skia had them in.
-  SkCodecs::Register(SkPngDecoder::Decoder());
-  SkCodecs::Register(SkJpegDecoder::Decoder());
-  SkCodecs::Register(SkWebpDecoder::Decoder());
-  SkCodecs::Register(SkGifDecoder::Decoder());
-  SkCodecs::Register(SkBmpDecoder::Decoder());
-  SkCodecs::Register(SkWbmpDecoder::Decoder());
-  SkCodecs::Register(SkIcoDecoder::Decoder());
-}
-
 // Though there can be multiple shells, some settings apply to all components in
 // the process. These have to be set up before the shell or any of its
 // sub-components can be initialized. In a perfect world, this would be empty.
@@ -131,7 +117,6 @@ void PerformInitializationTasks(Settings& settings) {
     } else {
       FML_DLOG(INFO) << "Skia deterministic rendering is enabled.";
     }
-    RegisterCodecsWithSkia();
 
     if (settings.icu_initialization_required) {
       if (!settings.icu_data_path.empty()) {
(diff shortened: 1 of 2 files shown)


b-luk:
At c7fc71ec68297a735b60cde861105c643e7de2a8:
```
❯ du -d 0 -k out/ios_release/Flutter.framework
15828	out/ios_release/Flutter.framework
```

At 7765d27a622348e20027abbfb1b0346bf6164742:
```
❯ du -d 0 -k out/ios_release/Flutter.framework
15696	out/ios_release/Flutter.framework
```

Showing a 132 KiB difference.

engine/src/flutter/shell/common/shell.cc
```diff
@@ -85,20 +85,6 @@ std::unique_ptr<Engine> CreateEngine(
...
-  // the wild" for a very small performance bump, but for now we mirror the
-  // order Skia had them in.
-  SkCodecs::Register(SkPngDecoder::Decoder());
-  SkCodecs::Register(SkJpegDecoder::Decoder());
-  SkCodecs::Register(SkWebpDecoder::Decoder());
-  SkCodecs::Register(SkGifDecoder::Decoder());
-  SkCodecs::Register(SkBmpDecoder::Decoder());
-  SkCodecs::Register(SkWbmpDecoder::Decoder());
-  SkCodecs::Register(SkIcoDecoder::Decoder());
-}
```
gaaclarke, line 100: Oh, you know what this was here for.  I'm just reading the comment now.  This is supposed to order decoders in a way so that more frequently used decoders are at the front of the list.  It was never to add them because they weren't there.  In all the years it was there no one ever took advantage of it though so I'm fine removing it.

gaaclarke (approved):
LGTM assuming it gets through the tests.  Thanks for measuring it locally beforehand.  Crazy we didn't do this sooner haha.

engine/src/flutter/tools/gn
```diff
@@ -469,6 +469,9 @@ def to_gn_args(args):
     gn_args['enable_unittests'] = False
 
   # Skia GN args.
+  if args.target_os == 'ios':
```
jmagman: Can this be removed for macOS too? `engine/src/flutter/shell/common/shell.cc` is built into both.

gaaclarke (approved):

engine/src/flutter/tools/gn
b-luk: Thanks for the suggestion. I updated the PR to apply to macOS as well.

b-luk:
Updated to remove the jpeg decoder on macOS as well.

At c7fc71ec68297a735b60cde861105c643e7de2a8:
```
❯ du -k -d 0 out/host_release_arm64/Flutter*.framework
19692	out/host_release_arm64/FlutterEmbedder.framework
20032	out/host_release_arm64/FlutterMacOS.framework
```

At 9fffd54133ded743f953e0ada3f34de074feec32:
```
❯ du -k -d 0 out/host_release_arm64/Flutter*.framework
19528	out/host_release_arm64/FlutterEmbedder.framework
19872	out/host_release_arm64/FlutterMacOS.framework
```

Shows 164 KiB difference for FlutterEmbedder.framework and 160KiB difference for FlutterMacOS.framework.

b-luk:
After discussion with gaaclarke, I'm closing this PR.

I did some performance testing, and it showed that using the fallback iOS/macOS ImageIO decoder has significantly worse performance when creating ImageGenerator objects.

ImageGenerator creation happens on the main thread, and the performance is bad enough that it potentially can cause UI jank.

I will write up my findings in https://github.com/flutter/flutter/issues/144438.

Issue #176605, opened 2025-10-06
SwiftPM macOS - Allow limiting architecture for release builds


abdelaziz-mahdy:
### Use case

I'm developing a Flutter plugin for ExecuTorch, which only provides arm64 binaries for macOS (no x86_64 support). Debug builds work correctly, but release builds fail because Flutter's `build_macos.dart` https://github.com/flutter/flutter/blob/5a6ba3c3550c146880da40b495c8dd1ffdc73636/packages/flutter_tools/lib/src/macos/build_macos.dart#L202  hardcodes the xcodebuild destination to `generic/platform=macOS`, forcing universal binaries.

**Issue:**
- Debug: Uses `-destination platform=macOS,arch=arm64` - Works
- Release: Uses `-destination generic/platform=macOS` - Forces arm64+x86_64 universal binary

**Current code** (build_macos.dart:202):
```dart
final String destination = buildInfo.isDebug
    ? 'platform=${XcodeSdk.MacOSX.displayName},arch=$arch'
    : XcodeSdk.MacOSX.genericPlatform; // Forces universal
```

**Attempted workarounds (all failed):**
- Xcode settings: `ARCHS=arm64`, `EXCLUDED_ARCHS=x86_64`, `ONLY_ACTIVE_ARCH=YES`
- Build flags: `--dart-define=DarwinArchs=arm64` (only affects AOT compilation, not xcodebuild destination)

**Reference:** ExecuTorch only supports arm64 on macOS: https://docs.pytorch.org/executorch/stable/using-executorch-ios.html

### Proposal

Allow developers to control architecture targeting for macOS release builds via:

**Option 1:** Respect `DarwinArchs` define for xcodebuild destination:
```dart
final String destination = buildInfo.isDebug || shouldLimitArchitecture
    ? 'platform=${XcodeSdk.MacOSX.displayName},arch=$arch'
    : XcodeSdk.MacOSX.genericPlatform;
```

**Option 2:** Add build flag:
```bash
flutter build macos --release --arch=arm64
```

**Option 3:** Package.swift configuration support to declare supported architectures.

This would enable plugins with architecture-specific dependencies to build successfully without requiring debug builds.

vashworth:
[...]

PR #176948, opened 2025-10-13
Pass EXCLUDED_ARCHS from Xcode project to xcodebuild for macOS builds

abdelaziz-mahdy:
Passes `EXCLUDED_ARCHS` from Xcode project build settings to the xcodebuild command, fixing macOS builds when
  dependencies don't support all architectures.

  ### Problem
  Release builds for macOS create universal binaries (arm64 + x86_64) by default. This causes build failures when 
  using native dependencies that only provide binaries for a single architecture. While developers can set 
  `EXCLUDED_ARCHS` in their Xcode project's xcconfig files, Swift Package Manager dependencies don't respect this
  project-level setting.
  
  **Fixes**: https://github.com/flutter/flutter/issues/176605

  ## Solution

Instead of adding a CLI flag, pass `EXCLUDED_ARCHS` from the Xcode project's build settings directly to xcodebuild. This ensures Swift Package Manager dependencies respect architecture exclusions set in the project configuration.

**How it works:**
When `macos/Runner/Configs/Release.xcconfig` contains:
```xcconfig
EXCLUDED_ARCHS = x86_64
```

The build command now includes `EXCLUDED_ARCHS=x86_64`, creating an arm64-only binary.

## Usage

**1. Edit your xcconfig file** (e.g., `macos/Runner/Configs/Release.xcconfig`):
```xcconfig
#include "../../Flutter/Flutter-Release.xcconfig"
#include "Warnings.xcconfig"

// Exclude x86_64 for dependencies that only support Apple Silicon
EXCLUDED_ARCHS = x86_64
```

**2. Build normally:**
```bash
flutter build macos --release
```

**Without exclusions (default behavior):**
```bash
# Creates universal binary (arm64 + x86_64)
flutter build macos --release
```

## Implementation

- Modified `buildMacOS()` to retrieve `EXCLUDED_ARCHS` from Xcode project build settings
- Pass `EXCLUDED_ARCHS` to xcodebuild command when set and non-empty
- Removed unused `activeArch` parameter
- Added inline documentation explaining the SPM compatibility fix

## Testing

**Manual verification:**
- ✅ Built ExecuTorch Flutter example with `EXCLUDED_ARCHS = x86_64`: arm64-only binary (518.6MB)
- ✅ Built without exclusions: universal binary (existing behavior)
- ✅ Verified architecture with `lipo -info`: arm64-only when excluded

**Unit tests:**
- ✅ Added test: "Passes EXCLUDED_ARCHS from Xcode project to xcodebuild command"
- ✅ Added test: "Does not pass EXCLUDED_ARCHS when not set in Xcode project"
- ✅ All 25 macOS build tests passing

## Pre-launch Checklist

- [x] I signed the [CLA].


vashworth:
If we can avoid adding any new flags, that would be preferred. Did you try setting the destination to `platform=macOS` to see if that fixed it?

abdelaziz-mahdy:
> If we can avoid adding any new flags, that would be preferred. Did you try setting the destination to `platform=macOS` to see if that fixed it?

  I tested platform=macOS,arch=$arch as suggested. It works, but breaks backward compatibility by forcing all release builds to single-architecture only.

  The problem:
  - Disables universal binary creation by default
  - Users wanting both architectures must build separately on Intel and Apple Silicon, then manually combine with lipo
  - No way to opt-out for projects without architecture constraints

  Why the flag is better:
  - Preserves current behavior (universal binaries by default)
  - Provides explicit opt-in for architecture-limited dependencies: --macos-arch=arm64
  - Clear, documentable solution: "Use --macos-arch if your dependencies don't support both architectures"

  The flag solves the edge case without breaking the common case.

abdelaziz-mahdy:
> > I tested platform=macOS,arch=$arch as suggested.
> 
> Try without an arch, just `platform=macOS`

 I tested just platform=macOS (without arch) as suggested. It still fails with the same error because xcodebuild
  treats platform=macOS as ambiguous and attempts to build for both architectures.

  The warning shows:
  xcodebuild: WARNING: Using the first of multiple matching destinations:
  { platform:macOS, arch:arm64, id:00006020-001820E60A33C01E, name:My Mac }
  { platform:macOS, arch:x86_64, id:00006020-001820E60A33C01E, name:My Mac }

  Despite picking arm64 first, xcodebuild still tries to compile for x86_64, causing the build to fail. The
  destination alone isn't sufficient—we must also specify ARCHS build settings to explicitly control which
  architectures to build.

  The --macos-arch flag approach works because it sets both the destination (platform=macOS,arch=arm64) and the
  build settings (ARCHS=arm64, ONLY_ACTIVE_ARCH=YES), giving explicit control without breaking the default universal
   binary behavior.

vashworth:
> > > I tested platform=macOS,arch=$arch as suggested.
> > 
> > 
> > Try without an arch, just `platform=macOS`
> 
> I tested just platform=macOS (without arch) as suggested. It still fails with the same error because xcodebuild treats platform=macOS as ambiguous and attempts to build for both architectures.
> 
> The warning shows: xcodebuild: WARNING: Using the first of multiple matching destinations: { platform:macOS, arch:arm64, id:00006020-001820E60A33C01E, name:My Mac } { platform:macOS, arch:x86_64, id:00006020-001820E60A33C01E, name:My Mac }
> 
> Despite picking arm64 first, xcodebuild still tries to compile for x86_64, causing the build to fail. The destination alone isn't sufficient—we must also specify ARCHS build settings to explicitly control which architectures to build.
> 
> The --macos-arch flag approach works because it sets both the destination (platform=macOS,arch=arm64) and the build settings (ARCHS=arm64, ONLY_ACTIVE_ARCH=YES), giving explicit control without breaking the default universal binary behavior.

Okay thank you for trying! I see now that it's not that the app binary is still compiling x86_64. It's SwiftPM and its dependencies

packages/flutter_tools/lib/src/macos/build_macos.dart
```diff
@@ -223,6 +237,11 @@ Future<void> buildMacOS({
         'COMPILER_INDEX_STORE_ENABLE=NO',
         if (disabledSandboxEntitlementFile != null)
           'CODE_SIGN_ENTITLEMENTS=${disabledSandboxEntitlementFile.path}',
+        // Add architecture-specific build settings when building for a specific arch
+        if (activeArch != null) ...<String>[
+          'ONLY_ACTIVE_ARCH=${onlyActiveArch ? 'YES' : 'NO'}',
+          'ARCHS=${activeArch.name}',
+        ],
```
vashworth: Okay so I played with this a little bit and I think a cleaner solution would be to pass `EXCLUDED_ARCHS` to the `xcodebuild` command so that developers can exclude the arch through Xcode. It seems like Swift Package Manager doesn't respect the `EXCLUDED_ARCHS` in the project unless it's sent with the command directly.

```diff
+ final String? excludedArches = buildSettings['EXCLUDED_ARCHS'];

  try {
    result = await globals.processUtils.stream(
      <String>[
        '/usr/bin/env',
        'xcrun',
        'xcodebuild',
        '-workspace',
        xcodeWorkspace.path,
        '-configuration',
        configuration,
        '-scheme',
        scheme,
        '-derivedDataPath',
        flutterBuildDir.absolute.path,
        '-destination',
        destination,
        'OBJROOT=${globals.fs.path.join(flutterBuildDir.absolute.path, 'Build', 'Intermediates.noindex')}',
        'SYMROOT=${globals.fs.path.join(flutterBuildDir.absolute.path, 'Build', 'Products')}',
        if (verboseLogging) 'VERBOSE_SCRIPT_LOGGING=YES' else '-quiet',
        'COMPILER_INDEX_STORE_ENABLE=NO',
        if (disabledSandboxEntitlementFile != null)
          'CODE_SIGN_ENTITLEMENTS=${disabledSandboxEntitlementFile.path}',
+        if (excludedArches != null && excludedArches.trim().isNotEmpty)
+          'EXCLUDED_ARCHS=$excludedArches',
        ...environmentVariablesAsXcodeBuildSettings(globals.platform),
      ],
```

What do you think?

abdelaziz-mahdy: I can give it a try, but why this instead of the command line arg? isnt that going to be a commandline arg too?

vashworth: > I can give it a try, but why this instead of the command line arg? isnt that going to be a commandline arg too?

No it wouldn't be a command line arg, the `buildSettings['EXCLUDED_ARCHS']` would come from the Xcode project's build settings. No flag would be passed into the CLI

abdelaziz-mahdy: done, fixes my issue, updated PR title and description.

vashworth (approved):
LGTM! Thank you for the changes!

bkonyi (approved):
LGTM

abdelaziz-mahdy:
I kept trying to find what was the issue and couldn't, can you show me how to figure it so I know in the future, and will fix it as soon I have access to the computer

vashworth:
> I kept trying to find what was the issue and couldn't, can you show me how to figure it so I know in the future, and will fix it as soon I have access to the computer

1. First click on the failing test, [`Linux analyze`](https://github.com/flutter/flutter/pull/176948/checks?check_run_id=55843987655) in this case

2. Click on "View more details on flutter-dashboard" at the bottom
Screenshot 2025-11-21 at 2 26 30 PM

3. Select the `stdout` of the step that failed

Screenshot 2025-11-21 at 2 28 35 PM

4. Scan the logs for failures

In this case, it says you need to run `dart format packages/flutter_tools/test/commands.shard/hermetic/build_macos_test.dart`

abdelaziz-mahdy:
> > I kept trying to find what was the issue and couldn't, can you show me how to figure it so I know in the future, and will fix it as soon I have access to the computer
> 
> 1. First click on the failing test, [`Linux analyze`](https://github.com/flutter/flutter/pull/176948/checks?check_run_id=55843987655) in this case
> 2. Click on "View more details on flutter-dashboard" at the bottom
> 
> Screenshot 2025-11-21 at 2 26 30 PM
> 3. Select the `stdout` of the step that failed
> 
> Screenshot 2025-11-21 at 2 28 35 PM
> 4. Scan the logs for failures
> 
> In this case, it says you need to run `dart format packages/flutter_tools/test/commands.shard/hermetic/build_macos_test.dart`

thank you, reformatted, now I can figure the ci/cd easily

PR #176974, opened 2025-10-13
Implements engine-side declarative pointer event handling for semantics.

### LINKED ISSUE #149001: [Flutter Web] dialogs dismissed prematurely with ensureSemantics
### Steps to reproduce

1. Run the code sample as a Web project in Chrome. 
2. Click on the "Show Dialog" button. 
3. Click anywhere inside the dialog that is not a form field. 
4. Observe the dialog being dismissed.

### Expected results

The dialog should not dismiss when any content inside it is clicked. The dialog should only dismiss if the area outside the content (the barrier) is clicked. It works as expected in MacOS builds.

### Actual results

Clicking on whitespace inside the dialog causes it to be dismissed in Web builds. This only occurs in Web builds and only when `WidgetsFlutterBinding.ensureInitialized().ensureSemantics();` is included.

### Code sample

```dart

void main() {
  runApp(const MyApp());
  WidgetsFlutterBinding.ensureInitialized().ensureSemantics();
}

  const MyApp({super.key});

  @override
    return const MaterialApp(
      title: 'Semantics Issue',
      home: MyHomePage(),
    );
  }
}

  @override
}

  void _onPressed() {
    showDialog(
      context: context,
      barrierColor: Colors.grey.shade400,
      builder: (context) {
        return _dialogContent;
      },
    );
  }

  @override
    return Scaffold(
      body: Center(
        child: Column(
          children: <Widget>[
            FilledButton(
              onPressed: _onPressed,
              child: const Text('Show Dialog'),
            )
          ],
        ),
      ),
    );
  }

  Widget get _dialogContent => Center(
        child: Container(
          margin: const EdgeInsets.all(50),
          child: const Material(
            child: SizedBox(
              height: 250,
              width: 250,
              child: Padding(
                padding: EdgeInsets.all(20),
                child: Column(
                  children: [
                    TextField(
                      decoration: InputDecoration(
                        border: OutlineInputBorder(),
                        labelText: 'Form field one',
                      ),
                    ),
                    SizedBox(height: 40),
                    TextField(
                      decoration: InputDecoration(
                        border: OutlineInputBorder(),
                        labelText: 'Form field two',
                      ),
                    ),
                  ],
                ),
              ),
            ),
          ),
        ),
      );
    }
```

</details>

### Screenshots or Video

<details open>

https://github.com/flutter/flutter/assets/1129265/fa9f2bce-0346-441e-ace7-905532ec23f2

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.19.2, on macOS 14.2.1 23C71 darwin-arm64, locale en-US)
    • Flutter version 3.19.2 on channel stable at /Users/rona/Sites/Tools/flutter
    • Framework revision 7482962148 (3 months ago), 2024-02-27 16:51:22 -0500
    • Engine revision 04817c99c9
    • Dart version 3.3.0
    • DevTools version 2.31.1

[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).


[!] Xcode - develop for iOS and macOS (Xcode 15.3)
    • Build 15E204a
    ✗ Unable to get list of installed Simulator runtimes.
    • CocoaPods version 1.15.2

[!] Android Studio (not installed)
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).

[✓] VS Code (version 1.89.1)
    • Flutter extension version 3.88.0

    • macOS (desktop) • macos  • darwin-arm64   • macOS 14.2.1 23C71 darwin-arm64
    • Chrome (web)    • chrome • web-javascript • Google Chrome 125.0.6422.76

[✓] Network resources

! Doctor found issues in 3 categories.
```

</details>

flutter-zl:
Implements engine-side declarative pointer event handling for semantics.

Framework-side to be implemented in next PR.

Fixes flutter/flutter#149001.

**Before change**
https://dialog-dismiss-before.web.app/

Click on the "Show Dialog" button.
Click anywhere inside the dialog that is not a form field.
Observe the dialog being dismissed. 

**After change**
https://dialog-dimiss-after.web.app/

Click on the "Show Dialog" button.
Click anywhere inside the dialog that is not a form field.
Observe the dialog not dismissed.


engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -590,11 +590,23 @@ abstract class SemanticRole {
...
+    // Only accept pointer events for leaf nodes with interactive semantics.
+    // This prevents non-interactive leaf nodes (like empty containers in dialogs)
+    // from intercepting pointer events meant for underlying widgets like modal
+    // barriers.
+    return semanticsObject.isTappable ||
+        semanticsObject.isButton ||
+        semanticsObject.flags.isTextField ||
+        semanticsObject.flags.isLink ||
+        semanticsObject.flags.isSlider ||
+        semanticsObject.isIncrementable;
```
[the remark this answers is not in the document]
flutter-zl: Great feedback. Updated code.

engine/src/flutter/lib/web_ui/test/engine/semantics/semantics_test.dart
```diff
@@ -1504,6 +1504,234 @@ void _testContainer() {
...
+      '#${kFlutterSemanticNodePrefix}0',
+    )!;
+    expect(
+      element.style.pointerEvents,
+      'none',
+      reason: 'Non-interactive leaf nodes should not intercept pointer events',
+    );
+
+    semantics().semanticsEnabled = false;
+  });
```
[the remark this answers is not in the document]
flutter-zl: Great feedback. Updated code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -590,11 +590,24 @@ abstract class SemanticRole {
...
+    // This prevents non-interactive leaf nodes (like empty containers in dialogs)
+    // from intercepting pointer events meant for underlying widgets like modal
+    // barriers.
+    return semanticsObject.isTappable ||
+        semanticsObject.isButton ||
+        semanticsObject.isCheckable ||
+        semanticsObject.flags.isTextField ||
+        semanticsObject.flags.isLink ||
+        semanticsObject.flags.isSlider ||
+        semanticsObject.isIncrementable;
```
mdebbar: All of these have their own Semantics role or behavior (except `isSlider` [1]). Let's override the `acceptsPointerEvents` in the corresponding classes.

[1] It looks like semantic sliders are not implemented? So let's not worry about it now.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -590,11 +590,24 @@ abstract class SemanticRole {
...
     }
-    // Ignore pointer events on all container nodes.
+    // Ignore pointer events on all container nodes, unless they define a route scope.
+    // Route-scoped containers (like dialogs) need to accept pointer events to prevent
+    // clicks from escaping to the underlying barrier.
     if (semanticsObject.hasChildren) {
-      return false;
-    }
-    return true;
+      return semanticsObject.scopesRoute;
```
chunhtai: This only solve this particular case. I thought the hittest should block sibling when there are multiple in hittest order.

This is what we do in framework
https://github.com/flutter/flutter/blob/c63c6b08a7114517575bcf3e45275ec7a9f4e87f/packages/flutter/lib/src/rendering/box.dart#L3336

chunhtai: Ideally this should always absorb hittest if position is in bound and no children wants to handle it

mdebbar: When `acceptsPointerEvents` returns false, the element becomes transparent for hit testing purposes. When it returns true, the element will be opaque for hit testing purposes.

This PR makes it so that `scopesRoute` elements behave as opaque for hit testing.

See:
https://github.com/flutter/flutter/blob/c63c6b08a7114517575bcf3e45275ec7a9f4e87f/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart#L1798-L1802

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Implemented engine-side declarative pointer event handling for semantics based on discussion. Please review.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Implemented engine-side declarative pointer event handling for semantics based on discussion. Please review.

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1156,6 +1156,34 @@ enum Tristate {
   }
 }
 
+/// Describes how a semantic node should behave during hit testing.
+///
+/// This enum is used by the web engine to determine whether semantic elements
```
chunhtai: This will likely to be used by other platform as well, we should make the language platform neutral
https://github.com/flutter/flutter/issues/177025
https://github.com/flutter/flutter/issues/177024

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1156,6 +1156,34 @@ enum Tristate {
...
+/// should accept or pass through pointer events via the CSS `pointer-events`
+/// property.
+///
+/// See also:
+///  * [SemanticsUpdateBuilder.updateNode], which accepts this enum.
+enum SemanticsHitTestBehavior {
+  /// The semantic element is opaque to hit testing, consuming any pointer
+  /// events within its bounds and preventing them from reaching elements
+  /// behind it.
```
chunhtai: is this blocking child or sibling in hittest order? or both? should document this more clearly

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1792,12 +1820,20 @@ abstract class SemanticsUpdateBuilder {
   /// not use this argument should use other ways to communicate validation
   /// errors to the user, such as embedding validation error text in the label.
   ///
+  /// The `hitTestBehavior` describes how this node should behave during hit
+  /// testing. If null, the platform will infer appropriate behavior based on
+  /// other semantic properties. This is primarily used by the web platform to
```
chunhtai: same for the comment here, should be platform neutral

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1792,12 +1820,20 @@ abstract class SemanticsUpdateBuilder {
   /// not use this argument should use other ways to communicate validation
   /// errors to the user, such as embedding validation error text in the label.
   ///
+  /// The `hitTestBehavior` describes how this node should behave during hit
+  /// testing. If null, the platform will infer appropriate behavior based on
```
chunhtai: instead of using null, consider give it a enum `SemanticsHitTestBehavior.defer` and makes it the default

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1913,6 +1949,7 @@ base class _NativeSemanticsUpdateBuilder extends NativeFieldWrapperClass1
     SemanticsRole role = SemanticsRole.none,
     required List<String>? controlsNodes,
     SemanticsValidationResult validationResult = SemanticsValidationResult.none,
+    SemanticsHitTestBehavior? hitTestBehavior,
```
chunhtai: should remove nullable and make required

engine/src/flutter/lib/ui/semantics.dart
chunhtai: what's the infer behavior? do they inherited from the parent?

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1156,6 +1156,34 @@ enum Tristate {
...
+  ///
+  /// On the web, this results in `pointer-events: all` CSS property.
+  opaque,
+
+  /// The semantic element is transparent to hit testing, allowing pointer
+  /// events to pass through to elements behind it.
+  ///
+  /// This is the default behavior for non-interactive semantic nodes.
+  ///
+  /// On the web, this results in `pointer-events: none` CSS property.
```
chunhtai: In framework transparent means this node receives hittest, but won't block the silbing and child behind it from receiving hittest

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,36 @@ abstract class SemanticRole {
   ///
   /// This boolean decides whether to set the `pointer-events` CSS property to
   /// `all` or to `none` on the semantics [element].
+  ///
+  /// The decision follows a multi-tier policy:
+  ///
+  /// **Tier 1: Explicit Behaviors** - Behaviors attached to this role can
```
chunhtai: Declaration should probably come first before this. Also I think you meant `Implicit behaviors`? since the behavior is inferred from semantics behavior/role

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,36 @@ abstract class SemanticRole {
...
+  /// override pointer event handling (e.g., [Tappable], [SemanticTextField]).
+  /// These take highest precedence as they represent interactive elements.
+  ///
+  /// **Tier 2: Framework Declaration** - When the framework provides explicit
+  /// [ui.SemanticsHitTestBehavior], it will be respected here. This allows
+  /// declarative control over pointer event handling from the framework layer.
+  ///
+  /// **Tier 3: Route-Scoped Containers** - Containers with [scopesRoute] set
```
chunhtai: This should be handled in framework by wrapping ModalRoute with SemanticsHitTestBehavior

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,36 @@ abstract class SemanticRole {
...
+  /// (like dialogs, bottom sheets) accept pointer events to act as barriers,
+  /// preventing clicks from escaping to underlying modal barriers. This is
+  /// the current fix for issue #149001.
+  ///
+  /// **Tier 4: Default Transparent** - Leaf nodes without interactive behaviors
```
chunhtai: we could also handle this on the framework side by inferring SemanticsHitTestBehavior if not provided

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
chunhtai: also ideally we should infer the `SemanticsHitTestBehavior` from the framework side if not provided. That way the platform just need to deal with SemanticsHitTestBehavior

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: You're right that framework should handle this. Our architecture fully supports that. When framework provides explicit values, engine just executes them (Tier 1). Engine inference is only a backward-compatible fallback (Tier 2) because popular packages like awesome_dialog don't know about SemanticsHitTestBehavior yet. Engine inference means they work immediately.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: You're right that framework should handle this. Our architecture fully supports that. When framework provides explicit values, engine just executes them (Tier 1). Engine inference is only a backward-compatible fallback (Tier 2) because popular packages like awesome_dialog don't know about SemanticsHitTestBehavior yet. Engine inference means they work immediately.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Thanks for reviewing. I updated the PR accordingly.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
mdebbar: > Declaration should probably come first before this.

This part of the comment was not addressed. When the framework says the node is transparent, we should treat it as transparent. It becomes really confusing when users set a button to transparent and notice that the button is still receiving clicks and blocking hit testing.

If we think it's a bad idea to allow interactive nodes to be transparent, then we should make that state illegal at the framework level. IMO the engine should just follow what the framework dictates.

engine/src/flutter/lib/ui/semantics.dart
mdebbar: > In framework transparent means this node receives hittest, but won't block the silbing and child behind it from receiving hittest

I think you meant "translucent"?
https://github.com/flutter/flutter/blob/542705c5e286153707edd7c0cd2507aa6362079d/packages/flutter/lib/src/rendering/proxy_box.dart#L144-L158

"transparent" is defined in platform view behaviors:
https://github.com/flutter/flutter/blob/542705c5e286153707edd7c0cd2507aa6362079d/packages/flutter/lib/src/rendering/platform_view.dart#L19-L32

- "opaque" matches `pointer-events: all` on the web.
- "transparent" matches `pointer-events: none` on the web.
- "translucent" (which is what you're describing) is not possible on web AFAICT.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Good catch. I thought it is about code style. I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Good catch.  I updated the PR accordingly.

engine/src/flutter/lib/ui/semantics.dart
```diff
@@ -1156,6 +1156,59 @@ enum Tristate {
...
+  ///
+  /// Note: This differs from the framework's `HitTestBehavior.translucent`,
+  /// which receives events while also allowing pass-through. Web's binary
+  /// `pointer-events` property (all or none) cannot support true translucent
+  /// behavior.
+  ///
+  /// Platform implementations:
+  ///  * On the web, this results in `pointer-events: none` CSS property.
+  ///
+  /// This is the default behavior for non-interactive semantic nodes when the
```
chunhtai: this paragraph should probably be in `defer` enum instead, and also mention this is web's default behavior if we want to include this info. (we don't know what other platform will do yet)

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
```diff
@@ -71,6 +71,9 @@ class SemanticIncrementable extends SemanticRole {
     _focusManager.manage(semanticsObject.id, _element);
   }
 
+  @override
+  bool get acceptsPointerEvents => true;
```
chunhtai: This is unfortunate name collisions, should make it something like shouldAcceptsPointerEvents.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,72 @@ abstract class SemanticRole {
...
+  /// makes an interactive element non-clickable. The framework layer is
+  /// responsible for ensuring valid semantic configurations.
+  ///
+  /// **Tier 2: Engine Inference** - When [ui.SemanticsHitTestBehavior.defer] is
+  /// set (the default), the engine infers the appropriate behavior based on:
+  /// - Interactive behaviors (buttons, text fields, sliders, etc.)
+  /// - Route-scoped containers (dialogs, bottom sheets)
+  /// - Default transparent for non-interactive leaf nodes
+  ///
+  /// This approach allows framework full control when specified, with intelligent
```
chunhtai:
```suggestion
  /// This approach allows framework full control when specified, with reasonable
```

chunhtai: should tune down the tone a bit

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,72 @@ abstract class SemanticRole {
...
+        // Used by modal surfaces like dialogs, bottom sheets, drawers.
+        return true;
+      case ui.SemanticsHitTestBehavior.transparent:
+        // Pass through pointer events to elements behind.
+        // Used for non-interactive decorative elements.
+        return false;
+      case ui.SemanticsHitTestBehavior.defer:
+        // This case should never be reached because defer is handled in
+        // acceptsPointerEvents before calling this method.
+        assert(false, 'defer should be handled by _inferAcceptsPointerEvents');
```
chunhtai: consider inline this switch case directly in acceptsPointerEvents body so that we don't need this assert

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,72 @@ abstract class SemanticRole {
...
+        return false; // Unreachable in debug; defensive in release
+    }
+  }
+
+  /// Infers whether pointer events should be accepted based on semantic properties.
+  ///
+  /// This method is called when [ui.SemanticsHitTestBehavior.defer] is set,
```
chunhtai: consider remove this paragraph and use assert(hitTestBehavior == ui.SemanticsHitTestBehavior.defer) to explain the assumption

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -590,11 +659,19 @@ abstract class SemanticRole {
...
     }
-    // Ignore pointer events on all container nodes.
+
+    // Route-scoped containers accept pointer events.
+    // This fixes the dialog dismissal bug (issue #149001) by ensuring that
+    // dialog containers act as barriers, preventing clicks on empty space
+    // from passing through to the modal barrier underneath.
     if (semanticsObject.hasChildren) {
-      return false;
+      return semanticsObject.scopesRoute;
```
chunhtai: this should be set from the framework side instead of using inferring here. The reason being that we don't want every platform has to duplicate this logic. ideally, platform inferring should be as less as possible.

engine/src/flutter/lib/ui/semantics.dart
flutter-zl: Great suggestion. I updated the documentation.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
flutter-zl: acceptsPointerEvents already exists in abstract class SemanticBehavior and abstract class SemanticBehavior. Please check https://github.com/flutter/flutter/blob/master/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart. do we want to change one of them to shouldAcceptsPointerEvents?

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Great suggestion. I updated the documentation.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Great suggestion. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: The intention is to let packages like https://pub.dev/packages/awesome_dialog which has not implemented our change to have the fix.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Great suggestion. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
chunhtai: In general, it will be their responsibility to opt in, but in this case as long as this is wrapped in https://github.com/flutter/flutter/blob/27d070b22463576a3118a328d5ac6c9f8ef0d9df/packages/flutter/lib/src/widgets/routes.dart#L2330

All the route/dialog should get this automaticially

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
chunhtai: we should rename this getter for every classes, it shouldn't be a problem since these are not public

flutter-zl: Thanks for the explanation. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Thanks for the explanation. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
```diff
@@ -71,6 +71,9 @@ class SemanticIncrementable extends SemanticRole {
     _focusManager.manage(semanticsObject.id, _element);
   }
 
+  @override
+  bool get shouldAcceptPointerEvents => true;
```
chunhtai: why does this hard code to true? does the refactor change the current behavior?

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1795,7 +1863,7 @@ class SemanticsObject {
     // Apply updates to the DOM.
     _updateRole();
 
-    if (semanticRole!.acceptsPointerEvents) {
+    if (semanticRole!.shouldAcceptPointerEvents) {
```
chunhtai: I think I misread part of the code in previous pr, I meant we should rename the semanticsBehavior.acceptsPointerEvents to semanticsBehavior.shouldAcceptPointerEvents

since based on the code, the `SemanticsBehavior` is more of a suggestion on whether to handle pointer, it is SemanticRole that makes the final decision

chunhtai: sorry about the confusion

chunhtai (commented):
Just some minor comments and question, otherwise LGTM

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Thanks for clarification. I updated code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
flutter-zl: I added a test('incrementable leaf nodes accept pointer events') case. Without it, Slider/text field would have pointer-events: none.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -581,7 +585,58 @@ abstract class SemanticRole {
...
+          // Used for non-interactive decorative elements.
+          return false;
+        case ui.SemanticsHitTestBehavior.defer:
+          break;
+      }
+    }
+
+    // TIER 2: Engine Inference
+    // When framework defers to engine, infer based on semantic properties.
+    return _inferAcceptsPointerEvents();
```
chunhtai:
```suggestion
      switch (hitTestBehavior) {
        case ui.SemanticsHitTestBehavior.opaque:
          // Absorb pointer events, blocking them from reaching elements behind.
          // Used by modal surfaces like dialogs, bottom sheets, drawers.
          return true;
        case ui.SemanticsHitTestBehavior.transparent:
          // Pass through pointer events to elements behind.
          // Used for non-interactive decorative elements.
          return false;
        case ui.SemanticsHitTestBehavior.defer:
          return _inferAcceptsPointerEvents();
      }
```

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
chunhtai: gotcha, this should at least check hittestbehavior first
```
return switch(hitTestBehavior) {
 HitTestBehavior.opaque => true,
 HitTestBehavior.transparent => false,
 _ => true,

}
```

flutter-zl: Great suggestion. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
flutter-zl: Great suggestion. I updated the code.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
```diff
@@ -71,6 +71,15 @@ class SemanticIncrementable extends SemanticRole {
     _focusManager.manage(semanticsObject.id, _element);
   }
 
+  @override
+  bool get acceptsPointerEvents {
+    return switch (semanticsObject.hitTestBehavior) {
+      ui.SemanticsHitTestBehavior.opaque => true,
+      ui.SemanticsHitTestBehavior.transparent => false,
+      _ => true,
```
chunhtai:
```suggestion
     ui.SemanticsHitTestBehavior.transparent => false,
      _ => true,
```

chunhtai: missed this

chunhtai (approved):
LGTM

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/incrementable.dart
flutter-zl: good catch. Updated.

PR #177031, opened 2025-10-14
Correct basque time format

justinmc:
Fixes https://github.com/flutter/flutter/issues/176677


justinmc:
cupertino_eu.arb and widgets_eu.arb don't have this key.

packages/flutter_localizations/lib/src/l10n/material_eu.arb
```diff
@@ -1,6 +1,6 @@
 {
   "scriptCategory": "English-like",
-  "timeOfDayFormat": "H:mm",
+  "timeOfDayFormat": "HH:mm",
```
QuncCccccc, line 3: We should also update generated_material_localizations.dart file by running `dart dev/tools/localization/bin/gen_localizations.dart --overwrite`.

QuncCccccc (commented):
Thanks for the fix!

This would be another issue that the localization team should help to fix in g3, otherwise next time when we do regular back-integration from g3, this might be reverted to H:mm again. To avoid it accidentally being reverted, can we add a unit test similar to [this](https://github.com/flutter/flutter/blob/e45c6c194fc055c2063d0123a863f008b9348c36/packages/flutter_localizations/test/material/translations_test.dart#L601)?

justinmc:
Thank you for the guidance @QuncCccccc! Ready for re-review.

packages/flutter_localizations/test/material/translations_test.dart
```diff
@@ -607,4 +607,15 @@ void main() {
     expect(localizations, isA<MaterialLocalizationIt>());
     expect(localizations.dateHelpText, 'gg/mm/aaaa');
   });
+
+  // Regression test for https://github.com/flutter/flutter/issues/176677
+  testWidgets('Basque translation for timeOfDayFormat', (WidgetTester tester) async {
+    const Locale locale = Locale('eu');
+    expect(GlobalCupertinoLocalizations.delegate.isSupported(locale), isTrue);
```
QuncCccccc: Oh I think some of the tests above might just copy paste tests from cupertino localizations and we didn't notice it! 😂

```suggestion
    expect(GlobalMaterialLocalizations.delegate.isSupported(locale), isTrue);
```

QuncCccccc (approved):
LGTM with just one minor comment above! Thanks for the fix!

packages/flutter_localizations/test/material/translations_test.dart
justinmc: Good catch! I'll go ahead and fix the other ones too.

Issue #59602, opened 2020-06-16
Create a manifest for template (.tmpl) files and use it as the source of truth


jonahwilliams:
Currently the flutter tool uses whatever is on disk to re-create templates. Unfortunately, this can cause issues if the tool logic gets out of sync with the tmp files.

Namely, this has happened because a tmpl file was removed which matched a gitignore filter, thus causing it to be persisted in the git repo.

This should also be usable by IDEs to figure out what templates and files are supported, but the initial version might not have everything.

jonahwilliams:
This would fix https://github.com/flutter/flutter/issues/57985 and prevent it from happening in the future

jonahwilliams:
Fixed via https://github.com/flutter/flutter/pull/59706 , there is a more specific issue for IDE wizard use case

Issue #57985, opened 2020-05-26
flutter create failed - "Value was missing for variable tag: androidFlutterJar"


johannes545:
## Steps to Reproduce

1. ...so when im trying to start a flutter project i run into the problem that says i dont have androidflutterjar
2. ...
3. ...

## Logs

```
```

```
![image](https://user-images.githubusercontent.com/65943941/82872802-d42ba500-9f33-11ea-8f22-f38511ac5cdc.png)

```
![problem](https://user-images.githubusercontent.com/65943941/82872885-f45b6400-9f33-11ea-8d1d-f17de0f33da3.JPG)

mrhmamun:
Probably it's solved here https://stackoverflow.com/questions/52946189/flutter-error-on-android-could-not-find-aapt2-proto-jar

TahaTesser:
Hi @johannes545 
can you please provide your `flutter doctor -v` and `flutter run --verbose`?
Also, to better address the issue, would be helpful if you could post a minimal code sample to reproduce the problem
Thank you

johannes545:
hey, so im very new to programing, only done basic apps in android studio with less than 100 lines of code.

so i dont know exactly how to adress it better. but il post the flutter doctor -v and the flutter run --verbose
![flutter doctor -v](https://user-images.githubusercontent.com/65943941/82981939-6643b400-9fed-11ea-9b9c-2238d0118f0a.JPG)
![verbose1](https://user-images.githubusercontent.com/65943941/82981952-6d6ac200-9fed-11ea-805a-7533e20562f7.JPG)
![verbose2](https://user-images.githubusercontent.com/65943941/82981954-7065b280-9fed-11ea-8fc3-69c517477756.JPG)
![verbose3](https://user-images.githubusercontent.com/65943941/82981970-75c2fd00-9fed-11ea-8d31-1279ae95ff71.JPG)
![verbose4](https://user-images.githubusercontent.com/65943941/82981973-7bb8de00-9fed-11ea-88ef-b43ae2386f19.JPG)

TahaTesser:
Hi @johannes545 
Can you please execute `flutter create --verbose testapp`  from your desktop folder and copy/paste all the logs, don't create project in user folder?
Thank you

johannes545:
not sure if its from my desktop folder. i have my flutter located at C:\src 
my android studio is also located there
[...]

PR #177034, opened 2025-10-15
Remove generated file from template manifest

robert-ancell:
I'm not sure what this manifest is being used for but none of the other templates also list the file that is generated from them, seems to have been a typo?


loic-sharma (approved):

loic-sharma:
@bkonyi as the flutter tool tech lead would be the right reviewer, but this LGTM :)

Some interesting background reading for templates manifests:

1. https://github.com/flutter/flutter/issues/59602
2. https://github.com/flutter/flutter/issues/57985#issuecomment-645043135

The template manifest seems to be an allowlist of files that can be used from the templates, in case the templates incorrectly have file.

bkonyi (approved):

PR #177059, opened 2025-10-15
Roll Abseil to Chromium's 5b92b04a2e (based on Abseil commit fc4481e968)

jason-simmons:
This also moves the Abseil sources from //flutter/third_party to //third_party to allow use of Chromium's GN scripts without modification.

See internal issue b/400498305


engine/src/build_overrides/angle.gni
```diff
@@ -18,7 +18,7 @@ if (!is_fuchsia) {
 }
 
 # Overrides for ANGLE's dependencies.
-angle_abseil_cpp_dir = "//flutter/third_party/abseil-cpp"
+angle_abseil_cpp_dir = "//third_party/abseil-cpp"
```
jtmcdole, line 21: Remind me - is this the top of flutter/flutter or is it the engine/src/flutter folder?

If we start storing files used to build the engine in root third_party, we need to update the content hash script.

jtmcdole (changes requested):

engine/src/build_overrides/angle.gni
jason-simmons: With this change, the Abseil sources will be placed in `engine/src/third_party/abseil-cpp`

jtmcdole: Late to follow up: that sounds good.  As long as it is under `engine/**` it will get included in the hash.

jtmcdole (approved):

reidbaker (approved):
RSLGTM

PR #177067, opened 2025-10-15
Fix create_updated_flutter_deps.py

mraleph:
1. Make sure that perfetto dependency is in Dart section. If it is in the wrong section the script duplicates it.
2. Fix the logic in the script which produced incorrectly quoted strings.

I am not entirely sure who still uses this script. Dart monorepo has a custom copy, which I am going to fix as well.

### DIFF
--- a/engine/src/tools/dart/create_updated_flutter_deps.py
+++ b/engine/src/tools/dart/create_updated_flutter_deps.py
@@ -115,6 +115,13 @@ def Main(argv):
                 updated_value = dart_v.replace(new_vars["dart_git"], "Var('dart_git') + '/")
                 updated_value = updated_value.replace(old_vars["chromium_git"], "Var('chromium_git') + '")
 
+                # We are aiming to produce a quoted string. The code below
+                # will add a closing ', however if we did not collapse dart_git
+                # or chromium_git paths into a Var(...) ref we need to add
+                # an opening '.
+                if updated_value == dart_v:
+                  updated_value = "'" + updated_value
+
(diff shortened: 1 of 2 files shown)


engine/src/tools/dart/create_updated_flutter_deps.py
```diff
@@ -115,6 +115,13 @@ def Main(argv):
                 updated_value = dart_v.replace(new_vars["dart_git"], "Var('dart_git') + '/")
                 updated_value = updated_value.replace(old_vars["chromium_git"], "Var('chromium_git') + '")
 
+                # We are aiming to produce a quoted string. The code below
```
jason-simmons, line 118: This can be avoided by:
* adding an `'android_git': 'https://android.googlesource.com'` mapping to Flutter's `DEPS`
* changing the `engine/src/flutter/third_party/dart/third_party/perfetto/src` entry in Flutter's `DEPS` to `Var('android_git') + '/platform/external/perfetto@13ce0c9e13b0940d2476cd0cff2301708a9a2e2b'`
* adding `updated_value = updated_value.replace(old_vars["android_git"], "Var('android_git') + '")` above in this script

stuartmorgan-g:
test-exempt: utility script whose output is validated as part of the build

engine/src/tools/dart/create_updated_flutter_deps.py
jason-simmons: Created a PR with support for `android_git` and packages like Perfetto that are placed in `src` subdirectories:
https://github.com/flutter/flutter/pull/177099

Issue #172553, opened 2025-07-22
Harden FlutterFragment and FlutterActivity by restricting how flags are set (especially in release mode)


reidbaker:
When auditing the flutter android embedder we discovered that we have design pattern that makes it easy to accidentally introduce vulnerabilities. Specifically in https://github.com/flutter/flutter/blob/404fb3f14fdd665afb57b08912df00090ba6117c/engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterShellArgs.java#L70 the pattern of trusting flags that can be set by a malicious app in release mode should be modified.

Non public docs: go/flutter-android-intent-based-flag-analysis go/flutter-fragment-flag-security-analysis 

This issue tracks both the design review process and implementation of a better system for enabling engine flags in release mode. 

Ideally there would be no way to set flags with intent values, possibly by using android manifest flags to set values in release builds. If that proves non feasible then creating an allow list of flags that can be set in release mode should be the next option.

Issue #140310, opened 2023-12-18
Update documentation to clarify that cache-sksl cannot be used with Impeller.


desmeit:
### Steps to reproduce

if I run

`flutter run --profile --cache-sksl --purge-persistent-cache`

and then 

`M`

it is not possible to get the sksl file.

I get the error:

```
No data was received. To ensure SkSL data can be generated use a physical device then:
  1. Pass "--cache-sksl" as an argument to flutter run.
  2. Interact with the application to force shaders to be compiled.
```

I checked this
https://github.com/flutter/flutter/issues/105580
and this
https://github.com/flutter/flutter/issues/74873
but it is not helping me because I have the newest flutter version.
  

### Expected results

writing sksk file.

### Actual results

No data was received. To ensure SkSL data can be generated use a physical device then:
  1. Pass "--cache-sksl" as an argument to flutter run.
  2. Interact with the application to force shaders to be compiled.

### Code sample

<details open><summary>Code sample</summary>

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.16.4, on macOS 14.1.2 23B92 darwin-arm64, locale de-DE)
    • Flutter version 3.16.4 on channel stable at /Users/xxx/development/tools/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 2e9cb0aa71 (vor 7 Tagen), 2023-12-11 14:35:13 -0700
    • Engine revision 54a7145303
    • Dart version 3.2.3
    • DevTools version 2.28.4

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/xxx/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/xxx/Library/Android/sdk
    • ANDROID_SDK_ROOT = /Users/xxx/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)
    • All Android licenses accepted.

[...]

PR #177127, opened 2025-10-16
[Android] Add mechanism for setting Android engine flags via Android manifest

camsim99:
# Overview

> [!NOTE]  
> This PR is based on conversation & feedback on go/flutter-android-harden-engine-shell-arguments.

Adds a mechanism for setting Android engine flags via the manifest. If a flag is specified on the command line and in manifest metadata, the value specified on the command line will take precedence. Documentation is added on this mechanism

Additionally, this PR removes the exposure of`--cache-sksl` command line flag as per [https://github.com/flutter/flutter/issues/140310#issuecomment-2708459007](https://www.google.com/url?q=https://github.com/flutter/flutter/issues/140310%23issuecomment-2708459007&sa=D&source=docs&ust=1761156167162464&usg=AOvVaw3a8ubXTtv3apknY2-P9dKe).

Additionally, this PR adds documentation for the only two supported ways of setting engine flags moving forward -- via the command line or manifest. The `Intent` mechanism will be removed when https://github.com/flutter/flutter/issues/180686 is completed (intended to be a follow up to this PR).

As the unit tests in this PR only cover setting flags via manifest in debug mode, I will follow up this PR with an integration test to test that flags are appropriately respected/ignored in release mode. See https://github.com/flutter/flutter/pull/178383 for my currently working but WIP draft.

Part of https://github.com/flutter/flutter/issues/172553.

# Follow up work:

## Add integration test for this new added mechanism
This will land as an immediate follow-up to this PR. WIP in https://github.com/flutter/flutter/pull/178383.

## Remove support for setting shell arguments via `Intent`
This task will be a follow up to this work + the integration test landing, and will complete work for https://github.com/flutter/flutter/issues/172553. See https://github.com/flutter/flutter/issues/180686 for details.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineCommandLineFlags.java
```diff
@@ -0,0 +1,243 @@
...
+              ENABLE_VULKAN_GPU_TRACING,
+              LEAK_VM));
+
+  /**
+   * Converts a command line flag string to camel case that can be used as a manifest metadata name
+   * if the flag is configurable by manifest. See {@link FlutterEngineManifestFlags} for the list of
+   * flags for such available flags.
+   *
+   * <p>Examples: " --old-gen-heap-size=" becomes "oldGenHeapSize" and "--use-test-fonts" becomes
+   * "useTestFonts".
```
camsim99: Looking for feedback on the naming. I know camel case is idiomatic to Android and this is a pretty simple method that does that but I realize that using a conversion method is funky. We could also hard code the manifest names.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
```diff
@@ -327,13 +331,24 @@ public void attachToActivity(
...
+          (Boolean)
+              applicationMetaData.get(
+                  FlutterEngineManifestFlags.ENABLE_SOFTWARE_RENDERING.metaDataKey);
+      final boolean useSoftwareRendering =
+          useSoftwareRenderingFromManifest != null ? useSoftwareRenderingFromManifest : false;
+      flutterEngine.getPlatformViewsController().setSoftwareRendering(useSoftwareRendering);
+    } catch (PackageManager.NameNotFoundException e) {
+      Log.e(TAG, "Could not find ApplicationInfo when attempting to read manifest metadata.", e);
+      return;
+    }
```
camsim99, line 356: Not 100% sure what the use case was for setting this flag via an `Intent` so need to double check that this refactor works as intended. I assume the call to `flutterEngine.getPlatformViewsController().setSoftwareRendering` is safe after the call to `flutterEngine.getPlatformViewsControllerDelegator().attach` so worst case this can still be set with the command line with the dynamic setting workaround.

gaaclarke:
> Flags that have no real need to be set statically in production apps (like --trace-skia) are now only settable via the command line.

Was that a requirement?  It seems like there may be a valid case where someone would want to use a 3rd party profiler that launches an app itself, outside of flutters command-line tool.  I don't think perfetto works that way, you have to launch yourself.  RenderDoc works that way for gpu frame debugging.

docs/engine/Android-Flutter-Shell-Arguments.md
```diff
@@ -0,0 +1,96 @@
...
+
+Notes:
+- Flags that take values use the `--flag=value` form (with `=`). The Flutter tool forwards them in that form to the Android embedding.
+- If you wish to statically set flags for your application, setting them via the manifest is recommended and modifying the embedding to allow that is encouraged.
+
+Note: If a flag is specified both on the command line and in the manifest, the command-line value
+takes precedence at runtime.
+
+## How to set engine flags in the manifest
+All manifest metadata keys must be prefixed with the package name `io.flutter.embedding.android` and are suffixed with the camel case of the flag you wish to set.
```
gaaclarke:
```suggestion

All manifest metadata keys must be prefixed with the package name `io.flutter.embedding.android` and are suffixed with the camel case of the flag you wish to set.
```

docs/engine/Android-Flutter-Shell-Arguments.md
```diff
@@ -0,0 +1,96 @@
...
+### When to use manifest vs command line
+
+- Use the manifest when you want a fixed, reproducible baseline of engine flags for your app across all launches. This is ideal for CI and for enforcing a consistent config for your team.
+- Use the manifest when you want to vary flags by build mode or product flavor via manifest merging. For example, place metadata in `src/debug/AndroidManifest.xml`, `src/profile/AndroidManifest.xml`, and `src/release/AndroidManifest.xml` (or per-flavor manifests) to tailor flags per variant.
+
+See below for details on using each method.
+
+## How to set engine flags from the command line
+
+When you run a standalone Flutter app with the Flutter tool, many engine flags can be passed directly and are forwarded to the Android engine. Examples:
```
gaaclarke: nit: We typically stick with 80 line columns ( https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks )

docs/engine/Android-Flutter-Shell-Arguments.md
```diff
@@ -0,0 +1,96 @@
...
+## How to set engine flags in the manifest
+All manifest metadata keys must be prefixed with the package name `io.flutter.embedding.android` and are suffixed with the camel case of the flag you wish to set.
+
+For flags that take values, set the numeric or string value (without the leading `--flag=` prefix). For boolean flags, use `android:value="true"` to enable; omit or set `false` to disable.
+
+Note: Manifest-provided values are overridden by command-line flags if both are present.
+
+### Examples
+Set the `--old-gen-heap-size=` flag to 322 MB:
+```xml
```
gaaclarke, line 72:
```suggestion
### Examples

Set the `--old-gen-heap-size=` flag to 322 MB:

```xml
```

gaaclarke: We don't enforce it on CI but I like the `markdownlint` vscode plugin for editing markdown files.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -332,9 +331,7 @@ private FlutterEngineGroup.Options addEntrypointOptions(FlutterEngineGroup.Optio
             + " this FlutterFragment.");
 
     FlutterEngineGroup group =
-        engineGroup == null
-            ? new FlutterEngineGroup(host.getContext(), host.getFlutterShellArgs().toArray())
-            : engineGroup;
+        engineGroup == null ? new FlutterEngineGroup(host.getContext()) : engineGroup;
```
gaaclarke: Does this still work?  I would have expected the enforcement to be happening at a different layer.  Now this isn't forwarding any flags to the group.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineCommandLineFlags.java
gaaclarke: Yea, this is funky, but I think it's fine.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
```diff
@@ -311,59 +273,148 @@ public void ensureInitializationComplete(
...
+              .getPackageManager()
+              .getApplicationInfo(
+                  applicationContext.getPackageName(), PackageManager.GET_META_DATA);
+      Bundle applicationMetaData = applicationInfo.metaData;
+      boolean oldGenHeapSizeSet = false;
+      boolean isLeakVMSet = false;
+
+      if (applicationMetaData != null) {
+        for (String metadataKey : applicationMetaData.keySet()) {
+          if (metadataKey.equals(FLUTTER_EMBEDDING_KEY)) {
```
gaaclarke: Iterating over entries will be more efficient here so you don't have to do another lookup.  `filter()` makes the loops a bit more clear.  I may have got my java syntax a bit wrong, but you get the idea.

```java
for (Map.Entry<K, V> entry : applicationMetaData.entrySet().stream()
        .filter(entry -> !entry.getKey().equals(FLUTTER_EMBEDDING_KEY))) {
}
```

engine/src/flutter/shell/platform/android/test/io/flutter/embedding/engine/loader/FlutterLoaderTest.java
```diff
@@ -298,7 +298,7 @@ public void itSetsEnableImpellerFromMetaData() {
     FlutterJNI mockFlutterJNI = mock(FlutterJNI.class);
     FlutterLoader flutterLoader = new FlutterLoader(mockFlutterJNI);
     Bundle metaData = new Bundle();
-    metaData.putBoolean("io.flutter.embedding.android.EnableImpeller", true);
+    metaData.putBoolean("io.flutter.embedding.android.enableImpeller", true);
```
gaaclarke: I don't think we want to rename these.  There is a lot of documentation online, not all of it that we control that this will conflict with.

gaaclarke (commented):
I gave it a first glance.  This approach is looking good to me.  My concerns:

1) Looks like we are changing the naming of some manifest metadata keys, we shouldn't be (as noted in my comment)
1) It looks like some things may not be wired up yet, like getting flags when using the FlutterEngineGroup?  Was that strictly handing Intent cases that we don't want anymore?
1) Do we have good existing integration tests for setting theses flags?  I imagine there are integration tests that rely on setting the flags?

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
camsim99: ~Oh oops...I didn't realize what I was changing here at the time. I'll need to rework this to use my FlutterEngineCommandLines-blah blah flags. Thanks for flagging!~ See note below!

camsim99: I retract my earlier statement. This was in fact just to support flags set via `Intent` so I believe this is safe to delete.

docs/engine/Android-Flutter-Shell-Arguments.md
```diff
@@ -0,0 +1,189 @@
...
+All flags available on Android can be set via the command line **and** via
+manifest metadata. See `src/flutter/shell/common/switches.cc` for
+the list of all supported flags, and see
+`src/flutter/shell/platform/android/io/flutter/embedding/engine/`
+`FlutterShellArgs.java` for the list of flags that can be set for the
+Android shell.
+
+## When to use manifest metadata versus the command line
+
+Use the manifest when:
```
reidbaker, line 17: This should also contain the information for when to use the command line.

reidbaker (commented):
After trying to not duplicate reviews from https://github.com/flutter/flutter/pull/178383 I just keep spotting the same things. 

Going to hold off until you have a chance to incorporate that feedback then do a second pass.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -344,6 +344,29 @@ private FlutterEngineGroup.Options addEntrypointOptions(FlutterEngineGroup.Optio
...
+    Set<String> extrasKeys = extras.keySet();
+
+    for (String extrasKey : extrasKeys) {
+      FlutterShellArgs.Flag flag = FlutterShellArgs.getFlagFromIntentKey(extrasKey);
+      if (flag != null) {
+        Log.w(
+            TAG,
+            "Engine flags can no longer be set via Intent on Android. If you wish to set "
+                + flag.commandLineArgument
+                + ", see flutter/docs/engine/Android-Flutter-Shell-Arguments.md for alternative methods.");
```
camsim99: I assume most use cases for using engine flags are internal, so pointing to an internal doc isn't a huge deal for now, but I'm considering making this a website page too for external developers. Reviewers, let me know if you have thoughts on this.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
```diff
@@ -327,13 +331,37 @@ public void attachToActivity(
...
+          applicationMetaData.getBoolean(
+              FlutterShellArgs.ENABLE_SOFTWARE_RENDERING.metadataKey, false);
+      flutterEngine.getPlatformViewsController().setSoftwareRendering(useSoftwareRendering);
+    } catch (PackageManager.NameNotFoundException e) {
+      Log.w(
+          TAG,
+          "Could not set software rendering because application manifest metadata not found for package with name: "
+              + appContext.getPackageName()
+              + ".");
+    }
```
camsim99, line 361: I found that this is the only place where software rendering is configured in the embedding for the `PlatformViewsController`, so I modified this to check for configuration via application manifest here and added a warning for folks still trying to set the flag via `Intent`.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterShellArgs.java
```diff
@@ -4,207 +4,334 @@
...
-    if (intent.hasExtra(ARG_KEY_DART_FLAGS)) {
-      args.add(ARG_DART_FLAGS + "=" + intent.getStringExtra(ARG_KEY_DART_FLAGS));
+    /**
+     * Creates a new Flutter shell flag.
+     *
+     * <p>{@param allowedInRelease} determines whether or not this flag is allowed in release mode.
+     * Whenever possible, it is recommended to NOT allow this flag in release mode. Many flags are
+     * designed for debugging purposes and if enabled in production, could expose sensitive
+     * application data or make the app vulnerable to malicious actors.
+     */
```
camsim99, line 67: Added note here about allowing flags in release mode. Because I think the javadoc + constructor without the `allowedInRelease` parameter will discourage folks from allowing flags in release, I did not add a String `reason` but happy to add one if folks feel strongly.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/ApplicationInfoLoader.java
```diff
@@ -10,24 +10,17 @@
...
-      FlutterLoader.class.getName() + '.' + FlutterLoader.VM_SNAPSHOT_DATA_KEY;
-  public static final String PUBLIC_ISOLATE_SNAPSHOT_DATA_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.ISOLATE_SNAPSHOT_DATA_KEY;
-  public static final String PUBLIC_FLUTTER_ASSETS_DIR_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.FLUTTER_ASSETS_DIR_KEY;
+  // AndroidManifest.xml metadata keys for setting internal respective Flutter configuration values.
   public static final String NETWORK_POLICY_METADATA_KEY = "io.flutter.network-policy";
   public static final String PUBLIC_AUTOMATICALLY_REGISTER_PLUGINS_METADATA_KEY =
-      "io.flutter." + FlutterLoader.AUTOMATICALLY_REGISTER_PLUGINS_KEY;
+      "io.flutter.automatically-register-plugins";
```
camsim99, line 26: @reidbaker I left some flags here because they were not part of FlutterShellArgs for some reason. Why they don't use the `io.flutter.embedding.android` package name I don't know, but I did some digging and it looks like both are historical artifacts:

- `io.flutter.automatically-register-plugins` -- used before automatic registration of plugins
- `io.flutter.network-policy` -- https://docs.flutter.dev/release/breaking-changes/network-policy-ios-android

Thinking both of these can be removed. If you agree, I'll file an issue and link here to do as a follow up.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
camsim99: cc @gmackall who may have thoughts on this. TLDR software rendering can no longer be configured by `Intent` which seems currently required to enable software rendering for the `PlatformViewsController` because this code is the only caller of https://github.com/flutter/flutter/blob/fde117b7b144e3d98bf9e9146edf892204533e3c/engine/src/flutter/shell/platform/android/io/flutter/plugin/platform/PlatformViewsController.java#L843. Instead, I'm changing it to look for the setting in the manifest.

docs/engine/Android-Flutter-Shell-Arguments.md
```diff
@@ -0,0 +1,191 @@
...
+
+See below for details on using each method.
+
+## How to set engine flags from the command line
+
+When you run a standalone Flutter app with the Flutter tool, engine flags
+can be passed directly and are forwarded to the Android engine. Examples:
+
+```bash
+flutter run -d android \
```
reidbaker: For documentation lets remove flags we do not need or use the expanded form (--device-id)

docs/engine/Android-Flutter-Shell-Arguments.md
reidbaker, line 1: Overall great documentation

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
reidbaker: Instead of linking to a path what if you linked to the path on github?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
```diff
@@ -327,13 +331,39 @@ public void attachToActivity(
...
-            ? activity
-                .getIntent()
-                .getBooleanExtra(FlutterShellArgs.ARG_KEY_ENABLE_SOFTWARE_RENDERING, false)
-            : false;
-    flutterEngine.getPlatformViewsController().setSoftwareRendering(useSoftwareRendering);
+    // Previously, --enable-software-rendering could be set here via Intent. Warn developers
+    // about the new method for doing so if this was attempted.
+    final Intent intent = activity.getIntent();
+    if (intent != null) {
+      if (intent.hasExtra("enable-software-rendering")) {
```
reidbaker, line 349: Can you file a clean up bug for us to delete this code after a stable release and reference that bug here?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
reidbaker: I think you made a reasonable call here. Could/should we move the evaluation of this flag to where we inflate the rest of the flags then reference that inflated flag here?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterShellArgs.java
reidbaker: I think this is reasonable. Another option would be do document the reason in a java doc.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/ApplicationInfoLoader.java
reidbaker: Sounds good

engine/src/flutter/shell/platform/android/test/io/flutter/embedding/engine/FlutterShellArgsTest.java
```diff
@@ -2,35 +2,87 @@
...
-@RunWith(AndroidJUnit4.class)
 public class FlutterShellArgsTest {
+
   @Test
-  public void itProcessesShellFlags() {
-    // Setup the test.
-    Intent intent = new Intent();
-    intent.putExtra("dart-flags", "--observe --no-hot --no-pub");
-    intent.putExtra("trace-skia-allowlist", "skia.a,skia.b");
+  public void allFlags_containsAllFlags() {
```
reidbaker, line 19: I know reflection is gross but I love reflection in tests that prevents me from making a mistake.

reidbaker (approved):
Fantastic work thanks you!

camsim99:
This is ready for review. I'm not convinced that the failed Google test is related to this change.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -344,6 +348,29 @@ private FlutterEngineGroup.Options addEntrypointOptions(FlutterEngineGroup.Optio
     isFlutterEngineFromHost = false;
   }
 
+  // Soon, engine flags will no longer be settable via Intent extras, so warn
```
reidbaker: The meaning of soon changes over time. Instead consider linking to a bug that tracks the removal work.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -344,6 +348,29 @@ private FlutterEngineGroup.Options addEntrypointOptions(FlutterEngineGroup.Optio
...
+  // developers that engine shell arguments set that way will be ignored.
+  private void warnIfEngineFlagsSetViaIntent(Intent intent) {
```
reidbaker: can you use @NonNullable here for the input

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -1090,9 +1117,8 @@ private void ensureAlive() {
     @NonNull
     Lifecycle getLifecycle();
 
-    /** Returns the {@link FlutterShellArgs} that should be used when initializing Flutter. */
     @NonNull
-    FlutterShellArgs getFlutterShellArgs();
+    String[] getFlutterShellArgs();
```
reidbaker, line 1122: Is this public? if so i think this is a breaking change.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
```diff
@@ -251,7 +250,7 @@ public static class NewEngineFragmentBuilder {
     private String initialRoute = "/";
     private boolean handleDeeplinking = false;
     private String appBundlePath = null;
-    private FlutterShellArgs shellArgs = null;
+    private String[] shellArgs = null;
```
reidbaker, line 253: I am not sure why any of these changes are required. Can we not keep using FlutterShellArgs?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineConnectionRegistry.java
```diff
@@ -327,12 +329,37 @@ public void attachToActivity(
...
+    boolean useSoftwareRendering =
         activity.getIntent() != null
             ? activity
                 .getIntent()
-                .getBooleanExtra(FlutterShellArgs.ARG_KEY_ENABLE_SOFTWARE_RENDERING, false)
+                .getBooleanExtra(
+                    FlutterShellArgsIntentUtils.ARG_KEY_ENABLE_SOFTWARE_RENDERING, false)
             : false;
+
+    // The ability to set --enable-software-rendering via Intent will soon be removed. Warn
```
reidbaker: The meaning of soon changes over time.  I think in this case you can just say "will be removed" or "is planned to be removed"

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/ApplicationInfoLoader.java
```diff
@@ -10,24 +10,19 @@
...
-      FlutterLoader.class.getName() + '.' + FlutterLoader.ISOLATE_SNAPSHOT_DATA_KEY;
-  public static final String PUBLIC_FLUTTER_ASSETS_DIR_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.FLUTTER_ASSETS_DIR_KEY;
+  // TODO(camsim99): Remove support for these flags:
+  // https://github.com/flutter/flutter/issues/179276.
+  // AndroidManifest.xml metadata keys for setting internal respective Flutter configuration values.
   public static final String NETWORK_POLICY_METADATA_KEY = "io.flutter.network-policy";
   public static final String PUBLIC_AUTOMATICALLY_REGISTER_PLUGINS_METADATA_KEY =
-      "io.flutter." + FlutterLoader.AUTOMATICALLY_REGISTER_PLUGINS_KEY;
+      "io.flutter.automatically-register-plugins";
```
reidbaker, line 25: Why not keep FlutterLoader.AUTOMATICALLY_REGISTER_PLUGINS_KEY

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
```diff
@@ -89,8 +56,7 @@ public class FlutterLoader {
 
   private static FlutterLoader instance;
 
-  @VisibleForTesting
-  static final String aotSharedLibraryNameFlag = "--" + AOT_SHARED_LIBRARY_NAME + "=";
+  private boolean enableSoftwareRendering = false;
```
reidbaker, line 59: Why does this show up as a new flag. I see it in existing code https://github.com/search?q=repo%3Aflutter%2Fflutter%20enableSoftwareRendering&type=code but the diff makes it look new.

reidbaker (approved):

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
camsim99: It's not public; the only implementers are FlutterActivity and FlutterFragment. I'll address why I'm making this change [here](https://github.com/flutter/flutter/pull/177127#discussion_r2718086278).

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
camsim99: It's not required, but I'm making it because I prefer FlutterShellArgs to refer to my new class for all of the available flags for Android. I could do some renaming to keep this as is, but I would like to keep my change since it's not breaking and for familiarity purposes, i.e. folks already familiar with where to look that need to add a new flag will automatically look at the right place.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/ApplicationInfoLoader.java
camsim99: The flag is not at all referenced in FlutterLoader, so it didn't make sense for the variable declaration to exist there to me.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
camsim99: You're right, the flag isn't new! I only added this new variable so I could reference its value from FlutterEngineConnectionRegistry to address your feedback about evaluating this flag separately from where the rest are evaluated in FlutterLoader: https://github.com/flutter/flutter/pull/177127#discussion_r2566343452

PR #176799, opened 2025-10-09
Update `engine.version`

camsim99:
Sets `engine.version` to release build from https://github.com/flutter/flutter/commit/3eead6b0480ba4aacab8f0e0d058ac77f550bc1d. Part of https://github.com/flutter/flutter/issues/176744

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[...]

PR #177178, opened 2025-10-17
3.38.0 beta engine version

justinmc:
The version I set in https://github.com/flutter/flutter/pull/176799 was incorrect so the release build failed. This updates it to the correct engine version.


camsim99 (approved):
This looks right to me since https://github.com/flutter/flutter/commit/ced5a9334ed517ae2102e120be6d334c6afcd7e2 didn't touch the engine!

camsim99:
Sigh maybe that commit doesn't work because it was pre the fix for https://github.com/flutter/flutter/issues/176838. Maybe @jtmcdole can confirm. If so, do we need introduce another bogus engine change? Maybe we can delete the file I added in https://github.com/flutter/flutter/pull/176842

justinmc:
Alright I have merged another PR that deletes the engine file (https://github.com/flutter/flutter/pull/177181) and updated the SHA to that merge commit. The analyzer test just failed because the engine artifacts haven't finished building yet. After they finish I will rerun and, if we're right, it should pass.

PR #177181, opened 2025-10-17
Trigger engine build, and clean up trigger file

justinmc:
My PR updating engine.version (https://github.com/flutter/flutter/pull/177178) for the 3.38.0 beta release isn't passing the analyzer because the engine artifacts don't exist. This PR will trigger another engine build, and then I'll use its merge SHA in engine.version. And it conveniently deletes the file used for this same purpose in https://github.com/flutter/flutter/pull/176842.

### DIFF
--- a/engine/TRIGGER.md
+++ b/engine/TRIGGER.md
@@ -1,2 +1,0 @@
-<!-- TODO(camsim99): Delete this file. -->
-This file is purely to trigger an engine release build and will be deleted in a future commit pre-release.


camsim99 (approved):
Hope this works! 🤞

jtmcdole (approved):

jtmcdole:
So the thought is: because the builder failed last time, we actually didn't produce all the artifacts?

camsim99:
That's my guess? Of course I don't understand the infrastructure _that well_ so maybe there was another way around that issue...if that was even the cause.

PR #177220, opened 2025-10-19
feat: add RoundedSuperellipseInputBorder

### LINKED ISSUE #176987: Implement RoundedSuperellipseInputBorder
### Use case

I want to use the squircle corners of iOS for my input fields. There is the `RoundedSuperellipseBorder` to do exactly that, but a `TextField` works only with `InputBorder`s that extend from `ShapeBorder`, which makes those classes sibling classes of `OutlinedBorder` which `RoundedSuperellipseBorder` extends from.

### Proposal

I believe there should be a `RoundedSuperellipseInputBorder` that fills the gap. On iOS native text fields use the same corner math, see the search bar in the contacts app, so this is necessary when aiming for platform respective UI.

rkishan516:
# Whats new

* Add RoundedSuperellipseInputBorder

fixes: #176987 

## Pre-launch Checklist

- [x] I signed the [CLA].


dkwingsmt:
Sorry, I thought I commented on this PR. Anyway, I was going to say that I'd like more discussion under https://github.com/flutter/flutter/issues/176987 before moving forward with actual coding. Can you take a look?

packages/flutter/lib/src/material/input_border.dart
```diff
@@ -574,3 +574,234 @@ class OutlineInputBorder extends InputBorder {
...
+    if (shape is OutlinedBorder) {
+      (shape as OutlinedBorder).paintInterior(canvas, rect, paint, textDirection: textDirection);
+    } else {
+      // Fallback for shapes that don't support paintInterior
+      canvas.drawPath(shape.getOuterPath(rect, textDirection: textDirection), paint);
+    }
+  }
+
+  @override
+  bool get preferPaintInterior => true;
```
dkwingsmt: Shouldn't it be `shape.preferPaintInterior`, since the provided shape might not have defined `paintInterior`?

packages/flutter/lib/src/material/input_border.dart
```diff
@@ -574,3 +574,234 @@ class OutlineInputBorder extends InputBorder {
   @override
   int get hashCode => Object.hash(borderSide, borderRadius, gapPadding);
 }
+
```
dkwingsmt, line 579: Can you also add this new class to `OutlineInputBorder `'s "See also"?

dkwingsmt (approved):
LGTM with nits. Thank you for the awesome changes! And sorry for late review.

packages/flutter/lib/src/material/input_border.dart
rkishan516: Updated.

packages/flutter/lib/src/material/input_border.dart
rkishan516: Updated.

packages/flutter/lib/src/material/input_border.dart
```diff
@@ -574,3 +576,234 @@ class OutlineInputBorder extends InputBorder {
...
+///
+/// ** See code in examples/api/lib/material/shaped_input_border/shaped_input_border.0.dart **
+/// {@end-tool}
+///
+/// See also:
+///
+///  * [OutlineInputBorder], a traditional rounded rectangle border.
+///  * [RoundedSuperellipseBorder], which can be used with this border for iOS-style shapes.
+///  * [UnderlineInputBorder], the default [InputDecorator] border which
+///    draws a horizontal line at the bottom of the input decorator's container.
```
justinmc, line 603: Nit: I would swap the order of RoundedSuperellipseBorder and UnderlineInputBorder so that the two subclasses of InputBorder are next to each other.

packages/flutter/lib/src/material/input_border.dart
```diff
@@ -574,3 +576,234 @@ class OutlineInputBorder extends InputBorder {
...
+  ///
+  /// If an application does not specify a [borderSide] parameter of
+  /// value [BorderSide.none], the input decorator substitutes its own, using
+  /// [copyWith], based on the current theme and [InputDecorator.isFocused].
+  ///
+  /// See also:
+  ///
+  ///  * [InputDecoration.floatingLabelBehavior], which should be set to
+  ///    [FloatingLabelBehavior.never] when the [borderSide] is
+  ///    [BorderSide.none]. If let as [FloatingLabelBehavior.auto], the label
```
justinmc: "let" => "left"

justinmc: Would it be possible/useful to catch this with an assert in InputDecoration?

packages/flutter/lib/src/material/input_border.dart
```diff
@@ -574,3 +576,234 @@ class OutlineInputBorder extends InputBorder {
...
+  }
+
+  @override
+  bool get preferPaintInterior => shape.preferPaintInterior;
+
+  Path _gapBorderPath(Rect rect, double start, double extent, {TextDirection? textDirection}) {
+    // Create a continuous path for the border with a gap in the top edge.
+    final Path outerPath = shape.getOuterPath(rect, textDirection: textDirection);
+
+    // If there's no meaningful gap, return the full outline
```
justinmc: Nit: Periods at the end of your comments here and elsewhere.

justinmc: Nit: Periods at the end of your comments here and elsewhere.

justinmc (approved):
LGTM with nits :+1:

packages/flutter/lib/src/material/input_border.dart
rkishan516: This behaviour is specific to `ShapedInputBorder`, I don't think coupling that in InputDecoration will be a good idea. And since `ShapedInputBorder` don't have access to floatingLabelBehavior, its not feasible to assert this.

dkwingsmt (approved):
Thank you!

PR #177244, opened 2025-10-20
Add assertion to Navigator.pop when stack is empty to improve developer experience

Masrafi:
## Motivation: 
Beginners frequently encounter a confusing "black screen" when Navigator.pop() is unintentionally called on an empty route stack. This behavior lacks a clear explanation in debug mode, leading developers to incorrectly assume the app is broken. This change clarifies the behavior by throwing an explicit, actionable assertion.

## Technical Change:

Assertion: An assert(() { ... }()) block was added within NavigatorState.pop() to check if any routes are present in the history before finalizing the pop operation.

Stability Fix (Critical): The entire core logic of pop() was wrapped in a try/finally block. This guarantees that the private _debugLocked flag is reliably reset to false in the finally block, ensuring that the NavigatorState does not enter a corrupted state when the assertion is triggered. This fix is necessary to prevent the !_debugLocked crash during cleanup.

## Testing:

New Test Case: A testWidgets case was added to navigator_test.dart to specifically invoke navigator.pop() on a single route stack. The test uses the throwsA(isA<AssertionError>()) matcher to confirm that the fatal assertion is correctly thrown.

Test Suite Compatibility: Several existing framework tests that previously relied on the silent, non-asserting failure of pop() on an empty stack were updated to explicitly expect the new AssertionError, resolving wide-spread CI failures.

## Conclusion
This change significantly improves the developer experience during development by converting an implicit runtime error into an explicit, debug-only assertion.


packages/flutter/lib/src/widgets/navigator.dart
```diff
@@ -5592,6 +5592,17 @@ class NavigatorState extends State<Navigator> with TickerProviderStateMixin, Res
...
       _flushHistoryUpdates(rearrangeOverlay: false);
+      assert(() {
+        if (kDebugMode) {
+          final bool hasPresentRoute = _history.any(_RouteEntry.isPresentPredicate);
+          if (!hasPresentRoute) {
+            debugPrint(
+              'Tried to pop a route on an empty stack. Ensure there is at least one route before calling Navigator.pop().',
+            );
+          }
+        }
```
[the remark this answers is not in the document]
Masrafi: The kDebugMode is removed. It is okay now.

packages/flutter/lib/src/widgets/navigator.dart
```diff
@@ -5592,6 +5592,15 @@ class NavigatorState extends State<Navigator> with TickerProviderStateMixin, Res
...
     if (entry.currentState == _RouteLifecycle.pop) {
       _flushHistoryUpdates(rearrangeOverlay: false);
+      assert(() {
+        final bool hasPresentRoute = _history.any(_RouteEntry.isPresentPredicate);
+        if (!hasPresentRoute) {
+          debugPrint(
+            'Tried to pop a route on an empty stack. Ensure there is at least one route before calling Navigator.pop().',
+          );
+        }
+        return true;
```
justinmc: Instead of using debugPrint, I would just return hasPresentRoute, and then add the string as an error message parameter to `assert`.

packages/flutter/test/widgets/navigator_test.dart
```diff
@@ -5219,6 +5219,36 @@ void main() {
...
+        }
+      };
+
+      await tester.pumpWidget(MaterialApp(home: Container()));
+
+      final NavigatorState navigator = tester.state(find.byType(Navigator));
+      navigator.pop();
+      await tester.pump(); // Pump again to ensure all async framework work is complete
+
+      debugPrint = oldDebugPrint;
```
justinmc: Then here, instead of checking debugMessages, use takeException and expect an AssertionError.

justinmc (commented):
Thanks for the PR! I agree that this is a useful error message. I think an app probably shouldn't knowingly pop the last route on its Navigator... Even in the case of a nested Navigator, you probably don't want to pop the last route. But we'll see what the Google tests say.

packages/flutter/test/widgets/navigator_test.dart
Masrafi: Absolutely. This is a much cleaner way to test assertions.
I've refactored the test case in navigator_test.dart to fully comply with this:

1. Removed the DebugPrintCallback logic.
2. The test now relies on try/catch to get the AssertionError object.
3. We use tester.takeException() to confirm the error was captured and to clear the internal test harness log.
4. The test verifies the type and uses a simplified contains() check to ensure the specific warning message is present on the exception object.

This successfully verifies the new behavior and the test is now passing. Thanks again!

packages/flutter/lib/src/widgets/navigator.dart
Masrafi: Thanks for the suggestion! I’ve updated the code to return hasPresentRoute directly and moved the message into the assertion parameter instead of using debugPrint.

packages/flutter/test/widgets/navigator_test.dart
Masrafi: Done! The test now uses takeException and expects an AssertionError instead of checking debug messages. This makes the test more robust and aligned with Flutter best practices.

packages/flutter/lib/src/widgets/navigator.dart
```diff
@@ -5592,6 +5592,13 @@ class NavigatorState extends State<Navigator> with TickerProviderStateMixin, Res
     }
     if (entry.currentState == _RouteLifecycle.pop) {
       _flushHistoryUpdates(rearrangeOverlay: false);
+      assert(
+        () {
+          final bool hasPresentRoute = _history.any(_RouteEntry.isPresentPredicate);
+          return hasPresentRoute;
+        }(),
+        'Tried to pop a route on an empty stack. Ensure there is at least one route before calling Navigator.pop().',
```
chunhtai: This message doesn't seem accurate, I think this check the stack is empty after the pop?

chunhtai: also 
> Ensure there is at least one route before calling Navigator.pop().

this is not good neither, having just one route before the pop means it stack will be empty after the pop

packages/flutter/test/widgets/navigator_test.dart
```diff
@@ -5219,6 +5219,26 @@ void main() {
...
+      await tester.pumpWidget(const MaterialApp(home: Text('Home')));
+      final NavigatorState navigator = tester.state(find.byType(Navigator));
+
+      Object? popException;
+      try {
+        navigator.pop();
+      } catch (e) {
+        popException = e;
+      }
+      expect(popException, isA<AssertionError>());
```
chunhtai: should check for the message.

Masrafi:
Hi team,

I've pushed the fix for asserting on an empty stack.

Code Fix: I've confirmed stability by implementing the necessary try/finally block in Navigator.pop() to guarantee the _debugLocked flag is released. This resolves the framework instability issue.

Current Issue: The CI is failing because existing framework tests are now correctly hitting this new assertion and crashing.

Could a team member please check the logs and provide the file paths for the failing tests? I need to update them with expect(..., throwsA(...)) to resolve the CI failures.

Thanks!

PR #177252, opened 2025-10-20
Adds a new CI build for Linux host DDM-enabled artifacts

goderbauer:
Work towards [b/452833651](b/452833651).

This adds a new flavor of linux_host_engine, which enables the DDM build flag with the goal of building a flutter_tester binary that supports loading DDMs in tests.

Tested locally via `et build --config ci/host_debug_ddm`.

This roughly follows https://github.com/flutter/flutter/pull/168717, which added a similar DDM-enabled build for iOS.

I've branched this off the `linux_host_engine.json`, added the build flag for DDM support, and removed things we don't need.

### DIFF
--- a/engine/src/flutter/ci/builders/linux_host_engine_ddm.json
+++ b/engine/src/flutter/ci/builders/linux_host_engine_ddm.json
@@ -18,7 +18,7 @@
                     "base_path": "out/ci/host_debug_ddm/zip_archives/",
                     "type": "gcs",
                     "include_paths": [
-                        "out/ci/host_debug_ddm/zip_archives/linux-x64/artifacts.zip"
+                        "out/ci/host_debug_ddm/zip_archives/linux-x64-ddm/artifacts.zip"
                     ],
                     "realm": "production"
                 }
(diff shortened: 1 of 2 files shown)


sigmundch (approved):
Thanks!

engine/src/flutter/ci/builders/linux_host_engine_ddm.json
```diff
@@ -0,0 +1,55 @@
...
+            "gclient_variables": {
+                "download_android_deps": false,
+                "download_jdk": false,
+                "use_rbe": true
+            },
+            "gn": [
+                "--target-dir",
+                "ci/host_debug_ddm",
+                "--runtime-mode",
+                "debug",
```
jtmcdole, line 39: the linux_host_engine.json has this as release?

jtmcdole (approved):
just a call out about debug vs release.

engine/src/flutter/ci/builders/linux_host_engine_ddm.json
goderbauer: The build that is producing the `artifacts.zip` (the one I am interested in here) seems to be configured to debug mode, see 

https://github.com/flutter/flutter/blob/ce25e25e2ce304cedabf6d883ac92721f144409e/engine/src/flutter/ci/builders/linux_host_engine.json#L44

Or am I misunderstanding how this works?

PR #177293, opened 2025-10-20
Roll reclient to version 185

jason-simmons:
* The reclient bootstrap shutdown command now requires the config file flag.
* The reproxystatus command must be run before stopping the reproxy in order to access RBE metrics.

### DIFF
--- a/engine/src/flutter/tools/pkg/engine_build_configs/lib/src/build_config_runner.dart
+++ b/engine/src/flutter/tools/pkg/engine_build_configs/lib/src/build_config_runner.dart
@@ -523,9 +523,7 @@ final class BuildRunner extends Runner {
         pid: result.pid, // pid,
       );
     }
-    if (okMessage == null) {
-      okMessage = bootstrapResult.stdout.trim();
-    }
+    okMessage ??= bootstrapResult.stdout.trim();
     eventHandler(
       RunnerResult(
         '${build.name}: RBE ${shutdown ? 'shutdown' : 'startup'}',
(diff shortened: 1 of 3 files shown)


engine/src/flutter/tools/pkg/engine_build_configs/lib/src/build_config_runner.dart
```diff
@@ -518,9 +523,8 @@ final class BuildRunner extends Runner {
         pid: result.pid, // pid,
       );
     }
-    String okMessage = bootstrapResult.stdout.trim();
-    if (shutdown) {
-      okMessage = await _reproxystatus() ?? okMessage;
+    if (okMessage == null) {
+      okMessage = bootstrapResult.stdout.trim();
     }
```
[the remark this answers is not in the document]
jason-simmons: done

engine/src/flutter/tools/pkg/engine_build_configs/lib/src/build_config_runner.dart
```diff
@@ -518,10 +523,7 @@ final class BuildRunner extends Runner {
         pid: result.pid, // pid,
       );
     }
-    String okMessage = bootstrapResult.stdout.trim();
```
zanderso, line 521: There are some tests of this code in https://github.com/flutter/flutter/blob/master/engine/src/flutter/tools/pkg/engine_build_configs/test/build_config_runner_test.dart.

Is there any way to write a test that would fail if the calls were out of order?

jason-simmons: Added a test

zanderso (approved):
Thanks!

jtmcdole:
lgtm

jason-simmons:
See https://github.com/flutter/flutter/issues/177356 - Fuchsia RBE builds on CI have often been falling back to local compilation after the recent roll of the Clang toolchain.

PR #177337, opened 2025-10-21
CupertinoSheetRoute with scrolling and dragging

### LINKED ISSUE #161687: Cupertino Sheet should have drag to dismiss and nested scrolling work together
On native iOS, when you have a sheet open with scrollable content, if you start scrolling, it will scroll normally. But then when you drag downwards, once you reach the top of the scrollable content, the sheet starts it's drag to dismiss gesture. Then if you drag back up again, once the the sheet reaches it's original height, the nested scroll starts back up again.

https://github.com/user-attachments/assets/72082b7b-414e-4093-a02a-27b8d2188dd3

In Flutter currently we can either give the nested scroll or the drag to dismiss gesture priority. And we can't switch back and forth between either animations on one drag event.

MitchellGoodwin:
Fixes #161687

Enables scrollable content within the sheet to work together with the dragging animation.

Flutter:

https://github.com/user-attachments/assets/3a9994e3-078a-410e-8288-95592713ab39

Native:

https://github.com/user-attachments/assets/c0bd46fb-26f5-41a6-bf42-6c78c23ddf34 

Fling when scrolling to the top vs fling when the scrollable content is already at the top

https://github.com/user-attachments/assets/e934facc-36d2-4e81-9ccd-a3e9dd7b6201

When scrolling is enabled, then the sheet will no longer have a drag gesture recognizer over the sheet content, and will instead rely on the scrollable content to trigger the drag. A non-scrolling area can be wrapped with `CupertinoSheetDragArea` to put a drag gesture recognizer only on that area, convenient for navbars. See [cupertino_sheet.3.dart](https://github.com/flutter/flutter/pull/177337/files#diff-3df71aa5e6ddb6184007a85e847a9a4883515697b7014764e1978d240d72e0e1) for a full example.

## Pre-launch Checklist

- [x] I signed the [CLA].


lukepighetti:
missing one minor bit of fidelity. if you are scrolled down and fling to the top it should spring at the top instead of slam to a halt

MitchellGoodwin:
The physics when scrolling to the top is definitely not there yet. It either stops suddenly or bounces when it shouldn't when the drag to dismiss happens. I opened this draft PR to get some feedback on the overall approach.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -153,18 +154,25 @@ Future<T?> showCupertinoSheet<T>({
...
+        ? CupertinoSheetRoute<T>(builder: effectiveBuilder, enableDrag: enableDrag)
+        : CupertinoSheetRoute<T>.scrollable(
+            scrollableBuilder: scrollableBuilder,
+            enableDrag: enableDrag,
+          );
+
+    return Navigator.of(context, rootNavigator: true).push<T>(route);
   } else {
-    widgetBuilder = (BuildContext context) {
+    Widget nestedNavigationContent(Widget child) {
```
Piinks: Since this is changed to define an inline function, is the `else` necessary?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -184,21 +192,28 @@ Future<T?> showCupertinoSheet<T>({
...
       );
-    };
-  }
+    }
 
-  return Navigator.of(
-    context,
-    rootNavigator: true,
-  ).push<T>(CupertinoSheetRoute<T>(builder: widgetBuilder, enableDrag: enableDrag));
+    final PageRoute<T> route = effectiveBuilder != null
```
Piinks: Can you simplify the logic to de-dupe and call on your new wrapper method based on `useNestedNavigation`?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -922,3 +1024,154 @@ class _CupertinoDragGestureController<T> {
...
+
+  @override
+  Drag drag(DragStartDetails details, VoidCallback dragCancelCallback) {
+    // Save this so we can call it later if we have to [goBallistic] on our own.
+    _dragCancelCallback = dragCancelCallback;
+    return super.drag(details, dragCancelCallback);
+  }
+}
+
+class _NeverUserScrollableScrollPhysics extends ScrollPhysics {
```
Piinks: Why is this needed over using NeverScrollableScrollPhysics?

MitchellGoodwin: I didn't want to set `allowImplicitScrolling` to false.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -567,7 +612,11 @@ class CupertinoSheetRoute<T> extends PageRoute<T> with _CupertinoSheetRouteTrans
         borderRadius: const BorderRadius.vertical(top: Radius.circular(12)),
         child: CupertinoUserInterfaceLevel(
           data: CupertinoUserInterfaceLevelData.elevated,
-          child: _CupertinoSheetScope(child: builder(context)),
+          child: Builder(
+            builder: (BuildContext context) {
+              return _CupertinoSheetScope(sheetContext: context, child: _effectiveBuilder(context));
```
Piinks: Why do we need to pass around context? That's an anti-pattern.

MitchellGoodwin: Ultimately it's so that `_CupertinoDragGestureDetector` is able to figure out the height of the sheet for knowing how much to adjust the page transition in response to a user drag.

It gets there through `CupertinoSheetDragArea` looking up the `_CupertinoSheetScope` above it and passing along the context to `_CupertinoDragGestureDetector`.

Piinks: Yeah, it's really not something we should do. Passing around build context is not safe. DraggableScrollableSheet keeps track of the sheet's extent. Are we able to do that here?

MitchellGoodwin: I think I can change it so that _CupertinoSheetScope tracks the height/extent rather than the context and that'll work pretty smoothly.

packages/flutter/lib/src/cupertino/sheet.dart
chunhtai, line 173: somewhere in the doc should mention how/when to use which builder

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -640,6 +691,9 @@ mixin _CupertinoSheetRouteTransitionMixin<T> on PageRoute<T> {
   /// If `true`, dragging is enabled; otherwise, it remains fixed.
   bool get enableDrag;
 
+  /// Determines whether the content can be scrolled.
+  bool get enableScroll;
```
chunhtai: can this be private?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -720,12 +788,18 @@ class _CupertinoDragGestureDetectorState<T> extends State<_CupertinoDragGestureD
...
+
+  BuildContext get sheetContext => widget.sheetContext ?? context;
+
   @override
   void initState() {
     super.initState();
     assert(_stretchDragController == null);
     _stretchDragController = _StretchDragControllerProvider.maybeOf(context);
     _recognizer = VerticalDragGestureRecognizer(debugOwner: this)
+      ..velocityTrackerBuilder = _cupertinoVelicityBuilder
```
chunhtai:
```suggestion
      ..velocityTrackerBuilder = _cupertinoVelocityBuilder
```

chunhtai (commented):
For Accessibility, we would need to focus on two things
1. dismissable

in talkback this is fine as user can do gesture down and left to issue a android back.

For voiceOver though, it is two fingers Z gesture and it calls Semantics(onDismiss). Looking at the semantics tree, I didnt see the semantics action is set.

2. the draghandle

In talkback, tap should be able to dismiss the sheet similar to https://github.com/flutter/flutter/blob/7b98d50631e52399041060ae2ae19163cd5cba87/packages/flutter/lib/src/material/bottom_sheet.dart#L367

For voicerOver, we may need to check the iOS behavior, do you know whether there is builtin app that uses the native draggable sheet?

MitchellGoodwin:
> also, while not yet wired up in the engine, can you also apply SemanticsRole.dragHandle to the handle so that we won't forget about this in the future when it is wired up

CupertinoSheet currently does not add a drag handle. For the native sheet widget, the drag handle is normally added when the sheet is resizable, which CupertinoSheet does not do by default, nor is this PR adding that.

However, CupertinoSheetDragArea _could_ be wrapped around a drag handle type widget manually added by a dev.

So should we either

1. Not do anything with the drag handle role for this PR.
2. Add either a `isDragHandle` or a more generic SemanticsRole property to CupertinoSheetDragArea.
3. Add an other SemanticsRole somewhere else that I'm missing.

Piinks:
> CupertinoSheet currently does not add a drag handle. For the native sheet widget, the drag handle is normally added when the sheet is resizable

@MitchellGoodwin have you checked the native behavior with accessibility controls enabled? Does SwiftUI expose the ability to add drag handles?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -89,14 +89,21 @@ const double _kSheetScaleFactor = 0.0835;
...
 /// This is a convenience method for displaying [CupertinoSheetRoute] for common,
-/// straightforward use cases. The Widget returned from `pageBuilder` will be
-/// used to display the content on the [CupertinoSheetRoute].
+/// straightforward use cases. There are two different build methods that are provided
+/// by the method for building the content of the [CupertinoSheetRoute]. Use `builder`
+/// if there is no content within the sheet that needs to scroll. This is good
+/// for sheets with a simple display, however scrolling gestures will conflict
+/// with the drag to dismiss gesture. `scrollableBuilder` will enable scrollable
+/// content within the sheet. See [CupertinoSheetRoute.scrollable] for more
+/// information.
```
justinmc: Nit: Should you mention that this is for vertical scrolling?

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @override
+  Size get preferredSize {
+    return child.preferredSize;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return CupertinoSheetDragArea(child: child);
```
justinmc: Could this be automatically built into CupertinoNavigationBar?

justinmc (commented):
Just trying to think through an alternative approach... Rather than a new constructor for CupertinoSheetRoute that passes a ScrollController, could you use `PrimaryScrollController.of` inside of CupertinoSheetRoute and ask users to do the same and pass that to their relevant scroll view? I'm no expert in this area but it seems kind of aligned with the intention of PrimaryScrollController.

If not PrimaryScrollController, then would a new similar InheritedWidget make sense? ...Or would it just make this more complicated and less discoverable?

Otherwise, do you know how this kind of thing works in SwiftUI? Is it always the entire CupertinoSheet that is scrollable?

Assuming none of these drive-by ideas is viable, and that this covers the main use cases from SwiftUI, then this approach looks good to me!

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: Good callout

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
MitchellGoodwin: It's possible, and it would be convenient as this would theoretically be a common use case. I'd just worry about this behavior being discoverable.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -89,14 +89,21 @@ const double _kSheetScaleFactor = 0.0835;
...
+
 /// Shows a Cupertino-style sheet widget that slides up from the bottom of the
 /// screen and stacks the previous route behind the new sheet.
 ///
 /// {@youtube 560 315 https://www.youtube.com/watch?v=5H-WvH5O29I}
 ///
 /// This is a convenience method for displaying [CupertinoSheetRoute] for common,
-/// straightforward use cases. The Widget returned from `pageBuilder` will be
-/// used to display the content on the [CupertinoSheetRoute].
+/// straightforward use cases. There are two different build methods that are provided
```
Piinks: I know this is not exactly part of this change, but lil nit:


```suggestion
/// This is a convenience method for displaying [CupertinoSheetRoute] for
/// most use cases. There are two different build methods that are provided
```

Reminded me of https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#use-the-passive-voice-recommend-do-not-require-never-say-things-are-simple

Follow-up question - what are the less common or straightforward sue bases, and what should people use in those scenarios?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -545,29 +570,87 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
   /// Creates a page route that displays an iOS styled sheet.
-  CupertinoSheetRoute({super.settings, required this.builder, this.enableDrag = true});
+  CupertinoSheetRoute({super.settings, this.builder, this.enableDrag = true})
+    : scrollableBuilder = null,
+      _sheetMode = _CupertinoSheetMode.dragOnly;
+
+  /// Creates a page route that displays an iOS styled sheet that is expected to
+  /// have scrollable content that spans the whole sheet.
+  ///
+  /// When this constructor is used, the content of the sheet will not watch for
```
Piinks: Can you explain this a bit more? If the content is a scrollable, won't it actually be watching for those drag gestures?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -545,29 +570,87 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
+  /// drag gestures to trigger the drag downwards to dismiss behavior. Instead,
+  /// if a vertical happens within a scrollable widget that uses the [ScrollController]
```
Piinks: I think a word is missing here.

I would focus this paragraph on the expected behavior. Some of the implementation details are not super relevant to the user.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -545,29 +570,87 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
+  /// provided by [scrollableBuilder], the sheet will trigger either scrolling or
+  /// the drag to dismiss gesture based on the state of the widget.
+  ///
+  /// In order to trigger the drag to dismiss gesture outside of the scrollable
+  /// content, wrap that area with a CupertinoSheetDragArea.
+  ///
+  /// {@tool dartpad}
+  /// This example shows how to show a Cupertino Sheet with scrollable content,
+  /// and a navbar that will trigger the drag only on a gesture started within
```
Piinks: Breadcrumbs would help here, what is the navbar?

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
Piinks: This does feel a bit kludgy from the developer experience side. 

What native behavior are we trying to emulate by exposing this drag wrapper for non-scrolling parts? If I had a header that I wanted to have drag the sheet, why not put it in a pinned sliver in a CustomScrollView as part of the scrollable? Would that achieve the same effect without adding more API for the developer to have to wire up?

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: > Follow-up question - what are the less common or straightforward sue bases, and what should people use in those scenarios?

showCupertinoSheet adds some boilerplate around nested navigation, and making sure the route is pushed to the right Navigator, which should work for most use cases of both. But for nested navigation in particular there might be configurations that our boilerplate might not fit, so in those cases they would use CupertinoSheetRoute directly and set things up themselves. This includes add-to-app situations with just the sheet, restorable state, custom Navigators, etc.

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: What this means is that unlike the non-scrolling version, a vertical drag gesture recognizer won't be placed over the whole sheet, and instead the vertical drags would be watched for by the scrolling content.

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
MitchellGoodwin: > What native behavior are we trying to emulate by exposing this drag wrapper for non-scrolling parts?

We want to be able to trigger the drag-to-dismiss behavior when a drag originates outside of the scroll area, and only the drag-to-dismiss, not a scroll. For example, if you start a drag on the header of a native sheet, it will not scroll the content below it, no matter what direction you drag. This is also the only way to trigger the stretch upwards drag effect on a sheet. If you start the drag from within the scrolling area then the draggable/scrollable behavior would happen.

> If I had a header that I wanted to have drag the sheet, why not put it in a pinned sliver in a CustomScrollView as part of the scrollable? Would that achieve the same effect without adding more API for the developer to have to wire up?

I believe in that case then a scroll would be triggered on a drag upwards originating in the pinned sliver, instead of the stretch animation.

From native: a drag starting within the navbar will not scroll and will stretch upwards. A drag starting below the navbar will scroll depending on the direction, and will not scroll upwards.
ScreenRecording2025-10-21at12 39 56PM-ezgif com-video-to-gif-converter

Basically outside of the scrolling area we want the sheet to work as if it's a non-scrolling sheet. This wrapper lets a developer wrap the original vertical drag gesture recognizer only around those areas.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -153,52 +161,69 @@ Future<T?> showCupertinoSheet<T>({
...
+          return <Route<void>>[
+            CupertinoPageRoute<void>(
+              builder: (BuildContext context) {
+                return PopScope(
+                  canPop: false,
+                  onPopInvokedWithResult: (bool didPop, Object? result) {
+                    if (didPop) {
+                      return;
+                    }
+                    Navigator.of(context, rootNavigator: true).pop(result);
```
chunhtai: I know this is not introduced by this pr, but what if there is multiple navigator above the nested navigator? I thought we may want to pop just one level above this nested navigator

MitchellGoodwin:
> Can you talk more about this? does this mean for non scrollable cupertino sheet user can start dragging anywhere in the sheet to start popping gesture? This sounds a bit unexpected.

Yes, that it is what it means, and that's how the native sheet works.

https://github.com/user-attachments/assets/3163fed9-4a46-430d-a8ee-9d483115d679

You can disable that with the `enableDrag` in the constructor. Then with this PR you could re-enable it for, say just the navbar with the drag area class added in this PR.

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: `showCupertinoSheet` always pushes to the root navigator, especially as it needs to slide the whole page down. So popping the sheet also targets the root navigator, as the last route in the stack should theoretically be the sheet you are meaning to pop, in the case of nested navigation.

If there is a situation where that does not work for their app, then we direct people to use `CupertinoSheetRoute` directly instead of `showCupertinoSheet`.

chunhtai: Ah ok, makes sense

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
Piinks: Would it be easier for the developer if, in the case of a scrollable sheet, we provide a header parameter and we can manage the gesture resolution ourselves internally?
What is the native experience like for developers building the UI you described above?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -153,52 +161,69 @@ Future<T?> showCupertinoSheet<T>({
   )
   WidgetBuilder? pageBuilder,
   WidgetBuilder? builder,
```
Piinks, line 184: Just curious since I've been learning a bit more about this area, what is the difference between pageBuilder and builder?

packages/flutter/lib/src/cupertino/sheet.dart
Piinks: Except for things not part of the scrollable?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -766,7 +884,11 @@ class _CupertinoDragGestureDetectorState<T> extends State<_CupertinoDragGestureD
     if (_stretchDragController == null) {
       return;
     }
-    _dragGestureController!.dragUpdate(details.primaryDelta!, _stretchDragController!.controller);
+    _dragGestureController!.dragUpdate(
+      // Divide by size of the sheet.
+      details.primaryDelta! / sheetHeight,
```
Piinks: Do we need to check for sheetHeight == 0?

examples/api/lib/cupertino/sheet/cupertino_sheet.3.dart
Piinks: > I believe in that case then a scroll would be triggered on a drag upwards originating in the pinned sliver, instead of the stretch animation.

Did you try it?

We have control over how the scrolling offset is applied here with the custom scroll position class in the scrollable sheet, could we divert it in this case?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -720,12 +832,18 @@ class _CupertinoDragGestureDetectorState<T> extends State<_CupertinoDragGestureD
   late VerticalDragGestureRecognizer _recognizer;
   _StretchDragControllerProvider? _stretchDragController;
 
+  static VelocityTracker _cupertinoVelocityBuilder(PointerEvent event) =>
+      IOSScrollViewFlingVelocityTracker(event.kind);
+
+  double get sheetHeight => widget.sheetHeight ?? context.size!.height;
```
Piinks: We should make sure the fallback is safe to access, has error message.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -777,7 +899,7 @@ class _CupertinoDragGestureDetectorState<T> extends State<_CupertinoDragGestureD
       return;
     }
     _dragGestureController!.dragEnd(
-      details.velocity.pixelsPerSecond.dy / context.size!.height,
```
Piinks, line 906: What about this change requires that sheet height has to be passed all the way down here instead of continuing to use context.size!.height?

Piinks (commented):
It would be best to update the design document about this change. In particular, some diagrams illustrating how gestures come in and are handled would be ideal. There are a lot of classes working together here, but it looks like there is some unfortunate leaking of information across boundaries we typically try to avoid. Updating the design document will help a lot with to see the whole picture here.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -545,29 +570,90 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
+        onStartPopGesture: () => _CupertinoSheetRouteTransitionMixin._startPopGesture<T>(this),
+        builder: scrollableBuilder!,
+      );
 
   @override
   final bool enableDrag;
 
   @override
   Widget buildContent(BuildContext context) {
+    final double sheetHeight =
```
Piinks: There should probably be a call to debugCheckHasMediaQuery before accessing it.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -607,7 +693,16 @@ class CupertinoSheetRoute<T> extends PageRoute<T> with _CupertinoSheetRouteTrans
...
+  const _CupertinoSheetScope({
+    required super.child,
+    required this.route,
+    required this.sheetHeight,
+  });
+
+  // The height of the sheet when fully open, and not stretched upwards.
+  final double sheetHeight;
+
+  final CupertinoSheetRoute<dynamic> route;
```
Piinks: Maybe something for the design doc to explain, if the CupertinoSheetRoute's role is to `see if another sheet is in the tree already`, why does it need the sheetHeight and the route itself?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -922,3 +1045,259 @@ class _CupertinoDragGestureController<T> {
...
+      oldPosition: oldPosition,
+      onDragStart: onDragStart,
+      onDragUpdate: onDragUpdate,
+      onDragEnd: onDragEnd,
+      sheetIsDraggedDown: sheetIsDraggedDown,
+    );
+  }
+}
+
+/// A scroll position that manages scroll activities for
```
Piinks, line 1187: These docs looks like leftover copy paste from the widgets class. 
These should be updated to reflect the right context. To avoid confusion, it might be best to give this class a unique name.

Piinks: Also something for the doc - since this was copied from the widgets layer and then changed to suit the use case, the document should talk about what is different and why.

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: With other routes, `builder` has only the context in the argument, but `pageBuilder` has context and two animations. `pageBuilder` was originally used incorrectly as the name here, as I was looking at similar routes that used `pageBuilder` for patterns and didn't pick up on the difference in arguments.

examples/api/lib/cupertino/sheet/cupertino_sheet.2.dart
```diff
@@ -71,7 +71,7 @@ class _RestorableSheetState extends State<RestorableSheet>
     Object? arguments,
   ) {
     return CupertinoSheetRoute<int?>(
-      builder: (BuildContext context) {
+      scrollableBuilder: (BuildContext context, ScrollController _) {
```
justinmc: Nit: No underscores.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -154,59 +157,82 @@ final Animatable<double> _kScaleTween = Tween<double>(begin: 1.0, end: 1.0 - _kS
...
+    'Use scrollableBuilder instead. '
     'This feature was deprecated after v3.33.0-0.2.pre.',
   )
   WidgetBuilder? pageBuilder,
+  @Deprecated(
+    'Use scrollableBuilder instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
   WidgetBuilder? builder,
+  ScrollableWidgetBuilder? scrollableBuilder,
```
justinmc, line 185: The name scrollableBuilder makes me think it's just supposed to build a Scrollable and the actual page content should be built elsewhere. I guess I don't have an idea for a better name, though. scrollablePageBuilder is probably too long.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -565,24 +595,61 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
-  final WidgetBuilder builder;
+  @Deprecated(
+    'Use scrollableBuilder instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
+  final WidgetBuilder? builder;
+
+  /// Builds the primary contents of the sheet route with a provided [ScrollController].
+  ///
+  /// Use this instead of [builder] when the contents of the sheet need to both
```
justinmc: builder is deprecated, so shouldn't people always use scrollableBuilder?

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -565,24 +595,61 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
+  WidgetBuilder get _effectiveBuilder =>
+      builder ??
+      (BuildContext context) => _CupertinoDraggableScrollableSheet<T>(
+        enabledCallback: () => enableDrag,
+        onStartPopGesture: () =>
+            _CupertinoSheetRouteTransitionMixin._startPopGesture<T>(this, topGap),
+        builder:
+            scrollableBuilder ??
+            (BuildContext context, ScrollController controller) => builder!(context),
+      );
```
justinmc: Nit: Consider breaking _effectiveBuilder into a normal function instead of a `=>` function for readability.

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -551,6 +577,10 @@ class _StretchDragControllerProvider extends InheritedWidget {
 /// The sheet will be dismissed by dragging downwards on the screen, or a call to
 /// [CupertinoSheetRoute.popSheet].
 ///
+/// Any scrollable content that uses the [ScrollController] provided by
+/// `scrollableBuilder` will trigger the drag to dismiss behavior when a downward
+/// gesture happens on the scrollable area when the content is already at the top.
```
justinmc: I think you should explain the common case here more thoroughly (and also in the docs for scrollableBuilder below). I think users are going to wonder what the scrollController parameter is for and they're probably going to look at these docs or the docs for scrollableBuilder. So maybe add something like:

"
Any time a CupertinoSheetRoute contains a large scrollable that might conflict with the dismiss drag gesture, pass the provided [ScrollController] to the scrollable. Otherwise, this parameter can be ignored. See below for an example.
"

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -565,24 +595,61 @@ class _StretchDragControllerProvider extends InheritedWidget {
...
+    'Use scrollableBuilder instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
+  final WidgetBuilder? builder;
+
+  /// Builds the primary contents of the sheet route with a provided [ScrollController].
+  ///
+  /// Use this instead of [builder] when the contents of the sheet need to both
+  /// scroll in the Y-axis, while still being able to drag the sheet downwards to
+  /// pop the route.
```
justinmc: Also, I would elaborate here in the same way as I mentioned above. And maybe include the example here as well, or mention that there is an example in [CupertinoSheetRoute].

justinmc (commented):
I'm on board with this approach. Passing the ScrollController to the builder seems straightforward, consistent with Flutter's patterns, and discoverable. Thanks for responding to all of my alternative ideas in my previous review.

What if a dev wants to use the ScrollController that is passed in, but they also want to customize the ScrollController? Like by listening to it or setting an initialOffset or onAttach etc.

I think @Piinks has a few outstanding comments with no answers but otherwise this looks good.

FYI there is a Google testing failure, but it appears to just be a merge conflict.

Piinks:
> What if a dev wants to use the ScrollController that is passed in, but they also want to customize the ScrollController?

I think it would break the sheet features it manages. The same would happen for DraggableScrollableSheet. Developers have to use that specific subclass of ScrollController in order for everything to work.

packages/flutter/lib/fix_data/fix_cupertino.yaml
```diff
@@ -500,4 +500,56 @@ transforms:
...
+        changes:
+          - kind: 'addParameter'
+            index: 3
+            name: 'scrollableBuilder'
+            style: optional_named
+            argumentValue:
+              expression: "(BuildContext context, ScrollController controller) {\n
+                          \     WidgetBuilder widgetBuilder = {% builder %};\n
+                          \     return widgetBuilder(context);\n
+                          \   }"
```
MitchellGoodwin, line 521: @Piinks do you know if there's a better way to do this? Ideally I'd like to just get the function body and put it with the added arguments, but I couldn't figure out how to do that or if it is possible.

So I'd like to turn a `builder` that looks like this
```dart
(BuildContext context) {
  return Container();
}
```

directly into
```dart
(BuildContext context, ScrollController controller) {
  return Container();
}
```

packages/flutter/lib/src/cupertino/sheet.dart
MitchellGoodwin: I also can't think of a better name, and I think it's better for it to be on the shorter side.

packages/flutter/lib/fix_data/fix_cupertino.yaml
```diff
@@ -500,4 +500,56 @@ transforms:
       opacity:
         kind: 'fragment'
         value: 'arguments[0]'
+
```
Piinks, line 503: Dropping comment to link to for reference.

justinmc (approved):
LGTM :+1: 

Hopefully we can resolve your dartdoc question in https://github.com/flutter/flutter/pull/177337#discussion_r2695330751 but I don't know the answer myself :)

packages/flutter/lib/fix_data/fix_cupertino.yaml
Piinks: We heard back from the folks who know dart fix best, @MitchellGoodwin do we want to make changes here based on that? Or is this ok to land given the current limitations?

MitchellGoodwin: @Piinks I think this is ok to land. It sounds like this is the best we can do at the moment, and it does what it's supposed to in smoothing over the transition for developers.

packages/flutter/lib/fix_data/fix_cupertino.yaml
MitchellGoodwin: It sounds like there is not a prettier way to do this at the moment, but this does what it's supposed to. The negative of the current state is that the code produced by the dart fix is a line longer than it ideally would be.

Piinks (approved):
LGTM as well

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -94,9 +94,12 @@ final Animatable<double> _kScaleTween = Tween<double>(begin: 1.0, end: 1.0 - _kS
...
 ///
-/// This is a convenience method for displaying [CupertinoSheetRoute] for common,
-/// straightforward use cases. The Widget returned from `pageBuilder` will be
-/// used to display the content on the [CupertinoSheetRoute].
+/// This is a convenience method for displaying [CupertinoSheetRoute] for most
+/// use cases. The Widget returned from `scrollableBuilder` will be used to display
+/// the content on the [CupertinoSheetRoute]. If the content of the sheet has a
+/// scrollable view, the [ScrollController] provided by `scrollableBuilder` can be
+/// used to enable the drag-to-dimiss gesture to work with the scrolling of the
+/// content. See [CupertinoSheetRoute] for an example.
```
chunhtai:
```suggestion
/// content. See [CupertinoSheetRoute.scrollableBuilder] for an example.
```

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -197,26 +216,34 @@ Future<T?> showCupertinoSheet<T>({
...
-      builder: widgetBuilder,
-      enableDrag: enableDrag,
-      settings: settings,
-      showDragHandle: showDragHandle,
-      topGap: topGap,
-    ),
-  );
+    final route = effectiveBuilder != null
+        ? CupertinoSheetRoute<T>(
+            builder: (BuildContext context) => nestedNavigationContent(effectiveBuilder),
```
chunhtai: since we are deprecating the builder, we should both use `scrollableBuilder`

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -1050,3 +1144,218 @@ class _CupertinoDragGestureController<T> {
...
+    required this.onDragUpdate,
+    required this.onDragEnd,
+    required this.sheetIsDraggedDown,
+  });
+
+  VoidCallback? _dragCancelCallback;
+  final Set<AnimationController> _ballisticControllers = <AnimationController>{};
+  bool get listShouldScroll => pixels > 0.0;
+
+  final void Function() onDragStart;
```
chunhtai: typdef these

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -157,27 +160,43 @@ final Animatable<double> _kScaleTween = Tween<double>(begin: 1.0, end: 1.0 - _kS
...
+    'Use scrollableBuilder instead. '
     'This feature was deprecated after v3.33.0-0.2.pre.',
   )
   WidgetBuilder? pageBuilder,
+  @Deprecated(
+    'Use scrollableBuilder instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
   WidgetBuilder? builder,
+  ScrollableWidgetBuilder? scrollableBuilder,
```
chunhtai, line 185: Can you also update the doc in ScrollableWidgetBuilder? the doc right now seems to be specific for DraggableScrollableSheet which is confusing. Since we can't mention cupertino in that doc, maybe make it more general for all cases?

chunhtai (approved):

PR #177342, opened 2025-10-21
Correct editable text and placeholder position in baseline aligned stack

### LINKED ISSUE #176817: CupertinoTextField text not vertically centered when fontSize differs between style and placeholderStyle.
### Steps to reproduce

When using different fontSize values for style and placeholderStyle in CupertinoTextField,
the rendered text appears visually lower (not centered) even though textAlignVertical is set to TextAlignVertical.center.

This issue occurs since Flutter 3.35.x, after PR [#166952](https://github.com/flutter/flutter/pull/166952) introduced _BaselineAlignedStack for baseline alignment between placeholder and editable text.
The new baseline-based layout causes the entire text area to appear visually shifted down when the two font sizes differ.

-------

1. Create a simple CupertinoTextField with different font sizes for text and placeholder:
```dart
CupertinoTextField(
  style: const TextStyle(fontSize: 20, height: 1.4),
  placeholder: 'Enter text',
  placeholderStyle: const TextStyle(fontSize: 16, height: 1.4, color: CupertinoColors.systemGrey),
  textAlignVertical: TextAlignVertical.center,
)
```
2. Run on iOS simulator (or macOS with Cupertino style).
3. Observe that the text is rendered slightly below the visual center of the input box.

### Expected results

The text (and placeholder) should appear visually centered within the field when textAlignVertical is center, regardless of font size differences.

Image

### Actual results

The text is baseline-aligned, making smaller placeholder fonts appear too low, and larger text fonts look offset downward relative to the center.

Image

### Code sample

```dart
CupertinoTextField(
  style: const TextStyle(fontSize: 20, height: 1.4),
  placeholder: 'Enter text',
  placeholderStyle: const TextStyle(fontSize: 16, height: 1.4, color: CupertinoColors.systemGrey),
  textAlignVertical: TextAlignVertical.center,
)
```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.5, on macOS 26.0 25A354 darwin-arm64, locale zh-Hans-CN) [393ms]
    • Flutter version 3.35.5 on channel stable at /Users/liwenhui/.fvm/versions/3.35.5
    • Framework revision ac4e799d23 (13 days ago), 2025-09-26 12:05:09 -0700
    • Engine revision d3d45dcf25
    • Dart version 3.9.2
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [2.2s]
    • Android SDK at /Users/liwenhui/Library/Android/sdk
    • Emulator version 36.2.11.0 (build_id 14212360) (CL:N/A)
    • Platform android-36, build-tools 36.1.0
    • ANDROID_HOME = /Users/liwenhui/Library/Android/sdk
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14018985-b1038.68)

[✓] Xcode - develop for iOS and macOS (Xcode 26.0.1) [910ms]
    • Build 17A400
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [73ms]

[✓] Android Studio (version 2025.1) [72ms]
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14018985-b1038.68)

[✓] IntelliJ IDEA Ultimate Edition (version 2025.1) [72ms]
    • IntelliJ at /Applications/IntelliJ IDEA.app
    • Flutter plugin version 85.2.4
    • Dart plugin version 251.25267.1

[✓] VS Code (version 1.101.2) [5ms]
    • Flutter extension can be installed from:
      🔨 https://marketplace.visualstudio.com/items?itemName=Dart-Code.flutter

[✓] Connected device (0 available) [0.2s]

[✓] Network resources [6.0s]

• No issues found!
```

</details>

victorsanni:
This PR makes CupertinoTextField's placeholder and editable text position respect `CupertinoTextField.textAlignVertical`.

<details open>
<summary>Image comparison</summary>

| | Before | After | 
| --- | --- | --- |
| top  | top b4 |  top |
| center  | top b4 |  center |
| bottom  | top b4 | bottom |

</details>

<details>
<summary>Sample code</summary>

```dart

void main() => runApp(const TextFieldApp());

class TextFieldApp extends StatelessWidget {
  const TextFieldApp({super.key});

  @override
    return const MaterialApp(home: HomePage());
  }
}

class HomePage extends StatefulWidget {
  const HomePage({super.key});

  @override
  State<HomePage> createState() => _HomePageState();
}

class _HomePageState extends State<HomePage> {
  TextEditingController commentController = TextEditingController();
  @override
    return Scaffold(
      backgroundColor: Colors.grey,
      body: SafeArea(
        child: Center(
          child: Padding(
            child: CupertinoTextField(
              style: const TextStyle(fontSize: 40),
              placeholderStyle: const TextStyle(
                fontSize: 80,
                color: CupertinoColors.inactiveGray,
              ),
              placeholder: 'Enter text',
              textAlignVertical: TextAlignVertical.bottom,
            ),
          ),
        ),
      ),
    );
  }
}

```

</details>

Fixes [CupertinoTextField text not vertically centered when fontSize differs between style and placeholderStyle.](https://github.com/flutter/flutter/issues/176817)


packages/flutter/lib/src/cupertino/text_field.dart
```diff
@@ -1392,6 +1392,7 @@ class _CupertinoTextFieldState extends State<CupertinoTextField>
                 child: _BaselineAlignedStack(
                   placeholder: placeholder,
                   editableText: editableText,
+                  textAlignVertical: _textAlignVertical,
```
justinmc, line 1395: Ok so previously the placeholder and the EditableText were aligned by baseline. However, when they have different font sizes, this might look wrong (as seen in the screenshots in the description of https://github.com/flutter/flutter/issues/176817). Now with this PR it will default to center, which looks correct when the font sizes differ, and otherwise it's configurable with the textAlignVertical parameter.

Am I understanding correctly?

justinmc (approved):
LGTM 👍 

I guess it's possible this will break visual diff tests, so let's see what the Google tests say.

packages/flutter/lib/src/cupertino/text_field.dart
victorsanni: Partly. 

This PR doesn't change the default. The wrong look happened because the position of the placeholder and editable text was slightly wrong (wrong calculations) and also wasn't configurable (even though the property `textAlignVertical` existed). 

`textAlignVertical` originally wrapped the text dependent attachments (placeholder, editable text etc.) in an `Align`. My understanding (but might be wrong) is in writing our own renderbox, we now arbitrarily fixed the position of the placeholder + editable text in the entire space the `Align` contained, making the `Align` uneffective. 

This PR basically copies over the alignment into the renderbox positioning so it is respected.

packages/flutter/lib/src/cupertino/text_field.dart
```diff
@@ -1698,11 +1699,13 @@ class _BaselineAlignedStack
     required this.editableTextBaseline,
     required this.placeholderBaseline,
     required this.editableText,
+    required this.textAlignVertical,
```
loic-sharma, line 1701: Should we make `textAlignVertical` before the `editableText` parameter?

It feels a little weird that `textAlignVertical` is in between `editableText` and `placeholder`. It feels like we're "interrupting" the section of children arguments, if that makes sense.

This would also match the order used by the fields below.

loic-sharma (approved):
LGTM

packages/flutter/lib/src/cupertino/text_field.dart
```diff
@@ -1392,6 +1392,7 @@ class _CupertinoTextFieldState extends State<CupertinoTextField>
                 child: _BaselineAlignedStack(
```
loic-sharma, line 1392: This seems like a useful widget. Should we add a generic version of this widget to the widgets library? I imagine other design systems would want to use this.

This isn't a blocker for landing this PR. Just an idea for future improvement :)

victorsanni: I'm wondering if we can just do this by adding a flag to `Stack`. The problem would be knowing which child's baseline should be the reference baseline. 

I do think this is very doable though. Off the top of my head I think the API could have two main properties: a child with a reference baseline and children for all the other widgets that should be aligned to that baseline. Although if we know the [baseline distance](https://api.flutter.dev/flutter/rendering/RenderBox/getDistanceToBaseline.html) for the reference child's renderbox, we can use that number to align other children in a `Stack` by wrapping each child in a [Baseline](https://api.flutter.dev/flutter/widgets/Baseline-class.html). 

But getting the reference child's renderbox outside layout might be problematic (using a post frame callback can mitigate this, but that has its problems as well). So it's a good argument for writing a custom renderobject for this, but (correct me if I'm wrong here) that will mean we will have to manually position the children instead of relying on `Baseline` to simplify the work.

loic-sharma: Maybe we could do something like this:

```dart
BaselineStack(
  alignment: Alignment.top,
  children: [
    child1,
    BaselineReference(child: child2),
    child3,
  ],
),
```

`BaselineReference` must be the child of a `BaselineStack` (similar to how `Flexible` and `Expanded` must be the direct child of a `Row`/`Colum`/`Flex`)

victorsanni: I've always been confused by the `Row`/`Column`/`Flex` and `Flexible`/`Expanded` API because forcing a widget to only have a specific parent/child relationship introduces additional cognitive complexity. 

Since all we need here is a reference baseline (so no `Positioned` widget as in `Stack`), I'm strongly inclined to an API that separates the child with the reference baseline from the other children. And this API means we can require the reference child but not necessarily the other children.

I do think we need some sort of intermediary configuration class that wraps both the reference child and other children. This is so that the [TextBaseline](https://api.flutter.dev/flutter/dart-ui/TextBaseline.html) for each child can be provided. Or is there an alternative way to provide the `TextBaseline` for each child?

loic-sharma: > Since all we need here is a reference baseline (so no Positioned widget as in Stack), I'm strongly inclined to an API that separates the child with the reference baseline from the other children. And this API means we can require the reference child but not necessarily the other children.

That's fair. Another alternative might be:

```dart
BaselineStack(
  alignment: Alignment.top,
  baselineReferenceIndex: 1,
  children: [
    child1,
    child2,
    child3,
  ],
),
```

PR #177343, opened 2025-10-21
[web] Use SkPathBuilder because SkPath is becoming immutable

mdebbar:
Skia is working on making `SkPath` immutable: https://skia-review.googlesource.com/c/skia/+/1075478

In Flutter Web's CanvasKit renderer, we use `SkPath` as a mutable object, which made the [Skia roll](https://github.com/flutter/flutter/pull/177184) fail. To fix this, we should start using `SkPathBuilder` instead.

Remaining work:
- [x] Figure out the deletion/disposal of `SkPath`s generated from `.snapshot()` calls.
- [ ] `LazyPath` should be restructured to better accommodate a world of immutable paths and path builders (coming in a future PR).


eyebrowsoffire (approved):
LGTM!

PR #177352, opened 2025-10-21
Change the root path of the license crawl to engine/src

jason-simmons:
Previously the license script looked for licenses within the engine/src/flutter tree.  This PR updates the script to support the move of Abseil and the Fuchsia SDK from engine/src/flutter to engine/src/third_party.

See https://github.com/flutter/flutter/pull/177059 and https://github.com/flutter/flutter/pull/177118


engine/src/flutter/tools/licenses_cpp/src/license_checker.cc
```diff
@@ -37,6 +37,8 @@ const std::array<std::string_view, 9> kLicenseFileNames = {
 const std::array<std::string_view, 2> kThirdPartyIgnore = {"pkg",
                                                            "vulkan-deps"};
 
+const char* kRootPackageName = "flutter";
```
gaaclarke: Can you pipe this through to `LicenseChecker::Flags`, please?  One of the big problems with the last license checker was that as people patched it they co-mingled data and logic piecemeal which made it hard to focus on fixing the actual logic.  There is already an example there for you.

engine/src/flutter/tools/licenses_cpp/data/exclude.txt
```diff
@@ -8,112 +8,118 @@
...
-^third_party/vulkan-deps/vulkan-headers/src/LICENSE\.md$
-^third_party/vulkan-deps/vulkan-loader/.*
-^third_party/vulkan-deps/vulkan-utility-libraries/src/LICENSE.md$
-^third_party/vulkan-deps/vulkan-utility-libraries/src/scripts/.*
-^tools/android_sdk/packages\.txt$
-^tools/fuchsia/.*
-^tools/licenses_cpp/data/.*
+^build_overrides/.*
+^fuchsia/.*
+^out/.*
```
gaaclarke, line 13: no action required: Unfortunately the way the license checker is designed is to be auditable.  So even though we don't parse any of these files, just looking at them causes a cost in disk access and logs.  This is going to make the logs much harder to audit.

gaaclarke (commented):
The code looks good (modulo the hardcoded name).  I'm a bit concerned with the approach because this is going to muddle up the audit logs.  I thought about this a bit more.  Here are some alternatives:
1) Can you just add a symbolic link between the new location and the old location?  Maybe the checker would have to be updated a bit to make sure that works.
1) We could make our own symbolic link structure in //engine/src/flutter/tools/licenses_cpp/data/links that maps the 2 directories.  Something like text files with 2 lines that specify the links source and destination?

The benefit there is that the folder we are scanning is more in line with what someone would have expected and we don't have to parse through tons of excluded files are come up with yet another way to exclude those files from being scanned.

engine/src/flutter/tools/licenses_cpp/src/license_checker.cc
jason-simmons: done - moved the root package name to `LicenseChecker::Flags`

engine/src/flutter/tools/licenses_cpp/data/exclude.txt
jason-simmons: I added these filters out of caution to ensure that those directories would never be crawled.  But from what I can tell they are not really necessary.

If I delete those three lines from `exclude.txt`, then when I run the license script it still produces the correct output.  I did not see anything in the license checker's logs that mentions paths within `out`, `fuchsia`, or `build_overrides`.

Should I remove these lines from `exclude.txt`?

gaaclarke: Here's the output from above: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8700357321091203633/+/u/test:_Run_license_cpp/stdout

It didn't add a bunch of `EXCLUDE: "out/` examples.  I forgot that's because on CI we put our build outputs elsewhere.  It did add a fair bit of `EXCLUDE: "third_party/fuchsia-sdk` but the execution time is the same.

> Should I remove these lines from exclude.txt?

No, I think it's prudent to keep them.  It won't affect CI runs but it will affect local runs.

engine/src/flutter/tools/licenses_cpp/src/license_checker.h
```diff
@@ -15,6 +15,7 @@ class LicenseChecker {
  public:
   struct Flags {
     bool treat_unmatched_comments_as_errors = false;
+    std::string root_package_name = "flutter";
```
gaaclarke: Can you pipe this up to the command line flags please?  That way when it's executed it looks like this:

`licenses_cpp --path foo/bar --root-package-name=bar`

Can you measure the time it takes to run the license checker locally before your change and after?  I'm interested to see what the exclusion of the `out/` directory is going to do to runtime (and the noise in the logs).  If it is reasonable I'm fine with this approach for now and leave it as a potential cleanup in the future to allow out-of-tree inclusions instead of dropping the scanned directory one folder higher.  I think that is the better solution.

gaaclarke: I'd make the root-package-name flag optional, where not supplying one keeps the old logic.

jason-simmons: Added a command line flag.

This PR does not significantly change the running time of the license checker.

Results for `time ci/licenses_cpp.sh host_release` on my Linux setup:

| With this PR | Without this PR |
| ----- | ----- |
| real    15.611s | real    14.943s |
| user    8.646s | user    8.693s |
| sys     5.779s | sys     5.464s |

engine/src/flutter/tools/licenses_cpp/src/license_checker.cc
```diff
@@ -159,9 +159,13 @@ std::string GetDirFilename(const fs::path& working_dir) {
 
 Package GetPackage(const Data& data,
                    const fs::path& working_dir,
-                   const fs::path& relative_path) {
+                   const fs::path& relative_path,
+                   const LicenseChecker::Flags& flags) {
+  std::string root_package_name = flags.root_package_name.has_value()
+                                      ? flags.root_package_name.value()
+                                      : GetDirFilename(working_dir);
```
gaaclarke, line 166: nit: `flags.root_package_name.value_or(GetDirFilename(working_dir))`

gaaclarke (approved):
lgtm, thanks jason

engine/src/flutter/tools/licenses_cpp/src/license_checker.cc
jason-simmons: I wanted to avoid calling `GetDirFilename` in the common case where the root package name has a value.  (The engine's current C++20 toolchain does not provide `std::optional::or_else`)

gaaclarke: I thought they were doing metaprogramming magic to only evaluate where there is no value.  That sucks.  `or_else` looks like more boilerplate than just using the ternary operator.  TIL.

PR #177404, opened 2025-10-22
Add tooltip windows to the windowing API alongside the window positioning logic

mattkae:
## What's new?
- Added tooltips to the windowing API, but they have no implementation for the time being
- Added the `WindowPostiioner` logic to the windowing API
- Wrote a lot of tests :test_tube: 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+///       child: AnimatedContainer(
+///         duration: const Duration(milliseconds: 200),
+///         color: _tooltipController != null ? Colors.blueAccent : Colors.blue,
+///         padding: const EdgeInsets.all(12),
+///         child: Row(children: children),
+///       ),
+///     );
+///   }
+///
+///   Rect? _getAnchorRect() {
```
loic-sharma: @justinmc do you know if there is a better way to do this? This seems hacky and not something we want customers to do. My understanding is this positions the tooltip on the text, but I expect this might not work as expected if the text hasn't been laid out yet, or if the text's position will be updated this frame.

loic-sharma: I see that `Tooltip` uses [`OverlayPortal.overlayChildLayoutBuilder`](https://github.com/flutter/flutter/blob/3c04c999c1cf59f2ac8db8d03388ce5dcfb60f4e/packages/flutter/lib/src/material/tooltip.dart#L1046-L1050), which provides the necessary layout information [here](https://api.flutter.dev/flutter/widgets/OverlayChildLayoutInfo-extension-type.html). 

We shouldn't need a `OverlayPortal` for this tooltip windows, but I suspect we want to reuse whatever layout magic `OverlayPortal` does to get the child's size/position.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+///   }
+/// }
+/// ```
+/// {@end-tool}
+///
+/// Children of a [TooltipWindow] widget can access the [TooltipWindowController]
+/// via the [WindowScope] inherited widget.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+abstract class TooltipWindowController extends BaseWindowController {
```
loic-sharma, line 695: I wonder if we should make this [`sealed`](https://dart.dev/language/class-modifiers#sealed) instead of `abstract`. This would prevent folks from extending/implementing `TooltipWindowController` outside of the widgets library.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+///
+/// {@tool snippet}
+/// An example usage might look like:
+///
+/// ```dart
+/// // TODO(mattkae): remove invalid_use_of_internal_member ignore comment when this API is stable.
+/// // ignore_for_file: invalid_use_of_internal_member
+/// import 'package:flutter/material.dart';
+/// import 'package:flutter/src/widgets/_window.dart';
+/// import 'package:flutter/src/widgets/_window_positioner.dart';
```
loic-sharma: We'll want to clean this sample up before we make these APIs public. However, I suspect we'll likely forget this.

What's the best way to remember? What do you think of filing an issue that's a checklist of things we need to do before we make windowing APIs public, and then linking to that issue here?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+  /// Users who want to instantiate a new [TooltipWindowController] should
+  /// always use the factory method to create a controller that is valid
+  /// for their particular platform.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  @protected
+  TooltipWindowController.empty();
+
+  /// The parent controller of this tooltip.
```
loic-sharma, line 748: Consider adding a quick comment that the tooltip will be destroyed if this parent is destroyed.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+/// tooltip window.
+///
+/// The user of this class is responsible for managing the lifecycle of the window.
+/// When the window is no longer needed, the user should call [destroy] on this
+/// controller to release the resources associated with the window.
+///
+/// {@tool snippet}
+/// An example usage might look like:
+///
+/// ```dart
```
loic-sharma: @justinmc This is a large sample, what do you think of moving this to switching to a `@tool dartpad` that points to something like `examples/api/lib/widgets/windows/tooltip.0.dart`?

If we do this, some things we need to figure out:

1. The `api` example app will create two top level windows (one from Dart, one from native entry point). Is this OK?
2. How do we properly call out that this is a sample that shows off experimental APIs? I suspect we'll want some sort of big disclaimer comment at the top.
3. How do we properly explain that this sample is only runnable on the main channel with the windowing feature flag? I suspect we might want to assert this in the `main` method.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+/// Delegate class for tooltip window controller.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+/// * [TooltipWindowController], the controller that creates and manages tooltip windows.
+/// * [TooltipWindow], the widget for a tooltip window.
+/// * [RegularWindowControllerDelegate], the delegate for regular window controllers.
+mixin class TooltipWindowControllerDelegate {
```
loic-sharma, line 633: Should we update `_window_macos.dar` now that that exists?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+/// if no edge is specified.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+enum WindowPositionerAnchor {
+  /// If the [WindowPositioner.parentAnchor] is set to [center], then the
+  /// child window will be positioned relative to the center
+  /// of the parent window.
+  ///
+  /// If [WindowPositioner.childAnchor] is set to  [center], then the middle
```
loic-sharma:
```suggestion
  /// If [WindowPositioner.childAnchor] is set to [center], then the middle
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+/// The specified anchor is used to derive an anchor point on the anchor rectangle that
+/// the child [BaseWindowController] will be positioned relative to.
+/// If a corner anchor is set (e.g. [topLeft] or [bottomRight]),
+/// the anchor point will be at the specified corner; otherwise, the derived anchor point
+/// will be centered on the specified edge, or in the center of the anchor rectangle
+/// if no edge is specified.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+enum WindowPositionerAnchor {
```
loic-sharma, line 21: I wonder if we should use [`Alignment`](https://api.flutter.dev/flutter/painting/Alignment-class.html) instead since it's so similar?

One huge caveat: even though most people use `Alignment` like an `enum`, you can also do stuff like `Alignment(0.3, 0.7)`. But I suspect that's OK?

packages/flutter/lib/src/widgets/_window_positioner.dart
loic-sharma: FYI it looks like this double-space problem is on the other comments below as well :)

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,238 @@ abstract class DialogWindowController extends BaseWindowController {
...
+  ///
+  /// The [delegate] argument can be used to listen to the window's
+  /// lifecycle. For example, it can be used to save state before
+  /// a window is closed.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  factory TooltipWindowController({
+    required BaseWindowController parent,
+    required Rect anchorRect,
+    required WindowPositioner positioner,
```
loic-sharma, line 718: Could/should this be optional? While I expect most people will want to provide a `positioner`, could we have a sensible default? I don't feel strongly, just seeing what people think.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+  /// [WindowPositionerAnchor.bottomRight]), the anchor point will be at the specified corner;
+  /// otherwise, the derived anchor point will be centered on the specified
+  /// edge, or in the center of the anchor rectangle if no edge is specified.
+  ///
+  /// This is used in conjunction with [parentAnchor] to determine the final position
+  /// of the child window.
+  final WindowPositionerAnchor childAnchor;
+
+  /// Specify the window position offset relative to the position of the
+  /// anchor on the anchor rectangle and the anchor on the child.\
```
loic-sharma:
```suggestion
  /// anchor on the anchor rectangle and the anchor on the child.
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+  /// The offset position of the window is the one used for constraint testing.
+  /// See [constraintAdjustment].
+  ///
+  /// An example use case is placing a popup menu on top of a user interface
+  /// element, while aligning the user interface element of the parent window
+  /// with some user interface element placed somewhere in the popup window.
+  final Offset offset;
+
+  /// Defines how Flutter will adjust the position of the window if the unadjusted
+  /// position would result in the window being partly constrained by the platform..
```
loic-sharma:
```suggestion
  /// position would result in the window being partly constrained by the platform.
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+      }
+    }
+
+    return defaultResult;
+  }
+}
+
+extension on WindowPositionerAnchor {
+  WindowPositionerAnchor flipX() {
+    switch (this) {
```
loic-sharma: Consider a switch expression here. Something like:

```dart
return switch (this) {
  WindowPositionerAnchor.center => WindowPositionerAnchor.center,
  ...
};
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+      if (_rectContains(outputRect, result & childSize)) {
+        return result & childSize;
+      }
+    }
+
+    return defaultResult;
+  }
+}
+
+extension on WindowPositionerAnchor {
```
loic-sharma: If we keep the `WindowPositionerAnchor` enum, could we move these to be methods on the enum? See: https://dart.dev/language/enums#declaring-enhanced-enums

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+        return rect.bottomRight;
+    }
+  }
+}
+
+bool _rectContains(Rect r1, Rect r2) {
+  return r1.left <= r2.left && r1.right >= r2.right && r1.top <= r2.top && r1.bottom >= r2.bottom;
+}
+
+Offset _constraintTo(Rect r, Offset p) {
```
loic-sharma: Is this a typo?

```suggestion
Offset _constrainTo(Rect r, Offset p) {
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+    this.offset = Offset.zero,
+    this.constraintAdjustment = const <WindowPositionerConstraintAdjustment>{},
+  });
+
+  /// Copy a [WindowPositioner] with some fields replaced.
+  WindowPositioner copyWith({
+    WindowPositionerAnchor? parentAnchor,
+    WindowPositionerAnchor? childAnchor,
+    Offset? offset,
+    Set<WindowPositionerConstraintAdjustment>? constraintAdjustment,
```
loic-sharma: Instead of making this a `Set<WindowPositionerConstraintAdjustment>`, what do you think of making `WindowPositionerConstraintAdjustment` a class with boolean members. Something like:

```dart
class WindowPositionerConstraintAdjustments {
  final bool flipX;
  final bool flipY;
}

File: packages/flutter/lib/src/widgets/_window_positioner.dart [hunk 16]
```diff
@@ -0,0 +1,546 @@
...
+  /// Whether a window is considered "constrained" is left to the platform
+  /// to determine. For example, the window may be partly outside the
+  /// output's 'work area', thus necessitating the child window's
+  /// position be adjusted until it is entirely inside the work area.
+  ///
+  /// The adjustments can be combined, according to a defined precedence:
+  ///
+  /// 1. [WindowPositionerConstraintAdjustment.flipX] and [WindowPositionerConstraintAdjustment.flipY]
+  /// 2. [WindowPositionerConstraintAdjustment.slideX] and [WindowPositionerConstraintAdjustment.slideY]
+  /// 3. [WindowPositionerConstraintAdjustment.resizeX] and [WindowPositionerConstraintAdjustment.resizeY]
```
loic-sharma: It wasn't clear to me how the adjustments are combined until I read the implementation. I'd consider explaining that it picks the first valid adjustment. Maybe something like:

```suggestion
  /// 3. [WindowPositionerConstraintAdjustment.resizeX] and [WindowPositionerConstraintAdjustment.resizeY]
  ///
  /// The first adjustment that results in the child window
  /// being entirely inside the work area will be picked.
```

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,546 @@
...
+  /// Computes the screen-space rectangle for a child window placed according to
+  /// this [WindowPositioner].
+  ///
+  /// [childSize] is the frame size of the child window.
+  ///
+  /// [anchorRect] is the rectangle relative to which the child window is placed.
+  ///
+  /// [parentRect] is the parent window's rectangle.
+  ///
+  /// [outputRect] is the output display area where the child window will be placed.
```
loic-sharma: Maybe consider calling this `displayRect`? I think it'd be a little clearer what this is without reading the docs.

packages/flutter/lib/src/widgets/_window.dart
mattkae: Yeah if there's any guidance, I'd appreciate it!

packages/flutter/lib/src/widgets/_window.dart
mattkae: Is that a restriction that we need? I feel like it's valid to provide your own implementation if you want to support some niche platform

packages/flutter/lib/src/widgets/_window.dart
mattkae: https://github.com/flutter/flutter/issues/177586 :tada:

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: This feels a bit more imprecise than what we were going for, so I will defer to @Saviq to see if he has any opinions one way or the other.

packages/flutter/lib/src/widgets/_window.dart
mattkae: The plan was for the "default" to be provided at a higher level (e.g. the design library that is describing a `Tooltip` widget).

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: Actually - one problem that I see with this is an inability to reason about flipping and moving the tooltip when it goes off the screen. Having well-defined positions makes that easier to reason about.

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: I personally prefer the flag approach, but maybe we get a third opinion :) Are flags usually done with `bool`s in the flutter package?

packages/flutter/lib/src/widgets/_window.dart
mattkae: True! I can merge up

packages/flutter/lib/src/widgets/_window_positioner.dart
Saviq: While I can see Alignment could express _much_ of what's going on here, Is it enough… or is it too much?

Like Matt's saying, consider a drop-down menu that doesn't fit on screen - to flip it correctly along the horizontal axis, we need to know the anchor points on the parent and child _and_ the offset, separately - so the offset can be flipped as well.

With Alignments (two, one for the parent anchor and another for the child anchor), we'd get "baked" values that we'd have to de-/reconstruct into the anchor points and offset… is an alignment of `(1.0, 0.75)` offset `0.25` from the bottom `bottomCenter`, or from `bottomRight`? …maybe the math can hold either way?

But then, say both alignments are outside the parent and child, we'd need to sum the offsets to reason about them in the positioning logic, but would the result be what the user intended, given the available API? Or did the user intend for the _parent_ offset to be static, and only the _child_ offset to be flipped if constrained (this wouldn't map to Wayland, FWIW)?

It all does add a little complexity, does it outweigh the cost of having a new, similar API?

Saviq: I made a picture to maybe better express why I think there's not enough in an `Alignment`:

Image

However many `Alignment`s, they basically just add up to an offset between the centers of the two rectangles. But when constraints come into play, we need to know the actual anchor points - we need to maintain the relationship between them (e.g. if the child anchor is centered, resizing would be relative to the center axis - if it's a corner, it would be relative to the vertical axis off that corner).

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: Oh derp. Ignore me, I'm being silly.

packages/flutter/lib/src/widgets/_window_positioner.dart
loic-sharma: Thanks for the detailed thoughts. Using separate enums sounds good to me, closing!

packages/flutter/lib/src/widgets/_window_positioner.dart
loic-sharma: Yup, the public API is usually done using `bool`s.

Internally, this can be implemented as a bitfields if needed. I'd maybe avoid this unless we need to pass this value to the embedders.

Here's prior art of bitfields, if you need it: https://github.com/flutter/flutter/blob/f0dbd7dc5adea84d642b3c294ab9325efe588a50/engine/src/flutter/lib/ui/window.dart#L920-L986

mattkae: Bools it is!

packages/flutter/lib/src/widgets/_window.dart
justinmc: Yeah I was definitely thinking we should move this to examples/ when I saw this large code block here. But will it be able to have its tests run against it in the normal infrastructure? @mattkae Can you try moving this to examples/ and adding a trivial test, pushing a commit, and seeing if CI passes? Or if we can't get CI to set the experimental flag, could the test just expect the `main` assertion error?

+1 for asserting for the feature flag in `main` and adding a disclaimer comment.

packages/flutter/lib/src/widgets/_window.dart
justinmc: Using a GlobalKey like this gives you the layout information from the previous frame. That's probably not ideal here because for example if the Text animates its position the tooltip will appear to lag slightly behind it.

But another thing to flag here is that it looks like the tooltip's position won't be updated at all if the Text moves? I haven't run the example myself though.

One thing you might be able to do is to wrap Text in a LayoutBuilder and use the position and size it gives you for the current frame. This might not work if the Text itself is positioning itself somewhere tricky inside of its constraints, like if it is aligned right or center. I don't know if that matters here.

Typically when a widget depends on a sibling widget's layout in order to do its own layout in the same frame, the solution is to use a MutliChildRenderObjectWidget.

A good example of a solution to an extreme version of this problem is [in RawAutocomplete](https://github.com/flutter/flutter/blob/c47233ab3827fbb5bb08ab0dc7de0e7aabd427fd/packages/flutter/lib/src/widgets/autocomplete.dart#L656). It has a floating options window that is positioned based on the layout of a text field while also being aware of its position on the screen so it doesn't overflow the screen. That solution works all in the same frame. Not sure if we need something so complicated here since the platform gives us its constraints though?

I think it's important for us to think about all of this because our users will probably encounter these same problems. They probably want tooltips that position themselves correctly out of the box, and we should try to make that easy for them.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+/// The specified anchor is used to derive an anchor point on the anchor rectangle that
+/// the child [BaseWindowController] will be positioned relative to.
+/// If a corner anchor is set (e.g. [topLeft] or [bottomRight]),
+/// the anchor point will be at the specified corner; otherwise, the derived anchor point
+/// will be centered on the specified edge, or in the center of the anchor rectangle
+/// if no edge is specified.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+enum WindowPositionerAnchor {
```
justinmc, line 21: Could [Alignment](https://main-api.flutter.dev/flutter/painting/Alignment-class.html) work for this? Or if not, maybe this should similarly wrap an x and y double instead of just being an enum.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+///
+/// [flipX] and [flipY] mean reverse the anchor points and offset along an axis.
+/// [slideX] and [slideY] mean adjust the offset along an axis.
+/// [resizeX] and [resizeY] mean adjust the client window size along an axis.
+///
+/// The adjustments can be combined, according to a defined precedence:
+///
+///  1. [flipX] and [flipY]
+///  2. [slideX] and [slideY]
+///  3. [resizeX] and [resizeY]
```
justinmc: This is making me think of a transformation matrix and its translate/rotate/scale components.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// translated in the Y-direction (either negative or positive) in order
+  /// to best display the window on screen.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  final bool slideY;
+
+  /// If [flipX] is `true` and the window would be displayed off the screen in the X-axis in one direction, then
+  /// it will be flipped to the opposite side of its parent in order to show
+  /// to best display the window on screen.
```
justinmc: Typo: "to show to best display"

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+@internal
+class WindowPositionerConstraintAdjustment {
+  const WindowPositionerConstraintAdjustment({
+    this.slideX = false,
+    this.slideY = false,
+    this.flipX = false,
+    this.flipY = false,
+    this.resizeX = false,
+    this.resizeY = false,
+  });
```
justinmc, line 215: Should there be any constraints on possible values here, or are all combinations valid?

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Does this mean that it's possible for multiple of flip/slide/resize to happen? Can you give an example?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// If `true` and the window would be displayed off the screen in the Y-axis, then
+  /// its height will be reduced such that it fits on screen.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  final bool resizeY;
+}
+
+/// The [WindowPositioner] defines how child windows are placed relative to
+/// their parents.
```
justinmc: This is referring to parent windows right? Maybe: "their parent windows."

Also, is it really multiple parents?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  ///
+  /// The specified anchor is used to derive an anchor point that will be positioned
+  /// relative to the [parentAnchor]. If a corner anchor is set (e.g. [WindowPositionerAnchor.topLeft] or
+  /// [WindowPositionerAnchor.bottomRight]), the anchor point will be at the specified corner;
+  /// otherwise, the derived anchor point will be centered on the specified
+  /// edge, or in the center of the anchor rectangle if no edge is specified.
+  ///
+  /// This is used in conjunction with [parentAnchor] to determine the final position
+  /// of the child window.
+  final WindowPositionerAnchor childAnchor;
```
justinmc: I'm still not 100% clear how these work after reading them. Is it that parentAnchor defines a point on the parent and childAnchor defines a point on the child, and the child is placed such that these two points lie on top of each other?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// For example if the anchor of the anchor rectangle is at (x, y), the window
+  /// has a [childAnchor] of [WindowPositionerAnchor.topLeft], and the [offset]
+  /// is (ox, oy), the calculated window position will be (x + ox, y + oy).
+  /// The offset position of the window is the one used for constraint testing.
+  /// See [constraintAdjustment].
+  ///
+  /// An example use case is placing a popup menu on top of a user interface
+  /// element, while aligning the user interface element of the parent window
+  /// with some user interface element placed somewhere in the popup window.
+  final Offset offset;
```
justinmc: The docs here make a lot of sense to me and I really appreciate the example.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// output's 'work area', thus necessitating the child window's
+  /// position be adjusted until it is entirely inside the work area.
+  ///
+  /// The adjustments can be combined, according to a defined precedence:
+  ///
+  /// 1. [WindowPositionerConstraintAdjustment.flipX] and [WindowPositionerConstraintAdjustment.flipY]
+  /// 2. [WindowPositionerConstraintAdjustment.slideX] and [WindowPositionerConstraintAdjustment.slideY]
+  /// 3. [WindowPositionerConstraintAdjustment.resizeX] and [WindowPositionerConstraintAdjustment.resizeY]
+  ///
+  /// The first adjustment that results in the child window being entirely inside the work area will be picked.
```
justinmc: This is also explained on the docs for WindowPositionerConstraintAdjustment, maybe deduplicate it with a macro?

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Below I see the line: "The first adjustment that results in the child window being entirely inside the work area will be picked." That indicates the answer to my question is no. I recommended deduplicating these docs with a macro.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// [childSize] is the frame size of the child window.
+  ///
+  /// [anchorRect] is the rectangle relative to which the child window is placed.
+  ///
+  /// [parentRect] is the parent window's rectangle.
+  ///
+  /// [displayRect] is the output display area where the child window will be placed.
+  ///
+  /// All sizes and rectangles are in physical coordinates.
+  Rect placeWindow({
```
justinmc, line 385: I see this isn't called anywhere yet, how will this be  used?

justinmc: Also, what is preventing this method from being static?

packages/flutter/test/widgets/window_positioner_test.dart
justinmc, line 1: I like all the fast unit tests here that don't use testWidget 👍

justinmc (commented):
Overall looks good. This kind of positioning stuff is fiddly but I think your way of doing things keeps it pretty clean.

My main concern is with the use of GlobalKeys in the example and how that should be done in the ideal situation, because I want to make sure we set users up to be able to get this right on the happy path.

I see that the framework is positioning the tooltip based on constraints from the platform. Just to make sure, do all platforms provide this kind of constraint size information?

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: Will most `Tooltip` widgets use the same `WindowPositioner` values? If they would all have the same default, I'd lean towards repeating the default here as well. If there are reasons why they would have different defaults, omitting it here SGTM 👍

packages/flutter/lib/src/widgets/_window_positioner.dart
Saviq: See above for a discussion on this from Loïc's review: https://github.com/flutter/flutter/pull/177404#discussion_r2468086538

packages/flutter/lib/src/widgets/_window.dart
Saviq: Don't think there is a "defendable" default here - downstream widgets might prefer tooltips above or below the cursor, or to the side of the widget they refer to…

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+///  3. [resizeX] and [resizeY]
+///
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+class WindowPositionerConstraintAdjustment {
+  const WindowPositionerConstraintAdjustment({
+    this.slideX = false,
+    this.slideY = false,
+    this.flipX = false,
+    this.flipY = false,
```
Saviq: Sort it according to precedence?

```suggestion
    this.flipX = false,
    this.flipY = false,
    this.slideX = false,
    this.slideY = false,
```

packages/flutter/lib/src/widgets/_window_positioner.dart
Saviq: They're all valid, yes - with precedence defined above.

packages/flutter/lib/src/widgets/_window_positioner.dart
Saviq: > such that these two points lie on top of each other?

Modulo offset, yes. Please do suggest better wording, this is somewhat etched into our brains by now :)

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: This can't be static, and it will be used by the platforms to place windows. It is shared logic for doing that in dart, which is very useful.

packages/flutter/lib/src/widgets/_window.dart
mattkae: Is it a strict requirement that a tooltip must follow a widget if the position of that widget changes? In most toolkits, it seems to be the case that the tooltip will close if the position of the anchor widget changes, as it assumed that a user might no longer be interested in it at that point.

I ask because it _might not_ be worth muddying up the API for a situation that most users won't find useful :)

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: Thanks for confirming! Marking resolved

packages/flutter/lib/src/widgets/_window.dart
justinmc: That is a great way to avoid all this complexity. I am good with closing the tooltip when the widget position changes. Later if users want it to follow we can address that then.

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Sorry to revive this thread, I had the same thought as Loic, but this is what I was imagining for @Saviq's example above:

```
parentAnchor: Alignment.bottomRight AKA Alignment(1.0, 1.0)
offset: (20, 20)
childAnchor: Alignment.topLeft AKA Alignment(0.0, 0.0)
```

And then no need for the WindowPositionerAnchor class.

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Ok I've replied in that thread. I still think Alignment might be viable there.

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: The use of the term "anchor rectangle" is tripping me up I think.

parentAnchor:

> Defines the point on the parent from which to position the child.
> 
> The child is positioned by placing [childAnchor] on top of [parentAnchor] and then translating by [offset].
> 
> Defaults to [WindowPositionerAnchor.center].


childAnchor:
> Defines the point on the child that is positioned relative to the parent.
> 
> The child is positioned by placing [childAnchor] on top of [parentAnchor] and then translating by [offset].
> 
> Defaults to [WindowPositionerAnchor.center].

offset:

> The offset with which to place the child relative to the parent.
> 
> The child is positioned by placing [childAnchor] on top of [parentAnchor] and then translating by [offset].
> 
> [...your example stuff...]
> 
> Defaults to [Offset.zero].

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// See [constraintAdjustment].
+  ///
+  /// An example use case is placing a popup menu on top of a user interface
+  /// element, while aligning the user interface element of the parent window
+  /// with some user interface element placed somewhere in the popup window.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  final Offset offset;
+
+  /// Defines how Flutter will adjust the position of the window if the unadjusted
```
justinmc, line 357: Looks like an incomplete sentence here.

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  return Offset(p.dx.clamp(r.left, r.right), p.dy.clamp(r.top, r.bottom));
+}
+
+Offset _flipX(Offset offset) {
+  return Offset(-offset.dx, offset.dy);
+}
+
+Offset _flipY(Offset offset) {
+  return Offset(offset.dx, -offset.dy);
+}
```
justinmc, line 520: These can all be static right?

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Oh right it uses childAnchor.

packages/flutter/lib/src/widgets/_window_positioner.dart
Saviq: We can't limit `Alignment` to just the corners + center, though, can we? Is it worth adding that complexity just to avoid the enum?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+      WindowPositionerAnchor.right => rect.centerRight,
+      WindowPositionerAnchor.topLeft => rect.topLeft,
+      WindowPositionerAnchor.bottomLeft => rect.bottomLeft,
+      WindowPositionerAnchor.topRight => rect.topRight,
+      WindowPositionerAnchor.bottomRight => rect.bottomRight,
+    };
+  }
+}
+
+/// The [WindowPositionerConstraintAdjustment] how a window will adjust
```
robert-ancell: Missing a verb after the class name?

packages/flutter/lib/src/widgets/_window_positioner.dart
```diff
@@ -0,0 +1,510 @@
...
+  /// it will be flipped to the opposite side of its parent in order
+  /// to best display the window on screen.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  final bool flipX;
+
+  /// If [flipY] is `true` and then window would be displayed off the screen in the Y-axis in one direction, then
+  /// it will be flipped to the opposite side of its parent in order
+  /// it on screen.
```
robert-ancell: Missing "to best display"? Probably "it will be flipped to the opposite side of its parent in order to best display the window on screen." to match the case for `flipX`.

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: These do not belong to any class

packages/flutter/lib/src/widgets/_window.dart
mattkae: Yup, let me try that

packages/flutter/lib/src/widgets/_window.dart
mattkae: Won't the `LayoutBuilder` only provide the size, and not the position?

Or is `_OverlayChildLayoutBuilder` the thing that I really want here?

packages/flutter/lib/src/widgets/_window.dart
mattkae: Added! Let's see if CI passes...

packages/flutter/lib/src/widgets/_window_positioner.dart
mattkae: Do you have a better wording in mind? That is the best that we could come up with :sweat_smile:

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: > Won't the `LayoutBuilder` only provide the size, and not the position?

That's my understanding as well. I believe `LayoutBuilder` is called when constraints are passed down from the parent to the child. The position isn't known until later, after the child has passed its size back up to the parent.

> Or is _OverlayChildLayoutBuilder the thing that I really want here?

I think you want something else. Overlays are how Flutter does "floating" content within a window. But here we're doing "floating" content outside a window.

I bubbled overlays  up as `_OverlayChildLayoutBuilder` figures out a [`OverlayChildLayoutInfo.childPaintTransform`](https://api.flutter.dev/flutter/widgets/OverlayChildLayoutInfo/childPaintTransform.html). I imagine whatever trick we're using to figure out the `childPaintTransform` is the same trick we'll want to use here as well for tooltip windows.

mattkae: Kk, we might have to expose some new helper for people to use in that case, or perhaps hide in the material library.

@knopp Did you have a recommended way of doing this somewhere? I remember you mentioning it.

loic-sharma: It looks like overlay portal's [trick](https://github.com/flutter/flutter/blob/main/packages/flutter/lib/src/widgets/overlay.dart#L2701) is to create a proxy render object that lays out its child and track its size and paint transform. It uses a layout callback to report this information back to the widget layer.

For this to work, I believe that whatever positions the tooltip needs to be a parent of the widget that is the target anchor. We _might_ need to change the API of `TooltipWindow` to reflect this.

knopp: It seems to me that this discussion should perhaps be happening on level above the window controller? I'd be happy with the controller taking anchor rect in global coordinates and providing function to update the anchorRect/positioner. I don't think we want to bake in any widget position tracking in it, that should be responsibility of whatever widget library is actually providing the tooltip. So we could address this when migrating material to `TooltipWindowController` for example.

knopp: Here is a very rough example of working tooltip that tracks anchor widget:
https://github.com/knopp/flutter/blob/ef90e8196005d2daa668efbab0a2da232d817cc4/examples/multiple_windows/lib/app/main_window.dart#L18

The tracking can probably be done better, the point is that all that's needed on `TooltipWindowController` is the `updatePosition` method 

https://github.com/knopp/flutter/blob/ef90e8196005d2daa668efbab0a2da232d817cc4/packages/flutter/lib/src/widgets/_window.dart#L644

(the rect is in global (root widget) coordinates)

mattkae: Would it be reasonable to leave this example using the `GlobalKey` method for now, and leave it to be updated by you when you land the tooltip implementations?

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: Meaning you don't want to allow `Alignment(0.1, 0.1)`? Why not?

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: That's static enough for me 😁

examples/api/lib/widgets/windows/tooltip.0.dart
```diff
@@ -0,0 +1,96 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// TODO(mattkae): remove invalid_use_of_internal_member ignore comment when this API is stable.
```
justinmc, line 5: Nit: Maybe link to a GitHub issue.

packages/flutter/lib/src/widgets/_window_positioner.dart
justinmc: See my previous comment. The `>` quotes are not actually quoting from the code, they are my suggestions for the new wording. Maybe quote wasn't the best way to communicate that 😄

packages/flutter/lib/src/widgets/_window.dart
justinmc: This sounds reasonable to me so we don't hold up this PR.  Maybe adding a comment in the code like "This positioning logic is simpler than you might want in production."

Also, can you create an issue for this so we don't forget?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -620,6 +621,148 @@ abstract class DialogWindowController extends BaseWindowController {
...
+  /// The tooltip will be destroyed if its parent is destroyed.
+  BaseWindowController get parent;
+
+  /// Request change to the constraints of the window.
+  ///
+  /// The [constraints] describes the new constraints that the window should
+  /// satisfy. If the constraints disagree with the current size of the window,
+  /// the platform might resize the window to satisfy the new constraints.
+  ///
+  /// The platform is free to ignore this request.
```
justinmc, line 759: Out of curiosity, is it possible to get the size of the tooltip window after calling this method to see what the size ended up being?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -657,6 +800,22 @@ abstract class WindowingOwner {
...
   });
+
+  /// Creates a [TooltipWindowController] with the provided properties.
+  ///
+  /// Most app developers should use [TooltipWindowController]'s constructor
+  /// instead of calling this method directly. This method allows platforms
+  /// to inject platform-specific logic.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
```
justinmc, line 811: Should this be protected?

justinmc (commented):
I think the only blocker for me is double checking that we really don't want to use Alignment, otherwise this looks good!

packages/flutter/lib/src/widgets/_window.dart
mattkae: No, because we want to enable implementers of the `TooltipWindowController` interface to call this method. It is public by design

packages/flutter/lib/src/widgets/_window.dart
mattkae: You can use `WindowScope.contentSizeOf`

examples/api/lib/widgets/windows/tooltip.0.dart
mattkae: I will link to the pre-launch checklist: https://github.com/flutter/flutter/issues/177586

packages/flutter/lib/src/widgets/_window.dart
mattkae: Here is the issue I made for this: https://github.com/flutter/flutter/issues/178829

justinmc (approved):
LGTM 👍 

I'm on board with the WindowPositionerAnchor enum instead of Alignment, especially given that it more closely maps to Waylands implementation. Thanks for all of the changes.

PR #177428, opened 2025-10-23
refactor: migrate FadeForwardsPageTransitionsBuilder to widgets

rkishan516:
Changes:

* Move FadeForwardsPageTransitionsBuilder from `material/page_transitions_theme.dart` to `widget/page_transitions_builder.dart`
* Replace `Colors.transparent` with `const Color(0x00000000)` in FadeForwardsPageTransitionsBuilder.
* Remove `ColorScheme.surface` with black as fallback color when animating.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/page_transitions_builder_test.dart
```diff
@@ -505,6 +506,469 @@ void main() {
...
+        coloredBox.color,
+        isNot(Colors.transparent),
+      ); // Color is not transparent during animation.
+
+      await tester.pump(const Duration(milliseconds: 801));
+      expect(find.text('page b'), findsOneWidget);
+      coloredBox = tester.widget(find.byType(ColoredBox).last);
+      expect(coloredBox.color, Colors.transparent); // Color is transparent during animation.
+    },
+    variant: TargetPlatformVariant.only(TargetPlatform.android),
```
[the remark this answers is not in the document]
rkishan516: Test was migrated from `page_transitions_theme_test`

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +340,227 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+
+    final bool isOpaque = ModalRoute.opaqueOf(context) ?? true;
+
+    if (!isOpaque) {
+      return builder;
+    }
+
+    return ColoredBox(
+      color: secondaryAnimation.isAnimating
+          // TODO(kishan): Fallback when animating should be surface color.
```
victorsanni: Nit: TODOs should use the github username.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +340,227 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+  static final Animatable<double> _fadeOutTransition = Tween<double>(
+    begin: 1.0,
+    end: 0.0,
+  ).chain(CurveTween(curve: const Interval(0.0, 0.25)));
+
+  // The default background color.
+  static const Color _defaultBackgroundColor = Color(0x00000000);
+
+  // The default background color.
+  static const Color _fallbackBackgroundColor = Color(0xFF000000);
```
victorsanni: I'm assuming this is the default surface color. But this is for `ColorScheme.light`, the default dark surface color is `Color(0xff121212)`.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +340,227 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+  // The fade out transition of the old page when the new page appears.
+  static final Animatable<double> _fadeOutTransition = Tween<double>(
+    begin: 1.0,
+    end: 0.0,
+  ).chain(CurveTween(curve: const Interval(0.0, 0.25)));
+
+  // The default background color.
+  static const Color _defaultBackgroundColor = Color(0x00000000);
+
+  // The default background color.
```
victorsanni: I don't think this comment is descriptive of the variable.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
victorsanni: Also `fallback` is basically `default`, can the name be changed to something more relevant? e.g `defaultLightSurfaceColor` with an additional `defaultDarkSurfaceColor`.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +340,227 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+  ).chain(CurveTween(curve: const Interval(0.0, 0.75)));
+
+  // The fade out transition of the old page when the new page appears.
+  static final Animatable<double> _fadeOutTransition = Tween<double>(
+    begin: 1.0,
+    end: 0.0,
+  ).chain(CurveTween(curve: const Interval(0.0, 0.25)));
+
+  // The default background color.
+  static const Color _defaultBackgroundColor = Color(0x00000000);
```
victorsanni, line 502: This color is only used when not animating? That should be specified in the comment.

victorsanni (commented):
Looks good so far. 

It does bug me that we lose any set `ColorScheme.surface` because that's only in the Material library. Maybe we can add a `surfaceColor` to the parameters in the constructor for the transitions builder? To at least make it customizable. Then users in the Material library can add `ColorScheme.of(context).surface` as that property. 

If we go with this approach, it will need to be communicated with developers. Maybe we can add an example in the Material library that does this. Maybe add a migration guide? What do you think @MitchellGoodwin?

Or is there another way out?

EDIT: Since the `backgroundColor` property already exists, we don't need a `surfaceColor` property. Users in Material can supply `ColorScheme.of(context).surface` to `backgroundColor`.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
rkishan516: Sure, will update.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
rkishan516: Okay, I will update. Since we can't access Theme here for brightness, should we use platformBrightness?

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +342,234 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+  // The fade out transition of the old page when the new page appears.
+  static final Animatable<double> _fadeOutTransition = Tween<double>(
+    begin: 1.0,
+    end: 0.0,
+  ).chain(CurveTween(curve: const Interval(0.0, 0.25)));
+
+  // The default background color, used when no transition is active.
+  static const Color _defaultBackgroundColor = Color(0x00000000);
+
+  // The default surface color for light theme, used as fallback during transitions.
```
victorsanni: `light theme` is a generic descriptor of what's actually happening. This color is used when the platform brightness is light, that should be clearly specified here instead.

packages/flutter/lib/src/widgets/page_transitions_builder.dart
```diff
@@ -338,3 +342,234 @@ class OpenUpwardsPageTransitionsBuilder extends PageTransitionsBuilder {
...
+  static const Color _defaultLightSurfaceColor = Color(0xFF000000);
+
+  // The default surface color for dark theme, used as fallback during transitions.
```
victorsanni: Same here.

chunhtai:
In regard to the color situation, I think the current approach may not be ideal. The existing customer would have to set the background color from color scheme to preserve the existing behavior. We should avoid making the code to be more verbose if possible.

I can think of 2 way forward:

1. have a DefaultPageTransitionStyle similar to DefaultTextStyle and DefaultSelectionStyle...etc. The material theme will builds a DefaultPageTransitionStyle based on its ColorTheme

2. Make the widget version's TransitionBuilder's color or other material theme related properties to be required properties. The material library still retain these class, but instead it builds on top of the widget version's with its require properties infer from material theme

rkishan516:
@victorsanni so whats the conclusion? Just move FadeForwardsPageTransitionsBuilder to widgets with backgroundColor as required

rkishan516:
> I'd recommend pausing this particular PR for now while we keep discussing the available options here. But feel free to open PRs for the other transition builders in the meantime.

Hey @victorsanni, I think we need to make a decision before proceeding further.

The remaining transition builders are ZoomPageTransitionsBuilder and CupertinoPageTransitionBuilder, but both have dependencies that need to be addressed:

  1. **ZoomPageTransitionsBuilder** - Has a dependency on surface color, similar to FadeForwardsPageTransitionsBuilder
  2. **CupertinoPageTransitionBuilder** - Is fully dependent on CupertinoRoute

rkishan516:
Thanks @victorsanni, Makes sense. Then only thing remaining is moving cupertino route to cupertino/ . I will try to push that ASAP.

Issue #81666, opened 2021-05-03
flutter build ios - Crash  - [tool_crash] FormatException : Missing extension byte (at offset 17)


eminovperviz:
## Command
```
flutter build ios
```

## Steps to Reproduce
1. ...
2. ...
3. ...

## Logs
FormatException
```
#0      _Utf8Decoder.convertChunked (dart:convert-patch/convert_patch.dart:1892:7)
#1      _Utf8ConversionSink.addSlice (dart:convert/string_conversion.dart:314:28)
#2      _Utf8ConversionSink.add (dart:convert/string_conversion.dart:310:5)
#3      _ConverterStreamEventSink.add (dart:convert/chunked_conversion.dart:72:18)
#4      _SinkTransformerStreamSubscription._handleData (dart:async/stream_transformers.dart:111:24)
#5      _rootRunUnary (dart:async/zone.dart:1436:47)
#6      _CustomZone.runUnary (dart:async/zone.dart:1335:19)
#7      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1244:7)
#8      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:341:11)
#9      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:271:7)
#10     _SyncStreamControllerDispatch._sendData (dart:async/stream_controller.dart:733:19)
#11     _StreamController._add (dart:async/stream_controller.dart:607:7)
#12     _StreamController.add (dart:async/stream_controller.dart:554:5)
#13     _Socket._onData (dart:io-patch/socket_patch.dart:2160:41)
#14     _rootRunUnary (dart:async/zone.dart:1444:13)
#15     _CustomZone.runUnary (dart:async/zone.dart:1335:19)
#16     _CustomZone.runUnaryGuarded (dart:async/zone.dart:1244:7)
#17     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:341:11)
#18     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:271:7)
#19     _SyncStreamControllerDispatch._sendData (dart:async/stream_controller.dart:733:19)
#20     _StreamController._add (dart:async/stream_controller.dart:607:7)
#21     _StreamController.add (dart:async/stream_controller.dart:554:5)
#22     new _RawSocket.<anonymous closure> (dart:io-patch/socket_patch.dart:1696:33)
[...]

PR #177470, opened 2025-10-23
[ Tool ] Add `Stream.transformWithCallSite` to provide more useful stack traces

bkonyi:
Exceptions thrown within a stream transformer don't provide any context as to where the call to `transform(...)` occurred within the program, often resulting in stack traces consisting of only Dart SDK sources.

This change adds a new extension method on `Stream` called `transformWithCallSite`, which captures the current `StackTrace` upon invocation. This stack trace is reported in the case of an error in order to provide context for better error reporting.

Example issue: https://github.com/flutter/flutter/issues/81666


jyameo (approved):
lgtm!

Issue #166173, opened 2025-03-28
[A11y] Further optimize geometry calculation in semantics


chunhtai:
Right now when a render object is mark semantics needs update, all ancestor render objects up to semantics boundary geometrics and are marked dirty, which is unnecessary.

We should only mark dirty the renderobject that mark needs semantics update, this will require us to also keep track of these renderobject besides the semantics boundary.

LongCatIsLooong:
Update:

After some initial attempts, it seems to be pretty difficult to apply the same collect-dirty-boundaries-and-flush strategy as we do in layout for semantics geometry calculation. One example is OverlayPortal (whose overlay child's semantics parent RenderObject may not be a real ancestor of the overlay child's RenderObject, that implementation may change soon but still the framework does not make the guarantee). Its cached transform must be invalidated if any render object within the path `semanticsParent` - `common ancestor` - `overlayChild` repaints. To know whose cache must be invalidated when any RenderObject is marked for repaint, we need to either store the dependents in each RenderObject (which is very expensive to maintain), or find the dependents on the fly (so `markNeedsPaint` and `markNeedsSemanticsUpdate` becomes expensive).

So at the moment I'm trying out just doing treewalk from the root to compute geometry (with caching), since the semantics tree is usually way shallower than the RenderObject tree. With a rudimentary implementation I'm seeing a 300-400us increase in `drawFrameDuration` from the `bench_material3_scroll_semantics` benchmark (from around 1800us). I'll see if I can shave some of the increase off by separating transform and clipping into different cache.

PR #177477, opened 2025-10-23
Make a11y `computeChildGeometry` slightly faster

LongCatIsLooong:
This is not a fix for #166173, it's only a by-product. The fix will be in separate PR, this includes the unrelated code changes to make that PR smaller.

According to `bench_material3_semantics`, this reduces `Semantics.ensureGeometry` by ~100 μs (321.6 μs -> 228.9 μs). 

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5578,47 +5583,48 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
...
-          in result.$1.whereType<_RenderObjectSemantics>()) {
-        assert(childSemantics.contributesToSemanticsTree);
-        if (childSemantics.shouldFormSemanticsNode) {
-          _children.add(childSemantics);
-        } else {
-          _children.addAll(childSemantics._children);
-          siblingMergeGroups.addAll(childSemantics.siblingMergeGroups);
-        }
-      }
+    if (!contributesToSemanticsTree) {
```
LongCatIsLooong, line 5608: This only changed indentation (returning early if `!contributesToSemanticsTree`).

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -6383,38 +6382,61 @@ final class _SemanticsGeometry {
...
       semanticsClipRect = parentSemanticsClipRect;
-      assert(parentToCommonAncestorTransform == null);
+      // Travrese from `parent`'s render object to `child`'s.
       for (int i = childToCommonAncestor.length - 1; i > 0; i -= 1) {
-        (paintClipRect, semanticsClipRect) = _computeClipRect(
-          childToCommonAncestor[i],
-          childToCommonAncestor[i - 1],
-          semanticsClipRect,
-          paintClipRect,
+        final RenderObject nodeParent = childToCommonAncestor[i];
```
LongCatIsLooong: For the more common codepath the transform is now computed in the same loop and there are less `inverse` calculations. The overlay portal path still uses `_computeClipRect` but that path usually isn't as hot.

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: (at first glance the `ancestorNode` may not have up-to-date geometry, if it is going to be merged as a sibling, and this function is being called in `_updateGeometry`?)

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5919,10 +5925,7 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
   }
 
   /// Builds the semantics subtree under the [cachedSemanticsNode].
-  void _buildSemanticsSubtree({
-    required Set<int> usedSemanticsIds,
-    List<SemanticsNode>? semanticsNodes,
-  }) {
+  void _buildSemanticsSubtree({required Set<int> usedSemanticsIds}) {
```
LongCatIsLooong, line 5956: The `semanticsNodes` argument wasn't used.

packages/flutter/lib/src/painting/matrix_utils.dart
```diff
@@ -72,6 +72,65 @@ abstract final class MatrixUtils {
     return null;
   }
 
+  /// Computes the result of matrix multiplication `a x b` and stores the result
+  /// in `b`.
+  ///
+  /// This function does not alter the argument `a`, and is thus useful when you
+  /// want to left-multiply in-place. For right-multiply in-place, see
+  /// [Matrix4.multiply].
+  static void multiplyInPlace(Matrix4 a, Matrix4 b) {
```
chunhtai, line 81: we should add a todo to clean this up
https://github.com/google/vector_math.dart/issues/250

packages/flutter/lib/src/rendering/object.dart
chunhtai: Now Qun's PR is merged, we no longer needed a special path for overlayportal.

packages/flutter/lib/src/rendering/object.dart
chunhtai: dirty nodes' ensureSemantics are called from top to bottom, so ancestor should always have up to date geometry

packages/flutter/lib/src/painting/matrix_utils.dart
LongCatIsLooong: created https://github.com/google/vector_math.dart/pull/354. I'll see if I can bump the dependencies after that PR is merged.

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: I think in theory people can still return a non-descendant node in their custom RenderObject's`visitChildrenForSemantics` implementation? (But now that `RenderObject.semantcsParent` is gone I'm not sure if an implementation like still works).

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: > ensureSemantics

do you ensureGeometry? What if there's a merge sibling ancestor (whose geometry is updated in `_updateSiblingNodesGeometries` which is called in `_produceSemanticsNode`? The geometry of that isn't updated when `ensureGeometry` is called I think?

chunhtai: I just realized this is grabing from cached Semantics Node, not the _SemanticsGeometry that stored in the _RenderObjectSemantics, this looks like a bug because ensureGeometry doesn't update the cached semantics node's geomtry.

> _updateSiblingNodesGeometries

I think if we fixes the the code to grab from _SemanticsGeometry of the _RenderObjectSemantics this should be fine?

packages/flutter/lib/src/rendering/object.dart
chunhtai: right, I don't think they can have this case where Render tree and semantics tree are differ without semanticsParent

LongCatIsLooong: Sounds good, I don't think I've touched that code path in this patch so I'll probably replace that with an assert in a different PR.

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: I guess if this code path is now obsolete then I guess we don't really have to fix this?

justinmc:
@LongCatIsLooong FYI this has conflicts.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -6370,97 +6370,70 @@ final class _SemanticsGeometry {
...
-      for (int i = childToCommonAncestor.length - 1; i > 0; i -= 1) {
-        (paintClipRect, semanticsClipRect) = _computeClipRect(
-          childToCommonAncestor[i],
-          childToCommonAncestor[i - 1],
-          semanticsClipRect,
-          paintClipRect,
-        );
-      }
-    } else {
-      // Otherwise we have to find the closest ancestor RenderObject that
```
LongCatIsLooong, line 6438: Ok I got rid of this path since `RenderObject.semanticsParent` is now gone so `parent` must be an ancestor of `child`.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -6370,97 +6368,70 @@ final class _SemanticsGeometry {
...
-            clipPath[i],
-            semanticsClipRect,
-            paintClipRect,
-          );
-          parent = clipPath[i];
-        }
-      }
+      final Rect? localPaintClipInParent = _transformRect(
+        nodeParent.describeApproximatePaintClip(node),
+        transform, // paint transform from nodeParent to parent
```
chunhtai: the comment seems to be inaccurate. Isn't this transform going to be from original child to the current ancestor?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -6485,54 +6456,18 @@ final class _SemanticsGeometry {
   }
 
   /// From parent to child coordinate system.
-  static Rect? _transformRect(Rect? rect, Matrix4 transform) {
+  static Rect? _transformRect(
+    Rect? rect,
+    Matrix4 transform, [
+    Rect Function(Matrix4, Rect) apply = MatrixUtils.inverseTransformRect,
```
chunhtai: we should type def this

chunhtai (commented):

chunhtai (approved):

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: The comment was ambiguous (rephrased to "paint transform of nodeParent in relation to parent") but I think it was correct. To avoid calculating the inverse in every step we'll have to start from the parent. The transform is applied in-place at the end of each for loop so when we calculate the "local" clips the transform is still the paint transform of nodeParent instead node.

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: Added a typedef. But I don't like the typedef because it doesn't tell me the exact signature of the function `_transformRect` takes unless I look it up.

chunhtai: inline the function signature is also hard to read

Issue #177071, opened 2025-10-15
Use androidx.exifinterface in android embedder


mboetger:
### Use case

https://github.com/flutter/flutter/pull/176860 adds some checking of exif metadata.  When trying to add androidx.exifinterface to the cipd bundle and the PR, the issue https://github.com/flutter/flutter/issues/177066 was found.  

### Proposal

After https://github.com/flutter/flutter/issues/177066 is fixed, update the code and the CIPD bundle to use androidx.exifinterface.

PR #177416, opened 2025-10-22
Bump Templates To Correct Versions

jesswrd:
I bumped templates to the correct versions. I also included a fix for kgp and agp compatability.

AGP 8.11.0 -> 8.11.1
Gradle 8.13 -> 8.14
KGP 2.2.0 -> 2.2.20

I ran `flutter analyze --suggestions` and `flutter build apk` to ensure these are safe versions.

Will cherry pick this to beta after the pr merges.

Fixes https://github.com/flutter/flutter/issues/177320

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #177508, opened 2025-10-24
Gradle Lockfile workaround and Android Embedder dependency change for ExifInterface

### LINKED ISSUE #177066: generate_gradle_lockfiles needs to work with local engine changes
### Use case

https://github.com/flutter/flutter/pull/176860 is currently blocked because it adds a new android embedder dependency (androidx.exifinterface).  Before the monorepo, it was expected this change would land and then a follow-up PR would update the lock files (see engine/src/flutter/tools/cipd/android_embedding_bundle/README.md).  

However, since it's now a monorepo- the changes need to happen in the same PR.  However,the script does not take into account local engine changes so running the script in the same PR does nothing. 

### Proposal

generate_gradle_lockfiles.dart needs to account for local engine changes when generating lockfiles.

mboetger:
Create a workaround for generating gradle lock files.  When the ignore file exists, gradle locking will be disabled. This will allow deps for the embedder to be checked in, and then in a follow-up PR the lockfiles can correctly be updated and the ignore file can be deleted.

Fixes: #177066, #177071

## Pre-launch Checklist

- [X] I signed the [CLA].


dev/integration_tests/android_views/android/.ignore-locking.md
reidbaker, line 1: Maybe put an actual reason here

gmackall (approved):
I think this is a reasonable solution given the difficulty of the alternative

reidbaker (changes requested):
I am for this proposal. That said it is unlikley that we will want to exempt only some of these projects. Is it possible to link to a ignore-locking.md file in the root? 

Second I think the ignore file should have a reason with a bug in the file that way someone can look up the progress or context for why locking is ignored.

dev/integration_tests/android_views/android/.ignore-locking.md
mboetger: I put it in the same place the exclusion file which made sense to me since it is 'config'

dev/tools/bin/config/.ignore-locking.md
```diff
@@ -0,0 +1 @@
+177508
```
reidbaker:
```suggestion
Reason: https://github.com/flutter/flutter/pull/177508
```

dev/tools/bin/generate_gradle_lockfiles.dart
```diff
@@ -217,42 +248,46 @@ String exec(String cmd, List<String> args, {String? workingDirectory}) {
...
-}
+  allprojects {
+      repositories {
+          google()
+          mavenCentral()
+      }
+  }
 
-rootProject.layout.buildDirectory.value(rootProject.layout.buildDirectory.dir("../../build").get())
+  rootProject.layout.buildDirectory.value(rootProject.layout.buildDirectory.dir("../../build").get())
```
reidbaker: Why add all of this indentation to every file?

dev/tools/bin/generate_gradle_lockfiles.dart
```diff
@@ -298,54 +333,58 @@ plugins {
...
+  // root_app/android/build.gradle.kts).
+  // After modification verify formatting with ktlint.
+  const String rootGradleKtsFileContent = r'''
+  // Copyright 2014 The Flutter Authors. All rights reserved.
+  // Use of this source code is governed by a BSD-style license that can be
+  // found in the LICENSE file.
+
+  // This file is auto generated.
+  // To update all the settings.gradle files in the Flutter repo,
+  // See dev/tools/bin/generate_gradle_lockfiles.dart.
```
reidbaker: Same here, migration to a function added indentation to string contents.

engine/src/flutter/tools/cipd/android_embedding_bundle/README.md
```diff
@@ -34,11 +34,16 @@ below explain how to fetch the license information for the dependencies.
...
-the Gradle lockfiles using the script at
-`<framework_repo>/dev/tools/bin/generate_gradle_lockfiles.dart`
-(run with the `--no-gradle-generation` and `--no-exclusion` flags).
\ No newline at end of file
+1. The Gradle lockfiles will need to be updated, but they cannot be
+   updated in this PR.  They will need to be updated in a follow-up
+   PR.  Instead, run
+   `<repo_root>/dev/tools/bin/generate_gradle_lockfiles.dart
+   --no-gradle-generation --no-exclusion --ignore-locking
+   --ignore-reaseon=<PR>`.  Replace <PR> with the PR number. This will
```
reidbaker:
```suggestion
   --ignore-reaseon=Reason: <ISSUE>`.  Replace <ISSUE> a link to the issue. This will
```

packages/integration_test/example/android/settings.gradle.kts
```diff
@@ -34,8 +34,8 @@ buildscript {
 
 plugins {
     id("dev.flutter.flutter-plugin-loader") version "1.0.0"
-    id("com.android.application") version "8.11.0" apply false
-    id("org.jetbrains.kotlin.android") version "2.2.0" apply false
+    id("com.android.application") version "8.11.1" apply false
+    id("org.jetbrains.kotlin.android") version "2.2.20" apply false
```
reidbaker, line 38: I wanted to confirm that you intended to update the AGP version and KGP version. These are the version I would expect.

reidbaker (approved):

dev/tools/bin/generate_gradle_lockfiles.dart
mboetger: Because the string got moved into a function and I indented the string.  This should be fixed.

packages/integration_test/example/android/settings.gradle.kts
mboetger: Not really - this was a side effect of re-generating the gradle files (with the new, singular, relative lock file).  I think it should be ok though.  Can try to unwind this if you prefer.

mboetger: Might have been since https://github.com/flutter/flutter/pull/177416 just landed?

reidbaker: Don't spend any time to unwind, it is fine.

jtmcdole:
> > Is it possible to link to a ignore-locking.md file in the root?
> 
> @reidbaker It seems like moving to a single file is causing the tests to fail again. It may be because there is not a full repo checkout in the shards? https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8699826480735377169/+/u/run_test.dart_for_build_tests_shard_and_subshard_1_5/stdout

You get a [full fetch](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8699826480735377169/+/u/Checkout_flutter_flutter/git_fetch/stdout) and [full checkout](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8699826480735377169/+/u/Checkout_flutter_flutter/git_checkout/stdout).

dev/tools/bin/generate_gradle_lockfiles.dart
```diff
@@ -41,8 +41,15 @@ void main(List<String> arguments) {
           'Run the script using the config file at ./configs/lockfile_exclusion.yaml to skip the specified subdirectories.',
       defaultsTo: true,
     )
-    ..addFlag('ignore-locking', help: 'Generate ignore file to disable gradle dependency locking. A reason must be given.')
-    ..addOption('ignore-reason', help: 'Reason to disable gradle dependency locking. A reason must be given if --ignore-locking is used.');
+    ..addFlag(
+      'ignore-locking',
+      help: 'Generate ignore file to disable gradle dependency locking. A reason must be given.',
+    )
+    ..addOption(
```
reidbaker: You can simplify this code by only using ignore-reason. There is no reason to use the ignore-reason flag without ignore-locking and there is no reason to use ignore-locking without a reason.

reidbaker (approved):

Issue #115926, opened 2022-11-23
Blur filter isn't clipped when using border radius


erikkiralj1:
## Steps to Reproduce

1. [GitHub repo](https://github.com/erikkiralj1/flutter_platform_blur) with reproducible example
2. Run the app on iOS device, preferably iOS simulator

**Expected results:** 
When adding a border radius to a widget, blur should be cut of to follow the radius of the widget

**Actual results:** 
When adding a border radius to a widget, the blur stays the original size of the widget without the border radius
<details>
<summary>Code sample</summary>

```dart
// BlurredBlock.dart
// Container as a BackdropFilter child wrapped in ClipRect that has a given border radius 
return ClipRect(
      child: BackdropFilter(
        filter: ImageFilter.blur(sigmaX: 10.0, sigmaY: 10.0),
        child: Container(
          width: 800,
          height: 800,
          decoration: BoxDecoration(
            color: Colors.black.withOpacity(0.5),
            borderRadius: BorderRadius.circular(500),
          ),
          child: const Center(
            child: Text('Blurred block 1'),
          ),
        ),
      ),
    );
```

```dart
// main.dart
// Renders Stack with Google Maps widget and BlurredBlock widget positioned on top of it in the center
 const Center(
              child: BlurredBlock(),
            )
```
</details>

<details>
  <summary>Logs</summary>

```
[✓] Flutter (Channel master, 3.5.0-12.0.pre.175, on macOS 13.0 22A380 darwin-x64, locale en-HR)
    • Flutter version 3.5.0-12.0.pre.175 on channel master at /Users/erikkiralj/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 92f10ed712 (2 weeks ago), 2022-11-08 17:57:52 -0800
    • Engine revision e7d7edab98
    • Dart version 2.19.0 (build 2.19.0-374.0.dev)
    • DevTools version 2.19.0

[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
    • Android SDK at /Users/erikkiralj/Library/Android/sdk
    • Platform android-33, build-tools 32.0.0
    • Java binary at: /Users/erikkiralj/Library/Application Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/213.7172.25.2113.9123335/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.13+0-b1751.21-8125866)
[...]

PR #177551, opened 2025-10-26
Platform view blur clipping - Rounded Rect (iOS)

Hari-07:
This PR aims to fix blurs on platform views not being clipped on ios. Currently the clips were not forwarded to the mutator stack and thus just ignored. The frame the blur is in was being used which would always behave like a clip rect ignoring any corner radii or clip paths etc

Now all rrect, rse and path clips are being forwarded to the mutator stack. 

**Currently only rounded rect has been implemented though, and even in that just a single uniform corner radii, which seems to be most of the use cases from users who have run into this issue.** 

Example: 
```dart
    return ClipRRect(
      borderRadius: BorderRadius.circular(50),
      child: Container(
        width: 200,
        height: 200,
        decoration: BoxDecoration(
          color: Colors.black.withOpacity(0.2),
          borderRadius: BorderRadius.circular(500),
        ),
        child: BackdropFilter(
          filter: ImageFilter.blur(sigmaX: 10.0, sigmaY: 10.0),
          child: const Center(
            child: Text('Blurred block 1'),
          ),
        ),
      ),
    );
```
Before:

After:

https://github.com/flutter/flutter/issues/115926

## Pre-launch Checklist

- [x] I signed the [CLA].


LukeMoody01:
Tested this fix in our production app (heavy usage of blurred overlays over platform views). The clipping works perfectly, but we're seeing significant performance impact:

- **Debug builds**: 2-3ms per raster frame (~15-25% slower)
- **Release builds**: 1-2ms per raster frame (~8-15% slower)

The bottleneck seems to be:
1. Platform view layer creation for each blur instance on every frame
2. SkRRect → CALayer geometry conversion per frame
3. No caching when clip geometry hasn't changed

Potential optimisations:
- Cache clip layers between frames when geometry is static
- Batch multiple BackdropFilters into a single platform view
- GPU shader-based clipping instead of CALayer composition

Happy to provide profiling data or test optimisation approaches if there's interest.

Test setup: Flutter 3.35.7 baseline vs custom build with this PR, iPhone 16 Pro (iOS 18.6.2)

Hari-07:
@LukeMoody01 diving a bit more into the bottlenecks and optimisations you have mentioned - any chance this was AI generated? Because this isn't really making sense to me based off of the code. For example

> No caching when clip geometry hasn't changed.
> Platform view layer creation for each blur instance on every frame

As I understand `@property(nonatomic) NSMutableArray<UIVisualEffectView*>* backdropFilterSubviews;` is storing the filters and those are being reused 
```
if (self.backdropFilterSubviews.count <= index) {
      backdropFilterView = filter.backdropFilterView;
      [self addSubview:backdropFilterView];
      [self.backdropFilterSubviews addObject:backdropFilterView];
    } else {
      [filter updateVisualEffectView:self.backdropFilterSubviews[index]];
    }
```

I did have some stray logs I had added in this PR that I have now removed, can you try rerunning your performance benchmark? I think the perf hit just came from those rather than anything else

Hari-07:
> Bottlenecks were indeed AI-generated based on your git diff, engine code isn’t my usual domain, so I’d take those suggestions with a healthy grain of salt.

Please avoid these IMO. While its obviously helpful to test it out and share the performance you saw, if unfamiliar its better not to comment on bottlenecks and fixes rather than use AI and give wrong suggestions which ended up wasting my time trying to make sense of it

gmackall (commented):
From android triage: just noting that the small android portion looks reasonable, to get it out of the android triage queue

engine/src/flutter/flow/embedded_views.h
```diff
@@ -37,7 +39,11 @@ enum class MutatorType {
   kClipPath,
   kTransform,
   kOpacity,
-  kBackdropFilter
+  kBackdropFilter,
+  kBackdropClipRect,
+  kBackdropClipRRect,
+  kBackdropClipRse,
```
hellohuanlin: Rse isn't a widely used acronym. I'd go with `RSuperellipse`

engine/src/flutter/flow/embedded_views.h
```diff
@@ -109,6 +171,11 @@ class Mutator {
       case MutatorType::kClipPath:
       case MutatorType::kClipRRect:
       case MutatorType::kClipRSE:
+      // TODO: Figure out if this matters
```
hellohuanlin: reminder on this before landing

engine/src/flutter/flow/layers/clip_rrect_layer.cc
```diff
@@ -23,4 +23,8 @@ void ClipRRectLayer::ApplyClip(LayerStateStack::MutatorContext& mutator) const {
   }
 }
 
+void ClipRRectLayer::PushClipToEmbedderStack(PrerollContext* context) const {
```
hellohuanlin: Nit: `PushClipToEmbeddedNativeViewMutatorStack`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -163,6 +165,9 @@ - (void)updateVisualEffectView:(UIVisualEffectView*)visualEffectView {
   visualEffectSubview.layer.backgroundColor = UIColor.clearColor.CGColor;
   visualEffectView.frame = _frame;
 
+  visualEffectView.layer.cornerRadius = _cornerRadius;
+  visualEffectView.clipsToBounds = YES;
```
hellohuanlin, line 169: will this cause regression?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -576,8 +583,21 @@ - (void)applyMutators:(const flutter::MutatorsStack&)mutatorsStack
         CGFloat blurRadius = (*iter)->GetFilterMutation().GetFilter().asBlur()->sigma_x();
         UIVisualEffectView* visualEffectView = [[UIVisualEffectView alloc]
             initWithEffect:[UIBlurEffect effectWithStyle:UIBlurEffectStyleLight]];
+
+        // TODO: Technically we should be using the largest corner radius, but currently
```
hellohuanlin: this should be tracked by a github issue

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -586,6 +606,31 @@ - (void)applyMutators:(const flutter::MutatorsStack&)mutatorsStack
...
+        clip.rect = boundingRect;
+        impeller::RoundingRadii radii = rrect.GetRadii();
+        clip.topLeftRadius = radii.top_left.width;
+        clip.topRightRadius = radii.top_right.width;
+        clip.bottomLeftRadius = radii.bottom_left.width;
+        clip.bottomRightRadius = radii.bottom_right.width;
+        [pendingClipRRects addObject:clip];
+      }
+      case flutter::MutatorType::kBackdropClipRse: {
+        // TODO: Pending Implementation
```
hellohuanlin: same here

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
Hari-07: I don't think so

`clipsToBounds` just prevents subviews of visualEffectView from going outside its bounds, that seems safe currently

engine/src/flutter/flow/embedded_views.h
Hari-07: I made this change but right above it does say `kClipRSE`

engine/src/flutter/flow/embedded_views.h
Hari-07: Looks like this thing is irrelevant now. 
`IsClipType()` was only used in one place earlier and that got deleted a while back in https://github.com/flutter/engine/pull/53826

Currently looks like IsClipType() isn't referenced by anything. Im making this false since they're not clips on the embedded view, and just modifies the backdrop on it. But dont think it matters either way

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
Hari-07: Filed as https://github.com/flutter/flutter/issues/179125

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
Hari-07: Filed as https://github.com/flutter/flutter/issues/179126

hellohuanlin (approved):
LGTM!

LouiseHsu:
Unfortunately I think there are legit failures, if you look at 
https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/2067659/overview

```
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.ChildOutsideBounds passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.FullyContainedChild passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.LayerCached passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.OpacityInheritance passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.OpacityInheritancePainting passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.OpacityInheritanceSaveLayerPainting passed 0 / 2 times.
  /Volumes/Work/s/w/ir/cache/builder/engine/src/out/ci/host_release_arm64_tests/flow_unittests ClipPathLayerTest.OvalPathReducesToClipRRect passed 0 / 2 times.
  ```
  
  Which definitely seems related to your changes. Can you take a look?

LouiseHsu (approved):
Looks good to me! Thanks for addressing the tests~

PR #177566, opened 2025-10-27
Fix cubic subdivision estimation using correct Wang’s formula

ColdPaleLight:
Replaced the incorrect Abs()/Max() operations with squared magnitude checks via dot products.


engine/src/flutter/impeller/geometry/wangs_formula.cc
```diff
@@ -13,9 +13,8 @@ namespace {
 // X and Y directions.
 constexpr static Scalar kPrecision = 4;
 
-inline Scalar length(Point n) {
-  Point nn = n * n;
-  return std::sqrt(nn.x + nn.y);
+inline Scalar length(Vector2 n) {
```
flar: These classes provide their own length formulae. - GetLength().

engine/src/flutter/impeller/geometry/wangs_formula.cc
```diff
@@ -25,10 +24,11 @@ Scalar ComputeCubicSubdivisions(Scalar scale_factor,
...
                                 Point p2,
                                 Point p3) {
-  Scalar k = scale_factor * .75f * kPrecision;
-  Point a = (p0 - p1 * 2 + p2).Abs();
-  Point b = (p1 - p2 * 2 + p3).Abs();
-  return std::sqrt(k * length(a.Max(b)));
+  const Scalar k = scale_factor * .75f * kPrecision;
+  const Vector2 a = p0 - p1 * 2 + p2;
+  const Vector2 b = p1 - p2 * 2 + p3;
+  const Scalar max_len_sq = std::max(a.Dot(a), b.Dot(b));
```
flar: Vector2 also provides GetLengthSquared, though it takes more characters, it spells out what you are doing.

engine/src/flutter/impeller/geometry/wangs_formula_unittests.cc
```diff
@@ -0,0 +1,22 @@
...
+// found in the LICENSE file.
+
+#include "gtest/gtest.h"
+
+#include "flutter/impeller/geometry/wangs_formula.h"
+
+namespace impeller {
+namespace testing {
+
+TEST(WangsFormulaTest, Cubic) {
```
flar, line 12: Is it also possible to write a test case that tests the practical results, as in requesting a few cubic curves to flatten to various precisions and then trying to verify that the polygons match the curves to within the requested precision?

flar (commented):
Just some comments to consider for now.

flar:
Can we include a link on where to find these formulas for reference?

engine/src/flutter/impeller/geometry/wangs_formula.cc
ColdPaleLight: Done.

engine/src/flutter/impeller/geometry/wangs_formula.cc
ColdPaleLight: Done.

engine/src/flutter/impeller/geometry/wangs_formula_unittests.cc
ColdPaleLight: Directly verifying the error between the polyline and the curve in a test case is actually more complicated than it might seem.

Wang’s formula is rigorously derived using the curve’s derivatives, so it already guarantees the polyline stays within the requested precision. Checking that it returns a reasonable number of segments is usually sufficient.

For the full derivation, see:
Goldman, Ron. (2003). 5.6.3 Wang's Formula. Pyramid Algorithms. Morgan Kaufmann Publishers.

ColdPaleLight:
> Here is another way to compute subdivisions, but I haven't read it closely or vetted it. We've used Raph's math in other areas to good results, so it bears investigating if you have the time.
> 
> https://raphlinus.github.io/graphics/curves/2019/12/23/flatten-quadbez.html


Thanks for sharing the link! I’ve actually looked into this approach already. It turns out it’s a bit more computationally expensive than Wang’s formula. Nical’s benchmarks show the numbers pretty clearly:
https://github.com/nical/bezier_flattening/blob/main/notes/bench-cubic-threadripper.md

So for our purposes, Wang’s formula still seems like the more efficient choice, and it already gives solid error guarantees.

flar:
> > Can we include a link on where to find these formulas for reference?
> 
> Just to clarify, unfortunately, Wang’s formula can’t be viewed through a simple link online—it’s only fully available in books. The most accessible reference is: Goldman, Ron. (2003). 5.6.3 Wang's Formula. Pyramid Algorithms. Morgan Kaufmann Publishers.
> 
> The formula was originally introduced by Guozhao Wang in the 1980s, but I couldn’t find a way to access the original paper online. Goldman’s book provides both the derivation and practical explanation.

I discovered the same looking for a link yesterday. I think a biblio reference at least clarifies what we are modeling after even if they can't click and read it directly.

flar (approved):
LGTM

flar (changes requested):
This PR is fine to go in except for one issue which is the presence of a rebase on a PR with goldens. We still don't know exactly what causes the problem, but it is possible for golden approvals to get "lost in the system" when there are rebases in the PR.

So, to prevent a nightmare of "where did the approved goldens go", I'm going to suggest that you resubmit these changes as a "clean PR" with a single commit and I'll approve that one and we can triage the goldens there and submit and avoid any problems. Don't create it from a rebase or anything, instead abandon this branch, squash it into a single commit, then go to ToT and cherry-pick the one commit into a brand new branch. That will hopefully give a clean base for the golden approvals to succeed.

Apologies that we have this glitch in the system, but I have another PR that I cannot submit because it never sees the golden failures during PR testing, but runs into them after submitting and I am still trying to figure how to get that work into the system.

ColdPaleLight:
> This PR is fine to go in except for one issue which is the presence of a rebase on a PR with goldens. We still don't know exactly what causes the problem, but it is possible for golden approvals to get "lost in the system" when there are rebases in the PR.
> 
> So, to prevent a nightmare of "where did the approved goldens go", I'm going to suggest that you resubmit these changes as a "clean PR" with a single commit and I'll approve that one and we can triage the goldens there and submit and avoid any problems. Don't create it from a rebase or anything, instead abandon this branch, squash it into a single commit, then go to ToT and cherry-pick the one commit into a brand new branch. That will hopefully give a clean base for the golden approvals to succeed.
> 
> Apologies that we have this glitch in the system, but I have another PR that I cannot submit because it never sees the golden failures during PR testing, but runs into them after submitting and I am still trying to figure how to get that work into the system.

Thanks for your information! Here is the "clean PR" https://github.com/flutter/flutter/pull/177758

Issue #169486, opened 2025-05-27
[Android] button actions in webview_flutter are blocked


letrungdo:
### Steps to reproduce

1. Open the screen containing the webview
2. Use 2 or more fingers to zoom in and out of the map

### Expected results

After zooming in and out, tap the buttons work normally.

### Actual results

Tap the buttons and no action will happen.

### Code sample

You can reproduce it on this repo:
https://github.com/letrungdo/flutter_webview_issue

I have wrapped the WebViewWidget in SingleTouchContainer for easy reproducibility. If I remove it it will be harder to reproducible and it still has the issue on

### Screenshots or Video

https://github.com/user-attachments/assets/51a2ea17-27de-4c74-9742-55e946a185d5

### Logs

No logs

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.32.0, on macOS 15.4.1 24E263 darwin-arm64, locale en-JP) [290ms]
    • Flutter version 3.32.0 on channel stable at /Users/letrungdo/fvm/versions/3.32.0
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision be698c48a6 (7 days ago), 2025-05-19 12:59:14 -0700
    • Engine revision 1881800949
    • Dart version 3.8.0
    • DevTools version 2.45.1

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [2.5s]
    • Android SDK at /Users/letrungdo/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • Java binary at:
      /Users/letrungdo/Library/Java/JavaVirtualMachines/corretto-17.0.13/Contents/Home/bin/java
      This JDK is specified in your Flutter configuration.
      To change the current JDK, run: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment Corretto-17.0.13.11.1 (build 17.0.13+11-LTS)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.3) [719ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16E140
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [12ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.2) [11ms]
[...]

PR #162895, opened 2025-02-07
Remove `android_verified_input`, it was never added to CI and does not run locally.

matanlurey:
Someone could always resurrect it via `git`-magic (or GitHub), but let's not maintain (including Gradle lock files, Dart analysis, etc) unused code that hasn't run in 12+ months.

/cc @johnmccutchan

PR #177572, opened 2025-10-27
fix: gesture blocked in android webview after zoom

letrungdo:
- Fix issue https://github.com/flutter/flutter/issues/169486


gmackall (changes requested):
Thank you for taking the time to put together this contribution. Unfortunately we cannot accept a change which regresses the verification status of these input events.

reidbaker:
> Thank you for taking the time to put together this contribution. Unfortunately we cannot accept a change which regresses the verification status of these input events.

You can read more about verification of input events here. 
https://developer.android.com/reference/android/hardware/input/InputManager#verifyInputEvent%28android.view.InputEvent%29

engine/src/flutter/shell/platform/android/io/flutter/plugin/platform/PlatformViewsController.java
```diff
@@ -731,19 +712,38 @@ public MotionEvent toMotionEvent(
...
     // will fail the verifiable input check.
     PointerProperties[] pointerProperties =
         parsePointerPropertiesList(touch.rawPointerPropertiesList)
             .toArray(new PointerProperties[touch.pointerCount]);
 
+    if (!usingVirtualDiplay && trackedEvent != null) {
+      // We need to use Flutter's calculated coordinates which may include complex transformations
+      // (zoom, scale, rotation) that can't be represented by simple offsetLocation().
+      // Creating a new MotionEvent with the correct coordinates ensures gestures work properly
+      // after transformations like zoom, but loses the verifiable input flag.
```
gmackall: For the record, I think this is not a correct description of the mechanism by which this change fixes the problem. If we print out coordinates for a given pointer, for both the original translated event and new constructed event (i.e. the event below), you will see the coordinates are identical.

Rather, the problem is that the SingleTouchContainer you added is pruning input events, leading to a different pointer count between the two `MotionEvents`, as we can see if we print out some of the contents of these two events:

```
E/HI GRAY,( 8361): tracking a new motion event: io.flutter.embedding.android.MotionEventTracker$MotionEventId@94abf41

E/HI GRAY ( 8361): original event info: 
E/HI GRAY ( 8361): pointer count is: 2
E/HI GRAY ( 8361): event action is: 2
-------------- LOOPING OVER POINTER COUNT --------------
E/HI GRAY ( 8361): pointer properties are: PointerProperties(id = 0, tooltype = 1, )
E/HI GRAY ( 8361): pointer coords are: (x=264.75,y=498.0) and size=0.0847811
E/HI GRAY ( 8361): point id is: 0
E/HI GRAY ( 8361): pointer properties are: PointerProperties(id = 1, tooltype = 1, )
E/HI GRAY ( 8361): pointer coords are: (x=392.25,y=1228.5) and size=0.09659486
E/HI GRAY ( 8361): point id is: 1


E/HI GRAY ( 8361): constructed event info: 
E/HI GRAY ( 8361): pointer count is: 1
E/HI GRAY ( 8361): event action is: 2
-------------- LOOPING OVER POINTER COUNT --------------
E/HI GRAY ( 8361): pointer properties are: PointerProperties(id = 0, tooltype = 1, )
E/HI GRAY ( 8361): pointer coords are: (x=264.75,y=498.0) and size=0.0847811
E/HI GRAY ( 8361): point id is: 0
```

Unfortunately it does not seem there is an api that would allow us to modify the pointer count of the original event.

letrungdo: Thanks @gmackall for the clarification! You're right - the issue is pointer count mismatch, not coordinates.

## Updated Fix

I've revised the implementation to **preserve verifiable input whenever possible**:

```java
if (trackedEvent.getPointerCount() == touch.pointerCount) {
    // Counts match → preserve verifiable flag
    translateMotionEvent(trackedEvent, pointerCoords);
    return trackedEvent;
}
// Counts mismatch → must reconstruct (loses verifiable flag)
return MotionEvent.obtain(...);
```
This minimizes security impact by only sacrificing the verifiable flag when pointer counts actually mismatch (e.g., gesture recognizer filtering).
Why Accept This Trade-off?
No alternative: Android has no API to modify pointer count while preserving verifiability
Functional > theoretical security: App is completely unusable after zoom without this fix
Minimal scope: Only affects pointer count mismatch cases, not normal gestures
Already validated: Events passed through Flutter framework validation
Would this approach be acceptable given the Android platform constraint?

gmackall: Yeah I think the approach of preserving the verifiability when possible, but falling back when not possible, is a good solution here (it was also the approach I was going to suggest based on some offline discussions).

The original integration test for this behavior was regrettably reverted without ever being relanded https://github.com/flutter/flutter/pull/140836/, https://github.com/flutter/flutter/pull/162895, so it would be easy to accidentally introduce a regression here where we are falling back in all cases to unverified events. But I think this approach is safe enough to land, in my opinion.

@reidbaker WDYT?

gmackall (approved):

engine/src/flutter/shell/platform/android/io/flutter/plugin/platform/PlatformViewsController.java
```diff
@@ -731,19 +731,58 @@ public MotionEvent toMotionEvent(
...
+      // Trade-off justification:
+      // - Functional correctness: Without this fix, gestures become completely blocked
+      //   after zoom operations, making the app unusable (see issue #169486)
+      // - Security impact minimal: The event has already passed through Flutter
+      //   framework validation before reaching this point
+      // - No alternative: Android does not provide an API to modify MotionEvent
+      //   pointer count while preserving the verifiable flag
+      //
+      // This approach ensures the app remains functional while minimizing security
+      // impact by only sacrificing verifiability when absolutely necessary.
```
gmackall: nit: I don't think we need to leave a tradeoff justification in the code, please delete

letrungdo: @gmackall I deleted the tradeoff comment.

gmackall (approved):
LGTM, manually confirmed that this passes the verified input test.

reidbaker (approved):

PR #177579, opened 2025-10-27
Fix ElevatedButton.icon breaks focus traversal and VoiceOver when toggling icon

bleroux:
## Description

This PR changes `ElevatedButton.icon` to avoid building a different widget. When a different widget is created the whole subtree is recreated which leads to various issues (Focus and A11y issues for instance).
The change is similar to https://github.com/flutter/flutter/pull/175810 which fixed the exact same problem for `OutlinedButton.icon`.

## Related Issue

[TextButton.icon breaks focus traversal and ink effect when toggling icon](https://github.com/flutter/flutter/issues/173944)
[Voiceover focus traversal breaks if a button's state changes to include an icon](https://github.com/flutter/flutter/pull/175810) 

## Tests

- Adds 2 tests


hannah-hyj (approved):
LGTM

PR #177604, opened 2025-10-27
Move glyph bounds calculation to creation of text frame.

eyebrowsoffire:
This is a draft testing out moving the computation of glyph bounds to when we create the text frame, instead of later when rasterization happens.


Piinks:
Greetings once more from stale triage, is this still active? :)

eyebrowsoffire:
Turns out, after further investigation, there are still more call sites deep inside Skia which use the shared strike cache that are unavoidable. So this approach fundamentally isn't going to work. We're going to need support from Skia to resolve this problem.

Issue #174691, opened 2025-08-29
CupertinoPicker ticking sound is playing at full volume through headphones


UCetinkaya94:
### Steps to reproduce

1. Connect headphones to your device
2. Under phone settings -> **_Sounds & Haptics_** set **_Ringtone and Alerts_** to the max volume
3. Open a Flutter app using `CupertinoPicker` and start scrolling

Tested in release mode on an iPhone 14 Pro (iOS 26 public beta)
and iPhone XR (iOS 18.1.1)

### Expected results

The ticking sound plays at a reasonable volume through the headphones or not at all, like the stock clock app on iOS  

### Actual results

The ticking sound plays at the max volume through the headphones which is very uncomfortable

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/cupertino.dart';
import 'package:flutter/material.dart';

void main() => runApp(MaterialApp(home: HomeScreen()));

class HomeScreen extends StatelessWidget {
  const HomeScreen({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Center(
        child: CupertinoTimerPicker(
          onTimerDurationChanged: (_) {},
        ),
      ),
    );
  }
}
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.35.2, on macOS 15.6.1 24G90 darwin-arm64, locale de-DE) [306ms]
    • Flutter version 3.35.2 on channel stable at /Users/ufukcetinkaya/Developer/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 05db968908 (4 days ago), 2025-08-25 10:21:35 -0700
    • Engine revision a8bfdfc394
    • Dart version 3.9.0
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-swift-package-manager, enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [1.125ms]
    • Android SDK at /Users/ufukcetinkaya/Library/Android/sdk
    • Emulator version 35.4.9.0 (build_id 13025442) (CL:N/A)
    • Platform android-36, build-tools 35.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
[...]

PR #177643, opened 2025-10-28
Add `enableSound` flag to `CupertinoPicker`

alex-medinsh:
I am not sure how to replicate the native iOS behaviour regarding the ticking sounds yet. For now here is a flag to opt-out of the ticking altogether.

Towards https://github.com/flutter/flutter/issues/174691

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/cupertino/picker.dart
+++ b/packages/flutter/lib/src/cupertino/picker.dart
@@ -94,6 +94,7 @@ class CupertinoPicker extends StatefulWidget {
     required this.onSelectedItemChanged,
     required List<Widget> children,
     this.selectionOverlay = const CupertinoPickerDefaultSelectionOverlay(),
+    this.soundEnabled = true,
     bool looping = false,
   }) : assert(diameterRatio > 0.0, RenderListWheelViewport.diameterRatioZeroMessage),
        assert(magnification > 0),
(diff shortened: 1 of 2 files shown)


Piinks (commented):
> For now here is a flag to opt-out of the ticking altogether.

Is this meant to be a temporary flag?

justinmc (commented):
Just a drive-by review to say that we have a bunch of widgets currently with `enableFeedback` parameters that cover both haptics and audio, e.g. [ButtonStyle.enableFeedback](https://main-api.flutter.dev/flutter/material/ButtonStyle/enableFeedback.html). Maybe this should have the same name and should also disable the haptics? Or if it stays audio-only, maybe it should be named `enableAudio` to be more aligned with the name of `enableFeedback`.

alex-medinsh:
Hi @Piinks, @justinmc!
I feel like my current implementation left the component in a bad state (which I apologize for), since the ticking sound might be quite unpleasant to headphone users, which might turn some app's users away. 

I don't currently know of an easy way to make the ticking sound follow the native iOS behavior, namely that it does not play the sound at all when headphones are connected, and unfortunately I don't currently have the bandwidth to investigate a proper solution, so the least I could do is add a flag to allow developers to opt-out of the sound. It can be a temporary flag until we find a better solution, or it can stay, maybe there is value in allowing the developer to disable or even customize the sound.

@justinmc good call on the name, I will update it to align with similar parameters. As for also controlling the haptics, I am not sure. On one hand there are similar toggles in other widgets, as you mentioned, on the other hand I think that some developers might still want to have the haptics enabled, but opt-out of the sound, since in it's current state it can be unpleasant to the users.

victorsanni (approved):
I think it's fair to give developers the option to disable the sound until the bug where it plays on full volume on headphones is fixed.

Piinks:
> until the bug where it plays on full volume on headphones is fixed.

Is this a bug on our side or iOS? We do not usually land temporary workarounds, rather we embrace the yak shave and fix the root issue. :)

MitchellGoodwin:
I agree with Kate that adding this for the purpose of a workaround isn't ideal. But in my opinion, adding it for the sake of customizability like other widgets have sounds good to me.

PR #157399, opened 2024-10-23
Reland "Fix TextField selects all content after the application is resumed"

bleroux:
## Description

Relands https://github.com/flutter/flutter/pull/156968 wich was reverted in https://github.com/flutter/flutter/pull/157378 

This PR makes `EditableText` aware of the lifecycle 'resumed' state to let the current selection unchanged when the application is resumed (on web and desktop, 'resumed' means the Flutter app window regained focus).

Before this PR, on web and desktop, the whole content of a `TextField` was selected whenever a `TextField` gained focus. This is the correct behavior when tabbing between fields but it is not when a field regains focus after the application is resumed 

## Related Issue

Fixes [When switching to another browser tab or window and then going back, all text on TextField is selected automatically](https://github.com/flutter/flutter/issues/156078).

## Tests

Adds 1 test.

PR #177658, opened 2025-10-28
Fix EditableText _justResumed is not accurate

bleroux:
## Description

This PR tweaks the selection logic added in https://github.com/flutter/flutter/pull/157399.
Before this PR (and since https://github.com/flutter/flutter/pull/157399) when the app resumed while a TextField was selected the selection of the TextField is maintained. This is the right behavior for the currently focused TextField. But when there are several TextFields, after the app resumed and the user move the focus to a another TextField, the behavior should be to select all the content of the newly focused TextField. To achieve this the `_justResumed`flag added in https://github.com/flutter/flutter/pull/157399 should be reset as soon as the focus move as it is needed only for the current focused TextField to restore its selection just after the app resumed.

## Related Issue

Fixes [Pressing tab does select all content when app is resumed for TextFields which were not focused](https://github.com/flutter/flutter/issues/177650)

## Tests

- Adds 1 test

### DIFF
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -3214,8 +3214,21 @@ class EditableTextState extends State<EditableText>
     widget.focusNode.addListener(_handleFocusChanged);
     _cursorVisibilityNotifier.value = widget.showCursor;
     _spellCheckConfiguration = _inferSpellCheckConfiguration(widget.spellCheckConfiguration);
-    _appLifecycleListener = AppLifecycleListener(onResume: () => _justResumed = true);
+    _appLifecycleListener = AppLifecycleListener(onResume: _onResume);
     _initProcessTextActions();
+  }
+
+  void _onResume() {
+    _justResumed = true;
+    // To prevent adding multiple listeners, remove any existing one first.
+    FocusManager.instance.removeListener(_resetJustResumed);
+    // Reset _justResumed as soon as there is a focus change.
+    FocusManager.instance.addListener(_resetJustResumed);
+  }
+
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -3214,10 +3214,23 @@ class EditableTextState extends State<EditableText>
...
+    _appLifecycleListener = AppLifecycleListener(onResume: _onResume);
     _initProcessTextActions();
   }
 
+  void _onResume() {
+    _justResumed = true;
+    // To prevent adding multiple listeners, remove any existing one first.
+    FocusManager.instance.removeListener(_resetJustResumed);
+    // Reset _justResumed as soon as there is a focus change.
+    FocusManager.instance.addListener(_resetJustResumed);
```
justinmc, line 3226: Ah good call. Those should be all the situations where we want to reset _justResumed that I can think of...

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -16620,12 +16623,82 @@ void main() {
       expect(focusNode.hasFocus, true);
       expect(controller.selection, collapsedAtEnd('Flutter!').selection);
 
-      await setAppLifeCycleState(AppLifecycleState.inactive);
-      await setAppLifeCycleState(AppLifecycleState.resumed);
+      await setAppLifecycleState(AppLifecycleState.inactive);
```
justinmc, line 16626: I think I agree with you that it should be Lifecycle and not LifeCycle, but it looks like we have both all over the code base 🤷

justinmc (approved):
LGTM 👍

PR #177659, opened 2025-10-28
Add discussion of federated plugin documentation

stuartmorgan-g:
Adds a canonical reference for the guidance we give in reviews about how to document platform-specific behaviors in federated plugins, and the reason for that guidance.


bparrishMines (approved):
LGTM

Issue #126787, opened 2023-05-14
Tooltip controller to manipulate its visibility and such things


mhbdev:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)
- [X] I have read the [guide to filing a bug](https://flutter.dev/docs/resources/bug-reports)

### Use case

 I was working on a widget that should connect to the server to bring the Tooltip message to the client. after fetching the data I needed to show the tooltip manually because its triggerMode has been set to `.tap`.

### Proposal

I was thinking of a controller to pass into the Tooltip constructor.

Controller be like:
```dart
typedef TooltipVisibilityCallback = void Function({bool immediately});

class TooltipController extends ChangeNotifier {
  TooltipController();

  late TooltipVisibilityCallback _open;

  TooltipVisibilityCallback get open => _open;

  set open(TooltipVisibilityCallback value) {
    notifyListeners();
    _open = value;
  }
}
```

Usage:
```dart
_controller.open(immediately: true);

Tooltip(
    controller: _controller,
);
```

We have to set Controller's `open` callback on Tooltip's `initState`
```dart
    widget.controller?.open = _showTooltip;
```

If you think the idea of implementing it like I told, seems ok, then I can work on this feature and will make a PR for that.

danagbemava-nc:
Hi @mhbdev, you can manually trigger a tooltip using the tooltip key. See https://api.flutter.dev/flutter/material/Tooltip-class.html#material.Tooltip.4

Is there any limitation you're facing using this approach?

mhbdev:
> Hi @mhbdev, you can manually trigger a tooltip using the tooltip key. See https://api.flutter.dev/flutter/material/Tooltip-class.html#material.Tooltip.4
> 
> Is there any limitation you're facing using this approach?

I know that but I think it would be muck more cleaner. I was working with MenuAnchor widget and saw the controller of it and I was wondering why whould they create a controller for MenuAnchor while we can open or close it using a globalKey.

Issue #97496, opened 2022-01-30
☂️ Reinforcement: Add more basic components to the core framework


Hawmex:
This issue tracks proposals and development for net-new widgets and functionalities to be added to the core widgets library.

This effort is the other side of the coin to our reinforcement work, see the [refactoring side](https://github.com/flutter/flutter/issues/53059) as well. While the refactoring issue focuses on moving existing code, this issue is dedicated to identifying and building new, foundational, style-agnostic primitives that are missing from the core widgets library today. The goal is to fill gaps in the core and make it an even more powerful foundation for building any design system.

> [!NOTE]
> Some of these may be best informed by a UXR study. Before considering contributing one of these net-new widgets, reach out to the team on Discord to confirm.

### Propose a New Core Widget! 💡

What foundational widget or capability do you wish existed in Flutter's core? If you have an idea for a new, un-opinionated "headless" widget that would be broadly useful, this is the place to propose it. Please provide a clear use case and rationale for why it belongs in the core framework.

### _Last update Mar 4, 2026_

So far, we have investigated many of the existing design components to determine if any refactoring can be done *before* decoupling, moving existing APIs to the widgets layer to a common surface (which is tracked in [https://github.com/flutter/flutter/issues/53059](https://github.com/flutter/flutter/issues/53059)). In some cases, we have decided not to break widgets in the design libraries to try to accomplish this, which leaves the opportunity for new, more generic APIs to be introduced in the widgets layer later.

As we have been focused on the opportunity to refactor existing widgets (#53059) ahead of actually separating the libraries, expect this work of adding _new_ widgets to continue *after* the design libraries land in their new packages, which is what this issue is meant to track.

To check out all the issues and aspects of the decoupling effort, see the [github project](https://github.com/orgs/flutter/projects/220), and the root parent issue: https://github.com/flutter/flutter/issues/101479.

#### Status:

- What's landed or being worked on wrt raw widgets:  
  - **Menus**  
    - [RawMenuAnchor](https://main-api.flutter.dev/flutter/widgets/RawMenuAnchor-class.html)  
    - [RawMenuAnchorGroup](https://main-api.flutter.dev/flutter/widgets/RawMenuAnchorGroup-class.html)  
    - [RawMenuOverlayInfo](https://main-api.flutter.dev/flutter/widgets/RawMenuOverlayInfo-class.html)  
  - **Radios**  
    - [RawRadio](https://main-api.flutter.dev/flutter/widgets/RawRadio-class.html)  
[...]

Issue #53059, opened 2020-03-22
☂️ Reinforcement: Refactor widgets from design into the core before decoupling


filleduchaos:
This issue tracks the essential refactoring work required to strengthen the core widgets library _**before**_ decoupling the material and cupertino libraries (https://github.com/flutter/flutter/issues/101479).

We will prioritize net-new additions **_after_** decoupling. 
**🔍 For feedback on net-new widgets to strengthen the core, see: link**

---

Our goal is to identify and move foundational, style-agnostic logic out of the existing design system libraries and into the core framework. This will ensure the core is robust enough to support multiple high-quality design systems and will provide a richer set of primitives for all developers.

> [!NOTE]
It is important to note that not _everything_ in the design libraries will be refactored into the core. Some widgets are too heavily tied to their design language to extract a meaningful, un-opinionated API that would satisfy developer needs. One of our guiding principles is to avoid the "[lowest common denominator](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-the-lowest-common-denominator)" problem, ensuring that any new core widget is genuinely powerful and flexible. 

### Community Feedback Needed 🗣️

This is a critical opportunity for community input. Please use this issue to provide feedback and suggest other widgets or functionalities you believe are fundamental and should be moved into the core widgets library before the main decoupling occurs.

#### Widgets we are investigating/refactoring before decoupling:

- https://github.com/flutter/flutter/issues/172929
- https://github.com/flutter/flutter/issues/172930
- https://github.com/flutter/flutter/issues/172931

#### Widgets that are unlikely to be refactored, will migrate as-is:

- **ColorScheme**
  - Very material, uses material color utilities package to generate material spec color scheme.
- **InkWell**
  - Relies on Material ancestor to paint, too complex to refactor to a basic component that paints in response to a gesture, which could be done in a different/simpler way
- **Scaffold**
  - Very material specific for layout. We discussed focusing on simple app bar, bottom bar widgets might be nice to have instead with built in safe area handling. A column could compose something similar to a Scaffold layout if the simple *bar components handled things like SafeArea out of the box. Simple composition from the widgets layer is still our focus.
- **SliverAppBar**
  - Highly complex and material-specific
  - SliverPersistentHeader, its base class is already in widgets, We added several atomic sliver header widgets last year that are meant to replace the SliverAppBar behemoth through composition in the widgets layer.
[...]

PR #83830, opened 2021-06-02
Make tooltip hoverable and dismissible

chunhtai:
fixes https://github.com/flutter/flutter/issues/75885

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #177678, opened 2025-10-28
Raw tooltip with smaller API surface that exposes tooltip widget

### LINKED ISSUE #45034: Custom Overlay for Tooltip Widget
Internal: b/148141781

## Use case

I want to provide a "bubble" in the bottom of  **Tooltip**. And also, "X" icon to close the tooltip. Currently,  Tooltip class doesn't provide any property to customize overlay. It is showing a simple Text widget inside a ConstrainedBox. It also lacks a way of showing/hiding tooltip programmatically. 

I'm achieving my use case with basically **copying the whole Tooltip class** and changing the __TooltipOverlay_ [build](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/tooltip.dart#L493) method. 

Tooltip decoration property is only useful to change the shape and background color. There is also no way to show/hide tooltip programmatically.

## Proposal

How about providing a **overlay** parameter in Tooltip?
If it is not provided, `_TooltipOverlay` can be used as a default overlay. 

I would be happy to give a PR with tests. I don't want to copy the whole **Tooltip** class from the framework to just modify a single method!

And also, exposing show/hide tooltip would make the widget very much useful!

victorsanni:
The raw widget used to make a [Tooltip](https://api.flutter.dev/flutter/material/Tooltip-class.html).

Design doc: flutter.dev/go/codeshare-tooltip

## Constructor
```dart
class RawTooltip extends StatefulWidget {
  const RawTooltip({
    super.key,
    required this.semanticsTooltip,
    required this.tooltipBuilder,
    this.enableTapToDismiss = true,
    this.triggerMode = TooltipTriggerMode.longPress,
    this.enableFeedback = true,
    this.onTriggered,
    this.hoverDelay = Duration.zero,
    this.touchDelay = const Duration(milliseconds: 1500),
    this.dismissDelay = const Duration(milliseconds: 100),
    this.animationStyle = _kDefaultAnimationStyle,
    this.positionDelegate,
    required this.child,
  });
```

## Properties
```dart
  final String semanticsTooltip;
  final TooltipComponentBuilder tooltipBuilder;
  final Duration hoverDelay;
  final Duration touchDelay;
  final Duration dismissDelay;
  final bool enableTapToDismiss;
  final TooltipTriggerMode triggerMode;
  final bool enableFeedback;
  final TooltipTriggeredCallback? onTriggered;
  final AnimationStyle animationStyle;
  final TooltipPositionDelegate? positionDelegate;
  final Widget child;
```

Part of [☂️ Reinforcement: Add more basic components to the core framework](https://github.com/flutter/flutter/issues/97496)
Part of [☂️ Reinforcement: Refactor widgets from design into the core before decoupling](https://github.com/flutter/flutter/issues/53059)
Fixes [Custom Overlay for Tooltip Widget ](https://github.com/flutter/flutter/issues/45034)


packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,819 @@
...
+/// [_ExclusiveMouseRegion], but moves outside of the bounds of the inner
+/// [_ExclusiveMouseRegion] that was initially blocking the outer widget.
+///
+/// Likewise, [onExit] is called when the a mouse pointer moves out of the paint
+/// bounds of this widget, or moves into another [_ExclusiveMouseRegion] that
+/// overlaps this widget in hit-testing order.
+///
+/// This widget doesn't affect [MouseRegion]s that aren't [_ExclusiveMouseRegion]s,
+/// or other [HitTestTarget]s in the tree.
+class _ExclusiveMouseRegion extends MouseRegion {
```
loic-sharma, line 175: Like you mentioned in the draft design doc, `RawTooltip` is visually unopinionated, but behaviorally opinionated. If a custom design system needs a different behavior, they might need to fork the logic in `RawTooltip`. 

Do you think `_ExclusiveMouseRegion`, `_TooltipOverlay`, and `_TooltipPositionDelegate` make sense as public APIs? Assuming these can be standalone APIs, that would make it easier to fork `RawTooltip` to customize its behavior.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,873 @@
...
+    this.enableFeedback = true,
+    this.onTriggered,
+    this.mouseCursor,
+    this.ignorePointer = false,
+    this.waitDuration = Duration.zero,
+    this.showDuration = const Duration(milliseconds: 1500),
+    this.exitDuration = const Duration(milliseconds: 100),
+    this.animationStyle = _kDefaultAnimationStyle,
+    this.positionDelegate,
+    this.child,
```
justinmc: So the child is typically the thing that needs to be hovered to show the tooltip, right? Does it actually make sense to support no child or should it be required?

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: This is a good idea. We should probably design these things as if they could be public anyway.

justinmc: To me a tooltip consists of:

 * The UI that shows the tooltip when you hover it (`child`).
 * The tooltip itself (tooltipBuilder). Could be split between the frame and the contents but that's probably for a design language to do.
 * The positioning logic (_TooltipOverlay, which includes _TooltipPositionDelegate).
 * The logic for how to listen to user events and decide when to show the tooltip (_ExclusiveMouseRegion plus Listener).
 * Semantics.

I think a user could reasonably want to customize any of those things besides Semantics. It would be cool to write a RawTooltip that allowed you to customize all of that, but that's a tall order, and it's not easy to get it right. You could always delay any/all of that to another PR, or even wait for feedback from users.

justinmc (commented):
Some thoughts about the API and making things public.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: With this API, the only one out of the 5 above that is not customizable is the _ExclusiveMouseRegion plus Listener logic. That logic is not as localized as the others, it wraps both the child and the _TooltipOverlay, and exposing it will require exposing more private methods like _scheduleShowTooltip.

Right now, I think we should wait for an actual user problem to get more clarity on how we can best expose the the _ExclusiveMouseRegion logic. It's kicking the can down the road, but if we eventually have to address it we will at least have a clearer scope.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Requiring the child lgtm. Currently `Tooltip` doesn't require a child, I don't know if this is because of `ensureTooltipVisible` which allows for programmatic triggering of the tooltip. But looking through the code I see the child null coalesces to SizedBox.shrink, so it's probably a good idea to require it.

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -962,242 +491,86 @@ class TooltipState extends State<Tooltip> with SingleTickerProviderStateMixin {
...
+      enableFeedback:
+          widget.enableFeedback ?? _tooltipTheme.enableFeedback ?? _defaultEnableFeedback,
+      waitDuration: widget.waitDuration ?? _tooltipTheme.waitDuration ?? _defaultWaitDuration,
+      enableTapToDismiss: widget.enableTapToDismiss,
+      onTriggered: widget.onTriggered,
+      mouseCursor: widget.mouseCursor,
+      ignorePointer: widget.ignorePointer ?? widget.message != null,
+      exitDuration: widget.exitDuration ?? _tooltipTheme.exitDuration ?? _defaultExitDuration,
+      positionDelegate: _getDefaultPositionDelegate,
+      child: widget.child ?? const SizedBox.shrink(),
```
chunhtai, line 520: can make widget.child non nullable and default to const SizedBox.shrink() in constructor

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,972 @@
...
+///  * [TooltipPositionDelegate], which uses this context to compute tooltip positions.
+@immutable
+class TooltipPositionContext {
+  /// Creates a tooltip position context.
+  const TooltipPositionContext({
+    required this.target,
+    required this.targetSize,
+    required this.tooltipSize,
+    required this.verticalOffset,
+    required this.preferBelow,
```
chunhtai: preferBelow is passed into the RawTooltip and piped here to be used by `TooltipPositionDelegate` seems unecessary, if material design tooltip has already known the preferBelow they could have uses it directly in `TooltipPositionDelegate` without going through the TooltipPositionContext

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,972 @@
...
+      Object.hash(target, targetSize, tooltipSize, overlaySize, verticalOffset, preferBelow);
+}
+
+/// Overrides the visibility of descendant [RawTooltip] widgets.
+///
+/// If disabled, the descendant [RawTooltip] widgets will not display a tooltip
+/// when tapped, long-pressed, hovered by the mouse, or when
+/// `ensureTooltipVisible` is called. This only visually disables tooltips but
+/// continues to provide any semantic information that is provided.
+class TooltipVisibility extends StatelessWidget {
```
chunhtai: I am not sure why we have this widget, but this seems like something that can remain as material concept for now until there is a clear use case.

chunhtai (commented):
I only took a quick glance, will finish the review later

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: RawTooltip checks `TooltipVisibility.of(context)` internally before displaying the tooltip.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: But TooltipPositionContext is already a public API with preferBelow required, so not piping it through will be a breaking change? I'm a little confused on whats the ask here.

packages/flutter/lib/src/widgets/tooltip.dart
chunhtai: we could implement it just for material version of tooltip, by returning Size.shrink in the tooltip builder

packages/flutter/lib/src/material/tooltip.dart
victorsanni: Actually this is causing several test failures, a lot of widgets pass a nullable child and even though in theory if the child is null it should get resolved, the type check still fails.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: I think there is a use case for RawTooltip, in the sense that a user can wrap a widget subtree with this widget to guarantee that no descendant tooltip will be displayed. 

> we could implement it just for material version of tooltip, by returning Size.shrink in the tooltip builder

I don't think this is possible, `TooltipVisibility.of(context)` sets a `_visible` bool that is checked before `_scheduleShowTooltip` and the gesture handlers. Won't moving it to Material require us to move that logic also?

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  });
+
+  /// The text to display in the tooltip's semantics announcement.
+  final String semanticsTooltip;
+
+  /// Builds the widget that will be displayed in the tooltip's overlay.
+  ///
+  /// The returned widget will be wrapped in an [IgnorePointer] widget if
+  /// [ignorePointer] is true.
+  final TooltipComponentBuilder tooltipBuilder;
```
justinmc, line 295: Can you explain the Animation parameter in the docs here?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  /// {@endtemplate}
+  final MouseCursor? mouseCursor;
+
+  /// Whether this tooltip should be invisible to hit testing.
+  ///
+  /// See also:
+  ///
+  /// * [IgnorePointer], for more information about how pointer events are
+  /// handled or ignored.
+  final bool ignorePointer;
```
justinmc: How common is this? I'm wondering if it's worth it or not to build this in. Users could also just wrap their tooltipBuilder in IgnorePointer themselves.

If you keep it, include a line about "Defaults to false."

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  ///  actions.
+  /// {@endtemplate}
+  final bool enableFeedback;
+
+  /// {@template flutter.widgets.RawTooltip.onTriggered}
+  /// Called when the Tooltip is triggered.
+  ///
+  /// The tooltip is triggered after a tap when [triggerMode] is
+  /// [TooltipTriggerMode.tap] or after a long press when [triggerMode] is
+  /// [TooltipTriggerMode.longPress].
```
justinmc, line 372: What about when hovered with a mouse? It seems like that does not cause onTriggered to be called. Maybe explicitly mention that.

justinmc: Clicking with a mouse also doesn't call this.

justinmc: Maybe include a "See also" that points to triggerMode.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  /// If null, the default positioning behavior is used (above or below the
+  /// target).
+  ///
+  /// See also:
+  ///
+  ///  * [TooltipPositionContext], which contains the positioning parameters.
+  ///  * [TooltipPositionDelegate], the function signature for custom
+  ///  positioning.
+  /// {@endtemplate}
+  final TooltipPositionDelegate? positionDelegate;
```
justinmc, line 448: Is there ever any positioning logic that this can't handle? I wonder if anyone would want to be able to pass the whole overlayChildBuilder.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+
+  /// The widget below this widget in the tree.
```
justinmc, line 450: Maybe explain that this is the "target" thing that triggers the tooltip to appear.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+    }
+    _activeHoveringPointerDevices.remove(event.device);
+    if (_activeHoveringPointerDevices.isEmpty) {
+      _scheduleDismissTooltip(withDelay: widget.exitDuration);
+    }
+  }
+
+  /// Shows the tooltip if it is not already visible.
+  ///
+  /// After made visible by this method, The tooltip does not automatically
```
justinmc: "The" => "the"

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  ///
+  /// This method returns true if it successfully dismisses the tooltips. It
+  /// returns false if there is no tooltip shown on the screen.
+  static bool dismissAllToolTips() {
+    if (_openedTooltips.isNotEmpty) {
+      // Avoid concurrent modification.
+      final List<RawTooltipState> openedTooltips = _openedTooltips.toList();
+      for (final state in openedTooltips) {
+        assert(state.mounted);
+        state._scheduleDismissTooltip(withDelay: Duration.zero);
```
justinmc: Nit: Maybe default withDelay to zero?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+
+  static final List<RawTooltipState> _openedTooltips = <RawTooltipState>[];
+
+  /// Dismiss all of the tooltips that are currently shown on the screen,
+  /// including those with mouse cursors currently hovering over them.
+  ///
+  /// This method returns true if it successfully dismisses the tooltips. It
+  /// returns false if there is no tooltip shown on the screen.
+  static bool dismissAllToolTips() {
+    if (_openedTooltips.isNotEmpty) {
```
justinmc: Nit: I would reverse the logic and early return.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+    required this.target,
+    required this.targetSize,
+    required this.tooltipSize,
+    required this.verticalOffset,
+    required this.preferBelow,
+    this.overlaySize = Size.infinite,
+  });
+
+  /// The center point of the target widget in the global coordinate system.
+  final Offset target;
```
justinmc, line 77: I realize this wasn't your decision, but why is this the center instead of the top left? It seems like the center makes it harder to calculate what you want in most practical cases.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  @override
+  int get hashCode =>
+      Object.hash(target, targetSize, tooltipSize, overlaySize, verticalOffset, preferBelow);
+}
+
+/// The method of interaction that will trigger a tooltip.
+/// Used in [RawTooltip.triggerMode].
+///
+/// On desktop, a tooltip will be shown as soon as a pointer hovers over
+/// the widget, regardless of the value of [RawTooltip.triggerMode].
```
justinmc: What about on a mobile/touch device with a mouse connected? I'm assuming it behaves the same as desktop. If so maybe reword this "When using a pointer device like a mouse, a tooltip will be..."

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  ///
+  /// See also:
+  ///
+  ///   * [GestureDetector.onTap], the event that is used for trigger.
+  ///   * [Feedback.forTap], the feedback method called when feedback is enabled.
+  tap,
+}
+
+/// Signature for when a tooltip is triggered.
+typedef TooltipTriggeredCallback = void Function();
```
justinmc: Could this just be a VoidCallback?

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: I agree with @victorsanni after playing around with this locally. I was able to satisfy all my customization needs. We should just be open to possible changes in the future based on user feedback, as Victor says.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+    if (outermost) {
+      // The outermost region resets the global states.
+      isOutermostMouseRegion = true;
+      foundInnermostMouseRegion = false;
+    }
+    return isHit;
+  }
+}
+
+/// A raw tooltip.
```
justinmc: Nit: A design-agnostic tooltip.

Also is it worth explaining what a tooltip is?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+///
+/// It can be triggered in three ways:
+///
+///  * By a long press gesture.
+///  * By a tap gesture.
+///  * By hovering over the widget with a mouse pointer.
+///  * Manually, using the [RawTooltipState.ensureTooltipVisible] function.
+class RawTooltip extends StatefulWidget {
```
justinmc: Add a "See also" add link to Tooltip.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+  /// The returned widget will be wrapped in an [IgnorePointer] widget if
+  /// [ignorePointer] is true.
+  final TooltipComponentBuilder tooltipBuilder;
+
+  /// Whether the tooltip defaults to being displayed below the widget.
+  ///
+  /// If there is insufficient space to display the tooltip in
+  /// the preferred direction, the tooltip will be displayed in the opposite
+  /// direction.
+  final bool preferBelow;
```
justinmc: Defaults to true.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,928 @@
...
+
+  /// Shows the tooltip if it is not already visible.
+  ///
+  /// After made visible by this method, The tooltip does not automatically
+  /// dismiss after `waitDuration`, until the user dismisses/re-triggers it, or
+  /// [RawTooltip.dismissAllToolTips] is called.
+  ///
+  /// Returns `false` when the tooltip shouldn't be shown or when the tooltip
+  /// was already visible.
+  bool ensureTooltipVisible() {
```
justinmc, line 764: I'm always extra skeptical of imperative APIs, but I think this one makes sense. It would be super cumbersome to try to make a declarative API via a "visible" parameter, since internally the widget needs to change visibility.

The only viable alternative to this pattern is a controller, but it seems to be overkill in this case.

packages/flutter_test/test/finders_test.dart
```diff
@@ -369,6 +369,20 @@ void main() {
...
+        _boilerplate(
+          RawTooltip(
+            semanticsTooltip: 'Tooltip Message',
+            tooltipBuilder: (BuildContext context, Animation<double> animation) =>
+                const Text('Tooltip Message'),
+            child: const Text('+'),
+          ),
+        ),
+      );
+      expect(find.byTooltip('Tooltip Message'), findsOneWidget);
```
justinmc, line 383: I think I didn't even realize byTooltip was a thing, good catch.

packages/flutter/test/material/tooltip_test.dart
justinmc, line 1: Now that we've got all of those tests in widgets/tooltip_test.dart, are there any tests in this file that are redundant and no longer needed? IMO if some feature's implementation is in RawTooltip, it can be tested in detail in widgets/tooltip_test.dart, and material/tooltip_test.dart only really needs a test to make sure that it can pass the parameter down or whatever.

packages/flutter/test/widgets/raw_tooltip_test.dart
```diff
@@ -0,0 +1,2747 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
justinmc, line 1: I assume this file was copied from material/tooltip_test.dart. Is there anything new I should specifically review in here?

justinmc (commented):
I think this is a good design and a useful example of a raw widget abstraction, most comments below are minor. There is a lot of nuance inside RawTooltip that would be a pain for a user to implement themselves. Despite that, the API keeps it quite customizable.

I played around with this locally and it worked pretty well. It seemed to be customizable enough to be able to do what I imagine design library authors would want it to do, and in a nice way. It also correctly tracked the target while inside of a scrollable too. I guess Tooltip worked like that before, but I wanted to make sure.

What will we do when multiwindow is available and supports tooltip windows on desktop?

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: I think the default was to have the tooltip overlay child center aligned with the tooltip child? I think I agree left is more conventional, but changing it will be breaking. This `target` is the same as in [positionDependentBox](https://api.flutter.dev/flutter/painting/positionDependentBox.html).

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Agree, but this was introduced before this PR. Maybe we should deprecate this?

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: done.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: done.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: true, I think we can keep it in Material Tooltip for a cleaner API.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: > Is there ever any positioning logic that this can't handle?

I don't think so, the return value is an `Offset` which is used to position the overlaychild in the Overlay, so it allows for completely arbitrary positioning.

>  I wonder if anyone would want to be able to pass the whole overlayChildBuilder

So basically override `_buildTooltipOverlay`? Maybe. One example is if they want an alternative to `_ExclusiveMouseRegion`. But with respect to UI and positioning, it is infinitely customizable.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: There is an open request for a controller: https://github.com/flutter/flutter/issues/126787

packages/flutter/test/widgets/raw_tooltip_test.dart
victorsanni: Not necessarily, but I added some boilerplate to deal with MaterialApp -> WidgetsApp. Most notably WidgetsApp.pageRouteBuilder.

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -583,274 +377,18 @@ class TooltipState extends State<Tooltip> with SingleTickerProviderStateMixin {
...
-  static const Duration _defaultHoverExitDuration = Duration(milliseconds: 100);
+  static const Duration _defaultExitDuration = Duration(milliseconds: 100);
   static const Duration _defaultWaitDuration = Duration.zero;
   static const bool _defaultExcludeFromSemantics = false;
   static const TooltipTriggerMode _defaultTriggerMode = TooltipTriggerMode.longPress;
   static const bool _defaultEnableFeedback = true;
   static const TextAlign _defaultTextAlign = TextAlign.start;
 
-  final OverlayPortalController _overlayController = OverlayPortalController();
+  final GlobalKey<RawTooltipState> _tooltipKey = GlobalKey();
```
chunhtai:
```suggestion
  final GlobalKey<RawTooltipState> _tooltipKey = GlobalKey<RawTooltipState>();
```

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -962,242 +491,87 @@ class TooltipState extends State<Tooltip> with SingleTickerProviderStateMixin {
...
-      size: size,
-      childSize: childSize,
-      target: target,
-      verticalOffset: verticalOffset,
-      preferBelow: preferBelow,
+    return RawTooltip(
+      key: _tooltipKey,
+      semanticsTooltip: widget.message ?? widget.richMessage?.toPlainText() ?? '',
+      tooltipBuilder: (BuildContext context, Animation<double> animation) => IgnorePointer(
+        ignoring: widget.ignorePointer ?? widget.message != null,
```
chunhtai: do you know why we have ignore pointer here? and only when there is no rich message?

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -962,242 +491,87 @@ class TooltipState extends State<Tooltip> with SingleTickerProviderStateMixin {
...
-                heightFactor: 1.0,
-                child: Text.rich(richMessage, style: textStyle, textAlign: textAlign),
-              ),
+    return ConstrainedBox(
+      constraints: constraints,
+      child: DefaultTextStyle(
+        style: textStyle,
+        textAlign: textAlign,
+        child: Semantics(
+          container: true,
```
chunhtai: this and the SemanticsRole.tooltip should be applied in the RawTooltip

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1767,6 +1769,18 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
       return true;
     }());
 
+    result = Focus(
```
chunhtai, line 1775: Should we use tabRegion.onTapOutside to handle this?

packages/flutter/lib/src/widgets/tooltip.dart
chunhtai: You can do something like  RawTooltipPositionContext that without preferbelow and then in the material Tooltip re-wrap the RawTooltipPositionContext with the Tooltip's preferBelow to create TooltipPositionContext

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,969 @@
...
+/// as a pointer hovers over the widget, regardless of the value of
+/// [RawTooltip.triggerMode].
+///
+/// See also:
+///
+///   * [RawTooltip.waitDuration], which defines the length of time that
+///     a pointer must hover over a tooltip's widget before the tooltip
+///     will be shown.
+enum TooltipTriggerMode {
+  /// Tooltip will only be shown by calling `ensureTooltipVisible`.
```
chunhtai: use code link RawTooltipState.ensureTooltipVisible and material TooltipState.ensureTooltipVisible

packages/flutter/lib/src/widgets/tooltip.dart
chunhtai: The material tooltip can just build the child directly instead of using RawTooltip is visibility is off. It still a bit hard for me to imagine the use case though.

packages/flutter/lib/src/material/tooltip.dart
victorsanni: This is preserving existing behavior in material Tooltip

packages/flutter/lib/src/widgets/app.dart
victorsanni: I think this is allowing the tooltip to be dismissed when the `esc` key is pressed? And onTapOutside will dismiss the tooltip when tapped outside which is a different feature?

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Instead of introducing a new class just to resolve this property, I'm thinking we can remove it from `RawTooltip` and do the same thing we did for verticalOffset here. That way `positionDelegate` can handle all position customizations. We can default it to position below, but if they want it above they can write a custom `positionDelegate` as such:

## Before

```dart
RawTooltip(
  ...
  preferBelow: false,  // default is true
)
```

## After
```dart
RawTooltip(
  ...
  positionDelegate: (TooltipPositionContext context) => positionDependentBox(
    size: context.overlaySize,
    childSize: context.tooltipSize,
    target: context.target,
    preferBelow: false,  // default is true
  ),
)
```

packages/flutter/lib/src/widgets/app.dart
chunhtai: oh I misread the code, how about the using WidgetsApp.defaultActions to handle this?

victorsanni: In `_defaultShortcuts` the `esc` key maps to `DismissIntent`:

https://github.com/flutter/flutter/blob/33ed9ba6fc0a666a666fd9bcc2018b8a17855ed7/packages/flutter/lib/src/widgets/app.dart#L1268

I don't know too much about Intents/Actions/Shortcuts but I think when the esc key is pressed, the nearest parent that has an action that consumes DismissIntent (i.e extends Action<DismissIntent>) will be triggered. 

So what happens if the tooltip is in a dialog that has an action to dismiss when the esc key is pressed? That would dismiss the dialog and not the tooltip I think? Which is undesired? 

I'm looking at the [PR](https://github.com/flutter/flutter/pull/83830) that added the code block and I'm thinking that is the same rationale (i.e always dismiss tooltips on esc).

chunhtai: oh wow I don't even remember I added it. I wasn't too familiar with the action shortcut when I first added the code.

> the tooltip is in a dialog that has an action to dismiss when the esc key is pressed? That would dismiss the dialog and not the tooltip I think? Which is undesired?

I don't think the current code handle it either, at least all the current dismissaction will consume the ESC.

victorsanni: I see. I want to add a class DismissTooltipsAction, should it be public or private? And should it live in widgets/app.dart or widgets/tooltip.dart?

victorsanni: I created this action:

```dart
class _DismissTooltipsAction extends DismissAction {
  _DismissTooltipsAction();

  @override
  Object invoke(DismissIntent intent) {
    return RawTooltip.dismissAllToolTips();
  }
}
```

and passed it to `WidgetsApp.defaultActions`:

```dart
DismissIntent: _DismissTooltipsAction()
```

but it doesn't work

chunhtai: nvm this for now, I have spotted several bugs in the action shortcut system that it will just crash due to assertion error in here https://github.com/flutter/flutter/blob/8edbbb739139ddc9a18bbd41b4d027e6d980b2d5/packages/flutter/lib/src/widgets/actions.dart#L866
basically the intent can't be null otherwise this crash. Not entirely sure about the intention for this assert

Even if I remove the assert, it seems like the ModalRoute's _DismissModalAction will steal the shortcut **even though**  it is not enabled, not entirely sure whether this is intentional, this mean dismiss action won't ever propagate up above the modalroute.

I think we may need to take a closer look at the action shortcuts mechanism, I don't think it achieve all the use cases we want. Either we implement a way for _DismissModalAction to give up the intent, or have a way for tooltip to **inject** its action to the nearlest ActionScope.

I think this is out of scope for this pr. Can you add a todo here and file an issue that the tooltip dismissal should rely on action shortcut?

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+///  * [TooltipPositionDelegate], which uses this context to compute tooltip positions.
+@immutable
+class TooltipPositionContext {
+  /// Creates a tooltip position context.
+  const TooltipPositionContext({
+    required this.target,
+    required this.targetSize,
+    required this.tooltipSize,
+    required this.verticalOffset,
+    this.preferBelow = true,
```
chunhtai, line 72: looks like this is never passed in and will always be true? this seems confusing for the consumer of RawTooltip widget. I think this is even worse than piping the RawTooltip.preferBewlow to this class.

I still think we should probably create a new class.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+  @override
+  void initState() {
+    super.initState();
+    // Listen to global pointer events so that we can hide a tooltip immediately
+    // if some other control is clicked on. Pointer events are dispatched to
+    // global routes **after** other routes.
+    GestureBinding.instance.pointerRouter.addGlobalRoute(_handleGlobalPointerEvent);
+  }
+
+  Widget _buildTooltipOverlay(BuildContext context, OverlayChildLayoutInfo layoutInfo) {
```
chunhtai, line 784: Maybe I missed something, but the Semantics(container:true) in the original file material/tooltip.dart line 1170 is missing. We also want to apply SemanticsRole.tooltip here as well

packages/flutter/lib/src/material/tooltip.dart
```diff
@@ -460,52 +291,16 @@ class Tooltip extends StatefulWidget {
...
-      // Avoid concurrent modification.
-      final List<TooltipState> openedTooltips = _openedTooltips.toList();
-      for (final state in openedTooltips) {
-        assert(state.mounted);
-        state._scheduleDismissTooltip(withDelay: Duration.zero);
-      }
-      return true;
-    }
-    return false;
+    return RawTooltip.dismissAllToolTips();
```
dkwingsmt, line 314: Make doc a `@macro`?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+
+const AnimationStyle _kDefaultAnimationStyle = AnimationStyle(
+  curve: Curves.fastOutSlowIn,
+  duration: Duration(milliseconds: 150),
+  reverseDuration: Duration(milliseconds: 75),
+);
+
+/// Signature for building the tooltip overlay child.
+///
+/// The `animation` parameter is an [Animation] that maps to the tooltip's show
```
dkwingsmt: Do you think it'd be better to move the detailed explanation to the class property (`RawTooltip.tooltipBuilder`), i.e. the function signature only has the first paragraph? This is because developers will probably only come it from `RawTooltip.tooltipBuilder` anyway.

Also link the method where this signature is used.

If you agree, apply this to other signature symbols.

packages/flutter/lib/src/widgets/raw_tooltip.dart
dkwingsmt: Wouldn't top left be biased toward a certain case? I think the center position, combined with `targetSize`, is the simplest way to provide just enough information.

packages/flutter/lib/src/widgets/raw_tooltip.dart
dkwingsmt: Why can't we pass `RawTooltip.preferBewlow` to this class?

If we really don't want `preferBewlow`, my suggestion is to rewrite `TooltipPositionContext.preferBewlow` as a getter, and mark it as deprecated.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+  int get hashCode =>
+      Object.hash(target, targetSize, tooltipSize, overlaySize, verticalOffset, preferBelow);
+}
+
+/// The method of interaction that will trigger a tooltip.
+/// Used in [RawTooltip.triggerMode].
+///
+/// When using a pointer device like a mouse, a tooltip will be shown as soon
+/// as a pointer hovers over the widget, regardless of the value of
+/// [RawTooltip.triggerMode].
```
dkwingsmt: I think it's clearer to explain that this enum only affects touch based events, and will not affect mouse or method-calling whatsoever. 
```suggestion
/// How touch events should trigger a tooltip.
///
/// When using a pointer device like a mouse, a tooltip is shown as soon as
/// a pointer hovers over the widget, regardless of the value of
/// [RawTooltip.triggerMode].
///
/// A tooltip might also be triggered through other means regardless of this
/// option, such as by calling `ensureTooltipVisible`.
///
/// See also:
///
///  * [RawTooltip.triggerMode], which uses this enum.
```

Also moved the link to the property as a "See also", since it can also be used in themed widgets.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+///
+/// See also:
+///
+///   * [RawTooltip.waitDuration], which defines the length of time that
+///     a pointer must hover over a tooltip's widget before the tooltip
+///     will be shown.
+enum TooltipTriggerMode {
+  /// Tooltip will only be shown by calling [RawTooltipState.ensureTooltipVisible].
```
dkwingsmt: In accompany with the change above:
```suggestion
  /// Tooltip will not be triggered by touch events.
```

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+      foundInnermostMouseRegion = false;
+    }
+    return isHit;
+  }
+}
+
+/// A design-language-agnostic tooltip.
+///
+/// A tooltip provides a text label which helps explain the function of a button
+/// or some other user interface action.
```
dkwingsmt: Instead of focusing on being "design-language-agnistic", which is well known since we're in the `widgets` library, the doc should focus more on "what a tooltip is". I propose the following doc after consulting AI:
```dart
/// A widget that wraps a child to display an informative overlay in response to
/// user interactions, such as hovering or long-pressing.
/// 
/// Tooltips provide essential context by displaying text labels or brief
/// descriptions that explain the function of a button or other user interface
/// elements.
```

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+///
+/// [RawTooltip] can be triggered in three ways:
+///  * By a long press gesture.
+///  * By a tap gesture.
+///  * By hovering over the widget with a mouse pointer.
+///  * Manually, using the [RawTooltipState.ensureTooltipVisible] function.
```
dkwingsmt: Slight rephrase:
```suggestion
/// The tooltip can be triggered in several ways:
///
///  * By hovering a mouse pointer over the widget.
///  * By touch interactions, such as a long press or a tap, depending on the
///    configuration of [triggerMode].
///  * Manually, by calling [RawTooltipState.ensureTooltipVisible].
```

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+///
+/// See also:
+///
+///   * [Tooltip], a higher-level widget in the Material library that wraps
```
dkwingsmt: Are there a lot of additional functionalities? Isn't it mostly just with a material design layout by default?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+
+  /// Builds the widget that will be displayed in the tooltip's overlay.
+  ///
+  /// The animation property exposes the underlying tooltip overlay child show
+  /// and hide animation. This can be used to drive animations that sync up with
+  /// the tooltip overlay child show/hide animation, for example to fade the
+  /// tooltip in and out.
+  final TooltipComponentBuilder tooltipBuilder;
+
+  /// Whether the tooltip's [semanticsTooltip] should be excluded from
```
dkwingsmt: Does it mean `semanticsTooltip` is useless if this boolean is false?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+    this.enableFeedback = true,
+    this.onTriggered,
+    this.mouseCursor,
+    this.animationStyle = _kDefaultAnimationStyle,
+    this.positionDelegate,
+    required this.child,
+  });
+
+  /// The text to display in the tooltip's semantics announcement.
+  final String semanticsTooltip;
```
dkwingsmt: Out of curiosity, I suppose this is intentional that we don't provide any convenient default setup to derive this property from the tooltip, right?

Can we at least provide some guidance, such as "typically you should put xxx string here and it will be announced like xxx"?

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+  ///
+  /// See also:
+  ///
+  /// * [TooltipTriggerMode], which defines the different ways a tooltip can be
+  ///   triggered.
+  /// {@endtemplate}
+  final TooltipTriggeredCallback? onTriggered;
+
+  /// {@template flutter.widgets.RawTooltip.mouseCursor}
+  /// The cursor for a mouse pointer when it enters or is hovering over the
```
dkwingsmt: Why do we need this instead of letting `child` defining it themselves?

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+  /// override the hide tooltip animation duration. If it is null, defaults to
+  /// 75ms.
+  ///
+  /// If [AnimationStyle.reverseCurve] is provided, it will be used to override
+  /// the hide tooltip animation curve. If it is null, the same curve will be
+  /// used as for the show tooltip animation.
+  ///
+  /// To disable the tooltip show/hide animation, use
+  /// [AnimationStyle.noAnimation].
+  final AnimationStyle animationStyle;
```
dkwingsmt, line 406: Have we considered delegating animation entirely to the themed widget? `AnimationStyle` might not be flexible enough and does not support simulation well (for now). `RawMenuAnchor` uses this design by not be aware of animation at all and letting the themed widget to take over animation.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+        positionDelegate,
+        defaultValue: null,
+      ),
+    );
+  }
+}
+
+/// Contains the state for a [RawTooltip].
+///
+/// This class can be used to programmatically show the [RawTooltip], see the
```
dkwingsmt:
```suggestion
/// This class can be used to programmatically show the [RawTooltip]. See the
```

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,910 @@
...
+///
+/// A tooltip provides a text label which helps explain the function of a button
+/// or some other user interface action.
+///
+/// [RawTooltip] can be triggered in three ways:
+///  * By a long press gesture.
+///  * By a tap gesture.
+///  * By hovering over the widget with a mouse pointer.
+///  * Manually, using the [RawTooltipState.ensureTooltipVisible] function.
+///
```
dkwingsmt, line 225: Add an example of how to call `ensureTooltipVisible` to open the tooltip programatically.

dkwingsmt (commented):
Sorry for the late review.

Generally looking good. Mostly doc suggestions, except that I'm wondering if we can make it unaware of animations and let themed widgets take over animations completely.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: I can readd `Semantics(container:true)` to Material/tooltip.dart, I removed it after a comment, but now I think since it wrapped the Material tooltipBuilder contents it should stay in Material.

Adding `SemanticsRole.tooltip` causes a lot of failures, I'm thinking this can be done in a subsequent PR to reduce the churn here, what do you think?

chunhtai: If we don't add it in this PR I think let's also skip the container:true if it doesn't break the test. I am not too sure why the container:true is needed here. Since it is a overlayportal, It should have created its own node already.

packages/flutter/lib/src/widgets/raw_tooltip.dart
chunhtai: The argument is that it is a redundant API. the consumer passing preferBewlow to RawTooltip just to be given back to them in the TooltipPositionContext. If the consumer knows what the preferBewlow should be, they could have uses it directly in position delegate

chunhtai: > If we really don't want preferBewlow, my suggestion is to rewrite TooltipPositionContext.preferBewlow as a getter, and mark it as deprecated.

This is a valid API for material version of Tooltip, because the consumer of Tooltip widget doesn't know about the preferBelow, it is derive internally in Tooltip. So we should keep it.

The RawTooltip is a different story though

packages/flutter/lib/src/widgets/raw_tooltip.dart
chunhtai: According to the proposal [go/normalize-design](http://go/normalize-design) (if we agree that is), Tooltip has a more strict [APG requirement](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/) that are aware of animation. it is probably best to keep the animation in RawTooltip so it has full control to achieve the APG.

chunhtai: for example,

> If the tooltip is invoked when a pointing cursor moves over the trigger element, then it remains open as long as the cursor is over the trigger or the tooltip.

This will be tricky to implement, if we let design library to drive the animation. (think about if the mouse move in and out multiple time when the dismiss/appearing animation is on going)

dkwingsmt: I don't think it's that hard to implement. Similar to `RawMenuAnchor`, you can let `RawTooltip` to tell the themed tooltip when the message is shown or hidden, and the themed tooltip is basically drawing an animating message. I agree that it might take some design for `RawTooltip` to know the rect of the message to invoke the triggers.

Anyway, I'm not strong on this design, since I think this PR is close to finish and is in good shape, and we can always add a more customizable constructor in the future.

chunhtai: I think it is implementable, but the debouncing of repeatedly enter and exit will have to be handled in the design library, which I think can have some corner cases. We will be kind of relying on the design library to do the right thing in order to fulfill APG.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Even without this PR, I think preferBelow should never have been exposed in TooltipPositionContext because it doesn't tell the user writing their Tooltip any new or difficult-to-compute information. Same as verticalOffset. 

Both are already properties on Tooltip that can be set directly, it doesn't make sense for a user to set them in the Tooltip constructor then receive the same values they set when writing the positionDelegate.

chunhtai: > Both are already properties on Tooltip that can be set directly, it doesn't make sense for a user to set them in the Tooltip constructor then receive the same values they set when writing the positionDelegate.

for material tooltip, they have to do theme resolution to get the effective value. but i am not strongly toward one way or another though. So if you think it is best to deprecate it, feel free to ignore my previous comment

dkwingsmt: > This is a valid API for material version of Tooltip, because the consumer of Tooltip widget doesn't know about the preferBelow, it is derive internally in Tooltip.

Doesn't `Tooltip` have a `preferBelow` property? https://api.flutter.dev/flutter/material/Tooltip/preferBelow.html
(Of course it also defaults to the theme value, but the consumer would know the theme value easily.)

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Another drawback if we remove the animation logic from RawTooltip is (I think) we also lose the timer logic. So all duration APIs (waitDuration, showDuration, exitDuration) which is kinda essential to how tooltips are used because of features like debouncing the tooltip and setting how long the tooltip should be shown.

victorsanni: Side note, I had Gemini create a RawTooltip that grabs the animation from tooltipBuilder and uses it to drive a spring simulation: [https://gist.github.com/victorsanni/a1f2c9b2da38284941962a59dfd807c7](https://www.google.com/url?q=https://gist.github.com/victorsanni/a1f2c9b2da38284941962a59dfd807c7&sa=D&source=docs&ust=1766174971881201&usg=AOvVaw2_JsAQtW8y-J0vYrXRPfGn)

If you set `RawTooltip.animationStyle.duration` and `RawTooltip.animationStyle.reverseDuration` to an infinitesimal value, it's very possible to have a physics-based animation chained to the internal animation so that the effective animation looks almost entirely physics-based. It's hacky though, and might be tricky to test.

packages/flutter/lib/src/widgets/raw_tooltip.dart
dkwingsmt: I propose we keep this value:
* Even though it's some information that is easily accessible outside of the callback, it's ok to provide it again. In fact, it's convenient for the implementer to have it at hand.
* Keeping it simplifies this PR.
* We can still come back and discuss about changing it in the future. Landing the PR in the current way doesn't make it easier or harder.

chunhtai: Thinking a bit more, this kind of render tooltipTheme.preferBelow  useless if we remove it. it will be no longer used by the material library.

chunhtai: actually I was wrong, there is still a default builder that uses preferbelow so nvm

chunhtai: yeah given that we are either removing both or keep both for rawtooltip and tooltip without creating a new RawTooltipPositioncontext. I am ok to make the decision in a later day, and land it as-is then

packages/flutter/lib/src/widgets/raw_tooltip.dart
chunhtai: If we were to change this I think we should probably just give a Rect, but that is probably outside of the scope.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: SGTM. I will add a TODO to consider removing `preferBelow` and `verticalOffset` from `TooltipPositionContext`.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: done, I added a small paragraph and moved the bulk of the docs to the property.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: I'll add a TODO here and do it in a separate PR, just to make this one a little easier to review.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: changed to "a Material-themed RawTooltip"

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Yes, different accessibility tools will announce in different ways. Maybe we should specify here that this is the string that will be used by accessibility tools like Talkback/Voiceover.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Yes, can we remove this flag and do this instead by providing an empty "semanticsTooltip"?

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: True, removed API

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1767,6 +1769,20 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
       return true;
     }());
 
+    // TODO(victorsanni): Use actions and shortcuts to dismiss tooltips when
+    // esc is pressed instead of using a Focus widget.
```
dkwingsmt: Add a TODO issue and link here

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1767,6 +1769,20 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
...
+    result = Focus(
+      canRequestFocus: false,
+      onKeyEvent: (FocusNode node, KeyEvent event) {
+        if ((event is! KeyDownEvent && event is! KeyRepeatEvent) ||
+            event.logicalKey != LogicalKeyboardKey.escape) {
+          return KeyEventResult.ignored;
+        }
+        return RawTooltip.dismissAllToolTips() ? KeyEventResult.handled : KeyEventResult.ignored;
```
dkwingsmt, line 1782: Does it mean that, if I have a dialog, whose body has a underlined text span with tooltip, and I press Esc with the tooltip on, the Esc will not close the dialog but the tooltip (while my mouse is still on the underlined text span)? It doesn't feel very intuitive to me... (Let me know it's intentional or is the current behavior.)

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,920 @@
...
+  Widget build(BuildContext context) {
+    // If message is empty then no need to create a tooltip overlay to show
+    // the empty black container so just return the wrapped child as is or
+    // empty container if child is not specified.
+    if (widget.semanticsTooltip.isEmpty) {
+      return widget.child;
+    }
+    assert(debugCheckHasOverlay(context));
+    final bool excludeFromSemantics = widget.excludeFromSemantics;
+    // TODO(victorsanni): Add SemanticsRole.tooltip.
```
dkwingsmt: Add a TODO issue?

packages/flutter/lib/src/widgets/tooltip.dart
dkwingsmt: I think it's definitely the more intuitive API, although I suggest asking for @chunhtai 's opinion first.

Maybe make it required but nullable, so that excluding from semantics is still an explicit configuration.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,920 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
dkwingsmt, line 1: I suggest renaming all files related to `RawTooltip` as `raw_tooltip`. This convention is used by other raw widgets and going against it makes it counterintuitive when I can't find anything searching for the file based on the widget name.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,920 @@
...
+  /// {@endtemplate}
+  final Duration waitDuration;
+
+  /// {@template flutter.widgets.RawTooltip.showDuration}
+  /// The length of time that the tooltip will be shown after a long press is
+  /// released (if triggerMode is [TooltipTriggerMode.longPress]) or a tap is
+  /// released (if triggerMode is [TooltipTriggerMode.tap]). This property
+  /// does not affect mouse pointer devices.
+  ///
+  /// Defaults to 1.5 seconds for long press and tap released
```
dkwingsmt: The first paragraph should be one-sentence long.

packages/flutter/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,920 @@
...
+  ///
+  /// See also:
+  ///
+  ///  * [exitDuration], which allows configuring the time until a pointer
+  /// disappears when hovering.
+  /// {@endtemplate}
+  final Duration showDuration;
```
dkwingsmt: I think the name `showDuration` is very misleading. It's more like `showDelay` to me, while `showDuration` sounds like the time before it automatically dismisses. Shall we take this opportunity to rename these properties, instead of fixating them with this new widget that goes into the next release?

victorsanni: I agree, `waitDuration` and `showDuration` are both delays, but `waitDuration` is the delay on hover and `showDuration` is the delay on tap/long press. Maybe `hoverDelay` for `waitDuration` and `touchDelay` for `showDuration`?

Even `exitDuration` imo should be changed to `dismissDelay` or `exitDelay`.

packages/flutter/lib/src/widgets/app.dart
victorsanni: Yes I think this is the existing behavior, I think it is intentional.

packages/flutter/lib/src/widgets/tooltip.dart
dkwingsmt: I like all your names! :)

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -253,7 +253,10 @@ class RawTooltip extends StatefulWidget {
   ///
   /// This string is used by assistive technologies, most notably screen readers
   /// like TalkBack and VoiceOver, to describe the tooltip's purpose.
-  final String semanticsTooltip;
+  ///
+  /// If non-null and non-empty, the tooltip will add a [Semantics] label that
+  /// is set to this value.
```
dkwingsmt: The last paragraph feels a bit awkward. Readers might take it as hinting that they should set it as null. I propose the following version based on consult with AI, which uses a more unbiased way to describe both cases.
```dart
  /// The text used by assistive technologies (e.g., TalkBack, VoiceOver) to
  /// describe the tooltip's purpose.
  ///
  /// Providing a non-empty string adds a [Semantics] label for assistive 
  /// technologies. If the tooltip should not have a semantic description, 
  /// this property must be explicitly set to null.
```

dkwingsmt (approved):
LGTM. Thank you!

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: Sounds good. No need to do anything here now until people are asking for it.

packages/flutter/lib/src/widgets/tooltip.dart
justinmc: For another PR probably.

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1767,6 +1769,21 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
       return true;
     }());
 
+    // TODO(victorsanni): https://github.com/flutter/flutter/issues/180319
+    // Use actions and shortcuts to dismiss tooltips when esc is pressed instead
+    // of using a Focus widget.
```
justinmc, line 1774: Good catch, this is something that would be nice to get right in the raw widget. They can override the Shortcut if they really want to.

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: @dkwingsmt's rewrite looks good to me. FYI @chunhtai .

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,916 @@
...
+  /// override the hide tooltip animation duration. If it is null, defaults to
+  /// 75ms.
+  ///
+  /// If [AnimationStyle.reverseCurve] is provided, it will be used to override
+  /// the hide tooltip animation curve. If it is null, the same curve will be
+  /// used as for the show tooltip animation.
+  ///
+  /// To disable the tooltip show/hide animation, use
+  /// [AnimationStyle.noAnimation].
+  final AnimationStyle animationStyle;
```
justinmc, line 406: If I wanted my tooltip to animate with a SpringSimulation, could I do that?

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,916 @@
...
+  ///   // Position tooltip to the right of the target
+  ///   return Offset(
+  ///     context.target.dx + context.targetSize.width / 2,
+  ///     context.target.dy - context.tooltipSize.height / 2,
+  ///   );
+  /// }
+  /// ```
+  ///
+  /// If null, the default positioning behavior is used (above or below the
+  /// target).
```
justinmc: Probably explain what determines whether it defaults to above or below.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,916 @@
...
+  /// other arbitrary positioning logic.
+  ///
+  /// Example:
+  /// ```dart
+  /// positionDelegate: (TooltipPositionContext context) {
+  ///   // Position tooltip to the right of the target
+  ///   return Offset(
+  ///     context.target.dx + context.targetSize.width / 2,
+  ///     context.target.dy - context.tooltipSize.height / 2,
+  ///   );
```
justinmc: Somewhere in these docs you should probably explain that the anchor is at the midpoint of the target, i.e. returning Offset.zero will position the top left corner of the toopltip on the center of the target.

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,916 @@
...
+  /// {@macro flutter.widgets.ProxyWidget.child}
+  final Widget child;
+
+  static final List<RawTooltipState> _openedTooltips = <RawTooltipState>[];
+
+  /// {@template flutter.widgets.RawTooltip.dismissAllToolTips}
+  /// Dismiss all of the tooltips that are currently shown on the screen,
+  /// including those with mouse cursors currently hovering over them.
+  ///
+  /// This method returns true if it successfully dismisses the tooltips. It
```
justinmc: Nit: "the tooltips" => "at least one tooltip"

packages/flutter/lib/src/widgets/raw_tooltip.dart
```diff
@@ -0,0 +1,916 @@
...
+      ),
+    );
+  }
+}
+
+/// Contains the state for a [RawTooltip].
+///
+/// This class can be used to programmatically show the [RawTooltip]. See the
+/// [ensureTooltipVisible] method.
+class RawTooltipState extends State<RawTooltip> with SingleTickerProviderStateMixin {
```
justinmc, line 521: Should we close https://github.com/flutter/flutter/issues/126787 and recommend that people use RawTooltipState instead?

justinmc (approved):
LGTM :+1: . FYI there are analyzer failures.

packages/flutter/lib/src/widgets/raw_tooltip.dart
chunhtai: LGTM for most part except that this sets [Semantics.tooltip], not `label`

chunhtai (approved):
LGTM

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Changed it to Semantics tooltip label, is this ok?

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Yeah, maybe I can add a TODO here to add an example of how to chain a physics based animation to the tooltip animation.

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: changed to 

```
  /// If null, defaults to positioning the tooltip below the target and
  /// center-aligned using [positionDependentBox].
```

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Yes, I can leave a comment that it is a nice to have but will be too much of a breaking change to justify adding without a clear use case.

packages/flutter/lib/src/widgets/raw_tooltip.dart
chunhtai: maybe Semantics tooltip string, the word `label` may confused with SemanticsProperties.label

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: Actually returning Offset.zero from positionDelegate will position the tooltip at the top left corner of the overlay. The offset here is relative to the overlay, that's why `context` is exposed to tell the developer the `RawTooltip.child`'s  position and size.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: So I am running into a problem with `find.byTooltip` if this flag is removed. `find.byTooltip` will attempt to find the widget with a tooltip message (i.e semanticsTooltip). However if an empty/null semanticsTooltip is used to indicate semantics exclusion, that can conflict with a developer who sets a semanticsTooltip to find the widget in tests.

This is indicating to me that semanticsTooltip might not be the right name here since the string serves an additional purpose. Maybe we should keep the `message` paradigm and explain its uses in the documentation. Or alternatively we could deprecate `find.byTooltip`, but this is used a ton in both our source code and by developers.

What do you think @dkwingsmt and @chunhtai ?

chunhtai: > that can conflict with a developer who sets a semanticsTooltip to find the widget in tests.

Can you explain a bit more about this? I am not following


Also agree using nullability in semanticsTooltip to indicate whether they want to exclude semantics

The original flag is also not clear whether they should remove the semantics in the pop up. It feels to me it should remove both the semantics tooltip and also the semantics in pop up. Otherwise we run into a weird situation that the screen reader think there is no tooltip, but the actual tooltip pop can still be read in screen reader

victorsanni: > Can you explain a bit more about this? I am not following

Basically if a developer sets `Tooltip.excludeFromSemantics` to true, they can't use `find.byTooltip` with a message. Because internally Tooltip will send a null message to RawTooltip.semanticsTooltip if the flag is true. And internally `find.byTooltip` compares the message passed in to `RawTooltip.semanticsTooltip` since the tooltip overlay child is a black box.

This [PR](https://github.com/flutter/flutter/pull/71587) turns on the flag for BottomNavigationBar items, which is causing uses of `find.byTooltip` for BottomNavigationBar items to now fail (when they were previously passing).

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: That's probably helpful, thanks.

packages/flutter/lib/src/widgets/raw_tooltip.dart
justinmc: Ah I misread that. More reason to explain it in the docs haha.

Is the comment "Position tooltip to the right of the target" incorrect for this example code then?

victorsanni: > Is the comment "Position tooltip to the right of the target" incorrect for this example code then?

No, the horizontal value of the returned Offset takes the horizontal value of the target (`context.target.dx`) and shifts it to the right by `context.targetSize.width / 2`.

packages/flutter/lib/src/widgets/tooltip.dart
victorsanni: Talked offline with @chunhtai, for now we are adding a separate check in `find.byTooltip` for `Tooltip`s with `excludeFromSemantics` turned on. This will check the `Tooltip.message`/`Tooltip.richMessage` directly instead of using `RawTooltip.semanticsTooltip`.

packages/flutter_test/lib/src/finders.dart
```diff
@@ -361,7 +361,7 @@ class CommonFinders {
     );
   }
 
-  /// Finds [Tooltip] widgets with the given `message`.
+  /// Finds [RawTooltip] widgets with the given `message`.
```
chunhtai: The doc should also mention Tooltip 
```suggestion
  /// Finds [RawTooltip] or [Tooltip] widgets with the given `message`.
```

chunhtai: In the future, if we decide to remove the material dependencies, we should consider either move this method out to the material package somehow, or deprecate this and create a new finder called bySemanticsTooltip

packages/flutter/lib/src/widgets/raw_tooltip.dart
victorsanni: > explain it in the docs haha.

Done with an `Offset.zero` example.

PR #177686, opened 2025-10-28
[web] Refactor LazyPath and separate immutable paths from path builders

mdebbar:
- `CkPath`:
    - An immutable Path backed by the immutable `SkPath`.
    - Not exposed to Flutter.
- `CkPathBuilder`:
    - A path builder backed by `SkPathBuilder`.
    - Not exposed to Flutter.
- `SkwasmPath`:
    - Implements both the immutable path interface and the path builder interface.
    - Not exposed to Flutter.
- `LazyPath`:
    - The only implementation of the mutable `ui.Path` interface.
    - Exposed to Flutter.
    - Manages the lifecycles of renderer-specific paths and path builders.


engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
```diff
@@ -567,11 +647,15 @@ class LazyPath implements LayerPath, Collectable {
...
   @override
   void addPath(ui.Path path, ui.Offset offset, {Float64List? matrix4}) {
+    // TODO(jacksongardner): Shouldn't we store a copy of `path` here? The `path` may change between
+    //                       now and when this command is executed later.
     _addCommand(AddPathCommand(path as LazyPath, offset, matrix4: matrix4));
   }
 
   @override
   void extendWithPath(ui.Path path, ui.Offset offset, {Float64List? matrix4}) {
+    // TODO(jacksongardner): Shouldn't we store a copy of `path` here? The `path` may change between
```
eyebrowsoffire: Yes, I think you're right in this comment. We should copy it.

engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
```diff
@@ -567,11 +647,15 @@ class LazyPath implements LayerPath, Collectable {
 
   @override
   void addPath(ui.Path path, ui.Offset offset, {Float64List? matrix4}) {
+    // TODO(jacksongardner): Shouldn't we store a copy of `path` here? The `path` may change between
```
eyebrowsoffire: Here too, yes copy it.

engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
```diff
@@ -396,15 +455,22 @@ class LazyPath implements LayerPath, Collectable {
...
+      );
+      final DisposablePathBuilder pathBuilder = constructors.fromPath(path);
+      path.dispose();
+      return pathBuilder;
+    });
   }
 
   DisposablePathConstructors constructors;
-  DisposablePath Function() initializer;
+  DisposablePathBuilder Function() initializer;
```
eyebrowsoffire: I think you might be able to just remove this completely, since you've simplified the factory constructors.

eyebrowsoffire (commented):
This is looking great so far!

engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
mdebbar: It's still used by `LazyPath.combined()`.

We may be able to remove it if we introduce a `CombinePathsCommand`? I'll see if that's doable.

mdebbar: Please look at my attempt to remove `initializer` here: https://github.com/flutter/flutter/pull/177686/commits/9c095eb02524ff5c1a526eeaf1966d8148b31bd4

I don't think it's an improvement, but let me know what you think.

engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
```diff
@@ -584,45 +686,53 @@ class LazyPath implements LayerPath, Collectable {
...
   LazyPathMetrics computeMetrics({bool forceClosed = false}) {
-    return LazyPathMetrics(path: this, forceClosed: forceClosed);
+    return LazyPathMetrics(path: LazyPath.fromLazyPath(this), forceClosed: forceClosed);
   }
 
   @override
   void collect() {
     _cachedPath?.dispose();
+    if (!identical(_cachedBuilder, _cachedPath)) {
+      // In some implementations (*cough* skwasm *cough*), the built path and the builder might be the same
```
eyebrowsoffire, line 730: Maybe we can file an issue that skwasm itself should use the pathbuilder pattern. Then you can add it as a todo in this comment.

eyebrowsoffire (dismissed):
Looks good to me! One minor suggestion.

engine/src/flutter/lib/web_ui/lib/src/engine/lazy_path.dart
mdebbar: @gemini-code-assist could you write a Github issue for changing Skwasm to use the PathBuilder pattern more effectively?

mdebbar: @gemini-code-assist you linked to an old issue. I'm not sure that you have access to create issues on Github. Just give me the markdown content and I'll do the creation.

eyebrowsoffire (approved):

AbdeMohlbi:
@mdebbar is there a reason why this did not make it to flutter 3.44.0 or was it because the stable was cut before this commit landed ?

PR #177687, opened 2025-10-28
Fixes RuntimeEffect when used with ImageFilter.compose and gaussian blur

### LINKED ISSUE #170820: `ImageFilter.blur` breaks `ImageFilter.shader` in a `BackdropFilter` together
### Steps to reproduce

When trying to find a workaround for #170792 I stumbled on another roadblock.

In an Impeller build, do the following:

1. create a `BackdropFilterLayer` with its filter set to `ImageFilter.combined`, with inner being a blur and outer being a shader
2. When you change `sigmaX` and `sigmaY`, the size of the texture that's passed to the shader seems to change, but not in a uniform way

Check the code sample for a reliable repro.

### Expected results

The red circle stays in the same position with the same size the entire time.

### Actual results

See video

### Code sample

```dart
// main.dart
import 'dart:ui';

void main() async {
  final shader = await FragmentProgram.fromAsset('shaders/circle.frag');

  runApp(MainApp(shader: shader));
}

  const MainApp({super.key, required this.shader});

  final FragmentProgram shader;

  @override
}

class _MainAppState extends State<MainApp> with SingleTickerProviderStateMixin {
  late final animationController = AnimationController(
    vsync: this,
    lowerBound: 0,
    upperBound: 10,
    duration: Duration(seconds: 2),
  )..repeat(reverse: true);

  @override
    return MaterialApp(
      home: ListenableBuilder(
        listenable: animationController,
        builder: (context, child) {
          return Scaffold(
            body: Stack(
              children: [
                const Center(child: FlutterLogo(size: 100)),
                Positioned.fill(
                  child: LayoutBuilder(
                    builder: (context, constraints) {
                      return BackdropFilter(
                        filter: ImageFilter.compose(
                          inner: ImageFilter.blur(
                            sigmaX: animationController.value,
                            sigmaY: animationController.value,
                            tileMode: TileMode.clamp,
                          ),
                          outer: ImageFilter.shader(
                            widget.shader.fragmentShader()
                              ..setFloat(2, 100) // uCirclePos.x (center)
                              ..setFloat(3, 100) // uCirclePos.y (center)
                              ..setFloat(4, 100.0), // uCircleRadius
                          ),
                        ),
                        child: Center(
                          child: Text(
                            animationController.value.toStringAsFixed(2),
                          ),
                        ),
                      );
                    },
                  ),
                ),
              ],
            ),
          );
        },
      ),
    );
  }
}

```

With this fragment shader:

```glsl
#version 320 es
precision mediump float;

uniform vec2 uResolution;
uniform vec2 uCirclePos;
uniform float uCircleRadius;
uniform sampler2D uTexture;

out vec4 fragColor;

#include <flutter/runtime_effect.glsl>

float sdfCircle(vec2 p, float radius) {
    return length(p) - radius;
}

void main() {
    vec2 fragCoord = FlutterFragCoord().xy;
    vec2 screenUV = fragCoord / uResolution;
    
    vec2 normalizedCoord = (fragCoord - uResolution * 0.5) / min(uResolution.x, uResolution.y);
    vec2 normalizedCirclePos = (uCirclePos - uResolution * 0.5) / min(uResolution.x, uResolution.y);
    
    float normalizedRadius = uCircleRadius / min(uResolution.x, uResolution.y);
    float distance = sdfCircle(normalizedCoord - normalizedCirclePos, normalizedRadius);
    float circleAlpha = 1.0 - smoothstep(-0.01, 0.01, distance);
    vec4 backgroundColor = texture(uTexture, screenUV);
    vec4 circleColor = vec4(1.0, 0.0, 0.0, circleAlpha);
    fragColor = mix(backgroundColor, circleColor, circleAlpha * 0.8);
}
```

</details>

### Screenshots or Video

<details open>

https://github.com/user-attachments/assets/8c2179f6-93ea-49be-9c52-aa8c78a466c7

</details>

### Logs

_No response_

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.32.4, on macOS 15.5 24F74 darwin-arm64, locale en-US)
[✓] Xcode - develop for iOS and macOS (Xcode 16.4)
[✓] Android Studio (version 2024.2)
[✓] VS Code (version 1.101.0)
    ! Error: Browsing on the local area network for iPhone von mir. Ensure the device is
      unlocked and attached with a cable or associated with the same local area network
      as this Mac.
    ! Error: Browsing on the local area network for Tim’s iPad. Ensure the device is
      unlocked and attached with a cable or associated with the same local area network
      as this Mac.
[✓] Network resources
```

</details>

### LINKED ISSUE #177611: Opengles texture upside-down in ImageFilter.compose golden tests
https://github.com/flutter/flutter/pull/177510 introduced a new golden test that draws textures but they are upside down on opengles

affected tests:
- `ComposePaintRuntimeOuter `

This may be the same issue as https://github.com/flutter/flutter/issues/177317

Adding in the suggest y coordinate swap for opengles fixes `ComposePaintRuntimeOuter` but breaks `ComposePaintRuntimeInner`

## observed
the textures are upside down on opengles, but right-side-up on vulkan/metal

## expected
all backends to render the same thing

gaaclarke:
fixes https://github.com/flutter/flutter/issues/170820
fixes https://github.com/flutter/flutter/issues/177611

This PR snapshots the input of a runtimeeffect if its transform is not identity.  This fixes the cases where we use compose with the runtime effect and the inner effect evaluates to a snapshot that doesn't match the size of x, such as`runtime_effect(blur(x))`.  Blur will perform at lower resolutions when the sigma is high as an optimization.  Blur also adds a halo around what it is blurring.  Previously we could optimize away this extra render pass when we assumed that the coverage of inner(x) was the same as x (but potentially just translated), but we can't do that in practice because of things like gaussian blur.

In order to make this PR work we had to come up with a way to disable the extra padding that is added to every `RenderToSnapshot`.  This is because users want to to reference specific fragments.  The extra border of padding would increase the texture size of inner(x) and x which created a jump in the locations calcuated in the fragment shaders when the inner snapshot's transform would jump from identity to non-identity.  The only other alternative would be to pass in the offset to the fragment shader but that would be a breaking change.

In order to get the shaders rendering in the right orientation on opengles I had to switch the shaders to do a `uv.y = 1.0 - uv.y`.  This is in the documentation for ImageFilter.shader but it is an annoying gotcha.  Hopefully this doesn't break someone if the logic was different previously for ImageFilter.compose but this is the correct thing to do and since it wasn't tested, there is no breaking change.

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke:
Goldens that look wrong when removing the offset by 1:
- ColorFilterAdvancedBlendNoFbFetch_Metal
- MaskBlurVariantTestSolidTranslucent_OpenGLES

All of the problems that look like the original reported issue the offset by 1 fixed were on opengles.

gaaclarke:
In the future we can potentially send the transform into the fragment shader, then we could avoid removing the one extra pixel of padding.  It looks like removing that pixel of padding is making us get slightly different results on different platforms when sampling outside of the texture.  This behavior is much better than what we had previously though.

There is a previous comment where jonah said we could pass in this transform as part of an `ImageFilter.shader2` api.  But actually now that we have uniform-by-name we don't need a new API at the dart level, we could just look for a magic uniform name and set it.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
```diff
@@ -46,11 +47,54 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
...
     return std::nullopt;
   }
+
   std::optional<Rect> maybe_input_coverage = input_snapshot->GetCoverage();
   if (!maybe_input_coverage.has_value()) {
     return std::nullopt;
   }
-  Rect input_coverage = maybe_input_coverage.value();
+
+  // If the input snapshot does not have an identity transform the
```
chinmaygarde, line 56: I don't think we document this performance consideration anywhere. Perhaps a section in `ImageFilter.shader`?

chinmaygarde: Another alternative that I am sure was suggested elsewhere could be to pass the transform into the shader as a uniform and push the responsibility of taking it into account during sampling. But I think this is a good compromise when it comes to ease-of-use and correctness.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
```diff
@@ -46,11 +47,54 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
...
+    });
+    TextureContents texture_contents;
+    texture_contents.SetTexture(input_snapshot->texture);
+    std::optional<Rect> bounds =
+        Rect::MakePointBounds(quad.begin(), quad.end());
+    if (bounds.has_value()) {
+      texture_contents.SetSourceRect(bounds.value());
+      texture_contents.SetDestinationRect(coverage);
+      texture_contents.SetStencilEnabled(false);
+      texture_contents.SetSamplerDescriptor(input_snapshot->sampler_descriptor);
```
chinmaygarde, line 85: Shouldn't the sampler be nearest-neighbor? The input snapshots sampler will then be applied when the texture is composited up?

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
```diff
@@ -102,7 +139,7 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
   sub_entity.SetContents(std::move(contents));
   sub_entity.SetBlendMode(entity.GetBlendMode());
   sub_entity.SetTransform(input_snapshot->transform *
-                          Matrix::MakeTranslation(-1.0f * snapshot_origin));
+                          snapshot_transform.Invert());
```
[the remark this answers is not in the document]
chinmaygarde: I had a hard time parsing this gem gem. If don't think this is an issue because the snapshot origin is captured outside the render proc. But I could be misunderstanding what you are trying to say.

chinmaygarde (approved):
A couple of nits. The thing about the sampling mode being the one that may affect correctness.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
chinmaygarde: > There is a previous comment where jonah said we could pass in this transform as part of an ImageFilter.shader2 api. But actually now that we have uniform-by-name we don't need a new API at the dart level, we could just look for a magic uniform name and set it.

Ah just read your followup. Yeah, that makes sense and should be easier to implement now with the uniform lookup by name stuff. But this patch is good from a correctness standpoint.

gaaclarke: Yea, we're definitely in the "make it work" phase, then we can "make it fast" later.  I don't think we can document it as being slow to the user because there isn't actionable information.  There is no good alternative.  If we do something like passing in the transform to avoid the RenderToSnapshot conditionally, that would be the time to make sure to document it for users.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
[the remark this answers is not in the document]
gaaclarke: I wouldn't think too hard about it.  This is likely AI slop... bad bot.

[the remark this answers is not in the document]
gaaclarke: Over those course of fixing these 2 bugs I kept trying to hail mary pass to gemini to see if it could fix it for me.  It couldn't.  I think that means 1) this stuff is complicated 2) the code could be easier to follow, even I have trouble keeping track of the coordinate space transitions.

In order for it to be really helpful we'd have to get something like a loop going where it generates the golden images and "looks" at the result and reiterates.  I don't think that's possible to get setup today.  We have the MCP server for the engine, in the future we could potentially get it serving up images when it can be parsed.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
gaaclarke: I don't think so.  For the blur case this texture we are rendering to a snapshot is going to potentially be 1/8th the resolution of what we are drawing it as.  So I think it's important we have linear sampling to make the blur look smooth.

gaaclarke: yea, looks much better with linear

## nearest

impeller_Play_AiksTest_ComposeBackdropRuntimeOuterBlurInner_Vulkan


## linear

impeller_Play_AiksTest_ComposeBackdropRuntimeOuterBlurInner_Vulkan

PR #177730, opened 2025-10-29
Document that `error` parameter must be initialized

logiclrd:
Flutter is built on top of GLib, and GLib's standard error mechanism uses a structure `GError` and a function `g_set_error`. This function takes a pointer to a pointer, setting it to a new `GError` for the outer caller to receive. The model is designed to capture the first error that occurs; as such, `g_set_error` checks if the supplied (by reference) pointer is _already_ pointing at something. If it is, then it simply emits a warning about this new error and discards it, assuming that `error` already points at a presumably more-pertinent `GError`.

This semantic is sensible, but it isn't automatically obvious. It makes sense when a function is calling multiple other functions, each of which could return an error, ensuring that the error that gets returned is the first one that happened. But, if you're only calling one function and you don't have prior exposure to this design, it might not be intuitively obvious that it has this intentional behaviour. Instead, the `GError** error` parameter just looks like any old `out` parameter. That can lead to code being written that doesn't bother to initialize the error variable; after all, if no error occurs, it's irrelevant, and if an error _does_ occur, then the function will overwrite it anyway, right? (This latter being the incorrect assumption.)

```
    // INCORRECT CODE
    GError *error;

    if (!fl_method_call_respond_success(method_call, result_value, &error)) {
      g_warning("Failed to send Flutter Platform Channel response: %s", error->message);
      g_error_free(error);
    }
```

GLib handles the fact that this isn't automatically obvious to someone with no prior experience with straightforward documentation:

> GLib ▶ set_error
>
> Does nothing if `err` is `NULL`; **if `err` is non-`NULL`, then `*err` must be `NULL`**. A new `GError` is created and assigned to `*err`.

Since this is, ultimately, the function used internally by Flutter API functions to return errors through this channel, Flutter effectively has this same semantic. As such:

```
    // CORRECT CODE
    GError *error = nullptr;

    if (!fl_method_call_respond_success(method_call, result_value, &error)) {
      g_warning("Failed to send Flutter Platform Channel response: %s", error->message);
      g_error_free(error);
    }
```

(edited per @robert-ancell) In addition, `g_autoptr` can be used to make a smart pointer that automatically ensures that the error, if any, is not leaked.

```
    // CORRECTER CODE
    g_autoptr(GError) error = nullptr;

    if (!fl_method_call_respond_success(method_call, result_value, &error)) {
      g_warning("Failed to send Flutter Platform Channel response: %s", error->message);
    }
```

Flutter should make a similar mention in its documentation so that people who are just discovering the system don't make the same mistake I did. 🙂

This PR updates the inline documentation comments for public functions that take `GError** error` parameters to warn concisely about this situation. Note that it is not required (at least by `g_set_error`) that they poin at `NULL`, specifically, just that they not be _uninitialized_.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/linux/public/flutter_linux/fl_texture_gl.h
```diff
@@ -92,7 +92,7 @@ struct _FlTextureGLClass {
    * @width: (inout): width of the texture in pixels.
    * @height: (inout): height of the texture in pixels.
    * @error: (allow-none): #GError location to store the error occurring, or
-   * %NULL to ignore.
+   * %NULL to ignore. If `error` is not NULL, `*error` must not be uninitialized.
```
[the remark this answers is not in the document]
logiclrd: The wording is not ambiguous at all. I meant exactly what I wrote: It must not be an uninitialized value. It could be initialized to `NULL`, or it could contain some other valid `GError *` pointer. This is the pattern used elsewhere; if a call fails but that particular failure isn't checked-for and subsequent calls get the same `GError **`, they see it with the pointer put in it by the first call. It would be inaccurate to say that `*error` must be `NULL`.

engine/src/flutter/shell/platform/linux/fl_engine_private.h
```diff
@@ -96,7 +96,7 @@ FlDisplayMonitor* fl_engine_get_display_monitor(FlEngine* engine);
  * fl_engine_start:
  * @engine: an #FlEngine.
  * @error: (allow-none): #GError location to store the error occurring, or %NULL
- * to ignore.
+ * to ignore. If `error` is not %NULL, `*error` must not be uninitialized.
```
robert-ancell: This is a double negative and thus hard to understand - suggest instead "If provided, `*error` must be initialized.".

robert-ancell (changes requested):
Thanks for this! I suggested a wording change but otherwise I think this change is good.

I also considered linking to the [upstream documentation](https://docs.gtk.org/glib/error-reporting.html) but I didn't find that any clearer describing why it should be initialized.

robert-ancell:
Mostly for the purposes of anyone who might stumble across this bug in the future - the recommended use of GError is to use `g_autoptr` to avoid accidentally leaking the error, i.e. the above case becomes:
```c
    // CORRECT CODE
    g_autoptr(GError) error = nullptr;

    if (!fl_method_call_respond_success(method_call, result_value, &error)) {
      g_warning("Failed to send Flutter Platform Channel response: %s", error->message);
    }
```

engine/src/flutter/shell/platform/linux/fl_engine_private.h
logiclrd: That's fair, though I feel that if it says _just_ "must be initialized", it may not be obvious to readers that both `NULL` and a pointer to an object are valid initialized values. Perhaps a parenthetical on this point...

logiclrd: Okay, this should be much easier to understand:

```
/**
 * fl_engine_start:
 * @engine: an #FlEngine.
 * @error: (allow-none): #GError location to store the error occurring, or %NULL
 * to ignore. If `error` is not %NULL, `*error` must be initialized (typically
 * %NULL, but an error from a previous call using GLib error handling is explicitly
 * valid).
 *
```

Is it good? :slightly_smiling_face:

Silly me, I jumped the gun and committed this change to one file. :facepalm:

logiclrd: Maybe "using GLib error handling" is unnecessary...

robert-ancell: Looks good. I think it's better to err on the side of being more verbose since consumers are less likely to understand the GLib conventions.

logiclrd: I'll propagate that wording out, then.

robert-ancell (approved):
Thanks for working on this!

mattkae (approved):
Looks good to me too!

PR #177743, opened 2025-10-29
[web] Upgrade Chrome to 141 (for engine tests)

### LINKED ISSUE #168686: [web] Engine test failures in Chrome v128.0.6543.0
Starting in Chrome version 128.0.6543.0 several web engine tests are failing.

Investigation revealed that focus/blur events aren't being dispatched following a call to `element.focus()` or `element.blur()` are called. This started with version `128.0.6543.0`. Earlier versions of Chrome dispatch focus/blur events as expected.

The issue could not be reproduced in a standalone jsfiddle.

mdebbar:
- Update Chrome to 141 for web engine tests.
- Improve image codec tests so they exercise all frames.
- Skip the frames of certain images that are known to cause problems in Chrome.

Chrome bug for the problematic images: https://issues.chromium.org/456445108

Fixes https://github.com/flutter/flutter/issues/168686


harryterkelsen (approved):
LGTM

PR #143341, opened 2024-02-12
[Re-re-land] Enforce a policy on supported Gradle, Java, AGP, and KGP versions

gmackall:
This is a direct revert of (the revert of (the reland of (the policy pr))): https://github.com/flutter/flutter/pull/143132. 

The only change is:
1. to put a conditional all on one line, because the packages repository has a test that uses an old flutter project to make sure nothing regresses. The old project uses an old gradle version, and the old gradle version bundles an old groovy version, and the old groovy version has a bug where lines that start with `&&` don't always work: https://issues.apache.org/jira/browse/GROOVY-7218 (I enjoy that the revert reason ends up providing another strong justification to go forward with the policy). Also thanks to @reidbaker for pointing out this bug.
2. I also made a slight formatting change to the messages that print when out of the support bounds, which I think looks slightly better.

I tested this with on a branch that included a revert of https://github.com/flutter/flutter/pull/142008, and was able to recreate the failure and verify that it was resolved by 1).

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #177753, opened 2025-10-29
Respect product flavor abiFilters by adding a `disable-abi-filtering` Android project flag.

### LINKED ISSUE #175845: [Android] reconfigure abiFilters not effect in productFlavors abiFilters
### Steps to reproduce

/android/app/build.gradle.kts

```
flavorDimensions += listOf("device", "type")
productFlavors {
    create("te288") {
        dimension = "device"
        ndk {
            abiFilters.clear()
            abiFilters.addAll(listOf("arm64-v8a"))
        }
    }
    create("rk399") {
        dimension = "device"
        ndk {
            abiFilters.clear()
            abiFilters.addAll(listOf("armeabi-v7a"))
        }
    }
    create("uk392") {
        dimension = "device"
        ndk {
            abiFilters.clear()
            abiFilters.addAll(listOf("armeabi-v7a"))
        }
    }
    create("common") {
        dimension = "type"
    }
    create("fingerprint") {
        dimension = "type"
    }
}
```

Then run build:
```
flutter build apk --release --flavor te288Common
```

### Expected results

The abi should only include arm64_v8a or armeabi-v7a

### Actual results

Apk is still including "x86_64", "arm64_v8a" and "armeabi-v7a".

Image

### Code sample

https://github.com/yiky84119/test1.git

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.4, on Microsoft Windows [版本 10.0.26100.6584], locale zh-CN)
[✓] Windows Version (Windows 11 or higher, 24H2, 2009)
[!] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
    ✗ cmdline-tools component is missing.
      Try installing or updating Android Studio.
      Alternatively, download the tools from https://developer.android.com/studio#command-line-tools-only and make sure
      to set the ANDROID_HOME environment variable.
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/to/windows-android-setup for more details.
[✓] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.14.14)
[✓] Android Studio (version 2025.1.3)
[✓] IntelliJ IDEA Ultimate Edition (version 2025.2)
[✓] Proxy Configuration
[✓] Network resources
```

</details>

mboetger:
This PR introduces a flag (`disable-abi-filtering`) that disables the FlutterPlugin's abiFiltering on buildTypes.  This should only be used when the developers know what they are doing.  Specifically this allows developers to set their own abiFilters in ProductFlavors or defaultConfig.

Because Gradle has complex merging logic, there is a hierarchy of priority when calculating settings (like abiFilters).  For example: defaultConfig < productFlavors < buildTypes when combining the three into the final build variant.  If abiFilters are set in buildType, it will override abiFilters in productFlavors.  

When the FlutterPlugin executes, it cannot know about the developers build.gradle settings and therefore we cannot add logic that checks to see if abiFilters are set in productFlavors.  Therefore, this flag gives developers an escape hatch so they can implement what they want.

Fixes: #175845 

## Pre-launch Checklist

- [X] I signed the [CLA].


gmackall (commented):
I think to truly mark the linked issue as fixed we would also need to expose this option through the flutter tool, something like the path that `android-skip-build-dependency-validation` takes in https://github.com/flutter/flutter/pull/143341

gmackall:
> > I think to truly mark the linked issue as fixed we would also need to expose this option through the flutter tool, something like the path that `android-skip-build-dependency-validation` takes in #143341
> 
> Technically it can be invoked by the flutter tool though, no? This is the test invocation I used:
> 
> ```
> flutter build apk --release --flavor arm64 -Pdisable-abi-filtering=true
> ```
> 
> I thought about adding at `flutter` level command line, but since this is such a niche use case and should only be used at the developers peril I didn't want to make it too easy for developers to find the escape hatch. I was going to add some documentation on https://docs.flutter.dev/release/breaking-changes/default-abi-filters-android but I was concerned adding a `flutter` level flag would invite users to try it out.
> 
> Anyway, that's my 2 cents. If it's a blocker to LGTM I'll add it.

Ah I had forgotten about the `-android-project-arg/-P` command. That's fair, I actually prefer passing it that way as well. 

Maybe we should add something to our [flavors docs](https://docs.flutter.dev/deployment/flavors) which makes brief mention of this requirement when using abiFilters on the product flavors though? Otherwise I think there is essentially no way for a developer to do this outside of either us telling them directly on an issue, or them knowing the internals of the FGP (which I think we should never assume they do).

gmackall (approved):
That additional documentation can be figured out independently though, this change lgtm

jtmcdole:
This was stuck in the merge queue and not progressing; removing and re-inserting.

PR #177816, opened 2025-10-30
Remove minimize API for dialogs

robert-ancell:
According to the [spec](https://docs.google.com/document/d/1eQG-IS7r4_S9_h50MY_hSGwUVtgSTRSLzu7MLzPNd2Y/edit?tab=t.0) dialogs should not be minimizable.

The table in the spec:
  | Regular | Dialog | Popup | Tip | Satellite
-- | -- | -- | -- | -- | --
Has parent | No | Yes | Yes | Yes | Yes
Can have input focus | Yes | Yes | Yes | No | Yes
Can be the active window 1 | Yes | Yes | No | No | Yes
When window or ancestor is not active window | Stays open | Stays open | Closes | Stays open | Hides
User-moveable | Yes | Yes | No | No | Yes
User-resizeable | Yes | Yes | No | No | Yes
Maximizable | Yes | No | No | No | No
Minimizeable | Yes | **No** | No | No | No
Fullscreen-able | Yes | No | No | No | No
Title Bar | Yes | Same as Parent | No | No | Yes
Close button | Yes | No | No | No | Yes

PR #177817, opened 2025-10-30
Implement dialog windows for the Linux platform

robert-ancell:


packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -724,3 +770,163 @@ class RegularWindowControllerLinux extends RegularWindowController {
...
+  String get title => _window.getTitle();
+
+  @override
+  @internal
+  bool get isActivated => _window.isActive();
+
+  @override
+  @internal
+  // NOTE: On Wayland this is never set, see https://gitlab.gnome.org/GNOME/gtk/-/issues/67
+  bool get isMinimized => (_window.getWindow().getState() & _GDK_WINDOW_STATE_ICONIFIED) != 0;
```
robert-ancell, line 885: I think the minimize API is wrong, see https://github.com/flutter/flutter/pull/177816

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -724,3 +770,163 @@ class RegularWindowControllerLinux extends RegularWindowController {
...
+       _parent = parent,
+       _window = _GtkWindow(),
+       super.empty() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    _window.setTypeHint(_GDK_WINDOW_TYPE_HINT_DIALOG);
+    if (parent != null) {
+      final controller = owner._regularWindowControllers[parent.rootView.viewId];
```
robert-ancell: Can dialogs only be children of regular windows? We silently ignore the parent if not, should this throw an exception?

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Spec was incorrect, we should support minimize (for modeless dialogs)

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Discussed in person, yes dialogs can be children of any window type. :shrug:

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -724,3 +779,167 @@ class RegularWindowControllerLinux extends RegularWindowController {
...
 }
+
+/// Implementation of [DialogWindowController] for the Linux platform.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [DialogWindowController], the base class for dialog windows.
+class DialogWindowControllerLinux extends DialogWindowController implements _LinuxWindowController {
```
mattkae: We should just not implement the hidden interface here

mattkae (approved):
LGTM aside from the comment

PR #177826, opened 2025-10-31
feat: added errorPadding property to `TextField`

### LINKED ISSUE #175834: Add Theme-level option to align error text with the text field border (remove default indent/padding)
### Use case

Currently, when using `TextField / TextFormField`, the error text (errorText) is rendered with an indent that matches the input/hint text padding. This behavior makes the error text look visually offset from the left border of the field.

At the moment, the only way to fix this is by overriding decoration properties (error widget, custom padding, etc.) on each individual field. This is tedious and redundant in larger apps. If the padding needs to be adjusted later, developers must revisit every field manually.

### Proposal

Introduce a theme-level property (e.g., in InputDecorationTheme) that controls the horizontal alignment/padding of error/helper text independently from the input/hint text.

Example API idea:
```
InputDecorationTheme(
  errorPadding: EdgeInsets.zero, // or Alignment options
  errorAlignment: TextAlign.start,
)
```

Why it matters:

Developers often want the input/hint text to be padded, but the error text to be flush with the text field border.

This improves visual consistency in UIs.

Reduces redundancy by eliminating per-field overrides.

abikko:
This PR closes: https://github.com/flutter/flutter/issues/175834

Added errorPadding property, 
Currently it's not ready yet.

## Pre-launch Checklist

- [x] I signed the [CLA].


examples/api/lib/material/text_field/text_field.4.dart
justinmc, line 1: It's probably not strictly necessary to include an example for a straightforward change like this. However, if you want to include it as a dartpad example in the errorPadding docs to show people how to solve the problem from the issue, that's probably worth it.

If you do keep this example, you'll have to add a test for it too.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -193,6 +194,7 @@ class _BorderContainer extends StatefulWidget {
   final Color fillColor;
   final Color hoverColor;
   final bool isHovering;
+  final EdgeInsetsGeometry? contentPadding;
```
justinmc, line 197: Is this used anywhere?

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -914,7 +919,7 @@ class _RenderDecoration extends RenderBox
 
   static BoxParentData _boxParentData(RenderBox box) => box.parentData! as BoxParentData;
 
-  EdgeInsetsDirectional get contentPadding => decoration.contentPadding;
+  EdgeInsetsDirectional get contentPadding => decoration.errorPadding ?? decoration.contentPadding;
```
justinmc, line 922: Won't this also affect the position of the content?

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -3322,6 +3340,9 @@ class InputDecoration {
   /// and `EdgeInsets.fromLTRB(12, 24, 12, 16)` when [isDense] is false.
   final EdgeInsetsGeometry? contentPadding;
 
+  ///TODO(abikko): add documentation after implementing error padding in text field
```
justinmc, line 3343: Don't forget this. This is where I'd add your example.

packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1670,81 +1670,90 @@ class _TextFieldState extends State<TextField>
...
-          stylusHandwritingEnabled: widget.stylusHandwritingEnabled,
-          enableIMEPersonalizedLearning: widget.enableIMEPersonalizedLearning,
-          contentInsertionConfiguration: widget.contentInsertionConfiguration,
-          contextMenuBuilder: widget.contextMenuBuilder,
-          spellCheckConfiguration: spellCheckConfiguration,
-          magnifierConfiguration:
-              widget.magnifierConfiguration ?? TextMagnifier.adaptiveMagnifierConfiguration,
-          hintLocales: widget.hintLocales,
+        child: Padding(
+          padding: contentPadding,
```
justinmc, line 1683: I think you probably want to implement the padding in the [layout](https://github.com/flutter/flutter/blob/ac3b91db0dcc463d8816d25417e80d08b79c7030/packages/flutter/lib/src/material/input_decorator.dart#L953) method of InputDecorator, not here.

justinmc (commented):
Thanks for jumping on fixing this. I agree with the need for an errorPadding property like this, but I think the implementation should be in _layout (see my comment below).

packages/flutter/lib/src/material/text_field.dart
abikko: Yes, today I was thinking about moving from here to implementing it to layout for RenderBox InputDecorator

thanks!

packages/flutter/lib/src/material/input_decorator.dart
abikko: There is a problem that _EditableText is "surrounded" by size of _Decorator which extends RenderBox

And we have to use errorPadding(assuming that contentPadding will be more than ErrorPadding)

packages/flutter/lib/src/material/input_decorator.dart
abikko: I am assuming that we will use it _EditableText RenderBox, for calculating the padding in _Decorator

examples/api/lib/material/text_field/text_field.4.dart
abikko: I think that in new release, I have to clearly explain how that ErrorPadding works! :)

abikko: Thanks for review, I appreciate that!

Tomorrow I will try to implement your suggestions, and write some tests and documentation

packages/flutter/lib/src/material/text_field.dart
justinmc: @abikko Do you still intend to come back to this PR and move this to the _layout method instead of a Padding widget?

justinmc (commented):
@danielkiing3 Thanks for chiming in here. When you say FormField I guess you mean TextFormField? I think you're right that any API we add here should probably affect both error and helper (which are positioned in the same place today).

dkwingsmt:
@abikko  Thanks for your contribution! Since I haven't seen a reply, I'm going to reluctantly close this pull request for now, but if you get time to address the comments, we'd be more than happy to take a look at a new patch. If you do send a new patch, it would be very useful to reference this one in the description in order to help reviewers see the previous context.

Issue #113977, opened 2022-10-25
Flutter build IPA with --export-options-plist not working


gkustas:
I have a white-label app that needs to be built from the command line. I am using "flutter build IPA" with an export options list to specify code signing specifics. It appears that the values are not getting passed to xcodebuild and the build fails because it doesn't find the provisioning profile.

```
error: No profiles for 'com.foundationtennis.app' were found: Xcode couldn't find any iOS App Development provisioning profiles matching 'com.foundationtennis.app'. Automatic signing is disabled and unable to generate a profile. To enable automatic signing, pass -allowProvisioningUpdates to xcodebuild. (in target 'Runner' from project 'Runner')
```

NOTE: if I open the XCode workspace and manually set the signing capabilities, flutter build IP works fine.

## Steps to Reproduce

1. Create ExportOptions.plist file like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>generateAppStoreInformation</key>
    <false/>
    <key>method</key>
    <string>app-store</string>
    <key>provisioningProfiles</key>
    <dict>
      <key>com.foundationtennis.app</key>
      <string>a38ecde3-53bc-4910-8068-604b01665598</string>
    </dict>
    <key>signingCertificate</key>
    <string>Apple Distribution</string>
    <key>signingStyle</key>
    <string>manual</string>
    <key>stripSwiftSymbols</key>
    <true/>
    <key>teamID</key>
    <string>844NRE74R7</string>
    <key>uploadBitcode</key>
    <false/>
    <key>uploadSymbols</key>
    <true/>
  </dict>
</plist>
```
2. Run flutter build IPA like so:

```
flutter build ios --release --verbose --export-options-plist=/Users/george/Develop/flutter_builds/foundation_tennis_v4/ios/ExportOptions.plist
```
3. See the output from the previous command below

## Logs

see attached for verbose output

</details>
[flutterbuildipa.log](https://github.com/flutter/flutter/files/9856148/flutterbuildipa.log)

huycozy:
@gkustas 
From your log, I see there is an error with CocoaPods:

```
Error output from CocoaPods:
           ↳
[        ]     Ignoring ffi-1.15.4 because its extensions are not built. Try: gem pristine ffi --version 1.15.4
[...]

Issue #106612, opened 2022-06-25
Support `flutter build ipa` with manual signing and provisioning profiles


michelesandroni:
Steps to Reproduce:

- on a MacBook M1, install vs code, flutter 3.0.2
- in vs code, create a new application project
- open the 'ios' project in xcode
- set the target's destination to any iOS Device (arm64, armv7)
- go to Targets > Runner > Signing & Capabilities
- turn OFF "Automatically manage signing"
- assign an existing Bundle Identifier 
- assign an existing Provisioning Profile
- go back to VS code and run the following in terminal:
`flutter build ipa`
- if you get an error "gen_snapshot_armv7 cannot be opened because the developer cannot be verified", click cancel, open the Settings application and go to Security & Privacy, then click "Allow Anyway" next to "gen_snapshot_armv7"
- after fixing the "gen_snapshot_armv7" error re-run:
`flutter build ipa`
- at this point the xcode build should start in the terminal, displaying the various phases (compiling, linking, signing) followed by archive generation and ipa generation, which should fail.

**Expected results:** flutter should generate an ipa file.

**Actual results:**

```
Archiving com.<COMPANY-REMOVED>.<APP-NAME-REMOVED> ...
Automatically signing iOS for device deployment using specified development team in Xcode project: <TEAM-NAME-REMOVED>
Running Xcode build...                                                  
 └─Compiling, linking and signing...                         7.0s
Xcode archive done.                                         24.8s
Built /Users/m1/Desktop/flutter_application_1/build/ios/archive/Runner.xcarchive.

💪 Building with sound null safety 💪

Building App Store IPA...                                           3.5s
Encountered error while creating the IPA:
error: exportArchive: "Runner.app" requires a provisioning profile.

Try distributing the app in Xcode: "open /Users/m1/Desktop/flutter_application_1/build/ios/archive/Runner.xcarchive"
```

**Note:**
I'm using manual signing because the actual app I'm working on uses the receive_sharing_intent package, which requires certain capabilities such as App Groups (required for the Share Extension to work) and Push Notifications (to avoid the warning email from Apple when submitting for app store publishing).
These capabilities in turn require the relative features enabled in the provisioning profile, which for some reason flutter (or xcode?) are apparently unable to handle automatically.

Manual signing is also convenient if you don't have an actual iOS device but you still want to export for the app store, otherwise xcode will throw the error "Failed to create provisioning profile. There are no devices registered in your account on the developer website.".

[...]

PR #177851, opened 2025-10-31
flutter_tools: correct iOS signing log for manual code signing (CODE_SIGN_STYLE=Manual)

MohammedTarigg:
## Summary

When running `flutter build ipa` with a project that uses manual code signing (`CODE_SIGN_STYLE=Manual`) for Release/Profile, the tool currently prints:

```
Automatically signing iOS for device deployment using specified development team in Xcode project: 93WG32C8SG
```

This is misleading in two ways:

1. The project is not using automatic signing for Release/Profile, it's explicitly using manual signing with a provisioning profile.
2. The build in this mode is typically for App Store/TestFlight distribution, not just device deployment.

Users who rely on manual signing (for example, Push Notifications + Sign in with Apple entitlements) often hit an `xcodebuild -exportArchive` failure later (`"Runner.app" requires a provisioning profile ...`) and assume Flutter overrode their signing, because of this log line. This confusion has been reported in the context of `flutter build ipa` where archive succeeds but export fails, and Flutter's messaging doesn't match the actual signing setup.

Related issues:
- https://github.com/flutter/flutter/issues/106612 - Users report the misleading "Automatically signing..." message even when using manual signing
- https://github.com/flutter/flutter/issues/113977 - ExportArchive failures with manual signing and ExportOptions.plist

## What this change does

* Before logging "Automatically signing iOS ...", we now check the active build configuration's `CODE_SIGN_STYLE`.
* If the style is `Manual`, we suppress the automatic-signing message (since manual signing is not automatic).
* If the style is `Automatic` or not set, we keep the original behavior.
* No functional behavior of signing or export is changed.

## Before

```
Automatically signing iOS for device deployment using specified development team in Xcode project: 93WG32C8SG
```

(This message appears even when `CODE_SIGN_STYLE=Manual`)

## After

```
(No message when CODE_SIGN_STYLE=Manual, since signing is not automatic)
```

(The message only appears when `CODE_SIGN_STYLE=Automatic` or not set)

## Tests

* Added test in `packages/flutter_tools/test/general.shard/ios/code_signing_test.dart` to assert we don't emit the automatic-signing message when `CODE_SIGN_STYLE=Manual`.
* All existing tests pass (40/40 tests in code_signing_test.dart).

## Why this helps

* Reduces confusion around manual signing / App Store export flows.
* Matches expectations from users who configure provisioning profiles manually and expect Flutter to respect that configuration.
* Prevents users from thinking Flutter is overriding their manual signing setup based on misleading log output.


MohammedTarigg:
Closing in favor of #177852 which contains only the intended changes on a clean branch.

PR #177852, opened 2025-10-31
flutter_tools: correct iOS signing log for manual code signing (CODE_SIGN_STYLE=Manual)

MohammedTarigg:
## Summary

When running `flutter build ipa` with a project that uses manual code signing (`CODE_SIGN_STYLE=Manual`) for Release/Profile, the tool currently prints:

```
Automatically signing iOS for device deployment using specified development team in Xcode project: x
```

This is misleading in two ways:

1. The project is not using automatic signing for Release/Profile, it's explicitly using manual signing with a provisioning profile.
2. The build in this mode is typically for App Store/TestFlight distribution, not just device deployment.

Users who rely on manual signing (for example, Push Notifications + Sign in with Apple entitlements) often hit an `xcodebuild -exportArchive` failure later (`"Runner.app" requires a provisioning profile ...`) and assume Flutter overrode their signing, because of this log line. This confusion has been reported in the context of `flutter build ipa` where archive succeeds but export fails, and Flutter's messaging doesn't match the actual signing setup.

**Related issues:**
- #106612 - Support `flutter build ipa` with manual signing and provisioning profiles
- #113977 - Flutter build IPA with --export-options-plist not working

## What this change does

* Before logging "Automatically signing iOS ...", we now check the active build configuration's `CODE_SIGN_STYLE`.
* If the style is `Manual`, we suppress the automatic-signing message (since manual signing is not automatic).
* If the style is `Automatic` or not set, we keep the original behavior.
* No functional behavior of signing or export is changed.

## Before

```
Automatically signing iOS for device deployment using specified development team in Xcode project: x
```

(This message appears even when `CODE_SIGN_STYLE=Manual`)

## After

```
(No message when CODE_SIGN_STYLE=Manual, since signing is not automatic)
```

(The message only appears when `CODE_SIGN_STYLE=Automatic` or not set)

## Tests

* Added test in `packages/flutter_tools/test/general.shard/ios/code_signing_test.dart` to assert we don't emit the automatic-signing message when `CODE_SIGN_STYLE=Manual`.
* All existing tests pass (40/40 tests in code_signing_test.dart).

## Why this helps

* Reduces confusion around manual signing / App Store export flows.
* Matches expectations from users who configure provisioning profiles manually and expect Flutter to respect that configuration.
* Prevents users from thinking Flutter is overriding their manual signing setup based on misleading log output.

## Breaking Changes

None. This change only affects log output and does not modify any signing behavior, export behavior, or CLI interface.

## Verification

Reviewers can verify this fix by:
1. Creating a test iOS project with `CODE_SIGN_STYLE=Manual` and `DEVELOPMENT_TEAM` set in Xcode project settings
2. Running `flutter build ipa --release --verbose`
3. Confirming the log does NOT show "Automatically signing iOS..."
4. Verifying the app still builds/signs correctly

The included unit test covers this scenario automatically.

---

Supersedes #177851 (that PR accidentally included unrelated commits from my fork history). This version only includes the intended change and tests.

### DIFF
--- a/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
+++ b/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
@@ -89,7 +89,7 @@ void main() {
             );
         expect(signingConfigs, isNull);
         // Should not log "Automatically signing..." when CODE_SIGN_STYLE is Manual.
-        expect(logger.statusText, isEmpty);
+        expect(logger.statusText, isNot(contains("Automatically signing iOS for device deployment using specified development team in Xcode project")));
         expect(logger.errorText, isEmpty);
         expect(logger.traceText, isEmpty);
       });
--- a/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
+++ b/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
@@ -89,7 +89,14 @@ void main() {
             );
         expect(signingConfigs, isNull);
         // Should not log "Automatically signing..." when CODE_SIGN_STYLE is Manual.
-        expect(logger.statusText, isNot(contains("Automatically signing iOS for device deployment using specified development team in Xcode project")));
+        expect(
+          logger.statusText,
+          isNot(
+            contains(
+              'Automatically signing iOS for device deployment using specified development team in Xcode project',
+            ),
+          ),
+        );
         expect(logger.errorText, isEmpty);
         expect(logger.traceText, isEmpty);
       });
--- a/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
+++ b/packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
@@ -70,6 +70,30 @@ void main() {
         expect(logger.traceText, isEmpty);
       });
 
+      testWithoutContext('No misleading log when CODE_SIGN_STYLE is Manual', () async {
+        final logger = BufferLogger.test();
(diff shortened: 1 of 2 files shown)


packages/flutter_tools/test/general.shard/ios/code_signing_test.dart
```diff
@@ -70,6 +70,30 @@ void main() {
...
+              logger: logger,
+              config: Config.test(),
+              terminal: FakeTerminal(),
+              fileSystem: MemoryFileSystem.test(),
+              fileSystemUtils: FakeFileSystemUtils(),
+              plistParser: FakePlistParser(),
+            );
+        expect(signingConfigs, isNull);
+        // Should not log "Automatically signing..." when CODE_SIGN_STYLE is Manual.
+        expect(logger.statusText, isEmpty);
```
navaronbracke: Just a drive-by comment, but you can match on that expectation explicitly. Testing on "isEmpty" doesn't feel like the right thing to test here.

```suggestion
        expect(logger.statusText, isNot(contains("Automatically signing iOS for device deployment using specified development team in Xcode project")));
```

MohammedTarigg: Updated the test to assert that the "Automatically signing..." message is not logged (instead of general isEmpty) when CODE_SIGN_STYLE is Manual. Thanks!

vashworth (approved):

LouiseHsu (approved):

PR #176597, opened 2025-10-06
[ios][platform_view]block gesture with hitTest by overlay layers

hellohuanlin:
This PR is an MVP version to address web view not tappable issue on iOS 26 (https://github.com/flutter/flutter/issues/175099). This issue cannot be fixed with `pointer_interceptor`, so if your issue can be fixed by it, it would be a different issue. 

This PR performance a **naive** hit test that decides to block platform view gestures if the touch falls within the "overlap" region between Flutter widgets and platform view. This is not a **complete** solution since there are a few drawback: 

1. There can be use cases where even if we touch outside of "overlap" (or no overlap at all), we still want to block gestures. 
2. overlaps are merged into a single UIView in our current implementation. This means if you have multiple overlaps, the blocking region would be a bounding box that contains all overlaps. 

To address the above drawbacks, a complete solution would require accessing Flutter's gesture arena, which will take longer time. 

https://github.com/user-attachments/assets/2fa26d8f-5342-4d87-bba2-5f5e5849816c

*List which issues are fixed by this PR. You must list at least one issue. An issue is not required if the PR fixes something trivial like a typo.*

Part of https://github.com/flutter/flutter/issues/175099

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[...]

Issue #176476, opened 2025-10-03
[iOS] - UiKitView behind a full-screen BottomSheet receives taps on iOS


Premnath2000:
### Steps to reproduce

1. Create a Flutter app with a CustomScrollView.

2. Add a SliverAppBar with a UiKitView in actions (e.g., right-side circle button).

3. Add a list below the app bar (e.g., SliverList with multiple items).

4. Add a bottom-aligned row of multiple UiKitViews.

5. Wrap list items with GestureDetector that shows a full-screen ModalBottomSheet (isScrollControlled: true).

6. Run on iOS (Simulator or device).

7. Tap the UiKitView inside the bottom sheet at a location overlapping the UiKitView behind.

### Expected results

No need to listen in the bottom sheet, and some rectangular shapes(Masked UikitView) should not be shown in the bottom sheet.

### Actual results

The UiKitView behind the bottom sheet also receives the tap.

Taps “pass through” the overlaying Flutter bottom sheet.

### Code sample

Main dart

```
import 'dart:math';
import 'package:flutter/material.dart';
import 'package:flutter/services.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      debugShowCheckedModeBanner: false,
      home: SliverAppBarDemo(),
    );
  }
}

class SliverAppBarDemo extends StatelessWidget {
  const SliverAppBarDemo({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: SafeArea(
        child: Stack(
          children: [
            CustomScrollView(
              slivers: [
                SliverAppBar(
                  toolbarHeight: 100,
                  backgroundColor: Colors.blue,
                  centerTitle: true,
                  floating: false,
                  pinned: false,
                  expandedHeight: 70,
                  leading: const SizedBox(width: 50),
                  title: const Text(
                    "Header Title",
                    style: TextStyle(
                      fontSize: 18,
                      color: Colors.white,
                      fontWeight: FontWeight.bold,
                    ),
                  ),
                  actions: [
                    Container(
                      margin: const EdgeInsets.only(right: 12),
                      width: 50,
                      height: 50,
                      color: Colors.red,
                      child: UiKitView(
                        key: const ValueKey('plusButton_glass_view'),
                        viewType: "circle_container_view",
                        creationParamsCodec: const StandardMessageCodec(),
                      ),
[...]

Issue #159814, opened 2024-12-04
[ios]Further investigate iOS 18.2 WebView behavior and potentially file a radar


hellohuanlin:
### Steps to reproduce

On iOS 18.2 the web view's gesture recognizer likely cached the old state of our `FlutterDelayingGestureRecognizer` (see PR description [here](https://github.com/flutter/engine/pull/56804)). 

This is likely a bug on iOS. We should probably create a reproducible project in UIKit and file the radar with Apple. 

### Expected results

NA

### Actual results

NA

### Code sample

<details open><summary>Code sample</summary>

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[Paste your output here]
```

</details>

gxl2024595:
So there's still an opportunity for asking Apple to fix this issue, and we go straight to upgrade iOS instead of upgrading Flutter sdk, right? Because upgrading Flutter SDK for some huge projects is a tough job. Perhaps solving this problem through Apple is the best solution for everyone.

pcba-dev:
@hellohuanlin any updates on this issue?

JDongKhan:
any progress?

DanielFrTB:
Hi,
Any progress with Apple about this issue?

Qingx:
Hi this issue has been fixed after flutter version 3.27.

jmagman:
Interesting. @hellohuanlin can you confirm? I wonder which commit changed this behavior on the Flutter side...

hellohuanlin:
> Also this is potentially a bug on iOS, so we may wanna create a reproducible project in UIKit and file the radar with Apple.

I am revisiting this. Will post here for any updates.

hellohuanlin:
Update: I was able to create a reproducible sample on native (see below). Will file a radar with Apple.  

```
import UIKit
import WebKit

class MyRecognizer: UIGestureRecognizer {
  var onTouchBegan: (() -> Void)? = nil
  init() {
    super.init(target: nil, action: nil)
  }
  override func shouldRequireFailure(of otherGestureRecognizer: UIGestureRecognizer) -> Bool {
    return false
  }
[...]

Issue #30143, opened 2019-03-28
WebView touchstart events are invoked even when the WebView is not hit tested on iOS


amirh:
Internal: b/126614142
Internal: b/285315999

Sample repro app:
```dart
import 'dart:async';
import 'package:flutter/material.dart';
import 'package:webview_flutter/webview_flutter.dart';

void main() => runApp(MaterialApp(home: WebViewExample()));

class WebViewExample extends StatelessWidget {
  final Completer<WebViewController> _controller =
      Completer<WebViewController>();

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: WebView(
        initialUrl: 'https://master-docs-flutter-io.firebaseapp.com/',
        javascriptMode: JavascriptMode.unrestricted,
        onWebViewCreated: (WebViewController webViewController) {
          _controller.complete(webViewController);
        },
        onPageFinished: (String url) {
          injectTouchListener();
        },
        javascriptChannels: <JavascriptChannel> [
          JavascriptChannel(name: 'Log', onMessageReceived: (JavascriptMessage msg) { print(msg.message); })
        ].toSet() ,
      ),
      floatingActionButton: FloatingActionButton(
        child: Container(color: Colors.orange,),
          onPressed: () {}
          ),
    );
  }
  Future<void> injectTouchListener() async {
    final WebViewController controller = await _controller.future;
    controller.evaluateJavascript("document.addEventListener('touchstart', (event) => { Log.postMessage(String(event)); }, true);");
  }
}
```
On iOS the message is printing even when tapping the FAB.

amirh:
cc @cyanglaz

cyanglaz:
Spent some time investigate this issue with @amirh and @chinmaygarde debugging this issue. We tried replacing 'touchstart' with a different event like 'click', it works fine. We think there could be some special case handling for events like 'touchstart' in iOS. We have not found any way to work around it.

Meanwhile, if anyone encounter this issue, I suggest working around it with a different event type.

asavartsov:
In my case, it even worse. When I tap a block in front of WebView, WebView receives touchstart, but does not receive touchend. And it will receive touchend ont he next tap with new touchstart right after it. This behavior breaks my Leaflet map completely. Adding IgnorePointer/AbsorbPointer does not help here. Somehow the page inside WebView still receives touchstart/touchend.

And it is probably connected to issues about broken map controls. I'm testing on iOS 13.6.

asavartsov:
[...]

PR #177859, opened 2025-10-31
[ios][pv]accept/reject gesture based on hit test

### LINKED ISSUE #165787: [iOS] - Ad Click Listener Breaks After Scroll Initiated on Ad (GAM Rich Media in Flutter ListView)
### Steps to reproduce

**Description**
Rich media ads served through Google Ad Manager stop responding to click events after a ListView is scrolled, provided that the scroll starts from the ad widget. This issue occurs specifically in Flutter apps but does not appear in fully native iOS applications.

**Steps to Reproduce**
1. Create campaign in GAM with provided ad code.
2. Embed ad from Google Ad Manager inside a Flutter ListView.
3. Ensure that text inside the ad is updated correctly when interacting with button inside the ad without scrolling the listview
4. Scroll the ListView, starting the gesture from the ad widget.
5. Try clicking the button with in the ad

**Affected Platforms**
✅ iOS (Flutter app & iOS platform views)
❌ Does not occur on fully native iOS apps

**Observations**
- Touchend event is not triggered on flutter app when touch is stopped. Once user taps background of the ad the touchend event is triggered.
- On fully native iOS app the touchend event is triggered once the user interaction stops.

**Tested on**
Flutter version: 3.29.0
Platform: iOS (18.4)

### Expected results

**Expected Behavior**
- The text should update correctly when clicking the button in the ad (indicating that click event listener is triggered).

### Actual results

**Actual Behavior**
- The ad’s click event listener no longer works after scrolling if the touch started on the ad widget.
- The issue also occurs when rendering the ad inside a Flutter plugin using a platform view on iOS.

### Code sample

Test Ad Script for GAM campaign:

```
<script>
 var newDiv = document.createElement('div');
 var textSpan = document.createElement('span');
 var newContent = document.createTextNode('Initial state');
 textSpan.appendChild(newContent);
 newDiv.appendChild(textSpan);
 newDiv.appendChild(document.createElement('br'));
 document.body.appendChild(newDiv);

 var button1 = document.createElement('button');
 button1.textContent = 'Example button';
 button1.style.cssText = 'width: 200px; height: 100px; margin-bottom: 25px;';
 newDiv.appendChild(button1);

 button1.addEventListener('click', function () { textSpan.textContent = 'Clicked'; });
 button1.addEventListener('touchstart', function () { textSpan.textContent = 'Touch Started'; });
 button1.addEventListener('touchend', function () { textSpan.textContent = 'Touch Ended'; });
</script>
```

### Screenshots or Video

<details open>

**iOS app**
https://github.com/user-attachments/assets/5a0f535d-3503-4421-80b4-7992a639e37b

**Flutter app**
https://github.com/user-attachments/assets/0a868956-da4d-4600-a801-7a2bc13f6d6c

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.29.0, on macOS 14.2 23C64 darwin-arm64, locale fi-FI)
[!] Xcode - develop for iOS and macOS (Xcode 15.4)
    ! CocoaPods 1.15.2 out of date (1.16.2 is recommended).
[✓] Android Studio (version 2023.2)
[✓] VS Code (version 1.98.0)
    ! Device emulator-5554 is offline.
[✓] Network resources

```

</details>

hellohuanlin:
Use hit test instead of delaying recognizer to block platform view gestures. This fixes web view and admob not tappable issue linked below. 

## Problem with delaying recognizer

There was an Apple bug that causes our delaying recognizer to conflict with web view's internal gesture recognizers. You can read more detailed research on this bug here: https://github.com/flutter/engine/pull/57032. Also, we repro'ed it in native app and [filed a radar with Apple](https://github.com/flutter/flutter/issues/159814#issuecomment-3215278821). 

## hit test & FFI to the rescue 

We perform a simple hitTest to decide whether to block the gesture or not. We had [an MVP version](https://github.com/flutter/flutter/pull/176597) (closed) that uses the platform view's "overlay layer". However, there are 2 issues with this MVP version: 
(1) There can be use cases where even if we touch outside of "overlap" (or no overlap at all), we still want to block gestures 
(2) overlaps are merged into a single UIView in our current implementation. This means if you have multiple overlaps, the blocking region would be a bounding box that contains all overlaps; 

There is also a [previous attempt](https://github.com/flutter/engine/pull/43778/files) to use hit test to block the gesture, where the embedder blocks the main thread, wait for the asynchronous response from framework, and decide to block or not. Given that UI & platform threads are merged, this will 100% cause deadlock (main thread waiting for main thread). 

Luckily, with FFI, we are able to query the framework and get the result **synchronously**, which solves the deadlock problem above. 

## Usage

To reduce risk of (serious) regression, I'd prefer to create a new policy `FlutterPlatformViewGestureRecognizersBlockingPolicyHitTest` (rather than reusing existing policy). Plugin owners would need to either use this policy, or create an API to allow change of polices. Developers can also overwrite their plugin locally, or use method swizzling. 

My next task would be to update 1p web_view & admob plugins, but I probably won't do it for 3p plugins. 

## Pointer interceptor deprecation

We had a pointer interceptor plugin created for a similar bug. Pointer interceptor doesn't work in https://github.com/flutter/flutter/issues/175099#issuecomment-3423752130 (webview) and https://github.com/flutter/flutter/issues/165787 (admob), because they don't have flutter widget on top of platform view. Another problem with pointer interceptor is that it requires developers to wrap their widgets. 

I can confirm that this hit test approach also fixes the [original issue](https://github.com/flutter/flutter/issues/30143) fixed by pointer interceptor. So with this, we will be able to deprecate pointer_interceptor plugin (which will be my next TODO item).

## Misc

This is not related to https://github.com/flutter/flutter/issues/176476, which is likely a separate Apple bug that we should file a radar

Fixes https://github.com/flutter/flutter/issues/175099
Fixes https://github.com/flutter/flutter/issues/165787

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -308,6 +308,11 @@ void _dispatchPointerDataPacket(ByteData packet) {
 }
 
 @pragma('vm:entry-point')
+bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
+  return PlatformDispatcher.instance._platformViewShouldAcceptGesture(viewId, x, y);
+}
+
+@pragma('vm:entry-point')
 void _dispatchSemanticsAction(int viewId, int nodeId, int action, ByteData? args) {
   PlatformDispatcher.instance._dispatchSemanticsAction(viewId, nodeId, action, args);
 }
@@ -414,6 +419,24 @@ void _invoke3<A1, A2, A3>(
     zone.runGuarded(() {
       callback(arg1, arg2, arg3);
     });
+  }
+}
+
+R? _invoke3WithReturn<A1, A2, A3, R>(
+  R Function(A1 a1, A2 a2, A3 a3)? callback,
+  Zone zone,
+  A1 arg1,
+  A2 arg2,
+  A3 arg3,
+) {
+  if (callback == null) {
+    return null;
+  }
+  if (identical(zone, Zone.current)) {
+    return callback(arg1, arg2, arg3);
+  } else {
+    // TODO: how to handle synchronous return? Should we assume zone is always current?
+    return null;
   }
 }
 
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -422,6 +422,12 @@ void _invoke3<A1, A2, A3>(
   }
 }
 
+/// Invokes [callback] inside the given [zone] passing it [arg1], [arg2], and [arg3],
+/// and returns a nullable result of the specified type.
+///
+/// The 3 in the name refers to the number of arguments expected by
+/// the callback (and thus passed to this function, in addition to the
+/// callback itself and the zone in which the callback is executed).
 R? _invoke3WithReturn<A1, A2, A3, R>(
   R Function(A1 a1, A2 a2, A3 a3)? callback,
   Zone zone,
@@ -435,8 +441,15 @@ R? _invoke3WithReturn<A1, A2, A3, R>(
   if (identical(zone, Zone.current)) {
     return callback(arg1, arg2, arg3);
   } else {
-    // TODO: how to handle synchronous return? Should we assume zone is always current?
-    return null;
+    // TODO(hellohuanlin): add `zone.runGuardedWithReturn` API.
+    try {
+      return zone.run(() {
+        return callback(arg1, arg2, arg3);
+      });
+    } catch (e, s) {
+      zone.handleUncaughtError(e, s);
+      return null;
+    }
   }
 }
 
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,6 +43,7 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptGesture].
 typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
 
 /// Signature for responses to platform messages.
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,6 +43,8 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
+
 /// Signature for responses to platform messages.
 ///
 /// Used as a parameter to [PlatformDispatcher.sendPlatformMessage] and
@@ -1336,6 +1338,14 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  PlatformViewShouldAcceptGestureCallback? get onPlatformViewShouldAcceptGesture =>
+      _onPlatformViewShouldAcceptGesture;
+  PlatformViewShouldAcceptGestureCallback? _onPlatformViewShouldAcceptGesture;
+  Zone _onPlatformViewShouldAcceptGestureZone = Zone.root;
+  set onPlatformViewShouldAcceptGesture(PlatformViewShouldAcceptGestureCallback? callback) {
+    _onPlatformViewShouldAcceptGesture = callback;
+  }
+
   // Called from the engine via hooks.dart.
   void _updateFrameData(int frameNumber) {
     final FrameData previous = _frameData;
@@ -1371,6 +1381,17 @@ class PlatformDispatcher {
         arguments: args,
       ),
     );
+  }
+
+  bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
+    return _invoke3WithReturn<int, double, double, bool>(
+          onPlatformViewShouldAcceptGesture,
+          _onPlatformViewShouldAcceptGestureZone,
+          viewId,
+          x,
+          y,
+        ) ??
+        false;
   }
 
   ErrorCallback? _onError;
--- a/engine/src/flutter/lib/ui/window/platform_configuration.cc
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.cc
@@ -77,6 +77,9 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  // The embedded platform view (e.g. UIView on iOS) is called "platform view"
+  // on framework side, but "embedded view" on engine side. On the other hand,
+  // "platform view" refers to the whole flutter view on engine side.
   embedded_view_should_accept_gesture_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library,
@@ -406,6 +409,9 @@ bool PlatformConfiguration::EmbeddedViewShouldAcceptGesture(
       embedded_view_should_accept_gesture_.Get(),
       {tonic::ToDart(view_id), tonic::ToDart(touch_began_location.x),
        tonic::ToDart(touch_began_location.y)});
+  if (tonic::CheckAndHandleError(dart_result)) {
+    return false;
+  }
   return tonic::DartConverter<bool>::FromDart(dart_result);
 }
 
--- a/engine/src/flutter/lib/ui/window/platform_configuration.cc
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.cc
@@ -77,6 +77,10 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  embedded_view_should_accept_gesture_.Set(
+      tonic::DartState::Current(),
+      Dart_GetField(library,
+                    tonic::ToDart("_platformViewShouldAcceptGesture")));
   dispatch_semantics_action_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchSemanticsAction")));
@@ -386,6 +390,23 @@ void PlatformConfiguration::DispatchPointerDataPacket(
 
   tonic::CheckAndHandleError(
       tonic::DartInvoke(dispatch_pointer_data_packet_.Get(), {data_handle}));
+}
+
+bool PlatformConfiguration::EmbeddedViewShouldAcceptGesture(
+    int64_t view_id,
+    const flutter::PointData& touch_began_location) {
+  std::shared_ptr<tonic::DartState> dart_state =
+      embedded_view_should_accept_gesture_.dart_state().lock();
+  if (!dart_state) {
+    return false;
+  }
+  tonic::DartState::Scope scope(dart_state);
+
+  Dart_Handle dart_result = tonic::DartInvoke(
+      embedded_view_should_accept_gesture_.Get(),
+      {tonic::ToDart(view_id), tonic::ToDart(touch_began_location.x),
+       tonic::ToDart(touch_began_location.y)});
+  return tonic::DartConverter<bool>::FromDart(dart_result);
 }
 
 void PlatformConfiguration::DispatchSemanticsAction(int64_t view_id,
--- a/engine/src/flutter/lib/ui/window/platform_configuration.cc
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.cc
@@ -395,7 +395,7 @@ void PlatformConfiguration::DispatchPointerDataPacket(
       tonic::DartInvoke(dispatch_pointer_data_packet_.Get(), {data_handle}));
 }
 
-bool PlatformConfiguration::EmbeddedViewShouldAcceptGesture(
+bool PlatformConfiguration::EmbeddedNativeViewShouldAcceptGesture(
     int64_t view_id,
     const flutter::PointData& touch_began_location) {
   std::shared_ptr<tonic::DartState> dart_state =
--- a/engine/src/flutter/lib/ui/window/platform_configuration.h
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.h
@@ -15,6 +15,7 @@
 #include "flutter/fml/time/time_point.h"
 #include "flutter/lib/ui/semantics/semantics_update.h"
 #include "flutter/lib/ui/window/platform_message_response.h"
+#include "flutter/lib/ui/window/point_data.h"
 #include "flutter/lib/ui/window/pointer_data_packet.h"
 #include "flutter/lib/ui/window/view_focus.h"
 #include "flutter/lib/ui/window/viewport_metrics.h"
@@ -461,6 +462,10 @@ class PlatformConfiguration final {
   ///
   void DispatchPointerDataPacket(const PointerDataPacket& packet);
 
+  bool EmbeddedViewShouldAcceptGesture(
+      int64_t view_id,
+      const flutter::PointData& touch_began_location);
+
   //----------------------------------------------------------------------------
   /// @brief      Notifies the framework that the embedder encountered an
   ///             accessibility related action on the specified node. This call
@@ -584,6 +589,7 @@ class PlatformConfiguration final {
   tonic::DartPersistentValue update_accessibility_features_;
   tonic::DartPersistentValue dispatch_platform_message_;
   tonic::DartPersistentValue dispatch_pointer_data_packet_;
+  tonic::DartPersistentValue embedded_view_should_accept_gesture_;
   tonic::DartPersistentValue dispatch_semantics_action_;
   tonic::DartPersistentValue begin_frame_;
   tonic::DartPersistentValue draw_frame_;
--- a/engine/src/flutter/lib/ui/window/platform_configuration.h
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.h
@@ -463,7 +463,8 @@ class PlatformConfiguration final {
   void DispatchPointerDataPacket(const PointerDataPacket& packet);
 
   //----------------------------------------------------------------------------
-  /// @brief      Requests from the engine if an embedded view should accept
+  /// @brief      Requests from the engine if an embedded native view should
+  /// accept
   ///             gesture at a given touch location.
   ///
   ///
@@ -474,7 +475,7 @@ class PlatformConfiguration final {
   /// @return     true if the embedded view should accept gesture; false
   /// otherwise.
   ///
-  bool EmbeddedViewShouldAcceptGesture(
+  bool EmbeddedNativeViewShouldAcceptGesture(
       int64_t view_id,
       const flutter::PointData& touch_began_location);
 
--- a/engine/src/flutter/lib/ui/window/platform_configuration.h
+++ b/engine/src/flutter/lib/ui/window/platform_configuration.h
@@ -462,6 +462,18 @@ class PlatformConfiguration final {
   ///
   void DispatchPointerDataPacket(const PointerDataPacket& packet);
 
+  //----------------------------------------------------------------------------
+  /// @brief      Requests from the engine if an embedded view should accept
+  ///             gesture at a given touch location.
+  ///
+  ///
+  /// @param[in]  view_id               The identifier of the flutter view that
+  ///                                   hosts the embedded view.
+  /// @param[in]  touch_began_location  The touch began location of a gesture.
+  ///
+  /// @return     true if the embedded view should accept gesture; false
+  /// otherwise.
+  ///
   bool EmbeddedViewShouldAcceptGesture(
       int64_t view_id,
       const flutter::PointData& touch_began_location);
--- a/engine/src/flutter/lib/ui/window/point_data.h
+++ b/engine/src/flutter/lib/ui/window/point_data.h
@@ -1,0 +1,17 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_LIB_UI_WINDOW_POINT_DATA_H_
+#define FLUTTER_LIB_UI_WINDOW_POINT_DATA_H_
+
+namespace flutter {
+
+// Represents a point on screen.
+struct PointData {
+  double x;
+  double y;
+};
+}  // namespace flutter
+
+#endif  // FLUTTER_LIB_UI_WINDOW_POINT_DATA_H_
--- a/engine/src/flutter/runtime/runtime_controller.h
+++ b/engine/src/flutter/runtime/runtime_controller.h
@@ -478,7 +478,8 @@ class RuntimeController : public PlatformConfigurationClient,
   bool DispatchPointerDataPacket(const PointerDataPacket& packet);
 
   //----------------------------------------------------------------------------
-  /// @brief      Requests from the engine if an embedded view should accept
+  /// @brief      Requests from the engine if an embedded native view should
+  /// accept
   ///             gesture at a given touch location.
   ///
   ///
@@ -489,7 +490,7 @@ class RuntimeController : public PlatformConfigurationClient,
   /// @return     true if the embedded view should accept gesture; false
   /// otherwise.
   ///
-  bool EmbeddedViewShouldAcceptGesture(
+  bool EmbeddedNativeViewShouldAcceptGesture(
       int64_t view_id,
       const flutter::PointData& touch_began_location);
 
--- a/engine/src/flutter/runtime/runtime_controller.h
+++ b/engine/src/flutter/runtime/runtime_controller.h
@@ -18,6 +18,7 @@
 #include "flutter/lib/ui/text/font_collection.h"
 #include "flutter/lib/ui/ui_dart_state.h"
 #include "flutter/lib/ui/window/platform_configuration.h"
+#include "flutter/lib/ui/window/point_data.h"
 #include "flutter/lib/ui/window/pointer_data_packet.h"
 #include "flutter/lib/ui/window/pointer_data_packet_converter.h"
 #include "flutter/lib/ui/window/view_focus.h"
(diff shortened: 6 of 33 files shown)


engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -417,6 +422,24 @@ void _invoke3<A1, A2, A3>(
...
+  A3 arg3,
+) {
+  if (callback == null) {
+    return null;
+  }
+  if (identical(zone, Zone.current)) {
+    return callback(arg1, arg2, arg3);
+  } else {
+    // TODO: how to handle synchronous return? Should we assume zone is always current?
+    return null;
```
hellohuanlin: I'm a bit iffy about this part, and we don't have prior example of synchronously return from the framework - when I test it, `zone` is always `current`, so maybe we can make a safe assumption here? I am also curious when will zone be different. @chunhtai

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -77,6 +77,10 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  embedded_view_should_accept_gesture_.Set(
```
hellohuanlin, line 83: For the embedder and the framework, I use the term "platform view". However, in the engine, I have to use the term "embedded view" to disambiguate with the "platform view" (which seems to represent the whole flutter view). This seems to be the existing convention if i'm reading it right.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,11 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+@pragma('vm:entry-point')
+bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
```
hellohuanlin, line 311: I'm debating myself about calling it `_platformViewShouldAcceptGesture` vs `_iOSPlatformViewShouldAcceptGesture` (here and rest of the PR). I'm open to input.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -417,6 +422,24 @@ void _invoke3<A1, A2, A3>(
   }
 }
 
+R? _invoke3WithReturn<A1, A2, A3, R>(
```
hellohuanlin, line 431: can't think of a better name, open to suggestion.

hellohuanlin:
@yjbanov just FYI since you are working on something related on the web.

engine/src/flutter/lib/ui/hooks.dart
chunhtai: gemini replied

Engine Callbacks: The Flutter engine itself, when it calls into Dart code (for example, to dispatch a pointer event or a platform message), might not be running in the application's zone. The 
 functions ensure that the framework's callbacks for these events are run in the correct zone.

so unless the code path from after dart ffi has introduced a new zone, which is controlled by us, I think they will always be the same.

packages/flutter_tools/lib/src/commands/assemble.dart
```diff
@@ -279,7 +279,9 @@ class AssembleCommand extends FlutterCommand {
       }
       final int indexEquals = chunk.indexOf('=');
       if (indexEquals == -1) {
-        throwToolExit('Improperly formatted define flag: $chunk');
+        // TODO: revert this change. seems like a tools bug.
+        // throwToolExit('Improperly formatted define flag: $chunk');
+        continue;
```
vashworth: This is fixed upstream by https://github.com/flutter/flutter/pull/177198

engine/src/flutter/lib/ui/hooks.dart
vashworth: So looking at `_invoke3`, it calls `zone.runGuarded` when the zones are not identical. If you step into that, it says that `runGuarded` is basically just running it in a try catch and handling the error on catch. So I wonder if you could do something like this to try and mimic it? And then file an issue against the Dart SDK to have them add a `runGuarded` with a return type?

```dart
 try {
   return zone.run(() {
      return callback(arg1, arg2, arg3);
    });
 } catch (e, s) {
   zone.handleUncaughtError(e, s);
   return null;
 }
```


https://github.com/dart-lang/sdk/blob/21ab9e1ac59fbeaf9eaa01dab934830d01aa224f/sdk/lib/async/zone.dart#L801-L814

engine/src/flutter/lib/ui/hooks.dart
vashworth: I think the name is inline with the existing invokes. Please add a doc comment, though

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -43,6 +43,8 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
```
vashworth, line 47: Please add doc comment to this and the others below

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -388,6 +392,23 @@ void PlatformConfiguration::DispatchPointerDataPacket(
...
+    int64_t view_id,
+    const flutter::PointData& touch_began_location) {
+  std::shared_ptr<tonic::DartState> dart_state =
+      embedded_view_should_accept_gesture_.dart_state().lock();
+  if (!dart_state) {
+    return false;
+  }
+  tonic::DartState::Scope scope(dart_state);
+
+  Dart_Handle dart_result = tonic::DartInvoke(
```
vashworth, line 408: Looks like most methods wrap `tonic::DartInvoke` in `tonic::CheckAndHandleError`, should yours also do that?

https://github.com/flutter/flutter/blob/0908c2e225e88af5a01c8b0435489cfaec874644/engine/src/flutter/third_party/tonic/logging/dart_error.h#L20-L58

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -461,6 +462,10 @@ class PlatformConfiguration final {
   ///
   void DispatchPointerDataPacket(const PointerDataPacket& packet);
 
+  bool EmbeddedViewShouldAcceptGesture(
```
vashworth: Please add doc comment

engine/src/flutter/runtime/runtime_controller.h
```diff
@@ -476,6 +477,10 @@ class RuntimeController : public PlatformConfigurationClient,
   ///
   bool DispatchPointerDataPacket(const PointerDataPacket& packet);
 
+  bool EmbeddedViewShouldAcceptGesture(
```
vashworth: Please add doc comment

engine/src/flutter/shell/common/engine.h
```diff
@@ -829,6 +829,22 @@ class Engine final : public RuntimeDelegate, PointerDataDispatcher::Delegate {
   void DispatchPointerDataPacket(std::unique_ptr<PointerDataPacket> packet,
                                  uint64_t trace_flow_id);
 
+  //----------------------------------------------------------------------------
+  /// @brief      Requests the engine if an embedded view should accept
```
vashworth:
```suggestion
  /// @brief      Requests from the engine if an embedded view should accept
```

engine/src/flutter/lib/ui/window/platform_configuration.cc
vashworth: I think this deserves a comment in code to explain the change

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -573,8 +580,26 @@ - (BOOL)containsWebView:(UIView*)view remainingSubviewDepth:(int)remainingSubvie
   return NO;
 }
 
+- (UIView*)hitTest:(CGPoint)point withEvent:(UIEvent*)event {
```
vashworth, line 583: Does the `_flutterViewController` need to be reset like it does in `touchesBegan`?

https://github.com/flutter/flutter/blob/aed2e42321b2783f7d0f9b3233c2656dd60e8c04/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L734-L737

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -573,8 +580,26 @@ - (BOOL)containsWebView:(UIView*)view remainingSubviewDepth:(int)remainingSubvie
...
+  if (_blockingPolicy == FlutterPlatformViewGestureRecognizersBlockingPolicyHitTest &&
+      event.type == UIEventTypeTouches) {
```
vashworth: Does this work with a mouse/keyboard? Does it work with scrolling?

engine/src/flutter/shell/common/platform_view.h
```diff
@@ -182,6 +182,22 @@ class PlatformView {
     virtual void OnPlatformViewDispatchPointerDataPacket(
         std::unique_ptr<PointerDataPacket> packet) = 0;
 
+    //--------------------------------------------------------------------------
+    /// @brief      Requests the delegate if an embedded view should accept
```
vashworth:
```suggestion
    /// @brief      Requests from the delegate if an embedded view should accept
```

engine/src/flutter/lib/ui/hooks.dart
LouiseHsu: since its ios specific i vote the second one

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: Ah woof, I made an assumption that this is doing some sort of async hopping. Thanks for stepping into the code.

engine/src/flutter/lib/ui/hooks.dart
chunhtai: Code after platformViewiOS should be platform neutral, in case other platform also have use case for this method. If you want to specify the use case in iOS, you can put it in the api doc that currently only iOS uses this.

engine/src/flutter/lib/ui/window/platform_configuration.cc
hellohuanlin: <s>I don't think the current API supports `DartInvoke` that returns a value. But I can open an issue in Dart. </s>

I misread - will add a check before return.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: no need - hitTest is called before `touchBegan` starts, so we still have `touchBegan` covering this reset logic.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: Good observation - that's something I am still testing. Will report back

packages/flutter_tools/lib/src/commands/assemble.dart
hellohuanlin: nice!

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: Good point. Though I couldn't find an existing convention with a platform prefix, so I am not sure... (both about this line, as well as all boilerplates in the engine)

@chunhtai do you have any thoughts on it?

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -388,6 +392,23 @@ void PlatformConfiguration::DispatchPointerDataPacket(
       tonic::DartInvoke(dispatch_pointer_data_packet_.Get(), {data_handle}));
 }
 
+bool PlatformConfiguration::EmbeddedViewShouldAcceptGesture(
```
chunhtai: We need to unify the naming, Replace EmbeddedView with PlatformView?

engine/src/flutter/lib/ui/window/platform_configuration.cc
chunhtai: At some point we should use ffigen to generate the ffi binding. https://github.com/flutter/flutter/issues/175816 hopefully this won't be needed in that reality

engine/src/flutter/lib/ui/hooks.dart
chunhtai: this name is fine, since platformView only has one meaning in framework side, and this makes sense

engine/src/flutter/lib/ui/window/platform_configuration.cc
chunhtai: here and everywhere, we should not change the naming convention in the middle of call chain

engine/src/flutter/lib/ui/window/point_data.h
```diff
@@ -0,0 +1,17 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_LIB_UI_WINDOW_POINT_DATA_H_
+#define FLUTTER_LIB_UI_WINDOW_POINT_DATA_H_
+
+namespace flutter {
+
+// Represents a point on screen.
+struct PointData {
```
chunhtai, line 11: do you forseen addition to the pointerData? feels a bit overkill

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -461,6 +462,10 @@ class PlatformConfiguration final {
...
+      int64_t view_id,
+      const flutter::PointData& touch_began_location);
```
chunhtai, line 480: consider just use two double x and y and pass by value

packages/flutter/lib/src/gestures/binding.dart
```diff
@@ -14,6 +14,7 @@ import 'dart:collection';
 import 'dart:ui' as ui show PointerDataPacket;
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter/rendering.dart';
```
chunhtai: gesture layer can't import rendering. this is layer violation

packages/flutter/lib/src/gestures/binding.dart
```diff
@@ -319,6 +322,17 @@ mixin GestureBinding on BindingBase implements HitTestable, HitTestDispatcher, H
...
 
+  bool _handlePlatformViewShouldAcceptGesture(int viewId, double x, double y) {
+    final HitTestResult result = HitTestResult();
+    hitTestInView(result, Offset(x, y), viewId);
+
+    if (result.path.isEmpty) {
+      return false;
+    }
+    final HitTestTarget firstHit = result.path.first.target;
+    return firstHit is RenderUiKitView;
```
chunhtai: we may need a more general way of doing this instead of RenderObject type check.

Consider create an abstract interface in this layer to be extends in Rendering layer.

also what if there are multiple PlatformView?

engine/src/flutter/lib/ui/window/platform_configuration.cc
hellohuanlin: See discussion here: https://github.com/flutter/flutter/pull/177859/files#r2482734556

"platform view" means the flutter view in engine. 

If we replace it with "PlatformView", we will end up with an API in platform view's delegate:  

```
OnPlatformViewPlatformViewShouldAcceptGesture
```

Where the first "PlatformView" refers to the flutter view, and the second "PlatformView" refers to the embedded view. This will be far more confusing.

engine/src/flutter/lib/ui/window/platform_configuration.h
hellohuanlin: `x`, `y` was my first attempt, and I changed it to `PointData` for more idiomatic practice followed by different platforms (e.g. Apple's `CGPoint`, dart's `Offset)

engine/src/flutter/lib/ui/window/point_data.h
hellohuanlin: Not sure if i follow (pointer_data is another type related to pointer packets). could you clarify?

packages/flutter/lib/src/gestures/binding.dart
hellohuanlin: I may need some help understanding this and how to resolve it. Thanks!

packages/flutter/lib/src/gestures/binding.dart
hellohuanlin: > Consider create an abstract interface in this layer to be extends in Rendering layer.

Could you elaborate more? thanks

> also what if there are multiple PlatformView?

We should be fine - the request was initiated from the platform view on top.

engine/src/flutter/lib/ui/window/platform_configuration.cc
chunhtai: embedded view sounds like an add-to-app. How about embeddedNativeView?

engine/src/flutter/lib/ui/window/point_data.h
chunhtai: was going to suggest using two doubles, but see your other comment. continue discussion there

engine/src/flutter/lib/ui/window/platform_configuration.h
chunhtai: sg

packages/flutter/lib/src/gestures/binding.dart
chunhtai: image

See this graph libraries can't import other libraries that stack on top, otherwise it will be a circular dependency.

One way to resolve this is to have a abstract class in this layer

```dart
// in gesture layer
abstract NativeHitTarget {}
``` 

```dart
// in Render layer
class RenderUIKitView extends NativeHitTarget{
}
```

and then in here you can check

```dart
target is NativeHitTarget
```

hellohuanlin: Makes sense! Thanks for the diagram!

engine/src/flutter/lib/ui/window/platform_configuration.cc
hellohuanlin: SG

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -573,8 +580,25 @@ - (BOOL)containsWebView:(UIView*)view remainingSubviewDepth:(int)remainingSubvie
   return NO;
 }
 
+- (UIView*)hitTest:(CGPoint)point withEvent:(UIEvent*)event {
+  if (_blockingPolicy == FlutterPlatformViewGestureRecognizersBlockingPolicyHitTest) {
+    CGPoint pointInFlutterView = [self convertPoint:point toView:self.flutterViewController.view];
+    // Block gesture if the framework instructed so (after performing its own hitTest).
+    if (![self.flutterViewController
+            platformViewShouldAcceptGestureAtTouchBeganLocation:pointInFlutterView]) {
+      return self;
```
vashworth, line 589: If I understand this correctly, if the point is within the FlutterView, the framework will return true. If it returns true, it goes on to super hitTest, which uses the frontmost view that contains the point. If it returns false, it uses the platform view.

Couple questions I'm a little confused on:
1. Why do we need to ask the framework at all?
2. What happens if there are multiple layered platform views?
3. What happens if the FlutterView is above/below the platform view?

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -77,6 +77,13 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  // The embedded platform view (e.g. UIView on iOS) is called "platform view"
+  // on framework side, but "embedded view" on engine side. On the other hand,
+  // "platform view" refers to the whole flutter view on engine side.
+  embedded_view_should_accept_gesture_.Set(
```
vashworth, line 83: Nit: Rename to "embedded_native_view..." to keep inline
```suggestion
  embedded_native_view_should_accept_gesture_.Set(
```

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -584,6 +602,7 @@ class PlatformConfiguration final {
   tonic::DartPersistentValue update_accessibility_features_;
   tonic::DartPersistentValue dispatch_platform_message_;
   tonic::DartPersistentValue dispatch_pointer_data_packet_;
+  tonic::DartPersistentValue embedded_view_should_accept_gesture_;
```
vashworth, line 605:
```suggestion
  tonic::DartPersistentValue embedded_native_view_should_accept_gesture_;
```

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -77,6 +77,13 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  // The embedded platform view (e.g. UIView on iOS) is called "platform view"
+  // on framework side, but "embedded view" on engine side. On the other hand,
```
vashworth, line 81:
```suggestion
  // on framework side, but "embedded native view" on engine side. On the other hand,
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: > If I understand this correctly, if the point is within the FlutterView, the framework will return true.

No, the query to the framework is "Is the top most hit target a platform view at given point?". 

Side note: "top most hit target" != "point inside platform view" (see the drop down menu example below)

> ... which uses the frontmost view that contains the point. 

Almost but not always - it also depends on the front-most view (which is the platform view). If the platform view itself overwrites `hitTest`, and say, redirect to another arbitrary target, it will be respected too. 

> If it returns false, it uses the platform view.

The opposite - if it returns false, that means platform view shouldn't receive the touch. HitTest is a **pre-order** tree traversal (visiting parent before children). We return `self` here to tell UIKit, "hey I (the touch interceptor view) will take care of the touch, so don't visit my children (the platform view)"

> Why do we need to ask the framework at all?

My MVP version actually skips the framework query and relies on overlay layer. However, there are cases where overlay layer doesn't work. For example, when drop down menu is shown, there will be an invisible "barrier" covering the whole screen, which won't create an overlay since it's invisible. 

> What happens if there are multiple layered platform views?

It's fine - UIKit will make sure the correct touch interceptor view receives the touch and sends requests to framework. 

> What happens if the FlutterView is above/below the platform view?

I assume FlutterView you mean a flutter widget covering the platform view? 

If we have flutter widget *above* the platform view - framework would return false from hitTest, so the touch interceptor would block the gesture (by returning `self` in hitTest). If *below*, then the opposite.

jmagman (changes requested):
> I'd prefer to create a new policy

I think this needs a serious design doc before it's landed, particularly around how devs are supposed to know how to use this policy, the migration and documentation story. 

Obviously this would be better if this works for everyone automatically without plugins chasing policies. 
Additionally, this will allow other platform teams to converge on similar solutions.

chunhtai:
(triage) @hellohuanlin is this pr ready for another look?

xiaopingtang:
@hellohuanlin any update for this?

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -43,6 +43,9 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptGesture].
+typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
```
loic-sharma, line 47: I believe with the current API adding a new parameter to this callback would be a breaking change.

I'd consider moving all these arguments to a `FooEvent` type to improve the API's forward compatibility (you can add a new property Adding a new property to `FooEvent` wouldn't be a breaking

(You might also want to consider returning an enum instead of a boolean, but I feel less strongly about this)

loic-sharma, line 47: Instead of providing the callback a `int viewId`, I'd consider providing a `FlutterView view`. The callback can easily get the ID from a view, but not vice versa

loic-sharma: I'd also consider accepting an `Offset offset` instead of `double x, double y` - that's more ergonomic for Flutter. See above: https://github.com/flutter/flutter/pull/177859#discussion_r2591197837

loic-sharma: I'd consider:

```suggestion
class HitTestRequest {
  const HitTestRequest({required this.view, required this.offset});

  final FlutterView view;
  final Offset offset;
}

class HitTestResponse {
  const HitTestResponse({required this.firstHitIsPlatformView});
  
  final bool firstHitIsPlatformView;
}

typedef HitTestCallback = HitTestResponse Function(HitTestRequest request);
```

This makes the API a bit more generic, less tied to iOS's specific needs, and allows us to add more features in the future.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: Apparently i was wrong - we have to set flutterViewController in hitTest, because the flutterViewController may not be set when creating the platform view (race condition). The comment here is probably incorrect and very misleading - we should just set it once so that it's not nil (rather than set it on every touch began). 

Credit: https://github.com/flutter/flutter/issues/175099#issuecomment-3635663560 by @holzgeist

chunhtai:
Hi @hellohuanlin should we close this one since there is another pr https://github.com/flutter/flutter/pull/179659?

hellohuanlin:
Closing in favor of https://github.com/flutter/flutter/pull/179659/

PR #177954, opened 2025-11-03
wires up SetSemanticsTreeEnabled to android accessibility bridge

chunhtai:
Before the change, the bridge only listens to updateSemantics when it receives signal from the OS

After the change, the bridge always listens to updateSemantics, but will clear out caches when SetSemanticsTreeEnabled is called with false

## Pre-launch Checklist

- [ ] I signed the [CLA].


engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -312,7 +313,8 @@ private void setAccessibleNavigation(boolean value) {
   private boolean isReleased = false;
 
   // Handler for all messages received from Flutter via the {@code accessibilityChannel}
-  private final AccessibilityChannel.AccessibilityMessageHandler accessibilityMessageHandler =
+  @VisibleForTesting
+  public final AccessibilityChannel.AccessibilityMessageHandler accessibilityMessageHandler =
```
mboetger: Can we remove the public if it's just for testing? Looks like the test is in the same package.

engine/src/flutter/shell/platform/android/test/io/flutter/embedding/engine/FlutterJNITest.java
```diff
@@ -274,6 +274,21 @@ public void createOverlaySurface_callsPlatformViewsController() {
     verify(platformViewsController, times(1)).createOverlaySurface();
   }
 
+  @Test
```
mboetger, line 278: Might as well verify the negative case (e.g. `setSemanticsTreeEnabled(false)`) while we're here.

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -389,6 +391,13 @@ public void updateSemantics(
         public void setLocale(String locale) {
           AccessibilityBridge.this.setLocale(locale);
         }
+
+        @Override
+        public void setSemanticsTreeEnabled(boolean enabled) {
```
mboetger: This doesn't do anything if `enabled` is true.  Should we just rename this to `resetSemanticsTree`?  What's the point of the `enabled`?  It does not stop the delegate from receiving updates from the framework (if they are sent).

mboetger (changes requested):
I'm not sure why we are passing a boolean when the method does nothing when the parameter is `true`.

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
chunhtai: renamed, in other platform, they start to initialize resource(AccessibilityBridge is created) when enabled = true and disposed when enabled= false.

The Android embedding, the resource between a11ybridge, a11ychannel, JNI is tangling together that it is hard to follow the other platform's practice.

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
chunhtai: looks like it is required

  /b/s/w/ir/cache/builder/engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java:184: error: accessibilityMessageHandler has private access in AccessibilityBridge
      verify(mockChannel).setAccessibilityMessageHandler(bridge.accessibilityMessageHandler);

mboetger: It looks to me like you set it to `private final`.  If you just remove the `public` and set it to JUST `final`, in java that will be considered 'package private' and since the tests are in the same package it should work.

chunhtai: fixed

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -1641,6 +1659,13 @@ void updateSemantics(
      * <p>Must be called on the main thread
      */
     void setLocale(@NonNull String locale);
+
+    /**
+     * Invoked by native to notify framework started or stopped compiling accessibility tree.
```
mboetger:
```suggestion
     * Invoked by native to notify embedder to reset accessibility tree.
```

mboetger (approved):

PR #177959, opened 2025-11-04
Impeller: Allows R32G32B32A32_SFLOAT images

### LINKED ISSUE #141289: Support for floating point textures
### Use case

This is an extension to #136025.

Fragment shaders sometimes need lookup tables, for example for color transformations. These tables usually use floating point, since that's the natural way math works on GPUs. GLSL handles this transparently, and APIs like OpenGL have support for this via floating point textures (`GL_RGBA32F`, `GL_R32F`, etc).

### Proposal

If Flutter adds support for sync texture construction, this should also allow supplying a floating point texture to enable these kinds of operations in fragment shaders.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/141289

design doc: https://docs.google.com/document/d/1zpkMutZkqo2GVdMhiKzFURpgN8JDZTjtIVwwqqHKM90/edit?tab=t.0

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/lib/ui/painting/image_decoder.h
```diff
@@ -34,14 +34,25 @@ class ImageDecoder {
 
   using ImageResult = std::function<void(sk_sp<DlImage>, std::string)>;
 
+  enum PixelFormat {
+    kUnknown,
+    kOptimal,
+    kR32G32B32A32Float,
+  };
+
+  struct Options {
```
chinmaygarde, line 45: Nit: Perhaps rename this to `DecodeOptions`?

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -75,7 +76,196 @@ bool IsWideGamut(const SkColorSpace* color_space) {
...
+        .makeColorSpace(SkColorSpace::MakeSRGB());
+  } else {
+    return base_image_info.makeWH(decode_size.width(), decode_size.height())
+        .makeColorType(ChooseCompatibleColorType(base_image_info.colorType()))
+        .makeAlphaType(alpha_type)
+        .makeColorSpace(SkColorSpace::MakeSRGB());
+  }
+}
+
+struct DecodedBitmap {
```
chinmaygarde, line 153: Nit: Since you have followed the absl::StaturOr pattern elsewhere, perhaps make this struct not contain a separate error field and just use absl::StatusOr (or optional).

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -75,7 +76,196 @@ bool IsWideGamut(const SkColorSpace* color_space) {
...
+    bitmap->setImmutable();
+  }
+  return {.bitmap = bitmap, .allocator = bitmap_allocator};
+}
+
+DecodedBitmap HandlePremultiplication(
+    const std::shared_ptr<SkBitmap>& bitmap,
+    std::shared_ptr<ImpellerAllocator> bitmap_allocator,
+    const std::shared_ptr<impeller::Allocator>& allocator) {
+  if (bitmap->alphaType() != SkAlphaType::kUnpremul_SkAlphaType) {
```
chinmaygarde, line 203: Add a trace event here?

engine/src/flutter/lib/ui/painting/image_decoder_impeller.h
```diff
@@ -58,15 +50,28 @@ class ImageDecoderImpeller final : public ImageDecoder {
...
-              uint32_t target_height,
+              const Options& options,
               const ImageResult& result) override;
 
+  struct ImageInfo {
+    impeller::ISize size;
+    impeller::PixelFormat format = impeller::PixelFormat::kUnknown;
+  };
+
+  struct DecompressResult {
```
chinmaygarde, line 66: Ditto on StatusOr instead instead of an error field.

engine/src/flutter/lib/ui/painting/image_decoder_impeller.h
```diff
@@ -58,15 +50,28 @@ class ImageDecoderImpeller final : public ImageDecoder {
...
+    std::shared_ptr<impeller::DeviceBuffer> device_buffer;
+    ImageInfo image_info;
+    std::optional<SkImageInfo> resize_info = std::nullopt;
```
chinmaygarde: Optionals are auto initialized to std::nullopt.

engine/src/flutter/lib/ui/painting/image_decoder.h
```diff
@@ -34,14 +34,25 @@ class ImageDecoder {
 
   using ImageResult = std::function<void(sk_sp<DlImage>, std::string)>;
 
+  enum PixelFormat {
```
chinmaygarde: Can this be called `TargetPixelFormat` or `DecodedPixelFormat`? Also, what is kUnknown? Is it the same as kOptimal or kDontCare? Perhaps add some docstrings?

chinmaygarde: Instead of kOptimal, I like kDontCare instead. With optimal, you have to explain how its optimal. With kDontCare, you are leaving it up to the implementation. Which is whats happening here.

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -75,7 +76,196 @@ bool IsWideGamut(const SkColorSpace* color_space) {
...
+          scaled_bitmap->pixmap(),
+          SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone))) {
+    FML_LOG(ERROR) << "Could not scale decoded bitmap data.";
+  }
+  scaled_bitmap->setImmutable();
+
+  std::shared_ptr<impeller::DeviceBuffer> buffer =
+      scaled_allocator->GetDeviceBuffer();
+  if (!buffer) {
+    return {.decode_error = "Unable to get device buffer"};
```
chinmaygarde: Ditto on statusor.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -1909,6 +1909,10 @@ enum PixelFormat {
   /// component, followed by: green, blue and alpha. Premultiplied alpha isn't
   /// used, matching [ImageByteFormat.rawExtendedRgba128].
   rgbaFloat32,
+
+  /// This is an unspecified pixel format that leaves it to Flutter's discretion
+  /// to choose the optimal pixel format.
+  optimal,
```
chinmaygarde: How about `kDontCare`.

Whatever name you choose, perhaps also specify when someone might want to use this. Something like "Use this when the image is going to be sampled the application doesn't care about the exact format."

chinmaygarde: Wait, this can't be right. This struct is given to decodeImageFromPixels. This isn't the target pixel format, its the incoming pixel format. How can Flutter divine this format?

chinmaygarde (commented):
The major thing I didn't quite get was how incoming pixel format could be "optimal".

gaaclarke:
> The major thing I didn't quite get was how incoming pixel format could be "optimal".

It couldn't be, that would result in a runtime error.  The PixelFormat type has 2 uses, it defines the input format for decoding pixels and it defines the target pixel format for the texture.  Not all values are valid for for each of those inputs.  Alternatively we could have made a different enum "TargetPixelFormat"?

chinmaygarde:
> Alternatively we could have made a different enum "TargetPixelFormat"

Ooh. I really like that idea. Instead of conflating the two ideas.

engine/src/flutter/lib/ui/painting.dart
flar: Don't Care sounds a bit judgmental, like you've given up on your expectations from a pesky implementation.

kImplementationDependent?
kAny?
kBestForConditions?
kUnspecified?

Or how about just leaving it null to mean "do what you want"? You only set this when you care.

engine/src/flutter/lib/ui/painting/image_decoder.h
gaaclarke: The full name is `ImageDecoder::Options` and `ImageDecoder` only does one thing, `Decode()` so I think this fine.

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
gaaclarke: done

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
gaaclarke: done

engine/src/flutter/lib/ui/painting/image_decoder_impeller.h
gaaclarke: done

engine/src/flutter/lib/ui/painting/image_decoder_impeller.h
gaaclarke: done

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -85,12 +275,13 @@ static std::optional<impeller::PixelFormat> ToPixelFormat(SkColorType type) {
       return impeller::PixelFormat::kR16G16B16A16Float;
     case kBGR_101010x_XR_SkColorType:
       return impeller::PixelFormat::kB10G10R10XR;
+    case kRGBA_F32_SkColorType:
+      return impeller::PixelFormat::kR32G32B32A32Float;
```
chinmaygarde, line 279: I am pretty sure this requires a capability check (or am I mistaken). At least for the GL backend I'd think. Perhaps we should fallback to the DontCare case?

chinmaygarde (approved):
The new updates LGTM. I had a concern about device capability checks that I think we can document our way out of.

engine/src/flutter/lib/ui/painting/image_decoder_impeller.cc
gaaclarke: Not according to my investigation in gpuinfo.org.  I have that in the design doc.  I think it says something like 95% compatibility but it also says that for r8g8b8a8.  It doesn't seem to go to 100%

gaaclarke: Not according to my investigation in gpuinfo.org. I have that in the design doc. I think it says something like 95% compatibility but it also says that for r8g8b8a8. It doesn't seem to go to 100%

gaaclarke:
@chinmaygarde thanks for giving it another look, i responded to all your comments but I forgot to hit "submit" then github lost them.

gaaclarke:
rebase to try to kick CI

PR #178033, opened 2025-11-05
Remove WindowingOwner.hasTopLevelWindows

robert-ancell:
This function doesn't seem to serve any purpose.
Top level is not a term we are currently using in the multi window code. If we want to track if we have open windows, this can be done in the Flutter framework - it doesn't need to be done at the platform level.


robert-ancell:
Perhaps this function was pre-FFI where you needed to get this information from native code? If the purpose is to track windows created in native code and not in the Dart platform then this should be `hasNativeWindows()` and the Dart created windows tracked in a platform independent way.

mattkae (approved):
Agreed! Just need to fix CI

PR #178081, opened 2025-11-05
Listen to text spacing overrides on the web

### LINKED ISSUE #142712: [web:a11y] respect text spacing browser settings
Internal bug: b/338058572

### Use case

Independent of the web renderers is it not possible to fulfil the WCAG Success Criterion 1.4.12 Text Spacing as described here: https://codepen.io/2kool2/pen/LYzVeLO.

The criterion requires text spacing depending on user settings.

### Proposal

The web engine should respect the user settings or should allow the style manipulation to fulfil the mentioned criterion.

Tested with flutter version: `Flutter (Channel stable, 3.16.9, on macOS 14.3 23D56 darwin-arm64, locale de-DE)`

https://github.com/flutter/flutter/assets/122024451/06d81da7-1f4a-44a1-8801-3b84176c69fe

<details>
  <summary>
  Used the flutter demo project:
  </summary>

  ```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        //
        // TRY THIS: Try running your application with "flutter run". You'll see
        // the application has a purple toolbar. Then, without quitting the app,
        // try changing the seedColor in the colorScheme below to Colors.green
        // and then invoke "hot reload" (save your changes or press the "hot
        // reload" button in a Flutter-supported IDE, or press "r" if you used
        // the command line to start the app).
        //
        // state is not lost during the reload. To reset the state, use hot
        // restart instead.
        //
        // This works for code too, not just values: Most code changes can be
        // tested with just a hot reload.
        useMaterial3: true,
      ),
    );
  }
}

  // how it looks.

  // always marked "final".

  final String title;

  @override
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });
  }

  @override
    //
    return Scaffold(
      appBar: AppBar(
        // TRY THIS: Try changing the color here to a specific color (to
        // Colors.amber, perhaps?) and trigger a hot reload to see the AppBar
        // change color while the other colors stay the same.
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          //
          // horizontal).
          //
          // TRY THIS: Invoke "debug painting" (choose the "Toggle Debug Paint"
          // action in the IDE, or press "p" in the console), to see the
          // wireframe for each widget.
          children: <Widget>[
            const Text(
            ),
            Text(
              '$_counter',
            ),
            const Spacer(),
            const Text(
                "Irure minim pariatur incididunt laboris aliquip dolor amet mollit excepteur. Laboris sed irure in adipiscing excepteur ut occaecat mollit commodo laborum aliquip id amet. Nostrud aliquip velit sit amet enim non commodo occaecat cillum fugiat minim reprehenderit tempor anim. Laborum ut eiusmod sunt quis adipiscing magna aliqua qui culpa lorem commodo cupidatat aute pariatur sint officia minim tempor. Irure do mollit sunt sint ut ipsum ex in proident pariatur quis laboris cillum nisi mollit occaecat. Sint excepteur voluptate nostrud nostrud aliqua nulla ad sint cillum anim nisi cupidatat non consectetur aliquip aliquip ut."),
            const Text(
                "Proident officia consectetur laborum in aute sint aute commodo amet incididunt. Enim officia in id duis pariatur occaecat dolore eu magna dolor magna excepteur enim eiusmod lorem velit esse. Dolor ea proident sunt ullamco nostrud deserunt nulla ipsum velit voluptate nisi aliquip do deserunt fugiat ipsum occaecat cupidatat."),
            const Text(
                "Duis dolor duis ad sed pariatur magna eiusmod velit et in nulla anim ut sunt irure incididunt do. Nisi veniam cillum consequat aliquip non ipsum minim elit tempor fugiat esse est proident minim dolore elit dolore duis. In ad aliquip sit enim mollit officia exercitation ad ut consequat. Aliqua voluptate aute aliquip enim dolore nulla aliqua enim non sunt laborum occaecat mollit deserunt."),
            const Text(
                "Ullamco lorem lorem cupidatat adipiscing deserunt id eiusmod magna velit. Elit aute elit duis fugiat sed labore amet cupidatat do incididunt labore ea. Amet sit laborum qui sint anim nisi ipsum nulla officia et officia. Duis irure et ipsum sint exercitation culpa lorem aute cupidatat pariatur reprehenderit fugiat. Veniam est consectetur ex mollit ipsum mollit incididunt commodo do tempor nulla enim veniam laboris ullamco ad minim culpa eu. Aliquip esse et esse ullamco eu nisi laboris irure magna."),
          ],
        ),
      ),
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
    );
  }
}
```

Flutter run commands: 
- `flutter run -d chrome --web-renderer canvaskit`
- `flutter run -d chrome --web-renderer html`

Pasted in terminal to reproduce the user setting:
```
javascript:(function(){const style=document.createElement('style');style.setAttribute('rel','stylesheet'),style.textContent='*{line-height:1.5 !important;letter-spacing:0.12em !important;word-spacing,0.16em !important;}p{margin-bottom:2em !important;}',document.head.appendChild(style);})();
```

</details>

### PR DESCRIPTION
**Pull request by Renzo-Olivares**
Original PR/Discussion: https://github.com/flutter/flutter/pull/172915

# Framework:
* `EditableText`/`SelectableText`, applies `lineHeightScaleFactorOverride`, `wordSpacingOverride`, and `letterSpacingOverride` to it's `TextStyle` similarly to how we already do for bold platform overrides. Note `SelectableText` is built on `EditableText` so it also applies these overrides.
* `Text`, applies `lineHeightScaleFactorOverride`, `wordSpacingOverride`, and `letterSpacingOverride` to it's `TextStyle` similarly to how we already do for bold platform overrides.
* Exposes line height override through `MediaQueryData.lineHeightScaleFactorOverride` and `maybeLineHeightScaleFactorOverrideOf(context)`.
* Exposes letter spacing override through `MediaQueryData.letterSpacingOverride` and `maybeLetterSpacingOverrideOf(context)`.
* Exposes word spacing override through `MediaQueryData.wordSpacingOverride` and `maybeWordSpacingOverrideOf(context)`.
* Exposes paragraph spacing override through `MediaQueryData.paragraphSpacingOverride` and `maybeParagraphSpacingOverrideOf(context)`.
* `MediaQuery.applyTextStyleOverrides()` \ `MediaQueryData.applyTextStyleOverrides()` to be able to reset/override the text spacing settings on `MediaQueryData`.

# Engine:
* Introduces new members on `PlatformDispatcher` API that hold the text spacing properties that are overridden on the web.
* We provide the `lineHeightScaleFactorOverride`, `letterSpacingOverride`, `wordSpacingOverride`, and `paragraphSpacingOverride` on the web by attaching a `ResizeObserver` to an off-screen hidden element, when its size changes we capture its text spacing CSS properties, and notify the framework through `onMetricsChanged`.

Fixes #142712

https://github.com/user-attachments/assets/aaaa3e74-c232-4956-acd2-ae1a4487e415

## Pre-launch Checklist

- [x] I signed the [CLA].

### CODE REVIEW

File: engine/src/flutter/lib/web_ui/lib/src/engine/safe_browser_api.dart [hunk 1]
```diff
@@ -56,6 +56,15 @@ num? parseFontSize(DomElement element) {
...
 }
 
+/// Parses the given style property [attributeName] of [element] and returns the
+/// [resolved value](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value) without a unit.
+///
+/// Returns `null` if the property value is not numeric (e.g., 'normal',
+/// 'auto') or cannot be parsed.
+num? parseNumericStyleProperty(DomElement element, String attributeName) {
+  return parseFloat(domWindow.getComputedStyle(element).getPropertyValue(attributeName));
+}
```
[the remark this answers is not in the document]
Renzo-Olivares: `parseFloat` already handles the `NaN` case.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1834,6 +1891,10 @@ class _PlatformConfiguration {
       defaultRouteName: defaultRouteName ?? this.defaultRouteName,
       systemFontFamily: systemFontFamily ?? this.systemFontFamily,
       configurationId: configurationId ?? this.configurationId,
+      lineHeightScaleFactorOverride: lineHeightScaleFactorOverride,
+      letterSpacingOverride: letterSpacingOverride,
+      wordSpacingOverride: wordSpacingOverride,
+      paragraphSpacingOverride: paragraphSpacingOverride,
```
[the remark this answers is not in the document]
Renzo-Olivares: This is intentional as this is a private class, these specific values cannot be reset to null from `copyWith`, and they are not currently being used so for the time being they are excluded from `copyWith`. The existing values are always preserved.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -4822,7 +4845,9 @@ class EditableTextState extends State<EditableText>
       textScaler: effectiveTextScaler,
       textHeightBehavior: widget.textHeightBehavior ?? DefaultTextHeightBehavior.maybeOf(context),
       locale: widget.locale,
-      structStyle: widget.strutStyle,
+      structStyle: widget.strutStyle.copyWith(
+        forceStrutHeight: lineHeightScaleFactor == null ? null : false,
+      ),
```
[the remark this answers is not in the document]
Renzo-Olivares: In `editable.dart` it looks like the `TextPainter` is getting created with the same `StrutStyle` that is passed down from `EditableText`.

mdebbar (approved):
web side looks good to me!

packages/flutter/lib/src/painting/strut_style.dart
```diff
@@ -340,27 +340,28 @@ class StrutStyle with Diagnosticable {
...
-       fontWeight = fontWeight ?? textStyle.fontWeight,
-       fontStyle = fontStyle ?? textStyle.fontStyle,
-       debugLabel = debugLabel ?? textStyle.debugLabel,
-       _package = package; // the textStyle._package data is embedded in the
-  // fontFamily names, so we no longer need it.
+  }) : this(
+         fontFamily: fontFamily != null
+             ? (package == null ? fontFamily : 'packages/$package/$fontFamily')
+             : textStyle.fontFamily,
+         fontFamilyFallback: fontFamilyFallback ?? textStyle.fontFamilyFallback,
```
LongCatIsLooong, line 353: the font family fallback here needs the package path prefix.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -4469,7 +4479,19 @@ class EditableTextState extends State<EditableText>
       }
 
       spellCheckResults = SpellCheckResults(text, suggestions);
-      renderEditable.text = buildTextSpan();
+      final double? lineHeightScaleFactor = mounted
```
LongCatIsLooong: why's the mounted check needed?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -6739,3 +6773,50 @@ class _EditableTextTapUpOutsideAction extends ContextAction<EditableTextTapUpOut
...
   }
 }
+
+/// A utility class for overriding the text styles of a [TextSpan] tree.
+// When changes are made to this class, the equivalent API in text.dart
+// must also be updated.
+// TODO(Renzo-Olivares): Remove after investigating a solution for overriding all
+// styles for children in an [InlineSpan] tree, see: https://github.com/flutter/flutter/issues/177952.
+class _OverridingTextStyleTextSpanUtils {
+  static TextSpan applyTextSpacingOverrides({
```
LongCatIsLooong, line 6782: nit: maybe remove the utility class an expose the functions instead?

packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -654,6 +680,53 @@ class MediaQueryData {
   ///    supported.
   final bool supportsShowingSystemContextMenu;
 
+  /// The height of the text, as a multiple of the font size.
```
LongCatIsLooong: nit: maybe: text override, same for the other two fields

packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -940,6 +1067,12 @@ class MediaQueryData {
     gestureSettings,
     Object.hashAll(displayFeatures),
     supportsShowingSystemContextMenu,
+    Object.hash(
```
LongCatIsLooong, line 1070: why are these 4 grouped together?

packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -1005,6 +1142,34 @@ class MediaQuery extends InheritedModel<_MediaQueryAspect> {
...
+  ///    affected properties of the [MediaQueryData].
+  MediaQuery.applyTextStyleOverrides({
+    super.key,
+    required BuildContext context,
+    required double? lineHeightScaleFactorOverride,
+    required double? letterSpacingOverride,
+    required double? wordSpacingOverride,
+    required double? paragraphSpacingOverride,
+    required super.child,
+  }) : data = MediaQuery.of(context).applyTextStyleOverrides(
```
LongCatIsLooong: This looks error-prone. `data` is evaluated on construction so if the user caches this widget this doesn't get rebuilt even if the ancestor media query changes.

LongCatIsLooong: I'd just wrap this in a builder, and you don't have to take `context` if you do that.

LongCatIsLooong: Or move this to `MediaQueryData`.

packages/flutter/lib/src/widgets/text.dart
```diff
@@ -706,6 +722,26 @@ class Text extends StatelessWidget {
...
+      wordSpacing: wordSpacing,
+      textSpan: TextSpan(
+        style: effectiveTextStyle,
+        text: data,
+        locale: locale,
+        children: textSpan != null ? <InlineSpan>[textSpan!] : null,
+      ),
+    );
+    final StrutStyle? effectiveStrutStyle = strutStyle != null && lineHeightScaleFactor != null
+        ? strutStyle!.merge(StrutStyle(height: lineHeightScaleFactor))
```
LongCatIsLooong: nit: don't do the strutStyle null check and use `?` instead?

LongCatIsLooong (approved):
LGTM with nits

packages/flutter/lib/src/widgets/editable_text.dart
Renzo-Olivares: Because this logic is happening in an async method, the lint complained we were not checking mounted. https://dart.dev/tools/linter-rules/use_build_context_synchronously

packages/flutter/lib/src/painting/strut_style.dart
Renzo-Olivares: I think it does end up doing this but through its `getter`. The main constructor also defers to this getter.

```dart
List<String>? get fontFamilyFallback {
  if (_package != null && _fontFamilyFallback != null) {
    return _fontFamilyFallback.map((String family) => 'packages/$_package/$family').toList();
  }
  return _fontFamilyFallback;
}
```

packages/flutter/lib/src/widgets/media_query.dart
Renzo-Olivares: We have exceeded the 20 objects that this method takes, so I ended up grouping them. Should I do something else instead in this case?

packages/flutter/lib/src/widgets/editable_text.dart
Renzo-Olivares: I think I like the class because i'm able to place the `TODO` on the entire class. Are there any benefits to placing them directly in `Text` as functions?

packages/flutter/lib/src/widgets/media_query.dart
Renzo-Olivares: Updated to `Overrides the height of the text...`

packages/flutter/lib/src/widgets/editable_text.dart
loic-sharma: On line 4476 we return if `suggestions == null`. Could we update that to return also if `!mounted`, and then simplify this logic? I don't feel strongly, just a thought.

packages/flutter/lib/src/widgets/media_query.dart
loic-sharma: Oh good catch. +1 to wrapping this in a builder and removing the `context` argument.

packages/flutter/test/widgets/selectable_text_test.dart
```diff
@@ -5598,6 +5598,31 @@ void main() {
...
+
+      final RenderEditable renderEditable = findRenderEditable(tester);
+      final TextStyle? resultTextStyle = renderEditable.text?.style;
+      expect(resultTextStyle?.height, 2.0);
+      expect(resultTextStyle?.letterSpacing, 2.0);
+      expect(resultTextStyle?.wordSpacing, 2.0);
+      expect(renderEditable.strutStyle?.height, 2.0);
+    },
+  );
   group('context menu', () {
```
loic-sharma, line 5627: Consider:

```suggestion

  group('context menu', () {
```

I don't feel strongly, feel free to keep as-is if you prefer

packages/flutter/test/widgets/media_query_test.dart
```diff
@@ -1104,6 +1104,78 @@ void main() {
...
+    late MediaQueryData withTextStyleOverrides;
+    await tester.pumpWidget(
+      MediaQuery(
+        data: const MediaQueryData(
+          lineHeightScaleFactorOverride: 9.0,
+          letterSpacingOverride: 9.0,
+          wordSpacingOverride: 9.0,
+          paragraphSpacingOverride: 9.0,
+        ),
+        child: Builder(
```
loic-sharma: If you switch to Weiyu's suggestion of adding a `Builder` in `MediaQuery.applyTextStyleOverrides`, don't forget to remove the `Builder`s from these tests :)

loic-sharma (approved):
LGTM, excellent work!

packages/flutter/lib/src/widgets/editable_text.dart
Renzo-Olivares: I think we should not exit early if  `suggestions != null` and `!mounted` in that case we still want the spell check to work and that should continue to work as expected as it didn't access the `context` before. 

`suggestions == null || !mounted` - returns early even when spell check might be expected

packages/flutter/test/widgets/media_query_test.dart
Renzo-Olivares: Good call, I did end up going with the `Builder` suggestion.

packages/flutter/lib/src/widgets/media_query.dart
Renzo-Olivares: Went with the `Builder` suggestion. Thank you!

packages/flutter/lib/src/widgets/editable_text.dart
LongCatIsLooong: Uhm the state should stop doing spell checking and stop receiving spell checking updates as soon as it's disposed so there should be no need for the mount check? Do we have that lint enabled in the framework?

Renzo-Olivares: Yeah this lint is currently enabled in the framework https://github.com/flutter/flutter/blob/e74c5954502f51a5cb2089320767dfab8f611168/analysis_options.yaml#L249 .

Renzo-Olivares: > On line 4476 we return if suggestions == null. Could we update that to return also if !mounted, and then simplify this logic? I don't feel strongly, just a thought.

I ended up going with this suggestion. It makes sense that we don't spell check at all while unmounted.

packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -719,6 +792,56 @@ class MediaQueryData {
...
+      paragraphSpacingOverride: paragraphSpacingOverride,
+    );
+  }
+
+  /// Creates a copy of this media query data but with the
+  /// `lineHeightScaleFactorOverride`, `letterSpacingOverride`,
+  /// `wordSpacingOverride`, and `paragraphSpacingOverride` replaced
+  /// with the given values.
+  ///
+  /// If an argument is null (the default), then this [MediaQueryData]
```
chunhtai, line 807: This seems like a weird API. Does this mean if they just one to change one property, they have to still provide the rest with the ancestor MediaQueryData? What is the use case of this method?

chunhtai (commented):
Just a question on the new media override wrapper, otherwise LGTM

packages/flutter/lib/src/widgets/media_query.dart
Renzo-Olivares: Yeah they still need to provide the rest with the ancestor `MediaQueryData`. The reason for this is that users will likely want to be able to disable any automatic text spacing at their discretion. `copyWith` may have been a good option here but we cannot reset values to nullable through it without using a sentinel approach or similar. This API was the alternative we arrived at so users can still disable their text spacing settings.

loic-sharma: Agreed this is a weird API, but Renzo, Weiyu, Justin, and I all discussed this in last week's text input sync ([Google internal link](https://docs.google.com/document/d/1vimZnqmmv5ZjTc9mDZ73cX3ztM6hlAXTSiznd5gB80c/edit?resourcekey=0-9Zl8Zp8pdQwYsSRm8sLcCA&tab=t.0#heading=h.95ldjol5ufja)), and this was the API we preferred.

This PR is getting a bit time sensitive (this needs to be rolled into google3 before Friday), so @Renzo-Olivares I'm good with merging this PR without fully addressing this comment. If needed, we can send a subsequent PR to address any follow-up comments by @chunhtai :)

Piinks: Yes, this has multiple approvals from relevant stakeholders and had extensive discussion about the API design. Let's get this in to unblock a critical cutoff date.

PR #178007, opened 2025-11-04
Revert "Refactor OverlayPortal semantics (#173005)"

chingjun:
This reverts commit ccf6466b14e144971a7d65983bdd16459d65a62a.

~This is a test PR for investigation, the commit seems to be breaking a customer app.~

The original PR caused b/457553134

PR #173005, opened 2025-07-30
Refactor OverlayPortal semantics

QuncCccccc:
Fixes https://github.com/flutter/flutter/issues/163576
Fixes https://github.com/flutter/flutter/issues/175184

This PR refactored the grafting part on `OverlayPortal`. Originally, the semantics tree of `OverlayPortal` was constructed/grafted in render object phase to make sure the correctness of the traversal order. However this resulted wrong hit-test order and the issue surfaced on web. With the fact that on web we are not able to graft/correct hit-test order tree, this PR:
* Reverts the original grafting of the `OverlayPortal` so the hit-test order is always correct. 
* Then, we adds the grafting and updates the traversal order when we send `childrenInTraversalOrder` to engine.
* Updating `childrenInTraversalOrder` causes it have different length from the length of `childrenInHitTestOrder` and wrong hit-test transform of the `OverlayPortal` children because when the transform is calculated, it assumes a correct traversal order. To fix these issues, this PR also:
  * recalculates the transform for `OverlayPortal` children.
  * adds `hitTestTransform` property and pass it to Android engine.
  * skip grafting for web because it assumes the same length of `childrenInTraversalOrder` and `childrenInHitTestOrder`.
  * added grafting by using `ARIA-owns` in web engine to fix the traversal order.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #178095, opened 2025-11-06
Reland "Refactor OverlayPortal semantics (#173005)"

QuncCccccc:
Reverts flutter/flutter#178007

This PR is to reland https://github.com/flutter/flutter/pull/173005 and add a fix to avoid infinite loop. The fix doesn't contain engine changes.


packages/flutter/lib/src/widgets/basic.dart
```diff
@@ -4006,7 +4006,7 @@ sealed class _SemanticsBase extends SingleChildRenderObjectWidget {
     required int? currentValueLength,
     required String? identifier,
     required Object? traversalParentIdentifier,
-    required Set<Object>? traversalChildIdentifier,
+    required Object? traversalChildIdentifier,
```
QuncCccccc, line 4009: This should be `Object` instead of `Set<Object>?`. This is missed because we haven't directly used it in OverlayPortal. Catched this when adding tests.

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -4022,13 +4022,30 @@ class SemanticsNode with DiagnosticableTreeMixin {
...
-        // grafted to be a child of an overlay portal parent node that has the
+        // If the child node is a traversal child, but the current node is
+        // not a traversal parent, it means the child node should be
+        // grafted to be a child of a traversal parent node that has the
         // same identifier as the child. So this child should be removed from
         // the current node's children list; i.e., we don't add it to
         // updatedChildren list.
-        continue;
+        //
+        // A corner case is the traversal parent of the traversal child, in paint
```
chunhtai, line 4032: so something like this?

```
// assume root node id is 0
Semantics(
   // assume semantics node id = 1
   traversalParent: 'key',
   child: Semantics(
      // assume semantics node id = 2
     traversalChildren: 'key',
   )
)
```

In this case I assume we want the hittest tree to be
0 -> 1 -> 2

but traversal tree to be

0 -> 2 -> 1

QuncCccccc: Talked with @chunhtai offline. In this case we throw exception to indicate wrong parent-child relationship. Also update SemanticsConfiguration.isCompatibleWith to avoid implicit merging.

chunhtai (approved):
LGTM

ValentinVignal:
Ok noted, thank you for the confirmation :)

PR #178102, opened 2025-11-06
Add new motion accessibility features to iOS.

### LINKED ISSUE #177330: Expose iOS "Customize onscreen motion" accessibility settings
_(Replaces #177195, which was about just one setting. This issue covers the whole bundle of related settings.)_

### Use case

In the Zulip app for iOS, we'd like to check whether the iOS 17+ "Auto-Play Animated Images" setting is enabled, so that when loading an image, we can choose between animated and still versions.

- iOS user doc: https://support.apple.com/guide/iphone/reduce-onscreen-motion-iph0b691d3ed/ios
- iOS developer doc: https://developer.apple.com/documentation/swiftui/environmentvalues/accessibilityplayanimatedimages

### Proposal

I propose adding new fields to [`AccessibilityFeatures`](https://api.flutter.dev/flutter/dart-ui/AccessibilityFeatures-class.html), providing "Auto-Play Animated Images" and, for completeness, all the related settings in that [Apple user doc](https://support.apple.com/guide/iphone/reduce-onscreen-motion-iph0b691d3ed/ios):

- [x] Reduce Motion _(already exposed, as [`AccessibilityFeatures.reduceMotion`](https://api.flutter.dev/flutter/dart-ui/AccessibilityFeatures/reduceMotion.html))_
- [ ] Vehicle Motion Cues
- [ ] Dim Flashing Lights
- [ ] Auto-Play Animated Images _(for Zulip's use case; was tracked in #177195)_
- [ ] Auto-Play Video Previews
- [ ] Auto-Play Message Effects
- [ ] Prefer Non-Blinking Cursor
- [ ] Limit Frame Rate

We would follow the existing pattern of iOS-only fields, such as [`AccessibilityFeatures.onOffSwitchLabels`](https://api.flutter.dev/flutter/dart-ui/AccessibilityFeatures/onOffSwitchLabels.html), by saying in the dartdocs "Only supported on iOS."

ksokolovskyi:
Closes https://github.com/flutter/flutter/issues/177330
Contributes to https://github.com/flutter/flutter/issues/175878

### Description

This PR implements support for three new iOS accessibility motion features.
- `Auto-Play Animated Images`: Informs the app when the user has chosen to pause automatically playing GIFs and other animated content.
- `Auto-Play Video Previews`: Informs the app when the user has disabled the automatic playback of video previews.
- `Prefer Non-Blinking Cursor`: Informs the app that the user prefers a non-blinking text insertion indicator in editable text fields.

The original issue requested seven features. I was able to find and implement features with clear, public, and general-purpose APIs.

The remaining features were not included for the following reasons:

- `Vehicle Motion Cues`: No public API found.
- `Auto-Play Message Effects`: No public API found. This is likely a feature exclusive to the Messages framework.
- `Limit Frame Rate`: No public API found.
- `Dim Flashing Lights`: There is a public API available [MADimFlashingLightsEnabled](https://developer.apple.com/documentation/mediaaccessibility/madimflashinglightsenabled()). It is intended for media playback frameworks. It's unclear if exposing this feature is necessary or actionable for a general-purpose application UI.

| Feature Name | API | Flutter AccessibilityFeatures |
| :-: | :-: | :-: |
| Vehicle Motion Cues | - | - |
| Dim Flashing Lights | [MADimFlashingLightsEnabled](https://developer.apple.com/documentation/mediaaccessibility/madimflashinglightsenabled()) |  - |
| Auto-Play Animated Images | [animatedImagesEnabled](https://developer.apple.com/documentation/accessibility/accessibilitysettings/animatedimagesenabled) | autoPlayAnimatedImages |
| Auto-Play Video Previews | [isVideoAutoplayEnabled](https://developer.apple.com/documentation/uikit/uiaccessibility/isvideoautoplayenabled) | autoPlayVideos |
| Auto-Play Message Effects | - | - |
| Prefer Non-Blinking Cursor | [prefersNonBlinkingTextInsertionIndicator](https://developer.apple.com/documentation/accessibility/accessibilitysettings/prefersnonblinkingtextinsertionindicator) | deterministicCursor |
| Limit Frame Rate | - | - |

The `AccessibilityFeatures.swift` was introduced to refactor all native iOS accessibility logic into a single, dedicated class. This change moves the responsibility of querying settings and observing notifications out of `FlutterViewController.mm`, which cleans up the controller, simplifies future maintenance, and makes logic unit-testable.

### Demo

https://github.com/user-attachments/assets/1b3e7a2f-03b4-4716-959e-dbeea938e4d2

<details closed><summary>Code sample</summary>

```dart

void main() {
  runApp(
    MaterialApp(
      home: Screen(),
    ),
  );
}

class Screen extends StatefulWidget {
  @override
  State<Screen> createState() => _ScreenState();
}

class _ScreenState extends State<Screen> with WidgetsBindingObserver {
  @override
  void initState() {
    super.initState();
    WidgetsBinding.instance.addObserver(this);
  }

  @override
  void didChangeAccessibilityFeatures() {
    super.didChangeAccessibilityFeatures();
    if (mounted) {
      setState(() {});
    }
  }

  @override
  void dispose() {
    WidgetsBinding.instance.removeObserver(this);
    super.dispose();
  }

  @override
    final accessibilityFeatures = View.of(
      context,
    ).platformDispatcher.accessibilityFeatures;

    return Scaffold(
      body: Center(
        child: Column(
          spacing: 10,
          children: [
            for (final feature in [
              'accessibleNavigation: ${accessibilityFeatures.accessibleNavigation}',
              'invertColors: ${accessibilityFeatures.invertColors}',
              'disableAnimations: ${accessibilityFeatures.disableAnimations}',
              'boldText: ${accessibilityFeatures.boldText}',
              'reduceMotion: ${accessibilityFeatures.reduceMotion}',
              'highContrast: ${accessibilityFeatures.highContrast}',
              'onOffSwitchLabels: ${accessibilityFeatures.onOffSwitchLabels}',
              'supportsAnnounce: ${accessibilityFeatures.supportsAnnounce}',
              'autoPlayAnimatedImages: ${accessibilityFeatures.autoPlayAnimatedImages}',
              'autoPlayVideos: ${accessibilityFeatures.autoPlayVideos}',
              'deterministicCursor: ${accessibilityFeatures.deterministicCursor}',
            ])
              Text(feature),
          ],
        ),
      ),
    );
  }
}
```

</details>

## Pre-launch Checklist

- [X] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterViewController.h
```diff
@@ -181,14 +181,6 @@ FLUTTER_DARWIN_EXPORT
  */
 - (id<FlutterPluginRegistry>)pluginRegistry;
 
-/**
- * A wrapper around UIAccessibilityIsVoiceOverRunning().
- *
- * As a C function, UIAccessibilityIsVoiceOverRunning() cannot be mocked in testing. Mock
- * this class method to testing features depends on UIAccessibilityIsVoiceOverRunning().
- */
-+ (BOOL)isUIAccessibilityIsVoiceOverRunning;
```
hellohuanlin, line 190: Why are we removing this API (which may cause breaking changes)?

ksokolovskyi: Thanks a lot for spotting this. I didn't realize that this would be a breaking change. I reverted the removal in the latest commit.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/AccessibilityFeatures.swift
```diff
@@ -0,0 +1,217 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
LouiseHsu, line 1: While this is quite nice and clean, is there a reason we are doing this pretty significant refactoring in this PR? it seems a bit out of scope, for this feature.

LouiseHsu (changes requested):

engine/src/flutter/shell/platform/darwin/ios/framework/Source/AccessibilityFeatures.swift
LouiseHsu: Also tagging @hannah-hyj who I seem to remember did some work refactoring accessibility flags for potential input

ksokolovskyi: Thanks for taking a look at this PR!

The new iOS 18 features require Apple's `AccessibilitySettings` API, which is Swift-only.

Since I had to add a Swift file to use this new API, I chose to consolidate all accessibility logic into it. The alternative would have been to split the logic across both `FlutterViewController.m` and this new Swift file, which felt less maintainable.

This approach keeps all related code in one place and makes the logic testable. Happy to discuss splitting it out if you still think it's best.

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -49,6 +49,9 @@ enum class AccessibilityFeatureFlag : int32_t {
   kHighContrast = 1 << 5,
   kOnOffSwitchLabels = 1 << 6,
   kNoAnnounce = 1 << 7,
+  kNoAutoPlayAnimatedImages = 1 << 8,
+  kNoAutoPlayVideoPreviews = 1 << 9,
+  kNonBlinkingCursor = 1 << 10,
```
chunhtai: we call this deterministicCursor in framework, should use this word instead

https://api.flutter.dev/flutter/widgets/EditableText/debugDeterministicCursor.html

engine/src/flutter/lib/ui/window.dart
```diff
@@ -985,6 +988,30 @@ class AccessibilityFeatures {
...
   bool get supportsAnnounce => _kNoAnnounceIndex & _index == 0;
 
+  /// Whether the platform allows auto-playing animated images.
+  ///
+  /// Only supported on iOS.
+  ///
+  /// Always returns `true` on other platforms.
+  // This index check is inverted (== 0 vs != 0) since most of the platforms
+  // don't have an option to disable animated images auto play.
+  bool get autoPlayAnimatedImages => _kNoAutoPlayAnimatedImagesIndex & _index == 0;
```
chunhtai, line 998: Does this also apply to gifs? If so we may need to wire this up to image widget. but that should probably be a separate PR cc @justinmc since you worked on that widget recently

engine/src/flutter/shell/platform/darwin/ios/framework/Source/AccessibilityFeatures.swift
```diff
@@ -0,0 +1,217 @@
...
+        return true
+    }
+
+    /// Notification name for changes to `Auto-Play Video Previews` setting.
+    @objc public static var videoPreviewsAutoPlayStatusDidChangeNotification: String {
+        return UIAccessibility.videoAutoplayStatusDidChangeNotification.rawValue
+    }
+
+    /// Whether `Auto-Play Video Previews` setting is enabled.
+    @objc public func isVideoPreviewsAutoPlayEnabled() -> Bool {
```
chunhtai: This feature flag doesn't convey the actually iOS settings, instead of just preview, it seems to disable autoplay for all videos

engine/src/flutter/lib/ui/window/platform_configuration.h
ksokolovskyi: TIL that there is such a functionality, thanks, I've renamed this a11y feature to `deterministicCursor`.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/AccessibilityFeatures.swift
ksokolovskyi: Indeed, looks like this setting disables the autoplay of all videos, not only previews. I've renamed this a11y feature as well.

chunhtai (approved):
LGTM for dart side

flutter-zl (approved):
LGTM from web ui side. Thanks for the change.

camsim99 (commented):
Android changes LGTM

engine/src/flutter/lib/ui/window.dart
justinmc: Looks like yes this is for gifs: https://developer.apple.com/documentation/accessibility/animated-images

I've created an issue: https://github.com/flutter/flutter/issues/180395

chunhtai:
google test failure doesn't seem related, @ksokolovskyi can you try rebase again?

LouiseHsu (approved):
ios parts lgtm!

PR #178116, opened 2025-11-06
[ Widget Preview ] Add embedded Widget Inspector support

### LINKED ISSUE #166423: Add first-class support for inspecting previews through the Widget Inspector

bkonyi:
This change allows for users to open an instance of the DevTools Widget Inspector within a web view embedded in the widget previewer. This will allow for developers to inspect their previews without requiring IDEs to create a special debug session for the widget previewer application.

DWDS is also rolled as part of this change to add DDS and DevTools support for the `web-server` device.

Fixes https://github.com/flutter/flutter/issues/166423

**Demo:**

https://github.com/user-attachments/assets/3e3a1098-0526-4c7f-8087-38fb84f28335


packages/flutter_tools/lib/src/commands/widget_preview.dart
```diff
@@ -456,8 +456,8 @@ final class WidgetPreviewStartCommand extends WidgetPreviewSubCommandBase with C
           webEnableHotReload: true,
         ),
         webEnableExposeUrl: false,
+        webEnableExpressionEvaluation: true,
         webRunHeadless: boolArg(kHeadless),
-        enableDevTools: boolArg(FlutterCommand.kEnableDevTools),
```
jyameo, line 484: Nit: While `DebuggingOptions` defaults `enableDevTools` to `true`, explicitly setting it here might be safer and clearer for others reading the code.

jyameo (approved):
LGTM!

packages/flutter_tools/lib/src/commands/widget_preview.dart
bkonyi: There's a lint that will prevent us from explicitly providing an argument if it matches the default for the parameter, so we can't do that.

jyameo: Ah! Got it. Thanks for the clarification

PR #178133, opened 2025-11-06
[web] Remove image codecs from canvaskit_chromium

### LINKED ISSUE #122331: [web] Remove image codecs from CanvasKit Chromium
**Update:** we had to put back image codecs into CanvasKit Chromium because there are still use cases that need the wasm codecs. See [PR](https://github.com/flutter/engine/pull/40501) for more info.

<hr>
As part of https://github.com/flutter/flutter/issues/113713, let's remove the image codecs from the CanvasKit Chromium variant.

This involves 2 things:

1. Adding the appropriate flags to the [canvaskit_chromium gn target](https://github.com/flutter/engine/blob/4429efe6c83d641141f1df0fbefbc091ed547cd5/third_party/canvaskit/BUILD.gn#L23-L25).
2. Adjusting the [`browserSupportsCanvaskitChromium`](https://github.com/flutter/engine/blob/832298bdd707ceabf6641d67d0550d231ac0acb7/lib/web_ui/lib/src/engine/browser_detection.dart#L271) boolean to check for availability of image codecs too.

mdebbar:
Fixes https://github.com/flutter/flutter/issues/122331

| Build | with image codecs (before) | without image codecs (after) | Difference % |
|-|-|-|-|
| canvaskit_chromium | 1.54 MB | 1.45 MB | $${\color{green}9\\%}$$ |
| canvaskit_webparagraph | 1.12 MB | 1.01 MB | $${\color{green}10\\%}$$ |


mdebbar:
> This makes uploaded bits smaller, right? By how much?

Yes, added the numbers to the PR description. It's ±100KB smaller.

> Have we confirmed on older Chrome builds? When was this fixed?

This PR is a no-op for apps (other than making CanvasKit's size smaller). The web engine has been using the browser `ImageDecoder` exclusively on Chrome for quite a while now. There was one remaining edge case that required Skia decoders, but that was addressed in https://github.com/flutter/engine/pull/53201 more than a year ago.

TL;DR: Flutter web apps today in Chrome are already not using the Skia decoders. This PR is removing the unused decoders to save size.

engine/src/flutter/lib/web_ui/test/canvaskit/canvaskit_api_test.dart
```diff
@@ -268,6 +268,11 @@ void _vertexModeTests() {
 
 void _imageTests() {
   test('MakeAnimatedImageFromEncoded makes a non-animated image', () {
+    if (configuration.canvasKitVariant == CanvasKitVariant.chromium) {
```
harryterkelsen, line 279: use `skip:` instead

harryterkelsen (dismissed):
LGTM with nit

engine/src/flutter/lib/web_ui/test/canvaskit/canvaskit_api_test.dart
mdebbar: That was the first thing I tried :)

It doesn't work because the `configuration` object is initialized *_after_* `skip` is evaluated. The `configuration` value is initialized in a `setUpAll` callback, so the only reliable place to get the correct value of `configuration.canvasKitVariant` is inside the `test` body.

yjbanov:
This is in our backlog to complete soon. Should we land this?

harryterkelsen (dismissed):
LGTM!

harryterkelsen (approved):
LGTM

PR #178139, opened 2025-11-06
[tool] clean up https cert configuration handling

kevmoo:
It's only valid if both the cert and the cert key are set.
Makes the code a lot cleaner, too.


packages/flutter_tools/lib/src/web/devfs_config.dart
```diff
@@ -189,28 +189,51 @@ WebDevServerConfig:
...
-  }
+  HttpsConfig copyWith({String? certPath, String? certKeyPath}) => HttpsConfig(
+    certPath: certPath ?? this.certPath,
+    certKeyPath: certKeyPath ?? this.certKeyPath,
+  );
 
-  final String? certPath;
-  final String? certKeyPath;
+  final String certPath;
+  final String certKeyPath;
```
ditman, line 236: I like that this is more strict now, and throwy in the static factory method!

packages/flutter_tools/lib/src/web/devfs_config.dart
```diff
@@ -189,28 +189,51 @@ WebDevServerConfig:
 /// Represents the [HttpsConfig] for the web dev server
 @immutable
 class HttpsConfig {
-  const HttpsConfig({this.certPath, this.certKeyPath});
-
+  const HttpsConfig({required this.certPath, required this.certKeyPath});
   factory HttpsConfig.fromYaml(YamlMap yaml) {
     final String? certPath = _validateType<String>(value: yaml[_kCertPath], fieldName: _kCertPath);
+    if (certPath == null) {
+      throw ArgumentError.value(yaml, 'yaml', '"$_kCertPath" must be defined');
```
ditman, line 196: These throws should have broken some test but it seems they didn't. Is there a test for this part of the tool that can assert that certPath and certKeyPath are being handled correctly?

(I guess `fromYaml` is used to load this from the config file only?)

ditman (approved):
LGTM, but cc/ @bkonyi as well!

packages/flutter_tools/lib/src/web/devfs_config.dart
kevmoo: I don't think there are tests here. Which is...sad...

kevmoo: tests added!

PR #178142, opened 2025-11-06
Reland "Update all uses of mutable SkPath methods to use SkPathBuilder"

jason-simmons:
This is a reland of https://github.com/flutter/flutter/pull/177738 with a fix to the `Op` function in `tools/path_ops`.

To match the previous behavior, `Op` will use `SkPathBuilder::operator=` to copy the result path so that attributes like the fill type will be copied.

### DIFF
--- a/engine/src/flutter/tools/path_ops/path_ops.cc
+++ b/engine/src/flutter/tools/path_ops/path_ops.cc
@@ -4,26 +4,26 @@
 
 #include "path_ops.h"
 
+#include "third_party/skia/include/core/SkPath.h"
+
 namespace flutter {
-SkPath* CreatePath(SkPathFillType fill_type) {
-  auto* path = new SkPath();
-  path->setFillType(fill_type);
-  return path;
+SkPathBuilder* CreatePath(SkPathFillType fill_type) {
+  return new SkPathBuilder(fill_type);
 }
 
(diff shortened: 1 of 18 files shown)


engine/src/flutter/tools/path_ops/path_ops.cc
```diff
@@ -33,25 +33,30 @@ void CubicTo(SkPath* path,
...
   path->reset();
 }
 
-void Op(SkPath* one, SkPath* two, SkPathOp op) {
-  Op(*one, *two, op, one);
+void Op(SkPathBuilder* one, SkPathBuilder* two, SkPathOp op) {
+  SkPath p1 = one->snapshot();
+  SkPath p2 = two->snapshot();
+  if (std::optional<SkPath> result = Op(p1, p2, op)) {
+    *one = result.value();
```
kjlubick, line 48: nit: use the explicit `SkPathBuilder` constructor here so we don't hide the change.

kjlubick (approved):

engine/src/flutter/tools/path_ops/path_ops.cc
jason-simmons: `SkPathBuilder` has an implementation of `operator=` that takes an `SkPath` - this is not implicitly constructing an `SkPathBuilder`.

PR #178205, opened 2025-11-08
fix #178204: update textediting controller notifier on user tap on field

### LINKED ISSUE #178204: TextEditingController listener is triggered on tap
### Steps to reproduce

While implementing a textfield search feature I noticed that the textEditingController listener is triggered when the user only taps on the field I tested this on web and android it's the same result always

Investigating the source of this behavior it's related to 
```
@override
  set value(TextEditingValue newValue) {
    assert(
      !newValue.composing.isValid || newValue.isComposingRangeValid,
      'New TextEditingValue $newValue has an invalid non-empty composing range '
      '${newValue.composing}. It is recommended to use a valid composing range, '
      'even for readonly text fields.',
    );
    super.value = newValue;
  }
```
and also 
```
 @override
  void userUpdateTextEditingValue(TextEditingValue value, SelectionChangedCause? cause) {
    // Compare the current TextEditingValue with the pre-format new
    // TextEditingValue value, in case the formatter would reject the change.
    final bool shouldShowCaret = widget.readOnly
        ? _value.selection != value.selection
        : _value != value;
    if (shouldShowCaret) {
      _scheduleShowCaretOnScreen(withAnimation: true);
    }

    // Even if the value doesn't change, it may be necessary to focus and build
    // the selection overlay. For example, this happens when right clicking an
    // unfocused field that previously had a selection in the same spot.
    if (value == textEditingValue) {
      if (!widget.focusNode.hasFocus) {
        _flagInternalFocus();
        widget.focusNode.requestFocus();
        _selectionOverlay ??= _createSelectionOverlay();
      }
      return;
    }
    _formatAndSetValue(value, cause, userInteraction: true);
  }
```
the selection is updated even if the `cause` is "tap"

### Expected results

Expected result could be just a focus on the field without a trigger of the Listener callback which could hold other logic or data fetch.

### Actual results

The actual results: the `addListener((){ // code })` the void callback is triggered upon user tap on the field on Android and Web

### Code sample

Implementation example
```
TextField(
    controller: TextEditingController()..addListener((){
      print('Text field updated');
    }
 ),
)
```

### Screenshots or Video

_No response_

### Logs

_No response_

### Flutter Doctor output

Flutter Doctor
```
[✓] Flutter (Channel main, 3.38.0-1.0.pre-60, on macOS 15.0.1 24A348 darwin-arm64, locale en-FR)
[!] Xcode - develop for iOS and macOS
    ✗ Xcode installation is incomplete; a full installation is necessary for iOS and macOS development.
      Download at: https://developer.apple.com/xcode/
      Or install Xcode via the App Store.
      Once installed, run:
        sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer
        sudo xcodebuild -runFirstLaunch
    ! CocoaPods 1.15.2 out of date (1.16.2 is recommended).
[✓] Network resources

```

koukibadr:
This PR updates the text editing controller listener trigger, current listner is invoked on user tap on field

Screen recording:
https://github.com/user-attachments/assets/c0327343-f2f3-4266-afc9-83322b90c861

**Updates**
- Update the `TextEditingController` class `set value` function to test if the current text value is different from the new text value, if it's the case the listner is triggered otherwise there's no notifyListener trigger.

Screen recording after updates
https://github.com/user-attachments/assets/f6aded4d-7e66-4531-abcd-1d4d53024541

related issue [issues/178204](https://github.com/flutter/flutter/issues/178204)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/rendering/editable.dart
+++ b/packages/flutter/lib/src/rendering/editable.dart
@@ -754,10 +754,12 @@ class RenderEditable extends RenderBox
         extentOffset: math.min(nextSelection.extentOffset, textLength),
       );
     }
-    _setTextEditingValue(
-      textSelectionDelegate.textEditingValue.copyWith(selection: nextSelection),
-      cause,
-    );
+    if (nextSelection.extentOffset > nextSelection.baseOffset) {
+      _setTextEditingValue(
+        textSelectionDelegate.textEditingValue.copyWith(selection: nextSelection),
+        cause,
+      );
+    }
   }
 
   @override
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -4290,7 +4290,7 @@ class EditableTextState extends State<EditableText>
     // text of [EditableText] is updated at the same time as the selection is
     // changed by a gesture event.
     final String text = widget.controller.value.text;
-    if (text.length < selection.end || text.length < selection.start) {
+    if (text.length < selection.end || text.length < selection.start || selection.end <= selection.start) {
       return;
     }
 


loic-sharma:
@koukibadr Thanks for the contribution! It looks like this pull request is missing tests - would you be able to add some?

koukibadr:
> @koukibadr Thanks for the contribution! It looks like this pull request is missing tests - would you be able to add some?

@loic-sharma yes I'll update the PR

loic-sharma:
Note from triage, there are active discussions on the issue: https://github.com/flutter/flutter/issues/178204

koukibadr:
Hello @Piinks, there's active discussion about this behavior I'll update the PR following the discussion result

PR #178224, opened 2025-11-10
[web] Enable semantics via query parameter for automation testing

### LINKED ISSUE #177670: [Flutter web-semantics] Feature request: Enable Accessibility button upper left 1 pixel instead of offscreen 1 pixel
### Use case

Automation tools such as Maestro and probably Appium seem to not be able to tap on
`      document.querySelector('flt-semantics-placeholder[aria-label="Enable accessibility"]')
                .click();` because it is at absolute positivion -1 left and -1 top of the screen (offscreen).

Maestro does appear to be able to tap on an invisible button with no text if it is at 1,1.

I am not sure if i have bandwidth to PR this but i wanted to get the discussion going.
See: https://docs.flutter.dev/ui/accessibility/web-accessibility#invisible-button

Can anyone think of a reason that the ('flt-semantics-placeholder[aria-label="Enable accessibility"]' has to be offscreen rather than "just" onepixel, invisible, transparent etc?

### Proposal

Change just 2 characters in the code to make the button be "+" 1 left and "+" one top insteda of "-" (or remove the negative)

flutter-zl:
Add support for enabling semantics via ?flutter-semantics query parameter. This allows automation testing tools like Maestro and Appium to enable semantics without modifying the application code or requiring special builds.

Usage: https://example.com/?flutter-semantics

After the change:
https://semantic-1107.web.app/
https://semantic-1107.web.app/?flutter-semantics

Fixes #177670

### Test WITHOUT Query Parameter

Navigate to: https://semantic-1107.web.app/

**In Chrome DevTools Console:**
```
// Check semantic elements
document.querySelectorAll('flt-semantics').length;
// Should return 0 - semantics NOT enabled
```
### Test WITH Query Parameter(disabling)

Navigate to:  https://semantic-1107.web.app/?flutter-semantics=false

**In Chrome DevTools Console:**
```
// Check semantic elements
document.querySelectorAll('flt-semantics').length;
// Should return 0 - semantics NOT enabled
```

### Test WITH Query Parameter(enabling)

Navigate to:  https://semantic-1107.web.app/?flutter-semantics or https://semantic-1107.web.app/?flutter-semantics=true

**In Chrome DevTools Console:**
// Check semantic elements
document.querySelectorAll('flt-semantics').length;
// Should return > 0 - semantics ARE enabled

// List all semantic buttons
document.querySelectorAll('[role="button"][aria-label]');
// Should show: "Go to Profile", "Go to Settings", "Go to Counter"

### Test Navigation Persistence

**With query parameter active (`?flutter-semantics`):**

1. Click "Profile" button
2. Check DevTools Console:
```
document.querySelectorAll('flt-semantics').length;
// Should still be > 0 - semantics PERSIST!
```

3. Click "Settings" button
4. Check again - semantics still present

5. Click "Back to Home"
6. Check again - semantics still present

7. Click "Counter" button
8. Increment counter a few times
9. Check semantic label:
```
document.querySelector('[aria-label^="Counter:"]').getAttribute('aria-label');
// Should show "Counter: [number]"
```

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
@@ -58,7 +58,16 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   /// This allows automation testing tools to enable semantics without modifying the app.
   /// Usage: https://example.com/?flutter-semantics
   void _checkQueryParameterForSemantics() {
-    if (Uri.base.queryParameters.containsKey('flutter-semantics')) {
+    checkUriForSemanticsParameter(Uri.base);
+  }
+
+  /// Checks if the given URI contains the flutter-semantics query parameter.
+  /// If present, enables semantics.
+  ///
+  /// This method is separated for testing purposes.
+  @visibleForTesting
+  void checkUriForSemanticsParameter(Uri uri) {
+    if (uri.queryParameters.containsKey('flutter-semantics')) {
       EngineSemantics.instance.semanticsEnabled = true;
     }
   }
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
@@ -48,6 +48,19 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     /// and ensures proper focus is set, enabling Flutter's focus restoration to work correctly
     /// when users navigate between pages.
     _addNavigationFocusHandler();
+
+    /// Check for query parameter to enable semantics programmatically after initialization.
+    /// This must be done asynchronously to avoid circular dependency during construction.
+    scheduleMicrotask(_checkQueryParameterForSemantics);
+  }
+
+  /// Check for query parameter to enable semantics programmatically.
+  /// This allows automation testing tools to enable semantics without modifying the app.
+  /// Usage: https://example.com/?flutter-semantics
+  void _checkQueryParameterForSemantics() {
+    if (Uri.base.queryParameters.containsKey('flutter-semantics')) {
+      EngineSemantics.instance.semanticsEnabled = true;
+    }
   }
 
   late StreamSubscription<int> _onViewDisposedListener;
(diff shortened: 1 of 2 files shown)


engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -48,6 +48,19 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
+    /// Check for query parameter to enable semantics programmatically after initialization.
+    /// This must be done asynchronously to avoid circular dependency during construction.
+    scheduleMicrotask(_checkQueryParameterForSemantics);
+  }
+
+  /// Check for query parameter to enable semantics programmatically.
+  /// This allows automation testing tools to enable semantics without modifying the app.
+  /// Usage: https://example.com/?flutter-semantics
+  void _checkQueryParameterForSemantics() {
+    if (Uri.base.queryParameters.containsKey('flutter-semantics')) {
```
ecolab-neil: @flutter-zl  is it "worth it" to check the value of this query parameter and if it is  value.toLowercase == "false"  then do not enable?  Or does that open up too big of a can of "worms". Also I'm "guessing" if it is already enabled in the flutter app and it's set to true "again" this way it causes no issues right?

flutter-zl: 1. I'd keep the presence-check only for simplicity. This is for enabling in automation tests, not disabling. Default is already off.

2. [The setter has a built-in guard](https://github.com/flutter/flutter/blob/master/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart#L2705).
`
set semanticsEnabled(bool value) {
  if (value == _semanticsEnabled) {
    return; // Early return - no work done if already set
  }
  // ... actual work only happens if value changes
}
`
I already verified in my local demo app. It works fine.

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -48,6 +48,19 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
+      EngineSemantics.instance.semanticsEnabled = true;
+    }
   }
```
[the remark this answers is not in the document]
flutter-zl: Added test cases.

ecolab-neil (commented):
@flutter-zl  i don't have access to approve, and i definitely would love to have it merged soon, like i said i just wonder if you would consider changing the code so that if the user puts `?flutter-semantics=false `that it would be false - since querystring is key value pairs this could be confusing checking *only* for the key

but i'd rather "skip" my suggestion than delay it much :)

Issue #178060, opened 2025-11-05
Tree Red: Missing dependencies on CI Ubuntu 22


jtmcdole:
In order to address the TASK_BOT died errors, we downgraded from Ubuntu 24 to Ubuntu 22. This rolled out slowly on Nov 4th and uncovered some test failures on the dashboard. These failures appear around many of the Android tests which do not include ninja / cmake / clang in their dependencies list of ci.yaml.

Updating multiple locations in: #178054 

```
          {"dependency": "clang", "version": "git_revision:5d5aba78dbbee75508f01bcaa69aedb2ab79065a"},
          {"dependency": "cmake", "version": "build_id:8787856497187628321"},
          {"dependency": "ninja", "version": "version:1.9.0"}
```

This has the tools installed in the cache on the CI machines and appear to fix the issue.

Outstanding issue: [Linux web_skwasm_tests_2](https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20web_skwasm_tests_2) appears to be failing with a chrome conneciton.

jtmcdole:
fyi @mdebbar for skwasm test failure

PR #178054, opened 2025-11-05
Add ninja / cmake deps to failing tests

jtmcdole:
Ubuntu 22 rollout triggered some failing tests due to missing deps.

PR #178243, opened 2025-11-10
[CP-Stable] Cherry-pick ninja+cmake deps

jtmcdole:
Pulls in #178054 and #178079

Please note there is one "bringup: true" in this change:
    bringup: true # Failing: https://github.com/flutter/flutter/issues/178060


vashworth (approved):

jtmcdole:
Given that 3.38 is upon us, I don't think we need this anymore.

PR #178253, opened 2025-11-10
Dynamically set MinimumOSVersion in App.framework

okorohelijah:
Currently, the MinimumOSVersion in App.framework/Info.plist is sourced from a static template file created during flutter create. This value can become out of sync with the -miphoneos-version-min flag used to compile the App.framework binary, especially after a Flutter SDK upgrade. This mismatch can lead to App Store submission warnings or rejections.

This change fixes the issue by making the build process the single source of truth for the MinimumOSVersion.

Fixes #176313 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/test/general.shard/build_system/targets/ios_test.dart
```diff
@@ -233,7 +233,18 @@ void main() {
...
+        FakeCommand(
+          command: <String>[
+            'plutil',
+            '-replace',
+            'MinimumOSVersion',
+            '-string',
+            '13.0',
+            infoPlist.path,
+          ],
+        ),
```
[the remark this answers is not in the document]
okorohelijah: done.

packages/flutter_tools/lib/src/build_system/targets/ios.dart
```diff
@@ -828,6 +831,27 @@ class ReleaseIosApplicationBundle extends _IosAssetBundleWithDSYM {
   }
 }
 
+/// Update the MinimumOSVersion key in the given Info.plist file.
+Future<void> _updateMinimumOSVersion(File infoPlist, Environment environment) async {
+  final minimumOSVersion = FlutterDarwinPlatform.ios.deploymentTarget().toString();
+  final plutilArgs = <String>[
+    'plutil',
+    '-replace',
```
vashworth, line 837: Does replace work if it's not already in the Info.plist?

packages/flutter_tools/lib/src/build_system/targets/ios.dart
```diff
@@ -828,6 +831,27 @@ class ReleaseIosApplicationBundle extends _IosAssetBundleWithDSYM {
...
+    minimumOSVersion,
+    infoPlist.path,
+  ];
+  final ProcessResult result = await environment.processManager.run(plutilArgs);
+  if (result.exitCode != 0) {
+    environment.logger.printError(
+      'Failed to update MinimumOSVersion in ${infoPlist.path}:\n'
+      '${result.stdout}\n---\n${result.stderr}',
+    );
+    throwToolExit('Failed to update MinimumOSVersion in ${infoPlist.path}');
```
vashworth: Instead of throwing an error, perhaps we can give a warning. I'm concerned that if this fails for some unknown reason, they won't be able to build their app and it doesn't give any valuable guidance on what to do.

Perhaps use `printXcodeWarning` and say something "Failed to update MinimumOSVersion in ${infoPlist.path}. This may cause AppStore validation failures. Please file an issue at https://github.com/flutter/flutter/issues/new."

vashworth (commented):
Like I mentioned offline, you'll need to do a migration. Looks like we already have one that migrates the App framework Info.plist, so we likely need to change this from updating the MinimumOSVersion to instead removing it:
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/ios/migrations/ios_deployment_target_migration.dart


You should add an integration test that verifies the contents of the `App.framework/Info.plist`. I think you add to
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/host_cross_arch.shard/ios_content_validation_test.dart

packages/flutter_tools/lib/src/build_system/targets/ios.dart
okorohelijah: yes, if it isnt there, it adds it

packages/flutter_tools/lib/src/build_system/targets/ios.dart
```diff
@@ -826,6 +829,29 @@ class ReleaseIosApplicationBundle extends _IosAssetBundleWithDSYM {
...
+  ];
+  final ProcessResult result = await environment.processManager.run(plutilArgs);
+  if (result.exitCode != 0) {
+    environment.logger.printError(
+      'Failed to update MinimumOSVersion in ${infoPlist.path}:\n'
+      '${result.stdout}\n---\n${result.stderr}',
+    );
+    printXcodeWarning(
+      'Failed to update MinimumOSVersion in ${infoPlist.path}. This may cause AppStore validation failures. Please file an issue at https://github.com/flutter/flutter/issues/new/choose',
+    );
```
vashworth: Why are both a print error and warning needed? Does the stdout/stderr not show in verbose mode unless you print to explicitly?

packages/flutter_tools/lib/src/ios/migrations/ios_deployment_target_migration.dart
```diff
@@ -58,10 +58,7 @@ class IOSDeploymentTargetMigration extends ProjectMigrator {
   <key>MinimumOSVersion</key>
   <string>12.0</string>
 ''';
-    const minimumOSVersionReplacement = '''
-  <key>MinimumOSVersion</key>
-  <string>13.0</string>
-''';
+    const minimumOSVersionReplacement = '';
```
vashworth, line 65: You'll also want to replace ones that are already `13.0`

packages/flutter_tools/lib/src/build_system/targets/ios.dart
vashworth: Also, I think it shows in both, but can you test to make sure the warning shows when run from both the Flutter CLI (in non-verbose) and when run from Xcode

packages/flutter_tools/test/host_cross_arch.shard/ios_content_validation_test.dart
```diff
@@ -172,6 +173,55 @@ void main() {
...
+              flutterBin,
+              ...getLocalEngineArguments(),
+              'build',
+              'ios',
+              '--verbose',
+              '--no-codesign',
+              '--${buildMode.cliName}',
+            ], workingDirectory: projectRoot);
+
+            templateInfoPlist.writeAsStringSync(originalContents);
```
vashworth: I don't think you need to write to the Info.plist or rebuild here. 

If you look at the structure of this test, in the `iOS app validation` group `setUpAll`, it creates an app (in which case the Info.plist should not have `MinimumOSVersion` within it):

https://github.com/flutter/flutter/blob/77d99d6ee290a39a8f2e6574dec5dc329bca30b4/packages/flutter_tools/test/host_cross_arch.shard/ios_content_validation_test.dart#L27-L42

You should add a test to validate that before it builds in the next group.

Then in the `build in ${buildMode.cliName} mode` group `setUpAll`, it builds the app:
 
https://github.com/flutter/flutter/blob/77d99d6ee290a39a8f2e6574dec5dc329bca30b4/packages/flutter_tools/test/host_cross_arch.shard/ios_content_validation_test.dart#L94-L105
 
So you shouldn't need to build, just validate the embedded Info.plist now has the correct `MinimumOSVersion`.

packages/flutter_tools/lib/src/build_system/targets/ios.dart
okorohelijah: yes it shows both but I had added both because i thought they could serve different purposes - the `printError` mainly for debugging(-v) and `the printXcodewarning` for Developer action

packages/flutter_tools/test/host_cross_arch.shard/ios_content_validation_test.dart
```diff
@@ -172,6 +173,31 @@ void main() {
...
+                .toString();
+
+            final ProcessResult result = processManager.runSync(<String>[
+              'plutil',
+              '-p',
+              appFrameworkInfoPlist.path,
+            ]);
+
+            expect(result.exitCode, 0);
+            expect(result.stdout, contains('"MinimumOSVersion" => "$expectedMinimumOSVersion"'));
```
vashworth: Why do you need to use `plutil`, why not just check just the contents using `readAsStringSync`?

packages/flutter_tools/lib/src/ios/migrations/ios_deployment_target_migration.dart
vashworth, line 1: Looks like there's not a pre-existing unit test for this class. One should be added under /Users/vashworth/Development/flutter/packages/flutter_tools/test/general.shard/migrations/

vashworth (approved):
LGTM!

PR #178258, opened 2025-11-10
Use glyph boundaries for horizontal character traversal

### LINKED ISSUE #34610: Mixing RTL and LTR text bugs
We are seeing strange behavior when mixing RTL and LTR text.

Notably:

- Type arabic, Type chinese, Type arabic, switch to chinese keyboard, delete. The arabic will delete all at once.
- Emojis work strangely when mixing with RTL and LTR.
- Caret positioning is not consistent.

There are a few other minor issues that I am not clear on how to describe.

### LINKED ISSUE #78660: Arrow keys in RTL move the wrong way
The left/right arrow keys should move the caret left/right respectively, regardless of the text direction.  In Flutter they move in the wrong direction in RTL text.

The methods that handle this in RenderEditable should be updated to perform the correct behavior without being renamed.  ~~So for example, the method [_moveGivenSelectionLeft](https://github.com/flutter/flutter/blob/8c44abc009a333b0494598c0cfda0fbc07358234/packages/flutter/lib/src/rendering/editable.dart#L735) actually moves the selection upstream right now, not necessarily left.  It should be modified to move the selection left like it says.~~ This is out of date, see below.

## Steps to Reproduce

1. Run the app below, which is just a TextField with some Arabic text in it, on a device with arrow keys.
2. Try to navigate with the left and right arrow keys.

**Expected results:** The cursor moves left with the left key and right with the right key.  Tested on TextEdit on a Mac after changing my keyboard to Arabic.

**Actual results:** The cursor movement direction is reversed.

<details>
  <summary>Code</summary>

```dart
 
void main() {
  runApp(MyApp());
}
 
  final TextEditingController controller = TextEditingController(text: 'شسيب');

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: TextField(
            controller: controller,
          ),
        ),
      ),
    );
  }
}
```

</details>

### LINKED ISSUE #120049: RTL textField Cursor movement has a problem with the last position and Emojies and text selection
## Steps to Reproduce

1. Execute `flutter run` on the code sample
2. type some things in text field like Screenshots that are attached 
3. try to move the cursor on different words and goes to the last words and move the cursor to the last position

**Expected results:** we want the cursor to move to the last position without any problems 

**Actual results:** As indicated in the screenshots the cursor stuck before the last position and can't move it to last position of words

for an additional report, this problem occurs when using a Persian or Arabic keyboard for typing and we use a special font for our text this makes other bugs that this issue repeats for each line of text for the normal situation this just occurs on the last line of the paragraph. also if using Emojies this problem repeats again or for text selection, we can select the last character of words.

another issue that should be mentioned here is the cursor movement problem for Persian keyboards when characters stick together this makes a bug that the issue stuck in the wrong position and when u typing something these words display in another position that does not match with your selected position

<details>
<summary>

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
    );
  }
}

  final String title;

  @override
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });
  }

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          children: <Widget>[
            const Text(
            ),
            Text(
              '$_counter',
            ),
            Padding(
              padding: const EdgeInsets.all(28.0),
              child: TextField(textDirection: TextDirection.rtl,maxLines: null,),
            )
          ],
        ),
      ),
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ),
    );
  }
}

</summary>

</details>

<details>
  <summary>Logs</summary>

```
```

[√] Flutter (Channel stable, 3.7.1, on Microsoft Windows [Version 10.0.19044.2486], locale en-US)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
[√] Visual Studio - develop for Windows (Visual Studio Professional 2019 16.11.22)  
[√] Android Studio (version 2021.1)
[√] IntelliJ IDEA Community Edition (version 2021.3)
[√] VS Code (version 1.73.1)
[√] Connected device (4 available)
[!] HTTP Host Availability
    X HTTP host "https://cocoapods.org/" is not reachable. Reason: An error occurred while checking the HTTP host: The connection terminated during
      handshake

```
```

</details>

image
The image about the cursor stuck behind the last Emojies

image
this image is about moving the cursor to the middle of the words and the cursor stuck in this position but if u typing, type in another position

image
this image is about the problem with the last char of a word that doesn't let you move the cursor behind it or select it

https://user-images.githubusercontent.com/28021064/216845203-1eccc103-6252-4e1c-9434-885992e0ccbc.mp4
the video that shows the problem with the cursor on Persian words

tgucio:
This PR implements horizontal character traversal (using arrow keys) in text fields and paragraphs using actual glyph information from text layout, as opposed to "guessing" character boundaries from the underlying text. This allows for correct horizontal caret traversal of RTL text as well as correctly positioning the caret between characters in complex scripts (e.g. Indic) that rely on text shaping.

Before:
before

After:
after

Fixes #34610
Fixes #78660
Maybe fixes #120049

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -5275,22 +5275,21 @@ class EditableTextState extends State<EditableText>
 
     int offset = extent.offset, nextOffset;
     // Correct for affinity.
-    if (extent.affinity == TextAffinity.upstream && extent.offset >= 0 && extent.offset < _value.text.length) {
+    if (extent.affinity == TextAffinity.upstream &&
+        extent.offset >= 0 &&
+        extent.offset < _value.text.length) {
       offset--;
     }
-
-    // TODO(tgucio): test newlines at RTL / LTR (skip one glyph?) -> FAIL
-    // TODO(tgucio): handle line ends after a single LTR/RTL change? [test]
 
     TextAffinity affinity = TextAffinity.downstream;
     TextRange boundary = textBoundary.getTextBoundaryAt(offset);
     // Use current boundary ends to find the next boundary.
     if (boundary.isValid) {
       nextOffset = switch ((forward, boundary.isNormalized)) {
-        (true, true) => boundary.end,        // LTR, forward
-        (true, false) => boundary.end - 1,   // RTL, forward
+        (true, true) => boundary.end, // LTR, forward
+        (true, false) => boundary.end - 1, // RTL, forward
         (false, true) => boundary.start - 1, // LTR, reverse
-        (false, false) => boundary.start,    // RTL, reverse
+        (false, false) => boundary.start, // RTL, reverse
       };
     } else {
       // Handle starting outside of text bounds.
@@ -5323,12 +5322,12 @@ class EditableTextState extends State<EditableText>
           offset = math.max(0, nextOffset);
           nextOffset = forward ? boundary.start : boundary.end - 1;
           nextBoundary = textBoundary.getTextBoundaryAt(nextOffset);
-        } while (nextOffset < _value.text.length
-            && nextOffset >= 0
-            && nextBoundary.isValid
-            && nextBoundary.isNormalized == boundary.isNormalized);
-          offset = math.max(0, boundary.start);
-          affinity = forward ? TextAffinity.upstream : TextAffinity.downstream;
+        } while (nextOffset < _value.text.length &&
+            nextOffset >= 0 &&
+            nextBoundary.isValid &&
+            nextBoundary.isNormalized == boundary.isNormalized);
+        offset = math.max(0, boundary.start);
+        affinity = forward ? TextAffinity.upstream : TextAffinity.downstream;
       } else {
         // RTL->LTR: start skipping at the current boundary.
         nextBoundary = boundary;
@@ -5337,12 +5336,12 @@ class EditableTextState extends State<EditableText>
           offset = math.max(0, nextOffset);
           nextOffset = forward ? boundary.start : boundary.end - 1;
           nextBoundary = textBoundary.getTextBoundaryAt(nextOffset);
-        } while (nextOffset < _value.text.length
-            && nextOffset >= 0
-            && nextBoundary.isValid
-            && nextBoundary.isNormalized == boundary.isNormalized);
-          offset = math.max(0, nextBoundary.end);
-          affinity = forward ? TextAffinity.downstream : TextAffinity.upstream;
+        } while (nextOffset < _value.text.length &&
+            nextOffset >= 0 &&
+            nextBoundary.isValid &&
+            nextBoundary.isNormalized == boundary.isNormalized);
+        offset = math.max(0, nextBoundary.end);
+        affinity = forward ? TextAffinity.downstream : TextAffinity.upstream;
       }
     }
 
@@ -5391,8 +5390,9 @@ class EditableTextState extends State<EditableText>
 
   // --------------------------- Text Editing Actions ---------------------------
 
-  TextBoundary _characterBoundary() =>
-      widget.obscureText ? _CodePointBoundary(_value.text) : renderEditable.glyphBoundaries.moveByGlyphBoundary;
+  TextBoundary _characterBoundary() => widget.obscureText
+      ? _CodePointBoundary(_value.text)
+      : renderEditable.glyphBoundaries.moveByGlyphBoundary;
   TextBoundary _nextWordBoundary() =>
       widget.obscureText ? _documentBoundary() : renderEditable.wordBoundaries.moveByWordBoundary;
   TextBoundary _linebreak() =>
@@ -5644,7 +5644,7 @@ class EditableTextState extends State<EditableText>
 
     // Delete
     DeleteCharacterIntent: _makeOverridable(
-      _DeleteTextAction<DeleteCharacterIntent>(this, _characterBoundary, _moveBeyondGlyphBoundary),
+      _DeleteTextAction<DeleteCharacterIntent>(this, _characterBoundary, _moveBeyondTextBoundary),
     ),
     DeleteToNextWordBoundaryIntent: _makeOverridable(
       _DeleteTextAction<DeleteToNextWordBoundaryIntent>(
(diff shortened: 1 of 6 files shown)


tgucio:
@LongCatIsLooong the implementation in this PR is actually based on `ui.Paragraph.getGlyphInfoAt` which exposes glyph direction. The logic to get boundaries is implemented in `GlyphBoundary` as an extension of `TextBoundary` to keep the current design in text_painter.dart. The idea is to use actual rendered glyph information instead of relying on the Characters package to "guess" where the glyphs should start/end.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -5237,6 +5237,117 @@ class EditableTextState extends State<EditableText>
...
+    int offset = extent.offset, nextOffset;
+    // Correct for affinity.
+    if (extent.affinity == TextAffinity.upstream &&
+        extent.offset >= 0 &&
+        extent.offset < _value.text.length) {
+      offset--;
+    }
+
+    TextAffinity affinity = TextAffinity.downstream;
+    TextRange boundary = textBoundary.getTextBoundaryAt(offset);
```
justinmc, line 5285: Does TextBoundary.getTextBoundaryAt possibly require iterating the entire string? I'm thinking about what the performance implications of this PR could be. I assume it will be negligible in most cases, but we should think it through.

packages/flutter/test/widgets/editable_text_test.dart
justinmc, line 1: We should probably include a high level test that actually puts a bidirectional string into an EditableText and traverses it with arrow keys.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -5550,7 +5662,7 @@ class EditableTextState extends State<EditableText>
       _UpdateTextSelectionAction<ExtendSelectionByCharacterIntent>(
         this,
         _characterBoundary,
-        _moveBeyondTextBoundary,
+        _moveBeyondGlyphBoundary,
```
justinmc, line 5665: Are there any other places that would benefit from _moveBeyondGlyphBoundary?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -5237,6 +5237,117 @@ class EditableTextState extends State<EditableText>
...
   }
 
+  // Returns the closest visual boundary location to `extent` but not including `extent`
+  // itself (unless already at the start/end of the text), in the direction
+  // specified by `forward`.
+  TextPosition _moveBeyondGlyphBoundary(
+    TextPosition extent,
+    bool forward,
+    TextBoundary textBoundary,
+  ) {
```
justinmc, line 5247: Nit: If you accept _value as a parameter, could this be static? Just thinking it might make this complex method easier to understand if so.

justinmc (commented):
@tgucio I really appreciate this ambitious PR. I left some quick thoughts but I'm interested in whether @LongCatIsLooong thinks this is feasible or not.

LongCatIsLooong:
Oh I only realized this is implementing visual order traversal after seeing the before and after gifs (I think @Renzo-Olivares or @loic-sharma might be interested). But still, I don't think SkParagraph currently exposes enough information for us to implement visual order traversal properly. Here's an example that I think we won't be able to handle without the bidi region information from SkParagraph (or maybe just ICU but SkParagraph needs that info for layout and already maintains the info for each character IIRC):
`aBcDe`. This can be render as `aBcDe` or `aDcBe` based on whether `BcD` are in the same bidi region. So we don't know if the next character visually after `a` is `B` or `D` without the bidi region info. Technically you can still get which character is visually after `a` using the hit-test method (getClosestGlyphAtPosition or something with a similar name) but that's going to be very slow and painful.

packages/flutter/lib/src/widgets/editable_text.dart
tgucio: That's correct, in the worst case (single text direction change), this would traverse the string towards either end. But I think the current `WordBoundary` implementation does so too.

packages/flutter/test/widgets/editable_text_test.dart
tgucio: Agree, this needs a couple of new tests here and there.

packages/flutter/lib/src/widgets/editable_text.dart
tgucio: Good question there. Perhaps Paragraph for changing text selection?

Renzo-Olivares:
Hi @tgucio, thank you for your contribution here. I've been meaning to review this pull request recently since it aligns with the RTL work I wanted to work on this quarter but I haven't had the bandwidth yet. Will try to get to this soon, appreciate the patience!

Renzo-Olivares:
Still taking a look at this PR but I had a few thoughts.

From testing on macOS web (chrome/safari/firefox), chrome seems to prefer logical order traversal for text selection with keyboard shortcuts. Safari and Firefox surprisingly differs in that it uses visual order for collapsed selection but logical order for uncollapsed selection with keyboard shortcuts.

On native, SwiftUI on macOS, visual order traversal was used, except with uncollapsed selections the selection could not span a mixed region.

On windows web (chrome)/and native (Notepad/Word) it also seems to prefer logic order traversal.

Still need to try on a linux machine (my guess is logical).

Because this behavior varies per platform i'm leaning towards guarding this with a platform check so only macOS native and Safari get this behavior.

LongCatIsLooong:
From text input triage: hello @tgucio looks like @Renzo-Olivares has left some comments.

Piinks:
Hey @tgucio would you like to continue working on this change?

PR #178309, opened 2025-11-11
[web] API to customize semantics placeholder message

### LINKED ISSUE #178172: [web] No way to set the placeholderMessage value
Internal Google issue: b/458487868

The web engine has a mutable top-level variable that specifies the label for the accessibility placeholder. It even documents that it can be updated (e.g. to internationalize it). However, it is declared in an internal engine library not exposed to the user.

Code: https://github.com/flutter/flutter/blob/3daaf5966eda062bf5a38a33c23a959cd9869803/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics_helper.dart#L27

We should move it to `dart:ui_web`, which is public, and can be actually set by the app developer.

## Reproduction code

There's no way to write code to reproduce this. There simply is no public setter to call.

mdebbar:
To customize the accessibility placeholder message:

```dart
import 'dart:ui_web' as ui_web;

void main() {
  ui_web.accessibilityPlaceholderMessage = 'My Custom Accessibility Message!';
  // ....
}
```

Fixes https://github.com/flutter/flutter/issues/178172


yjbanov (approved):
lgtm

Issue #178326, opened 2025-11-11
Unpin `package:dwds` from `26.1.0`


bkonyi:
https://github.com/flutter/flutter/pull/178267 needed to be reverted due to possible [failures](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8698521905652382113/+/u/run_web_benchmarks_ddc_hot_reload/stdout) introduced by the update to `package:dwds` `26.2.2`.

These failures should be investigated and the version pin removed.

PR #178327, opened 2025-11-11
Pin package:dwds to 26.1.0

bkonyi:
See https://github.com/flutter/flutter/issues/178326


Piinks (approved):
LGTM so we can land the pub roller, although to follow up on in the upcoming dash forum proposal to unpin deps.

PR #178335, opened 2025-11-11
[ Tool ] Only process a single unhandled tool exception

### LINKED ISSUE #178318: `flutter_tool` crash handler should not report multiple crashes for a single process
While debugging https://github.com/flutter/flutter/issues/178317, it became clear that it's possible for multiple crash events to be sent for a single process while the tool's shutdown hooks are executing.

Once the root error handler in the tool encounters an exception, it reports the crash to our crash analytics service before running all registered shutdown hooks. Once these shutdown hooks complete, the tool exits.

However, since the shutdown hooks can be asynchronous, it's possible for additional exceptions to be thrown due to the introduction of asynchronous gaps. In the referenced issue, the original exception was thrown from a file watcher callback that was processing files after a `flutter pub get`, resulting in exceptions being thrown from the same location repeatedly while the tool was shutting down. This caused the root error handler to be invoked repeatedly, causing multiple crash reports to be sent for a single process and the shutdown hooks to be invoked again. Eventually, the tool is able to exit successfully, but only after reporting multiple crashes.

In order to prevent the root error handler from attempting to report additional crashes while executing the shutdown hooks, we should prevent the error handler from being reentrant and return from the handler immediately if we're already in the process of shutting down the tool due to an exception.

bkonyi:
Without this change, if multiple asynchronous exceptions are thrown while processing an exception, multiple exception analytics events can be sent for a single process crash, skewing crashlytics data.

Fixes https://github.com/flutter/flutter/issues/178318

### DIFF
--- a/packages/flutter_tools/test/general.shard/runner/runner_test.dart
+++ b/packages/flutter_tools/test/general.shard/runner/runner_test.dart
@@ -323,10 +323,10 @@ void main() {
 
         // This is the main check of this test.
         //
-        // We are checking that, even though crash reporting failed with an
-        // exception on the first attempt, the second attempt tries to report the
-        // *original* crash, and not the crash from the first crash report
-        // attempt.
+        // We are checking that, even though multiple asynchronous errors were
+        // thrown, only a single crash report is sent. This ensures that a
+        // single process crash can't result in multiple crash events.
+
         // This test only makes sense if we've thrown more than one exception.
         expect(exceptionCount, greaterThan(1));
         expect(exceptionCount, command.exceptionCount);
--- a/packages/flutter_tools/test/general.shard/runner/runner_test.dart
+++ b/packages/flutter_tools/test/general.shard/runner/runner_test.dart
@@ -882,8 +882,6 @@ class CrashingFlutterCommand extends FlutterCommand {
 }
 
 class MultipleExceptionCrashingFlutterCommand extends FlutterCommand {
-  MultipleExceptionCrashingFlutterCommand();
-
   final _completer = Completer<void>();
 
   @override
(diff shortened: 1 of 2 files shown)


packages/flutter_tools/test/general.shard/runner/runner_test.dart
```diff
@@ -819,6 +881,36 @@ class CrashingFlutterCommand extends FlutterCommand {
...
+
+  @override
+  String get description => '';
+
+  @override
+  String get name => 'crash';
+
+  Future<void> get doneThrowing => _completer.future;
+
+  var exceptionCount = 0;
```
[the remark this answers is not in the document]
bkonyi: Conflicts with a lint rule.

packages/flutter_tools/test/general.shard/runner/runner_test.dart
```diff
@@ -289,6 +290,67 @@ void main() {
...
+          runZonedGuarded<Future<void>?>(
+            () {
+              unawaited(
+                runner.run(
+                  <String>['crash'],
+                  () => <FlutterCommand>[command],
+                  // This flutterVersion disables crash reporting.
+                  flutterVersion: '[user-branch]/',
+                  reportCrashes: true,
+                  shutdownHooks: ShutdownHooks(),
```
vashworth, line 310: Can we verify shutdown hooks are only called once?

bkonyi: Looking at the `ShutdownHooks` implementation, we actually guard against running the hooks more than once as it's possible to invoke `exitWithHooks` after returning from a command and for it to be invoked due to an outstanding asynchronous event that throws.

I've updated the PR description to reflect this. The main thing we're testing here is that we don't try and sent multiple events due to asynchronous exceptions being thrown in quick succession. I've verified this test fails without the fix.

reidbaker (approved):

bkonyi:
> Are you at all worried about the loss of data by throwing the first tool error we see and none of the rest?

No, I don't think so. Errors thrown after the initial error are likely due to a cascading failure, and we'd have to explicitly check for multiple reports (possibly with different stack traces) from an individual client within a short timeframe to even identify if multiple exceptions were thrown during a crash.

vashworth (approved):
LGTM

PR #122585, opened 2023-03-14
Make InkDecoration not paint if the ink is not visible

tvolkert:
Make InkDecoraiton not paint if the ink is not visible
    
This introduces a `isVisible` field in `InkDecoration` to control whether
the decoration is painted. It also introduces a `Visibility.of()` method
that can be used to determine the visibility state of widgets in the tree
based on Visibility ancestry state. This then updates Ink to consult
`Visibility.of()` and set the visible flag on its ink decoration
accordingly.
    
Finally, this updates `IndexedStack` to add a `Visibility` widget around
each of the stack's children, so that non-visible children of the stack
will be able to use `Visibility.of()` to know their visibility.
    
Putting it all together, this means that ink no longer paints itself when
it's in a non-selected index of an indexed stack.
    
https://github.com/flutter/flutter/issues/59963

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #178368, opened 2025-11-12
Fix InkDecoration painting in Visibility replacement.

### LINKED ISSUE #177989: InputChip does not draw a Color as Visibility replacement
### Steps to reproduce

If an InputChip is used somewhere in the tree as a replacement of a Visibility, it does not show the defined color property.

Tested on iPhone (Hardware and Simulator)

### Expected results

Image

### Actual results

Image

### Code sample

```dart
    var chip = Material(
      color: Colors.red,
      child: InputChip(
        showCheckmark: false,
        color: WidgetStateProperty.resolveWith((states) {
          print(states);

          return Colors.blue;
        },),
        onPressed: () {},
        label: const Text("Something"),
      ),
    );

    return Visibility(
      replacement:chip,
      visible: true, // if false, color is not shown
      child: chip,
    );

``` 

### Screenshots or Video

_No response_

### Logs

_No response_

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.6, on macOS 15.5 24F74 darwin-arm64, locale
    de-DE)
[✓] Xcode - develop for iOS and macOS (Xcode 16.4)

```

</details>

ksokolovskyi:
Fixes https://github.com/flutter/flutter/issues/177989

### Description
- Updates `Visibility` widget to not report `visibility` as `false` to the replacement widget and it's descendants, as replacement widget is actually visible
- Adds tests to verify new behavior

For reference, the `Visibility.of()` was introduced in https://github.com/flutter/flutter/pull/122585

| BEFORE | AFTER |
| - | - |
| before | after |

<details closed><summary>Code sample</summary>

```dart

void main() {
  runApp(const App());
}

class App extends StatelessWidget {
  const App({super.key});

  @override
    return MaterialApp(
      home: Scaffold(
        backgroundColor: Colors.white,
        body: Center(
          child: Column(
            spacing: 20,
            crossAxisAlignment: CrossAxisAlignment.center,
            children: [
              Visibility(
                visible: false,
                replacement: Material(
                  color: Colors.red,
                  child: Padding(
                    padding: const EdgeInsets.all(8),
                    child: InputChip(
                      color: WidgetStateProperty.resolveWith(
                        (states) => Colors.blue,
                      ),
                      onPressed: () {},
                      label: const Text("Something"),
                    ),
                  ),
                ),
                child: const SizedBox.shrink(),
              ),
            ],
          ),
        ),
      ),
    );
  }
}

```

</details>

## Pre-launch Checklist

- [X] I signed the [CLA].


victorsanni (approved):

packages/flutter/lib/src/widgets/visibility.dart
```diff
@@ -286,7 +293,7 @@ class Visibility extends StatelessWidget {
         result = visible ? child : replacement;
       }
     }
-    return _VisibilityScope(isVisible: visible, child: result);
+    return _VisibilityScope(isVisible: visible || result == replacement, child: result);
```
QuncCccccc, line 296: Will this be a breaking change that Visibility.of() will always return true even if we set something like:
```
Visibility(
  visible: false,
  child: xxx
)
```
Before this change, it will return false; but now it seems will return true.

ksokolovskyi: Hi, thanks a lot for taking a look at this PR!

With the change in this PR, the `Visibility.of(context)` will return `true` for the `replacement` widget if it is rendered as it is effectively visible.
```
Visibility(
  visible: false,
  child: xxx,
  replacement: xxx,)
```

However, when we have `maintainState: true`, then the `replacement` is not being rendered and the `Visibility.of(context)` will return `false` for the `child` widget.
```
Visibility(
  maintainState: true,
  visible: false,
  child: xxx,
  replacement: xxx,
)
```

I haven't thought that this would be a breaking change, but now realize that even though the tests haven't failed, the customers relying on `Visibility.of(context)` in `replacement` widgets will be affected.

What can you suggest doing in this case? Adding a breaking changes doc for the website or considering other solutions for the issue?

QuncCccccc: > With the change in this PR, the Visibility.of(context) will return true for the replacement widget if it is rendered as it is effectively visible.

I'm thinking if we have:
```dart
Visibility(
  visible: false,
  child: xxx,
  replacement: SizedBox.shrink(),)
```

Should `Visibility.of(context)` return true as well? It feels a bit odd that we cannot see anything but the `of` method return true.

Also for the case I gave in the above comment:
```
Visibility(
  visible: false,
  child: xxx
)
```
This case should be used a lot and I think the of() method should definitely be false. Maybe checking whether the replacement is a `SizedBox.shrink()`? But I'm not sure whether it's elegant enough:)

QuncCccccc (commented):
Sorry for the late response. I was OOO last week.

packages/flutter/lib/src/widgets/visibility.dart
```diff
@@ -237,10 +237,17 @@ class Visibility extends StatelessWidget {
...
+  /// ancestor tree, or if all ancestor [Visibility] widgets are effectively
+  /// visible.
+  ///
+  /// A [Visibility] widget is "effectively visible" if its [visible]
+  /// property is `true`, or if its [visible] property is `false` but it
+  /// is currently showing its [replacement] widget.
+  ///
+  /// If *any* ancestor [Visibility] widget is "effectively invisible" (meaning
+  /// its [visible] property is `false` and it is *not* showing its
+  /// [replacement]), then this method will return `false`.
```
dkwingsmt, line 250: I find the new doc a bit repetitive, and making the idea of "effectively visible" harder to understand than what should have been (if I understand correctly). I propose the following doc. Let me know what you think.
```dart
  /// Whether the element is visible based on the configuration of its 
  /// [Visibility] ancestors.
  ///
  /// Specifically, when a [Visibility] ancestor's [visible] property is false, 
  /// its [child] subtree is considered hidden, while its [replacement] 
  /// subtree is considered visible for the purposes of that ancestor.
  ///
  /// If there are no [Visibility] ancestors in the element's path, 
  /// this returns true.
```

dkwingsmt:

PR #178370, opened 2025-11-12
New optimized general convex path shadow algorithm

### LINKED ISSUE #170764: [Impeller] Tessellation for convex shadows (non-uniform RRects)
Currently Impeller has two techniques for rendering a shadow:
  * Fast RRect/Rect/Circle path that works in a single draw.
  * Generic Gaussian Blur 3-Pass

The second case is required for blurring rasters or complex shapes, but is much slower than the fast path. Skia has an intermediate approach, where they compute the tessellated shadow for convex and concave shapes. This is useful in case buttons and such end up non-uniform.

We should devise a similar approach, which as a motivating case can use the application in:  https://github.com/flutter/flutter/issues/169423

Sources:
  * https://github.com/google/skia/blob/b6c161a0f51f7af0749917ddd222f9d297cb076e/src/utils/SkShadowTessellator.cpp#L263

flar:
Convex paths will now use an optimized mesh-based algorithm to render shadows. The algorithm is based on the code in SkShadowTessellator.

Fixes https://github.com/flutter/flutter/issues/170764

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/impeller/toolkit/interop/impeller_unittests.cc
```diff
@@ -520,10 +520,10 @@ TEST_P(InteropPlaygroundTest, CanRenderShadows) {
     builder.DrawShadow(path_builder.Build(), shadow_color, 4.0f, false, 1.0f);
   }
   {
-    hpp::Paint red_paint;
-    red_paint.SetColor(
-        ImpellerColor{1.0, 0.0, 0.0, 1.0, kImpellerColorSpaceSRGB});
-    builder.DrawRect(box, red_paint);
+    // hpp::Paint red_paint;
```
flar: Whoops, not sure why I deleted this code...

flar:
It looks like the new Blur contents overrides Render instead of SetPassInfo. I need to check if it is bypassing some of the setup in the parent class.

gaaclarke (commented):
I'll hold off reviewing until the goldens are good.  Ping me when you're ready.

flar:
Updating to head as a possible de-flake because the "smiley face android engine test" does not look related and doesn't reproduce locally.

flar:
The new goldens all make sense to me so this is ready for review now.

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
```diff
@@ -78,4 +78,17 @@ float16_t IPSigmoid(float16_t x) {
   return 1.03731472073hf / (1.0hf + exp(-4.0hf * x)) - 0.0186573603638hf;
 }
 
+/// Converts a fraction in the range [0,1] to a guassian weighted distribution
+/// over the same range ([0,1]).
+float16_t IPFractionToGaussian(float16_t fraction) {
```
gaaclarke:
```suggestion
float16_t IPHalfFractionToGaussian(float16_t fraction) {
```

This is the naming convention of the other functions.

gaaclarke: This is making a gaussian cumulative distribution function, not the probability distribution one would usually associate with a "gaussian distribution". Can you update the name and the docstring to mention "cumulative distribution function".  It would help me personally if you threw in there that it was sigmoid shaped.  Maybe name it something like IPHalfFractionToGaussianCDF?

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
```diff
@@ -78,4 +78,17 @@ float16_t IPSigmoid(float16_t x) {
...
+  // IPErf produces outputs over [0, 1] from an input range of [-2, +2].
```
gaaclarke, line 84: Erf goes goes [-1,1].

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -0,0 +1,869 @@
...
+#include "flutter/impeller/entity/geometry/shadow_path_geometry.h"
+#include "flutter/testing/testing.h"
+
+namespace impeller {
+namespace testing {
+
+using namespace flutter;
+
+namespace {
+class PathReflector : public PathReceiver {
```
gaaclarke, line 22: Docstring please.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -0,0 +1,869 @@
...
+  static PathReflector ReflectAroundY(Scalar y_coordinate) {
+    return PathReflector(1.0f, 0.0f, -1.0f, y_coordinate * 2.0f);
+  }
+
+  static PathReflector ReflectAround(const Point& anchor) {
+    return PathReflector(-1.0f, anchor.x * 2.0f, -1.0f, anchor.y * 2.0f);
+  }
+
+  // |PathReceiver|
+  virtual void MoveTo(const Point& p2, bool will_be_closed) {
```
gaaclarke: Use `override` here and below please.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -0,0 +1,869 @@
...
+    path = ReflectPath(path);
+  }
+  return path;
+}
+}  // namespace
+
+TEST_P(AiksTest, DrawShadowDoesNotOptimizeHourglass) {
+  DisplayListBuilder builder;
+  builder.Clear(DlColor::kWhite());
+  builder.Scale(GetContentScale().x, GetContentScale().y);
```
gaaclarke, line 237: I thought we got rid of the need to do this, did you find you had to do it?

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -0,0 +1,869 @@
...
+  path_builder.Close();
+  DlPath path = path_builder.TakePath();
+  if (!clockwise) {
+    path = ReflectPath(path);
+  }
+  return path;
+}
+}  // namespace
+
+TEST_P(AiksTest, DrawShadowDoesNotOptimizeHourglass) {
```
gaaclarke, line 234: Is the main thing you are testing with these goldens the wireframes?  If so do we need to draw them on all 3 platforms since I imagine the code for generating the wireframes is crossplatform.

engine/src/flutter/impeller/display_list/canvas.h
```diff
@@ -283,31 +283,17 @@ class Canvas {
...
-        geom_;  // optional stack allocation
+    virtual ~BlurShape() = default;
+    virtual Rect GetBounds() const = 0;
+    virtual std::shared_ptr<SolidBlurContents> BuildBlurContent(
+        Sigma sigma) = 0;
+    virtual Geometry& BuildDrawGeometry() = 0;
   };
+  class RRectBlurShape;
+  class RSuperellipseBlurShape;
+  class PathBlurShape;
```
gaaclarke, line 296: This is kind of odd to me, why not just make BlurShape public and put these in an anonymous namespace in the .cc file?  Usually my rule of thumb is the less in the header, the better since that helps with compilation and with understanding.

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -177,24 +181,94 @@ static std::unique_ptr<EntityPassTarget> CreateRenderTarget(
...
-Geometry& Canvas::RSuperellipseBlurShape::BuildGeometry(Rect rect,
-                                                        Scalar radius) {
-  return geom_.emplace(rect, radius);
-}
+  Geometry& BuildDrawGeometry() override {
+    return geom_.emplace(rect_, Size(corner_radius_));
+  }
+
+ private:
+  const Rect& rect_;
```
gaaclarke: Don't keep a potential dangling reference, just copy the rect here.

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -177,24 +181,94 @@ static std::unique_ptr<EntityPassTarget> CreateRenderTarget(
...
+    contents->SetShape(rect_, corner_radius_);
+    return contents;
+  }
+
+  Geometry& BuildDrawGeometry() override {
+    return geom_.emplace(rect_, corner_radius_);
+  }
+
+ private:
+  const Rect& rect_;
```
gaaclarke: Same here.

engine/src/flutter/impeller/display_list/canvas.h
```diff
@@ -283,31 +283,17 @@ class Canvas {
...
-    Geometry& BuildGeometry(Rect rect, Scalar radius) override;
-
-   private:
-    std::optional<RoundSuperellipseGeometry>
-        geom_;  // optional stack allocation
+    virtual ~BlurShape() = default;
+    virtual Rect GetBounds() const = 0;
+    virtual std::shared_ptr<SolidBlurContents> BuildBlurContent(
+        Sigma sigma) = 0;
+    virtual Geometry& BuildDrawGeometry() = 0;
```
gaaclarke: Shouldn't this be `const Geometry& BuildDrawGeometry()`?

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -177,24 +181,94 @@ static std::unique_ptr<EntityPassTarget> CreateRenderTarget(
...
+
+ private:
+  // This value was determined by empirical eyesight tests so that the
+  // shadow mesh results will match the results of the shape-specific
+  // optimized shadow shaders.
+  static constexpr Scalar kSigmaScale = 2.8f;
+
+  const PathSource& source_;
+  Tessellator& tessellator_;
+  const Matrix& matrix_;
```
gaaclarke: More raw pointers.  Can you use raw_ptr if you want to keep these as raw and also add documentation to explain why these won't be dangling.

engine/src/flutter/impeller/display_list/canvas.h
```diff
@@ -395,22 +381,24 @@ class Canvas {
...
   bool AttemptDrawAntialiasedCircle(const Point& center,
                                     Scalar radius,
                                     const Paint& paint);
 
-  bool AttemptDrawBlurredRRect(const Rect& rect,
-                               Size corner_radii,
-                               const Paint& paint);
+  /// Returns the radius common to both width and height of all corners,
+  /// or -1 if the radii are not uniform.
+  Scalar GetCommonRRectLikeRadius(const RoundingRadii& radii);
```
gaaclarke:
```suggestion
  Scalar GetCommonRRectLikeRadius(const RoundingRadii& radii) const;
```

gaaclarke: Are these other ones const too?  Looks like itAttemptDrawBlurredRRect

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -689,8 +813,20 @@ void Canvas::DrawOval(const Rect& rect, const Paint& paint) {
     return;
   }
 
-  if (AttemptDrawBlurredRRect(rect, rect.GetSize() * 0.5f, paint)) {
-    return;
+  if (IsShadowBlurDrawOperation(paint)) {
+    if (rect.IsSquare()) {
+      RRectBlurShape shape(rect, rect.GetWidth() * 0.5f);
```
gaaclarke, line 848: Pull out constant please.

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -762,22 +898,23 @@ void Canvas::DrawArc(const Arc& arc, const Paint& paint) {
...
       return;
     }
+  }
 
-    if (paint.style == Paint::Style::kFill) {
-      Entity entity;
-      entity.SetTransform(GetCurrentTransform());
-      entity.SetBlendMode(paint.blend_mode);
+  auto& rect = round_rect.GetBounds();
+  auto& radii = round_rect.GetRadii();
```
gaaclarke:
```suggestion
  auto rect = round_rect.GetBounds();
  auto radii = round_rect.GetRadii();
```

These functions are surely returning trivial data structures not references.  I'm surprised this doesn't crash some places.

Can you remove the type deduction please ( https://google.github.io/styleguide/cppguide.html#Type_deduction )

engine/src/flutter/impeller/display_list/canvas.cc
```diff
@@ -811,12 +948,14 @@ void Canvas::DrawDiffRoundRect(const RoundRect& outer,
...
 void Canvas::DrawRoundSuperellipse(const RoundSuperellipse& round_superellipse,
                                    const Paint& paint) {
+  if (IsShadowBlurDrawOperation(paint)) {
+    if (AttemptDrawBlurredRSuperellipse(round_superellipse, paint)) {
+      return;
+    }
+  }
+
   auto& rect = round_superellipse.GetBounds();
   auto& radii = round_superellipse.GetRadii();
```
gaaclarke: (same here)

engine/src/flutter/impeller/entity/contents/shadow_vertices_contents.h
```diff
@@ -0,0 +1,51 @@
...
+#include "impeller/geometry/color.h"
+
+namespace impeller {
+
+/// A vertices contents for (optional) per-color vertices + texture and any
+/// blend mode.
+class ShadowVerticesContents final : public SolidBlurContents {
+ public:
+  static std::shared_ptr<ShadowVerticesContents> Make(
+      const ShadowPathGeometry* geometry);
```
gaaclarke: Please get rid of the raw pointer ( https://google.github.io/styleguide/cppguide.html#Ownership_and_Smart_Pointers )

Can't you use a unique_ptr here and pass ownership of the geometry to the shadowverticescontents?

engine/src/flutter/impeller/entity/geometry/fill_path_geometry.h
```diff
@@ -49,6 +49,21 @@ class FillPathSourceGeometry : public Geometry {
   FillPathSourceGeometry& operator=(const FillPathSourceGeometry&) = delete;
 };
 
+/// @brief A Geometry that produces fillable vertices from a |PathSource| object
+///        using the |FillPathSourceGeometry|.
+class FillPathFromSourceGeometry final : public FillPathSourceGeometry {
+ public:
+  explicit FillPathFromSourceGeometry(
+      const PathSource& source,
+      std::optional<Rect> inner_rect = std::nullopt);
```
gaaclarke: Can you document inner_rect please.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+    //
+    // It's initial value is actually the uninitialized constant so that the
+    // algorithm can treat it specially the first time it is encountered.
+    Scalar umbra_fraction = kFractionUninitialized;
+
+    // Pointers used to create a circular linked list while pruning the umbra
+    // polygon. The final list of vertices that remain in the umbra polygon
+    // are the vertices that remain on this linked list from a "head" pin.
+    UmbraPin* pNext = nullptr;
+    UmbraPin* pPrev = nullptr;
```
gaaclarke: Can't this be inferred from the index in the vector?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+    Scalar last_direction_ = 0.0f;
+    size_t change_count = 0u;
+
+    void AccumulateDirection(Scalar new_direction) {
+      if (last_direction_ == 0.0f || last_direction_ * new_direction < 0.0f) {
+        last_direction_ = std::copysign(1.0f, new_direction);
+        change_count++;
+      }
+    }
+  } x_direction_detector_, y_direction_detector_;
```
gaaclarke: nit: split out ivar declaration and struct declaration.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+
+  // The vertex mesh result that represents the shadow, to be rendered
+  // using a modified indexed variant of DrawVertices that also adjusts
+  // the alpha of the colors on a per-pixel basis by mapping their linear
+  // gaussian coefficients into the associated gaussian integral values.
+  const impeller::Tessellator::Trigs& trigs_;
```
gaaclarke: raw_ptr with lifetime comment

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+  static constexpr Scalar kCrossTolerance = 1.0f / 2048.0f;
+  static constexpr Scalar kIntersectionTolerance = 1.0e-6f;
+
+  static constexpr Scalar FiniteVectorLengthSquared(Vector2 v) {
+    return !v.IsFinite() ? -1.0f : v.Dot(v);
+  }
+
+  static constexpr inline bool OutsideInterval(Scalar numer,
+                                               Scalar denom,
+                                               bool denomPositive) {
```
gaaclarke:
```suggestion
                                               bool denom_positive) {
```

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+
+  // Structure to store the result of computing the intersection between
+  // 2 pins, containing the point of intersection and the relative fractions
+  // at which the pins intersected (expressed as a ratio of 0 to 1 where
+  // 0 represents intersecting at the path outline and 1 represents
+  // intersecting at the tip of the pin where the umbra is darkest.
+  struct PinIntersection {
+    Point intersection;
+    Scalar fraction0;
+    Scalar fraction1;
```
gaaclarke, line 378: std::array?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: We don't typically use hungarian notation.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+  // After the umbra_vertices of the pins are accumulated and linked into a
+  // ring using their pPrev/pNext pointers, compute the best surviving umbra
+  // vertex for each pin and set its location and index into the UmbraPin.
+  void PopulateUmbraVertices();
+
+  // Appends a fan of penumbra vertices centered on the path vertex of the
+  // |p_curr_pin| starting from the absolute point |fan_start| and ending
+  // at the absolute point |fan_end|, both of which should be equi-distant
+  // from the path vertex. The index of the vertex at |fan_start| should
+  // already be in the vector of vertices at an index given by |start_index|.
```
gaaclarke: nit: This is in Dart docstring format.  I would prefer the style we use in the engine which documents the arguments in their own section and what is returned.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+                         const Matrix& matrix,
+                         Scalar occluder_height,
+                         const Tessellator::Trigs& trigs)
+    : inverted_matrix_(matrix.Invert()),
+      occluder_height_(occluder_height),
+      trigs_(trigs) {
+  if (!matrix.IsInvertible()) {
+    Invalidate();
+    return;
+  }
```
gaaclarke, line 501: Instead of creating invalid objects, please use absl::StatusOr with a factory function.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+  // shape_area_ is accumulating sum(quad area).
+  //
+  // The final combined average weight factor will be (3 * sum(quad area)).
+  centroid_ += (v0 + v1) * quad_area;
+  shape_area_ += quad_area;
+
+  return result;
+}
+
+void PolygonInfo::FinalizeCentroid() {
```
gaaclarke: nit: make this a const function so that the side effects are clear at the call site:

```c++
centroid_ = FinalizeCentroid();
```

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+// at that sub-pixel grid are ignored, collinear points are reduced to just
+// the endpoints, and the centroid is updated from the remaining non-duplicate
+// grid points.
+void PolygonInfo::Write(Point point) {
+  if (!is_valid_) {
+    return;
+  }
+
+  if (path_ended_) {
+    Invalidate();
```
gaaclarke: Return an absl::Status

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+// centroid accumulation do its math for ever segment in the path, but
+// going forward we don't need the extra pin in the shape so we verify that
+// it is a duplicate and then we delete it.
+void PolygonInfo::EndContour() {
+  if (!is_valid_) {
+    return;
+  }
+
+  if (path_ended_) {
+    Invalidate();
```
gaaclarke: Same

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+
+  // PathTessellator always ensures the path is closed back to the origin
+  // by an extra call to Write(Point).
+  FML_DCHECK(pins_.front().path_vertex == pins_.back().path_vertex);
+  pins_.pop_back();
+  path_ended_ = true;
+}
+
+// Adjust the device point to its nearest sub-pixel grid location.
+Point PolygonInfo::ToDeviceGrid(Point point) {
```
gaaclarke:
```suggestion
Point PolygonInfo::ToDeviceGrid(Point point) const {
```

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+}
+
+// Use the direction change accumulators to count the number of times
+// that edges change direction in X and Y and return whether the path
+// can still be classified as simple and convex.
+bool PolygonInfo::CheckEdgeDirection(const Vector2 edge_vector) {
+  x_direction_detector_.AccumulateDirection(edge_vector.x);
+  y_direction_detector_.AccumulateDirection(edge_vector.y);
+  return x_direction_detector_.change_count <= 3u &&
+         y_direction_detector_.change_count <= 3u;
```
gaaclarke: pull out magic number to constant

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+// naming and differences in the methods on Point and Vertex2.
+std::optional<PolygonInfo::PinIntersection> PolygonInfo::ComputeIntersection(
+    UmbraPin* pPin0,
+    UmbraPin* pPin1) {
+  Vector2 v0 = pPin0->path_delta;
+  Vector2 v1 = pPin1->path_delta;
+  Vector2 tip_delta = pPin1->pin_tip - pPin0->pin_tip;
+  Vector2 w = tip_delta;
+  Scalar denom = pPin0->path_delta.Cross(pPin1->path_delta);
+  bool denomPositive = (denom > 0);
```
gaaclarke:
```suggestion
  bool denom_positive = (denom > 0);
```

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+// intersection.
+//
+// The intersection structure will be reset to empty otherwise.
+//
+// This method was converted nearly verbatim from the Skia source files
+// SkShadowTessellator.cpp and SkPolyUtils.cpp, except for variable
+// naming and differences in the methods on Point and Vertex2.
+std::optional<PolygonInfo::PinIntersection> PolygonInfo::ComputeIntersection(
+    UmbraPin* pPin0,
+    UmbraPin* pPin1) {
```
gaaclarke: variable naming

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+// The intersection structure will contain the fractional distances along the
+// pins of the intersection and the intersection point itself if there is an
+// intersection.
+//
+// The intersection structure will be reset to empty otherwise.
+//
+// This method was converted nearly verbatim from the Skia source files
+// SkShadowTessellator.cpp and SkPolyUtils.cpp, except for variable
+// naming and differences in the methods on Point and Vertex2.
+std::optional<PolygonInfo::PinIntersection> PolygonInfo::ComputeIntersection(
```
gaaclarke, line 860: It seems like you are using std::optional for error handling?  If i'm reading that right, please use absl::StatusOr

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+    UmbraPin* pPin0,
+    UmbraPin* pPin1) {
+  Vector2 v0 = pPin0->path_delta;
+  Vector2 v1 = pPin1->path_delta;
+  Vector2 tip_delta = pPin1->pin_tip - pPin0->pin_tip;
+  Vector2 w = tip_delta;
+  Scalar denom = pPin0->path_delta.Cross(pPin1->path_delta);
+  bool denomPositive = (denom > 0);
+  Scalar sNumer, tNumer;
```
gaaclarke: I don't know what s and t mean.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+    // segments are parallel, but not collinear
+    if (!ScalarNearlyZero(tip_delta.Cross(pPin0->path_delta),
+                          kCrossTolerance) ||
+        !ScalarNearlyZero(tip_delta.Cross(pPin1->path_delta),
+                          kCrossTolerance)) {
+      return std::nullopt;
+    }
+
+    // Check for zero-length segments
+    Scalar v0dotv0 = FiniteVectorLengthSquared(v0);
```
gaaclarke: instead of having numbers in these names, could you just use an std::array?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -0,0 +1,1264 @@
...
+  } x_direction_detector_, y_direction_detector_;
+
+  // The vertex mesh result that represents the shadow, to be rendered
+  // using a modified indexed variant of DrawVertices that also adjusts
+  // the alpha of the colors on a per-pixel basis by mapping their linear
+  // gaussian coefficients into the associated gaussian integral values.
+  const impeller::Tessellator::Trigs& trigs_;
+  std::vector<Point> vertices_;
+  std::vector<uint16_t> indices_;
+  std::vector<Scalar> gaussians_;
```
gaaclarke, line 337: Add docstring please.  It's not clear what these are.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.h
```diff
@@ -0,0 +1,108 @@
...
+  /// This number is also the count of gaussian coefficients in the mesh
+  /// since the two are assigned 1:1.
+  size_t GetVertexCount() const { return vertices_.size(); }
+
+  /// The count of the indices that define the mesh.
+  size_t GetIndexCount() const { return indices_.size(); }
+
+  const std::vector<Point>& GetVertices() const { return vertices_; }
+  const std::vector<uint16_t>& GetIndices() const { return indices_; }
+  const std::vector<Scalar>& GetGaussians() const { return gaussians_; }
```
gaaclarke, line 54: Or maybe that comment would be better here.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.h
```diff
@@ -0,0 +1,108 @@
...
+  /// The tessellator is used to get a cached set of |Trigs| for the
+  /// radii associated with the mesh around various corners in the path.
+  static std::shared_ptr<ShadowVertices> MakeAmbientShadowVertices(
+      Tessellator& tessellator,
+      const PathSource& source,
+      Scalar occluder_height,
+      const Matrix& matrix);
+
+ private:
+  const std::shared_ptr<ShadowVertices> shadow_vertices_;
```
gaaclarke: Does this need to be a shared_ptr?  Who else will maintain ownership of this?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
```diff
@@ -0,0 +1,344 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
gaaclarke, line 3: There is a large amount of logic in shader_path_geometry.cc.  These unit tests seem thin in comparison.  Do we have coverage for all that logic?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
```diff
@@ -0,0 +1,344 @@
...
+
+namespace impeller {
+namespace testing {
+
+using namespace flutter;
+
+namespace {
+
+#if SHOW_VERTICES
+static void ShowVertices(
```
gaaclarke: nit: no need for static functions inside of an anonymous namespace.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
```diff
@@ -0,0 +1,344 @@
...
+
+#include "flutter/third_party/skia/src/core/SkVerticesPriv.h"  // nogncheck
+#include "flutter/third_party/skia/src/utils/SkShadowTessellator.h"  // nogncheck
+
+#define SHOW_VERTICES false
+
+namespace impeller {
+namespace testing {
+
+using namespace flutter;
```
gaaclarke: using namespace is technically forbidden, can you list the symbols you need?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
```diff
@@ -0,0 +1,344 @@
...
+#include "flutter/impeller/entity/geometry/shadow_path_geometry.h"
+
+#include "flutter/display_list/geometry/dl_path.h"
+#include "flutter/display_list/geometry/dl_path_builder.h"
+#include "gtest/gtest.h"
+
+#include "flutter/third_party/skia/src/core/SkVerticesPriv.h"  // nogncheck
+#include "flutter/third_party/skia/src/utils/SkShadowTessellator.h"  // nogncheck
+
+#define SHOW_VERTICES false
```
gaaclarke:
```suggestion
#define IMPELLER_SHOW_VERTICES false
```

Please add a prefix to avoid any collisions.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
```diff
@@ -0,0 +1,344 @@
...
+  }
+  if (SimilarPoint(p1_1, p2_2) && SimilarPoint(p1_2, p2_1)) {
+    return true;
+  }
+  return false;
+}
+
+static bool SimilarPointTrio(Point p1_1,
+                             Point p1_2,
+                             Point p1_3,  //
```
gaaclarke: std::array<Point, 3> ?

engine/src/flutter/impeller/geometry/path_source.h
```diff
@@ -109,6 +109,38 @@ class EllipsePathSource : public PathSource {
...
+
+  void CubicTo(const Point& cp1, const Point& cp2, const Point& p2) override {
+    receiver_.CubicTo(matrix_ * cp1, matrix_ * cp2, matrix_ * p2);
+  }
+
+  void Close() override { receiver_.Close(); }
+
+ private:
+  PathReceiver& receiver_;
+  const impeller::Matrix& matrix_;
```
gaaclarke, line 142: Raw pointers.

engine/src/flutter/impeller/geometry/point.h
```diff
@@ -201,9 +200,83 @@ struct TPoint {
...
+  constexpr Type GetLength() const { return std::sqrt(GetLengthSquared()); }
+
+  /// Returns the distance (squared) from this point to the closest point on
+  /// the line segment p0 -> p1.
+  ///
+  /// If the projection of this point onto the line defined by the two points
+  /// is between them, the distance (squared) to that point is returned.
+  /// Otherwise, we return the distance (squared) to the endpoint that is
+  /// closer to the projected point.
+  constexpr Type GetDistanceToSegmentSquared(TPoint p0, TPoint p1) const {
```
gaaclarke: Is this ever used as a constexpr?  It would be nice to have the definition in the .cc file.

gaaclarke (changes requested):
Woo, I'm going to need a nap now.  I wish there was a way to make these PRs smaller in the future =). Looking good for the most part but it's a big PR so there was a lot to go over.

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
flar: Not the function I'm calling... `IPErf` perhaps it is misnamed as well?

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: The Scale? Did we? `git grep` found 214 instances of that call.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: It gives a view that shows how the mesh is formed for edification purposes only. Also, at one point (if you look back in the commits) I was also creating meshes using the equivalent Skia call to see if our 2 versions of the algorithms agreed.

They can be eliminated if we don't want them.

engine/src/flutter/impeller/display_list/canvas.h
flar: There is a private method `AttemptDrawBlur` that takes an instance of this class. It is internal only and should not leak outside this class.

Originally these 3 classes were defined inline in the header, but I moved them to the cc file to remove implementation from the header. They cannot implement this private interface unless they are mentioned in the class's (private) definition, so I needed the forward declaration.

So, implementation in header file?
Or publicly leaked internal interface?
Or forward declaration?

engine/src/flutter/impeller/display_list/canvas.h
flar: It looks like it can.

engine/src/flutter/impeller/display_list/canvas.cc
flar: I'm following conventions used in other areas to prevent copying of data during rendering. I'll look and see what the potential consequences are.

engine/src/flutter/impeller/display_list/canvas.h
flar: Even better - static functions. At least on this one. I'll check the others.

flar: `IsShadowBlurDrawOperation` was the only other one I could find that could be const or static and I made it static.

engine/src/flutter/impeller/display_list/canvas.cc
flar: A constant for the ratio of a radius to a diameter?

engine/src/flutter/impeller/entity/contents/shadow_vertices_contents.h
flar: I'm going to try something else first. That class is just a thin veneer over an existing shared_ptr to a ShadowVertices class. I'll break it out so that this class works directly on a ShadowVertices and that way I can just pass around the already created shared ptr.

flar: That seems to work...

engine/src/flutter/impeller/entity/geometry/fill_path_geometry.h
flar: Unused, deleted.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: It cannot be inferred because the linked list is constantly being edited.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: That's an odd place for that field to appear. The comment is about the 3 vectors that follow and has nothing to do with the trigs.

Trigs lifetimes have been a sore point for a while. I'd like to devise a better system. Its lifetime is complicated by being sometimes a shared object and sometimes an allocated object. Not a good solution.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: No, it's not a list of fractions, it's the 2 fractions from the perspective of each endpoint. Perhaps better names might make that clearer.

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
gaaclarke: It's been a week but I'm pretty sure I mapped the math we have now in octave and it was -1, 1.  I can double check if you need it.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
gaaclarke: Ugh, wonder what happened there, dreamtime coding session where I fixed all my annoyances?

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
gaaclarke: Did you forget to push your PR?  I don't know if the thumbs up means "i did it" or "good idea, let's do it" haha.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: I renamed some variables and clarified the fields with a comment.

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
flar: I'm accumulating a bunch of changes and got hung up on some of them. I can push them incrementally, but I've renamed this to `IPHalfFractionToFastGaussianCDF` in my local repo.

engine/src/flutter/impeller/compiler/shader_lib/impeller/gaussian.glsl
flar: I also mapped it and the function appears to do the right thing with the way I call it. I adjust [0,1] to [-2,+2] and get the right answers out of it. Here was my graph of the math: https://www.desmos.com/calculator/i4kuidzal5

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: One thing to note, it shows how the mesh differs between the CW and CCW versions of the path (the diagonal lines on the quads reverse).

engine/src/flutter/impeller/display_list/canvas.cc
flar: I removed the type deduction and references, though RoundingRadii is pretty large to copy for this. It is only used to fetch information below. Hmm, it might be cheaper to just call GetRadii in multiple places instead of moving it to a local variable.

flar: Done, no local caching of values any more.

engine/src/flutter/impeller/display_list/canvas.cc
flar: Deleted as well.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: I renamed a lot of variables in this file to eliminate hungarian notation, hopefully I got them all.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: That would be odd. It modifies the centroid variable which is not clear from the new calling convention you suggest. It would be more clear if it was `centroid = FinalizeCentroid(centroid);` as a static function, but its operation is so entwined with the rest of the algorithm that it doesn't make sense to have it be an independent static function.

I don't see a better way here.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: static.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: I moved the test into the struct and commented on why 3 makes sense.

flar: This isn't a tunable characteristic where you might want to change the value it compares against, it is a mathematical axiom so I think a comment on why that number makes sense at the point where we implement the new `IsConvex()` function on the accumulator struct does a better job than simply pulling it out into a separate constant which would also move its definition away from where it is applied.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Already changed when I adjusted the naming and associations with the return value fractions.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Not error handling at all. optional is exactly the right answer here as it is answering both "is there an intersection and if so what are its parameters".

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar:
```
// This method was converted nearly verbatim from the Skia source files
// SkShadowTessellator.cpp and SkPolyUtils.cpp, except for variable
// naming and differences in the methods on Point and Vertex2.
```

I'm keeping this and a couple other methods as close to the Skia version until I have time to circle back and really evaluate the process. Making arbitrary changes makes it harder to verify that we are using the same code Skia used until we have a chance to understand it, make it our own, and document its behavior for future development.

The most useful thing for reviewing this method in the short term is to check it against the Skia version and verify that it is the same code but with minor style differences.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: It's a Scalar, why use an array?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: It's there, the declaration of trigs_ got in the way, so I moved it.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.h
flar: It will be copied into other structures like Geometry/Entity/Contents.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
flar: Most of the unit test files use the entire flutter namespace.

We even have implementation files that do this as well. :(

engine/src/flutter/impeller/geometry/point.h
flar: There is no cc file for point.h (it's empty with a comment), it's a template.
The value of constexpr here is likely negligible, though.

flar:
I've fixed pretty much all of the minor stuff, and the main things left are related to lifetimes of refs and ptrs which I want to rethink. I'm going to push a new commit that gets the minor stuff out of the way while I think about that. I'll leave it to @gaaclarke to "resolve" the comments that I've addressed to his satisfaction.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: But, yes, it is more for visual confirmation that the meshes match what the algorithm is intending to produce...

engine/src/flutter/impeller/display_list/canvas.h
flar: Also, AttemptDrawBlur would rather be a member method since it uses other Canvas data and methods itself, so moving it to an anonymous function in the .cc file would require it to be a friend and take a Canvas instance.

If it is a private method then it needs a parameter that is the BlurShape class and so the implementors of that class need to also be considered members of the class...

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -17,46 +17,57 @@ namespace testing {
...
   }
 
+  /// Reflect a path horizontally and vertically around the given coordinate.
   static PathReflector ReflectAround(const Point& anchor) {
     return PathReflector(-1.0f, anchor.x * 2.0f, -1.0f, anchor.y * 2.0f);
   }
 
   // |PathReceiver|
-  virtual void MoveTo(const Point& p2, bool will_be_closed) {
+  virtual void MoveTo(const Point& p2, bool will_be_closed) override {
```
gaaclarke: nit: "virtual" is redundant once you have "override"

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
gaaclarke: Yea, I was just wondering if it made more sense to assert in code versus doing it visually.  I'm not recommending it, just posing the question.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -21,50 +21,211 @@ using impeller::Tessellator;
 using impeller::Trig;
 using impeller::Vector2;
 
+// Each point in the polygon form of the path is turned into a structure
+// that tracks the gradient of the shadow at that point in the path. The
```
gaaclarke: These should be doxygen comments (3 forward slashes)

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -21,50 +21,211 @@ using impeller::Tessellator;
...
+    kNonConvex,
+    kMultipleContours,
+  };
+
+  UmbraPinAccumulator() = default;
+  ~UmbraPinAccumulator() = default;
+
+  // Reserve enough pins for the indicated number of path vertices to
+  // avoid having to grow the vector during processing.
+  void reserve(size_t vertex_count) { pins_.reserve(vertex_count); }
```
gaaclarke: Method names should start with a capital letter.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -21,50 +21,211 @@ using impeller::Tessellator;
...
+
+  // Return the status properties of the path.
+  PathStatus GetStatus() { return GetResults().status; }
+
+  // Returns a reference to the accumulated vector of UmbraPin structs.
+  // Only valid if the status is kConvex.
+  std::vector<UmbraPin>& GetPins() { return pins_; }
```
gaaclarke, line 188: `const std::vector<UmbraPin>& GetPins() const { return pins_; }`?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: Each ivar should have its own docstring.  If you want to spell out all those details only once just add a docstring that says "see X"

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
```diff
@@ -75,210 +236,62 @@ class PolygonInfo : impeller::PathTessellator::VertexWriter {
...
+  // 2 pins, pin0 and pin1, containing the point of intersection and the
+  // relative fractions at which the 2 pins intersected (expressed as a
+  // ratio of 0 to 1 where 0 represents intersecting at the path outline
+  // and 1 represents intersecting at the tip of the pin where the umbra
+  // is darkest.
   struct PinIntersection {
-    Point intersection;
-    Scalar fraction0;
-    Scalar fraction1;
+    Point intersection;  // Point of the intersection between the pins
```
gaaclarke: Our docstring format (doxygen compatible) has the comment before the ivar.  I think there is a way to add it afterwards, it's something l ike `///<`?

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: This still has the old name.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: Sounds fine, can you add a comment then in the meantime?

engine/src/flutter/impeller/display_list/canvas.cc
gaaclarke: Yea, or a named helper function.  There is no context here that communicates you are converting from a diameter to a radius.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: Got rid of invalidate elsewhere.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
gaaclarke: I mentioned it because there is indexing in the name.  The feedback would be more applicable to the v0 variable.  It wasn't apparent to me that the name of the variable is just the operation that was used to derive it.  Usually we name things what they represent, not how they were calculated.  I'd prefer `v0_length_squared`.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
gaaclarke: leave the campsite cleaner that you found it =)

gaaclarke (commented):
This is looking really close, thanks @flar.  There are a few straggling comment issues and naming issues but otherwise is looking good.  Can you do me a favor and respond to my issues that I haven't marked "resolved".  When there is a big review it's helpful to get some feedback that a comment has been addressed, even if it just says "done".  Thanks.

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: Haha, cut and paste oversight. Done.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Oops. Done.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Can't be const. We're not done with it. The Accumulator class just initialized the structs. The real work is yet to be done. (I originally made it const and got lots of complaints.)

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
flar: I don't want to assert the values of the mesh because it can vary depending on platform (Intel vs ARM). I do assert the number of vertices in the other test file at least, and I had to adjust the numbers in one of those tests because Linux was producing one fewer vertex on the corner fans so I made the numbers not be so close to the cutoffs.

flar: This is pretty optional, but it helps to verify that our meshes "look right".

engine/src/flutter/impeller/display_list/canvas.cc
flar: I commented it.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Completely rewritten in the latest code. The new helper class is auto-finalizing and it returns the data in a Results struct that is <optional> until finalized.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: I think I got them all now.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Now named consistent with the pin0/pin1 at least.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc
flar: Stale code, but I'll try to update ivar comments.

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
flar: I found that one of the biggest blocks in the logic cannot be executed because it is implemented in a "general polygon" module that doesn't realize that the polygon was normalized by the calling code. I haven't proven that it can no longer be executed, so I added a big comment about the issue to the block. I plan to come back and delete the block later when I explore the 2 methods that I left "as is" from the Skia code.

flar: Other than that, I surmised all of the ways in which collinear points can mess with the path and decided to break the path analysis into a 2 step process for simplicity. First the code receives the points from the Source and normalizes the points to the grid and removes duplicates and collinear points. Then a second pass very efficiently runs through the normalized points and computes their data once all of the duplication and collinear points have been removed. I then added some more test cases that verify that we aren't thrown off by extra collinear points.

It's feeling pretty well tested at this point.

What remains is that the algorithm (verified that the Skia version of the code also does this) can throw up its hands on some valid convex paths and fail to produce an umbra ring. I plan to come back, figure out why that is, fix it, and then add more tests for those conditions. For now I focused on "negative" tests (i.e. we don't produce a mesh for paths that shouldn't have a mesh).

engine/src/flutter/impeller/entity/geometry/shadow_path_geometry_unittests.cc
flar: Did it and realized it was only 4 classes after all of that.

gaaclarke (approved):
lgtm, thanks jim

flar:
I reran my blur size comparison tests and I'm happy with the consistency of the output between the optimized special-case maskblur implementations (the old AttemptRRectBlurShape which is still used for basic rects and rrects), the general case maskblur (i.e. uses a GaussianBlur ImageFilter for any kind of shape) and the new general convex path shader (this PR).

engine/src/flutter/impeller/display_list/aiks_dl_shadow_unittests.cc
```diff
@@ -99,18 +99,28 @@ void DrawShadowMesh(DisplayListBuilder& builder,
...
-  DlPaint paint;
-  paint.setDrawStyle(DlDrawStyle::kStroke);
   bool should_optimize = path.IsConvex();
-  Point shadow_translate;
-  Point path_translate = Point(0, elevation * dpr * 0.5f);
+  Matrix matrix = builder.GetMatrix();
+
+  // From dl_dispatcher, making a MaskFilter.
+  Scalar light_radius = 800 / 600;
+  EXPECT_EQ(light_radius, 1.0f);  // Value in dl_dispatcher is bad.
```
gaaclarke, line 107: What's this?  Integer division that is cast to a float, why?

flar: Yep. I should file a separate issue. It's been there since that code was first written.

flar: https://github.com/flutter/flutter/issues/180238

PR #178376, opened 2025-11-12
Make sure that a Theme doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the Theme widget.


victorsanni (commented):
Theme doesn't render anything I think, do we need a 0x0 test here? @dkwingsmt

ahmedsameha1:
> Theme doesn't render anything I think, do we need a 0x0 test here? @dkwingsmt

And what about TooltipVisibility?

PR #178461, opened 2025-11-13
Make sure that a Tooltip doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the Tooltip widget.


packages/flutter/test/material/tooltip_test.dart
```diff
@@ -3663,6 +3663,23 @@ void main() {
...
+      MaterialApp(
+        home: Center(
+          child: SizedBox.shrink(
+            child: Tooltip(key: key, message: 'X'),
+          ),
+        ),
+      ),
+    );
+    expect(tester.getSize(find.byType(Tooltip)), Size.zero);
+    key.currentState!.ensureTooltipVisible();
```
victorsanni, line 3679: Should we check that the tooltip is of size zero after ensuring it is visible?

victorsanni (approved):

packages/flutter/test/material/tooltip_test.dart
dkwingsmt: I suggest we should use set the view size to zero since the tooltip is an overlay.

victorsanni (approved):

dkwingsmt (approved):

packages/flutter/test/material/tooltip_test.dart
```diff
@@ -3663,6 +3663,24 @@ void main() {
     expect(tooltipPosition.dx, closeTo(targetCenter.dx + 25, 5.0));
     expect(tooltipPosition.dy, closeTo(targetCenter.dy - 25 - 32, 5.0));
   });
+
+  testWidgets('Tooltip does not crash at zero area', (WidgetTester tester) async {
+    final GlobalKey<TooltipState> key = GlobalKey<TooltipState>();
```
victorsanni:
```suggestion
    final key = GlobalKey<TooltipState>();
```

victorsanni (approved):

Issue #178395, opened 2025-11-12
12% regression in `dynamic_path_tessellation`


gaaclarke:
introduced in https://github.com/flutter/flutter/pull/178142

https://flutter-flutter-perf.luci.app/e/?begin=1762480945&end=1762954756&keys=Xf6704922d5ba2f7983cab8113fc69582&num_commits=50&request_type=0&xbaroffset=2805065

**sub_result:** `average_frame_build_time_millis`
**test:** `dynamic_path_tessellation_perf_ios__timeline_summary`
**device:** `iPhone_16_Pro`

cc @jason-simmons, I'm guessing some of this was expected?  I just wanted to file this to make sure someone who was involved can give it the green light or not.

jason-simmons:
The new Skia immutable path API has some unavoidable cost.  The `CanvasPath` object wrapped by the Dart `Path` now holds an `SkPathBuilder`.  When the path is added to a display list, the `SkPathBuilder` contents now need to be copied into an immutable `SkPath` that is used to construct the `DlPath`.

(After the `CanvasPath` builds the `DlPath`, it will not be rebuilt unless the `CanvasPath` is later modified)

In the old API where `SkPath` was mutable, the `SkPath` would be given to the `DlPath` without copying.  The mutable `SkPath` had a copy-on-write implementation where its data would only need to be copied if the app modified it after the `DlPath` was created.

AFAICT there is currently no way to create the `SkPath` without copying the data.  The implementation of `SkPathBuilder::detach` is calling the `SkPathBuilder::snapshot` method that `CanvasPath` is using, which makes a copy.

For the `dynamic_path_tessellation` benchmark, another reason for the performance difference is the benchmark's frequent use of `Path.transform`.  The `CanvasPath::transform` implementation is currently making multiple copies of the path data, and that can be improved.

jason-simmons:
Skia plans to switch to the new `SkPathData` path implementation soon (see https://skia-review.googlesource.com/c/skia/+/1092240)

We should revisit this after that is fully landed.  In particular, it may make sense for Skia to add an implementation of `SkPathBuilder::detach` that moves the path data from the builder to the `SkPath` without a copy.

With that API, Flutter path wrappers like `CanvasPath` could call `detach`, give the resulting immutable `SkPath` to the display list, and also retain a local clone of the `SkPath`.  If the user later modifies the `CanvasPath`, then `CanvasPath` can reinitialize the `SkPathBuilder` using the `SkPath`.

[...]

PR #178512, opened 2025-11-14
Reduce the data copying in CanvasPath related to the SkPathBuilder API migration

jason-simmons:
* In CanvasPath::transform/shift, copy the source path into the target SkPathBuilder and then execute the transform within the SkPathBuilder
* Obtain SkPath snapshots through CanvasPath::path, which will cache the SkPath within a DlPath

See https://github.com/flutter/flutter/issues/178395


gaaclarke (approved):
lgtm!

Issue #177822, opened 2025-10-31
[Windows] HardwareKeyboard misses next key press after alt is pressed


josh-audio:
### Steps to reproduce

These steps use Windows 11 with Flutter 3.35.7 (see doctor output below).

1. Create a new Flutter project with `flutter create`
2. In `main.dart`, add the following in `main()` after `runApp()`:
   ```dart
   HardwareKeyboard.instance.addHandler((event) {
     if (event is KeyDownEvent) {
       print('Key pressed: ${event.logicalKey.debugName}');
     }
     return false;
   });
   ```
3. Run the project
4. Try the following two things:
   1. Press the left alt key 4 times
   2. Press alt, then release alt, then press another key

### Expected results

I expect an output like the following:

```
(for the first)
Key pressed: Alt Left
Key pressed: Alt Left
Key pressed: Alt Left
Key pressed: Alt Left

(for the second)
Key pressed: Alt Left
Key pressed: Key A
```

### Actual results

Instead, I get the following output:

```
(for the first)
Key pressed: Alt Left
Key pressed: Alt Left

(for the second)
Key pressed: Alt Left
```

Note that this does not happen for other keys, including other modifiers like shift and control.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';
import 'package:flutter/services.dart';

void main() {
  runApp(const MyApp());

  HardwareKeyboard.instance.addHandler((event) {
    if (event is KeyDownEvent) {
      print('Key pressed: ${event.logicalKey.debugName}');
    }
    return false;
  });
}

// Rest of file unchanged from flutter create output
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.35.7, on Microsoft Windows [Version 10.0.26200.6899], locale en-US) [411ms]
    • Flutter version 3.35.7 on channel stable at C:\Users\qbgee\Documents\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision adc9010625 (9 days ago), 2025-10-21 14:16:03 -0400
    • Engine revision 035316565a
    • Dart version 3.9.2
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop,
      enable-android, enable-ios, cli-animations, enable-lldb-debugging

[...]

PR #178523, opened 2025-11-14
Fix Windows Enter key after focus loss

letrungdo:
Fix issue: https://github.com/flutter/flutter/issues/169447

Problem: On Windows, pressing Enter after refocusing the app could drop the first keydown because the embedder still thought the key was pressed, causing a double-press requirement and HardwareKeyboard assertions.
Fix: In KeyboardKeyEmbedderHandler, synthesize a key-up whenever Win32 reports a “fresh” down for a key still tracked as pressed, so Flutter sees a release before the new down. This keeps modifier/lock-key synchronization intact.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/windows/keyboard_key_embedder_handler.cc
+++ b/engine/src/flutter/shell/platform/windows/keyboard_key_embedder_handler.cc
@@ -225,22 +225,24 @@ void KeyboardKeyEmbedderHandler::KeyboardHookImpl(
   uint64_t result_logical_key;
 
   if (is_event_down) {
+    if (had_record && !was_down) {
+      // Windows delivered a new down event without a matching up event. This
+      // happens if the window lost focus before it could receive the key up.
+      // Synthesize an up event so the framework releases the key before
+      // processing the incoming down.
+      SendSynthesizeUpEvent(physical_key, last_logical_record);
+      last_logical_record_iter = pressingRecords_.find(physical_key);
+      had_record = last_logical_record_iter != pressingRecords_.end();
+      last_logical_record = had_record ? last_logical_record_iter->second : 0;
+    }
+
     if (had_record) {
-      if (was_down) {
-        // A normal repeated key.
-        type = kFlutterKeyEventTypeRepeat;
-        FML_DCHECK(had_record);
-        ConvertUtf32ToUtf8_(character_bytes, character);
-        eventual_logical_record = last_logical_record;
-        result_logical_key = last_logical_record;
-      } else {
-        // A non-repeated key has been pressed that has the exact physical key
-        // as a currently pressed one, usually indicating multiple keyboards are
-        // pressing keys with the same physical key, or the up event was lost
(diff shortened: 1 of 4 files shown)


engine/src/flutter/shell/platform/windows/keyboard_key_embedder_handler.cc
```diff
@@ -225,22 +225,24 @@ void KeyboardKeyEmbedderHandler::KeyboardHookImpl(
   uint64_t result_logical_key;
 
   if (is_event_down) {
+    if (had_record && !was_down) {
+      // Windows delivered a new down event without a matching up event. This
+      // happens if the window lost focus before it could receive the key up.
+      // Synthesize an up event so the framework releases the key before
+      // processing the incoming down.
```
mattkae, line 232: Thanks for doing this work! This might even affect a bug that I'm dealing with in the alt tab switcher.

I have a question though: Would it be better to move this logic to when the window gains focus instead? in pseudocode, it would be:

```
if window_gaining_focus:
   if had_key_down:
      synthesize_matching_up_event()
```

My reasoning for this is that the current implementation has to wait for another down event to be clicked to synthesize the up event for the previous key, which may add an unnecessary delay and may be incorrect in some cases (e.g. if "alt" is held and you return back to the app before clicking another key, you may move your mouse with a modifier held. This might mean something for certain apps).

Just my two cents, let me know what you think!

mattkae (commented):
A thought, let me know what you think!

engine/src/flutter/shell/platform/windows/keyboard_key_embedder_handler.cc
letrungdo: Thanks for the idea.
I considered doing it on focus, but there are a couple of reasons I kept it in the key path:
KeyboardKeyEmbedderHandler isn’t told about focus changes, so "release on focus" would need new plumbing from FlutterWindow/WindowsLifecycleManager into the handler. I wanted the minimal, targeted fix.
Synthesizing ups on focus can be wrong if the OS still considers the key held (e.g., you alt‑tab back while still holding Alt). We’d force-release a modifier that the user is actually still holding, and any subsequent real keyup would then look out of order.
By deferring to the next real key event, we only synthesize when Windows tells us a fresh down is happening without a corresponding up, which aligns with the actual input stream.

mattkae: Understandable 👍 

My only worry with this implementation is that the enter key _could_ still be considered held while we mouse over widgets after receiving focus. This _might_ mean something to some apps.

At any rate, we can address that in the future, as this PR does fix the most common use case and doesn't make the case that I outlined above any worse.

mattkae (approved):
The fix looks good to me, thanks! I think it will solve my issues with https://github.com/flutter/flutter/issues/177822 as well, but I will have to see (next week probably)

PR #178015, opened 2025-11-05
[Android] Encode the original pointer count in messages that represent Android touch events

jason-simmons:
Android touch events include updates to multiple pointers, but each pointer data message sent from the embedder to the framework represents a single pointer.  So the Android embedder will send multiple messages for each touch event, and the framework's AndroidViewController will reassemble the messages and forward the resulting event to the platform view.

The AndroidViewController tracks the number of active pointers in its own local state.  If that state is out of sync with the event handled by the Android embedder, then the AndroidViewController may send duplicate events to the platform view.

This PR encodes the Android touch event's pointer count in the messages sent to the framework.  This allows the AndroidViewController to reliably determine whether it has received all of the pointer messages that originated from an event.

Fixes https://github.com/flutter/flutter/issues/176574

PR #178528, opened 2025-11-14
fix: check both pointer count and action before reusing MotionEvent

letrungdo:
Fix issue https://github.com/flutter/flutter/issues/169486

## Problem

After [PR #178015](https://github.com/flutter/flutter/pull/178015) was merged, Android platform views experience gesture blocking after multi-touch operations (e.g., pinch-to-zoom). The issue occurs because:

1. PR #178015 changed framework behavior to encode original pointer count in events
2. Framework now sends events with **different actions** than the original MotionEvent
   - Example: Framework sends `ACTION_MOVE (2)` while original event is `ACTION_POINTER_UP (6)`
3. The existing engine code only checked **pointer count**, not action
4. This caused the engine to use original event with wrong action → gesture blocked

### Logs Showing the Issue

trackedEvent.pointerCount=2, touch.pointerCount=2, trackedEvent.action=6 (ACTION_POINTER_UP), touch.action=2 (ACTION_MOVE) → Pointer counts MATCH, using original event → WebView receives ACTION_POINTER_UP but expects ACTION_MOVE → Gesture blocked! ❌


gmackall (commented):
Can I ask what behavior you expect to see? I no longer see the behavior from https://github.com/flutter/flutter/pull/178528 that your original PR intended to fix (the platform view becoming unresponsive) on master. I'm not sure I can see a behavior difference when testing your app on this branch vs master:

on your branch:
https://github.com/user-attachments/assets/47fe92b4-e71f-4830-8cb1-1e808512dd7e

on master:
https://github.com/user-attachments/assets/8f7fb804-7c78-4522-9763-4980c26523f9

Naively, they both look correct to me (they discard multi touch gestures, as one would expect from a `SingleTouchContainer` and otherwise don't become unresponsive).

I also am doubtful that https://github.com/flutter/flutter/pull/178015 introduced the action mismatch you are highlighting here. Anecdotally I had noticed that this bug existed before either your original PR or @jason-simmons's pr had landed. 

Given that your original PR still used the action from the tracked event when making the reconstructed event, I would expect no effect from @jason-simmons's PR on the behavior you see with regard to action delivered.

gmackall (approved):
Ok it turns out this actually resolves https://github.com/flutter/flutter/issues/179275! 

I also wanted to double check if we could, for the case where only the actions mismatch, call `MotionEvent.setAction()` and maintain verifiability. But it turns out that method renders the event not verified, so it seems pointer mismatch and action mismatch are the two cases where we can't fix the event and thus need to construct a new event. So this LGTM.

jason-simmons (approved):

PR #178550, opened 2025-11-14
[ Widget Preview ] Gracefully handle unexpected analysis context disposal

bkonyi:
It's possible that the preview detector is in the middle of processing sources after a source file was changed while the widget previewer is shutting down if we're not holding the mutex. This can result in the analysis context being disposed of when we don't expect it, causing requests to the analyzer to return errors.

This change guards against this scenario by ensuring that all calls to `findPreviewFunctions` are guarded by the mutex. This change also removes assumptions related to the return type from analyzer APIs to account for the possibility that an error response was returned.

Fixes https://github.com/flutter/flutter/issues/178472


jyameo (approved):
Looks great! Thanks Ben!

PR #169715, opened 2025-05-29
Adjust handle anchor offset for collapsed text selection for Android

muhammadkamel:
The text selection anchor/handle in InputField was slightly misaligned, shifting slightly to the left. This PR fixes the handle positioning to ensure proper alignment with the text cursor.

Before:
baloon_before

After fix:
baloon_after_fixed

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #178551, opened 2025-11-14
Fix text selection handle alignment, cursor width updates, and Android handle auto-dismissal like native approach

### LINKED ISSUE #45556: Android Cursor Pointer is off center
## Steps to Reproduce
Steps to reproduce 
1. create new project flutter create -i swift -a kotlin cursor_pointer
2. add a text field 

<details>

```import 'package:flutter/material.dart';

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        //
        // Try running your application with "flutter run". You'll see the
        // application has a blue toolbar. Then, without quitting the app, try
        // changing the primarySwatch below to Colors.green and then invoke
        // "hot reload" (press "r" in the console where you ran "flutter run",
        // or simply save your changes to "hot reload" in a Flutter IDE).
        // is not restarted.
        primarySwatch: Colors.blue,
      ),
      home: MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

  MyHomePage({Key key, this.title}) : super(key: key);

  // how it looks.

  // always marked "final".

  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });
  }

  @override
    //
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          //
          // Invoke "debug painting" (press "p" in the console, choose the
          // "Toggle Debug Paint" action from the Flutter Inspector in Android
          // Studio, or the "Toggle Debug Paint" command in Visual Studio Code)
          // to see the wireframe for each widget.
          //
          // horizontal).
          children: <Widget>[
            Text(
            ),
            Text(
              '$_counter',
              style: Theme.of(context).textTheme.display1,
            ),
            TextField(),
          ],
        ),
      ),

        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: Icon(Icons.add),
    );
  }
}
```

</details>

Screenshot_1574707239

Occurs on all android devices

<details>
  <summary>logs</summary>

```
jasons-mbp:cursor_pointer jasonzheng$ flutter run --verbose
[  +22 ms] executing: [/Volumes/Data/Sources/flutter/] git -c log.showSignature=false log -n 1 --pretty=format:%H
[  +42 ms] Exit code 0 from: git -c log.showSignature=false log -n 1 --pretty=format:%H
[        ] e70236e36ce1d32067dc68eb55519ec3e14b6b01
[        ] executing: [/Volumes/Data/Sources/flutter/] git describe --match v*.*.* --first-parent --long --tags
[  +19 ms] Exit code 0 from: git describe --match v*.*.* --first-parent --long --tags
[        ] v1.10.7-0-ge70236e36
[   +7 ms] executing: [/Volumes/Data/Sources/flutter/] git rev-parse --abbrev-ref --symbolic @{u}
[  +13 ms] Exit code 0 from: git rev-parse --abbrev-ref --symbolic @{u}
[        ] origin/beta
[        ] executing: [/Volumes/Data/Sources/flutter/] git ls-remote --get-url origin
[  +11 ms] Exit code 0 from: git ls-remote --get-url origin
[        ] https://github.com/flutter/flutter.git
[  +62 ms] executing: [/Volumes/Data/Sources/flutter/] git rev-parse --abbrev-ref HEAD
[  +12 ms] Exit code 0 from: git rev-parse --abbrev-ref HEAD
[        ] beta
[  +14 ms] executing: sw_vers -productName
[  +27 ms] Exit code 0 from: sw_vers -productName
[        ] Mac OS X
[        ] executing: sw_vers -productVersion
[  +23 ms] Exit code 0 from: sw_vers -productVersion
[        ] 10.14.6
[        ] executing: sw_vers -buildVersion
[  +26 ms] Exit code 0 from: sw_vers -buildVersion
[        ] 18G103
[  +37 ms] executing: /usr/bin/xcode-select --print-path
[  +17 ms] Exit code 0 from: /usr/bin/xcode-select --print-path
[        ] /Applications/Xcode.app/Contents/Developer
[   +1 ms] executing: /usr/bin/xcodebuild -version
[ +125 ms] Exit code 0 from: /usr/bin/xcodebuild -version
[   +3 ms] Xcode 11.2.1
           Build version 11B500
[  +63 ms] executing: /Users/jasonzheng/Library/Android/sdk/platform-tools/adb devices -l
[   +8 ms] Exit code 0 from: /Users/jasonzheng/Library/Android/sdk/platform-tools/adb devices -l
[        ] List of devices attached
           emulator-5554          device product:sdk_gphone_x86 model:Android_SDK_built_for_x86 device:generic_x86 transport_id:5
[  +19 ms] executing: /Volumes/Data/Sources/flutter/bin/cache/artifacts/libimobiledevice/idevice_id -h
[  +51 ms] /usr/bin/xcrun simctl list --json devices
[ +150 ms] More than one device connected; please specify a device with the '-d <deviceId>' flag, or use '-d all' to act on all
devices.
[   +5 ms] /Users/jasonzheng/Library/Android/sdk/platform-tools/adb -s emulator-5554 shell getprop
[  +38 ms] ro.hardware = ranchu
[   +5 ms] Android SDK built for x86 • emulator-5554                        • android-x86 • Android 9 (API 28) (emulator)
[   +1 ms] iPhone 8                  • 80FF753B-2016-49AD-B175-4663153BAB6D • ios         •
com.apple.CoreSimulator.SimRuntime.iOS-13-2
           (simulator)
[  +18 ms] "flutter run" took 650ms.

#0      throwToolExit (package:flutter_tools/src/base/common.dart:28:3)
#1      RunCommand.validateCommand (package:flutter_tools/src/commands/run.dart:276:7)
<asynchronous suspension>
#2      FlutterCommand.verifyThenRunCommand (package:flutter_tools/src/runner/flutter_command.dart:528:11)
<asynchronous suspension>
#3      FlutterCommand.run.<anonymous closure> (package:flutter_tools/src/runner/flutter_command.dart:457:33)
<asynchronous suspension>
#4      AppContext.run.<anonymous closure> (package:flutter_tools/src/base/context.dart:157:29)
<asynchronous suspension>
#5      _rootRun (dart:async/zone.dart:1124:13)
#6      _CustomZone.run (dart:async/zone.dart:1021:19)
#7      _runZoned (dart:async/zone.dart:1516:10)
#8      runZoned (dart:async/zone.dart:1463:12)
#9      AppContext.run (package:flutter_tools/src/base/context.dart:156:18)
<asynchronous suspension>
#10     FlutterCommand.run (package:flutter_tools/src/runner/flutter_command.dart:446:20)
#11     CommandRunner.runCommand (package:args/command_runner.dart:197:27)
<asynchronous suspension>
#12     FlutterCommandRunner.runCommand.<anonymous closure> (package:flutter_tools/src/runner/flutter_command_runner.dart:416:21)
<asynchronous suspension>
#13     AppContext.run.<anonymous closure> (package:flutter_tools/src/base/context.dart:157:29)
<asynchronous suspension>
#14     _rootRun (dart:async/zone.dart:1124:13)
#15     _CustomZone.run (dart:async/zone.dart:1021:19)
#16     _runZoned (dart:async/zone.dart:1516:10)
#17     runZoned (dart:async/zone.dart:1463:12)
#18     AppContext.run (package:flutter_tools/src/base/context.dart:156:18)
<asynchronous suspension>
#19     FlutterCommandRunner.runCommand (package:flutter_tools/src/runner/flutter_command_runner.dart:367:19)
<asynchronous suspension>
#20     CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:112:25)
#21     new Future.sync (dart:async/future.dart:224:31)
#22     CommandRunner.run (package:args/command_runner.dart:112:14)
#23     FlutterCommandRunner.run (package:flutter_tools/src/runner/flutter_command_runner.dart:251:18)
#24     run.<anonymous closure>.<anonymous closure> (package:flutter_tools/runner.dart:63:22)
<asynchronous suspension>
#25     _rootRun (dart:async/zone.dart:1124:13)
#26     _CustomZone.run (dart:async/zone.dart:1021:19)
#27     _runZoned (dart:async/zone.dart:1516:10)
#28     runZoned (dart:async/zone.dart:1500:12)
#29     run.<anonymous closure> (package:flutter_tools/runner.dart:61:18)
<asynchronous suspension>
#30     AppContext.run.<anonymous closure> (package:flutter_tools/src/base/context.dart:157:29)
<asynchronous suspension>
#31     _rootRun (dart:async/zone.dart:1124:13)
#32     _CustomZone.run (dart:async/zone.dart:1021:19)
#33     _runZoned (dart:async/zone.dart:1516:10)
#34     runZoned (dart:async/zone.dart:1463:12)
#35     AppContext.run (package:flutter_tools/src/base/context.dart:156:18)
<asynchronous suspension>
#36     runInContext (package:flutter_tools/src/context_runner.dart:63:24)
<asynchronous suspension>
#37     run (package:flutter_tools/runner.dart:50:10)
#38     main (package:flutter_tools/executable.dart:65:9)
<asynchronous suspension>
#39     main (file:///Volumes/Data/Sources/flutter/packages/flutter_tools/bin/flutter_tools.dart:8:3)
#40     _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:303:32)
#41     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:172:12)

jasons-mbp:cursor_pointer jasonzheng$ flutter analyze
Analyzing cursor_pointer...                                             
No issues found! (ran in 2.8s)

```

</details>

```
Jasons-MacBook-Pro:flutter jasonzheng$ flutter doctor -v
[✓] Flutter (Channel beta, v1.11.0, on Mac OS X 10.14.6 18G103, locale en-US)
    • Flutter version 1.11.0 at /Volumes/Data/Sources/flutter
    • Framework revision 856a90e67c (2 weeks ago), 2019-11-08 18:00:01 -0800
    • Engine revision af04338413
    • Dart version 2.7.0

 
[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /Users/jasonzheng/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling
      support)
    • Platform android-28, build-tools 28.0.3
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build
      1.8.0_152-release-1343-b01)

[✓] Xcode - develop for iOS and macOS (Xcode 11.2.1)
    • Xcode 11.2.1, Build version 11B500
    • CocoaPods version 1.8.4

[✓] Android Studio (version 3.4)
    • Flutter plugin version 35.2.1
    • Dart plugin version 183.6270
    • Java version OpenJDK Runtime Environment (build
      1.8.0_152-release-1343-b01)

[✓] VS Code (version 1.40.1)
    • Flutter extension version 3.6.0

    • Android SDK built for x86 • emulator-5554                        •
      android-x86 • Android 9 (API 28) (emulator)
    • iPhone 8                  • 80FF753B-2016-49AD-B175-4663153BAB6D • ios
      • com.apple.CoreSimulator.SimRuntime.iOS-13-2 (simulator)

• No issues found!
```

muhammadkamel:
This is a new PR to replace the previously closed PR: https://github.com/flutter/flutter/pull/169715

The text selection anchor/handle in InputField was slightly misaligned, shifting slightly to the left. This PR fixes the handle positioning to ensure proper alignment with the text cursor.

Cursor Width Updates
Previously, updating cursorWidth in EditableText did not reflect immediately without a hot restart. This PR updates EditableTextState.didUpdateWidget to detect cursorWidth changes and rebuild the selection overlay.

Android Handle Auto-Dismissal
On Android, text selection handles now automatically dismiss after 4 seconds of inactivity (specifically when the selection is collapsed), matching native Android behavior. A strict unit test ensures this timeout is respected.

Before:
baloon_before

After fix:
baloon_after_fixed

Fixes #45556 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/cupertino/text_selection.dart
```diff
@@ -155,7 +155,11 @@ class CupertinoTextSelectionControls extends TextSelectionControls {
   ///
   /// See [TextSelectionControls.getHandleAnchor].
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    required double cursorWidth,
+  }) {
```
justinmc, line 159: We should watch out for breaking changes due to changing this method signature.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -447,14 +464,43 @@ class TextSelectionOverlay {
     _updateTextSelectionOverlayVisibilities();
   }
 
+  void _disposeHandlesDismissalTimer() {
+    _handlesDismissalTimer?.cancel();
+    _handlesDismissalTimer = null;
+  }
+
+  // The duration for auto-dismissal of handles on Android.
+  static const Duration _androidHandlesDismissalDuration = Duration(seconds: 4);
```
justinmc, line 566: Could you add a sentence explaining where you got the 4 seconds from? Like "Measured on a physical Pixel 9 running Android 16." or whatever.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -447,14 +464,43 @@ class TextSelectionOverlay {
...
+
+  void _scheduleHandlesDismissal() {
+    _disposeHandlesDismissalTimer();
+    // Android treats specific handle visibility differently.
+    // If the selection is collapsed (cursor), handles should auto-dismiss.
```
justinmc, line 571: Thanks for catching this. I see the same thing on my Android 16 phone.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1254,6 +1308,11 @@ class SelectionOverlay {
   /// Called when the users start dragging the start selection handles.
   final ValueChanged<DragStartDetails>? onStartHandleDragStart;
 
+  /// Cursor width
+  /// This value is used for calculating the position of the text selection
+  /// handles.
```
justinmc: Period after "width", and also add an empty line of "///".

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1899,6 +1940,54 @@ void main() {
...
+    ); // End of 'A'
+    final Offset caretCenterGlobal = renderEditable.localToGlobal(caretRect.center);
+
+    // Now find the handle.
+    // In Material, the collapsed handle uses _TextSelectionHandlePainter.
+    final Finder handleFinder = find
+        .byType(CustomPaint)
+        .last; // Usually the last one drawn in the overlay.
+    // To be safer, we can check the painter type if we could access the private class, but we can't easily.
+    // Let's assume the handle is the one that is visible and in the overlay.
```
justinmc: You can search for private widgets:

```
find.byWidgetPredicate((Widget w) => '${w.runtimeType}' == '_MyPrivateWidget')
```

But _TextSelectionHandlePainter is not a widget...

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1899,6 +1940,54 @@ void main() {
...
+
+    final RenderBox handleBox = tester.renderObject(handleFinder);
+    final Offset handleCenter = handleBox.localToGlobal(handleBox.size.center(Offset.zero));
+
+    // The handle's center x should match the caret's center x.
+    expect(handleCenter.dx, closeTo(caretCenterGlobal.dx, 0.5));
+
+    // Also verify specifically that the visual center of the handle (which is the "onion" shape) points to the caret.
+    // The handle logic in getHandleAnchor should have compensated for the handle size and cursor width.
```
justinmc: Did you intend to add more code to verify this?

justinmc (commented):
Thank you for being persistent with this change! I think it looks like a great improvement overall. We'll have to make sure that it doesn't break any customers. Can you fix the analyzer error?

packages/flutter/lib/src/widgets/text_selection.dart
muhammadkamel: This value follows the native Android behavior. In native implementations (Kotlin/Java), the selection handle is automatically dismissed after roughly 4–5 seconds. This was measured on a physical Android device and matches the behavior seen across native apps.

dev/integration_tests/windowing_test/lib/main.dart
```diff
@@ -10,6 +10,7 @@ import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+// ignore: implementation_imports
```
Renzo-Olivares: Why is this change needed?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1254,6 +1312,12 @@ class SelectionOverlay {
   /// Called when the users start dragging the start selection handles.
   final ValueChanged<DragStartDetails>? onStartHandleDragStart;
 
+  /// Cursor width.
```
Renzo-Olivares: I think we can just reference `EditableText.cursorWidth` for this first line of the documentation.

```dart
/// {@macro flutter.widgets.editableText.cursorWidth}
```

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,15 @@ abstract class TextSelectionControls {
   /// Get the anchor point of the handle relative to itself. The anchor point is
   /// the point that is aligned with a specific point in the text. A handle
   /// often visually "points to" that location.
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight);
+  ///
+  /// The [cursorWidth] argument is the thickness of the cursor, which may be
+  /// used to offset the anchor point.
```
Renzo-Olivares: After the last sentence here, we should also include a bit about the default value being 2.0.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -340,6 +352,7 @@ class TextSelectionOverlay {
     required LayerLink endHandleLayerLink,
     required this.renderObject,
     this.selectionControls,
+    required this.cursorWidth,
```
Renzo-Olivares: I think it makes sense to make this required but it might be a breaking change.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -400,6 +414,10 @@ class TextSelectionOverlay {
   /// {@macro flutter.widgets.SelectionOverlay.selectionControls}
   final TextSelectionControls? selectionControls;
 
+  /// The width of the cursor.
```
Renzo-Olivares: This documentation should instead reference the `SelectionOverlay.cursorWidth` documentation through a doc template. Similar to `selectionControls` above.

packages/flutter/lib/src/material/selection_area.dart
```diff
@@ -54,6 +54,7 @@ class SelectionArea extends StatefulWidget {
     this.contextMenuBuilder = _defaultContextMenuBuilder,
     this.magnifierConfiguration,
     this.onSelectionChanged,
+    this.cursorWidth = 2.0,
```
Renzo-Olivares: I'm a bit weary about adding this to `SelectionArea` since i'm not sure how it will be used. A collapsed selection should show no cursor at all on static text. Though I think the reason it is being added is because `TextSelectionOverlay/SelectionOverlay.cursorWidth` is a required argument.

I see both sides:

It is important for `TextSelectionOverlay/SelectionOverlay.cursorWidth` to be a required argument so it can be taken into account when calculating the handle anchor. Making it nullable would mean if a user does not provide it they may not get an accurate handle anchor. Though in the case of `SelectableRegion` and static text we do not show a cursor with a collapsed selection, so adding this parameter is not very useful at the moment.

I think maybe we should hold off on changes to `SelectionArea` and `SelectableRegion` for now since the cursor is not used with static text. We can do this by keeping `TextSelectionOverlay.cursorWidth` required but making `SelectionOverlay.cursorWidth` nullable.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,15 @@ abstract class TextSelectionControls {
...
   /// often visually "points to" that location.
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight);
+  ///
+  /// The [cursorWidth] argument is the thickness of the cursor, which may be
+  /// used to offset the anchor point.
+  ///
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    double cursorWidth = 2.0,
```
Renzo-Olivares, line 134: I'm a bit on the fence about having a default value for this. If someone uses this method, has a custom cursorWidth, and does not provide their cursorWidth, this will default to 2.0 and the handle will be misaligned. Another option would be to make it required, which might be a breaking change but would make the handle alignment fixed for all users. Making it nullable be similar to having a default value, if the user chooses not to provide it then the handle will remain misaligned. What do others think about these options?

Renzo-Olivares (commented):
@muhammadkamel thank you for your patience and contribution! This is a great fix and adds a nice feature and improves native fidelity! I left a few comments regarding the surface of this change and thoughts regarding cursorWidth having a default value. Let me know what you think.

dev/integration_tests/windowing_test/lib/main.dart
muhammadkamel: @Renzo-Olivares 

I have done this to fix the flutter analyze warnings
But I can revert it if you want

muhammadkamel:
@Renzo-Olivares Thank you for the thorough review!

Could you please review new changes now.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -400,6 +415,9 @@ class TextSelectionOverlay {
   /// {@macro flutter.widgets.SelectionOverlay.selectionControls}
   final TextSelectionControls? selectionControls;
 
+  /// {@macro flutter.widgets.TextSelectionControls.getHandleAnchor.cursorWidth}
+  final double cursorWidth;
```
Renzo-Olivares: I think instead of exposing `cursorWidth` on `TextSelectionOverlay`, `TextSelectionOverlay` should just adhere to the `cursorWidth` in its already provided `renderObject`.

It feels a little strange that `TextSelectionOverlay/SelectionOverlay` exposes `cursorWidth` when it is not the one responsible for painting the cursor. Maybe it needs a more general name related to "anchor".

dev/integration_tests/windowing_test/lib/main.dart
Renzo-Olivares: That's surprising given no other changes in this file. This PR may just need a rebase to master.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -400,6 +414,11 @@ class TextSelectionOverlay {
   /// {@macro flutter.widgets.SelectionOverlay.selectionControls}
   final TextSelectionControls? selectionControls;
 
+  /// {@macro flutter.widgets.TextSelectionControls.getHandleAnchor.cursorWidth}
+  ///
+  /// This value is read directly from [renderObject.cursorWidth].
+  double get cursorWidth => renderObject.cursorWidth;
```
Renzo-Olivares: I don't think this public getter is needed anymore.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1254,6 +1313,12 @@ class SelectionOverlay {
   /// Called when the users start dragging the start selection handles.
   final ValueChanged<DragStartDetails>? onStartHandleDragStart;
 
+  /// {@macro flutter.widgets.TextSelectionControls.getHandleAnchor.cursorWidth}
+  ///
+  /// This value is used for calculating the position of the text selection
+  /// handles.
+  final double? cursorWidth;
```
Renzo-Olivares: This should have a getter and setter similar to `SelectionOverlay.lineHeightAtStart`, this way `TextSelectionOverlay` can update it in `TextSelectionOverlay._updateSelectionOverlay`.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -2056,6 +2148,11 @@ class FakeRenderEditable extends RenderEditable {
   @override
   bool hasFocus = false;
 
+  @override
+  Rect? getRectForComposingRange(TextRange range) {
```
Renzo-Olivares: Why is this needed?

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -538,6 +538,46 @@ void main() {
...
+          cursorWidth: cursorWidth,
+          selectionControls: materialTextSelectionControls,
+        ),
+      );
+    }
+
+    await tester.pumpWidget(buildWidget(2.0));
+    final EditableTextState state = tester.state(find.byType(EditableText));
+
+    // Ensure valid selection and focus
```
Renzo-Olivares: nit: end comments with a period, here and elsewhere in this file.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1859,6 +1868,36 @@ void main() {
...
+      (WidgetTester tester) async {
+        final TextSelectionOverlay overlay = await pumpApp(
+          tester,
+          value: const TextEditingValue(
+            text: 'test',
+            selection: TextSelection.collapsed(offset: 0),
+          ),
+        );
+
+        // Show handles
```
Renzo-Olivares: nit: end comments with a period, here and elsewhere in this file.

Renzo-Olivares (commented):
@muhammadkamel thank you for the quick changes left a few more comments. In general this change is looking good, but I still have some concerns over the naming of `cursorWidth`. It feels strange to me that `SelectionOverlay` has a `cursorWidth` field when it does not paint or control the cursor. Maybe `handleAnchorOffset` is appropriate here, what do you think? Open to other names.

packages/flutter/test/widgets/text_selection_test.dart
muhammadkamel: Ohh, I will remove it!

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1254,6 +1310,21 @@ class SelectionOverlay {
...
   final ValueChanged<DragStartDetails>? onStartHandleDragStart;
 
+  /// {@macro flutter.widgets.editableText.cursorWidth}
+  ///
+  /// This value is used for calculating the position of the text selection
+  /// handles.
+  double? _handleAnchorOffset;
+
+  /// Updates the handle anchor offset used for handle positioning.
+  void updateHandleAnchorOffset(double? value) {
```
Renzo-Olivares: nit: this can just be a setter on a public `handleAnchorOffset` similar to other fields in this class.

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for the quick changes. I've been thinking more about the naming around `cursorWidth` and `handleAnchorOffset`. I think a more appropriate name would be `targetWidth`, and some documentation could be as follows:

```dart
/// The [targetWidth] represents the width of the object (e.g. the cursor)
/// that the handle is pointing to. This can be used to center the handle
/// horizontally on that object.
```

Curious what other think of this name, my main concern with the naming (`cursorWidth`) is that this API is being added to `SelectionOverlay` which is used by `SelectableRegion` and does not currently have a visual representation for a collapsed selection because in the framework it only supports static text.

My other thought is around having it default to 2.0 (the same as the default EditableText.cursorWidth). While this matches the behavior in `EditableText`, if there was some other input field implementation with a different cursor width then the fix would not work and the developer would still need to figure out why their handles are not aligned. If it is null by default then that user will need to discover the fix. If it is required then the user will be made aware automatically but this is a breaking change. I'm leaning towards the breaking change so users are notified automatically about having to provide this new member but curious what others think. cc @loic-sharma @justinmc

justinmc:
@muhammadkamel Would you be able to fix the conflicts here? Also I owe you a review, sorry for the delay.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,16 @@ abstract class TextSelectionControls {
   /// Get the anchor point of the handle relative to itself. The anchor point is
   /// the point that is aligned with a specific point in the text. A handle
   /// often visually "points to" that location.
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight);
+  ///
+  /// {@template flutter.widgets.TextSelectionControls.getHandleAnchor.cursorWidth}
+  /// The [cursorWidth] argument is the thickness of the cursor, which may be
```
loic-sharma: Silly nitpick, our technical writing style guide prefers to use "may" for legal context only.

```suggestion
  /// The [cursorWidth] argument is the thickness of the cursor, which can be
```

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1254,6 +1310,22 @@ class SelectionOverlay {
   /// Called when the users start dragging the start selection handles.
   final ValueChanged<DragStartDetails>? onStartHandleDragStart;
 
+  /// {@macro flutter.widgets.editableText.cursorWidth}
+  ///
+  /// This value is used for calculating the position of the text selection
```
loic-sharma:
```suggestion
  /// This value is used to calculate the position of the text selection
```

dkwingsmt:
Hi! @muhammadkamel  Are you able to address the comments above so we can work together to get this landed? This is a great addition and I'm looking forward to it!

packages/flutter/lib/src/cupertino/desktop_text_selection.dart
```diff
@@ -70,7 +70,11 @@ class CupertinoDesktopTextSelectionControls extends TextSelectionControls {
 
   /// Gets the position for the text selection handles, but desktop has none.
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
```
Renzo-Olivares, line 73: This is part of a public API so it should be deprecated and not removed.

packages/flutter/lib/src/cupertino/text_selection.dart
```diff
@@ -153,9 +153,13 @@ class CupertinoTextSelectionControls extends TextSelectionControls {
 
   /// Gets anchor for cupertino-style text selection handles.
   ///
-  /// See [TextSelectionControls.getHandleAnchor].
+  /// See [TextSelectionControls.calculateHandleAnchor].
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
```
Renzo-Olivares, line 158: This is part of a public API so it should be deprecated and not removed.

packages/flutter/lib/src/material/desktop_text_selection.dart
```diff
@@ -71,7 +71,11 @@ class DesktopTextSelectionControls extends TextSelectionControls {
 
   /// Gets the position for the text selection handles, but desktop has none.
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
```
Renzo-Olivares, line 74: This is part of a public API so it should be deprecated and not removed.

packages/flutter/lib/src/material/text_selection.dart
```diff
@@ -107,12 +107,16 @@ class MaterialTextSelectionControls extends TextSelectionControls {
 
   /// Gets anchor for material-style text selection handles.
   ///
-  /// See [TextSelectionControls.getHandleAnchor].
+  /// See [TextSelectionControls.calculateHandleAnchor].
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
```
Renzo-Olivares, line 112: This is part of a public API so it should be deprecated and not removed.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -306,9 +332,11 @@ class EmptyTextSelectionControls extends TextSelectionControls {
   }
 
   @override
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
```
Renzo-Olivares, line 309: This is a public API so we should probably deprecate this instead of removing it.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,33 @@ abstract class TextSelectionControls {
   /// Get the anchor point of the handle relative to itself. The anchor point is
   /// the point that is aligned with a specific point in the text. A handle
   /// often visually "points to" that location.
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight);
+  @Deprecated(
+    'Use `calculateHandleAnchor` instead. '
+    'This feature was deprecated after v3.32.0-0.0.pre.',
+  )
+  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight) {
+    return calculateHandleAnchor(type, textLineHeight, targetWidth: 2.0);
```
Renzo-Olivares: This method previously did not have a default implementation, is there a reason it should now? If not we should omit it, which makes it clear to implementers that they need to implement it.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,33 @@ abstract class TextSelectionControls {
...
+  /// horizontally on that object.
+  ///
+  /// Subclasses should override either this method or [getHandleAnchor].
+  /// {@endtemplate}
+  Offset calculateHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    required double targetWidth,
+  }) {
+    return getHandleAnchor(type, textLineHeight);
```
Renzo-Olivares: I don't think this method needs a default implementation. This can be left to the implementer.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,33 @@ abstract class TextSelectionControls {
...
+  ///
+  /// The anchor point is the point that is aligned with a specific point in
+  /// the text. A handle often visually "points to" that location.
+  ///
+  /// {@template flutter.widgets.TextSelectionControls.calculateHandleAnchor.targetWidth}
+  /// The [targetWidth] represents the width of the object (e.g. the cursor)
+  /// that the handle is pointing to. This can be used to center the handle
+  /// horizontally on that object.
+  ///
+  /// Subclasses should override either this method or [getHandleAnchor].
```
Renzo-Olivares: I would move this last bit out of the template.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -447,14 +477,45 @@ class TextSelectionOverlay {
...
+    _handlesDismissalTimer?.cancel();
+    _handlesDismissalTimer = null;
+  }
+
+  // The duration for auto-dismissal of handles on Android.
+  // This matches the native Android behavior observed on physical devices
+  // (e.g., Pixel 9 running Android 16).
+  static const Duration _androidHandlesDismissalDuration = Duration(seconds: 4);
+
+  void _scheduleHandlesDismissal() {
```
Renzo-Olivares: nit: maybe `_scheduleHandleDismissalIfSupported`.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1083,10 +1152,12 @@ class SelectionOverlay {
     )
     Offset? toolbarLocation,
     this.magnifierConfiguration = TextMagnifierConfiguration.disabled,
+    required double targetWidth,
```
Renzo-Olivares: This is technically a breaking change. 

Option 1: Make this nullable.

Option 2: Add a default value.

If a user:
1. Sets a different cursor width than the default for the area that uses `SelectionOverlay`.
2. Uses `MaterialTextSelectionControls` or another `TextSelectionControls` that uses `targetWidth`.
Then their handle anchor may not be accurately calculated if the user does not know to provide an accurate targetWidth to `SelectionOverlay`.

In both options the handle anchor is left inaccurate if the user does not know to provide an accurate targetWidth to `SelectionOverlay`. I'm leaning towards making this nullable since this was previously not working before and also certain "selectable" widgets like static text do not have a collapsed cursor so they do not need this, also Cupertino handles do not consider `targetWidth`. I don't feel too strongly about either option, ideally users would be notified someway that they need to provide targetWidth after this change but that is a breaking change. Curious to hear others thoughts on this.

By making it required users are broken and notified that they must provide an accurate targetWidth so their handle anchor is properly calculated.

If a user is creating their own `TextSelectionControls` then from implementing `calculateHandleAnchor` they would know to provide an accurate `targetWidth` to `SelectionOverlay`.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -114,7 +114,33 @@ abstract class TextSelectionControls {
...
+  /// the text. A handle often visually "points to" that location.
+  ///
+  /// {@template flutter.widgets.TextSelectionControls.calculateHandleAnchor.targetWidth}
+  /// The [targetWidth] represents the width of the object (e.g. the cursor)
+  /// that the handle is pointing to. This can be used to center the handle
+  /// horizontally on that object.
+  ///
+  /// Subclasses should override either this method or [getHandleAnchor].
+  /// {@endtemplate}
+  Offset calculateHandleAnchor(
```
Renzo-Olivares, line 159: I think adding this new method will technically be a breaking change. So implementers will be notified of this. Just something to keep in mind.

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for coming back to this and addressing the feedback! I left a few comments.

muhammadkamel:
> Hi @muhammadkamel Are you able to address the comments above so we can work together to get this landed? This is a great addition and I'm looking forward to it!

Hi @dkwingsmt 

Sorry for the delay — I was on vacation. I’ll address the comments and keep you updated.

dkwingsmt:
(From triage) FYI: Let us know when this is good for another round of review. Thanks!

muhammadkamel:
> (From triage) FYI: Let us know when this is good for another round of review. Thanks!

Hi @dkwingsmt 

Could you please review the latest commit changes?

packages/flutter/lib/src/cupertino/text_selection.dart
```diff
@@ -151,11 +151,12 @@ class CupertinoTextSelectionControls extends TextSelectionControls {
     }
   }
 
-  /// Gets anchor for cupertino-style text selection handles.
```
Renzo-Olivares, line 154: We should re-add this documentation snippet to the deprecated API and the new API as well.

packages/flutter/lib/src/cupertino/desktop_text_selection.dart
```diff
@@ -68,12 +68,28 @@ class CupertinoDesktopTextSelectionControls extends TextSelectionControls {
     return const SizedBox.shrink();
   }
 
-  /// Gets the position for the text selection handles, but desktop has none.
```
Renzo-Olivares, line 71: We should re-add this documentation snippet to the deprecated API and the new API as well.

packages/flutter/lib/src/cupertino/text_selection.dart
```diff
@@ -178,6 +179,19 @@ class CupertinoTextSelectionControls extends TextSelectionControls {
...
+
+  @Deprecated(
+    'Use `calculateHandleAnchor` instead. '
+    'This feature was deprecated after v3.32.0-0.0.pre.',
+  )
+  @override
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    double cursorWidth = 2.0,
```
Renzo-Olivares, line 191: The deprecated API should not have `cursorWidth`. That is reserved for the new API. Also we changed `cursorWidth` -> `targetWidth`.

packages/flutter/lib/src/cupertino/desktop_text_selection.dart
```diff
@@ -68,12 +68,28 @@ class CupertinoDesktopTextSelectionControls extends TextSelectionControls {
...
 
+  @Deprecated(
+    'Use `calculateHandleAnchor` instead. '
+    'This feature was deprecated after v3.32.0-0.0.pre.',
+  )
+  @override
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    double cursorWidth = 2.0,
```
Renzo-Olivares, line 88: The deprecated API should not have `cursorWidth`. That is reserved for the new API. Also we changed `cursorWidth` -> `targetWidth`.

packages/flutter/lib/src/material/desktop_text_selection.dart
```diff
@@ -69,12 +69,28 @@ class DesktopTextSelectionControls extends TextSelectionControls {
...
 
+  @Deprecated(
+    'Use `calculateHandleAnchor` instead. '
+    'This feature was deprecated after v3.32.0-0.0.pre.',
+  )
+  @override
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    double cursorWidth = 2.0,
```
Renzo-Olivares, line 89: The deprecated API should not have `cursorWidth`. That is reserved for the new API. Also we changed `cursorWidth` -> `targetWidth`.

packages/flutter/lib/src/material/desktop_text_selection.dart
```diff
@@ -69,12 +69,28 @@ class DesktopTextSelectionControls extends TextSelectionControls {
     return const SizedBox.shrink();
   }
 
-  /// Gets the position for the text selection handles, but desktop has none.
```
Renzo-Olivares, line 72: We should re-add this documentation snippet to the deprecated API and the new API as well.

packages/flutter/lib/src/material/text_selection.dart
```diff
@@ -105,18 +105,32 @@ class MaterialTextSelectionControls extends TextSelectionControls {
     };
   }
 
-  /// Gets anchor for material-style text selection handles.
```
Renzo-Olivares, line 108: We should re-add this documentation snippet to the deprecated API and the new API as well.

packages/flutter/lib/src/material/text_selection.dart
```diff
@@ -105,18 +105,32 @@ class MaterialTextSelectionControls extends TextSelectionControls {
...
 
+  @Deprecated(
+    'Use `calculateHandleAnchor` instead. '
+    'This feature was deprecated after v3.32.0-0.0.pre.',
+  )
+  @override
+  Offset getHandleAnchor(
+    TextSelectionHandleType type,
+    double textLineHeight, {
+    double cursorWidth = 2.0,
```
Renzo-Olivares, line 129: The deprecated API should not have cursorWidth. That is reserved for the new API. Also we changed cursorWidth -> targetWidth.

packages/flutter/lib/src/widgets/text_selection.dart
Renzo-Olivares: A way around this would be to make this method a stub..

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -111,10 +115,73 @@ abstract class TextSelectionControls {
     VoidCallback? onTap,
   ]);
 
-  /// Get the anchor point of the handle relative to itself. The anchor point is
-  /// the point that is aligned with a specific point in the text. A handle
-  /// often visually "points to" that location.
-  Offset getHandleAnchor(TextSelectionHandleType type, double textLineHeight);
```
Renzo-Olivares, line 117: We should re-add this documentation snippet to the deprecated API.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -526,20 +665,23 @@ class TextSelectionOverlay {
   /// synchronously. This means that it is safe to call during builds, but also
   /// that if you do call this during a build, the UI will not update until the
   /// next frame (i.e. many milliseconds later).
-  void update(TextEditingValue newValue) {
-    if (_value == newValue) {
+  void update(TextEditingValue newValue, {double? targetWidth}) {
```
Renzo-Olivares, line 676: I recommend exposing a setter for `targetWidth` on `TextSelectionOverlay` rather than changing `update`.

Renzo-Olivares:
Since `Material` and `Cupertino` libraries are currently under a code freeze I will mark this as a draft. This will be migrated over when the code freeze ends https://github.com/flutter/flutter/issues/184093.

Piinks:

Piinks:

Piinks:
If you decide to return to this, material_ui and cupertino_ui are now accepting contributions. :)

The portions of this PR that affect the material and cupertino libraries can be ported over to flutter/packages, although they may have to wait for the changes here in the core SDK to reach stable first.

PR #178626, opened 2025-11-17
[fuchsia] Add wrapper for zx_iob_writev

sutes-work:
Unfortunately, I don't think there's an easy way to add a test for this because it's possible to create IOBuffers from dart.

I need to add this because Fuchsia logging now uses IOBuffers rather than sockets, so some logging code that I have will need this.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
+++ b/engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
@@ -239,6 +239,10 @@ base class System extends NativeFieldWrapperClass1 {
   @pragma('vm:external-name', 'System_VmoMap')
   external static MapResult vmoMap(Handle vmo);
 
+  // Iob operations.
+  @pragma('vm:external-name', 'System_IobWrite')
+  external static int iobWrite(Handle iob, int regionIndex, ByteData data);
+
   // Time operations.
   static int clockGetMonotonic() {
     if (zirconFFIBindings == null) {
--- a/engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/sdk_ext/system.cc
+++ b/engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/sdk_ext/system.cc
@@ -16,6 +16,7 @@
 #include <unistd.h>
 #include <zircon/process.h>
 #include <zircon/processargs.h>
+#include <zircon/syscalls/iob.h>
 
 #include "flutter/fml/unique_fd.h"
 #include "third_party/tonic/dart_binding_macros.h"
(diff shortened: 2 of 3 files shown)


engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
```diff
@@ -239,6 +239,10 @@ base class System extends NativeFieldWrapperClass1 {
   @pragma('vm:external-name', 'System_VmoMap')
   external static MapResult vmoMap(Handle vmo);
 
+  // Iob operations.
+  @pragma('vm:external-name', 'System_IobWrite')
+  external static int iobWrite(Handle iob, int regionIndex, ByteData data);
```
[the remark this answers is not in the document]
sutes-work: None of the other methods have these comments and adding this comment would make this appear inconsistent with the rest of the file. If comments are to be added, they should be added for every method in this file.

engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/sdk_ext/system.h
```diff
@@ -58,6 +58,10 @@ class System : public fml::RefCountedThreadSafe<System>,
 
   static Dart_Handle VmoMap(fml::RefPtr<Handle> vmo);
 
+  static zx_status_t IobWrite(fml::RefPtr<Handle> iob,
+                              uint32_t region_index,
+                              const tonic::DartByteData& data);
```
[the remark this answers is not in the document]
sutes-work: Similarly, none of the other methods in this file have these comments, so adding this comment would make this stand out and appear inconsistent from the rest of the file.

engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/sdk_ext/system.cc
```diff
@@ -565,6 +566,23 @@ Dart_Handle System::VmoMap(fml::RefPtr<Handle> vmo) {
   return ConstructDartObject(kMapResult, ToDart(ZX_OK), object);
 }
 
+zx_status_t System::IobWrite(fml::RefPtr<Handle> iob,
+                             uint32_t region_index,
+                             const tonic::DartByteData& data) {
+  if (!iob || !iob->is_valid()) {
+    data.Release();
```
zijiehe-google-com, line 573: Since destructor of DartByteData calls Release(), what's the suggested scenario of calling Release() directly?

zijiehe-google-com (approved):

engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon/sdk_ext/system.cc
sutes-work: This is modeled after System::VmoWrite which does this. If we want this changed here, then we would change VmoWrite too. I'm happy to do that, but I went for consistency here.

sutes-work: Let me know what you want to do.

zijiehe-google-com: @jason-simmons any suggestion if Release() is preferred being called explicitly?

jason-simmons: Calling `Release` as soon as possible is recommended.

`tonic::DartByteData` wraps a Dart API that acquires a pointer to the data within a Dart TypedData object.  The release should be done promptly because other Dart APIs that allocate memory can not be used until the pointer is released.

zijiehe-google-com: Thank you Jason. Then no further action is needed.

jason-simmons (approved):

PR #177029, opened 2025-10-14
Some cleanup of cross library test imports

justinmc:
There is a lot of cross-library importing in the framework unit tests that I'd like to clean up, see the design doc for more: https://docs.google.com/document/d/1UHxALQqCbmgjnM1RNV9xE2pK3IGyx-UktGX1D7hYCjs/edit?tab=t.0

This PR cleans up a few obvious instances and adds TODOs for others. I created this while doing an investigation for the design doc linked above. I hope that we'll be able to follow up with fixes for all of the problematic tests (tracked in the issue below).

Part of https://github.com/flutter/flutter/issues/177028

PR #130523, opened 2023-07-13
Add lint check to make sure samples are linked and have tests

gspencergoog:
## Description

This adds a checker that will check all of the API docs examples to make sure that they are linked from at least one source file in the packages.

It also checks to make sure that all of the examples have an associated test. Since there are a large number that don't have tests at the moment, there is also a large exception list that can be burned down (burn down list is in https://github.com/flutter/flutter/issues/130459).

Because there are missing links currently, this PR will only pass after https://github.com/flutter/flutter/pull/130521 is merged.

## Related Issues

- https://github.com/flutter/flutter/issues/129956
- https://github.com/flutter/flutter/issues/130459

## Tests

- Added test for the checker.

PR #178693, opened 2025-11-18
Test cross import lint

justinmc:
This PR adds a lint rule to catch imports between Material, Cupertino, and Widgets  in tests. Spun out of https://github.com/flutter/flutter/pull/177029 and the [Decoupling Tests design doc](https://docs.google.com/document/d/1UHxALQqCbmgjnM1RNV9xE2pK3IGyx-UktGX1D7hYCjs/edit?pli=1&tab=t.0).

For now, Material will be the place to put tests that need to import both Material and Cupertino. See https://github.com/flutter/flutter/pull/178693#discussion_r2550510287.

I used https://github.com/flutter/flutter/pull/130523 as a reference to create this PR.


dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,602 @@
...
+    );
+    final Set<File> widgetsTestsImportingCupertino = getFilesWithImports(
+      filesByLibrary[Library.widgets]!,
+      Library.cupertino,
+    );
+    final Set<File> cupertinoTestsImportingMaterial = getFilesWithImports(
+      filesByLibrary[Library.cupertino]!,
+      Library.material,
+    );
+
```
navaronbracke: What about Material importing Cupertino?

justinmc: Material importing Cupertino is allowed, at least for now. Material will continue to contain platform adaptive code (like Switch.adaptive), and it will also be responsible for testing interactions that involve both Material and Cupertino.

dev/bots/analyze.dart
```diff
@@ -288,6 +288,13 @@ Future<void> run(List<String> arguments) async {
   // Ensure integration test files are up-to-date with the app template.
   printProgress('Up to date integration test template files...');
   await verifyIntegrationTestTemplateFiles(flutterRoot);
+
+  // Check for cross-library imports in tests.
```
loic-sharma: What this means might not be immediately obvious to a reader. What do you think of adding a quick example here? For example:

```suggestion
  // Check for cross-library imports in tests. For example,
  // the widget library's tests should not import the Material library.
```

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+/// Checks the tests in the Widgets and Cupertino libraries for cross imports.
+///
+/// Excludes known tests that contain cross imports, i.e.
+/// [TestsCrossImportChecker.knownWidgetsCrossImports] and
+/// [TestsCrossImportChecker.knownCupertinoCrossImports].
+///
+/// The Material library should contain tests that verify behaviors involving
+/// multiple libraries, such as platform adaptivity. Otherwise, these libraries
+/// should not import each other in tests.
```
loic-sharma: I really like the bullet point rules in the "Overview" section of your [flutter.dev/go/decoupling-framework-tests](http://flutter.dev/go/decoupling-framework-tests) doc:

> * The Cupertino library/tests should never import Material. It should test its widgets under a full-Cupertino scenario.
> * The Material library should test its widgets in a full-Material scenario.
> * Design languages are responsible for testing their interoperability with Widgets.
> * Tests that cover interoperability between Material and Cupertino should go in Material. See the next section for when and how interoperability should be tested.
> * The Widgets library/tests should never import Cupertino or Material.

 What do you think of copying the bullet point rules here? I don't feel strongly about this, feel free to skip.

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+        .where((File? file) {
+          if (file == null) {
+            return false;
+          }
+          if (filenamePattern == null) {
+            return true;
+          }
+          return file.absolute.path.contains(filenamePattern);
+        })
+        .map<File>((File? s) => s!)
```
loic-sharma: You might be able to simplify this using [`whereType`](https://api.flutter.dev/flutter/dart-core/Iterable/whereType.html):

```suggestion
        .whereType<File>()
        .where((File file) {
          if (filenamePattern == null) {
            return true;
          }
          return file.absolute.path.contains(filenamePattern);
        })
```

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+  );
+
+  /// Returns the Set of paths in `knownPaths` that are not in `files`.
+  static Set<String> _differencePaths(Set<String> knownPaths, Set<File> files) {
+    final Set<String> testPaths = files.map((File file) {
+      final int index = file.absolute.path.indexOf('packages/flutter/test');
+      final indexNormalized = index == -1 ? 0 : index;
+      return file.absolute.path.substring(indexNormalized);
+    }).toSet();
+    return knownPaths.difference(testPaths);
```
loic-sharma, line 365: Will this work on Windows? File paths might use `\` as the directory separator.

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+
+  // Get a list of all the filenames in the source directory that end in "_test.dart".
+  static Set<File> _getTestFiles(Directory directory, _Library library) {
+    return _getFiles(directory.childDirectory(library.directory), RegExp(r'_test\.dart$')).toSet();
+  }
+
+  /// Returns true only if the file imports the given Library.
+  static bool _containsImport(File testFile, _Library library) {
+    final String contents = testFile.readAsStringSync();
+    return contents.contains(library.regExp);
```
loic-sharma: How long does it take to run this script? (If it's slow, we can likely speed it up by only reading each test file once. But I wouldn't bother with this optimization unless it's necessary)

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+  static final RegExp widgetsImportRegExp = RegExp(r"import 'package:flutter\/widgets.dart'");
+
+  /// The RegExp that finds an import of this library.
+  RegExp get regExp {
+    return switch (this) {
+      _Library.widgets => widgetsImportRegExp,
+      _Library.material => materialImportRegExp,
+      _Library.cupertino => cupertinoImportRegExp,
+    };
+  }
```
loic-sharma: Should we use string contains instead of regexes here? The regexes don't seem to do any special matching.

I think it'd let us simplify this logic a bit, we could move the string patterns to a constructor argument. Something like:

```
  widgets(
    directory: 'widgets',
    name: 'Widgets',
    importPattern: "import 'package:flutter/widgets.dart'",
  ),
```

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+  widgets(directory: 'widgets', name: 'Widgets'),
+  material(directory: 'material', name: 'Material'),
+  cupertino(directory: 'cupertino', name: 'Cupertino');
+
+  const _Library({required this.directory, required this.name});
+
+  final String directory;
+  final String name;
+
+  static final RegExp cupertinoImportRegExp = RegExp(r"import 'package:flutter\/cupertino.dart'");
```
loic-sharma: If you stick to regexes, don't forget to escape the `.`s here:

```suggestion
  static final RegExp cupertinoImportRegExp = RegExp(r"import 'package:flutter\/cupertino\.dart'");
```

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,572 @@
...
+    }
+    return filesWithCrossImports;
+  }
+
+  /// Returns the error message for the given known paths that no longer have a
+  /// cross import.
+  ///
+  /// `library` must not be `Library.Material`, because Material is allowed to
+  /// cross-import.
+  static String _getFixedImportError(Set<String> fixedPaths, _Library library) {
```
loic-sharma, line 420: Nice, this is a great idea 👍

dev/bots/check_tests_cross_imports.dart
justinmc: I like it, I hope we can properly communicate the ideas in that design doc. This also reminded me to create an issue (https://github.com/flutter/flutter/issues/179429) about explaining this sort of thing in the test README files.

dev/bots/check_tests_cross_imports.dart
justinmc: Just a couple of seconds, which I guess is ok given that the analyze CI check takes an hour?

```
$ time bin/cache/dart-sdk/bin/dart --enable-asserts dev/bots/check_tests_cross_imports.dart                                                                                                                                                                                              ✔  4s   15:50:48 

▌15:51:15▐ No errors were detected with test cross imports.
bin/cache/dart-sdk/bin/dart --enable-asserts   5.60s user 0.51s system 153% cpu 3.971 total
```

loic-sharma: Ok gotcha, that seems fine 👍

dev/bots/check_tests_cross_imports.dart
justinmc: That's probably exactly why the Windows check is failing right now, thanks!

dev/bots/check_tests_cross_imports.dart
justinmc: Now irrelevant as I switched to string contains, but good catch!

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,563 @@
...
+        throw ArgumentError('All files must include $prefix in their path.', 'files');
+      }
+      final String comparablePath = file.absolute.path
+          .substring(index)
+          .replaceAll('/', Platform.isWindows ? r'\' : '/');
+      return !knownPaths.contains(comparablePath);
+    }).toSet();
+  }
+
+  // Get a list of all the filenames in the source directory that end in "_test.dart".
```
Renzo-Olivares: super nit: `///` to be consistent with other comments in this file.

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,563 @@
...
+        filesWithCrossImports.add(testFile);
+      }
+    }
+    return filesWithCrossImports;
+  }
+
+  /// Returns the error message for the given known paths that no longer have a
+  /// cross import.
+  ///
+  /// `library` must not be `Library.Material`, because Material is allowed to
```
Renzo-Olivares: Can we assert this?

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,563 @@
...
+      foundError(
+        _getImportError(
+          files: unknownCupertinoTestsImportingMaterial,
+          testLibraryName: _Library.cupertino.name,
+          importedLibraryName: _Library.material.name,
+        ).split('\n'),
+      );
+    }
+
+    // Find any known cross imports that weren't found, and are therefore fixed.
```
Renzo-Olivares, line 544: Should we add a TODO to remove this logic when the known imports lists are removed?

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,563 @@
...
+  }
+
+  /// Returns the File's relative path.
+  String _getRelativePath(File file, [Directory? root]) {
+    root ??= flutterRoot;
+    return path.relative(file.absolute.path, from: root.absolute.path);
+  }
+
+  /// Returns the import error for the `files` in testLibraryName which import
+  /// importedLibraryName.
```
Renzo-Olivares: should this also have a doc comment about the `testLibraryName` not being `Material` because it allows cross imports.

dev/bots/check_tests_cross_imports.dart
justinmc: Way more concise, thanks!

dev/bots/check_tests_cross_imports.dart
justinmc: I do below!

```dart
      _Library.material => throw UnimplementedError(
        'Material is responsible for testing its interactions with Cupertino, so it is allowed to cross-import.',
      ),
```

dev/bots/check_tests_cross_imports.dart
justinmc: I refactored it to only accept valid import errors, and I added a doc comment about what that means.

loic-sharma (approved):
Looks great!

dev/bots/check_tests_cross_imports.dart
```diff
@@ -0,0 +1,577 @@
...
+  };
+
+  static final Set<String> _knownCrossImports = knownWidgetsCrossImports.union(
+    knownCupertinoCrossImports,
+  );
+
+  /// Returns the Set of paths in `knownPaths` that are not in `files`.
+  static Set<String> _differencePaths(Set<String> knownPaths, Set<File> files) {
+    final Set<String> testPaths = files.map((File file) {
+      final prefix = RegExp('packages.flutter.test');
```
loic-sharma: Would this work?

```suggestion
      final prefix = RegExp('packages[\/\\]flutter[\/\\]test');
```

I'm OK with the current solution though 👍

loic-sharma: If you update this, don't forget to also update `_getUnknowns` too

justinmc: Good call, more strict!

Issue #178547, opened 2025-11-14
Flutter 3.38 has issues rendering paths in canvas


CodeDoctorDE:
### Steps to reproduce

Call toImage on picture recorder or RenderPaintBoundaries containing an canvas where it has a background and rendered paths on it.

Tested on Linux Fedora 43 with flutter 3.38.1, on web this problem doesn't exist. Last stable 3.35.7 didn't have the issue.

The problem is really visible in my note taking app https://github.com/LinwoodDev/Butterfly

### Expected results

Rendered output like in the app. Background yellow and above the paths.

### Actual results

The background behind the strokes won't be rendered

### Code sample

<details open><summary>Code sample</summary>

https://gist.github.com/CodeDoctorDE/0132efc42499331637776db6b6a96da8

```dart
import 'dart:math';
import 'dart:ui' as ui;
import 'dart:io';
import 'package:flutter/material.dart';
import 'package:flutter/foundation.dart';
import 'package:flutter/rendering.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Rainbow Paths',
      theme: ThemeData(
        // Using a modern Material 3 theme.
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const RainbowPathsPage(),
    );
  }
}

class RainbowPathsPage extends StatefulWidget {
  const RainbowPathsPage({super.key});

  @override
  State<RainbowPathsPage> createState() => _RainbowPathsPageState();
}

class _RainbowPathsPageState extends State<RainbowPathsPage> {
  List<List<Offset>> _pathsData = [];
  final GlobalKey _repaintBoundaryKey = GlobalKey();

  // Configuration for path generation:
  final int _numberOfPaths = 100; // How many distinct paths to draw.
  final int _segmentsPerPath =
      15; // How many line segments each path will have.
  // Maximum length a segment can extend from the previous point, to keep paths somewhat continuous.
  final double _maxSegmentLength = 100.0;

  @override
  void initState() {
    super.initState();
    // Schedule the path generation after the first frame has been rendered.
    // This ensures that `context` is fully initialized and has valid layout information
    // (like the screen size) when `_generatePathsForSize` is called.
    WidgetsBinding.instance.addPostFrameCallback((_) {
      if (mounted) {
[...]

PR #178697, opened 2025-11-18
Restore OpenGL state modified by fl_compositor_opengl_present_layers

jason-simmons:
The OpenGL context used there may be shared with Skia, and Skia assumes that other code has not changed its state.

See https://github.com/flutter/flutter/issues/178547


robert-ancell (approved):
Annoying but required :)

PR #178701, opened 2025-11-18
Fix RefreshIndicator sensitivity with NestedScrollView by using render size.

### LINKED ISSUE #147363: NestedScrollView with RefreshIndicator: dragging effect of RefreshIndicator seems to disappear with a slight drag
### Steps to reproduce

1.After placing RefreshIndicator inside NestedScrollView,
2.When there's a lot of content in NestedScrollView's headerSliverBuilder,
3.The dragging effect for pulling down to refresh disappears, and a slight drag immediately triggers the refresh action.

### Expected results

The consistent dragging effect when content is less in headerSliverBuilder

### Actual results

After nesting RefreshIndicator within NestedScrollView, when the content of NestedScrollView's headerSliverBuilder is extensive, the dragging effect of pulling down for refresh seems to disappear. A slight drag immediately triggers the refresh action.

### Code sample

```dart

  final String title;

  @override
}

class _MyHomePageState extends State<MyHomePage>
    with SingleTickerProviderStateMixin {
  late TabController _tabController;

  @override
  void initState() {
    super.initState();
    _tabController = TabController(length: 3, vsync: this);
  }

  @override
  void dispose() {
    _tabController.dispose();
    super.dispose();
  }

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: RefreshIndicator(
        notificationPredicate: (notification) {
          return notification.depth == 2;
        },
        onRefresh: () async {
          await Future.delayed(const Duration(seconds: 2), () {});
        },
        child: NestedScrollView(
          headerSliverBuilder: (
            BuildContext context,
            bool innerBoxIsScrolled,
          ) {
            return [
              SliverToBoxAdapter(
                child: Container(
                  height: 1000, // Adjusting the simulated height of content in headerSliverBuilder
                  color: Colors.amber,
                  alignment: Alignment.center,
                  child: const Text("header"),
                ),
              ),
            ];
          },
          body: TabBarView(
            controller: _tabController,
            children: [
              ListView(),
              ListView(),
              ListView(),
            ],
          ),
        ),
    );
  }
}
```

</details>

### Screenshots or Video

https://github.com/flutter/flutter/assets/17977044/44ff84d3-b465-4871-b94e-d7e96735fafb

### Logs

```console
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.19.5, on macOS 14.4.1 23E224 darwin-x64, locale zh-Hans-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
[!] Xcode - develop for iOS and macOS (Xcode 14.2)
    ! CocoaPods 1.10.2 out of date (1.13.0 is recommended).
        CocoaPods is used to retrieve the iOS and macOS platform side's plugin code that responds to your plugin usage on the Dart side.
        For more info, see https://flutter.dev/platform-plugins
      To upgrade see https://guides.cocoapods.org/using/getting-started.html#updating-cocoapods for instructions.
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.87.2)
```

</details>

moritasoshi:
This PR fixes overly sensitive pull-to-refresh behavior in RefreshIndicator when used with NestedScrollView that has large headers.

### Problem

When RefreshIndicator wraps a NestedScrollView with a large header (e.g., 550px), the inner scrollable viewport becomes very small (e.g., 50px). The refresh trigger threshold was previously calculated based on this small viewport dimension (50px × 0.25 = 12.5px), making the indicator extremely sensitive—even a slight 13px drag would trigger refresh. This eliminates the visual drag feedback users expect.

### Solution

Changed the threshold calculation to always use the RefreshIndicator's actual rendered size instead of the viewport dimension reported by the surrounding Scrollable. This provides more natural drag sensitivity for all layouts, especially those with small scrollable areas.

With this change, the threshold is based on the full RefreshIndicator size (e.g., 600px × 0.25 = 150px), requiring a more reasonable drag distance to trigger refresh.

### Changes

- Modified `_checkDragOffset()` to use `RenderBox.size.height` of the RefreshIndicator itself instead of viewport dimensions
- Added comprehensive test demonstrating the fix for NestedScrollView with large headers

Fixes #147363

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/refresh_indicator.dart
```diff
@@ -316,6 +319,20 @@ class RefreshIndicator extends StatefulWidget {
   /// Defaults to 2.0.
   final double elevation;
 
+  /// Whether to derive the container extent from the [RefreshIndicator]'s
+  /// rendered size.
+  ///
+  /// By default, [_checkDragOffset] uses the viewport dimension reported by the
```
victorsanni: Referencing a private method in public API docs is causing the `Linux docs test` failure.

moritasoshi: Thanks! I removed the private method reference.

packages/flutter/lib/src/material/refresh_indicator.dart
```diff
@@ -528,6 +546,19 @@ class RefreshIndicatorState extends State<RefreshIndicator>
     }
   }
 
+  double _resolveContainerExtent(double containerExtent) {
```
victorsanni: what happens if we always use the renderbox size instead of the containerExtent?

packages/flutter/lib/src/material/refresh_indicator.dart
```diff
@@ -528,6 +546,19 @@ class RefreshIndicatorState extends State<RefreshIndicator>
...
+    if (!widget.useIndicatorExtent) {
+      return containerExtent;
+    }
+
+    final RenderObject? renderObject = context.findRenderObject();
+    if (renderObject is RenderBox && renderObject.hasSize) {
+      return renderObject.size.height;
```
victorsanni: I'm wondering if this is one of those bugs that only happens if containerExtent > size or vice versa (in that case, we can just use the max/min). Can you explain more here?

victorsanni (commented):
It would be nice if we can fix this bug without increasing the API surface.

packages/flutter/lib/src/material/refresh_indicator.dart
moritasoshi: There would be no difference in most cases.

RefreshIndicator wraps its child as a Stack, so its RenderBox size typically matches the child's size. The `containerExtent` (from ScrollNotification's viewport dimension) should also match this size in properly structured layouts.
Differences only appear when there's extra Padding or size constraints between RefreshIndicator and the Scrollable—which shouldn't happen in well-written code.

packages/flutter/lib/src/material/refresh_indicator.dart
moritasoshi: Yes.
The issue occurs when `containerExtent < size.height` - when the viewport dimension is too small compared to the indicator's rendered size.

In NestedScrollView, scroll notifications from both the outer (header section) and inner (body section) are dispatched. Due to NestedScrollView's design, when both outer and inner are expanded, **overscroll values can only be obtained from the inner scroll** (the outer scroll coordinator prevents overscroll on the outer scrollable, so overscroll notifications are never fired from the outer side no matter how far you pull). Therefore, we must use `notification.depth == (inner's depth)` in the notificationPredicate, but when the header section is large, that viewport becomes quite narrow, causing this issue.

packages/flutter/lib/src/material/refresh_indicator.dart
victorsanni: So why not use the renderbox size always? That way we won't have to increase the API surface.

moritasoshi: Yeah, thanks.

I was a bit hesitant because it would change the behavior slightly.

But it seems like any reasonable code wouldn't experience regressions. I'll modify the code to always use the size from the context.

victorsanni: Can you remove the flag and always use the size? So I can approve and run the `Google testing` CI check.

moritasoshi: Sure! I’ve removed the flag and updated it to always use the size.
The fix is complete, so please go ahead and run the Google testing CI check.

moritasoshi: Looks like some tests are failing... Any thoughts?

victorsanni: Rebased to fix

packages/flutter/test/material/refresh_indicator_test.dart
```diff
@@ -1270,4 +1270,72 @@ void main() {
...
+            ),
+          ),
+        ),
+      ),
+    );
+
+    final double indicatorHeight = tester.getSize(find.byType(RefreshIndicator)).height;
+    final double threshold = indicatorHeight * 0.25;
+    // The threshold is based on RefreshIndicator's render size, not the inner body viewport.
+    // A small drag of 15px should NOT trigger refresh
```
victorsanni:
```suggestion
    // A small drag of 15px should NOT trigger refresh.
```

packages/flutter/test/material/refresh_indicator_test.dart
```diff
@@ -1270,4 +1270,72 @@ void main() {
...
+    await tester.fling(find.text('Large Header'), const Offset(0.0, 15.0), 1000.0);
+    await tester.pump();
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
+    expect(refreshCalled, false);
+
+    // A drag meeting the render-size-based threshold should trigger refresh
```
victorsanni:
```suggestion
    // A drag meeting the render-size-based threshold should trigger refresh.
```

victorsanni (approved):

packages/flutter/test/material/refresh_indicator_test.dart
```diff
@@ -1270,4 +1270,72 @@ void main() {
...
+
+    final double indicatorHeight = tester.getSize(find.byType(RefreshIndicator)).height;
+    final double threshold = indicatorHeight * 0.25;
+    // The threshold is based on RefreshIndicator's render size, not the inner body viewport.
+    // A small drag of 15px should NOT trigger refresh
+    await tester.fling(find.text('Large Header'), const Offset(0.0, 15.0), 1000.0);
+    await tester.pump();
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
```
victorsanni: It's a little confusing to me that we need all these pumps.

packages/flutter/test/material/refresh_indicator_test.dart
```diff
@@ -1270,4 +1270,72 @@ void main() {
...
+    expect(refreshCalled, false);
+
+    // A drag meeting the render-size-based threshold should trigger refresh
+    refreshCalled = false;
+    await tester.fling(find.text('Large Header'), Offset(0.0, threshold), 1000.0);
+    await tester.pump();
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
+    await tester.pump(const Duration(seconds: 1));
```
victorsanni: here also.

MitchellGoodwin (approved):
LGTM as well pending Victor's suggestions.

packages/flutter/test/material/refresh_indicator_test.dart
moritasoshi: Thanks! Fixed it.

packages/flutter/test/material/refresh_indicator_test.dart
moritasoshi: Thank you for your comment.
I've updated the test to use `pumpAndSettle()` instead. The multiple pumps were needed because `onRefresh` is only called after the RefreshIndicator's snap animation finishes.

packages/flutter/lib/src/material/refresh_indicator.dart
```diff
@@ -514,8 +514,10 @@ class RefreshIndicatorState extends State<RefreshIndicator>
     return true;
   }
 
-  void _checkDragOffset(double containerExtent) {
+  void _checkDragOffset() {
     assert(_status == RefreshIndicatorStatus.drag || _status == RefreshIndicatorStatus.armed);
+    final renderBox = context.findRenderObject()! as RenderBox;
```
victorsanni: So now I'm a little confused here. Isn't this the renderbox of the refresh indicator? Or does the refresh indicator take up the size of the screen?

Piinks: I think so. This is a clever solution since the NestedScrollView's scroll notification metrics represent the individual scroll views rather than the sum of both. The refresh indicator wraps the whole Nested scroll view, so it has an accurate dimension for the full nested scrolling window.


We should put some assertions in place though. 
I would consider removing the `!` or adding a clear error message - for example if the context is invalidated while the gesture is being handled here.
Also, you should check that renderBox.hasSize before accessing size, if not that would also crash.

Piinks (changes requested):
Thanks for fixing this! Just a few adjustment to make before we can land it. There have been other PRs that tried to fix this, it looks like this is the right approach. 👍

Piinks (commented):
For the failing google test here, I would also check and consider how this change impacts this type of composition:

Column -> Expanded -> RefreshIndicator -> Container -> SingleChildScrollView (always scrollable scroll physics) -> with minimal content, not enough to scroll. Currently the refresh indicator can be used in this case, but with this change it breaks.

I can see about forming a runnable test case, I need to extract the Google bits.

packages/flutter/lib/src/material/refresh_indicator.dart
moritasoshi: Thank you. I have added the null check and size check to the code.

packages/flutter/lib/src/material/refresh_indicator.dart
```diff
@@ -514,8 +514,13 @@ class RefreshIndicatorState extends State<RefreshIndicator>
     return true;
   }
 
-  void _checkDragOffset(double containerExtent) {
+  void _checkDragOffset() {
     assert(_status == RefreshIndicatorStatus.drag || _status == RefreshIndicatorStatus.armed);
+    final renderBox = context.findRenderObject() as RenderBox?;
+    if (renderBox == null || !renderBox.hasSize) {
```
victorsanni, line 520: maybe we should assert here instead of silently returning

moritasoshi:
@victorsanni 
Thanks for the detailed explanation. I understand the issue now.
The core problem is that these two use cases require different reference values for calculating the drag threshold:

1. NestedScrollView with large headers: The containerExtent (from ScrollNotification) represents only the inner scrollable viewport, which can be much smaller than the visible area the user is interacting with.
2. RefreshIndicator wrapping padded content: The RefreshIndicator's rendered size includes non-scrolling areas (like padding), which is larger than the actual scrollable viewport.

In case 1, containerExtent is too small. In case 2, the rendered size is too large. There's no single value that correctly represents the "intended scrollable area" for both cases.
Given this fundamental conflict, I believe an optional parameter is necessary to let developers specify which calculation is appropriate for their layout.

Piinks:

moritasoshi:
Hi @Piinks, I've reworked the approach based on your feedback. Could you reopen this PR?

The fix now **only targets the NestedScrollView case** instead of impacting all scrollables:

```dart
if (notification.depth > 0) {
  final NestedScrollViewState? nsvState =
      notification.context?.findAncestorStateOfType<NestedScrollViewState>();
  if (nsvState != null) {
    final RenderBox? nsvRenderBox = nsvState.context.findRenderObject() as RenderBox?;
    if (nsvRenderBox != null && nsvRenderBox.hasSize) {
      containerExtent = nsvRenderBox.size.height;
    }
  }
}
```

- `depth > 0` guard ensures depth==0 scrollables (including padded SingleChildScrollView) are unaffected
- `findAncestorStateOfType<NestedScrollViewState>()` ensures only NestedScrollView cases use the adjusted extent
- Null safety is fully guarded with fallback to original `viewportDimension`
- Added regression tests for both NestedScrollView and non-NestedScrollView nested cases

Piinks:
Thanks for coming back! I am not sure isolating to NestedScrollView is the right approach either. Our style guide calls out that widgets should not have knowledge of each other, or behave differently based on the widget tree structure. This can lead to a poor developer experience when a widget behaves one way in one context, and quite differently in another.
I am also not sure, does this resolve the issue in miscalculation if the scroll view in the body of the nested scroll view in surrounded by other widgets like Padding?
Let's move discussion into the issue to see if we can work out a solution. :)

Issue #99963, opened 2022-03-11
Support for Linux RISC-V


TheOneWithTheBraid:
## Use case

RISC-V is an upcoming CPU architecture which can similarly to ARM result in very energy-efficient processors. I won't start a discussion on benefits or issues related to RISC-V architecture; there are plenty out in the internet and in scientific environment.

Even though RISC-V is not very common for desktop or mobile devices yet, there is definitely momentum in the development RISC-V. Most major Linux and BSD distributions support RISC-V, such as Ubuntu, Debian or FreeBSD.

- **Present:** not many use cases. There are a few Linux-capable development boards as well as one fully featured desktop SoC available.
- **Future:** expectedly more. There's a lot of ongoing development in the RISC-V community.

And what makes Linux all-day usable: Flutter of course. I would like to have this issue as a tracker for possibly upcoming RISC-V support and implementation.

## Proposal

Flutter currently supports amd64 and aarch64 on Linux. I'd propose to add support for RISC-V in addition.

This should mainly consist of:
- patches in the engine and Flutter tool to handle RISC-V
- pre-built Engine snapshots for RISC-V

## Related:

- Dart SDK MR: https://github.com/dart-lang/sdk/issues/38587

mit-mit:
RISC-V now has experimental support in Dart, and is available for Linux SDKs from our beta channel.

maks:
Given the [slow CPU and low performance I've seen on the Allwinner D1 when running the Dart compiler](https://github.com/dart-lang/sdk/issues/49253), which at the moment is in the cheapest and most readily available [development board](https://linux-sunxi.org/Sipeed_Lichee_RV), I think it would be most useful to start with supporting Linux on RISC-V using the same approach as Flutter has for mobile OS's (Android/iOS) as it would be a much better developer experience vs  trying to do on-device development the way Flutter currently supports on Linux (and MacOS, Windows).

iapicca:
after [apple considering risc-v architecture](https://9to5mac.com/2021/09/03/risc-v-chips-apple/)
google recently announced the [android risc-v porting](https://youtu.be/70O_RmTWP58) (article [here](https://arstechnica.com/gadgets/2023/01/google-announces-official-android-support-for-risc-v/))

@zanderso maybe the priority of this is issue should be taken up a notch?

xypron:
> Given the [slow CPU and low performance I've seen on the Allwinner D1 when running the Dart compiler]
[...]

PR #178711, opened 2025-11-18
[ Tool ] Add support for linux riscv64 architecture

vhaudiquet:
This adds support for buildind desktop linux applications for riscv64 on the flutter tool, as well as basic riscv64 support for the tool.

This is a first move towards fixing #99963. Along with https://github.com/flutter/flutter/pull/178712 that allows building an engine for `riscv64`, this PR allows me to build and run the flutter 'hello world' app on RISC-V hardware:

Screenshot from 2025-11-18 08-58-03

This allows someone who built a flutter engine and exposed it on a webserver to run commands like:
```
export FLUTTER_STORAGE_BASE_URL=http://your-server-url
flutter
flutter doctor
cd example/hello_world && flutter build linux
```

This means that even if Google does not want to support Flutter on RISC-V officially (hosting the artifacts), community-managed servers could host the RISC-V flutter engines/artifacts. It also means that internally, companies using Flutter on RISC-V would just have to build their engine and setup such a server, and it allows them to use the upstream Flutter tool.

## Pre-launch Checklist

- [x] I signed the [CLA].


reidbaker (approved):
Approved but I would want someone from the engine team to take a look before this lands.

reidbaker (approved):
Not sure where you are on the process of landing this feature but I would have expected to see a test that built for risc-v as part of this pr to ensure we dont regress on the ability to build.

reidbaker:
At the minimum I would expect that you could author a test that attempted to build the risc-v code. That code would be required before we can publish an RISC-V engine binary.

vhaudiquet:
I tried to understand the test infrastructure and add a test that is basically a copy of the one present on arm64, adapted for riscv64. Is that what you had in mind?

reidbaker (approved):
Thank you for adding a build test. This should still be approved by someone on the engine team but I no longer see any issues.

gaaclarke (approved):
LGTM.  @reidbaker the tests here assert that the tool will attempt to build riscv64 but we don't actually buildit as part of CI right?  We should consider adding that as a follow up.

vhaudiquet:
I also submitted #178712 to allow building the engine (either locally or in CI). I started a discussion there, that is still ongoing it seems. 

Having official tool support could still be useful even if you don't plan/want to build the engine on CI, as it would allow third-party actors to use the official tool with their third-party built engine using `export FLUTTER_STORAGE_BASE_URL` as described; this is why I submitted two different PRs.

bkonyi (approved):
LGTM! Let's figure out our plan for https://github.com/flutter/flutter/pull/178712 first and try and land this PRs together. I'll leave final approval to @jtmcdole.

packages/flutter_tools/lib/src/isolated/native_assets/native_assets.dart
```diff
@@ -666,6 +666,7 @@ OS getNativeOSFromTargetPlatform(TargetPlatform platform) {
       return OS.macOS;
     case TargetPlatform.linux_x64:
     case TargetPlatform.linux_arm64:
+    case TargetPlatform.linux_riscv64:
       return OS.linux;
```
dcharkes, line 670: Also packages/flutter_tools/lib/src/isolated/native_assets/targets.dart

```dart
  static List<AssetBuildTarget> targetsFor({

      case TargetPlatform.linux_arm64:
        return _linuxTarget(supportedAssetTypes, Architecture.arm64);
```

vhaudiquet: ~~Thanks, it seems I missed some of them. I used compiler/test errors to add everything, so I thought that was all. I guess I need to do another manual check of every TargetPlatform usage.~~

Edit: forget my comment above, I thought you were pointing out missing changes. What are you pointing out exactly?

packages/flutter_tools/lib/src/commands/build_linux.dart
```diff
@@ -80,8 +82,7 @@ class BuildLinuxCommand extends BuildSubCommand {
       throwToolExit('"build linux" only supported on Linux hosts.');
     }
     // Cross-building for x64 targets on arm64 hosts is not supported.
```
robert-ancell: Comment needs updating - "Cross-building for x64 targets on other hosts is not supported."?

robert-ancell: And is arm64 to/from riscv64 cross-compilation also allowed?

robert-ancell (approved):
Some small questions about the cross-compilation checks, but otherwise I can't see any issues.

robert-ancell (approved):
Some small questions about the cross-compilation checks, but otherwise I can't see any issues.

zanderso (changes requested):
Noting that the other PR indicates that this is still experimental, the ability to target riscv64 should be guarded by a feature flag to disable by default on any channel except main.

See
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/features.dart  and
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/config.dart

packages/flutter_tools/lib/src/commands/build_linux.dart
vhaudiquet: I guess the original intent of that code was to only allow cross-build on a x64 host. I can update that comment indeed.

vhaudiquet:
I pushed an update. This makes the `linux-riscv64` target available to the tool behind a feature flag, except on a riscv64 host. I feel like it makes sense to keep it on a riscv64 host, as anyway otherwise you can't cross-compile from riscv64. 

However, maybe that was not the right place in the code? Disabling it at the argument-parsing level can make sure that users will not even see the feature, but maybe this is not what we want.

Please review and tell me, I can change that, this is just a first draft for the feature. Thank you!

EDIT: It seems that this broke multiple tests. Fixing....

packages/flutter_tools/lib/src/commands/build_linux.dart
zanderso, line 26: My suggestion would be to parse the argument, then if riscv64 is selected but the feature has not been enabled, you can emit a warning explaining that the feature is experimental and how to enable it.

vhaudiquet: Ok, that seems more reasonable indeed. Thanks!

vhaudiquet: Solved with last change

vhaudiquet:
The google testing is the only one failing now. Is that a real fail, or a temporary/infra thing?

zanderso (approved):
lgtm

A Googler will need to follow step 5 of https://github.com/flutter/flutter/blob/master/docs/contributing/Feature-flags.md. @bkonyi would you be able to help out with that 2 line change?

jtmcdole:
This PR is 700+ commits behind master; updating to pick up changes.

jtmcdole:
Applying another fix internally

PR #178712, opened 2025-11-18
Add support for riscv64 desktop linux engine

vhaudiquet:
This allows to cross-compile a Flutter engine for riscv64 desktop linux from an x64 linux host.

This is a first move towards fixing https://github.com/flutter/flutter/issues/99963. Along with https://github.com/flutter/flutter/pull/178711 that allows using the flutter tool on riscv64, this PR allows me to build and run the flutter 'hello world' app on RISC-V hardware:

Screenshot from 2025-11-18 08-58-03

This allows to build a Flutter engine on x64 host, and then expose the artifacts with a webserver to the flutter tool of #178711:
```
# Host
et build -c ci/linux_debug_riscv64 && et build -c ci/linux_profile_riscv64 && et build -c ci/linux_release_riscv64
...http server...
# RISC-V
export FLUTTER_STORAGE_BASE_URL=http://host-server-url
flutter
```

This means that even if Google does not want to support Flutter on RISC-V officially (hosting the artifacts), community-managed servers could host the RISC-V flutter engines/artifacts. It also means that internally, companies using Flutter on RISC-V would just have to build their engine with upstream flutter instead of relying on third-party patches.

For now, building the engine on RISC-V still requires another patch:
```
diff --git a/src/Reactor/BUILD.gn b/src/Reactor/BUILD.gn
index 67dfeb0ec..dcb13614a 100644
--- a/src/Reactor/BUILD.gn
+++ b/src/Reactor/BUILD.gn
@@ -307,7 +307,7 @@ if (supports_subzero) {

 if (supports_llvm) {
   swiftshader_source_set("swiftshader_llvm_reactor") {
-    llvm_dir = "../../third_party/llvm-10.0"
+    llvm_dir = "../../third_party/llvm-16.0"

     deps = [
       ":swiftshader_reactor_base",
```
However once that dependency is upgraded, everything builds without issues.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/ci/builders/linux_riscv_host_engine.json
+++ b/engine/src/flutter/ci/builders/linux_riscv_host_engine.json
@@ -1,0 +1,151 @@
+{
+    "_comment": [
+        "The builds defined in this file should not contain tests, ",
+        "and the file should not contain builds that are essentially tests. ",
+        "The only builds in this file should be the builds necessary to produce ",
+        "release artifacts. ",
+        "Tests to run on linux hosts should go in one of the other linux_ build ",
+        "definition files."
+    ],
+    "luci_flags": {
+      "upload_content_hash": true
+    },
+    "builds": [
+        {
+            "archives": [
+                {
+                    "name": "ci/linux_profile_riscv64",
+                    "type": "gcs",
+                    "base_path": "out/ci/linux_profile_riscv64/zip_archives/",
+                    "include_paths": [
+                        "out/ci/linux_profile_riscv64/zip_archives/linux-riscv64-profile/linux-riscv64-flutter-gtk.zip"
+                    ],
+                    "realm": "production"
+                }
+            ],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Linux"
+            ],
+            "gclient_variables": {
+                "download_android_deps": false,
+                "download_jdk": false,
+                "use_rbe": true
+            },
+            "gn": [
+                "--target-dir",
+                "ci/linux_profile_riscv64",
+                "--runtime-mode",
+                "profile",
+                "--target-os=linux",
+                "--linux-cpu=riscv64",
+                "--prebuilt-dart-sdk",
+                "--no-lto",
+                "--rbe",
+                "--no-goma"
+            ],
+            "name": "ci/linux_profile_riscv64",
+            "description": "Produces profile mode artifacts to target riscv64 Linux from a Linux host.",
+            "ninja": {
+                "config": "ci/linux_profile_riscv64",
+                "targets": [
+                    "flutter/shell/platform/linux:flutter_gtk"
+                ]
+            }
+        },
+        {
+            "archives": [
+                {
+                    "name": "ci/linux_debug_riscv64",
+                    "type": "gcs",
+                    "base_path": "out/ci/linux_debug_riscv64/zip_archives/",
+                    "include_paths": [
+                        "out/ci/linux_debug_riscv64/zip_archives/linux-riscv64/artifacts.zip",
+                        "out/ci/linux_debug_riscv64/zip_archives/linux-riscv64/impeller_sdk.zip",
+                        "out/ci/linux_debug_riscv64/zip_archives/linux-riscv64/linux-riscv64-embedder.zip",
+                        "out/ci/linux_debug_riscv64/zip_archives/linux-riscv64/font-subset.zip",
+                        "out/ci/linux_debug_riscv64/zip_archives/linux-riscv64-debug/linux-riscv64-flutter-gtk.zip",
+                        "out/ci/linux_debug_riscv64/zip_archives/dart-sdk-linux-riscv64.zip"
+                    ],
+                    "realm": "production"
+                }
+            ],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Linux"
+            ],
+            "gclient_variables": {
+                "download_android_deps": false,
+                "download_jdk": false,
+                "use_rbe": true
+            },
+            "gn": [
+                "--target-dir",
+                "ci/linux_debug_riscv64",
+                "--runtime-mode",
+                "debug",
+                "--target-os=linux",
+                "--linux-cpu=riscv64",
+                "--prebuilt-dart-sdk",
+                "--no-lto",
+                "--rbe",
+                "--no-goma"
+            ],
+            "name": "ci/linux_debug_riscv64",
+            "description": "Produces debug mode artifacts to target riscv64 Linux from a Linux host.",
+            "ninja": {
+                "config": "ci/linux_debug_riscv64",
+                "targets": [
+                    "flutter/build/archives:artifacts",
+                    "flutter/build/archives:dart_sdk_archive",
+                    "flutter/tools/font_subset",
+                    "flutter/shell/platform/linux:flutter_gtk",
+                    "flutter/impeller/toolkit/interop:sdk",
+                    "flutter/build/archives:embedder"
+                ]
+            }
+        },
+        {
+            "archives": [
+                {
+                    "name": "ci/linux_release_riscv64",
+                    "type": "gcs",
+                    "base_path": "out/ci/linux_release_riscv64/zip_archives/",
+                    "include_paths": [
+                        "out/ci/linux_release_riscv64/zip_archives/linux-riscv64-release/linux-riscv64-flutter-gtk.zip"
+                    ],
+                    "realm": "production"
+                }
+            ],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Linux"
+            ],
+            "gclient_variables": {
+                "download_android_deps": false,
+                "download_jdk": false,
+                "use_rbe": true
+            },
+            "gn": [
+                "--target-dir",
+                "ci/linux_release_riscv64",
+                "--runtime-mode",
+                "release",
+                "--target-os=linux",
+                "--linux-cpu=riscv64",
+                "--prebuilt-dart-sdk",
+                "--no-lto",
+                "--rbe",
+                "--no-goma"
+            ],
+            "name": "ci/linux_release_riscv64",
+            "description": "Produces release mode artifacts to target riscv64 Linux from a Linux host.",
+            "ninja": {
+                "config": "ci/linux_release_riscv64",
+                "targets": [
+                    "flutter/shell/platform/linux:flutter_gtk"
+                ]
+            }
+        }
+    ]
+}
--- a/engine/src/build/config/compiler/BUILD.gn
+++ b/engine/src/build/config/compiler/BUILD.gn
@@ -401,6 +401,9 @@ config("compiler") {
       cflags += [ "--target=aarch64-linux-gnu" ]
       ldflags += [ "--target=aarch64-linux-gnu" ]
       cflags += [ "-DBORINGSSL_CLANG_SUPPORTS_DOT_ARCH" ]
+    } else if (current_cpu == "riscv64") {
+      cflags += [ "--target=riscv64-linux-gnu" ]
+      ldflags += [ "--target=riscv64-linux-gnu" ]
     }
   }
 
--- a/engine/src/build/config/compiler/BUILD.gn
+++ b/engine/src/build/config/compiler/BUILD.gn
@@ -411,6 +411,9 @@ config("compiler") {
       cflags += [ "--target=aarch64-linux-gnu" ]
       ldflags += [ "--target=aarch64-linux-gnu" ]
       cflags += [ "-DBORINGSSL_CLANG_SUPPORTS_DOT_ARCH" ]
+    } else if (current_cpu == "riscv64") {
+      cflags += [ "--target=riscv64-linux-gnu" ]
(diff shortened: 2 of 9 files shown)


engine/src/flutter/ci/builders/linux_riscv_host_engine.json
```diff
@@ -0,0 +1,151 @@
...
+            }
+        },
+        {
+            "archives": [
+                {
+                    "name": "ci/linux_debug_riscv64",
+                    "type": "gcs",
+                    "base_path": "out/ci/linux_debug_riscv64/zip_archives/",
+                    "include_paths": [
+                        "out/ci/linux_debug_riscv/zip_archives/linux-riscv/artifacts.zip",
```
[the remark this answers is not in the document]
jtmcdole: this looks like a bug?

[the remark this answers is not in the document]
vhaudiquet: Indeed! It seems I left a few typos here and there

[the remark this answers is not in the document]
vhaudiquet: I might have not seen it, as during my testing I copied the generated zip archives by hand to the right place for my webserver to serve it to the tool.

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
jtmcdole, line 1: Do you only want to support this for local builds right now?

vhaudiquet: I would love to be able to build that engine in CI. It would help the RISC-V community. But I don't think you want to officially build and serve that, right? So in the mean time, if it could be built by third-party CI pipelines, and unofficially distributed by the RISC-V community, I would be happy. Building locally was for testing, but that seems to be automatically buildable, right?

jtmcdole: We could build this in post submit rather than blocking in the merge queue. The question is do we need to support this on release builds or only at master? I'll start a chat to follow up

vhaudiquet: If I understand correctly, you would be ready to build this in CI but without that build being required to push/merge a commit. That is wonderful news. Release builds would be better as it would almost be first-class support for RISC-V. 

The main issue for me is that patch to swiftshader, which will be missing for the RISC-V build, but it seemed that patch was previously applied and reverted because it caused failure in Chromium/Chrome CI builds. I'm not sure if that is still the case though. 

Please let me know about your decision!

vhaudiquet:
I've only had to apply a patch to third_party swiftshader (as explained above). Other patches seem to have already been applied, or to not be necessary anymore?

I did not need anything except the swiftshader patch and this PR to build the engine.

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
vhaudiquet: @jtmcdole any updates on this?

robert-ancell (dismissed):
Change looks generally good to me. If this is still experimental then perhaps some comments noting that or a message when building. On the other hand, this only runs if you have a RISCV builder right? So you probably know the limitations if you've done that.

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
jtmcdole: I missed the github notifications. Until there's a .ci.yaml update that has something like `name: Linux linux_host_engine`, the artifacts are not built for master or release. We do not have a riscv machine in our pools - can this be built on arm or x64?

jtmcdole (approved):
This only builds locally for now. If we identify what machines we would need to build this at master, we could produce artifacts at a later time.

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
vhaudiquet: This can be built on x64 - but with the patch I mentioned in the PR description to `third_party/swiftshader`. I'm not sure how that part works, or how to send a patch there. Plus, this patch had already been sent but rolled back AFAIK (for something related to failing chromium builds).

vhaudiquet:
Hello @jtmcdole, happy new year! :smiley: 
Any updates? Is there anything still misssing?

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
jtmcdole: There's a way to do patches in [engine/src/flutter/third_party](https://github.com/flutter/flutter/tree/68d3d7b421e65bcdf5f8fb09227f0ec5396b7030/engine/src/flutter/third_party) - @jason-simmons might be better able to help there.

So right now, lets look at landing this; then get a patch in; then consider a x64 builder that can make artifacts at head.

jtmcdole (dismissed):

jtmcdole:
i've created a mirror. I'll see about landing the patch there and then updating DEPs.

gaaclarke:
still waiting on the patch from @jtmcdole, I've pinged him on that pr

gaaclarke:
@jtmcdole friendly ping

jtmcdole:
third_party/swiftshader doesn't build for me on infra. I've not had time to look into what is missing to get that building from the mirror. I can try an update the mirror / patch set sometime this week.

engine/src/flutter/ci/builders/linux_riscv_host_engine.json
jtmcdole: So.. took a while, but this did just land. Required more than your patch since llvm 16 and swiftshader were not happy with windows arm64.

jtmcdole:
> third_party/swiftshader doesn't build for me on infra. I've not had time to look into what is missing to get that building from the mirror. I can try an update the mirror / patch set sometime this week.

it does now.

vhaudiquet:
Hi @jtmcdole, thank you for your work on this!
What is left now to merge this one?
If needed, I can rebase this on top of master, and retry a build? Or do we want to get it in first, and land fixups later if the engine builds are not working? I will anyway try to find time this week to test an engine build with the new updated `third_party/swiftshader`.

jtmcdole:
I think a rebase and we should be able to review it, @vhaudiquet

vhaudiquet:
I have rebased on top of master.

I have retested a build, which **works**, but only after *others* small **patches**.

- **(1)** Unfortunately, it seems that a workaround used in `engine/src/flutter/shell/platform/linux/fl_view_accessible.cc` breaks the build on `trixie` sysroots; and right now only `riscv64` uses trixie (from the previous Flutter Android riscv64 builds, not this PR). Reading the workaround description, @robert-ancell is the one who fixed this upstream; but the fix has not made it into trixie (only in trixie-backports).
  We need a different workaround for trixie, as the code inside `atk.h` now uses real C++ features, so putting it all inside `extern C` is not possible. The more surgical fix is to only apply @robert-ancell fix to `atkversion.h`, but that file is not tracked in Git here... it is only inside the sysroot pulled by gclient.
  ```diff
  // Remove the old workaround that puts everything behind "extern C"
  diff --git a/engine/src/flutter/shell/platform/linux/fl_view_accessible.cc b/engine/src/flutte
  r/shell/platform/linux/fl_view_accessible.cc
  index df3a986ab9f..95cd63fe17f 100644
  --- a/engine/src/flutter/shell/platform/linux/fl_view_accessible.cc
  +++ b/engine/src/flutter/shell/platform/linux/fl_view_accessible.cc
  @@ -4,9 +4,7 @@
  
  - // Workaround missing C code compatibility in ATK header.
  - // Fixed in https://gitlab.gnome.org/GNOME/at-spi2-core/-/merge_requests/219
  -extern "C" {
   #include <atk/atk.h>
  -}
  
   #include "flutter/shell/platform/linux/fl_accessible_node.h"
   #include "flutter/shell/platform/linux/fl_accessible_text_field.h"
   ```
   ```diff
  // Surgically fix atkversion.h inside the sysroot
   --- a/engine/src/build/linux/debian_bullseye_amd64-sysroot/usr/include/atk-1.0/atk/atkversion.h
   +++ b/engine/src/build/linux/debian_bullseye_amd64-sysroot/usr/include/atk-1.0/atk/atkversion.h
   @@ -30,6 +30,8 @@

    #include <glib.h>

   +extern "C" {
   +
    /**
     * ATK_MAJOR_VERSION:
   @@ -474,5 +476,7 @@
    #define atk_interface_age atk_get_interface_age ()

   +}
   +
    #endif /* __ATK_VERSION_H__ */
  ```
- **(2)** A second issue seems to come from the fact that Dart SDK is built from source here (as there are no published dart riscv64 tarballs AFAIK). Because of that, something related to wasm-opt seems to break (unfamiliar with how it works exactly).
  ```diff
  diff --git a/engine/src/flutter/tools/gn b/engine/src/flutter/tools/gn
  index b7c34d8cc89..a97eed32fe6 100755
  --- a/engine/src/flutter/tools/gn
  +++ b/engine/src/flutter/tools/gn
  @@ -719,6 +719,16 @@ def to_gn_args(args):
       # dartdevc, web SDK kernel and source files.
       gn_args['dart_platform_sdk'] = not args.full_dart_sdk
  
  +  # When the Dart SDK is built from source (i.e. no usable prebuilt is
  +  # available for the target, such as for riscv64 Linux cross builds), exclude
  +  # the wasm-opt target. Binaryen's wasm-opt is built with exceptions enabled
  +  # and links the toolchain libc++, whose bundled libcxxabi exception symbols
  +  # collide with the no-exceptions stubs in the engine's libcxxabi. The
  +  # prebuilt Dart SDK already ships a wasm-opt binary for platforms that have
  +  # one, so this only affects from-source builds.
  +  if not gn_args.get('flutter_prebuilt_dart_sdk'):
  +    gn_args['dart_include_wasm_opt'] = False
  +
     if args.build_glfw_shell is not None:
       gn_args['build_glfw_shell'] = args.build_glfw_shell
  ```
   I had AI's help on this one, and here is the AI summary:
  >  - riscv64 has no prebuilt Dart SDK (flutter/prebuilts/ only has linux-x64 and linux-arm64;
  > can_use_prebuilt_dart returns False for linux_cpu=riscv64). So flutter_prebuilt_dart_sdk
  > stays false and the Dart SDK is built from source.
  > - Building the SDK from source builds wasm-opt (Binaryen), which is compiled with
  >   //build/config/compiler:enable_exceptions.
  > - The engine's own libcxxabi
  >   (flutter/build/secondary/flutter/third_party/libcxxabi/BUILD.gn) compiles
  >   cxa_noexception.cpp on Linux — stubs that define __cxa_current_primary_exception,
  >   __setExceptionClass, etc. as no-op/terminate.
  > - On Linux the build does not pass -nostdlib++ (build/config/compiler/BUILD.gn:644 only adds
  >   it if (!is_linux)), so the clang driver auto-links the toolchain's libc++.a. The riscv64
  >   buildtools/.../clang/lib/riscv64-unknown-linux-gnu/libc++.a bundles libcxxabi's
  >   cxa_exception.cpp.o (the real exception implementations).
  > - wasm-opt uses exceptions → the linker resolves __cxa_throw etc. by pulling cxa_exception.o
  >    out of the toolchain libc++.a. That object also defines
  >   __cxa_current_primary_exception/__setExceptionClass/… → duplicate symbols with the
  >   already-linked libcxxabi.cxa_noexception.o.

After those two changes:
```
$ et build -c ci/linux_debug_riscv64
[2026-08-12T11:26:47.574][ci/linux_debug_riscv64: GN]: OK
[2026-08-12T11:26:56.767][ci/linux_debug_riscv64: ninja]: 100.0% (5/5) ACTION //flutter/build/archives:artifacts(//build/toolchain/linux:clang_riscv64)
[2026-08-12T11:26:56.783][ci/linux_debug_riscv64: ninja]: OK
```

I am not sure how to proceed here. I believe the second issue should be patched here, and I will add a commit on the PR for that patch (unless you advise me not to). It seems to me that it makes sense to disable that feature for every "from-source" builds, as it will break every from-source build otherwise, AFAIU.

For the first issue, it gets a bit more tricky. The right solution would be to send the fix to Debian so that it gets into Trixie. Current Trixie version for that package is [`2.56.2-1+deb13u1`](https://packages.debian.org/source/stable/at-spi2-core), and the fix is available starting from upstream version `2.59.0`. I'm not completely familiar with how patches are sent to non-devel Debian versions, but I believe that this is the cleanest way. 

Another issue that we would have, though, is that Bullseye might be too old to patch (it is `oldoldstable`). 

The only other way IMO would be a script here that applies this patch directly.

Thoughts? Does everything makes sense? Do we want to take in the PR without the Debian fix, so that then we can build by manually patching, or do we want to wait until everything is ready first? Or maybe add another trixie-specific workaround that applies the patch, while hoping that Debian takes it?

vhaudiquet:
The Debian route seems convoluted, and I have confirmed that there will be no way to fix bullseye. I have added a patch function in the sysroot setup, I believe that this is the cleaner fix, and it will also allow updating sysroots to Trixie for other arches desktop builds, so I believe this is a win-win?

With this patch, the engine build succeeds :smiley: 
If you are willing to take all 3 commits, we will officially be able to cross-build the engine for riscv64

Issue #172561, opened 2025-07-22
[web] ☂️ WebParagraph


mdebbar:
## What is WebParagraph?

It's a new implementation of Flutter's text layout system on the web (ParagraphBuilder, Paragraph, etc). This new implementation takes advantage of new browser APIs, namely [Enhanced TextMetrics](https://chromestatus.com/feature/5075532483657728).

## Why?

The browser already contains a full text layout system, but because it's not exposed to web apps, we had to bundle a lot of the same bits in each Flutter web app. By utilizing Enhanced TextMetrics, we can achieve the following:

1. **Reduce code size**: Delete some parts of the text layout system that we don't need anymore: HarfBuzz, Font management, Freetype, Brotli decoder (was needed for Woff2), etc.
2. **Eliminate fallback fonts**: Fallback fonts cost extra network bandwidth and add complexity. They were needed because we didn't have access to system/browser fonts.
3. **Eliminate the tofu box issue!**
4. **Apple Emoji !**

Here's a list of issues that we aim to close with WebParagraph:

- https://github.com/flutter/flutter/issues/163823
- https://github.com/flutter/flutter/issues/157918

## How do I try it out?

Thanks for your interest in this project! Please subscribe to this issue to be notified when WebParagraph is ready for testing by our users.

PR #178748, opened 2025-11-18
WebParagrah: ellipsis

Rusino:
Part of https://github.com/flutter/flutter/issues/172561


engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/layout.dart
```diff
@@ -114,6 +116,23 @@ class TextLayout {
...
+    if (paragraph.paragraphStyle.ellipsis == null) {
+      _ellipsisBidiLevel = 0;
+    } else if (_ellipsisBidiLevel == null) {
+      final List<BidiRegion> regions = canvasKit.Bidi.getBidiRegions(
+        paragraph.paragraphStyle.ellipsis!,
+        ui.TextDirection.ltr,
+      );
+      assert(
+        regions.isNotEmpty && regions.length == 1,
+        'Ellipsis must be in a single direction only',
```
mdebbar:
```suggestion
        'The entire ellipsis must have the same text direction',
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/layout.dart
```diff
@@ -360,9 +405,22 @@ class TextLayout {
...
+      } else {
+        // We place the ellipsis block aat the beginning of the line (for RTL paragraph)
+        line.visualBlocks.insert(0, ellipsisBlock);
+      }
+    }
+
     // Now when we calculated all line metrics we have to correct placeholders that depend on it
     for (final LineBlock block in line.visualBlocks) {
-      if (block is TextBlock) {
+      if (block is TextBlock || block is EllipsisBlock) {
```
mdebbar:
```suggestion
      if (block is! PlaceholderBlock) {
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/layout.dart
```diff
@@ -1050,10 +1109,15 @@ abstract class LineBlock {
...
-  final double shiftFromLineStart;
+  double shiftFromLineStart;
 
   // TODO(mdebbar): Remove when possible!
   double get spanShiftFromLineStart;
+
+  String getText(int start, int end) {
+    assert(false);
+    return '';
+  }
```
mdebbar: Add a comment explaining why this is implemented the way it is.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paint.dart
```diff
@@ -177,7 +180,9 @@ class TextPaint {
         .translate(lineOffset.dx, lineOffset.dy);
 
     if (WebParagraphDebug.logging) {
-      final String text = paragraph.getText1(webTextCluster.start, webTextCluster.end);
+      final String text = block is EllipsisBlock
+          ? block.getText(webTextCluster.start, webTextCluster.end)
```
mdebbar: Would it be simpler to just do:

```suggestion
          ? paragraph.paragraphStyle.ellipsis!
```
and remove the `getText` methods from `LineBlock` and `EllipsisBlock`?

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/painter.dart
```diff
@@ -77,7 +77,18 @@ abstract class Painter {
 class CanvasKitPainter extends Painter {
   @override
   void paintBackground(ui.Canvas canvas, LineBlock block, ui.Rect sourceRect, ui.Rect targetRect) {
-    canvas.drawRect(targetRect, block.style.background!);
+    // We need to snap the block edges because Skia draws rectangles with subpixel accuracy
+    // and we end up overlaps (only is a problem when the colors have transparency)
+    // or gaps between blocks (which looks unacceptable - vertical likes between blocks).
+    // Whether we snap to floor or ceil is irrelevant as long as we are consistent on both sides
+    // (and will possibly have problems when glyph boundaries are outside of advantage rectangles)
```
mdebbar: did you mean
```suggestion
    // (and will possibly have problems when glyph boundaries are outside of advance rectangles)
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/painter.dart
```diff
@@ -77,7 +77,18 @@ abstract class Painter {
 class CanvasKitPainter extends Painter {
   @override
   void paintBackground(ui.Canvas canvas, LineBlock block, ui.Rect sourceRect, ui.Rect targetRect) {
-    canvas.drawRect(targetRect, block.style.background!);
+    // We need to snap the block edges because Skia draws rectangles with subpixel accuracy
+    // and we end up overlaps (only is a problem when the colors have transparency)
+    // or gaps between blocks (which looks unacceptable - vertical likes between blocks).
```
mdebbar:
```suggestion
    // or gaps between blocks (which looks unacceptable - vertical lines between blocks).
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/painter.dart
```diff
@@ -77,7 +77,18 @@ abstract class Painter {
 class CanvasKitPainter extends Painter {
   @override
   void paintBackground(ui.Canvas canvas, LineBlock block, ui.Rect sourceRect, ui.Rect targetRect) {
-    canvas.drawRect(targetRect, block.style.background!);
+    // We need to snap the block edges because Skia draws rectangles with subpixel accuracy
+    // and we end up overlaps (only is a problem when the colors have transparency)
```
mdebbar:
```suggestion
    // and we end up with overlaps (this is only a problem when colors have transparency)
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/painter.dart
```diff
@@ -77,7 +77,18 @@ abstract class Painter {
...
+    // or gaps between blocks (which looks unacceptable - vertical likes between blocks).
+    // Whether we snap to floor or ceil is irrelevant as long as we are consistent on both sides
+    // (and will possibly have problems when glyph boundaries are outside of advantage rectangles)
+    final snappedRect = ui.Rect.fromLTRB(
+      targetRect.left.roundToDouble(),
+      targetRect.top,
+      targetRect.right.roundToDouble(),
+      targetRect.bottom,
```
mdebbar: Why not round everything?

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paragraph.dart
```diff
@@ -614,9 +614,19 @@ class PlaceholderSpan extends ParagraphSpan {
...
+    required super.end,
+    required super.style,
+    required this.text,
+    required this.textDirection,
+  });
 
   final String text;
+  // We use TextSpan to get metrics from Chrome in many places,
+  // including empty spans (for example when measuring strut) and
+  // ellipsis span (which inherits textDirection from the span it attaches to).
```
mdebbar, line 628: > and ellipsis span (which inherits textDirection from the span it attaches to)

This doesn't seem to be accurate. The ellipsis span text direction is derived from the bidi regions of the ellipsis text.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paragraph.dart
```diff
@@ -627,12 +637,18 @@ class TextSpan extends ParagraphSpan {
...
+    // We need to set in up because we otherwise in RTL text without textDirection
+    // Canvas2D will return all clusters placed right to left starting from 0.
+    // Also, we have a separate (possibly, different) textDirection for the ellipsis.
+    layoutContext.direction = textDirection == ui.TextDirection.ltr ? 'ltr' : 'rtl';
     return layoutContext.measureText(text);
   }
 
+  double? advanceWidth() {
+    return _metrics.width;
+  }
```
mdebbar, line 650: Does this make more sense as a getter?
```suggestion
  double? get advanceWidth => _metrics.width;
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
```diff
@@ -369,4 +382,90 @@ class _LineBuilder {
...
+        (_layout.lines.length + 1) < _layout.paragraph.paragraphStyle.maxLines!) {
+      return false;
+    }
+    if (_layout.paragraph.paragraphStyle.ellipsis == null ||
+        _layout.paragraph.paragraphStyle.ellipsis!.isEmpty) {
+      // No ellipsizing needed, but we have reached max lines
+      return true;
+    }
+    // We need to shape the ellipsis here because only here we know the span/textStyle we ellipsize with
+    final String ellipsis = _layout.paragraph.paragraphStyle.ellipsis!;
```
mdebbar:
```suggestion
    final String? ellipsis = _layout.paragraph.paragraphStyle.ellipsis;
    if (ellipsis == null || ellipsis.isEmpty) {
      // No ellipsizing needed, but we have reached max lines
      return true;
    }
    // We need to shape the ellipsis here because only here we know the span/textStyle we ellipsize with
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
```diff
@@ -369,4 +382,90 @@ class _LineBuilder {
...
+      cutOffWidth += widthCluster;
+      if (_isWhitespace(cluster)) {
+        // We skip whitespaces when cutting off for ellipsis, so just continue
+        WebParagraphDebug.log('Ellipsize: whitespace');
+      } else if (_widthConsumedText +
+              _widthWhitespaces +
+              _widthPendingText -
+              cutOffWidth +
+              ellipsisSpan.advanceWidth()! <=
+          _maxWidth) {
```
mdebbar: Can we do this instead:

```suggestion
      } else if (canFit(ellipsisSpan.advanceWidth()! - cutOffWidth)) {
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
```diff
@@ -57,8 +58,12 @@ class TextWrapper {
         // Break the line and then continue with the current cluster as usual
         WebParagraphDebug.log('isHardLineBreak: $index');
 
-        line.consumePendingText();
```
mdebbar, line 60: Are we not consuming pending text anymore here?

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
```diff
@@ -369,4 +382,90 @@ class _LineBuilder {
...
+      return false;
+    }
+    return _layout.lines.length >= _layout.paragraph.paragraphStyle.maxLines!;
+  }
+
+  bool ellipsize(int clusterIndex) {
+    if (_layout.paragraph.paragraphStyle.maxLines == null ||
+        (_layout.lines.length + 1) < _layout.paragraph.paragraphStyle.maxLines!) {
+      return false;
+    }
```
mdebbar: Can we express this in terms of a method call? It looks to me like this code is checking whether this is the last line or not. So maybe something like:

```suggestion
    if (!_isLastLine()) {
      return false;
    }
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
```diff
@@ -369,4 +382,90 @@ class _LineBuilder {
 
     return height;
   }
+
+  bool reachedMaxLines() {
+    if (_layout.paragraph.paragraphStyle.maxLines == null) {
+      return false;
+    }
+    return _layout.lines.length >= _layout.paragraph.paragraphStyle.maxLines!;
```
mdebbar:
```suggestion
    final int? maxLines = _layout.paragraph.paragraphStyle.maxLines;
    if (maxLines == null) {
      return false;
    }
    return _layout.lines.length >= maxLines;
```

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paragraph.dart
Rusino: _metrics only gets calculated with _getMetrics call. Before that it's not available. 
I should probably have a check on that (or lazy call) but in any case I don't want
to make it too easy.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
Rusino: It is a method:
  bool reachedMaxLines() {
    if (_layout.paragraph.paragraphStyle.maxLines == null) {
      return false;
    }
    return _layout.lines.length >= _layout.paragraph.paragraphStyle.maxLines!;
  }

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
Rusino: Have no explanation here. It's a mistake.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paragraph.dart
Rusino: The default text direction that we use to shape the ellipsis comes from the attaching span. The default text direction matters, too. The ellipsis text direction comes from the ellipsis itself, of course.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/layout.dart
```diff
@@ -1050,10 +1109,15 @@ abstract class LineBlock {
...
-  final double shiftFromLineStart;
+  double shiftFromLineStart;
 
   // TODO(mdebbar): Remove when possible!
   double get spanShiftFromLineStart;
+
+  String getText(int start, int end) {
+    assert(false, 'Only implemented for the ellipsis block');
+    return '';
+  }
```
mdebbar: I think this can be deleted now.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/layout.dart
```diff
@@ -1164,6 +1228,27 @@ class PlaceholderBlock extends LineBlock {
...
+
+  @override
+  String getText(int start, int end) {
+    if (span.text.isEmpty) {
+      return span.text;
+    }
+    assert(start >= 0);
+    assert(end <= span.text.length);
+    return span.text.substring(start, end);
+  }
```
mdebbar: Can be deleted.

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/paint.dart
```diff
@@ -126,12 +125,27 @@ class TextPaint {
...
             paintContext.restore();
           case StyleElements.text:
-            painter.fillTextCluster(clusterText, layout.isDefaultLtr);
+            painter.fillTextCluster(
+              clusterText,
+              // We shape ellipsis with default direction coming from the attaching block
+              // and all the other blocks with the default paragraph direction
+              block is EllipsisBlock
+                  ? block.isLtr
+                  : layout.paragraph.paragraphStyle.textDirection == ui.TextDirection.ltr,
```
mdebbar, line 149: Why are we using paragraph direction to draw all blocks, but we use block direction for ellipsis block? The comment above says "what" we are doing but it doesn't say "why".

engine/src/flutter/lib/web_ui/lib/src/engine/web_paragraph/wrapper.dart
mdebbar: So can we use the `reachedMaxLines()` method here instead of spelling out the entire condition?

mdebbar (approved):

PR #178824, opened 2025-11-19
Add contentTextStyle support to SimpleDialog

### LINKED ISSUE #59462: Add dialog theme for Simple Dialogs
Currently the Dialog Theme defined in Material App is used for Simple Dialogs just for the background color, it would be useful if Simple Dialogs also adopt titleTextStyle and contentTextStyle defined in the Dialog Theme, instead of just the default headline16.

777genius:
## Description

This PR adds `contentTextStyle` support to `SimpleDialog`, matching the functionality already available in `AlertDialog`.

### Problem

`SimpleDialog` was ignoring `DialogTheme.contentTextStyle`, even though the `DialogThemeData.contentTextStyle` documentation explicitly states it applies to `SimpleDialog.children`:

> "Overrides the default value for DefaultTextStyle for SimpleDialog.children and AlertDialog.content."

However, only `AlertDialog` actually implemented this behavior, forcing developers to specify text styles individually on each child widget.

### Solution

Added `contentTextStyle` parameter to `SimpleDialog` and applied it using `DefaultTextStyle`, following the same pattern as `AlertDialog`. The implementation uses the standard fallback chain:

`contentTextStyle` parameter → `DialogTheme.contentTextStyle` → M2/M3 defaults

### Testing

Added a new test `'SimpleDialog Custom Content Text Style'` that verifies `contentTextStyle` is correctly applied to the children of `SimpleDialog`.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/59462

## Pre-launch Checklist

- [x] I signed the [CLA].


dkwingsmt (approved):
LGTM. Thanks for your work! Also sorry for late review.

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1288,7 +1301,7 @@ class SimpleDialog extends StatelessWidget {
     // The paddingScaleFactor is used to adjust the padding of Dialog
     // children.
     final TextStyle defaultTextStyle =
```
QuncCccccc: nit: I think we can rename the variable to effectiveTextStyle. Just feel defaultTextStyle is a bit confusing and effectiveXXX is convention we use for this path.

packages/flutter/test/material/dialog_test.dart
```diff
@@ -303,6 +303,22 @@ void main() {
     expect(materialWidget.clipBehavior, Clip.antiAlias);
   });
 
+  testWidgets('SimpleDialog Custom Content Text Style', (WidgetTester tester) async {
```
QuncCccccc, line 298: Can we add one more unit test to berify DialogTheme.contentTextStyle works?

QuncCccccc (approved):
Thanks for adding this! LGTM. I'll add a g3fix to fix the Google testing. The failures look expected since now we added the missing default value for content text style (`defaults.contentTextStyle`).

packages/flutter/lib/src/material/dialog.dart
777genius: Good catch! Renamed it to `effectiveTitleTextStyle` to be more specific — this way it's consistent with `effectiveTitlePadding` and `effectiveContentPadding` in the same method. Let me know if you'd prefer just           
  `effectiveTextStyle` instead.

packages/flutter/test/material/dialog_test.dart
777genius: Done! Added the test for `DialogTheme.contentTextStyle`. Also added Material3/Material2 fallback tests since the docs mention those defaults — figured it's good to have them covered too.

packages/flutter/test/material/dialog_test.dart
```diff
@@ -295,6 +295,71 @@ void main() {
...
+    await tester.pumpWidget(_buildAppWithDialog(dialog));
+
+    await tester.tap(find.text('X'));
+    await tester.pumpAndSettle();
+
+    final RenderParagraph content = _getTextRenderObjectFromDialog(tester, contentText);
+    expect(content.text.style, contentTextStyle);
+  });
+
+  testWidgets('SimpleDialog Custom Content Text Style - DialogTheme', (WidgetTester tester) async {
```
QuncCccccc, line 314: The linux analyzer is complaining some type issues but should be simple to fix:)
Image

777genius: Done! Linux analyze passes now. The issue was that the branch was based on an older master.

777genius: Thank you for review!

PR #178873, opened 2025-11-20
Fix DropdownMenuEntry.style not resolved when entry is highlighted

### LINKED ISSUE #177363: `DropdownMenuEntry` style's text style is not resolving with states
### Steps to reproduce

`textStyle` in `ButtonStyle` for `DropdownMenuEntry` is not resolving with states. It doesn't change fontWeight or fontSize for focused entry. 

You can use the following `textStyle` in the `ButtonStyle`;

```dart
textStyle: WidgetStateProperty.resolveWith(
                (states) => TextStyle(
                  fontWeight: states.contains(WidgetState.focused)
                      ? FontWeight.bold
                      : FontWeight.w400,
                  fontSize: 14,
                  height: 20 / 14,
                ),
              ),
```

`WidgetStateProperty.resolveWith` is working with other properties like `foregroundColor` as you can see in the reproducible code below.

### Expected results

It is expected to change text style according to the state of button in `DropdownMenu`

### Actual results

Not resolving with states on `textStyle`.

### Code sample

```dart
DropdownMenu(
        alignmentOffset: Offset(0, 6),
        menuStyle: MenuStyle(
          maximumSize: WidgetStatePropertyAll(Size.fromHeight(200)),
          padding: const WidgetStatePropertyAll(EdgeInsets.all(4)),
        ),
        dropdownMenuEntries: [
          DropdownMenuEntry(
            value: 'all',
            label: 'All',
            style: ButtonStyle(
              textStyle: WidgetStateProperty.resolveWith(
                (states) => TextStyle(
                  fontWeight: states.contains(WidgetState.focused)
                      ? FontWeight.bold
                      : FontWeight.w400,
                  fontSize: 14,
                  height: 20 / 14,
                ),
              ),
              padding: const WidgetStatePropertyAll(
                EdgeInsets.symmetric(horizontal: 10, vertical: 8),
              ),
              foregroundColor: WidgetStateProperty.resolveWith(
                (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade800
                    : Colors.black,
              ),
              backgroundColor: WidgetStateProperty.resolveWith(
                (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade50
                    : states.contains(WidgetState.hovered)
                        ? Colors.grey
                        : Colors.transparent,
              ),
              shape: WidgetStatePropertyAll(
                RoundedRectangleBorder(
                  borderRadius: BorderRadius.all(Radius.circular(10)),
                ),
              ),
            ),
          ),
          DropdownMenuEntry(
            value: 'upcoming',
            label: 'Upcoming',
            style: ButtonStyle(
              textStyle: WidgetStateProperty.resolveWith(
                    (states) => TextStyle(
                  fontWeight: states.contains(WidgetState.focused)
                      ? FontWeight.bold
                      : FontWeight.w400,
                  fontSize: 14,
                  height: 20 / 14,
                ),
              ),
              padding: const WidgetStatePropertyAll(
                EdgeInsets.symmetric(horizontal: 10, vertical: 8),
              ),
              foregroundColor: WidgetStateProperty.resolveWith(
                    (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade800
                    : Colors.black,
              ),
              backgroundColor: WidgetStateProperty.resolveWith(
                    (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade50
                    : states.contains(WidgetState.hovered)
                    ? Colors.grey
                    : Colors.transparent,
              ),
              shape: WidgetStatePropertyAll(
                RoundedRectangleBorder(
                  borderRadius: BorderRadius.all(Radius.circular(10)),
                ),
              ),
            ),
          ),
          DropdownMenuEntry(
            value: 'past',
            label: 'Past',
            style: ButtonStyle(
              textStyle: WidgetStateProperty.resolveWith(
                    (states) => TextStyle(
                  fontWeight: states.contains(WidgetState.focused)
                      ? FontWeight.bold
                      : FontWeight.w400,
                  fontSize: 14,
                  height: 20 / 14,
                ),
              ),
              padding: const WidgetStatePropertyAll(
                EdgeInsets.symmetric(horizontal: 10, vertical: 8),
              ),
              foregroundColor: WidgetStateProperty.resolveWith(
                    (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade800
                    : Colors.black,
              ),
              backgroundColor: WidgetStateProperty.resolveWith(
                    (states) => states.contains(WidgetState.focused)
                    ? Colors.teal.shade50
                    : states.contains(WidgetState.hovered)
                    ? Colors.grey
                    : Colors.transparent,
              ),
              shape: WidgetStatePropertyAll(
                RoundedRectangleBorder(
                  borderRadius: BorderRadius.all(Radius.circular(10)),
                ),
              ),
            ),
          ),
        ],
      ),
```

</details>

### Screenshots or Video

<details open>

### Expected Output

Image

### Current Output

Image

https://github.com/user-attachments/assets/46ebb877-ce62-4035-b907-58bb270b8ed2

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.35.1, on macOS 26.0.1 25A362 darwin-arm64, locale
    en-PK) [397ms]
    • Flutter version 3.35.1 on channel stable at
      /Users/ahmad/development/flutter
    • Framework revision 20f8274939 (10 weeks ago), 2025-08-14 10:53:09 -0700
    • Engine revision 1e9a811bf8
    • Dart version 3.9.0
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop,
      enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-lldb-debugging

    [1,403ms]
    • Android SDK at /Users/ahmad/Library/Android/sdk
    • Platform android-36, build-tools 34.0.0
      Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on
      this machine.
      To manually set the JDK path, use: `flutter config
      --jdk-dir="path/to/jdk"`.

[✓] Xcode - develop for iOS and macOS (Xcode 26.0.1) [984ms]
    • Build 17A400
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [50ms]

[✓] Android Studio (version 2025.1) [49ms]

[✓] VS Code (version 1.105.0) [19ms]
    • Flutter extension version 3.120.0

[✓] Connected device (4 available) [6.2s]
    • Ahmad's iPhone (wireless) (mobile) • 00008110-001238C02E98801E
      • ios            • iOS 26.0.1 23A355
    • iPhone 17 Pro (mobile)             • 828FAD79-1E9B-4C45-8729-656155AF3944
      • ios            • com.apple.CoreSimulator.SimRuntime.iOS-26-0 (simulator)
    • macOS (desktop)                    • macos
      • darwin-arm64   • macOS 26.0.1 25A362 darwin-arm64
    • Chrome (web)                       • chrome
      • web-javascript • Google Chrome 141.0.7390.109

[✓] Network resources [625ms]

• No issues found!
```

</details>

bleroux:
## Description

This PR adds logic to resolve `DropdownMenuEntry.style` properties with the correct `WidgetState` when an item is highlighted.

When `MenuItemButton` are highlighted the focused state is not added automatically because the focus does not move to the items (it stays on the `TextField` in order to let the user enters text to filter the items list). This PR sets the `MenuItemButton.statesController` with a forced focused state to let `MenuItemButton` know that the focused state should be use to resolve the properties.

## Related Issue

Fixes [DropdownMenuEntry style's text style is not resolving with states](https://github.com/flutter/flutter/issues/177363)

## Tests

- Adds 1 tests.

### DIFF
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -727,6 +727,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
   TextEditingController get _effectiveTextEditingController =>
       widget.controller ?? (_localTextEditingController ??= TextEditingController());
   final FocusNode _internalFocudeNode = FocusNode();
+  WidgetStatesController? _highlightedItemStatesController;
 
   FocusNode? _localTrailingIconButtonFocusNode;
   FocusNode get _trailingIconButtonFocusNode =>
@@ -759,6 +760,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
     _internalFocudeNode.dispose();
     _localTrailingIconButtonFocusNode?.dispose();
     _localTrailingIconButtonFocusNode = null;
+    _highlightedItemStatesController?.dispose();
     super.dispose();
   }
 
@@ -931,7 +933,13 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
       // Simulate the focused state because the text field should always be focused
       // during traversal. Include potential MenuItemButton theme in the focus
       // simulation for all colors in the theme.
-      if (entry.enabled && i == focusedIndex) {
+      final bool entryIsSelected = entry.enabled && i == focusedIndex;
+      if (entryIsSelected) {
+        _highlightedItemStatesController?.dispose();
+        _highlightedItemStatesController = WidgetStatesController(<WidgetState>{
+          WidgetState.focused,
+        });
+
         // Query the Material 3 default style.
         // TODO(bleroux): replace once a standard way for accessing defaults will be defined.
         // See: https://github.com/flutter/flutter/issues/130135.
@@ -986,6 +994,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
           excluding: excludeSemantics,
           child: MenuItemButton(
             key: enableScrollToHighlight ? buttonItemKeys[i] : null,
+            statesController: entryIsSelected ? _highlightedItemStatesController : null,
             style: effectiveStyle,
             leadingIcon: entry.leadingIcon,
             trailingIcon: entry.trailingIcon,
--- a/packages/flutter/test/material/dropdown_menu_test.dart
+++ b/packages/flutter/test/material/dropdown_menu_test.dart
@@ -506,6 +506,60 @@ void main() {
         paints..rect(color: defaultOverlayColor.withAlpha(0)),
       );
     });
+
+    // Regression test for https://github.com/flutter/flutter/issues/177363.
+    testWidgets('textStyle property is resolved when item is highlighted', (
+      WidgetTester tester,
+    ) async {
+      const TestMenu selectedItem = TestMenu.mainMenu3;
+      const TestMenu nonSelectedItem = TestMenu.mainMenu2;
+
+      final customButtonStyle = ButtonStyle(
+        textStyle: WidgetStateProperty.resolveWith(
+          (Set<WidgetState> states) => TextStyle(
+            fontWeight: states.contains(WidgetState.focused) ? FontWeight.bold : FontWeight.normal,
+          ),
+        ),
+      );
+
+      final menuEntries = <DropdownMenuEntry<TestMenu>>[];
+      for (final item in menuChildren) {
+        menuEntries.add(
+          DropdownMenuEntry<TestMenu>(
+            value: item.value,
+            label: item.label,
+            style: customButtonStyle,
+          ),
+        );
+      }
+
+      TextStyle? getItemLabelStyle(String label) {
+        final RenderObject paragraph = tester
+            .element<StatelessElement>(
+              find.descendant(of: findMenuItemButton(label), matching: find.text(label)),
+            )
+            .renderObject!;
+        return (paragraph as RenderParagraph).text.style;
+      }
+
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Scaffold(
+            body: DropdownMenu<TestMenu>(
+              initialSelection: selectedItem,
+              dropdownMenuEntries: menuEntries,
+            ),
+          ),
+        ),
+      );
+
+      // Open the menu.
+      await tester.tap(find.byType(DropdownMenu<TestMenu>));
+      await tester.pump();
+
+      expect(getItemLabelStyle(selectedItem.label)?.fontWeight, FontWeight.bold);
+      expect(getItemLabelStyle(nonSelectedItem.label)?.fontWeight, FontWeight.normal);
+    });
   });
 
   testWidgets('Inner TextField is disabled when DropdownMenu is disabled', (


packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -875,7 +877,13 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
       // Simulate the focused state because the text field should always be focused
       // during traversal. Include potential MenuItemButton theme in the focus
       // simulation for all colors in the theme.
-      if (entry.enabled && i == focusedIndex) {
+      final bool entryIsSelected = entry.enabled && i == focusedIndex;
+      if (entryIsSelected) {
+        _highlightedItemStatesController?.dispose();
+        _highlightedItemStatesController = WidgetStatesController(<WidgetState>{
+          WidgetState.focused,
+        });
```
bleroux, line 941: I opted for the creation of a new states controller each time because I don't think `WidgetStatesController` are supposed to be reused across several widgets (as they have listeners and a list of states which is updated by InkWell).
For instance, each `InkWell` (inside each `MenuItemButton`) will create a `Focus` widget with an `onFocusChange` callback that might remove the focused state. If the states controller is shared and this callback is called once then the controller will no more have the focus state turned on. Currently there is an `ExcludeFocus` widget surrongind each `MenuItemButton`, so this might not happen but it seems to me that reusing states controller for different widgets can lead to subtle issues.

@justinmc Maybe I'm too cautious here? Let me know if you encountered something similar.

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -917,7 +919,13 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
       // Simulate the focused state because the text field should always be focused
       // during traversal. Include potential MenuItemButton theme in the focus
       // simulation for all colors in the theme.
-      if (entry.enabled && i == focusedIndex) {
+      final bool entryIsSelected = entry.enabled && i == focusedIndex;
+      if (entryIsSelected) {
+        _highlightedItemStatesController?.dispose();
```
justinmc, line 938: Even if the entry is not selected should you still dispose _highlightedItemStatesController and set it to null?

packages/flutter/lib/src/material/dropdown_menu.dart
justinmc: By "reusing it for different widgets" you just mean for each MenuItemButton that is created each time this method is called?

In the docs for statesController I see:

```
  /// Classes based on this one can provide their own
  /// [WidgetStatesController] to which they've added listeners.
  /// They can also update the controller's [WidgetStatesController.value]
  /// however, this may only be done when it's safe to call
  /// [State.setState], like in an event handler.
```

It's not safe to call setState in the build phase, so I think you wouldn't be able to set the value back to `focused` here. So I think you are right, and you have to do it the way that you've done it here.

justinmc (approved):
LGTM :+1: . I can't think of a better way to do the WidgetStatesController than what you've done, but if you have any other ideas let me know.

packages/flutter/lib/src/material/dropdown_menu.dart
bleroux: > Even if the entry is not selected should you still dispose _highlightedItemStatesController and set it to null?

It will be set to null and disposed in two cases:
- when another entry is selected.
- when the DropdownMenu will be disposed.

I think it's ok as either there is one entry selected and the controller will be properly destroyed or there is none and the controller will be destroyed when the DropdownMenu is (this second case seems similar to usual states controllers which have the same life time than the widget they are attached to).

PR #178886, opened 2025-11-20
[web] Flutter errors should be reported with console.error()

### LINKED ISSUE #178340: Flutter web error messages use `console.log` instead of `console.error`.
### Steps to reproduce

Run the counter app, but add a runtime error inside the `_increment` function. Launch this app on chrome and click the button, then open up the console window.

You will see these are not printed as errors, but just normal text.

### Expected results

I would expect all errors to be printed as errors.

Note that on other platforms, these errors are emitted on `stderr` and not `stdout` like print messages are.

This is ultimately the root cause of https://github.com/dart-lang/ai/issues/283 - we listen to the stderr VM service stream for flutter app runtime errors, but we never get any.

The dwds tool which converts the chrome debug protocol to the VM service protocol treats `console.log` messages as "stdout" messages, and `console.error` messages as "stderr", which is I think the best we can do.

### Actual results

I do see the errors but on the stdout stream, which means we can't tell the difference between errors and regular debugging output.

mdebbar:
Fixes https://github.com/flutter/flutter/issues/178340


jason-simmons (approved):

harryterkelsen (approved):
LGTM

justinmc (approved):
LGTM 👍 

Should nothing in foundation be using debugPrint now? If so ideally we could have a lint for that, otherwise maybe just document it somewhere.

PR #178895, opened 2025-11-21
Add error description for nbsp character(\u202f)

### LINKED ISSUE #177479: Improve error message on string matching
We should improve the error message on string matching in flutter_test.

We have been looking at how to best roll in updates to date localizations that have (in a few cases) updated typical whitespace (U+0020) to be NBSP (non breaking space, U+202F). The localizations team confirmed this is WAI, and in some locales (like French) is a required typography rule.

We have found users' might experience this update like this:

Image

Which would be super frustrating to debug! We should improve the error message here to specify what the whitespace character is in the error message.

Separately we also discussed that in some cases (maybe not this particular case) users might like a test method that allows them to match strings without attention paid to whitespace, tracked in https://github.com/flutter/flutter/issues/177478

A little bit more context on this from the localizations team is available in b/442629033

QuncCccccc:
This PR is to add a more clear error description when there is a special white space NNBSP(\u202f). This whitespace character is used to connect time and "am"/"pm".

In this PR, I only updated the description in `_MatchesSemanticsData` and still not sure if it's necessary to create another custom matcher class to ignore special whitespace. I think it's valid to directly update users' test results since the expected results have indeed changed.

This is my experiment for the whitespace change when we update generated_date_localization.dart: cl/831625146 
I went through all Scuba diffs and all results look expected. Once this is landed, we can use the CL change (except the image change) as g3fix to update the date localizations.

Fixes https://github.com/flutter/flutter/issues/177479

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_test/lib/src/matchers.dart
+++ b/packages/flutter_test/lib/src/matchers.dart
@@ -2695,36 +2695,19 @@ class _MatchesSemanticsData extends Matcher {
     return true;
   }
 
-  String? _ignoreNBSP(String? input) {
-    if (input == null) {
-      return null;
-    }
-    return input.replaceAll('\u202f', ' ');
-  }
-
-  bool _checkStringMismatch(
-    Map<dynamic, dynamic> matchState,
-    String prefixText,
-    String? expected,
-    String? actual,
-  ) {
-    if (expected != null && actual != null && _ignoreNBSP(expected) == _ignoreNBSP(actual)) {
-      final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f');
-      return failWithDescription(matchState, '$prefixText was $actualWithNBSP');
-    }
-    return failWithDescription(matchState, '$prefixText was $actual');
+  bool _checkStringMismatch(Map<dynamic, dynamic> matchState, String prefixText, String actual) {
+    final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f');
+    return failWithDescription(matchState, '$prefixText was $actualWithNBSP');
   }
 
   bool _checkStringAttributeMismatch(
     Map<dynamic, dynamic> matchState,
     String prefixText,
-    AttributedString? expected,
-    AttributedString? actual,
+    AttributedString actual,
   ) {
-    if (expected!.string != actual!.string) {
-      return _checkStringMismatch(matchState, prefixText, expected.string, actual.string);
-    }
-    return failWithDescription(matchState, '$prefixText was: $actual');
+    final String actualStringWithNBSP = actual.string.replaceAll('\u202f', r'\u202f');
+    final actualWithNBSP = AttributedString(actualStringWithNBSP, attributes: actual.attributes);
+    return failWithDescription(matchState, '$prefixText was: $actualWithNBSP');
   }
 
   @override
@@ -2744,7 +2727,7 @@ class _MatchesSemanticsData extends Matcher {
     };
 
     if (label != null && label != data.label) {
-      return _checkStringMismatch(matchState, 'label', label, data.label);
+      return _checkStringMismatch(matchState, 'label', data.label);
     }
     if (attributedLabel != null &&
         (attributedLabel!.string != data.attributedLabel.string ||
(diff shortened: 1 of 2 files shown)


dkwingsmt:
Note: I was skeptical of the design of the error messages, because I thought it would be better to show the original nbsp character in the expected string and instead explain the details with appended descriptions, such as "where the 5th character is \nbsp", so that the shown expected string itself is true and copiable. 

I talked with Qun, and understands now that it is intentional that the expected string contains the escaped string, so that maintainers can simply copy the entire string to the source file.

The design looks good to me. I'll make a full review after the checks are green.

packages/flutter_test/lib/src/matchers.dart
```diff
@@ -2695,6 +2695,38 @@ class _MatchesSemanticsData extends Matcher {
...
+    Map<dynamic, dynamic> matchState,
+    String prefixText,
+    String? expected,
+    String? actual,
+  ) {
+    if (expected != null && actual != null && _ignoreNBSP(expected) == _ignoreNBSP(actual)) {
+      final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f');
+      return failWithDescription(matchState, '$prefixText was $actualWithNBSP');
+    }
+    return failWithDescription(matchState, '$prefixText was $actual');
```
chunhtai: why not also apply `final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f');` here?

expect('another\u202fString' , 'some String'); 

will first show 'another String', and then developer copy over they will see 'another\u202fString'  which is unecessary two step correction

packages/flutter_test/lib/src/matchers.dart
```diff
@@ -2695,6 +2695,38 @@ class _MatchesSemanticsData extends Matcher {
...
+  bool _checkStringAttributeMismatch(
+    Map<dynamic, dynamic> matchState,
+    String prefixText,
+    AttributedString? expected,
+    AttributedString? actual,
+  ) {
+    if (expected!.string != actual!.string) {
+      return _checkStringMismatch(matchState, prefixText, expected.string, actual.string);
+    }
+    return failWithDescription(matchState, '$prefixText was: $actual');
```
chunhtai: same should consider add `final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f')`
otherwise if the expect string are correctly using `\u202f` they error message will display string without the `\u202f`

dkwingsmt:
> Yes, in the unit test, we set the actual value to be "AAA\u202fBBB", and expected to be "AAA BBB", the error description shows the value should be "AAA\u202fBBB". If we reverse the actual and expected, like setting actual to be "AAA BBB", and set expected to be "AAA\u202fBBB", then the description will say the value should be "AAA BBB", and developers can still just copy and paste the message. Does that make sense? I can talk about it offline anytime:)!

Sounds good. Can you add a unit test for it?

packages/flutter_test/lib/src/matchers.dart
QuncCccccc: Ah it makes sense! Updated!

packages/flutter_test/lib/src/matchers.dart
QuncCccccc: Updated! Thanks!

chunhtai (approved):
LGTM

packages/flutter_test/lib/src/matchers.dart
```diff
@@ -2695,6 +2695,21 @@ class _MatchesSemanticsData extends Matcher {
...
+  }
+
+  bool _checkStringAttributeMismatch(
+    Map<dynamic, dynamic> matchState,
+    String prefixText,
+    AttributedString actual,
+  ) {
+    final String actualStringWithNBSP = actual.string.replaceAll('\u202f', r'\u202f');
+    final actualWithNBSP = AttributedString(actualStringWithNBSP, attributes: actual.attributes);
+    return failWithDescription(matchState, '$prefixText was: $actualWithNBSP');
```
dkwingsmt: nit: Rename it with "escaped", a more common way to call this process
```suggestion
    final String actualStringEscaped = actual.string.replaceAll('\u202f', r'\u202f');
    final actualEscaped = AttributedString(actualStringEscaped, attributes: actual.attributes);
    return failWithDescription(matchState, '$prefixText was: $actualEscaped');
```

packages/flutter_test/lib/src/matchers.dart
```diff
@@ -2695,6 +2695,21 @@ class _MatchesSemanticsData extends Matcher {
     return true;
   }
 
+  bool _checkStringMismatch(Map<dynamic, dynamic> matchState, String prefixText, String actual) {
+    final String actualWithNBSP = actual.replaceAll('\u202f', r'\u202f');
+    return failWithDescription(matchState, '$prefixText was $actualWithNBSP');
```
dkwingsmt: nit: Rename it with "escaped", a more common way to call this process
```suggestion
    final String actualEscaped = actual.replaceAll('\u202f', r'\u202f');
    return failWithDescription(matchState, '$prefixText was $actualEscaped');
```

packages/flutter_test/test/matchers_test.dart
```diff
@@ -913,6 +913,339 @@ void main() {
...
+        final node = _FakeSemanticsNode(data);
+
+        // Label
+        expect(
+          () => expect(node, matchesSemantics(label: 'label\u202f0')),
+          throwsA(
+            isA<TestFailure>().having(
+              (TestFailure e) => e.message,
+              'message',
+              contains(r'label was label 0'),
```
dkwingsmt: Will the error message in this case show a "space" in both expected and actual, since I think you only escaped one of them? In that case this PR is not solving this problem completely, but I'll leave it to you to decide whether it's acceptable.

QuncCccccc: I see! I think it makes sense. :) I updated the error description for both actual and expected, so it now contains 202f regardless of whether it's in actual or expected.

dkwingsmt (approved):

PR #178891, opened 2025-11-21
Update .ci.yaml and TESTOWNERS for arc macrobenchmark tests

b-luk:
The benchmarks were added in #178690, but that PR didn't update the .ci.yaml and TESTOWNERS files.

PR #178939, opened 2025-11-21
Sort entries in TESTOWNERS

b-luk:
Sorting these entries was suggested by the gemini-code-assist bot in another PR which modified the TESTOWNERS file: #178891 (see comments at https://github.com/flutter/flutter/pull/178891#discussion_r2548209393 and https://github.com/flutter/flutter/pull/178891#discussion_r2548209396)

This is a follow-up PR to perform the sorting. We don't want to do it in the PR where it was suggested #178891 because sorting the entries is out of scope of that PR's goal.


flar (commented):
Please add a description to the PR. I would recommend pointing to the other PR where the review bot is suggesting that this ordering should be maintained.

flar (approved):
LGTM

PR #178941, opened 2025-11-21
Resolving and piping the view ID  through the WidgetController and the TestPointer so that clicks wind up on the right view

mattkae:
## What's new?
- While coming up with a scheme for testing multi-window, I realize that issuing taps in `testWidgets` across views was not working because the view was always being set to `0`
- The fix is to resolve the view when we can, and provide an optional view in the other methods

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_test/lib/src/controller.dart
```diff
@@ -1040,12 +1040,15 @@ abstract class WidgetController {
     int buttons = kPrimaryButton,
     bool warnIfMissed = true,
     PointerDeviceKind kind = PointerDeviceKind.touch,
+    int? viewId,
```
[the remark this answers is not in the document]
loic-sharma: It seems like we should remove this `viewId` argument?

packages/flutter_test/lib/src/controller.dart
```diff
@@ -1567,6 +1608,7 @@ abstract class WidgetController {
     double touchSlopX = kDragSlopDefault,
     double touchSlopY = kDragSlopDefault,
     PointerDeviceKind kind = PointerDeviceKind.touch,
+    int? viewId,
```
loic-sharma: Can this be a `FlutterView?` instead of an `int?`. Ideally we should pass the `FlutterView` around down through all public APIs, and only convert to an `int` as late as possible.

It can be difficult to figure out where this `int` should come from, but you can find out how to acquire a `FlutterView` from its docs.

packages/flutter_test/lib/src/controller.dart
```diff
@@ -1675,6 +1727,7 @@ abstract class WidgetController {
     int buttons = kPrimaryButton,
     double frequency = 60.0,
     bool warnIfMissed = true,
+    int? viewId,
```
loic-sharma: Consider making this a `FlutterView` as well

packages/flutter_test/lib/src/controller.dart
```diff
@@ -1701,6 +1755,7 @@ abstract class WidgetController {
     int? pointer,
     int buttons = kPrimaryButton,
     double frequency = 60.0,
+    int viewId = 0,
```
loic-sharma: Consider making this a `FlutterView` here, as well as all locations below

packages/flutter_test/lib/src/controller.dart
```diff
@@ -1683,6 +1736,7 @@ abstract class WidgetController {
       pointer: pointer,
       buttons: buttons,
       frequency: frequency,
+      viewId: viewId ?? 0,
```
loic-sharma: Can we look this up from the finder if needed?

We shouldn't hardcode `0` as the implicit view ID: embedders should be able to stuff information in the view IDs, like a pointer, so the implicit view's ID might not be 0 on all embedders.

In locations where we do need a fallback view ID, can we look up the `WidgetsBinding.instance.platformDispatcher.implicitView` and throw if that's `null`?

packages/flutter_test/lib/src/controller.dart
mattkae: Yes I think it is fair to do this as a `FlutterView` at this level :+1:

knopp:
I am wondering a bit whether `viewId` couldn't be a property on `TestPointer` instead of being specified in each method?

packages/flutter_test/lib/src/controller.dart
```diff
@@ -781,6 +781,14 @@ abstract class WidgetController {
     ).view;
   }
 
+  FlutterView? _maybeViewOf(finders.FinderBase<Element> finder) {
+    try {
+      return _viewOf(finder);
```
chunhtai: this should be the other way around

_viewOf should use _maybeViewOf and throw if it returns null (if we really want it to throw)

chunhtai (approved):
overall lgtm, just a minor comment

packages/flutter_test/lib/src/test_pointer.dart
```diff
@@ -130,6 +136,7 @@ class TestPointer {
       _buttons = buttons;
     }
     return PointerDownEvent(
+      viewId: view?.viewId ?? WidgetsBinding.instance.platformDispatcher.implicitView?.viewId ?? 0,
```
loic-sharma: Instead of falling back to `0`, could we assert that either `view` or `WidgetsBinding.instance.platformDispatcher.implicitView` are non-null? Same comment for the rest of the file

loic-sharma (approved):
Looks good to me once pending comments are addressed!

packages/flutter_test/lib/src/test_pointer.dart
loic-sharma: Ah Matt posted this in Discord:

> ... we should still fall back to viewId 0 because test like test/live_widget_controller_test.dart override the WidgetsFlutterBinding

Could we fix those tests to explicitly pass in a view if needed? Alternatively, could we update those tests' override of `WidgetsFlutterBinding` to have a fake implicit view? I'm OK with falling back to `0` here, but I'd prefer these other options if possible.

mattkae: We actually do not need the `?? 0` there. It should resolve no problem!

PR #177410, opened 2025-10-22
[ios] Dynamic Content Resizing

LouiseHsu:
This PR fixes https://github.com/flutter/flutter/issues/169147

This is the iOS implementation of https://github.com/flutter/flutter/issues/149033

In Add-to-App scenarios, a common desired use case is to have the embedded FlutterView be able to size itself based on it's content. Up till now, the size had to be manually specified, which can be tedious and inaccurate.

This PR adds a new flag, `isAutoResizable`, which enables the ability for the embedded flutter view to dynamically set it's own size based off it's content. Note that this feature will *NOT* work if the the embedded app is wrapped in flexible or unsized widgets, such as Scaffold. The FlutterView will just not display at all. 

To use the new flag, just set it like
```
flutterViewController.autoResizable = true;
```
or in Swift

```
flutterViewController.isAutoResizable = true
```
## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #178958, opened 2025-11-22
Revert "[ios] Dynamic Content Resizing"

flar:
Reverts flutter/flutter#177410

The change is causing the task `wide_gamut_ios` to hang every time. Local testing shows that reverting this commit fixes the problem.


jtmcdole:
Added `emergency` label since this was not a RevertBot revert.

jtmcdole (approved):

PR #179109, opened 2025-11-26
Make sure that a CupertinoListTile doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the CupertinoListTile widget.


victorsanni (approved):

packages/flutter/test/cupertino/list_tile_test.dart
```diff
@@ -548,4 +548,17 @@ void main() {
     expect(firstPosition.value, 1.0);
     expect(lastPosition.value, 1.0);
   });
+
+  testWidgets('CupertinoListTile does not crash at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      const CupertinoApp(
+        home: Center(
+          child: SizedBox.shrink(
+            child: CupertinoListTile(title: Text('X'), trailing: CupertinoListTileChevron()),
```
victorsanni, line 557: I think it puts the chevron in by default? My request was we should expect that the chevron is correctly shrunk.

victorsanni (approved):

packages/flutter/test/cupertino/list_tile_test.dart
ahmedsameha1: Do you want me to assert that the size of the CupertinoListTileChevron is zero here?

dkwingsmt: I think this test is fine since we only want to verify that ListTile has size zero.

dkwingsmt (approved):
Checklist:

PR #102339, opened 2022-04-21
Fix ScrollMetricsNotifiation for SingleChildScrollView

Piinks:
Fixes https://github.com/flutter/flutter/issues/102181
Necessary to re-land https://github.com/flutter/flutter/pull/101460

`SingleChildScrollView` has its own  viewport render object, and only performs layout once (since it only has one child), so it never calls `applyContentDimensions` after the initial layout.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #87076, opened 2021-07-26
Add a hook for scroll position to notify scrolling context when dimen…

chunhtai:
The scrollable needs to know if the content dimensions in the scroll position has changed in order to update its semantics node correctly. This pr adds an API to ScrollingContext to listen to the change.

fixes https://github.com/flutter/flutter/issues/40419

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #179133, opened 2025-11-26
Send ScrollMetricsNotification when SingleChildScrollView is scrolled.

yiiim:
Background: #173857.

`ScrollMetricsNotification` was originally not intended to be triggered during scrolling, but `CustomScrollView` triggers a layout during scrolling, which also triggers `ScrollMetricsNotification`. `SingleChildScrollView` does not trigger a layout when scrolling, resulting in different behaviors between `SingleChildScrollView` and `CustomScrollView`. This PR unifies their behavior so that both trigger `ScrollMetricsNotification` when scrolling. 

@chunhtai, this breaks some of your tests in #87076, mainly because triggering `ScrollMetricsNotification` during scrolling happens one frame earlier than during layout, resulting in one less frame. Is this correct?

The original purpose was to update `_lastMetrics`. I tried three approaches: 
1) As shown in this PR, trigger `ScrollMetricsNotification` and update `_lastMetrics` on every scroll; 

2) Cancel triggering `ScrollMetricsNotification` during `CustomScrollView` scrolling, keep `applyContentDimensions` unchanged, and update only the `pixels` property in `_lastMetrics` during scrolling without sending notifications. Since pixels is already updated, `applyContentDimensions` will not trigger notifications either. This will cause more test failures.

3) Maintain the current `ScrollMetricsNotification` triggering, similar to #102339. 

I prefer the first approach, which keeps the behavior of `SingleChildScrollView` and `CustomScrollView` consistent.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/flutter/test/widgets/scroll_notification_test.dart
+++ b/packages/flutter/test/widgets/scroll_notification_test.dart
@@ -45,13 +45,13 @@ void main() {
     await gesture.up();
     await tester.pump(const Duration(seconds: 1));
 
-    expect(events.length, 5);
-    // user scroll do not trigger the ScrollMetricsNotification.
+    expect(events.length, 6);
     expect(events[0] is ScrollStartNotification, true);
     expect(events[1] is UserScrollNotification, true);
     expect(events[2] is ScrollUpdateNotification, true);
-    expect(events[3] is ScrollEndNotification, true);
-    expect(events[4] is UserScrollNotification, true);
+    expect(events[3] is ScrollMetricsNotification, true);
+    expect(events[4] is ScrollEndNotification, true);
+    expect(events[5] is UserScrollNotification, true);
 
     events.clear();
     // Change the content dimensions again.
@@ -222,10 +222,12 @@ void main() {
     ScrollNotification? notification;
 
     void handleNotification(ScrollNotification value) {
-      if (value is ScrollStartNotification ||
-          value is ScrollUpdateNotification ||
-          value is ScrollEndNotification) {
-        notification = value;
+      switch (value) {
+        case ScrollStartNotification():
+        case ScrollUpdateNotification(dragDetails: DragUpdateDetails _):
+        case ScrollEndNotification():
+          notification = value;
+        default:
       }
     }
 
--- a/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
+++ b/packages/flutter/lib/src/widgets/single_child_scroll_view.dart
@@ -7,12 +7,15 @@
 /// @docImport 'page_view.dart';
 library;
 
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:flutter/gestures.dart' show DragStartBehavior;
 import 'package:flutter/rendering.dart';
+import 'package:flutter/scheduler.dart';
 
 import 'basic.dart';
+import 'binding.dart';
 import 'focus_manager.dart';
 import 'focus_scope.dart';
 import 'framework.dart';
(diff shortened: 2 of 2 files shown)


packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -45,13 +45,13 @@ void main() {
...
 
-    expect(events.length, 5);
-    // user scroll do not trigger the ScrollMetricsNotification.
+    expect(events.length, 6);
     expect(events[0] is ScrollStartNotification, true);
     expect(events[1] is UserScrollNotification, true);
     expect(events[2] is ScrollUpdateNotification, true);
-    expect(events[3] is ScrollEndNotification, true);
-    expect(events[4] is UserScrollNotification, true);
+    expect(events[3] is ScrollMetricsNotification, true);
```
chunhtai, line 52: I just catch this now, but if only pixel (scroll position) is changing shouldn't this be `ScrollUpdateNotification` ? in that case line 51 has already included that, do we need to fire another ScrollMetricsNotification?

yiiim: This is the tricky part. For a `CustomScrollView`, every scroll triggers a relayout, so it dispatches a `ScrollMetricsNotification`, but this is not the case for a `SingleChildScrollView`. My idea is to make them consistent. I think the most ideal way is to not dispatch `ScrollMetricsNotification` on scroll, but this would be a breaking change for `CustomScrollView`. Alternatively, we could maintain the status quo and use the reopened #102339 to update `_lastMetrics`. @Piinks, do you remember why this PR was closed at the time?

The solution in this PR is to have `SingleChildScrollView` also dispatch a `ScrollMetricsNotification` on every scroll. As the PR description says, this also breaks some tests. I'd like to know which solution you think we should adopt? @chunhtai @Piinks.

Piinks:
> This PR unifies their behavior so that both trigger ScrollMetricsNotification when scrolling.

Why do we need to unify the behavior? What I am wondering here considering this change is, what is the problem this is trying to solve? Looking at the older linked issues, these are not meant to behave the same way.

yiiim:
> > ScrollMetricsNotification was originally not intended to be triggered during scrolling, but CustomScrollView triggers a layout during scrolling, which also triggers ScrollMetricsNotification.
> 
> Is this always true? Or only when the metrics have changed?

When a CustomScrollView scrolls, it inevitably triggers a layout pass, which calls applyContentDimensions. In applyContentDimensions, as shown here: https://github.com/flutter/flutter/blob/4de6fcc369e7619e216f29cb66aa53338a18de1d/packages/flutter/lib/src/widgets/scroll_position.dart#L670, since pixels has already changed, _isMetricsChanged() is always true. Therefore, scrolling a CustomScrollView will always trigger a ScrollMetricsNotification.

yiiim:
> It seems to me the CustomScrollView's behavior is incorrect unless the scrolling causes content size changes. Maybe what we want to do is that we should make sure it is not only scroll position change when the metrics changes before sending the notification

Understood. I previously tried preventing CustomScrollView from dispatching ScrollMetricsNotification while scrolling, but that caused many issues and broke a lot of tests (e.g., #87076). I’ll likely need to address those failures incrementally before proceeding.

chunhtai:
Agree with @Piinks , discuss this in a doc seems better so that you don't have to go back and forth with the implementation until we figure out a way forward. Also appreciate all the work you have been put into this so far!

Piinks:
Hi @yiiim! I hope you have been well, I just came across this in old PR triage. Is this something you would like to revisit?

PR #179147, opened 2025-11-26
Implementation of tooltip windows for win32

mattkae:
## What's new?
- Implemented tooltip windows for the Win32 platform
- Implemented size-to-content on Win32
- Wired up the window metrics event with constraints
- Updated the example application to demonstrate tooltips
- Wrote a bunch of unit tests

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/embedder/embedder.cc
```diff
@@ -1649,10 +1657,14 @@ MakeViewportMetricsFromWindowMetrics(
            "physical height or width.";
   }
 
-  metrics.physical_min_width_constraint = metrics.physical_width;
-  metrics.physical_max_width_constraint = metrics.physical_width;
-  metrics.physical_min_height_constraint = metrics.physical_height;
-  metrics.physical_max_height_constraint = metrics.physical_height;
+  if (metrics.physical_width != 0) {
```
robert-ancell: It's not clear to me why this is overriding the values set above - perhaps needs a comment?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -413,6 +425,8 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   // to prevent screen tearing.
   bool NeedsVsync() const;
 
+  bool IsSizedToContent() const;
```
robert-ancell, line 435: Missing documentation comment

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
```diff
@@ -0,0 +1,147 @@
...
+  if (width_ == width && height_ == height) {
+    return;
+  }
+  width_ = width;
+  height_ = height;
+  UpdatePosition();
+}
+
+WindowRect HostWindowTooltip::GetWorkArea() const {
+  WindowRect work_area = {0, 0, 10000, 10000};
```
robert-ancell: Suggest to make constants for max size.

engine/src/flutter/shell/platform/windows/host_window_tooltip.h
```diff
@@ -0,0 +1,47 @@
...
+                          private FlutterWindowsViewSizingDelegate {
+ public:
+  // Creates a tooltip window.
+  HostWindowTooltip(WindowManager* window_manager,
+                    FlutterWindowsEngine* engine,
+                    const BoxConstraints& constraints,
+                    GetWindowPositionCallback get_position_callback,
+                    HWND parent);
+
+  void UpdatePosition();
```
robert-ancell, line 27: Docstring?

examples/multiple_windows/lib/app/window_settings_dialog.dart
```diff
@@ -2,9 +2,28 @@
...
 // found in the LICENSE file.
 
+// ignore_for_file: invalid_use_of_internal_member
+
 import 'package:flutter/material.dart';
+// ignore: implementation_imports
+import 'package:flutter/src/widgets/_window_positioner.dart';
 import 'models.dart';
 
+// Helper to convert enum to display name
```
robert-ancell: Trailing '.'

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+      PlatformDispatcher.instance.engineId!,
+      contentSizeConstraints,
+      _Win32PlatformInterface.getWindowHandle(
+        PlatformDispatcher.instance.engineId!,
+        parent.rootView.viewId,
+      ),
+      _onGetWindowPosition.nativeFunction,
+    );
+    if (viewId < 0) {
+      throw Exception('Windows failed to create a regular window with a valid view id.');
```
robert-ancell: regular -> tooltip

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+  int? handleWindowsMessage(
+    FlutterView view,
+    HWND windowHandle,
+    int message,
+    int wParam,
+    int lParam,
+  ) {
+    if (view.viewId == parent.rootView.viewId) {
+      if (message == _WM_SIZE) {
+        // Tooltips should close when their parent window is deactivated.
```
robert-ancell: Is this comment correct? This is relating to resizing, not deactivating, right?

robert-ancell (dismissed):
Mostly small stuff

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
```diff
@@ -0,0 +1,148 @@
...
+      WindowSize{width_, height_},
+      WindowRect{parent_top_left.x, parent_top_left.y,
+                 parent_bottom_right.x - parent_top_left.x,
+                 parent_bottom_right.y - parent_top_left.y},
+      work_area);
+  SetWindowPos(window_handle_, nullptr, rect->left, rect->top, rect->width,
+               rect->height, SWP_NOACTIVATE | SWP_NOOWNERZORDER);
+  free(rect);
+
+  // The positiner constrained the dimentions more than current size, apply
```
justinmc: positiner => positioner

examples/multiple_windows/lib/app/main_window.dart
```diff
@@ -179,6 +185,10 @@ class _WindowCreatorCard extends StatelessWidget {
                   child: const Text('Regular'),
                 ),
                 const SizedBox(height: 8),
+                TooltipButton(
+                  parentController: windowManager.windows.first.controller,
```
justinmc: Is this a safe way to get the right controller? As always I'm trying to help out the users that will copy this code without reading it.

examples/multiple_windows/lib/app/models.dart
```diff
@@ -61,18 +62,27 @@ class WindowManagerAccessor extends InheritedNotifier<WindowManager> {
...
+
 /// Settings that control the behavior of newly created windows.
 class WindowSettings {
   WindowSettings({
     this.regularSize = const Size(400, 300),
     this.dialogSize = const Size(200, 200),
+    this.positioner = const WindowPositioner(
+      parentAnchor: WindowPositionerAnchor.right,
+      childAnchor: WindowPositionerAnchor.left,
+    ),
```
justinmc, line 75: This seems pretty straightforward to use 👍 . I remember we talked a lot about WindowPositioner in the last PR.

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,176 @@
...
+
+// ignore_for_file: invalid_use_of_internal_member
+// ignore_for_file: implementation_imports
+
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+import 'models.dart';
+
+/// Tracks global rect of an [Element].
+class ElementPositionTracker {
```
justinmc: Make this private if you can.

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,176 @@
...
+  /// compared to last result of [getGlobalRect].
+  void Function(Rect rect)? onGlobalRectChange;
+
+  static bool _persistentCallbackRegistered = false;
+  static final _trackers = <ElementPositionTracker>[];
+
+  static void _register(ElementPositionTracker tracker) {
+    if (!_persistentCallbackRegistered) {
+      _persistentCallbackRegistered = true;
+      WidgetsBinding.instance.addPersistentFrameCallback((_) {
```
justinmc: Avoid the `_` for parameters https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-anonymous-parameter-names

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,176 @@
...
+      onPressed: () {
+        if (_tooltipController != null) {
+          _tooltipController!.destroy();
+          _tooltipTracker?.dispose();
+          setState(() {
+            _tooltipController = null;
+            _tooltipTracker = null;
+          });
+        } else {
+          // Tooltip is not shown, show it
```
justinmc: Period at the end here.

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,176 @@
...
+
+  static void _unregister(ElementPositionTracker tracker) {
+    _trackers.remove(tracker);
+  }
+
+  static void _update() {
+    final trackersCopy = List<ElementPositionTracker>.from(
+      _trackers,
+      growable: false,
+    );
```
justinmc: Is it necessary to make a copy here?

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,176 @@
...
+  void Function(Rect rect)? onGlobalRectChange;
+
+  static bool _persistentCallbackRegistered = false;
+  static final _trackers = <ElementPositionTracker>[];
+
+  static void _register(ElementPositionTracker tracker) {
+    if (!_persistentCallbackRegistered) {
+      _persistentCallbackRegistered = true;
+      WidgetsBinding.instance.addPersistentFrameCallback((_) {
+        _update();
```
justinmc: Why is it necessary to update all trackers and not just this one? I want to make sure that it's necessary to use this singleton pattern with static state. It sometimes makes things hard to test, etc.

examples/multiple_windows/lib/app/tooltip_window_content.dart
```diff
@@ -0,0 +1,98 @@
...
+// ignore_for_file: implementation_imports
+
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+
+class TooltipWindowContent extends StatefulWidget {
+  /// Creates a tooltip window widget.
+  /// [controller] the controller for this window
+  /// [child] the content to render into this window
+  /// [key] the key for this widget
```
justinmc: I know these docs are not super critical inside an example, but typically I would separate each of these lines into its own paragraph with an empty `///` line above and below.

examples/multiple_windows/lib/app/tooltip_window_edit_dialog.dart
```diff
@@ -0,0 +1,234 @@
...
+// found in the LICENSE file.
+
+// ignore_for_file: invalid_use_of_internal_member
+// ignore_for_file: implementation_imports
+
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+import 'package:flutter/src/widgets/_window_positioner.dart';
+
+// Helper to convert enum to display name
```
justinmc: Period here.

examples/multiple_windows/lib/app/tooltip_window_edit_dialog.dart
```diff
@@ -0,0 +1,234 @@
...
+String _anchorToString(WindowPositionerAnchor anchor) {
```
justinmc: Nit: I think I would usually make this a static method on _TooltipWindowEditDialogState, but no big deal either way.

justinmc: Or make it a util somewhere since I see it's duplicated in another file.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [TooltipWindowController], the base class for tooltip windows.
+class TooltipWindowControllerWin32 extends TooltipWindowController
+    implements _WindowsMessageHandler {
+  /// Creates a new tooltip window controller for Win32.
+  ///
+  /// When this constructor completes the native window has been created and
```
justinmc: Nit: Add a comma: "When this constructor completes,"

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+       super.empty() {
+    _owner._addMessageHandler(this);
+    _onGetWindowPosition =
+        ffi.NativeCallable<
+          ffi.Pointer<_Rect> Function(
+            ffi.Pointer<_Size> childSize,
+            ffi.Pointer<_Rect> parentRect,
+            ffi.Pointer<_Rect> outputRect,
+          )
+        >.isolateLocal(_handleOnGetWindowPosition);
```
justinmc: This is the biggest type parameter I've ever seen. Any way to shorten it with a typedef or something? No worries if not, it just tripped me up.

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+  }
+
+  final WindowingOwnerWin32 _owner;
+  final TooltipWindowControllerDelegate _delegate;
+  final BaseWindowController _parent;
+  WindowPositioner _positioner;
+  Rect _anchorRect;
+  bool _destroyed = false;
+
+  ffi.Pointer<_Rect> _handleOnGetWindowPosition(
```
justinmc: Nit: Drop the "On"? So `_handleGetWindowPosition` https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

packages/flutter/lib/src/widgets/_window_win32.dart
```diff
@@ -680,6 +690,246 @@ class DialogWindowControllerWin32 extends DialogWindowController {
...
+  }
+
+  late final ffi.NativeCallable<
+    ffi.Pointer<_Rect> Function(
+      ffi.Pointer<_Size> childSize,
+      ffi.Pointer<_Rect> parentRect,
+      ffi.Pointer<_Rect> outputRect,
+    )
+  >
+  _onGetWindowPosition;
```
justinmc: Another super long type parameter to consider typedefing.

justinmc (commented):
Overall looks good (I mainly reviewed the Dart and only skimmed the C++). I still don't have a working local Windows engine setup so I haven't run this myself.

Was there a TODO in the codebase somewhere that should be removed with this PR or am I thinking of something else?

engine/src/flutter/shell/platform/embedder/embedder.cc
mattkae: @knopp Can you comment on this?

knopp: The original code unconditionally set the constraints to physical size (tight). After the PR this is only done when physical size is specified, otherwise it would always override user specified constraints. 

As for the decision why original code overrode the constraints, I'm not sure to be honest, I just preserved the behavior when not sized to content.

examples/multiple_windows/lib/app/tooltip_window_content.dart
mattkae: I should just simplify it, its a bit superfluous

examples/multiple_windows/lib/app/main_window.dart
mattkae: Good catch! We should use `WindowScope.of(context)`

examples/multiple_windows/lib/app/tooltip_button.dart
mattkae: So this needs to be a singleton because `addPersistentFrameCallback` may only be called once per app, since it can't be unregistered. I have updated the code to make this whole dance a bit more explicit.

examples/multiple_windows/lib/app/tooltip_button.dart
mattkae: I _believe_ it is because the list may be changed from another thread. @knopp do you know why this is?

mattkae:
@justinmc @loic-sharma @robert-ancell This should be good for another look!

engine/src/flutter/shell/platform/embedder/embedder.cc
```diff
@@ -1619,6 +1619,14 @@ MakeViewportMetricsFromWindowMetrics(
 
   metrics.physical_width = SAFE_ACCESS(flutter_metrics, width, 0.0);
   metrics.physical_height = SAFE_ACCESS(flutter_metrics, height, 0.0);
+  metrics.physical_min_width_constraint =
+      SAFE_ACCESS(flutter_metrics, min_width_constraint, 0.0);
```
loic-sharma: If an embedder is using an old embedder API from before constraints were added, we should fallback to tight constraints using the physical width / height sizes.

engine/src/flutter/shell/platform/embedder/embedder.cc
loic-sharma: > As for the decision why original code overrode the constraints, I'm not sure to be honest, I just preserved the behavior when not sized to content.

We did this just so that we could introduce view constraints in `dart:ui` without actually needing to update the embedders. That was just a stopgap solution to get sensible constraint values. This is not a behavior we need to keep.

On the Dart size, I might want to know the current size of my tooltip, as well as how small/big I'm allowed to make the tooltip.

For example, a tooltip could currently have the size 200x200 but be allowed to be size 100x100 to 300x300. I'd want to send a metric with `width`/`height` of `200`, but constraints `100x100` to `300x300`.

loic-sharma: I expect that many embedders will update their embedder.h and not realize they need to add constraints to their metrics.

Could we add a `FML_LOG(ERROR)` message if the constraints are all `0` but the `width` and `height` are non-zero? I don't believe there's any valid use cases for this scenario, but this indicates an embedder that has forgotten to pass constraint values.

engine/src/flutter/shell/platform/embedder/embedder.h
```diff
@@ -1067,6 +1067,14 @@ typedef struct {
   FlutterEngineDisplayId display_id;
   /// The view that this event is describing.
   int64_t view_id;
+  /// Minimum physical width of the window.
```
loic-sharma, line 1073: Could we explain that this must be less than or equal to `width`. Same comment for the other constraint values.

engine/src/flutter/shell/platform/embedder/embedder.cc
mattkae: I logged an error in the case that no constraint is set, so that the platform developer knows to update per embedder.h. I think this is a solid middleground.

engine/src/flutter/shell/platform/embedder/embedder.cc
mattkae: So the current logic is as follows:

1. We set the constraints to the constraints or (if missing) the physical size
2. If the max width/height == min width/height == 0, then we default them to tight constraints again, because a 0 constraint is nonsensical and probably results from an unset value.

engine/src/flutter/shell/platform/embedder/embedder.cc
```diff
@@ -1619,6 +1619,48 @@ MakeViewportMetricsFromWindowMetrics(
...
+           " may be using an outdated version of the FlutterWindowMetricsEvent"
+           " struct. Please update the struct to include these values.";
+  }
+
+  // Set the constraints, defaulting to the physical size if they are not
+  // provided.
+  //
+  // If all of the constraints are zero, we interpret that as no constraints
+  // being provided and default them to the physical size. A constraint of zero
+  // is nonsensical in practice and probably results from an unset field.
```
loic-sharma: Sorry, I think my comment was unclear. I meant to suggest something like:

1. Grab the constraints from the metric. If the metric does not have constraint fields yet, fallback to the physical values.
2. Verify the physical size is within the constraints. If it isn't, return an error.

Also I'm wondering if it'd be better to have an explicit `metrics.has_constraints` flag instead of interpreting `0` values. This approach has the benefit of making forward compatibility easier: it defaults to `false`, which means `0` values still get interpreted as unbounded constraints.

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -29,6 +29,18 @@ namespace flutter {
 // A unique identifier for a view.
 using FlutterViewId = int64_t;
 
+// Optional delegate for views that are sized to contents.
+class FlutterWindowsViewSizingDelegate {
+ public:
+  // This method may be called from raster thread.
+  virtual bool ViewIsSizedToContent() const = 0;
+
+  // These methods will always be called from the main thread.
```
loic-sharma:
```suggestion
  // These methods will always be called from the platform thread.
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -185,6 +200,11 @@ void FlutterWindowsView::ForceRedraw() {
 
 // Called on the platform thread.
 bool FlutterWindowsView::OnWindowSizeChanged(size_t width, size_t height) {
+  if (IsSizedToContent()) {
+    // No resize synchronization needed for views sized to content.
```
loic-sharma, line 204: Would it be worthwhile to assert that the size is within the bounds here?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -29,6 +29,18 @@ namespace flutter {
...
+  virtual Size GetMinimumViewSize() const = 0;
+  virtual Size GetMaximumViewSize() const = 0;
```
loic-sharma: Would it be easier to expose `virtual const BoxConstraints& GetConstraints() const = 0;` here instead?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -247,6 +259,10 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   // Returns true if the view was focused.
   virtual bool Focus();
 
+  void SetSizingDelegate(FlutterWindowsViewSizingDelegate* delegate) {
+    sizing_delegate_ = delegate;
+  }
```
loic-sharma: Could this be a constructor argument?

engine/src/flutter/shell/platform/windows/host_window.cc
```diff
@@ -225,35 +226,39 @@ std::unique_ptr<HostWindow> HostWindow::CreateDialogWindow(
                            parent ? parent : std::optional<HWND>()));
 }
 
+std::unique_ptr<HostWindow> HostWindow::CreateTooltipWindow(
+    WindowManager* window_manager,
+    FlutterWindowsEngine* engine,
+    const WindowConstraints& preferred_constraints,
+    GetWindowPositionCallback get_position_callback,
+    HWND parent) {
+  return std::unique_ptr<HostWindowTooltip>(new HostWindowTooltip(
```
loic-sharma, line 236: Can this be `std::make_unique` instead of `new`?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -413,6 +429,9 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   // to prevent screen tearing.
   bool NeedsVsync() const;
 
+  // If true, the view is sized to its content via a sizing delegate.
```
loic-sharma, line 431: Consider adding something like:

```suggestion
  // If true, the view is sized to its content via a sizing delegate.
  // If false, the view is sized by its parent HWND or by the user.
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -29,6 +29,18 @@ namespace flutter {
 // A unique identifier for a view.
 using FlutterViewId = int64_t;
 
+// Optional delegate for views that are sized to contents.
+class FlutterWindowsViewSizingDelegate {
+ public:
+  // This method may be called from raster thread.
```
loic-sharma: In my first read through I incorrectly assumed this was only called from the raster thread. I'd call out that this can be called from the platform threadd too.

```suggestion
  // This method may be called from the platform or raster threads.
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -154,6 +155,20 @@ bool FlutterWindowsView::OnFrameGenerated(size_t width, size_t height) {
   // Called on the raster thread.
   std::unique_lock<std::mutex> lock(resize_mutex_);
 
+  if (IsSizedToContent()) {
+    if (!ResizeRenderSurface(width, height)) {
```
loic-sharma, line 164: Would it be possible to check that the `width` and `height` satisfy the view's constraints?

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -154,6 +155,20 @@ bool FlutterWindowsView::OnFrameGenerated(size_t width, size_t height) {
...
+      return false;
+    }
+    std::weak_ptr<int> weak_view_alive = view_alive_;
+    engine_->task_runner()->PostTask([this, width, height, weak_view_alive]() {
+      if (weak_view_alive.lock().get() == nullptr) {
+        return;
+      }
+      sizing_delegate_->DidUpdateViewSize(width, height);
+    });
```
loic-sharma: Would it be better to push this logic (switch to UI thread, check the view is alive) down to the sizing delegate? The view is already really complex.

engine/src/flutter/shell/platform/embedder/embedder.cc
```diff
@@ -1619,6 +1619,48 @@ MakeViewportMetricsFromWindowMetrics(
...
+    metrics.physical_min_width_constraint = metrics.physical_width;
+    metrics.physical_max_width_constraint = metrics.physical_width;
+  }
+
+  if (metrics.physical_min_height_constraint ==
+          metrics.physical_max_height_constraint &&
+      metrics.physical_max_height_constraint == 0) {
+    metrics.physical_min_height_constraint = metrics.physical_height;
+    metrics.physical_max_height_constraint = metrics.physical_height;
+  }
```
loic-sharma, line 1637: Could you add a test for the embedder API changes?

Something like:
 https://github.com/flutter/flutter/blob/19633544a4ae1f2f99d31384ec6546d225977d6d/engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc#L4068-L4122

engine/src/flutter/shell/platform/windows/host_window.h
```diff
@@ -103,6 +115,25 @@ class HostWindow {
...
+    WindowArchetype archetype;
+    DWORD window_style;
+    DWORD extended_window_style;
+    const BoxConstraints& box_constraints;
+    Rect const initial_window_rect;
+    LPCWSTR title;
+    std::optional<HWND> const& owner_window;
+    int nCmdShow = SW_SHOWNORMAL;
+    FlutterWindowsViewSizingDelegate* sizing_delegate = nullptr;
+  };
```
loic-sharma, line 130: This is nice, I haven't seen this pattern before 👍

engine/src/flutter/shell/platform/windows/window_manager.cc
```diff
@@ -161,6 +189,26 @@ void InternalFlutterWindows_WindowManager_SetWindowSize(
...
 
+void InternalFlutterWindows_WindowManager_DestroyWindow(HWND hwnd) {
+  flutter::HostWindow* window = flutter::HostWindow::GetThisFromHandle(hwnd);
+  HWND flutter_view_handle = nullptr;
+  if (window) {
+    // First reparent the FlutterView to null parent. Otherwise destroying
+    // the window HWND will immediately destroy the FlutterView HWND as well,
+    // which will cause crash when raster thread tries to reallocate surface.
+    // The FlutterView may only be destroyed safely when
+    // FlutterWindowsEngine::RemoveView finishes.
```
loic-sharma, line 201: Who calls `RemoveView` for hosted windows?

I'm fuzzy on how this works for windows created by the runner app - it looks like `FlutterWindowsViewController::Destroy` calls `FlutterWindowsEngine::RemoveView`. That code path won't be run for hosted windows though? Let me know if I'm misunderstanding something :)

engine/src/flutter/shell/platform/embedder/embedder.cc
mattkae: Yup! That is probably a good idea

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
mattkae: It shouldn't return a const& because the constraints may be calculated on the fly (e.g. if the work area is smaller than we require), but it would make sense to returns constraints

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
mattkae: I don't think so, and I wouldn't want breaking in the event that that happens. Is that ok?

engine/src/flutter/shell/platform/windows/host_window.cc
mattkae: Unfortunately, no. The `HostWindow` has a protected constructor, so unique_ptr can't resolve it. C++ 😄

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
mattkae: @knopp Can you comment here?

engine/src/flutter/shell/platform/windows/window_manager.cc
mattkae: @knopp Can you comment here?

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
mattkae: We sent the constraints down to the view already. By the time we get here, we should assume that it followed those constraints and is sending up a size that matches them

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
knopp: I don't see why not? Should be fairly easy change. The comment for `FlutterWindowsViewSizingDelegate::DidUpdateViewSize` needs to be updated to reflect that it is called from raster thread.

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: I honestly don't know if we have all the wiring in place and working well for runner created views. This is not supported on macOS right now, and while I know that it has been requested on Windows, I'm not sure if it's doable right now.

engine/src/flutter/shell/platform/windows/host_window.cc
loic-sharma: Ahh gotcha, thanks for the explanation!

engine/src/flutter/shell/platform/windows/window_manager.cc
loic-sharma: Runner created views _should_ work as expected and call "remove view" before destroyed the child window.

However, my concern here is with Dart created windows. It looks like nothing calls `FlutterWindowsEngine::RemoveView` for tooltip windows?

The reason I ask: `FlutterWindowsEngine::RemoveView` blocks until the raster thread has confirmed it will not present to the view. We should make sure `RemoveView` completes before we start destroying things. It's possible that `DestroyWindow` is already doing this correctly, but I'd double-check that.

engine/src/flutter/shell/platform/embedder/embedder.h
```diff
@@ -1067,6 +1067,26 @@ typedef struct {
   FlutterEngineDisplayId display_id;
   /// The view that this event is describing.
   int64_t view_id;
+  /// If `true`, the window has size constraints.
+  /// If `false`, the constraint values are ignored.
+  bool has_constraints;
+  /// Minimum physical width of the window.
+  ///
+  //// This must be less than or equal to |max_width_constraint| and |width|.
```
loic-sharma: Consider clarifying for this and below:

```suggestion
  //// If |has_constraints| is `true, this must be less than or equal to |max_width_constraint| and |width|.
```

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: I'm bit confused here. Isn't `FlutterWindowsEngine::RemoveView` called through `FlutterWindowsViewController` destructor?

knopp: The `FlutterWindowsViewController` should get destroyed when the `HostWindow` holding it gets destroyed, which should be inside `WindowManager::HandleMessage` on `WM_NCDESTROY`.

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
loic-sharma: The problem is that the rasterization is asynchronous. The raster thread **attempts** to drop frames whose sizes are incorrect, but it is ultimately the embedder's responsibility to verify the presented content's size is correct.

For example:

1. I have a content sized view with constraints 0x0 to 100x100
2. The framework produces a frame of size 50x50
3. The raster thread checks 50x50 is within the constraints. It is. It begins rasterizing the frame.
4. The platform thread changes the content sized view's constraints to 0x0 to 20x20
5. The raster thread finishes rasterizing the frame and asks the embedder to present content of size 50x50 to a view whose max size is 20x20.

The embedder is responsible with ensuring that the content is dropped. Please let me know if I misunderstood something though!

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -29,6 +29,19 @@ namespace flutter {
...
+// Optional delegate for views that are sized to contents.
+class FlutterWindowsViewSizingDelegate {
+ public:
+  // This method may be called from the platform or raster threads.
+  virtual bool ViewIsSizedToContent() const = 0;
+
+  // This method is called from the platform thread.
+  virtual BoxConstraints GetConstraints() const = 0;
+
+  // This method is called from the raster thread.
```
loic-sharma:
```suggestion
  // This method is called from the raster thread
  // after the view's surface has been resized but
  // before the frame has been presented on the view.
```

engine/src/flutter/shell/platform/windows/window_manager.h
```diff
@@ -38,6 +38,19 @@ struct WindowConstraints {
...
 
+// Coordinates are in physical pixels.
+struct WindowRect {
+  int32_t left;
+  int32_t top;
+  int32_t width;
+  int32_t height;
+};
+
+struct WindowSize {
```
loic-sharma, line 50: Consider adding just to be super clear:

```suggestion
// Sizes are in physical pixels.
struct WindowSize {
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -413,6 +427,10 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   // to prevent screen tearing.
   bool NeedsVsync() const;
 
+  // If true, the view is sized to its content via a sizing delegate.
+  // If false, the view is sized by its parent HWND or by the user.
```
loic-sharma, line 432:
```suggestion
  // If false, the view is sized by its parent HWND or by the user.
  //
  // This method can be called from the platform or raster threads.
```

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
knopp: The reason for ignoring wrong size during rasterization is to ensure that the resizing is as fast as possible and the content keeps up with window size.

But for changing constraints when sized to contents the worst thing that will happen is that you see one extra frame with old size. I'm not sure if this is a bad thing. Is the constraint change expected to be frequent? And what if we change the constraints in animation, but then during animation the intermediate frames get dropped because the size no longer matches new constraints, wouldn't this cause bunch of missed frames?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -413,6 +427,10 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
...
+  bool IsSizedToContent() const;
```
loic-sharma, line 435: Before this PR, the view managed the size data that could be mutated by the platform thread but read by both the the platform or raster threads. In my opinion, centralizing this sizing data makes it easier to reason about the thread-safety.

Currently with this PR, some of the sizing data is in `FlutterWindowsViewSizingDelegate`.

I wonder if it'd be better if the `FlutterWindowsView` was the source of truth for the view's:

1. Mode: parent-sized or content-sized
2. View's size
3. View's constraints, if content-sized

The view would use this information to present content. It would take care of making sure reads/writes to this data are thread-safe.

The `HostWindowTooltip` would still have its own separate  constraints and size, but it wouldn't need to worry about reads from the raster thread. `FlutterWindowsViewSizingDelegate` would just be used to receive view resize notifications.

What are your thoughts?

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: We don't. The view gets destroyed inside `FlutterWindowsViewController::Destroy()` (`view_.reset();`) after `FlutterWindowsEngine::RemoveView` completes. What am I missing here?

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
knopp: The counter argument to this, that I think you made at different place is that `FlutterWindowsView` is already quite complex. If I understand what you're suggesting is that instead of `FlutterView` pulling the constraints from the delegate, the class implementing the delegate would be pushing those to the view?

engine/src/flutter/shell/platform/windows/window_manager.cc
```diff
@@ -161,6 +189,26 @@ void InternalFlutterWindows_WindowManager_SetWindowSize(
...
+    flutter_view_handle = window->GetFlutterViewWindowHandle();
+    ShowWindow(flutter_view_handle, SW_HIDE);
+    SetParent(flutter_view_handle, nullptr);
+  }
+  DestroyWindow(hwnd);
+  if (flutter_view_handle) {
+    // Now the flutter view HWND can be destroyed safely.
+    DestroyWindow(flutter_view_handle);
```
loic-sharma, line 209: Could you expand on why we destroy the parent window before we destroy the child window?

For runner created views we destroy the child window before the parent window: https://github.com/flutter/flutter/blob/9c47d57dbb38ec552959b2272ab31fdff911ba87/examples/hello_world/windows/runner/flutter_window.cpp#L47-L51

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: Note that there are two things here - destroying `FlutterWindowsView` and destroying the underlying HWND. The comment refers to destroying the underlying `FlutterWindowsView ` HWND prematurely (during destruction of parent's HWND), before we had a chance to destroy `FlutterWindowsView`. In other words, `FlutterWindowsView` HWND got destroyed while `FlutterWindowsView` was still alive and raster threads attempted to recreate the surface (which crashed because of invalid HWND).

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
loic-sharma: I don't strongly about this one, marking resolved.

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
knopp: One of the issues here is that getting the constraints is non-trivial, as the positioner can enforce additional constraints on top of user constraints (i.e. if something doesn't fit the screen). Right now the single source of truth for the constraints is `FlutterWindowsViewSizingDelegate`. If we want to push the constraints to the view itself, we'd be introducing a state for something that can be easily derived. Personally I prefer to avoid doing that.

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
```diff
@@ -0,0 +1,165 @@
...
+      .nCmdShow = SW_SHOWNOACTIVATE,
+      .sizing_delegate = this,
+  });
+  SetWindowLongPtr(window_handle_, GWLP_HWNDPARENT,
+                   reinterpret_cast<LONG_PTR>(parent_));
+}
+
+bool HostWindowTooltip::ViewIsSizedToContent() const {
+  return box_constraints_.biggest().width() != 0 &&
+         box_constraints_.biggest().height() != 0;
```
loic-sharma: Could you remind me when a tooltip window would have `0,0` max constraints? Consider adding a comment here about this scenario, it's not immediately obvious to me why we'd want to disable content sizing here.

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: Actually, we don't. This is called from WM_DESTROY. That means the child HWND is already destroyed by the time we reset `flutter_controller_`. This is racy and may crash if we reallocate the surface when closing window (because the flutter view HWND has been destroyed already). 

It probably slipped under the radar, because you would need to be manually resizing the window while destroying it, which is very unlikely the happen with regular windows. I only noticed the crash because I had the tooltip window being resized in animation and then closing it programatically.

What happens when you close window is `WM_CLOSE` -> Default handler calls `DestroyWindow(TopLevelHWND)`.

`DestroyWindow(TopLevelHWND`)` does two things: 
- destroy all child HWNDs
- produce `WM_DESTROY`, in which we reset `flutter_controller`_.

Unfortunately the order of the operations is that the child HWNDs are destroyed first.

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
knopp: Right now `BoxConstraints(0,0,0,0)` simply means special value for manually sized windows, assumption being there is no good reason why user would want content laid out at tight constraints with zero width and height. We can make an explicit boolean for that.

engine/src/flutter/shell/platform/windows/window_manager.cc
knopp: We might want to see if calling `FlutterWindowsEngine::RemoveView` from `WM_DESTROY` on FlutterView's HWNDs would solve this, but I'd file that as a separate issue.

engine/src/flutter/shell/platform/embedder/embedder.cc
knopp: What does "current size" mean? The size of last produced surface? How would that be accessed on dart side (i.e. during layout)? Right now specifying size results in tight constraints, while specifying the constraints may result in loose constraints. If you specify both size and constraints how would this be accessed during layout?

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
loic-sharma: An explicit boolean sounds good to me 👍

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
loic-sharma: > But for changing constraints when sized to contents the worst thing that will happen is that you see one extra frame with old size. I'm not sure if this is a bad thing.

Eventually, we'll want to support content-sized views in add-to-app. The Flutter view might be nested in some native Windows app content. If the native Windows app updates the Flutter view's constraints from a max size of 200x200 to a max size of 100x100, the Flutter view should respect that and drop frames that would exceed its new constraints.

> Is the constraint change expected to be frequent? 

Off-hand, I don't think so, no.

> And what if we change the constraints in animation, but then during animation the intermediate frames get dropped because the size no longer matches new constraints, wouldn't this cause bunch of missed frames?

Can we make this work the same way as window resizing? The platform thread would try to wait until it receives a frame that meets the new constraints before letting the constraints change again.

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
mattkae: I will add it

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
knopp: That seems like a bit of hypothetical scenario right now. Also the platform thread can't "just wait" until it receives a frame. For resizing we block in a particular event (WM_SIZE IIRC), as a best effort to try to delay the compositor updating the window border until content is available. I don't quite see the parallel with views sized to content. When exactly would be "block the platform thread"? 

Also we don't really block the platform thread, we pump the event loop until content is available, which is relatively okay to do inside WM_SIZE, but can be prone to reentrancy issues if we do at some random place.

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
mattkae: Done!

engine/src/flutter/shell/platform/windows/window_manager.cc
loic-sharma: > This is called from WM_DESTROY.

Ah thanks, this was the key bit that I was missing. From the name `InternalFlutterWindows_WindowManager_DestroyWindow` I incorrectly assumed that this method is called when the Dart apps requests to destroy the window. I didn't realize this was in response to the `WM_DESTROY` message on the host window.

Thanks for the detailed explanation!

engine/src/flutter/shell/platform/windows/window_manager.h
```diff
@@ -166,8 +199,14 @@ void InternalFlutterWindows_WindowManager_SetFullscreen(
     HWND hwnd,
     const flutter::FullscreenRequest* request);
 
+FLUTTER_EXPORT
+void InternalFlutterWindows_WindowManager_DestroyWindow(HWND hwnd);
```
loic-sharma: Consider:

```suggestion
// Invoked by the framework when the host window receives WM_DESTROY.
FLUTTER_EXPORT
void InternalFlutterWindows_WindowManager_OnDestroyWindow(HWND hwnd);
```

engine/src/flutter/shell/platform/windows/window_manager.cc
loic-sharma: Thanks for the explanations. I misunderstood when this method was called, this comment addressed my concerns: https://github.com/flutter/flutter/pull/179147/files#r2633679265

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
loic-sharma: > If I understand what you're suggesting is that instead of `FlutterView` pulling the constraints from the delegate, the class implementing the delegate would be pushing those to the view?

Yup exactly.

> The counter argument to this, that I think you made at different place is that `FlutterWindowsView` is already quite complex. 

I agree, I do think we should split up `FlutterWindowsView`. The abstractions in the Windows embedder are far from ideal.

However, I'd prefer that we have very clear boundaries around the data / logic the raster thread can access, especially if that data can also be written by the platform thread. In my mind we'd be able to keep these boundaries clearer if we introduce this new state instead of directly relying on the sizing delegate.

engine/src/flutter/shell/platform/embedder/embedder.cc
loic-sharma: > What does "current size" mean? The size of last produced surface?

Yes that's right.

> How would that be accessed on dart side (i.e. during layout)?

For example, during the build phase my app might make the current window bigger if the view's size is too small for some content I'd like to display.

> If you specify both size and constraints how would this be accessed during layout?

Layout [only uses the view's constraints](https://github.com/flutter/flutter/blob/3c0914480ca45d04dac87375cde5cce768249e5a/packages/flutter/lib/src/rendering/view.dart#L289-L296). Layout effectively ignores the view's size.

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
loic-sharma: > That seems like a bit of hypothetical scenario right now.

That's fair. After all, desktop doesn't support add-to-app yet :)

That said, Gemini's visual layout mode (see this [blog post](https://blog.google/products/gemini/gemini-3-gemini-app/)) does use add-to-app to embed  content-sized Flutter views on mobile and web. I don't know the Gemini team's plans, but they might need this scenario if they decide to create native desktop apps.

> When exactly would be "block the platform thread"?

I didn't have anything particular in mind, just spit balling.

> Also we don't really block the platform thread

Yes I'm aware, hence why I wrote "wait" instead of "block" :)

---

It does seem like we have a mismatch here. If the Dart app updates a window's constraints, these shouldn't be enforced retroactively for frames that are already in the pipeline. However, if the native app updates the Flutter view's constraints, these should be enforced immediately. I'm OK punting on this this problem for now since desktop doesn't support add-to-app today.

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
mattkae: I agree that, for now, it makes sense to move this information to the View instead of the Delegate. The Delegate should just be called to notify of a change.

engine/src/flutter/shell/platform/windows/flutter_windows.cc
```diff
@@ -92,8 +92,8 @@ static FlutterDesktopViewControllerRef CreateViewController(
     engine = std::unique_ptr<flutter::FlutterWindowsEngine>(engine_ptr);
   }
 
-  std::unique_ptr<flutter::FlutterWindowsView> view =
-      engine_ptr->CreateView(std::move(window_wrapper));
+  std::unique_ptr<flutter::FlutterWindowsView> view = engine_ptr->CreateView(
+      std::move(window_wrapper), false, flutter::BoxConstraints());
```
loic-sharma, line 96: For readability, consider adding a comment to hint the parameter name here and in other places that pass a value for this:

```suggestion
      std::move(window_wrapper), /*is_sized_to_content=*/false, flutter::BoxConstraints());
```

(Alternatively consider making this an enum, but up to you, I'm good with either approach)

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
loic-sharma: Thanks everyone for the wonderful discussion. I'll mark this as resolved since we decided to punt this (with approval from our overlord Gemini!)

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -372,8 +392,26 @@ void FlutterWindowsView::SendWindowMetrics(size_t width,
   FlutterEngineDisplayId display_id = binding_handler_->GetDisplayId();
   FlutterWindowMetricsEvent event = {};
   event.struct_size = sizeof(event);
-  event.width = width;
-  event.height = height;
+  if (IsSizedToContent()) {
+    auto const constraints = GetConstraints();
+    event.has_constraints = true;
```
loic-sharma: Looks like this is missing the width/height values here.

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -848,4 +906,22 @@ bool FlutterWindowsView::NeedsVsync() const {
   return !windows_proc_table_->DwmIsCompositionEnabled();
 }
 
+bool FlutterWindowsView::IsSizedToContent() const {
+  return is_sized_to_content_;
+}
+
+BoxConstraints FlutterWindowsView::GetConstraints() const {
+  Size smallest = box_constraints_.smallest();
+  Size biggest = box_constraints_.biggest();
```
loic-sharma, line 903: My understanding is this is unbounded constraints by default? If we're not sized to content, should we instead return constraints that are the current width/height? (If we do this change, I imagine that'll let us simplify the window metric logic too)

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -38,7 +47,10 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   FlutterWindowsView(
       FlutterViewId view_id,
       FlutterWindowsEngine* engine,
+      bool is_sized_to_content,
+      const BoxConstraints& box_constraints,
       std::unique_ptr<WindowBindingHandler> window_binding,
+      FlutterWindowsViewSizingDelegate* sizing_delegate = nullptr,
       std::shared_ptr<WindowsProcTable> windows_proc_table = nullptr);
```
loic-sharma, line 57: Maybe consider making the `is_sized_content` and `box_constraints` arguments optional? Just a thought, feel free to skip, I don't feel strongly.

engine/src/flutter/shell/platform/windows/window_manager.h
loic-sharma: @mattkae Re-opening this one as I think this might've gotten missed: what do you think of adding the `On` prefix to `OnDestroyWindow`? I think it makes it a little clearer that this is called in response to `WM_DESTROY`.

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,183 @@
...
+        growable: false,
+      );
+      for (final tracker in trackersCopy) {
+        tracker._updateSelf();
+      }
+    });
+  }
+
+  static final _instance = _ElementPositionTrackerManager._();
+  factory _ElementPositionTrackerManager() => _instance;
```
loic-sharma: Using a `factory` like this isn't idiomatic, I'd make the `instance` public and then reference that directly. For example, [`WidgetsBinding.instance`](https://api.flutter.dev/flutter/widgets/WidgetsBinding/instance.html).

Then `_ElementPositionTrackerManager().foo(...)` would become `_ElementPositionTrackerManager.instance.foo(...)`.

examples/multiple_windows/lib/app/tooltip_button.dart
```diff
@@ -0,0 +1,183 @@
...
+      });
+    } else {
+      // Tooltip is not shown, show it.
+      final tracker = _ElementPositionTracker(
+        element: _tooltipButtonKey.currentContext!,
+      );
+      _ElementPositionTrackerManager().add(tracker);
+      final UniqueKey key = UniqueKey();
+      final controller = TooltipWindowController(
+        anchorRect: tracker.getGlobalRect()!,
```
loic-sharma, line 133: Sorry, I know I've asked before but I've forgotten 😅. What's our plan again for `anchorRect` long-term to replace `_ElementPositionTrackerManager`?

examples/multiple_windows/lib/app/tooltip_window_content.dart
```diff
@@ -0,0 +1,95 @@
...
+  late final AnimationController _animationController;
+
+  @override
+  void initState() {
+    super.initState();
+    _animationController = AnimationController(
+      duration: const Duration(seconds: 1),
+      vsync: this,
+    );
+    _animationController.repeat(reverse: true);
```
loic-sharma: I think you can simplify this using the brand new [`RepeatingAnimationBuilder`](https://main-api.flutter.dev/flutter/widgets/RepeatingAnimationBuilder-class.html) 👀 

That should let you switch to a `StatelessWidget`, without a ticker provider nor an animation controller.

examples/multiple_windows/lib/app/window_settings_dialog.dart
```diff
@@ -59,23 +86,335 @@ class _WindowSettingsEditorState extends State<_WindowSettingsEditor> {
...
+                    }
+                  },
+                ),
+              ),
+              ListTile(
+                contentPadding: const EdgeInsets.symmetric(horizontal: 16),
+                title: const Text('Child Anchor'),
+                subtitle: DropdownButton<WindowPositionerAnchor>(
+                  isExpanded: true,
+
```
loic-sharma, line 112:
```suggestion
```

examples/multiple_windows/lib/app/window_settings_dialog.dart
```diff
@@ -59,23 +86,335 @@ class _WindowSettingsEditorState extends State<_WindowSettingsEditor> {
...
+                  ],
+                ),
+              ),
+              const Divider(),
+              ListTile(
+                contentPadding: const EdgeInsets.symmetric(horizontal: 16),
+                title: const Text('Parent Anchor'),
+                subtitle: DropdownButton<WindowPositionerAnchor>(
+                  isExpanded: true,
+
```
loic-sharma:
```suggestion
```

examples/multiple_windows/lib/app/window_settings_dialog.dart
```diff
@@ -59,23 +86,335 @@ class _WindowSettingsEditorState extends State<_WindowSettingsEditor> {
...
+                  value: _parentAnchor,
+                  items: WindowPositionerAnchor.values.map((anchor) {
+                    return DropdownMenuItem(
+                      value: anchor,
+                      child: Text(anchorToString(anchor)),
+                    );
+                  }).toList(),
```
loic-sharma: FYI, you can also do this:

```suggestion
                  items: [
                    for (WindowPositionerAnchor value in WindowPositionerAnchor.values)
                      DropdownMenuItem(
                        value: anchor,
                        child: Text(anchorToString(anchor)),
                      ),
                  ],
```

Just a heads up in case you weren't aware. Feel free to keep as-is if that's your preference :)

examples/multiple_windows/lib/app/window_settings_dialog.dart
```diff
@@ -59,23 +86,335 @@ class _WindowSettingsEditorState extends State<_WindowSettingsEditor> {
...
+                child: Row(
+                  mainAxisAlignment: MainAxisAlignment.end,
+                  children: [
+                    TextButton(
+                      onPressed: widget.onClose,
+                      child: const Text('Cancel'),
+                    ),
+                    const SizedBox(width: 12),
+                    FilledButton(
+                      onPressed: () {
```
loic-sharma: This build method is quite big already, I'd consider moving this apply logic to a separate method.

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
mattkae: Updated! `GetWorkArea` still logically lives on the tooltip window, but everything else concerning the data of "how" a view is constrained has been moved to the view

engine/src/flutter/shell/platform/windows/window_manager.h
mattkae: Oh, whoops. I thought the comment was just about adding the comment. I will add that!

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
mattkae: I'm going to leave them explicit for now, because I fear that people might forget about them and then be surprised by the behavior. If it ends up being a burden, we can change it it in the future.

I did, however, place them after the `window_binding` in the latest update, so the option is open.

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
mattkae: True, that will make things easier to reason about at this level.

examples/multiple_windows/lib/app/tooltip_button.dart
mattkae: Our idea for now was to leave this as the simplest way to do things until we have a reason to optimize otherwise. The issue is that most use cases will simply hide the tooltip if the parent widget moves in some way. This tends to be the common case across platforms. For this reason, we are wary of investing too much time in a complicated solution

mattkae:
@loic-sharma Should be ready for a re-review!

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
loic-sharma: That sounds good to me, thanks!

engine/src/flutter/shell/platform/windows/flutter_windows_view.cc
```diff
@@ -372,8 +392,23 @@ void FlutterWindowsView::SendWindowMetrics(size_t width,
...
+      static_cast<size_t>(constraints.smallest().width());
+  event.min_height_constraint =
+      static_cast<size_t>(constraints.smallest().height());
+  event.max_width_constraint =
+      static_cast<size_t>(constraints.biggest().width());
+  event.max_height_constraint =
+      static_cast<size_t>(constraints.biggest().height());
+  if (IsSizedToContent()) {
+    event.width = event.min_width_constraint;
+    event.height = event.min_height_constraint;
```
loic-sharma: The width and height here should be the view's current width and height (see [`FlutterView.physicalSize`](https://api.flutter.dev/flutter/dart-ui/FlutterView/physicalSize.html)). Should this be `event.width = width` and `event.height = height`?

Same comment below for `CreateWindowMetricsEvent`

engine/src/flutter/shell/platform/windows/flutter_windows_view.h
```diff
@@ -455,6 +479,15 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate {
   // The accessibility bridge associated with this view.
   std::shared_ptr<AccessibilityBridgeWindows> accessibility_bridge_;
 
+  // If `true`, the view is sized to its content via a sizing delegate.
```
loic-sharma, line 482: Consider adding something like:

```suggestion
  // If `true`, the view is sized to its content via a sizing delegate.
  // If `false`, the view is sized by its parent HWND.
```

I don't feel strongly though :)

engine/src/flutter/shell/platform/windows/host_window_tooltip.cc
```diff
@@ -0,0 +1,136 @@
...
+      .owner_window = parent,
+      .nCmdShow = SW_SHOWNOACTIVATE,
+      .sizing_delegate = this,
+      .is_sized_to_content = is_sized_to_content});
+  SetWindowLongPtr(window_handle_, GWLP_HWNDPARENT,
+                   reinterpret_cast<LONG_PTR>(parent_));
+}
+
+void HostWindowTooltip::DidUpdateViewSize(int32_t width, int32_t height) {
+  std::weak_ptr<int> weak_view_alive = view_alive_;
```
loic-sharma, line 48: I suspect we'll forget about the threading considerations here, so just so that it's obvious consider adding:

```suggestion
  // This method is called on the raster thread.
  std::weak_ptr<int> weak_view_alive = view_alive_;
```

examples/multiple_windows/lib/app/tooltip_button.dart
loic-sharma: Gotcha, thanks for the details!

loic-sharma (approved):
Excellent work! :)

PR #179162, opened 2025-11-26
[web] Add clone method to LayerPicture and dispose pictures in PictureLayer

### LINKED ISSUE #82878: Implement Layer.dispose for CanvasKit layers
The framework calls this when it is done with a particular layer. On the VM side, we can't dispose of all native resources at this point because something else might still need to hold on to the raster layer to rasterize a scene on the raster thread. On CanvasKit, it's slightly less obvious to me what we need to do - multi-threaded races aren't a concern, but we need to figure out when and whether to release resources on the native side.

/cc @yjbanov @hterkelsen

harryterkelsen:
Fixes https://github.com/flutter/flutter/issues/82878

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer_scene_builder.dart
```diff
@@ -64,7 +64,11 @@ class LayerSceneBuilder implements ui.SceneBuilder {
     bool isComplexHint = false,
     bool willChangeHint = false,
   }) {
-    currentLayer.add(PictureLayer(picture as LayerPicture, offset, isComplexHint, willChangeHint));
+    // Add a clone of the picture to the layer tree so that the original picture
+    // can be disposed.
+    currentLayer.add(
+      PictureLayer((picture as LayerPicture).clone(), offset, isComplexHint, willChangeHint),
```
mdebbar: It really feels like a `picture` should be a `CountedRef`, and `PictureLayer` should somehow do `picture.ref(this);`. That would avoid the unnecessary cloning of pictures, and the `_isClone` thing.

But (ugh!) `CountedRef` is designed to only work with C++/JSObject, so... probably outside the scope of this PR.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/picture.dart
```diff
@@ -85,9 +93,11 @@ class CkPicture implements LayerPicture {
       _debugDisposalStackTrace = StackTrace.current;
       return true;
     }());
-    ui.Picture.onDispose?.call(this);
+    if (!_isClone) {
+      ui.Picture.onDispose?.call(this);
+    }
```
mdebbar, line 98: By doing it this way, we are reporting the disposal of the picture as soon as the framework calls `dispose` on the original picture, even though the "picture" is still alive in the engine. This probably defeats the purpose of the callback.

I'm guessing the purpose of the callback is to report when the picture is actually disposed, i.e. when all clones have been disposed and the native object is released.

harryterkelsen: This matches how the native engine does it: https://github.com/flutter/flutter/blob/d57f6f81c67b1d27298515d4c0dd790cf8dc92e3/engine/src/flutter/lib/ui/painting.dart#L7658

Once `Picture.dispose()` is called, the framework has no way to access the picture anymore.

The reason the native engine can call `_dispose` right away and doesn't need to create a clone of the image is that the native implementation of Picture.dispose just derefs the backing display_list, and when addPicture is called, it creates a new reference to the backing display_list. On the web, the Picture object is both how we represent the underlying Skia picture to the framework and how we hold the reference to the Skia picture ourselves, so we `clone` the Picture

engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer_scene_builder.dart
harryterkelsen: Yeah, there's a bit of clunkiness because we use our implementations of the `dart:ui` classes to be both how we represent the Skia objects to the framework but also to be how the engine holds onto the Skia objects itself. In the native engine there is another layer of abstraction in the C++ UI library between the `dart:ui` types and the underlying Skia objects.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/picture.dart
mdebbar: Got it, I see now. Okay if it matches the native behavior, then sounds good!

engine/src/flutter/lib/web_ui/test/ui/platform_view_test.dart
```diff
@@ -866,23 +866,26 @@ Future<void> testMain() async {
...
 
       implicitView.debugPhysicalSizeOverride = const ui.Size(100, 100);
       implicitView.debugForceResize();
-      await renderScene(sb.build());
+      await renderScene(testScene());
       _expectSceneMatches(<_EmbeddedViewMarker>[_overlay, _platformView]);
 
       implicitView.debugPhysicalSizeOverride = const ui.Size(200, 200);
       implicitView.debugForceResize();
-      await renderScene(sb.build());
```
mdebbar, line 885: This should've been failing all along. According to the `SceneBuilder.build` [docs](https://api.flutter.dev/flutter/dart-ui/SceneBuilder/build.html):

> After calling this function, the scene builder object is invalid and cannot be used further.

Can we make `LayerSceneBuilder.build` explicitly throw when it's called more than once? (not necessarily in this PR).

engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer.dart
```diff
@@ -22,10 +22,10 @@ abstract class Layer implements ui.EngineLayer {
   /// Implement layer visitor.
   R accept<R>(LayerVisitor<R> visitor);
 
-  // TODO(dnfield): Implement ui.EngineLayer.dispose for CanvasKit.
-  // https://github.com/flutter/flutter/issues/82878
   @override
-  void dispose() {}
+  void dispose() {
+    // This is a no-op for all Layer types except PictureLayer.
+  }
```
mdebbar: I'm not seeing `ui.EngineLayer.dispose` being called inside the engine, so I'm assuming that the framework is calling it when it's done rendering the scene.

If that's the case, why not override `RootLayer.dispose` and do the `clonedPictures` cleanup there?

engine/src/flutter/lib/web_ui/test/common/rendering.dart
```diff
@@ -22,12 +22,15 @@ void setUpRenderingForTests() {
       EnginePlatformDispatcher.instance
           .render(_sceneToRender!)
           .then<void>((_) {
+            _sceneToRender?.dispose();
+            _sceneToRender = null;
```
mdebbar, line 26: Good candidate for `whenComplete`?

mdebbar (approved):
LGTM

engine/src/flutter/lib/web_ui/test/common/rendering.dart
harryterkelsen: Done

engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer.dart
harryterkelsen: That's basically what I changed it to, except I overrode `dispose()` in `ContainerLayer` and dispose of any straggling PictureLayers there.

engine/src/flutter/lib/web_ui/test/ui/platform_view_test.dart
harryterkelsen: Yeah. Let's make it so that the `SceneBuilder` disposes itself after `build()` is called, which is basically what the native engine does https://github.com/flutter/flutter/blob/8e952d62976dfe77e54849e9e110e5ebb7ef7f71/engine/src/flutter/lib/ui/compositing/scene_builder.cc#L298

I'll add a TODO

engine/src/flutter/lib/web_ui/test/common/rendering.dart
```diff
@@ -21,13 +21,18 @@ void setUpRenderingForTests() {
...
+          .whenComplete(() {
+            _sceneToRender?.dispose();
+            _sceneToRender = null;
+          })
           .then<void>((_) {
             _sceneCompleter?.complete();
           })
           .catchError((Object error) {
+            _sceneToRender?.dispose();
+            _sceneToRender = null;
```
mdebbar: This is not necessary here since it's already done in `whenComplete`

engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer.dart
```diff
@@ -53,6 +63,11 @@ class RootLayer extends ContainerLayer {
   R accept<R>(LayerVisitor<R> visitor) {
     return visitor.visitRoot(this);
   }
+
+  /// The pictures that were cloned into this layer tree.
+  ///
+  /// They need to be disposed of after the tree has been rendered.
+  final List<LayerPicture> clonedPictures = <LayerPicture>[];
```
mdebbar: I think this isn't needed anymore?

mdebbar (approved):
LGTM!

engine/src/flutter/lib/web_ui/test/common/rendering.dart
harryterkelsen: Oops! Good catch

engine/src/flutter/lib/web_ui/lib/src/engine/layer/layer.dart
harryterkelsen: Right! Done

PR #179163, opened 2025-11-27
Fix crash when text editing value changes between scrolls

### LINKED ISSUE #179164: `EditableText` crashes when text editing value changes between consecutive scrolls
### Steps to reproduce

1. Run the test below, it will crash with a `type error, null check operator used on a null value`.

```dart
testWidgets(
  'Does not crash when editing value changes between consecutive scrolls',
  (WidgetTester tester) async {
    // Regression test for https://github.com/flutter/flutter/issues/
    final controller = TextEditingController(text: 'text ' * 10000);

    await tester.pumpWidget(
      MaterialApp(
        home: Material(child: TextField(controller: controller, maxLines: null)),
      ),
    );

    final Finder textField = find.byType(TextField);
    final EditableTextState editableTextState = tester.state<EditableTextState>(
      find.byType(EditableText),
    );
    // Long press to select the first word and show the toolbar.
    await tester.longPressAt(textOffsetToPosition(tester, 0));
    await tester.pumpAndSettle();
    expect(editableTextState.selectionOverlay?.toolbarIsVisible, true);
    expect(controller.selection, const TextSelection(baseOffset: 0, extentOffset: 4));

    // Scroll down so selection is not visible, and toolbar is scheduled to be shown
    // when the selection is once again visible.
    final TestGesture gesture = await tester.startGesture(tester.getCenter(textField));
    await gesture.moveBy(const Offset(0.0, -200.0));
    await tester.pump();
    await gesture.up();

    // Scroll again before the post-frame callback from the first scroll is run to invalidate
    // the data from the first scroll.
    controller.value = const TextEditingValue(text: 'a different value');

    await gesture.down(tester.getCenter(textField));
    await gesture.moveBy(const Offset(0.0, -100.0));
    await tester.pump();
    await gesture.up();
    await tester.pump();
    // This test should reach the end without crashing.
  },
  variant: TargetPlatformVariant.only(TargetPlatform.android),
);
```

### Expected results

Running the test above should pass.

### Actual results

Test above, crashes with a `type error, null check operator used on a null value`.

### Root cause

If the text editing value changes between two scrolls, when the second scroll ends `_dataWhenToolbarShowScheduled` will be invalidated. If the second scroll ends before the post frame callback scheduled by the first scroll has a chance to execute, there will be a crash because `_dataWhenToolbarShowScheduled` is being accessed while null.

Renzo-Olivares:
This change fixes a crash in `EditableText` that occurs when the text editing value changes between two scrolls, when the second scroll ends `_dataWhenToolbarShowScheduled` will be invalidated. If the second scroll ends before the post frame callback scheduled by the first scroll has a chance to execute, there will be a crash because `_dataWhenToolbarShowScheduled` is being accessed while null.

This change also invalidates a scheduled toolbar if the text editing value has changed before the post-frame callback has a chance to run.

Fixes #179164

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/editable_text.dart
+++ b/packages/flutter/lib/src/widgets/editable_text.dart
@@ -4231,7 +4231,13 @@ class EditableTextState extends State<EditableText>
       _showToolbarOnScreenScheduled = true;
       SchedulerBinding.instance.addPostFrameCallback((Duration _) {
         _showToolbarOnScreenScheduled = false;
-        if (!mounted) {
+        if (!mounted || _dataWhenToolbarShowScheduled == null) {
+          return;
+        }
+        if (_dataWhenToolbarShowScheduled!.value != _value) {
+          // Value has changed so we should invalidate any toolbar scheduling.
+          _dataWhenToolbarShowScheduled = null;
+          _disposeScrollNotificationObserver();
           return;
         }
         final Rect deviceRect = _calculateDeviceRect();
--- a/packages/flutter/test/cupertino/text_field_test.dart
+++ b/packages/flutter/test/cupertino/text_field_test.dart
@@ -10821,4 +10821,92 @@ void main() {
     variant: TargetPlatformVariant.all(),
     skip: kIsWeb, // [intended] on web the browser handles the context menu.
   );
+
+  testWidgets(
+    'Does not crash when editing value changes between consecutive scrolls',
+    (WidgetTester tester) async {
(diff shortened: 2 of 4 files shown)


justinmc (approved):
LGTM 👍 

One nit: Are you sure it's worth it to include tests for TextField and CupertinoTextField, or do the tests on EditableText cover everything already?

loic-sharma (approved):

PR #179168, opened 2025-11-27
Support removing the ability to expand/collapse ExpansionPanel (#179167)

### LINKED ISSUE #179167: Allow ExpansionPanel with null body
### Use case

Some items in my ExpansionPanelList have no body. SizedBox will be the status quo for this situation, but it displays the expand/collapse icon, which is misleading. 

### Proposal

Allow a null body, which hides the expand/collapse icon.

benji-farquhar:
Some items in my ExpansionPanelList have no body. SizedBox will be the status quo for this situation, but it displays the expand/collapse icon, which is misleading.

Allow ExpansionPanel null body which hides expand/collapse icon. 

Fixes #179167

### DIFF
--- a/packages/flutter/lib/src/material/expansion_panel.dart
+++ b/packages/flutter/lib/src/material/expansion_panel.dart
@@ -404,14 +404,14 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
       final Widget headerWidget = child.headerBuilder(context, _isChildExpanded(index));
 
       Widget expandIconPadded = Visibility(
-        visible: child.expandable,
+        visible: child.expandable || _isChildExpanded(index),
         maintainSize: true,
         maintainAnimation: true,
         maintainState: true,
         child: Padding(
           padding: const EdgeInsetsDirectional.only(end: 8.0),
           child: IgnorePointer(
-            ignoring: child.canTapOnHeader || !child.expandable,
+            ignoring: child.canTapOnHeader || (!child.expandable && !_isChildExpanded(index)),
             child: ExpandIcon(
               color: widget.expandIconColor,
               isExpanded: _isChildExpanded(index),
@@ -450,7 +450,7 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
           expandIconPadded,
         ],
       );
-      if (child.canTapOnHeader && child.expandable) {
+      if (child.canTapOnHeader && (child.expandable || _isChildExpanded(index))) {
         header = MergeSemantics(
           child: InkWell(
             splashColor: child.splashColor,
--- a/packages/flutter/lib/src/material/expansion_panel.dart
+++ b/packages/flutter/lib/src/material/expansion_panel.dart
@@ -80,6 +80,7 @@ class ExpansionPanel {
     required this.headerBuilder,
     required this.body,
     this.isExpanded = false,
+    this.expandable = true,
     this.canTapOnHeader = false,
     this.backgroundColor,
     this.splashColor,
@@ -98,6 +99,14 @@ class ExpansionPanel {
   ///
   /// Defaults to false.
   final bool isExpanded;
+
+  /// Whether the panel is expandable.
+  ///
+  /// When true, the expand/collapse icon is shown.
+  /// When false, the expand/collapse icon is hidden.
+  ///
+  /// Defaults to true.
+  final bool expandable;
 
   /// Defines the splash color of the panel if [canTapOnHeader] is true,
   /// or the splash color of the expand/collapse IconButton if [canTapOnHeader]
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -80,6 +80,7 @@ class ExpansionPanel {
     required this.headerBuilder,
     required this.body,
     this.isExpanded = false,
+    this.expandable = true,
```
victorsanni, line 83: The convention is a flag named `enabled`.

benji-farquhar: hmm i don't know, IgnorePointer has "ignoring", and Visibility has "visible", AbsorbPointer has "absorbing". I think enabled is a different concept where it is greyed out and potentially can become disabled/enabled. This feature is more to stop the behaviour, like the 3 widgets I mentioned.

benji-farquhar: I changed to `expandable` field name as i think @dkwingsmt was not liking the nullable body as it was unclear what its purpose was. What do you prefer @dkwingsmt? i am not fussed, i just need the feature, and we are not greying anything out, we are removing the expand completely.

benji-farquhar: Hi @victorsanni @dkwingsmt. I'm unsure about using "enabled" as the flag name because if someone wants a "greyed out" look in the future, they might just change the way "enabled: false" presents itself in the UI (adding the expand/collapse icon back in but grey).

victorsanni: > if someone wants a "greyed out" look in the future, they might just change the way "enabled: false" presents itself in the UI (adding the expand/collapse icon back in but grey).

I agree, this makes sense.

victorsanni (commented):
Left https://github.com/flutter/flutter/issues/179167#issuecomment-3844906747 in the linked issue.

packages/flutter/lib/src/material/expansion_panel.dart
victorsanni, line 83: The main purpose of this flag is to show/hide the icon. I was worried this might be conflating expansion behavior with icon visuals, but since the panel is expandable only by tapping the icon, I think this is fine.

packages/flutter/lib/src/material/expansion_panel.dart
victorsanni, line 79: What about `ExpansionPanelRadio`?

packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -394,17 +403,23 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
...
-          ignoring: child.canTapOnHeader,
-          child: ExpandIcon(
-            color: widget.expandIconColor,
-            isExpanded: _isChildExpanded(index),
-            padding: _kExpandIconPadding,
-            splashColor: child.splashColor,
-            highlightColor: child.highlightColor,
-            onPressed: (bool isExpanded) => _handlePressed(isExpanded, index),
+      Widget expandIconPadded = Visibility(
+        visible: child.expandable || _isChildExpanded(index),
```
victorsanni, line 407: This should be called out in the documentation.

packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -394,17 +403,23 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
...
+        maintainSize: true,
+        maintainAnimation: true,
+        maintainState: true,
```
victorsanni, line 410: Do we need this? Seems like the child can be recreated basically on demand. https://github.com/flutter/flutter/blob/6593592c450e5399a703353f42a6cc91b128b888/packages/flutter/lib/src/widgets/visibility.dart#L138

packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -394,17 +403,23 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
...
-            color: widget.expandIconColor,
-            isExpanded: _isChildExpanded(index),
-            padding: _kExpandIconPadding,
-            splashColor: child.splashColor,
-            highlightColor: child.highlightColor,
-            onPressed: (bool isExpanded) => _handlePressed(isExpanded, index),
+      Widget expandIconPadded = Visibility(
+        visible: child.expandable || _isChildExpanded(index),
+        maintainSize: true,
+        maintainAnimation: true,
```
victorsanni, line 409: Same here.

packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -394,17 +403,23 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
...
-          child: ExpandIcon(
-            color: widget.expandIconColor,
-            isExpanded: _isChildExpanded(index),
-            padding: _kExpandIconPadding,
-            splashColor: child.splashColor,
-            highlightColor: child.highlightColor,
-            onPressed: (bool isExpanded) => _handlePressed(isExpanded, index),
+      Widget expandIconPadded = Visibility(
+        visible: child.expandable || _isChildExpanded(index),
+        maintainSize: true,
```
victorsanni, line 408: Cannot be set without setting the other two, so same.

packages/flutter/lib/src/material/expansion_panel.dart
```diff
@@ -435,7 +450,7 @@ class _ExpansionPanelListState extends State<ExpansionPanelList> {
           expandIconPadded,
         ],
       );
-      if (child.canTapOnHeader) {
+      if (child.canTapOnHeader && (child.expandable || _isChildExpanded(index))) {
```
victorsanni, line 453: Reusing this, so should we make a getter for it?

PR #179189, opened 2025-11-27
Add DropdownMenu.selectOnly

### LINKED ISSUE #178009: Allow DropdownMenu to be non-editable and focusable (select control)
### Use case

A `DropdownMenu` configured as a non-editable dropdown (select control) should not allow text input in its text field, regardless of how it receives focus. When focused, the component as a whole should be highlighted, allowing the user to open the menu with keyboard actions such as the `Space` or `Enter` key. This behavior would align with standard dropdown components, such as the [Material Web select component](https://material-web.dev/components/select/#usage).

Right now, it is not possible to have a non-editable dropdown and also provide a `FocusNode`.

To clarify, by non-editable dropdown I mean a dropdown that works like a normal select control without a text field, only a selector.

In many real situations we still need to use a `FocusNode` object, for example:

1. When submitting the previous field and programmatically moving focus to the dropdown.
2. When submitting a form and focusing the first invalid field, which might be the dropdown.

---

**Example of current behavior:**

<details>
<summary>Details</summary>

To create a `DropdownMenu` that functions as a simple, non-editable selector, the current approach involves setting `requestFocusOnTap: false`, `enableSearch: false`, and `enableFilter: false`. While this correctly prevents the text field from activating on a mouse click, it does not prevent keyboard input when the widget gains focus through other means.

In this example, pressing the button gives focus to the `DropdownMenu`, which incorrectly allows text input.

Flutter version:

```text
stable : 3.35.7
master : 3.38.0-1.0.pre-407
```

```dart

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  @override
    return MaterialApp(
      home: Scaffold(
        body: Container(
          padding: const EdgeInsets.all(24.0),
          alignment: Alignment.center,
          child: Column(
            spacing: 24,
            children: [
              OutlinedButton(
                onPressed: () => focusNode.requestFocus(),
                child: const Text('Focus DropdownMenu'),
              ),
              DropdownMenu<Option>(
                focusNode: focusNode,
                requestFocusOnTap: false,
                enableSearch: false,
                enableFilter: false,
                initialSelection: Option.a,
                dropdownMenuEntries: Option.values
                    .map(
                      (option) =>
                          DropdownMenuEntry(value: option, label: option.name),
                    )
                    .toList(),
                onSelected: (Option? option) =>
                    setState(() => selectedOption = option),
              ),
            ],
          ),
        ),
      ),
    );
  }

  Option? selectedOption;
  FocusNode focusNode = FocusNode();

  @override
  dispose() {
    focusNode.dispose();
    super.dispose();
  }
}

enum Option { a, b }

```

Image

</details>

#### Proposal 1

If [enableFilter](https://main-api.flutter.dev/flutter/material/DropdownMenu/enableFilter.html) is set to `false` and [enableSearch](https://main-api.flutter.dev/flutter/material/DropdownMenu/enableSearch.html) is also set to `false`, the widget should behave like a select control without a text field, regardless of how focus is requested or whether a custom `FocusNode` is provided. It should not depend on `requestFocusOnTap` for that behavior.

The [requestFocusOnTap](https://main-api.flutter.dev/flutter/material/DropdownMenu/requestFocusOnTap.html) parameter should be removed.

---

Keep in mind that if we choose this proposal, we must also consider another scenario that is currently not implemented:

**When `DropdownMenu` is used as a select, it should focus/highlight the item that matches keyboard input even when there is no text field.**

This matters because the current [enableSearch](https://main-api.flutter.dev/flutter/material/DropdownMenu/enableSearch.html) setting disables highlight behavior when set to `false`, which could conflict with this new behavior if implemented in the future.

<details>
<summary>Example</summary>

**Case 1**

1. Open dropdown

   Image

2. Press **"a"** (Go to the first item that starts with "a")

   Image

3. Press **"a"** again (Go to the next item that starts with "a")

   Image

4. Press **"a"** again (Return to the first item that starts with "a" since there are no more matches)

   Image

**Case 2**

Image

2. Quickly type a combination of keys like `"a4"` — this focuses the item that matches that prefix.
   In this case it jumps to `a4` and skips `a2` and `a3`.

   Image

</details>

---

#### Proposal 2

Add another parameter to make the component work strictly as a select.

Possible names:

* `showSearch`: `bool`
* `searchable` : `bool`
* `selectOnly`: `bool`
* `enableTextFieldSearch`: `bool`
* `mode`: `enum DropdownMenuMode {select, textField}`

The [requestFocusOnTap](https://main-api.flutter.dev/flutter/material/DropdownMenu/requestFocusOnTap.html) parameter should be removed.

In this case, make sure the current [enableSearch](https://main-api.flutter.dev/flutter/material/DropdownMenu/enableSearch.html) does not cause confusion with this new parameter.

bleroux:
## Description

This PR adds `DropdownMenu.selectOnly`. This property allows users to get the DropdownMenu behave as a 'select' component.
It is meant as a future replacement for `DropdownMenu.requestFocusOnTap`.

## Motivation

On mobile, a dropdown menu widget is usually used as a ‘select’ widget because opening the keyboard for searching/filtering is not convenient.

In Flutter, this is currently implemented through `DropdownMenu.requestFocusOnTap` which defaults to false on mobile and true on desktop.
The `DropdownMenu.requestFocusOnTap` property is currently used to set `FocusNode.canRequestFocus`. This leads do difficulties mainly related to focus traversal:
- Keyboard traversal requires workarounds (for instance relying on the trailing icon to be focusable).
- Keyboard shortcuts require also a workaround (currently relying on a Focus widget in a Stack).
- The `DropdownMenu` decoration does not reflect the focus state.

This PR proposes a new property named `DropdownMenu.selectOnly` which does not require `DropdownMenu.requestFocusOnTap` to be false to make the `DropdownMenu` behave like a select widget.

With this property the `DropdownMenu`:
- Supports keyboard navigation on mobile and desktop.
- Has a correct decoration when focused.
- Does not rely on the trailing icon to be focusable (see https://github.com/flutter/flutter/issues/174096).

In the future this property could be used as a replacement for `DropdownMenu.requestFocusOnTap`. For the moment, for compatibility, it does not replace `DropdownMenu.requestFocusOnTap`.

## Related Issue

Fixes [Allow DropdownMenu to be non-editable and focusable (select control) ](https://github.com/flutter/flutter/issues/178009)
Also related to [Make DropdownMenu's trailing icon not focusable by default](https://github.com/flutter/flutter/issues/174096)  and [[Material3] DropdownMenu Keyboard Accessibility](https://github.com/flutter/flutter/issues/123797).

## Tests

- Adds 9 tests.

### DIFF
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -672,7 +672,7 @@ class DropdownMenu<T extends Object> extends StatefulWidget {
 }
 
 class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
-  static const Map<ShortcutActivator, Intent> _commonShortcuts = <ShortcutActivator, Intent>{
+  static const Map<ShortcutActivator, Intent> _editableShortcuts = <ShortcutActivator, Intent>{
     SingleActivator(LogicalKeyboardKey.arrowLeft): ExtendSelectionByCharacterIntent(
       forward: false,
       collapseSelection: true,
@@ -683,6 +683,14 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
     ),
     SingleActivator(LogicalKeyboardKey.arrowUp): _ArrowUpIntent(),
     SingleActivator(LogicalKeyboardKey.arrowDown): _ArrowDownIntent(),
+  };
+
+  static const Map<ShortcutActivator, Intent> _selectOnlyShortcuts = <ShortcutActivator, Intent>{
+    SingleActivator(LogicalKeyboardKey.arrowUp): _ArrowUpIntent(),
+    SingleActivator(LogicalKeyboardKey.arrowDown): _ArrowDownIntent(),
+    // When selectOnly is true, a shortcut for the enter key is needed because
+    // the text field won't provide one.
+    SingleActivator(LogicalKeyboardKey.enter): _EnterIntent(),
   };
 
   final GlobalKey _anchorKey = GlobalKey();
@@ -1320,12 +1328,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
               );
 
         return Shortcuts(
-          shortcuts: <ShortcutActivator, Intent>{
-            ..._commonShortcuts,
-            // When selectOnly is true, a shortcut for the enter key is needed because
-            // the text field won't provide one.
-            if (selectOnly) const SingleActivator(LogicalKeyboardKey.enter): const _EnterIntent(),
-          },
+          shortcuts: selectOnly ? _selectOnlyShortcuts : _editableShortcuts,
           child: body,
         );
       },
--- a/packages/flutter/lib/src/material/dropdown_menu.dart
+++ b/packages/flutter/lib/src/material/dropdown_menu.dart
@@ -1021,7 +1021,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
     return result;
   }
 
-  void handleUpKeyInvoke(_ArrowUpIntent _) {
+  void handleUpKey(_ArrowUpIntent _) {
     setState(() {
       if (!widget.enabled || !_menuHasEnabledItem || !_controller.isOpen) {
         return;
@@ -1041,7 +1041,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
     });
   }
 
-  void handleDownKeyInvoke(_ArrowDownIntent _) {
+  void handleDownKey(_ArrowDownIntent _) {
     setState(() {
       if (!widget.enabled || !_menuHasEnabledItem || !_controller.isOpen) {
         return;
@@ -1367,8 +1367,8 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
 
     return Actions(
       actions: <Type, Action<Intent>>{
-        _ArrowUpIntent: CallbackAction<_ArrowUpIntent>(onInvoke: handleUpKeyInvoke),
-        _ArrowDownIntent: CallbackAction<_ArrowDownIntent>(onInvoke: handleDownKeyInvoke),
+        _ArrowUpIntent: CallbackAction<_ArrowUpIntent>(onInvoke: handleUpKey),
+        _ArrowDownIntent: CallbackAction<_ArrowDownIntent>(onInvoke: handleDownKey),
         _EnterIntent: CallbackAction<_EnterIntent>(onInvoke: handleEnterKey),
         DismissIntent: DismissMenuAction(controller: _controller),
       },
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1330,7 +1369,7 @@ class _DropdownMenuState<T extends Object> extends State<DropdownMenu<T>> {
       actions: <Type, Action<Intent>>{
         _ArrowUpIntent: CallbackAction<_ArrowUpIntent>(onInvoke: handleUpKeyInvoke),
         _ArrowDownIntent: CallbackAction<_ArrowDownIntent>(onInvoke: handleDownKeyInvoke),
-        _EnterIntent: CallbackAction<_EnterIntent>(onInvoke: (_) => _handleSubmitted()),
+        _EnterIntent: CallbackAction<_EnterIntent>(onInvoke: handleEnterKey),
```
QuncCccccc, line 1371: nit: maybe

```suggestion
        _EnterIntent: CallbackAction<_EnterIntent>(onInvoke: handleEnterKeyInvoke),
```

for consistency with the handle methods above.

QuncCccccc (commented):
Sorry, I'm still a bit confused. What's the difference between this new API and `requestFocusOnTap`? Feels like super similar:)

packages/flutter/lib/src/material/dropdown_menu.dart
bleroux: I choosed to renamed the existing callbacks for consistency.

bleroux:
> Are there a way to just keep the current API (requestFocusOnTap) and add the behaviors by updating the implementation?

While both properties target a similar use case: selecting an entry without editing the text field (so without opening the virtual keyboard on mobile), there is a main difference which is that `requestFocusOnTap` achieves this by not allowing the component to be focused, `selectOnly` does not prevent the component to be focused (it is mainly mapped to `TextField.readOnly`).

Maybe We can summarize this with:

```dart
requestFocusOnTap = selectOnly + canRequestFocus
```

(just for illustration because at the boolean level it is not correct :sweat_smile: )

Changing the current `requestFocusOnTap` behaviors would be problematic as the name will directly contradict the implementation (even when false the component will get the focus). Also it will break a lot of existing tests which assume that the `DropdownMenu` is not focusable at all when `requestFocusOnTap` is false.

I think both properties can coexist for the moment and at one time we might consider that `requestFocusOnTap` is only `canRequestFocus` and that it could be deprecated as one can rely on `FocusNode.canRequestFocus`. But to do so we first need to separate the behaviors related to the 'select' use case, this is what this PR does by adding this new DropdownMenu.selectOnly property.

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -506,6 +507,24 @@ class DropdownMenu<T extends Object> extends StatefulWidget {
...
+  /// a 'select' widget (i.e., the user can only select from the list, not edit
+  /// the text field to search or filter).
+  ///
+  /// When true, the inner text field is read-only.
+  ///
+  /// If the text field is also focusable (see [requestFocusOnTap]), the following
+  /// behaviors are also activated:
+  ///
+  ///  * Pressing Enter when the menu is closed opens it.
+  ///
```
QuncCccccc:
```suggestion
```

QuncCccccc (approved):
LGTM:) Thanks!

Issue #178463, opened 2025-11-13
Android's predictive back feature doesn't use the device's corner radius


MaherSafadii:
Since predictive back is now enabled by default, it would be better to have it polished and match the native behaviour, currently it uses a fixed border radius when the effect is active, which looks bad, Android provided APIs that get the devices corner radius which can be used to achieve a more seamless effect and match the native side.

Current implementation:

<img width="256" alt="Image" src="https://github.com/user-attachments/assets/0d665af8-6e62-49b1-9494-516a3ad596f2" />

in my custom implementation, all I did was use a ClipRRect on a whole page and set the corner radius to the device's using this package [corner_radius_plugin](https://pub.dev/packages/corner_radius_plugin), it uses official Android APIs to get the devices corner radius.

the result is a more native behavior:

<img width="256" alt="Image" src="https://github.com/user-attachments/assets/44b01499-6f4d-4f50-9dc3-a3ec75dac44a" />

tirth-patel-nc:
@MaherSafadii thanks for the report. labelling it as a new proposal. 

#166608 #97349 might be related.

maRci002:
Flutter uses a fixed radius for now because device-specific corner radii aren't exposed yet. Once **#97349** lands, this code can read the actual device radius instead of the hard-coded value.

https://github.com/flutter/flutter/blob/a5067be8868cc37820d2dd4df29043737302e75b/packages/flutter/lib/src/material/predictive_back_page_transitions_builder.dart#L361-L365

PR #179219, opened 2025-11-28
[Android] Add display corner radii support.

### LINKED ISSUE #97349: RoundedCorner and getRoundedCorner
It looks like Android 12 has introduced some nice features which allow developers to get the corner radius of a device.  This is a potential game changer for designs at the edge of the screen. 

Any chance this can be implemented in Flutter?

https://developer.android.com/reference/android/view/RoundedCorner
https://developer.android.com/reference/android/view/WindowInsets?hl=en#getRoundedCorner(int)

ksokolovskyi:
Closes https://github.com/flutter/flutter/issues/97349
Unblocks https://github.com/flutter/flutter/issues/178463

### Description

This PR focuses only on the Android side.

As was discussed in https://github.com/flutter/flutter/issues/97349, the iOS API for getting display corner radii is private. The only safe way to get the radii values is to generate a lookup table ahead and use its values at runtime. I would be happy to work on the iOS implementation in a separate PR after this lands.

- Adds `displayCornerRadii` in the physical pixels to the engine's window metrics
- Adds `displayCornerRadii` support on the Android API 31+
- Adds `displayCornerRadii` in the logical pixels to the `MediaQuery`
- Adds and updates tests

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -796,6 +826,7 @@ class MediaQueryData {
       letterSpacingOverride: letterSpacingOverride,
       wordSpacingOverride: wordSpacingOverride,
       paragraphSpacingOverride: paragraphSpacingOverride,
+      displayCornerRadii: displayCornerRadii,
```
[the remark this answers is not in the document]
ksokolovskyi: The `displayCornerRadii` parameter was skipped from the `copyWith` intentionally.
As `displayCornerRadii` is nullable, the code you suggested won't allow setting it to `null`:

```dart
displayCornerRadii: displayCornerRadii ?? this.displayCornerRadii
```

While `applyDisplayCornerRadii` method introduced in this PR allows both overriding `displayCornerRadii` and setting it to `null`.

mdebbar (commented):
The web stubs look good to me!

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
```diff
@@ -786,6 +787,20 @@ navigationBarVisible && guessBottomKeyboardInset(insets) == 0
       delegate.growViewportMetricsToCaptionBar(getContext(), viewportMetrics);
     }
 
+    if (Build.VERSION.SDK_INT >= API_LEVELS.API_31) {
+      RoundedCorner topLeft = insets.getRoundedCorner(RoundedCorner.POSITION_TOP_LEFT);
+      RoundedCorner topRight = insets.getRoundedCorner(RoundedCorner.POSITION_TOP_RIGHT);
+      RoundedCorner bottomRight = insets.getRoundedCorner(RoundedCorner.POSITION_BOTTOM_RIGHT);
+      RoundedCorner bottomLeft = insets.getRoundedCorner(RoundedCorner.POSITION_BOTTOM_LEFT);
+
+      viewportMetrics.displayCornerRadiusTopLeft = topLeft != null ? topLeft.getRadius() : 0;
```
mboetger, line 862: Should this be 0 or unsetValue?

mboetger (commented):
Android-side plumbing looks good.  One question below. Also we'll need someone from the framework to look at this, I believe.

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterView.java
ksokolovskyi: Hi @mboetger, thanks a lot for your review.

> Should this be 0 or unsetValue?

The [docs](https://developer.android.com/reference/android/view/WindowInsets#getRoundedCorner(int)) for `getRoundedCorner` say the following: `Returns the rounded corner of the given position. Returns null if there is none, or the rounded corner area is not inside the application's bounds`. 

From my understanding, `Returns null if there is none` means that the device corner is not rounded, thus the radius equals `0`.

Regarding `Returns null if the rounded corner area is not inside the application's bounds`, I tried to find a case when `getRoundedCorner` returns null on the real devices and emulators, but I haven't found one. When the app is in split-screen mode, the radius values are still non-null. However, this doesn't mean that this case doesn't exist.

Additionally, I map the `DisplayCornerRadii` from the engine to `BorderRadius` on the framework side, and `BorderRadius` doesn't have an option to mark the corner as unset. Though the mapping of `DisplayCornerRadii` to `BorderRadius` is only my choice, and I would like to hear the feedback from the framework team regarding this.

@justinmc, may I ask you to take a look at this PR when you have time? Thanks in advance!

mboetger: Seems reasonable to me.

engine/src/flutter/shell/platform/android/test/io/flutter/embedding/android/FlutterViewTest.java
```diff
@@ -1204,6 +1205,52 @@ public void setPaddingTopToZeroForFullscreenModeLegacy() {
     validateViewportMetricPadding(viewportMetricsCaptor, 100, 0, 100, 0);
   }
 
+  @Test
+  @TargetApi(31)
+  @Config(sdk = API_LEVELS.API_31)
```
mboetger:
```suggestion
  @Config(minSdk = API_LEVELS.API_31)
```

Just to ensure it works on 31 and above :)

mboetger (approved):
LGTM from the Android side - please wait to submit until you get LGTMs from the web and framework side though.

ksokolovskyi:
@mboetger, could you please take a look at why Google Testing failed?

mboetger:
> @mboetger, could you please take a look at why Google Testing failed?

Working on it.

packages/flutter/lib/src/widgets/media_query.dart
```diff
@@ -727,6 +746,17 @@ class MediaQueryData {
   /// for text paragraph spacing.
   final double? paragraphSpacingOverride;
 
+  /// The radii of the display corners in logical pixels.
+  ///
+  /// This is currently populated only on Android API 31+. On earlier Android
```
Piinks, line 751: I would just add mention of what platforms this does not work on yet, like iOS

mboetger:
Since my PR was reverted and added back in - I resolved the conflicts for you.  Hopefully the Google Testing should just start passing again.

engine/src/flutter/shell/platform/android/platform_view_android_jni_impl.cc
mboetger: When I resolved the conflicts - this didn't come up.  You'll probably need to add another four 'I' to this.

packages/flutter/lib/src/widgets/media_query.dart
ksokolovskyi: Hi @Piinks, thanks a lot for taking a look at this PR!

Would something like this work? 

```
/// The radii of the display corners in logical pixels.
///
/// This is currently populated only on Android API 31+. On earlier Android 
/// versions, iOS, and other platforms, this value is `null`.
///
/// See also:
///
///  * [FlutterView.displayCornerRadii], which returns the display corner
///    radii in physical pixels.
```

engine/src/flutter/shell/platform/android/platform_view_android_jni_impl.cc
ksokolovskyi: Thanks a lot for resolving the conflicts! I've updated the signature in the latest commit. Now let's wait for Google Testing to finish.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -2749,6 +2759,60 @@ enum DisplayFeatureState {
   postureHalfOpened,
 }
 
+/// A set of radii for the four corners of the display.
```
justinmc, line 2762: Nit: Maybe say "device's physical display" to make it more clear what this is?

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -2749,6 +2759,60 @@ enum DisplayFeatureState {
...
+///
+/// This is currently populated only on Android API 31+.
+class DisplayCornerRadii {
+  /// Creates a [DisplayCornerRadii].
+  const DisplayCornerRadii({
+    required this.topLeft,
+    required this.topRight,
+    required this.bottomRight,
+    required this.bottomLeft,
+  });
```
justinmc: Nit: Would it be useful to add asserts preventing negative numbers here?

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -2749,6 +2759,60 @@ enum DisplayFeatureState {
...
+        topRight == other.topRight &&
+        bottomRight == other.bottomRight &&
+        bottomLeft == other.bottomLeft;
+  }
+
+  @override
+  int get hashCode => Object.hash(topLeft, topRight, bottomRight, bottomLeft);
+
+  @override
+  String toString() {
```
justinmc, line 2813: Consider mixing in Diagnosticable and using debugFillProperties here. https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#override-tostring

engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -401,6 +401,47 @@ enum DisplayFeatureType { unknown, fold, hinge, cutout }
...
+        topRight == other.topRight &&
+        bottomRight == other.bottomRight &&
+        bottomLeft == other.bottomLeft;
+  }
+
+  @override
+  int get hashCode => Object.hash(topLeft, topRight, bottomRight, bottomLeft);
+
+  @override
+  String toString() {
```
justinmc, line 442: Same Diagnosticable comment here.

engine/src/flutter/lib/ui/fixtures/ui_test.dart
```diff
@@ -889,6 +889,10 @@ void hooksTests() async {
       0.0, // maxWidth
       0.0, // minHeight
       0.0, // maxHeight
+      -1.0, // display corner radius top left
```
justinmc, line 900: I guess we're assuming that the radius is always circular. Is that an assumption that the Android and iOS APIs make too?

packages/flutter/lib/src/widgets/media_query.dart
justinmc: @Piinks has likely started the holidays already, but your change looks fine to me. Consider doing the same in the other places where this is explained in this PR.

justinmc (approved):
LGTM with nits 👍 Thank you for doing this @ksokolovskyi, it's going to be a big improvement for predictive back!

Is there any reason to think that we couldn't reuse the same mechanism to pass this value from iOS as well some day? Seems like it should work to me but want to make sure.

A major edge case: What about a foldable phone that the user unfolds? Does that change the radius values?

engine/src/flutter/lib/ui/platform_dispatcher.dart
ksokolovskyi: This sounds great, but `Diagnosticable` is defined in `flutter/lib/src/foundation` and thus can't be used in `engine` layer.

engine/src/flutter/lib/ui/fixtures/ui_test.dart
ksokolovskyi: Both platforms represent device corner radii as a single scalar value rather than an elliptical X/Y pair.

On Android, the `RoundedCorner` API specifically defines the corner as a 'quarter circle approximation' and only provides a single [`getRadius()`](https://developer.android.com/reference/android/view/RoundedCorner#getRadius()) value.

On iOS, Apple uses 'continuous' curves (squircles) for a smoother look. Still, they are defined by a single radius value available via the private `_displayCornerRadius` property.

justinmc: Got it, thanks for reassuring me :+1:

engine/src/flutter/lib/ui/platform_dispatcher.dart
justinmc: Darn you're right.

engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart
justinmc: This is not possible as discussed elsewhere.

justinmc (approved):
LGTM :+1: . Thanks for the followup. Let's wait for a web review and then this is good to go.

mdebbar (approved):

PR #179224, opened 2025-11-28
165977 Add focus support for CupertinoAlertDialogAction

O-Hannonen:
## This PR 
- fixes #165977 by allowing `CupertinoDialogAction` to gain focus
- updates `CupertinoFocusHalo` to use superellipse shape

### How the new implementation looks and behaves

https://github.com/user-attachments/assets/c33beb31-e30f-4878-b44b-f04622019e73

### How the native counterpart looks and behaves

https://github.com/user-attachments/assets/8ea57f08-3906-4d95-bbb7-94176267207e

### Sample code

<details>

<summary>Sample code</summary>

```dart
import 'package:flutter/cupertino.dart';
import 'package:flutter/material.dart';

void main() => runApp(const AlertDialogApp());

class AlertDialogApp extends StatelessWidget {
  const AlertDialogApp({super.key});

  @override
  Widget build(BuildContext context) {
    return CupertinoApp(
      theme: CupertinoThemeData(brightness: Brightness.light),
      home: AlertDialogExample(),
    );
  }
}

class AlertDialogExample extends StatelessWidget {
  AlertDialogExample({super.key});

  FocusNode twoFocusNode = FocusNode();

  Future<void> _showModal(BuildContext context) {
    return showCupertinoModalPopup(
      context: context,
      builder: (BuildContext context) => CupertinoActionSheet(
        title: Text('title'),
        actions: <Widget>[
          CupertinoActionSheetAction(
            child: Text('text'),
            onPressed: () {
              Navigator.pop(context);
            },
          ),
          CupertinoActionSheetAction(
            child: Text('image'),
            onPressed: () {
              Navigator.pop(context);
            },
          ),
        ],
        cancelButton: CupertinoActionSheetAction(
          child: Text('cancel'),
          isDefaultAction: true,
          onPressed: () {
            Navigator.pop(context);
          },
        ),
      ),
    );
  }

  Future<void> _showAlertDialog(BuildContext context) {
    return showCupertinoDialog<void>(
      context: context,
      builder: (BuildContext context) => CupertinoAlertDialog(
        title: GestureDetector(
          child: const Text('Alert'),
          onTap: () {
            twoFocusNode.requestFocus();
            print(twoFocusNode.hasPrimaryFocus);
          },
        ),
        content: const Text('Proceed with destructive action?'),
        actions: <CupertinoDialogAction>[
          CupertinoDialogAction(
            isDefaultAction: true,
            onPressed: () {
              print('No');
              Navigator.pop(context);
            },
[...]

PR #179225, opened 2025-11-28
167114 add keyboard focus support for cupertino context menu action

### LINKED ISSUE #167114: CupertinoContextMenuAction: Add focus support
### Use case

`CupertinoContextMenuAction` is not able to receive focus so `CupertinoContextMenu` can't be navigated with keyboard when it has `CupertinoContextMenuAction`s as children.

Found while implementing fixes for #166398 and #165977, so highly related to those.

### Proposal

The native iOS context menu components work with keyboard like so:
https://github.com/user-attachments/assets/4d6891bb-c947-4730-9e3f-3e952dc3276a

Here is a reproducible example of how this does not work in Flutter.

<details>

<summary>Example</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
      ),
    );
  }
}

  final String title;

  @override
}

  @override
    return Scaffold(
      appBar: AppBar(

        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          spacing: 16,
          children: <Widget>[
            CupertinoContextMenu(
              actions: <Widget>[
                CupertinoContextMenuAction(
                  child: Text('Action 1'),
                  onPressed: () {
                    Navigator.pop(context);
                  },
                ),
                CupertinoContextMenuAction(
                  child: Text('Action 2'),
                  onPressed: () {
                    Navigator.pop(context);
                  },
                ),
                CupertinoContextMenuAction(
                  isDestructiveAction: true,
                  child: Text('Delete'),
                  onPressed: () {
                    Navigator.pop(context);
                  },
                ),
              ],
              child: Material(
                child: Container(
                  width: 100,
                  height: 100,
                  color: Colors.blue,
                  child: Center(child: Text('Hold me')),
                ),
              ),
            ),
          ],
        ),
      ),
    );
  }
}

```

</details>

O-Hannonen:
## This PR
- fixes #167114 by adding focus support for `CupertinoContextMenuAction`
- makes it possible to focus `CupertinoContextMenu`
- makes it possible to open `CupertinoContextMenu` with tab + m keyboard shortcut (iOS system keyboard traversal default behavior)
- updates `CupertinoFocusHalo` to use superellipse borders

### Sample how the new code looks and behaves

https://github.com/user-attachments/assets/27ca6a72-0585-4c67-81dd-1e832fc1adb4

### Sample how context menu traversal looks and behaves on iOS

https://github.com/user-attachments/assets/5d67c8de-09e4-447a-b672-feabdee23544

### Code sample

<details>

```dart

void main() => runApp(const ContextMenuApp());

class ContextMenuApp extends StatelessWidget {
  const ContextMenuApp({super.key});

  @override
    return CupertinoApp(
      theme: CupertinoThemeData(brightness: Brightness.light),
      home: ContextMenuExample(),
    );
  }
}

class ContextMenuExample extends StatefulWidget {
  const ContextMenuExample({super.key});

  @override
  State<ContextMenuExample> createState() => _ContextMenuExampleState();
}

class _ContextMenuExampleState extends State<ContextMenuExample> {
  @override
    return CupertinoPageScaffold(
      navigationBar: CupertinoNavigationBar(

        middle: Text('Context Menu Example'),
      ),
      child: Center(
        child: Column(
          spacing: 16,
          children: <Widget>[
            CupertinoContextMenu(
              actions: <Widget>[
                CupertinoContextMenuAction(
                  child: Text('Action 1'),
                  onPressed: () {
                    print('Action 1');
                    Navigator.pop(context);
                  },
                ),
                CupertinoContextMenuAction(
                  child: Text('Action 2'),
                  onPressed: () {
                    print('Action 2');
                    Navigator.pop(context);
                  },
                ),
                CupertinoContextMenuAction(
                  isDestructiveAction: true,
                  child: Text('Delete'),
                  onPressed: () {
                    print('Delete');
                    Navigator.pop(context);
                  },
                ),
              ],

              child: Material(
                child: Container(
                  width: 100,
                  height: 100,
                  color: Colors.orange,
                  child: Center(child: Text('Hold me')),
                ),
              ),
            ),
            TextButton(
              onPressed: () {
                print('TextButton1 pressed');
              },
              child: Text('Another button 1'),
            ),
            TextButton(
              onPressed: () {
                print('TextButton2 pressed');
              },
              child: Text('Another button 2'),
            ),
          ],
        ),
      ),
    );
  }
}
```

</details>
## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -11,10 +11,13 @@ import 'dart:ui' as ui;
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/scheduler.dart';
+import 'package:flutter/semantics.dart';
 import 'package:flutter/services.dart' show HapticFeedback;
 import 'package:flutter/widgets.dart';
 
+import '../services/hardware_keyboard.dart';
```
victorsanni: I see `package:flutter/services.dart` is already imported, maybe we can show what is necessary? or remove the show entirely so the entire library is available?

packages/flutter/test/cupertino/action_sheet_test.dart
```diff
@@ -1978,7 +1978,6 @@ void main() {
     },
     variant: TargetPlatformVariant.only(TargetPlatform.iOS),
   );
-
```
victorsanni, line 1979: undo

packages/flutter/test/cupertino/context_menu_action_test.dart
```diff
@@ -5,7 +5,9 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
```
victorsanni: Can we not import Material?

packages/flutter/lib/src/cupertino/cupertino_focus_halo.dart
```diff
@@ -95,11 +95,13 @@ class _CupertinoFocusHaloState extends State<CupertinoFocusHalo> {
...
-          border: _childHasFocus
-              ? Border.fromBorderSide(BorderSide(color: _effectiveFocusOutlineColor, width: 3.5))
-              : null,
+        decoration: ShapeDecoration(
+          shape: RoundedSuperellipseBorder(
+            side: _childHasFocus
+                ? BorderSide(color: _effectiveFocusOutlineColor, width: 3.5)
+                : BorderSide.none,
+            borderRadius: widget._borderRadius,
+          ),
```
O-Hannonen: These small changes were copied to here too from [another PR](https://github.com/flutter/flutter/pull/179224), but now that there are more related changes coming from the other PR it would be maybe beneficial to wait with this PR until the other one is merged.

O-Hannonen: These were reverted here and are now pending for [this](https://github.com/flutter/flutter/pull/180724/files)

packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -602,7 +626,32 @@ class _CupertinoContextMenuState extends State<CupertinoContextMenu> with Ticker
...
+      if (_lastKeyEvent case KeyDownEvent(logicalKey: LogicalKeyboardKey.tab)) {
+        FocusScope.of(context).nextFocus();
+      }
+      _lastKeyEvent = event;
+      return KeyEventResult.handled;
+    }
+
+    _lastKeyEvent = event;
+
+    return KeyEventResult.ignored;
```
dkwingsmt: Why do you have to handle key events manually instead of using the `Shortcuts` framework? What's the recorded `_lastKeyEvent` used for?

packages/flutter/lib/src/cupertino/context_menu_action.dart
```diff
@@ -43,6 +48,17 @@ class CupertinoContextMenuAction extends StatefulWidget {
...
 
+  /// {@macro flutter.widgets.Focus.focusNode}
+  final FocusNode? focusNode;
+
+  /// The color of the background that highlights active focus.
+  ///
+  /// A transparency of [kCupertinoButtonTintedOpacityLight] (light mode) or
+  /// [kCupertinoButtonTintedOpacityDark] (dark mode) is automatically applied to this color.
+  ///
+  /// When [focusColor] is null, defaults to [CupertinoColors.activeBlue].
```
dkwingsmt: Will it be better to embed the transparency into the default color, so that the app can have full control over the transparency?

packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -359,13 +369,20 @@ class CupertinoContextMenu extends StatefulWidget {
   /// Defaults to false.
   final bool enableHapticFeedback;
 
+  /// {@macro flutter.widgets.Focus.focusNode}
+  final FocusNode? focusNode;
+
+  /// The radius of the border rendered around this widget when focused.
```
dkwingsmt: Maybe we can start promoting this term
```suggestion
  /// The radius of the focus halo rendered around this widget when focused.
```

packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -564,7 +585,10 @@ class _CupertinoContextMenuState extends State<CupertinoContextMenu> with Ticker
     _onTapCompleted();
   }
 
-  void _onTapDown(TapDownDetails details) {
+  void _onTapDown(TapDownDetails details) => _handleOpenIntent();
+
+  void _handleOpenIntent([Intent? intent]) {
```
dkwingsmt: The current way of making `_onTapDown` call `handleOpenIntent` with a null parameter omitted is a bit weird to read. I suggest renaming the implementation method as `_performOpenMenu({required bool fromTap})` and making both tapping and actions call this method.

dkwingsmt (commented):
Generally ok except for how we handle the tab key

packages/flutter/lib/src/cupertino/context_menu_action.dart
O-Hannonen: Makes sense!

packages/flutter/lib/src/cupertino/context_menu.dart
O-Hannonen: The `Shortcuts` framework handles tab quite poorly as a modifier. Immediately when tab is pressed down, the focus moves to next item. So its not really possible to use only it if we want to have a focusable widget that uses tab as modifier key for the shortcut. 

The reason for manually listening to key events is to prevent the tab down from bubbling up and only move focus when the tab is released.

But now that I'm saying this aloud, should that be fixed elsewhere so we could use the shortcuts here? So eg. should tab focus traversal be modified to only move to next item when tab is released (vs. the on tab down that its now)? Or should a similar fix to this be added to the `Shortcuts` (so is it a common use case that we want to support that tab can be modifier)?

Here is a small reproduction highlighting the issue:

<details>
<summary>Sample code</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Tab as modifier key bug',
      theme: ThemeData(
        useMaterial3: true,
      ),
      home: const TabModifierKeyBugDemo(),
    );
  }
}

class TabModifierKeyBugDemo extends StatefulWidget {
  const TabModifierKeyBugDemo({super.key});

  @override
  State<TabModifierKeyBugDemo> createState() => _TabModifierKeyBugDemoState();
}

class _TabModifierKeyBugDemoState extends State<TabModifierKeyBugDemo> {
  int tabMCount = 0;
  void _handleTabM() {
    setState(() {
      tabMCount++;
    });
  }

  @override
    return Scaffold(
      appBar: AppBar(
        title: const Text('Tab + Modifier Key Bug Reproduction'),
      ),
      body: Padding(
        child: Column(
          crossAxisAlignment: CrossAxisAlignment.start,
          spacing: 16,
          children: [
            const Card(
              color: Color.fromARGB(255, 255, 240, 245),
              child: Padding(
                padding: EdgeInsets.all(12.0),
                child: Column(
                  crossAxisAlignment: CrossAxisAlignment.start,
                  children: [
                    Text(
                      'Instructions:',
                    ),
                    SizedBox(height: 8),
                    Text(
                      '1. Press Tab to navigate between buttons\n'
                      '2. When focused on Button 2, press Tab down. While holding Tab, press M.\n'
                      'BUG: Focus should NOT move to Button 3, only Tab+M action should trigger. Focus should move to next button only if Tab is released and no other key is pressed.\n',
                    ),
                  ],
                ),
              ),
            ),
            Text(
              'Tab+M pressed count: $tabMCount',
            ),
            ElevatedButton( 
              onPressed: () {},
              child: const Text('Button 1'),
            ),
            Shortcuts(
              shortcuts: {
                LogicalKeySet(LogicalKeyboardKey.tab, LogicalKeyboardKey.keyM):
                    const _TabMIntent(),
              },
              child: Actions(
                actions: {
                  _TabMIntent: CallbackAction<_TabMIntent>(
                    onInvoke: (_) {
                      _handleTabM();
                      return null;
                    },
                  ),
                },
                child: Container(
                  decoration: BoxDecoration(
                    border: Border.all(color: Colors.orange, width: 2),
                    borderRadius: BorderRadius.circular(8),
                  ),
                  padding: const EdgeInsets.all(12),
                  child: ElevatedButton(
                    onPressed: () {},
                    child: const Text('Button 2 (Has Tab+M shortcut)'),
                  ),
                ),
              ),
            ),
            ElevatedButton(
              onPressed: () {},
              child: const Text('Button 3'),  
            ),
          ],
        ),
      ),
    );
  }
}

class _TabMIntent extends Intent {
  const _TabMIntent();
}

```

</details>

dkwingsmt: Flutter has always defined tabbing as a keydown trigger, which is defined by all occurrences you can find by searching `SingleActivator(LogicalKeyboardKey.tab): NextFocusIntent()`, where `SingleActivator` only supports keyDown or keyRepeat trigger, not keyUp. Flutter doesn't support keyUp trigger shortcuts for now. 

Can you explain why you want a keyUp trigger? My test on XCode also shows that iOS uses keyDown trigger (to be precise, keyDown and keyRepeat).

packages/flutter/lib/src/cupertino/context_menu.dart
```diff
@@ -359,13 +369,20 @@ class CupertinoContextMenu extends StatefulWidget {
   /// Defaults to false.
   final bool enableHapticFeedback;
 
+  /// {@macro flutter.widgets.Focus.focusNode}
+  final FocusNode? focusNode;
+
+  /// The radius of the focus halo rendered around this widget when focused.
```
dkwingsmt: Explain default

O-Hannonen:
Sorry for the delay, did some further testing myself now too:
> There is no API to query whether FKA is enabled. Also, the shortcuts enabled by FKA are customizable, including Tab-M, and there is no API to query them either.

This is quite unfortunate :/

> Tab-M only works for context menu on the main screen, but not app widgets that use .contextMenu {  }.
Interesting finding, I had only tested on the main screen. Super weird and inconsistent decision from Apple. 

Did you figure out any way to open the context menu on keyboard inside apps? Eg. while testing in the Photos app, I was not able to open the context menu in any way with keyboard... 

> Flutter comes with focus halo support by default. I guess this is actually related to a11y. Following is the material MenuAnchor displaying the focus halo on an iPad simulator with FKA on.

MenuAnchor is *quite* similar but has few key differences. AFAIK the menu anchor is not intended to have any other onTap action than opening the anchor? Context menu usually has both onTap action in addition to opening the context menu (so both onTap and onLongPress) -> can't just use the regular tap event (space on keyboard) to open the menu. 

If its really true that Apple FKA does not support opening context menu on keyboard (other than on home screen), we could simply remove the keyboard shortcut part from this PR? So just make `ContextMenuAction` focusable by keyboard? Any thoughts?

O-Hannonen:
I've removed the keyboard shortcut from this PR for now. I still need to dig a bit deeper into the other points discussed, which I’m hoping to find time for shortly.

Based on what I've seen so far, I have two potential theories on how we should move forward (though further investigation might prove both wrong! 😄). Either way, I don't think creating `CupertinoFocusHalo` was a mistake. Because Flutter supports basic keyboard traversal even when Full Keyboard Access (FKA) is OFF, we are going to need a focus ring widget for that situation regardless. Then, when FKA is **ON**, this widget could potentially serve as the mechanism that passes the relevant data to the accessibility bridge.

Here is how I'm thinking about the two paths forward:

### Scenario 1: FKA is customizable

If iOS allows us to customize the native FKA focus ring, `CupertinoFocusHalo` would serve a dual purpose. For basic keyboard traversal (FKA OFF), it would render the visual halo itself. For FKA traversal (FKA ON), it would pass the custom shape through the accessibility bridge to the platform. As FKA overlay is so intrusive, we wouldn't necessarily even have to know if FKA is on or not (AFAIK we have no way of knowing), but we could render the visual halo ourselves and let FKA render its own halo on top of it. 

In this case, `CupertinoFocusHalo` would function exactly as you suspected `UIFocusHaloEffect` does natively, acting as the bridge to style the FKA halo. I still need to investigate exactly how `UIFocusHaloEffect` behaves on iOS to confirm this.

### Scenario 2: FKA is strictly separate and not customizable

It’s also possible that FKA is (and is intended to be) a completely separate system from basic keyboard traversal, with no room for customization. If that’s the case, our goal is simpler: we just need to ensure that our Cupertino components look and behave correctly under *both* systems independently. So handling our own basic traversal visuals when FKA is OFF, and letting the system take over when FKA is ON.

Issue #173843, opened 2025-08-15
[Windows] Windows tasks aren't processed if the Dart event loop is never drained


krll-kov:
# Updated details

The Windows event loop never gets a chance to process native tasks if the Dart event loop is never empty.

<details>
<summary>Sample app...</summary>

```dart
import 'package:flutter/material.dart';
import 'dart:async';

void main() {
  runApp( MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Directionality(
      textDirection: TextDirection.ltr,
      child: GestureDetector(
        onTap: () => unawaited(_startEvilLoop()),
        child: Text('Tap me!'),
      ),
    );
  }

  Future<void> _startEvilLoop() async {
    // Spam the Dart event loop with tasks.
    // This isn't ideal, but the app should continue to work.
    while (true) {
      await Future.delayed(Duration.zero);
    }
  }
}
```

</details>

On Windows with merged threads enabled, tapping on the text causes the app to become unresponsive. However, this app works as expected on macOS, or on Windows if merged threads are disabled. 

----

# Original issue

1. Upgrade to flutter 3.35.1
2. Update dependencies to these (needed just for custom window movement)
3. Paste code into main.dart
4. Long press the left mouse button at the center part of the app to start app window dragging 
5. Unpress mouse button - onPanEnd is not called6. Downgrade flutter to previous version and launch the sam app, onPanEnd will be called there

Reason for using custom dragging is that you can remove default System tabbar for app by changed WS_OVERLAPPEDWINDOW to WS_SYSMENU | WS_MINIMIZEBOX to create a window window native tabbar (custom UI)

pubsspeck:

```dart
dependencies:
  ffi: ^2.1.4
  win32: ^5.14.0
  flutter:
    sdk: flutter
```

<details>
<summary>main.dart...</summary>

```dart
import 'dart:ffi';
import 'dart:ui';

import 'package:ffi/ffi.dart';
import 'package:flutter/material.dart';
import 'package:win32/win32.dart';

int isDragging = 0;
int _hWnd = 0;

int get getAppWindowHWND {
  if (_hWnd == 0) {
    final Pointer<Utf16> _lpClassName = "FLUTTER_RUNNER_WIN32_WINDOW"
        .toNativeUtf16();
    _hWnd = FindWindow(_lpClassName, nullptr);
    malloc.free(_lpClassName);
  }
  return _hWnd;
}

Future<void> startAppWindowDragging(
[...]

PR #179249, opened 2025-11-30
[win32] Replace threadpool timer with custom background thread timer

### LINKED ISSUE #178916: Windows significant performance regression after update from 3.35.7 to 3.38.2
# Workaround

If your Flutter Windows application is affected by this issue, please update your `windows/runner/main.cpp` file:

```diff
  flutter::DartProject project(L"data");

+ // TODO: Remove this.
+ // This forces Flutter to use a separate thread for Dart.
+ // This mode will be removed in a future version of Flutter.
+ project.set_ui_thread_policy(flutter::UIThreadPolicy::RunOnSeparateThread);
```

This is a temporary workaround until we fix this issue. 

# Issue

FPS chart (from MSI Afterburner) shows that release version of the same flutter app works good on previous stable release and bad on new stable release.

https://github.com/user-attachments/assets/80cfe49b-d198-4694-a77b-4cc2110fd136

Old stable release: constant 120fps during animations
New stable release: up to 40fps during animations

Also, the task manager shows that overall GPU usage on new flutter version is x3-4 times lower then on previous stable version (for some reason flutter does not use all gpu power now and limits itself), while cpu usage is very low on both versions.

### flutter run -d windows --profile tests:
3.35.7:
Image

[dart_devtools_2025-11-21_11_59_04.351.json](https://github.com/user-attachments/files/23674190/dart_devtools_2025-11-21_11_59_04.351.json)

3.38.2:
Image

[dart_devtools_2025-11-21_12_07_01.041.json](https://github.com/user-attachments/files/23674328/dart_devtools_2025-11-21_12_07_01.041.json)

Issue is NOT linked to this specific code sample, i can reproduce it with any flutter app, but this is the code that was used to demonstrate regression

```dart

void main() {
  runApp(const MaterialApp(
    home: WindowsPerformanceDemo(),
  ));
}

class WindowsPerformanceDemo extends StatelessWidget {
  const WindowsPerformanceDemo({super.key});

  @override
    return Scaffold(
      backgroundColor: const Color(0xFFF0F2F5),
      body: Center(
        child: SizedBox(
          width: 1000,
          height: 600,
          child: GridView.builder(
            padding: const EdgeInsets.all(20),
            gridDelegate: const SliverGridDelegateWithFixedCrossAxisCount(
              crossAxisCount: 4,
              crossAxisSpacing: 20,
              mainAxisSpacing: 20,
              childAspectRatio: 16 / 9,
            ),
            itemCount: 8,
            itemBuilder: (context, index) {
              return DesktopHoverCard(index: index);
            },
          ),
        ),
      ),
    );
  }
}

class DesktopHoverCard extends StatefulWidget {
  final int index;
  const DesktopHoverCard({super.key, required this.index});

  @override
  State<DesktopHoverCard> createState() => _DesktopHoverCardState();
}

class _DesktopHoverCardState extends State<DesktopHoverCard> {
  bool _isHovered = false;

  @override
    return MouseRegion(
      onEnter: (_) => setState(() => _isHovered = true),
      onExit: (_) => setState(() => _isHovered = false),

      cursor: SystemMouseCursors.click,
      child: AnimatedScale(
        scale: _isHovered ? 1.05 : 1.0,
        duration: const Duration(milliseconds: 200),
        curve: Curves.easeOutCubic,
        child: Container(
          decoration: BoxDecoration(
            color: Colors.white,
            borderRadius: BorderRadius.circular(16),
            boxShadow: [
              BoxShadow(
                color: Colors.black.withOpacity(_isHovered ? 0.2 : 0.1),
                blurRadius: _isHovered ? 20 : 10,
                offset: const Offset(0, 4),
              ),
            ],
          ),
          clipBehavior: Clip.antiAlias,
          child: Stack(
            fit: StackFit.expand,
            children: [
              Image.network(
                'https://picsum.photos/id/${widget.index + 50}/500/300',
                fit: BoxFit.cover,
              ),
              Container(
                decoration: const BoxDecoration(
                  gradient: LinearGradient(
                    begin: Alignment.topCenter,
                    end: Alignment.bottomCenter,
                    colors: [Colors.transparent, Colors.black54],
                  ),
                ),
              ),
              Align(
                alignment: Alignment.bottomLeft,
                child: Padding(
                  child: Text(
                    'Windows Item ${widget.index + 1}',
                    style: const TextStyle(
                      color: Colors.white,
                      fontSize: 18,
                      fontWeight: FontWeight.bold,
                    ),
                  ),
                ),
              ),
            ],
          ),
        ),
      ),
    );
  }
}
```

Doctor v

```dart
[√] Flutter (Channel stable, 3.38.2, on Microsoft Windows [Version 10.0.26100.6899], locale en-US) [359ms]
    • Flutter version 3.38.2 on channel stable at C:\SDKS\flutter
    • Framework revision f5a8537f90 (3 days ago), 2025-11-18 09:27:21 -0500
    • Engine revision b5990e5ccc
    • Dart version 3.10.0
    • DevTools version 2.51.1
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file,
      enable-lldb-debugging

[√] Windows Version (Windows 11 or higher, 24H2, 2009) [4.5s]

[X] Android toolchain - develop for Android devices [239ms]
    X Unable to locate Android SDK.
      (or visit https://flutter.dev/to/windows-android-setup for detailed instructions).


[X] Chrome - develop for the web (Cannot find Chrome executable at .\Google\Chrome\Application\chrome.exe) [161ms]

[√] Visual Studio - develop Windows apps (Visual Studio Professional 2022 17.14.21 (November 2025)) [158ms]
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Professional
    • Visual Studio Professional 2022 version 17.14.36717.8

[√] Connected device (2 available) [196ms]
    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [Version 10.0.26100.6899]
    • Edge (web)        • edge    • web-javascript • Microsoft Edge 142.0.3595.90

[√] Network resources [380ms]

```

knopp:
Fixes https://github.com/flutter/flutter/issues/178916

It seems that for some reason the win32 thread pool timer is not precise enough for our use case which causes major frame pacing issues on Windows. This PR replaces it with a simple custom built background thread timer.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/windows/task_runner_unittests.cc
+++ b/engine/src/flutter/shell/platform/windows/task_runner_unittests.cc
@@ -2,13 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include <atomic>
-#include <chrono>
 #include "flutter/shell/platform/windows/task_runner.h"
 
+#include <atomic>
+#include <chrono>
+
 #include "flutter/fml/macros.h"
+#include "flutter/shell/platform/windows/task_runner_window.h"
 #include "gtest/gtest.h"
-#include "shell/platform/windows/task_runner_window.h"
 
 namespace flutter {
 namespace testing {
--- a/engine/src/flutter/shell/platform/windows/task_runner_unittests.cc
+++ b/engine/src/flutter/shell/platform/windows/task_runner_unittests.cc
@@ -2,10 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <atomic>
+#include <chrono>
 #include "flutter/shell/platform/windows/task_runner.h"
 
 #include "flutter/fml/macros.h"
 #include "gtest/gtest.h"
+#include "shell/platform/windows/task_runner_window.h"
 
 namespace flutter {
 namespace testing {
@@ -86,6 +89,30 @@ TEST(TaskRunnerTest, MaybeExecuteTaskOnlyExpired) {
 
   std::set<uint64_t> only_task_expired_before_now{task_expired_before_now};
   EXPECT_EQ(executed_task, only_task_expired_before_now);
+}
+
+TEST(TaskRunnerTest, TimerThreadDoesNotCancelEarlierScheduledTasks) {
+  std::atomic_bool signaled = false;
+  std::optional<std::chrono::high_resolution_clock::time_point> callback_time;
+  TimerThread timer_thread([&]() {
+    callback_time = std::chrono::high_resolution_clock::now();
+    signaled = true;
+  });
+  timer_thread.Start();
+  auto now = std::chrono::high_resolution_clock::now();
+  // Make sure that subsequent call to schedule does not cancel earlier task
+  // as documented in TimerThread::ScheduleAt.
+  timer_thread.ScheduleAt(now + std::chrono::milliseconds(20));
+  timer_thread.ScheduleAt(now + std::chrono::seconds(10));
+
+  while (!signaled.load()) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(5));
+  }
+
+  EXPECT_TRUE(callback_time.has_value());
+  EXPECT_GE(*callback_time, now + std::chrono::milliseconds(20));
+
+  timer_thread.Stop();
 }
 
 }  // namespace testing
--- a/engine/src/flutter/shell/platform/windows/task_runner_window.cc
+++ b/engine/src/flutter/shell/platform/windows/task_runner_window.cc
@@ -15,81 +15,75 @@
 
 namespace flutter {
 
-// Background timer thread. Necessary because neither SetTimer nor
-// CreateThreadpoolTimer have good enough accuracy not to affect the
-// framerate.
-class TimerThread {
- public:
-  explicit TimerThread(std::function<void()> callback)
-      : callback_(std::move(callback)),
-        next_fire_time_(
-            std::chrono::time_point<std::chrono::high_resolution_clock>::max()),
-        thread_([this]() { TimerThreadMain(); }) {}
+TimerThread::TimerThread(std::function<void()> callback)
+    : callback_(std::move(callback)),
+      next_fire_time_(
+          std::chrono::time_point<std::chrono::high_resolution_clock>::max()) {}
 
-  void Stop() {
-    {
-      std::lock_guard<std::mutex> lock(mutex_);
-      callback_ = nullptr;
+void TimerThread::Start() {
+  FML_DCHECK(!thread_);
+  thread_ =
+      std::make_optional<std::thread>(&TimerThread::TimerThreadMain, this);
+}
+
+void TimerThread::Stop() {
+  if (!thread_) {
+    return;
+  }
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    callback_ = nullptr;
+  }
+  cv_.notify_all();
+  thread_->join();
+}
+
+TimerThread::~TimerThread() {
+  // Ensure that Stop() has been called if Start() has been called.
+  FML_DCHECK(callback_ == nullptr || !thread_);
+}
+
+// Schedules the callback to be called at specified time point. If there is
+// already a callback scheduled earlier than the specified time point, does
+// nothing.
+void TimerThread::ScheduleAt(
+    std::chrono::time_point<std::chrono::high_resolution_clock> time_point) {
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (time_point < next_fire_time_) {
+      next_fire_time_ = time_point;
     }
+    ++schedule_counter_;
     cv_.notify_all();
-    thread_.join();
   }
+}
 
-  ~TimerThread() {
-    // Ensure that Stop() has been called.
-    FML_DCHECK(callback_ == nullptr);
-  }
-
-  // Schedules the callback to be called at specified time point. If there is
-  // already a callback scheduled earlier than the specified time point, does
-  // nothing.
-  void ScheduleAt(
-      std::chrono::time_point<std::chrono::high_resolution_clock> time_point) {
-    {
-      std::lock_guard<std::mutex> lock(mutex_);
-      if (time_point < next_fire_time_) {
-        next_fire_time_ = time_point;
-      }
-      ++schedule_counter_;
-      cv_.notify_all();
+void TimerThread::TimerThreadMain() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (callback_ != nullptr) {
+    cv_.wait_until(lock, next_fire_time_, [this]() {
+      return std::chrono::high_resolution_clock::now() >= next_fire_time_ ||
+             callback_ == nullptr;
+    });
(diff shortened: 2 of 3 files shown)


engine/src/flutter/shell/platform/windows/task_runner_window.cc
```diff
@@ -7,11 +7,85 @@
...
+  // Schedules the callback to be called at specified time point. If there is
+  // already a callback scheduled earlier than the specified time point, does
+  // nothing.
+  void ScheduleAt(
+      std::chrono::time_point<std::chrono::high_resolution_clock> time_point) {
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      if (time_point < next_fire_time_) {
+        next_fire_time_ = time_point;
+      }
```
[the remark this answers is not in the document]
knopp: No. The behavior is on purpose. If anything the name of the function may need to be changed to reflect that.

engine/src/flutter/shell/platform/windows/task_runner_window.cc
```diff
@@ -143,18 +203,7 @@ void TaskRunnerWindow::ProcessTasks() {
...
-        std::chrono::duration_cast<std::chrono::microseconds>(when).count();
-    ULARGE_INTEGER ticks;
-    ticks.QuadPart = -static_cast<LONGLONG>(microseconds * 10);
-    FILETIME ft;
-    ft.dwLowDateTime = ticks.LowPart;
-    ft.dwHighDateTime = ticks.HighPart;
-    SetThreadpoolTimer(timer_, &ft, 0, 0);
-  }
+  timer_thread_->ScheduleAt(std::chrono::high_resolution_clock::now() + when);
 }
```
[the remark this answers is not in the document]
knopp: Nice one.

engine/src/flutter/shell/platform/windows/task_runner_window.cc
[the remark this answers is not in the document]
knopp: Added test to ensure this works as expected.

engine/src/flutter/shell/platform/windows/task_runner_unittests.cc
```diff
@@ -2,10 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <atomic>
+#include <chrono>
 #include "flutter/shell/platform/windows/task_runner.h"
 
 #include "flutter/fml/macros.h"
 #include "gtest/gtest.h"
+#include "shell/platform/windows/task_runner_window.h"
```
loic-sharma: Nit:

```suggestion
#include "flutter/shell/platform/windows/task_runner.h"

#include <atomic>
#include <chrono>

#include "flutter/fml/macros.h"
#include "flutter/shell/platform/windows/task_runner_window.h"
#include "gtest/gtest.h"
```

engine/src/flutter/shell/platform/windows/task_runner_window.cc
```diff
@@ -7,30 +7,90 @@
...
+    std::chrono::time_point<std::chrono::high_resolution_clock> time_point) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (time_point < next_fire_time_) {
+    next_fire_time_ = time_point;
+  }
+  ++schedule_counter_;
+  cv_.notify_all();
+}
+
+void TimerThread::TimerThreadMain() {
```
loic-sharma, line 59: Should we be setting a thread priority here?

knopp: I'm not sure really. I don't know much about windows scheduler, even with the default values I'm getting sub-ms accuracy, but maybe the situation is worse with less CPU cores.

knopp: I'm leaning towards yes...

engine/src/flutter/shell/platform/windows/task_runner_window.cc
```diff
@@ -44,18 +104,10 @@ TaskRunnerWindow::TaskRunnerWindow() {
   }
 
   thread_id_ = GetCurrentThreadId();
-
-  // Increase timer precision for this process (the call only affects
-  // current process since Windows 10, version 2004).
-  timeBeginPeriod(1);
```
loic-sharma, line 50: Should we also remove the `timeEndPeriod(1)` in the destructor?

engine/src/flutter/shell/platform/windows/task_runner_window.cc
knopp: We could also use Multimedia Class Scheduler Service for the thread scheduling. This might need some experimenting. But maybe we can do that as a follow-up PR.

engine/src/flutter/shell/platform/windows/task_runner_window.cc
knopp: Yes. Done.

engine/src/flutter/shell/platform/windows/task_runner_window.cc
loic-sharma: We use "above normal" for the UI and raster threads: https://github.com/flutter/engine/pull/31778

I'd be OK with skipping thread priority for now and adding this in later if/when we find cases where this improves things.

loic-sharma (approved):
Thanks for the excellent fix!

Predidit:
Excuse me, could this fix be cherry-picked to the stable branch?

krll-kov:
> The fix here might have regressed #173843. With the new timer implementation now actually working as expected sometimes the one millisecond delay is not enough to for other windows messages to be processed :-/

So there's no sense in this fix because it does not do anything anymore?

knopp:
No. This PR fixes the framerate. But in some cases 

```dart
Future<void> _startEvilLoop() async {
    // Spam the Dart event loop with tasks.
    // This isn't ideal, but the app should continue to work.
    while (true) {
      await Future.delayed(Duration.zero);
    }
  }
```
Might lock up the application. Not that is a good idea to do this.

The initial workaround for the bug worked because the timer delay was rather big, which causes the framerate issue but at the same time avoided draining dart run loop too fast. Now that the framerate issue is fixed, we drain the dart event loop too fast taking precedence over win32 messages. To fix this with new timer we need better solution.

krll-kov:
> No. This PR fixes the framerate. But in some cases
> 
> ```dart
> Future<void> _startEvilLoop() async {
>     // Spam the Dart event loop with tasks.
>     // This isn't ideal, but the app should continue to work.
>     while (true) {
>       await Future.delayed(Duration.zero);
>     }
>   }
> ```
> 
> Might lock up the application. Not that is a good idea to do this.
> 
> The initial workaround for the bug worked because the timer delay was rather big, which causes the framerate issue but at the same time avoided draining dart run loop too fast. Now that the framerate issue is fixed, we drain the dart event loop too fast taking precedence over win32 messages. To fix this with new timer we need better solution.

That's what I mean, without timer and with it, the outcome is the same , that original issue with app freeze during loops with await of 0ms is still present, so it's presence seems pointless now

Issue #178839, opened 2025-11-20
[Impeller] Filled arcs with transparent colors have incorrect colors on iOS and macOS


b-luk:
### Steps to reproduce

Draw a filled arc with a transparent color on iOS and macOS with Impeller enabled

### Expected results

<img width="284" height="259" alt="Image" src="https://github.com/user-attachments/assets/a163c459-55e9-42da-b8ec-7a6683cf4d5f" />

### Actual results

<img width="300" height="260" alt="Image" src="https://github.com/user-attachments/assets/3fcacc39-13ea-41af-8047-e7788ff93698" />

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'dart:math' as math show pi;

import 'package:flutter/material.dart';

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Container(
      color: Colors.white,
      child: CustomPaint(painter: TestPainter()),
    );
  }
}

class TestPainter extends CustomPainter {
  @override
  void paint(Canvas canvas, Size size) {
    canvas.drawArc(
      Rect.fromCircle(center: size.center(Offset.zero), radius: 100),
      0,
      1.5 * math.pi,
      true,
      Paint()
        ..color = Colors.blue.withAlpha(127)
        ..style = PaintingStyle.fill,
    );
  }

  @override
  bool shouldRepaint(covariant CustomPainter oldDelegate) => false;
}

```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

See above

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[Paste your output here]
```

</details>

tototo23:
Hi!
I'm having the same issue here.
With the same source code : 
- works well on Windows, Android, web
- have this wrong behavior only on iOS (tested in iOS 17 and 26)

The render was good with Flutter 3.32.8.
I am now with Flutter 3.35.7 where this iOS render is wrong.
Didn't try yet with Flutter 3.38.2 to see if the issue is solved (my tests is scheduled for next week)

flar:
Bisecting turned up this commit which means it has been in there as long as the "new arc tessellator" has been in there.

It doesn't happen if we can use triangle fans - which would explain why it doesn't reproduce on Android if Android supports fans.

[...]

PR #179312, opened 2025-12-02
Use kPreventOverdraw for arcs with overlapping stroke caps

### LINKED ISSUE #178746: [Impeller] Fast arc stroke generator can overdraw when caps intersect
### Steps to reproduce

Draw a circular stroked arc with a transparent color. Use square stroke caps, and use a high sweep angle and stroke width so that the two square caps intersect.

This was an issue discovered and discussed with @flar. The fast arc stroke tessellation introduced in #169402 can draw overlapping geometry. The fast arc stroke tessellation does not take overdraw into account, causing incorrect coloring issues when used with transparent colors. This is similar to #174579.

Currently this overlap can only happen when using square caps. After #169400 is fixed, this will also happen with round caps.

This can be fixed by using `GeometryResult::Mode::kPreventOverdraw`, or by not using the fast arc stroke geometry for arcs with overlapping caps. We may want to performance comparisons for these two approaches in order to decide which one to use.

### Expected results

Overlapping caps should be the same color as the rest of the arc.

Image

### Actual results

Overlapping caps get overdrawn. The color is composited multiple times, causing a color mismatch with the rest of the arc.

Image

### Code sample

```dart
import 'dart:math' as math show pi;

void main() {
  runApp(MyApp());
}

  @override
    return Container(
      color: Colors.white,
      child: CustomPaint(painter: TestPainter()),
    );
  }
}

class TestPainter extends CustomPainter {
  @override
  void paint(Canvas canvas, Size size) {
    canvas.drawArc(
      Rect.fromCircle(center: size.center(Offset.zero), radius: 50),
      0,
      1.5 * math.pi,
      false,
      Paint()
        ..color = Colors.blue.withAlpha(127)
        ..strokeWidth = 80
        ..strokeCap = StrokeCap.square,
    );
  }

  @override
  bool shouldRepaint(covariant CustomPainter oldDelegate) => false;
}

```

</details>

### Screenshots or Video

See screenshots above.

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel master, 3.39.0-1.0.pre-181, on macOS 15.7.1 24G231 darwin-arm64, locale en)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Xcode - develop for iOS and macOS (Xcode 16.3)
[✓] Network resources

• No issues found!
```

</details>

b-luk:
Fixes #178746

Before:
image

After:
image

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,45 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+      auto half_width = stroke_width_ / 2.0f;
+      auto inner_radius = radius - half_width;
+      auto inner_arc_end_x =
+          cos(Radians(arc_.GetSweep()).radians) * inner_radius;
+      auto inner_square_cap_end_x =
+          inner_arc_end_x +
+          cos(Radians(arc_.GetSweep() + Degrees(90)).radians) * half_width;
+      return inner_square_cap_end_x > inner_radius;
+    }
+    case Cap::kRound: {
```
flar, line 118: Is this easier to compute in the method that has the Arc::Iterations structure?

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,45 @@ bool ArcGeometry::IsAxisAlignedRect() const {
   return false;
 }
 
+bool ArcGeometry::CapsOverlap() const {
+  FML_DCHECK(arc_.GetSweep().degrees >= 0.0f);
+  FML_DCHECK(arc_.GetSweep().degrees <= 360.0f);
+
+  if (stroke_width_ < 0 || cap_ == Cap::kButt ||
```
flar, line 92: What if stroke_width is much larger than the size of the arc? Does it overlap just by turning a little?

flar (changes requested):
Some food for thought. I don't know if this can be moved somewhere where we compute the Arc::Iterator or not. It would give you helpful delta vectors to compute where the ends are.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Stroke widths up to `2 * radius` behave properly. With these stroke widths, there is overlap only if `sweep_angle > 180`. And we have a check in the next line, line 93, to return "no overlap" when `arc_.GetSweep().degrees <= 180`. 

Example with a stroke width close to 2\*r. (radius is 50, 2\*r is 100). Overlap only occurs when the sweep angle passes 180 degrees.
2r-stroke-width-overlap

For stroke widths greater than `2 * radius`:

I had tested this before (apparently only with Impeller enabled), and thought that we didn't need to handle this case because Impeller doesn't render caps at all for arcs with stroke widths greater than `2 * radius`.

I just tested this case again. This time I also tested the behavior with Impeller disabled, using Skia. Skia does render caps for stroke widths greater than `2 * radius`.

Skia:
big-stroke-width-skia

Impeller:
Note that the when the arc switches to a filled arc, it has a stripey pattern. This is a separate issue tracked by #178839.
big-stroke-width-impeller

Impeller falls back to [this code](https://github.com/flutter/flutter/blob/2c4d19119ec302dea203a7ca5288fa4301735118/engine/src/flutter/impeller/display_list/canvas.cc#L721-L731) and renders the arc as a filled arc. This drops the caps from the arc. I'll file a new issue about this. *Edit: opened https://github.com/flutter/flutter/issues/179369*

So for the current behavior of Impeller, the caps can only overlap when `sweep_angle > 180`. The check in this commit sufficiently handles all cases where caps may overlap.

If we change Impeller so that it renders `stroke_width > 2 * radius` arcs in the same way as Skia, then there could be overlap for smaller sweep angles. If/when we make this change we should make sure the overlap check here is updated to handle these cases.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Right now the Arc class and its Iterations helper structure define an arc that consists of bounds, start/sweep angles, and include_center. The ArcGeometry class combines an Arc with a Paint, giving it a stroke width and cap type.

These cap overlap calculations depend on the stroke width and cap type. So I compute them in the ArcGeometry class, rather than for Arc or Arc::Iterations which have no knowledge about these Paint properties.

We could consider moving the Paint properties to be in Arc instead of ArcGeometry. And that way it would make more sense for the calculations to happen in the Arc class. But that would be a larger change, and I don't know if there are intentional reasons why these properties were put in ArcGeometry instead of the Arc class. (Maybe there are other Geometrys or other users of Arcs which don't deal with Paint/caps/strokes, so it this doesn't belong in Arc?)

Looking at the implementations of Arc::Iterations and Arc::ComputeIterations, I don't really see anything there which would help make the caps overlap calculations easier to compute. But I could be missing something.

So based on my current understanding of things, I think these calculations make sense here in ArcGeometry. Let me know if this makes sense or if you think otherwise.

flar: It looks like the Iterations was only computed temporarily in the tessellation functions so it is long gone by the time this code executes.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,45 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+      // https://github.com/flutter/flutter/issues/178746#issuecomment-3554526727
+      auto radius = arc_.GetOvalSize().width / 2.0f;
+      auto half_width = stroke_width_ / 2.0f;
+      auto inner_radius = radius - half_width;
+      auto inner_arc_end_x =
+          cos(Radians(arc_.GetSweep()).radians) * inner_radius;
+      auto inner_square_cap_end_x =
+          inner_arc_end_x +
+          cos(Radians(arc_.GetSweep() + Degrees(90)).radians) * half_width;
+      return inner_square_cap_end_x > inner_radius;
```
flar, line 116: This seems odd only calculating the X coordinate. Does it work at all angles? I notice you only test this with 0->330 arcs. You should add tests that cover a wide variety of angle conditions.

flar (changes requested):

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: The video at https://github.com/flutter/flutter/issues/178746#issuecomment-3554526727 shows that overlap can be determined using only the x coordinate. It should work for any sweep angle between 0 and 360. The video shows the relation between inner_square_cap_end_x (red line) and inner_radius (blue line) as the arc's sweep angle goes through a variety of angles across all 4 quadrants. Overlap occurs iff the red line > the blue line.

The video only shows a start angle of 0. These calculations are entirely independent of start angle, so we can treat any arc as if its start angle is 0 for the purposes of these calculation.

I updated the unit tests to test a lot more sweep angles.

flar: Now I see what my confusion was. I was thinking along the lines of "but what if it starts at 45 degrees or 90 degrees then don't you need to use Y instead or combine the two"?

But, the point I was missing is that the start angle has no bearing on the results (other than to perhaps complicate the test) so the test can be made as if the start angle was 0. And if the start angle is 0, then only the X coordinate is needed.

Please add a comment mentioning that start angle is not relevant and computing as if it starts at 0 only requires a test on the X coordinate...?

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
flar: OK with the subsequent issue filed...

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -47,7 +48,11 @@ GeometryResult ArcGeometry::GetPositionBuffer(const ContentContext& renderer,
     auto generator =
         renderer.GetTessellator().StrokedArc(transform, arc_, cap_, half_width);
 
-    return ComputePositionGeometry(renderer, generator, entity, pass);
+    auto result = ComputePositionGeometry(renderer, generator, entity, pass);
+    if (CapsOverlap()) {
+      result.mode = GeometryResult::Mode::kPreventOverdraw;
```
flar, line 53: Perhaps in a follow-on PR, we should consider having the generator specify the overlap type rather than patching it after the fact.

engine/src/flutter/impeller/entity/entity_unittests.cc
```diff
@@ -2313,6 +2313,98 @@ TEST_P(EntityTest, FillPathGeometryGetPositionBufferReturnsExpectedMode) {
...
+  {
+    StrokeParameters stroke = {.width = 50.0f, .cap = Cap::kButt};
+    for (auto sweep = 0; sweep < 360; sweep += 5) {
+      auto geometry = Geometry::MakeStrokedArc(oval_bounds, Degrees(0),
+                                               Degrees(sweep), stroke);
+
+      GeometryResult result =
+          geometry->GetPositionBuffer(*GetContentContext(), {}, mock_pass);
+
+      EXPECT_EQ(result.mode, GeometryResult::Mode::kNormal);
```
flar: 2 (yeah, that's the right number) suggestions:

- when a test is parameterized, only the line number appears in the output so it is hard to figure out which parameter is failed. This can be solved by using the fact that the EXPECT and ASSERT macros return a stream object so you can append ` << "sweep: " << sweep` to the end of the line and that will print out after a failure (try adding in a test that fails to see this in operation).
- It couldn't hurt to add a for loop for start angles to make sure this works for all start angles as well. (As well as `<< start` stream indicators to the EXPECT macros.)
- (Ignore that this is point #3) I assume that sweeps of 360+ turn into circles so we might not be testing the same tessellation code. One thing that does happen is that if you include the center, it doesn't turn into a circle for stroking because both spokes still need to be drawn.
- (Ignore the count of points, I'm on a roll) Hmmm, does the case of include center stroked arcs that go beyond 360 degrees of sweep need overlap protection?

flar (changes requested):
It all looks good, but I think I might have uncovered some related issues in my evaluation of the unit tests that bears investigation while we are solving the overlap issues.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Done. Added a comment. I also changed the round cap test to treat the start angle as 0, because this simplifies that test a bit too.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Acknowledged. I'll follow up with this.

engine/src/flutter/impeller/entity/entity_unittests.cc
b-luk: - Added sweep angle output.
- Added a for loop with start angles and added start angle output.
- Arcs that include the center do not use this tessellation code. [This check](https://github.com/flutter/flutter/blob/26f311ca5e8519f31b10a2fca6f10b6eaa028e04/engine/src/flutter/impeller/display_list/canvas.cc#L733-L740) makes stroked arcs that include the center use the path-based ArcStrokeGeometry code. These arcs will use the overdraw protection that is used by path-based strokes.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,52 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+    return false;
+  }
+
+  switch (cap_) {
+    case Cap::kSquare: {
+      // Square caps overlap if the inner corner of the ending cap extends
+      // inside the inner edge of the start cap. For a visualization of when
+      // this occurs, see
+      // https://github.com/flutter/flutter/issues/178746#issuecomment-3554526727
+      // Note that testing for overlap is completely non-dependent on the arc's
```
flar: independent of

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,52 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+    case Cap::kRound: {
+      // Round caps overlap if the distance between the arc's start and end
+      // points is less than the stroke width.
+      // https://github.com/flutter/flutter/issues/178746#issuecomment-3554526727
+      // Note that testing for overlap is completely non-dependent on the arc's
+      // start angle. To simplify the overlap test, we treat the arc as if its
+      // start angle is 0.
+      auto radius = arc_.GetOvalSize().width / 2.0f;
+      auto start_point = Point(radius, 0);
+      auto sweep_angle = Radians(arc_.GetSweep()).radians;
```
flar: nit: sweep_radians since you've resolved it to radians and not just an angle

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,52 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+      auto end_point = Point(cos(sweep_angle), sin(sweep_angle)) * radius;
+      return (end_point - start_point).GetLength() < stroke_width_;
```
flar: `start_point.GetDistanceSquared(end_point) < stroke_width * stroke_width` to avoid the sqrt

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
```diff
@@ -80,4 +85,52 @@ bool ArcGeometry::IsAxisAlignedRect() const {
...
+      // inside the inner edge of the start cap. For a visualization of when
+      // this occurs, see
+      // https://github.com/flutter/flutter/issues/178746#issuecomment-3554526727
+      // Note that testing for overlap is completely non-dependent on the arc's
+      // start angle. To simplify the overlap test, we treat the arc as if its
+      // start angle is 0. This allows the test to only require checking the x
+      // coordinate of the ending cap, rather than needing to calculate overlap
+      // based on both x and y positions of both caps.
+      auto radius = arc_.GetOvalSize().width / 2.0f;
+      auto half_width = stroke_width_ / 2.0f;
```
flar: `* 0.5f` might be faster than `/ 2.0f` depending on compiler opts.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
flar: Good comment. Grammar suggestion...

flar (changes requested):
LGTM, but we should have a golden that also verifies the output but I didn't see any golden changes from existing tests. Maybe a variety of angles and widths with each end cap. I don't think you need to test the non-overlaps (maybe 1 for reference?), just make sure a variety of overlaps don't overdraw. It might require 3 separate tests? Or if you can pack in a reasonably visible set of tests into one image that would be cool too. But we should have had golden changes for this and the fact that we didn't indicates a hole in our testing.

Some minor nits in the comments that would be nice to adopt.

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Done

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Done

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Done

engine/src/flutter/impeller/entity/geometry/arc_geometry.cc
b-luk: Done

flar (approved):
LGTM

flar:
Goldens LGTM

PR #179355, opened 2025-12-02
Implementation of `TestWindowingOwner` for testing multi-window

mattkae:
## What's new?
This pull request makes it so that multi window applications can be tested via `testWidgets`. This PR implements the following:

- A `_TestWindowingOwner` that is used by the `TestWidgetsFlutterBinding`
- A `_TestRegularWindowController` implementation
- A `_TestDialogWindowController` implementation
- A `_TestFlutterView` and `_TestDisplay` for use by the controllers
- Proper `isActivated` status across controllers
- Proper minimization, maximization, and fullscreen status
- Proper sizing (based on constraints)
- Using the new testing abilities in the multiple windows example app

I have purposefully not implemented tooltip windows yet, as they are still awaiting an implementation in Win32 to be useful.

## Pre-launch Checklist

- [x] I signed the [CLA].


examples/multiple_windows/test/multiple_windows_test.dart
```diff
@@ -0,0 +1,121 @@
...
+See: https://github.com/flutter/flutter/issues/30701.
+''';
+    // ignore: avoid_print
+    print(windowingDisabledErrorMessage);
+    exit(0);
+  }
+
+  testWidgets('Multiple windows smoke test', (WidgetTester tester) async {
+    multiple_windows
+        .main(); // builds the app and schedules a frame but doesn't trigger one
```
loic-sharma: The formatter's formatting is a bit funky with long trailing comments, consider moving the comment above (here and all subsequent tests)

```suggestion
    // Builds the app and schedules a frame but doesn't trigger one
    multiple_windows.main();
```

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -1489,7 +1489,7 @@ See: https://github.com/flutter/flutter/issues/30701.
   /// This setter can be used to provide a custom [WindowingOwner].
   ///
   /// {@macro flutter.widgets.binding.window.experimental}
-  @internal
+  @protected
```
loic-sharma: I'd prefer to keep `@internal` here if possible - it is a strong signal that users should not use this API, even if they're extending widgets binding.

I'm guessing you changed this so that it could be set in `flutter_test` without causing a diagnostic? Could we instead suppress the diagnostic where needed? I see you already added `// ignore_for_file: invalid_use_of_internal_member` - was that not enough?

packages/flutter_test/lib/src/binding.dart
```diff
@@ -400,6 +898,10 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
       binding.setupHttpOverrides();
     }
     _testTextInput = TestTextInput(onCleared: _resetFocusedEditable);
+
+    if (isWindowingEnabled) {
```
loic-sharma: We should probably do something similar to text input: this mock should be on for unit tests, but off for integration tests.

See:

https://github.com/flutter/flutter/blob/430800f8c01589b76f150dcbe82e87eb88e7cd61/packages/flutter_test/lib/src/binding.dart#L310-L333

And it looks like this is where integration tests turn it off:

https://github.com/flutter/flutter/blob/430800f8c01589b76f150dcbe82e87eb88e7cd61/packages/integration_test/lib/integration_test.dart#L98-L99

packages/flutter_test/lib/src/binding.dart
```diff
@@ -148,6 +154,498 @@ class CapturedAccessibilityAnnouncement {
...
+    BoxConstraints? preferredConstraints,
+    String? title,
+  }) : _delegate = delegate,
+       _size = preferredSize ?? const Size(800, 600),
+       _constraints = preferredConstraints ?? BoxConstraints.loose(const Size(1920, 1080)),
+       _title = title ?? 'Test Window',
+       super.empty() {
+    _constrainToBounds();
+    // QUESTION: Do I have to do anything smart to get this into the list of views
+    // from the PlatformDispatcher?
```
loic-sharma: It looks like the best way to do this would be to add new members to `TestPlatformDispatcher`, similar to how we let you mock `PlatformDispatcher.locales` today:

https://github.com/flutter/flutter/blob/5cf560e08ce5156af661307c9899541790478193/packages/flutter_test/lib/src/window.dart#L275-L290

I'd consider adding this plumbing as part of this PR, otherwise tests might have weird bugs due to views that don't exist in the platform dispatcher.

packages/flutter_test/lib/src/binding.dart
loic-sharma: I'd be OK with this being a TODO for later

examples/multiple_windows/test/multiple_windows_test.dart
mattkae: Unnecessary comment anyway in my opinion :)

packages/flutter_test/lib/src/binding.dart
mattkae: I can add it now!

examples/multiple_windows/test/multiple_windows_test.dart
```diff
@@ -0,0 +1,115 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: invalid_use_of_internal_member
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+// ignore: avoid_relative_lib_imports
+import '../lib/main.dart' as multiple_windows;
```
justinmc: Can this be imported via `package:multiple_windows`? Like:

https://github.com/flutter/flutter/blob/f63cf73d310065a11b67d82ba88618e342227c7c/examples/hello_world/test/hello_test.dart#L6

examples/multiple_windows/test/multiple_windows_test.dart
```diff
@@ -0,0 +1,115 @@
...
+  testWidgets('Can create a regular window', (WidgetTester tester) async {
+    multiple_windows.main();
+    await tester.pump(); // triggers a frame
+
+    final toTap = find.widgetWithText(OutlinedButton, 'Regular');
+    expect(toTap, findsOneWidget);
+    await tester.tap(toTap);
+    await tester.pump();
+
+    expect(find.widgetWithText(AppBar, 'Regular Window'), findsOneWidget);
```
justinmc, line 52: It's pretty cool that this can be found just like anything else, despite being in a separate window (in a real app).

packages/flutter_test/lib/src/binding.dart
```diff
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: invalid_use_of_internal_member
+// ignore_for_file: implementation_imports
```
justinmc, line 6: I guess we can get rid of these when we remove the multiwindow feature flag?

packages/flutter_test/lib/src/binding.dart
```diff
@@ -148,6 +154,482 @@ class CapturedAccessibilityAnnouncement {
...
+  /// Removes and destroys all child window controllers.
+  void removeAllChildren() {
+    for (final BaseWindowController child in _children) {
+      child.destroy();
+    }
+    _children.clear();
+  }
+
+  /// Returns the first activateable window in this window's hierarchy.
+  BaseWindowController getFirstActivateableChild() {
```
justinmc: Nit: We seem to use the spelling "activatable" a few places in the codebase and never include the "e", so maybe switch to that spelling.

packages/flutter_test/lib/src/binding.dart
```diff
@@ -148,6 +154,482 @@ class CapturedAccessibilityAnnouncement {
...
+  @override
+  TooltipWindowController createTooltipWindowController({
+    required TooltipWindowControllerDelegate delegate,
+    required BoxConstraints preferredConstraints,
+    required ui.Rect anchorRect,
+    required WindowPositioner positioner,
+    required BaseWindowController parent,
+  }) {
+    // TODO(mattkae): implement createTooltipWindowController
+    throw UnimplementedError();
```
justinmc, line 629: Is this for another PR? If so maybe link to an issue if you think it would help to remember it.

packages/flutter_test/lib/src/binding.dart
```diff
@@ -332,6 +814,20 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
   @protected
   bool get registerTestTextInput => true;
 
+  /// Determines whether the binding automatically register [windowingOwner] to
```
justinmc: "register" => "registers"

packages/flutter_test/lib/src/binding.dart
```diff
@@ -332,6 +814,20 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
...
+  /// the fake windowing owner implementation.
+  ///
+  /// Unit tests make use of this to mock out windowing system communication for
+  /// widgets. An integration test would set this to false, to test real windowing
+  /// system input.
+  ///
+  /// This property should not change the value it returns during the lifetime
+  /// of the binding. Changing the value of this property risks very confusing
+  /// behavior as the [WindowingOwner] may be inconsistently registered or
+  /// unregistered.
```
justinmc, line 827: I guess it's unlikely for this to happen since this is `protected`.

justinmc (approved):
LGTM 👍 . Thanks for circling back to this. It looks very straightforward for people to write tests for a multiwindow app just like they would for a regular Flutter app.

Just to check, what would it look like if I wanted to write a test that finds something only inside of any DialogWindow? What about only inside of some specific window? Maybe we should include tests like that at some point, but maybe not in this PR.

examples/multiple_windows/test/multiple_windows_test.dart
mattkae: The analyzer disliked this, but let me try again!

packages/flutter_test/lib/src/binding.dart
mattkae: Yup! There is a task in the pre-release checklist to rid ourselves of all of these at once

packages/flutter_test/lib/src/binding.dart
mattkae: I have it on my list to integrate it as part of the reset of the tooltip work, so it will be landing soon. No need for a ticket in my opinion

packages/flutter_test/lib/src/binding.dart
mattkae: True, I guess it's just a note for extenders, which is good to have

examples/multiple_windows/test/multiple_windows_test.dart
mattkae:
```
  error • Target of URI doesn't exist: 'package:multiple_windows/main.dart' • examples/multiple_windows/test/multiple_windows_test.dart:9:8 • uri_does_not_exist

```

PR #179359, opened 2025-12-02
Add project-specific analysis and suggestions to flutter doctor

jwren:
- Adds `ProjectAnalysisValidator` to run `dart analyze` on the current project.
- Integrates `dart pub outdated` to check for outdated dependencies.
  - Uses `--json` output to parse and report the exact count of outdated packages.
  - Displays actionable errors when outdated packages are found.
- Adds comprehensive tests for success, errors, warnings, lints, and outdated package checks.

### DIFF
--- a/packages/flutter_tools/lib/src/commands/doctor.dart
+++ b/packages/flutter_tools/lib/src/commands/doctor.dart
@@ -81,6 +81,7 @@ class DoctorCommand extends FlutterCommand {
     final bool isProject = project.manifest.appName.isNotEmpty;
     final bool disableProjectValidators = boolArg('disable-project-validators');
 
+    final projectValidatorTasks = <ValidatorTask>[];
     if (isProject && !disableProjectValidators) {
       final projectAnalysisValidator = ProjectAnalysisValidator(
         project: project,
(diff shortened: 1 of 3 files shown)


bkonyi (commented):
I did some quick performance measurements by running `flutter doctor` on the `$FLUTTER_ROOT/examples/splash` project with and without this change:

**With project-specific analysis**

`flutter doctor  13.80s user 3.95s system 171% cpu 10.348 total`

**Without project-specific analysis**

`flutter doctor  2.96s user 0.94s system 55% cpu 7.009 total`

There's a fairly significant performance regression caused by the extra analysis step, and this regression will likely scale with the size of the project. We run `flutter doctor` all the time in testing and on our CI to ensure the tool is properly initialized, so I'm not sure we should be doing project specific analysis as part of the default `flutter doctor` invocation.

Do we have plans for additional project-specific validators? If we can't enable this behavior by default, I think we'd need to have additional use cases in mind to make this a better solution than simply running `flutter analyze --suggestions`.

jwren:
Thanks for running and getting these metrics.

If `flutter analyze` is too slow, then what about not having it be enabled by default, and just having a message at the end of `flutter doctor` to run with the a `enable-project-validators` to have the `analyze`, `pub outdated` (and possibly future verifiers like `fix`) enabled.

As for `flutter analyze --suggestions` -- this is too slow to include in `flutter doctor`, and with the CLI output it hangs.  After this PR landed, my intention was to use the UX to point users to "validate their project with Gradle by calling `flutter analyze --suggestions`"

CC @reidbaker as FYI and any recommendations on the "validate their project with Gradle by calling `flutter analyze --suggestions`" verbiage



> I did some quick performance measurements by running `flutter doctor` on the `$FLUTTER_ROOT/examples/splash` project with and without this change:
> 
> **With project-specific analysis**
> 
> `flutter doctor 13.80s user 3.95s system 171% cpu 10.348 total`
> 
> **Without project-specific analysis**
> 
> `flutter doctor 2.96s user 0.94s system 55% cpu 7.009 total`
> 
> There's a fairly significant performance regression caused by the extra analysis step, and this regression will likely scale with the size of the project. We run `flutter doctor` all the time in testing and on our CI to ensure the tool is properly initialized, so I'm not sure we should be doing project specific analysis as part of the default `flutter doctor` invocation.
> 
> Do we have plans for additional project-specific validators? If we can't enable this behavior by default, I think we'd need to have additional use cases in mind to make this a better solution than simply running `flutter analyze --suggestions`.

jwren:
I haven't had time, closing for the time being.

PR #179387, opened 2025-12-03
docs: Impeller usage requirements for ImageFilter.shader and isShader…

### LINKED ISSUE #178767: Improve doc comments RE ImageFilter.shader and . isShaderFilterSupported
The doc comments should front-load the details about the API only usable with Impeller.

Perhaps using

> [!IMPORTANT]
> some loud markdown to get folks attention

I'd also argue that `isShaderFilterSupported` should reference how things tie to Impeller as well!

gowsikraja:
Summary

This PR improves the API documentation for ImageFilter.shader and isShaderFilterSupported by making the Impeller requirement more explicit and visible.

What was changed

Added an “Important” notice to clearly state that these APIs are only supported with Impeller.
Updated the doc comments to front-load this detail instead of placing it later in the text.
Added cross-reference notes explaining how isShaderFilterSupported relates to Impeller.
Why this change is helpful

Developers frequently run into confusion when using shader-based filters without realizing they require Impeller. This change makes the requirement obvious and easier to discover.

Issue reference

Fixes #178767

### DIFF
--- a/engine/src/flutter/lib/ui/painting.dart
+++ b/engine/src/flutter/lib/ui/painting.dart
@@ -4408,7 +4408,7 @@ abstract class ImageFilter {
   ///
   /// > [!WARNING]
   /// > This property indicates whether the Impeller rendering engine is enabled.
-  /// > If `false`, attempting to use [ImageFilter.shader] will throw an [UnsupportedError].
+  /// > Attempting to create an [ImageFilter.shader] when this property is `false` will throw an [UnsupportedError].
   static bool get isShaderFilterSupported => _impellerEnabled;
 
   // Converts this to a native DlImageFilter. See the comments of this method in
--- a/engine/src/flutter/lib/ui/painting.dart
+++ b/engine/src/flutter/lib/ui/painting.dart
@@ -4338,10 +4338,10 @@ abstract class ImageFilter {
   /// > This API is only supported when using the Impeller rendering engine.
   /// > On other backends, an [UnsupportedError] will be thrown.
   ///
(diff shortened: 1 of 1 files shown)


engine/src/flutter/lib/ui/painting.dart
```diff
@@ -4338,6 +4348,10 @@ abstract class ImageFilter {
   }
 
   /// Whether [ImageFilter.shader] is supported on the current backend.
+  ///
+  /// > [!WARNING]
+  /// > This property indicates whether the Impeller rendering engine is enabled.
```
chinmaygarde: This documents an implementation detail of isShaderFilterSupported. How about just saying, "Attempting to create an ImageFilter.shader when isShaderFilterSupported will throw and exception."

gowsikraja: Updated the isShaderFilterSupported doc to be implementation‑agnostic: it now states that attempting to create ImageFilter.shader when the property is false will throw an UnsupportedError — PTAL.

chinmaygarde: You can do the same thing for the warning. For instance, "This property will only return true when the Impeller rendering engine is enabled."

gowsikraja: Makes sense — updated the getter to be implementation‑agnostic and changed the warning too. PTAL.

chinmaygarde (approved):

jason-simmons (approved):

chinmaygarde:
LGTM but the analysis is not able to complete successfully.

PR #179415, opened 2025-12-03
Made wide gamut tests give more information in their failure

gaaclarke:
fixes https://github.com/flutter/flutter/issues/179396

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
+++ b/dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
@@ -145,36 +145,14 @@ List<double> _deepRed = <double>[1.0931, -0.2268, -0.1501];
   return (foundColor, closestColor);
 }
 
-(bool, List<double>) _findColor(
-  List<dynamic> result,
-  List<double> color, {
-  double epsilon = 0.01,
-}) {
+(bool, List<double>) _findColor(List<dynamic> result, List<double> color, {double epsilon = 0.01}) {
   expect(result, isNotNull);
   expect(result.length, 4);
   final [int width, int height, String format, Uint8List bytes] = result;
   return switch (format) {
-    'MTLPixelFormatBGR10_XR' => _findBGR10Color(
-      bytes,
-      width,
-      height,
-      color,
-      epsilon: epsilon,
-    ),
-    'MTLPixelFormatBGRA10_XR' => _findBGRA10Color(
-      bytes,
-      width,
-      height,
-      color,
-      epsilon: epsilon,
-    ),
-    'MTLPixelFormatRGBA16Float' => _findRGBAF16Color(
-      bytes,
-      width,
-      height,
-      color,
-      epsilon: epsilon,
-    ),
+    'MTLPixelFormatBGR10_XR' => _findBGR10Color(bytes, width, height, color, epsilon: epsilon),
+    'MTLPixelFormatBGRA10_XR' => _findBGRA10Color(bytes, width, height, color, epsilon: epsilon),
+    'MTLPixelFormatRGBA16Float' => _findRGBAF16Color(bytes, width, height, color, epsilon: epsilon),
     _ => fail('Unsupported pixel format: $format'),
   };
 }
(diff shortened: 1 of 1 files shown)


dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -54,6 +60,8 @@ bool _findRGBAF16Color(
   expect(bytes.lengthInBytes, width * height * 8);
   expect(bytes.lengthInBytes, byteData.lengthInBytes);
   var foundDeepRed = false;
```
b-luk: `foundDeepRed` should probably be named `foundColor` in all these functions.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -64,11 +72,16 @@ bool _findRGBAF16Color(
         _isAlmost(blue, color[2], epsilon)) {
       foundDeepRed = true;
```
b-luk: The "found" check is based on an epsilon for each rgb component, but the "closest" check is based on vector distance (squared) of the colors. I think it's weird that we're using a different color comparison check for finding the direct match vs finding the closest color.

To me, the vector distance seems like a more appropriate method of comparing colors than than the individual rgb component checks. Should we consider changing the "found" check to be `distance < epsilon` instead of comparing each individual component against epsilon? This way both the "found" and "closest" checks are based on the same method of color comparison. A side benefit is that the code would be simplified considerably too.

Making this change would be changing the behavior of the current "found" check. But I wouldn't expect it to affect any results assuming the epsilon is reasonable.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
gaaclarke: done

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
gaaclarke: Yea, that's a fair observation.  The reason the epsilon is applied per component is that it's just easier to reason about.  Whereas what does it mean for one color to be 0.01 euclidean distance from another?

I'd rather keep it this way for now.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -79,6 +92,8 @@ bool _findBGRA10Color(
   expect(bytes.lengthInBytes, width * height * 8);
   expect(bytes.lengthInBytes, byteData.lengthInBytes);
   var foundDeepRed = false;
```
b-luk: This function looks like it got updated incorrectly. Make sure to update it to match the others.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
b-luk: A color being 0.01 distance from another makes an equal amount of sense to me as a color's r/g/b value being 0.01 distance from another color's.

But I don't feel super strongly about this, so if you prefer keeping it as-is I'm fine with it.

b-luk (approved):

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
gaaclarke: thanks!

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
b-luk: Thinking more about what it actually means, a 0.01 euclidean distance means that in a 3D color space, we're seeing if the color we're checking is contained within a sphere of diameter 0.02 that is centered on our target color.

And with the epsilon check for each individual rgb component, we're seeing if the color we're checking is contained within a cube with side length=0.02 that is centered on our target color.

gaaclarke: Color comparison is actually more complicated than that and it's a deep rabbit hole to get into.  Check this out: https://en.wikipedia.org/wiki/CIE_1931_color_space

The Euclidean distance falls apart because that's not how human perception works.  The deltas in the 3d space aren't uniformly perceived, that's why it's hard to chose an epsilon for distances.

PR #179426, opened 2025-12-03
Use depth buffer to implement geometry overdraw protection

### LINKED ISSUE #179107: [Impeller] Geometry overdraw protection can be handled by the depth compare function
(Replaces https://github.com/flutter/flutter/issues/178951)

Currently if a geometry creates a set of triangles that might overlap then it can request that we use a stencil-then-cover operation to resolve their overlapping areas before we apply the rendering operation. This adds time by requiring multiple passes to complete the operation. The stencil-then-cover processing is set up in `clip_contents.cc`: https://github.com/flutter/flutter/blob/9d96df2364317334b55547ccb47c81aa9418eb71/engine/src/flutter/impeller/entity/contents/color_source_contents.h#L219

Another way to prevent overdraw would be to use the fact that we already test against the depth buffer. This test is meant to clip primitives by only allowing drawing at Z values that are greater or equal (`kGreaterEqual`) to the current value in the depth buffer. This is used to perform complex clipping by rendering an inversion of a clip at a depth that is higher than the primitives it will clip - thus causing them to fail their `Z >= depth_buffer_sample` test. This test is set up in `contents.cc`: https://github.com/flutter/flutter/blob/9d96df2364317334b55547ccb47c81aa9418eb71/engine/src/flutter/impeller/entity/contents/contents.cc#L29

Using the`kGreaterEqual` prevents drawing where the clip has already drawn higher (in Z) values, but does not prevent drawing over previous rendering operations that were drawn at lower (in Z) values. It also does not prevent drawing over parts of the current rendering operation when its geometry (triangular tessellation) has multiple triangles that contain the same pixels.

But, the depth comparison function `kGreater` would also enforce clipping for the same reason and allow drawing over prior rendering operations, again for the same reason, but in this case, the current rendering primitive could only visit the pixel once and then its own Z value put into the depth buffer would prevent the same primitive from painting that pixel a second time.

Since simply changing the depth compare function doesn't require a second pass, all cases where the Geometry objects request overdraw protection could provide that service in the normal pixel processing by simply using `kGreater`.

b-luk:
Use depth buffer to implement geometry overdraw protection as described in #179107.

Also deletes code that was used only for the old stencil-based overdraw protection.

Fixes #179107

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/impeller/entity/contents/clip_contents.cc
```diff
@@ -102,9 +102,6 @@ bool ClipContents::Render(const ContentContext& renderer,
       break;
     case GeometryResult::Mode::kNormal:
     case GeometryResult::Mode::kPreventOverdraw:
-      pass.SetCommandLabel("Clip stencil preparation (Increment)");
-      options.stencil_mode =
-          ContentContextOptions::StencilMode::kOverdrawPreventionIncrement;
       break;
```
b-luk, line 124: I'm a bit confused about this section of code.

Do I need anything here to replace the old code?

`GeometryResult::Mode::kNormal` used to fall through to the `GeometryResult::Mode::kPreventOverdraw` case and use the `kOverdrawPreventionIncrement` stencil mode. Was that intentional? What's the purpose of this?

flar: I'm not 100% sure on this. It may have been that kNormal just didn't care about these particular settings because, for instance, it didn't actually run the stencil pass anyway.
@gaaclarke @chinmaygarde ?

engine/src/flutter/impeller/entity/contents/color_source_contents.h
```diff
@@ -241,19 +239,7 @@ class ColorSourceContents : public Contents {
 
     pass.SetPipeline(pipeline_callback(options));
 
-    if (!pass.Draw().ok()) {
-      return false;
-    }
-
-    // If we performed overdraw prevention, a subsection of the clip heightmap
```
flar, line 248: This puzzles me. Why would we have needed to restore the clip depth in the first place? Any clip that is still in effect will have a higher depth value and we wouldn't have overwritten it by our actions. It should be OK to leave stale depth values in the buffer as each rendering op is processed at a slightly higher depth value. Also, didn't the previous solution use the stencil buffer and not the depth buffer?

flar (commented):
I would say LGTM, but this is mostly as far as I understand how things work. I'd like to see reviews from @gaaclarke or @chinmaygarde as well.

engine/src/flutter/impeller/entity/contents/color_source_contents.h
```diff
@@ -216,15 +216,13 @@ class ColorSourceContents : public Contents {
...
-    // If overdraw prevention is enabled (like when drawing stroke paths), we
-    // increment the stencil buffer as we draw, preventing overlapping fragments
-    // from drawing. Afterwards, we need to append another draw call to clean up
-    // the stencil buffer (happens below in this method). This can be skipped
-    // for draws that are fully opaque or use src blend mode.
+    // If overdraw prevention is requested (like when drawing stroke paths), set
+    // up the depth buffer and depth comparison function to prevent the same
+    // pixel from being painted multiple times.
     if (geometry_result.mode == GeometryResult::Mode::kPreventOverdraw &&
         options.blend_mode != BlendMode::kSrc) {
```
ColdPaleLight: Since the new rendering approach uses depth testing to eliminate overdraw, we no longer need to check whether the `BlendMode` is` kSrc`. In the past, this check helped skip an extra stencil pass when drawing opaque stroke paths, but overdraw still occurred (Some drivers may optimize this away, but such behavior is not guaranteed). With depth testing, overdraw for opaque stroke paths is avoided entirely.

flar: Actually, it can potentially make things a little faster for blend mode since it avoids storing a new pixel value on a previously rendered pixel.

If we trust this approach well enough then we can eventually just make this all unconditional.

b-luk: So is the suggestion to remove the `blend_mode != kSrc` check here?

In line 213 above, when `BlendMode` is `kSrc`, then `depth_write_enabled` is set to `true` (and `depth_compare ` remains its default `kGreaterEqual`). The comment says this is "to allow reordering". But I don't fully understand what this means or what this does.

Keeping the `blend_mode != kSrc` check here preserves this existing behavior when `BlendMode` is `kSrc`. I didn't want to change the existing behavior because I didn't know whether changing this behavior would do something unintended.

So current behavior with this PR is:
|   | BlendMode != kSrc | BlendMode == kSrc| 
| ----- | ----- |----- |
| **kPreventOverdraw false** | use default options | depth_write_enabled=true |
| **kPreventOverdraw true** | depth_write_enabled=true depth_compare=kGreater | depth_write_enabled=true |


If the `options.blend_mode != BlendMode::kSrc` here is removed, the behavior would be:
|   | BlendMode != kSrc | BlendMode == kSrc| 
| ----- | ----- |----- |
| **kPreventOverdraw false** | use default options | depth_write_enabled=true |
| **kPreventOverdraw true** | depth_write_enabled=true depth_compare=kGreater | depth_write_enabled=true depth_compare=kGreater |

Is this change in the behavior for the `(BlendMode == kSrc) && (kPreventOverdraw == true)` case acceptable and/or desired?

Or is the suggestion to both remove the `blend_mode != kSrc` check here, and also remove setting `depth_write_enabled` when `BlendMode` is `kSrc`? The behavior would be:

|   | BlendMode != kSrc | BlendMode == kSrc| 
| ----- | ----- |----- |
| **kPreventOverdraw false** | use default options | use default options |
| **kPreventOverdraw true** | depth_write_enabled=true depth_compare=kGreater | depth_write_enabled=true depth_compare=kGreater |

engine/src/flutter/impeller/entity/contents/clip_contents.cc
chinmaygarde: If memory serves, `GeometryResult::Mode::kNormal` means there are no overlapping triangles. Now I don't think you can entirely disregard whats on the stencil buffer and ignore the stencil test directly. Since normal is the simplest case of the lot, I suspect it was lumped together with PreventOverdraw. But disregarding the test entirely does seem off though.

chinmaygarde: Now that you are no longer using the stencil buffer, lumping it together with either the EO or NZ modes should be fine I think.

b-luk: I think you're right. I moved the kNormal case to be lumped with kNonZero, and I've verified locally that the [broken iOS app scenario tests](https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/2070592/overview) are now passing with with change applied. I updated the PR to do this.

Should I do anything with the kPreventOverdraw case here? Is that case never expected to happen for a ClipContents? That's my initial thought, but my understanding how all this code works is very incomplete. If that case is unexpected and incompatible with ClipContents, I can add a FML_UNREACHABLE() here. Otherwise, should I also lump kPreventOverdraw in with one of the other cases?

b-luk:
> Hmm, did you already get access to the goldens results? I couldn't find the link. In any case, I think the stencil mode setter might be off for the reason I mentioned in the inline comment.

The golden results are [here on FRoB](https://frob.corp.google.com/#/pr/179426). The TAP results for the original revision of this PR are [here](https://fusion2.corp.google.com/presubmit/839933947/OCL:839933947:BASE:839972608:1764810026632:9c62b84d/targets). The old TAP results show some failures where it looks like certain UI got clipped out or aren't rendering. Maybe the most recent [fix](https://github.com/flutter/flutter/pull/179426/commits/775e23d4204cf68063097fd5dfc9cf6cc21d850e) that you suggested will fix it in the new in progress TAP run.

b-luk:
I tracked down the golden test failures. These tests draw a stroked rounded rectangle with a MaskFilter which blurs parts of the rectangle. The blurring process is outlined here:

https://github.com/flutter/flutter/blob/775e23d4204cf68063097fd5dfc9cf6cc21d850e/engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc#L703-L711

The test was failing specifically when using a [BlurStyle](https://api.flutter.dev/flutter/dart-ui/BlurStyle.html) of `outer` or `inner`. These are the two blur styles where step 5 above will apply a clip mask to the output: They call `ApplyClippedBlurStyle` [here](https://github.com/flutter/flutter/blob/775e23d4204cf68063097fd5dfc9cf6cc21d850e/engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc#L584-L592).

`ApplyClippedBlurStyle` creates a `ClipContents` [here](https://github.com/flutter/flutter/blob/775e23d4204cf68063097fd5dfc9cf6cc21d850e/engine/src/flutter/impeller/entity/contents/filters/gaussian_blur_filter_contents.cc#L542-L552). The `Geometry` used is the geometry of the stroked rounded rectangle, which is a `StrokeSegmentsGeometry`. The [`GetPositionBuffer` call](https://github.com/flutter/flutter/blob/775e23d4204cf68063097fd5dfc9cf6cc21d850e/engine/src/flutter/impeller/entity/geometry/stroke_path_geometry.cc#L740) for this geometry returns a `GeometryResult` with `GeometryResult::Mode::kPreventOverdraw`.

So for this case, use a `ClipContents` with a `kPreventOverdraw` mode.

The problem with this is the same as the issue Chinmay pointed out above, where I originally changed the `kNormal` and `kPreventOverdraw` cases in `ClipContents::render` to not set any stencil_mode. Chinmay pointed out that this was likely incorrect for `kNormal`, and I fixed it for that case. The scenario for these tests also demonstrate that `ClipContents` with a `kPreventOverdraw` is a valid case, and it is fixed in the same way. With this change, the clip mask used for an inner/outer blur of a stroked path geometry works properly again.

engine/src/flutter/impeller/entity/contents/color_source_contents.h
ColdPaleLight: >  So is the suggestion to remove the blend_mode != kSrc check here?

Yes, it is.

> In line 213 above, when BlendMode is kSrc, then depth_write_enabled is set to true (and depth_compare  remains its default kGreaterEqual). The comment says this is "to allow reordering". But I don't fully understand what this means or what this does.

Impeller automatically [rewrites the BlendMode from kSrcOver to kSrc](https://github.com/flutter/flutter/blob/d50dfb9bf7766a48420818a9585669547ecaa551/engine/src/flutter/impeller/display_list/canvas.cc#L1833-L1836) when it detects that the content being drawn is fully opaque. In kSrc mode, the Render backend [turns blending off entirely](https://github.com/flutter/flutter/blob/d50dfb9bf7766a48420818a9585669547ecaa551/engine/src/flutter/impeller/entity/contents/content_context.cc#L347).

Given that, imagine we’re drawing three opaque shapes A → B → C (from back to front). Because blending is disabled for these draws, their visual result no longer depends on the draw order. **Once depth writes are enabled**, we can safely reorder them—for example, drawing C → B → A instead of the original order.

The benefit of this reordering is that it allows the GPU to reject more fragments early. Drawing the frontmost opaque geometry first lets early-z eliminate a large portion of the fragments from the shapes behind it, which reduces overdraw and lowers fragment-shader cost overall.

> If the options.blend_mode != BlendMode::kSrc here is removed, the behavior would be:

|                              | BlendMode != kSrc            | BlendMode == kSrc            |
|------------------------------|------------------------------|------------------------------|
| **kPreventOverdraw = false** | use default options          | depth_write_enabled = true   |
| **kPreventOverdraw = true**  | depth_write_enabled = true; depth_compare = kGreater | depth_write_enabled = true; depth_compare = kGreater |
> Is this change in the behavior for the (BlendMode == kSrc) && (kPreventOverdraw == true) case acceptable and/or desired?

Yes, this change is acceptable — in fact, it’s exactly the behavior we want.

b-luk: Thanks for the thorough explanation. I removed the `blend_mode != kSrc` check as you suggested.

engine/src/flutter/impeller/entity/contents/clip_contents.cc
b-luk: Ok, after studying this a bit, I think I have a much better understanding of how things work for ClipContents.

With the existing behavior, ClipContents always performs a clip by using stencil-and-cover. It does a render pass for the preparation stencil, which sets the stencil buffer according to the geometry. And then it does a second cover render pass that writes the depth (performing the actual clip) and resets the stencil buffer. It uses this stencil even for geometries kNormal and kPreventOverdraw. For those geometries, it uses a stencil mode that always fills in a 1 value regardless of any triangle overlap in the geometry.

My first PR revision removed the stencil step for kNormal and kPreventOverdraw. This was incorrect because it would render an empty stencil for the stencil step. So the cover step would clip everything. This caused the test breakages in my first revision, where the goldens showed a lot of content was inadvertently clipped out completely.

In a subsequent revision, I followed Chinmay's advice to make kNormal and kPreventOverdraw fall back to the kNonZero stencil case. The stencil is now properly filled out and these cases now work pretty close to how they did before.

I realize that we can do a little better specifically for difference clips.

Clipping needs to use stencil-and-cover for nonzero or evenodd geometry cases because those cases have overlapping triangles, and the stencil pass is needed to determine whether each of these overlaps should be included in the depth write.

But for kNormal and kPreventOverdraw, we don't need stencil-and-cover. The kNormal case has no overlapping triangles, so the depth write has no overlapping triangles to deal with. For the kPreventOverdraw case, any overlapping triangles should unconditionally be included in the depth write (unlike the nonzero or evenodd cases). So kNormal and kPreventOverdraw geometries can do a difference clip with a single render pass and does not need to use a stencil.

I just updated my PR to do this.

Edit: It turns out there are slight differences between the kStencilNonZeroFill stencil mode and what we want for kNormal/kPreventOverdraw geometries, which show up as a few small pixel differences in the golden tests. This means we can't fall through to kStencilNonZeroFill for kNormal/kPreventOverdraw. Instead, I made a new kStencilIncrementAll stencil mode which does what we want: set the stencil for everything covered by a triangle regardless of any triangle overlaps. This is the same as what the old kOverdrawPreventionIncrement stencil mode did.

flar (approved):
LGTM.

I'm a little puzzled by the golden changes, I don't know why those colors would be tweaked by "1" like that, but it looks fine.

Don't forget to approve the goldens before autosubmitting...

chinmaygarde (approved):

PR #179442, opened 2025-12-04
Make sure that a CupertinoScrollbar doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the CupertinoScrollbar widget.


packages/flutter/test/cupertino/scrollbar_test.dart
```diff
@@ -1347,4 +1347,14 @@ void main() {
...
+
+  testWidgets('CupertinoScrollbar does not crash at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      const CupertinoApp(
+        home: Center(
+          child: SizedBox.shrink(child: CupertinoScrollbar(child: CustomScrollView())),
+        ),
+      ),
+    );
+  });
```
victorsanni, line 1360: No expects?

ahmedsameha1: I am sorry.

victorsanni (approved):

dkwingsmt (approved):
Checklist:

PR #179467, opened 2025-12-04
Bump engine.version.

eyebrowsoffire:
We need to bump the engine version once more so that we're pointing at artifacts that were actually built for release (and therefore are signed.


jtmcdole (approved):
```
██╗      ██████╗████████╗███╗   ███╗
██║     ██╔════╝╚══██╔══╝████╗ ████║
██║     ██║  ███╗  ██║   ██╔████╔██║
██║     ██║   ██║  ██║   ██║╚██╔╝██║
███████╗╚██████╔╝  ██║   ██║ ╚═╝ ██║
╚══════╝ ╚═════╝   ╚═╝   ╚═╝     ╚═╝
```

Issue #179450, opened 2025-12-04
Semantics focus does not move outside viewport when moving up or down


zemanux:
### Steps to reproduce

1. Enable TalkBack on an Android device.
2. Move accessibility focus to the first/last visible ListTile.
3. Perform the `Previous item` / `Next item` gesture (Swipe Left/Right).

### Expected results

The focus should move to the previous / next `ListTile`, scrolling the ListView to make it visible.

### Actual results

Semantic focus move to the `SliverAppBar` / `TalkBack shortcut` instead.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() => runApp(const MaterialApp(home: MyApp()));

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: CustomScrollView(
        slivers: <Widget>[
          const SliverAppBar(
            pinned: true,
            expandedHeight: 100.0,
            title: Text('Pinned Semantic Focus'),
            backgroundColor: Colors.blue,
          ),
          // The List
          SliverList(
            delegate: SliverChildBuilderDelegate((BuildContext context, int index) {
              return ListTile(title: Text('Item $index'), subtitle: const Text('Scroll down then navigate back up'));
            }, childCount: 50),
          ),
        ],
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[[Upload media here]](https://github.com/user-attachments/assets/85a0ff56-780e-4d45-97df-d6f44ae75643)

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel master, 3.39.0-1.0.pre-408, on macOS 15.7.2 24G325 darwin-arm64, locale fr-FR) [3,2s]
    • Flutter version 3.39.0-1.0.pre-408 on channel master at /Users/emmanuel_lefebvre/fvm/versions/master
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision b3770ec6f4 (4 hours ago), 2025-12-04 05:15:39 -0500
    • Engine revision b3770ec6f4
    • Dart version 3.11.0 (build 3.11.0-200.1.beta)
    • DevTools version 2.53.0
    • Pub download mirror https://artifact.socrate.vsct.fr/artifactory/api/pub/all-pub
[...]

PR #179480, opened 2025-12-05
Fix - Semantics focus does not move outside viewport when moving up or down

jwlilly:
Fixes regression caused by PR [#174374](https://github.com/flutter/flutter/pull/174374). SliverLists are not scrolling with TalkBack. Update fixes the issue by adding the ScrollView or HorizontalScrollView if the scroll actions are present. 

Test code used: 

``` dart

void main() => runApp(const MaterialApp(home: MyApp()));

  const MyApp({super.key});

  @override
    return Scaffold(
      body: CustomScrollView(
        slivers: <Widget>[
          const SliverAppBar(
            pinned: true,
            expandedHeight: 100.0,
            title: Text('Pinned Semantic Focus'),
            backgroundColor: Colors.blue,
          ),
          // The List
          SliverList(
            delegate: SliverChildBuilderDelegate((BuildContext context, int index) {
              return ListTile(title: Text('Item $index'), subtitle: const Text('Scroll down then navigate back up'));
            }, childCount: 50),
          ),
        ],
      ),
    );
  }
}

```

https://github.com/user-attachments/assets/a3780f6d-8e22-40b7-83eb-38bc406b8821

- Fixing #179450

## Pre-launch Checklist

- [x] I signed the [CLA].


reidbaker (approved):

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -925,6 +925,14 @@ public AccessibilityNodeInfo createAccessibilityNodeInfo(int virtualViewId) {
       // the visible viewport of a scrollable, unless the node itself does not
       // allow implicit scrolling - then we leave the className as view.View.
       result.setScrollable(true);
+      if (semanticsNode.hasFlag(Flag.HAS_IMPLICIT_SCROLLING)) {
```
chunhtai, line 928: why we no longer check for shouldSetCollectionInfo ? looking at the code it looks like it check for at least 2 children and the ancestor has focus before setting the class. They make sense to me. How does this change fix the issue?

jwlilly: shouldSetCollectionInfo checks on line 947. But whether a view has collection info or not shouldn't impact whether it scrollable. Collection Info should only impact whether it should be a semantic list or not. 
I adjusted the code to add the ScrollView or HorizontalScrollView className to the nodeInfo based on the on if it has the scroll accessibility actions. With a Sliver list, it wasn't scroll with TalkBack because the ScrollView className wasn't being applied

chunhtai (approved):
LGTM

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
chunhtai: thanks for response, though I still not sure why sliver list didn't have the scrollable class before given it should have passed the shouldSetCollectionInfo check, but looking at the code again, the current code structure makes more sense after the change. so I think this is fine.

jwlilly: That was my fault with the changes for collectionItemInfo. I changed the logic of shouldSetCollectionInfo to target only ListViews. I'm guessing Sliver lists are slightly different so the scrollable class wasn't being applied because shouldSetCollectionInfo doesn't pass for Sliver lists

PR #179484, opened 2025-12-05
Implement flutter/accessibility channel

### LINKED ISSUE #172023: Nonfunctioning semantic announcements on Ubuntu desktop
### Steps to reproduce

Flutter on Ubuntu Linux desktop doesn't seem to support semantic announcements or live regions. `SemanticsService.isAnnounceSupported()` outputs `true`.

1. Create the default new Flutter application via VSCode
2. Add a `SemanticsService.announce` to the `_incrementCounter` function
3. Enable the screen reader on Ubuntu desktop
4. Run the app
5. Press the + button to increment the counter
6. Observe the screen reader not announcing anything

In my example below, I've also added a live region and the output of `SemanticsService.isAnnounceSupported()`.

I have no special settings for the screen reader or Ubuntu, everything should be near defaults.

### Expected results

The screen reader should read aloud when a live region changes content, or when a semantics announcement is manually triggered.

### Actual results

No announcements are read by the screen reader.

### Code sample

Slightly modified version of the default new application via VSCode:

```dart
import 'package:flutter/semantics.dart';

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
      ),
    );
  }
}

  final String title;

  @override
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });

    SemanticsService.announce(
      'Incrementing counter',
      TextDirection.ltr,
    ); // doesn't actually say anything
  }

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          children: <Widget>[
            const Text('You have pushed the button this many times:'),
            Semantics(
              liveRegion:
                  true, // doesn't actually result in any expected behavior
              child: Text(
                '$_counter',
              ),
            ),
            Text(
              'Supports announce: ${SemanticsService.isAnnounceSupported()}', // outputs true
            ),
          ],
        ),
      ),
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ),
    );
  }
}
```

</details>

### Logs

[Logs (txt file)](https://github.com/user-attachments/files/21188715/logs.txt)

### Flutter Doctor output

```console
$ flutter doctor -v
[✓] Flutter (Channel stable, 3.32.5, on Ubuntu 24.04.2 LTS 6.11.0-29-generic, locale en_US.UTF-8) [27ms]
    • Flutter version 3.32.5 on channel stable at /home/[USERNAME]/fvm/versions/3.32.5
    • Framework revision fcf2c11572 (2 weeks ago), 2025-06-24 11:44:07 -0700
    • Engine revision dd93de6fb1
    • Dart version 3.8.1
    • DevTools version 2.45.1

[✗] Android toolchain - develop for Android devices [33ms]
    ✗ Unable to locate Android SDK.
      (or visit https://flutter.dev/to/linux-android-setup for detailed instructions).


[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome) [12ms]

[✓] Linux toolchain - develop for Linux desktop [117ms]
    • Ubuntu clang version 18.1.3 (1ubuntu1)
    • cmake version 3.28.3
    • ninja version 1.11.1
    • pkg-config version 1.8.1
    ! Unable to access driver information using 'eglinfo'.
      It is likely available from your distribution (e.g.: apt install mesa-utils)

[!] Android Studio (not installed) [10ms]
      (or visit https://flutter.dev/to/linux-android-setup for detailed instructions).

[✓] VS Code (version unknown) [10ms]
    • VS Code at /snap/code/current/usr/share/code
    • Flutter extension can be installed from:
      🔨 https://marketplace.visualstudio.com/items?itemName=Dart-Code.flutter
    ✗ Unable to determine VS Code version.

[✓] Connected device (1 available) [37ms]
    • Linux (desktop) • linux • linux-x64 • Ubuntu 24.04.2 LTS 6.11.0-29-generic

[✓] Network resources [178ms]

! Doctor found issues in 3 categories.
```

</details>

robert-ancell:
Fixes https://github.com/flutter/flutter/issues/172023


engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
```diff
@@ -0,0 +1,167 @@
...
+  if (strcmp(type, kAnnounceType) == 0) {
+    process_announce(self, data);
+  } else if (strcmp(type, kTooltipType) == 0) {
+  } else if (strcmp(type, kLongPressType) == 0) {
+  } else if (strcmp(type, kTapType) == 0) {
+    // Only used by Android
+  } else if (strcmp(type, kFocusType) == 0) {
+    // Only used by Android and iOS.
+  } else {
+    // Silently ignore unknown types.
```
mattkae, line 142: Should we warn here? It seems that we're explicitly silencing other types. Would we want to silence "new" unknown types too?

engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  FlValue* text_direction_value =
+      fl_value_lookup_string(data, kTextDirectionKey);
+  if (text_direction_value == nullptr ||
+      fl_value_get_type(text_direction_value) != FL_VALUE_TYPE_INT) {
+    g_warning("Missing/invalid text direction in accessibility announce event");
+    return;
+  }
+  FlTextDirection text_direction =
+      static_cast<FlTextDirection>(fl_value_get_int(text_direction_value));
```
mattkae: Nit: Should we check if this fits into our enum too? I can imagine a scenario where new values are added and suddenly we are failing

engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
```diff
@@ -0,0 +1,167 @@
...
+      fl_value_lookup_string(data, kAssertivenessKey);
+  FlAssertiveness assertiveness = FL_ASSERTIVENESS_POLITE;
+  if (assertiveness_value != nullptr) {
+    if (fl_value_get_type(assertiveness_value) != FL_VALUE_TYPE_INT) {
+      g_warning("Invalid assertiveness in accessibility announce event");
+      return;
+    }
+    assertiveness =
+        static_cast<FlAssertiveness>(fl_value_get_int(assertiveness_value));
+  }
```
mattkae: Sam as above. Might be a good idea to check that it is within the confines of the expected values

engine/src/flutter/shell/platform/linux/fl_accessibility_handler_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  fl_mock_binary_messenger_send_standard_message(
+      messenger, "flutter/accessibility", message,
+      [](FlMockBinaryMessenger* messenger, FlValue* response,
+         gpointer user_data) {
+        EXPECT_EQ(fl_value_get_type(response), FL_VALUE_TYPE_NULL);
+        gboolean* called = static_cast<gboolean*>(user_data);
+        *called = TRUE;
+      },
+      &called);
+  EXPECT_TRUE(called);
```
mattkae, line 197: Should you also be asserting `EXPECT_FALSE(signalled)`?

engine/src/flutter/shell/platform/linux/fl_accessibility_handler_test.cc
```diff
@@ -0,0 +1,171 @@
...
```
mattkae, line 229: Should you also be asserting `EXPECT_FALSE(signalled)`?

mattkae (changes requested):
A few minor things, but looks solid :rock:

engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
robert-ancell: I think it would be a bad idea to log these here - we really have no idea what future a11y types might come and how frequent they might occur. This channel is badly designed - ideally these would have been method calls and we'd send a response if they've been handled. Then the Dart side could decide if it was worth logging. I discussed some of these limitations with other Flutter developers and they said this channel really should be replaced by a C interface in the embedder.

engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
robert-ancell: (falls down rabbit hole of C/C++ enum documentation)

I was originally checking the values but I decided that it was safer to accept everything as we don't want to fail the event if a new value turns up (these are essentially hints and not used in the Linux embedder anyway). I think perhaps it will be safer to generate a warning for unknown types and default to a value.

engine/src/flutter/shell/platform/linux/fl_accessibility_channel.cc
robert-ancell: See above.

mattkae (approved):
Nice 😄

robert-ancell:
The version of Debian is bullseye (oldoldstable), but is trixie (stable) for riscv builds. Ideally we'd update all the images to trixie. https://chromium.googlesource.com/chromium/src/+/refs/heads/main/build/linux/sysroot_scripts/sysroot_creator.py#24

robert-ancell:
I don't think we can realistically update to trixie until Chromium does. The tarballs have to be built and uploaded to the chromium servers and it probably doesn't make sense to support this for Flutter before Chromium. I guess we'll have to find a workaround for now :(

PR #179512, opened 2025-12-05
Add FlutterFramework as an Xcode SwiftPM local package override

### LINKED ISSUE #172427: [packages] Design analysis solution for SwiftPM packages
See https://github.com/flutter/flutter/issues/171442 for background. tl;dr is that the `xcode-analyze` command in flutter/packages repo tooling has traditionally analyzed both the native test code in the Runner project, and also the code of CocoaPods packages the project imported—importantly, the plugin code itself. However, the observed behavior is that if code in included in a project via SwiftPM, that code is *not* analyzed as part of the project, so the `xcode-analyze` command as currently designed doesn't analyze the plugin code, which is the main point of the command.

We need a solution to analyze the plugin code when using SwiftPM, since CocoaPods will stop working at some point. In theory that should be as simple as finding a command that allows running Xcode analysis of a Swift Package directly from the command line, but we'll presumably run into the usual problem that causes us to use example apps to interact with plugin packages, which is that the package is not necessarily well-formed on its own. In particular, Flutter imports are presumably broken. @vashworth this may inform some of the planning around the alternate SwiftPM integration model you've been looking at?

vashworth:
This PR adds the FlutterFramework as an Xcode local package override by adding it as a "file" to the Xcode project. In addition, it also adds the plugin as a local package override if we detect the app is a plugin's example app.

Screenshot 2025-12-04 at 10 29 16 PM

Fixes https://github.com/flutter/flutter/issues/179032. Fixes https://github.com/flutter/flutter/issues/172427.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/lib/src/migrations/swift_package_manager_integration_migration.dart
```diff
@@ -1179,3 +1296,47 @@ class ParsedProject {
   final String identifier;
   final List<String>? packageReferences;
 }
+
+class _PluginRelativeToExample {
```
stuartmorgan-g: Optional: you could just used a record with named fields, instead of defining a type just to return two things.

packages/flutter_tools/lib/src/migrations/swift_package_manager_integration_migration.dart
```diff
@@ -1179,3 +1296,47 @@ class ParsedProject {
...
+  ///
+  /// If the [xcodeProject] is not within an example app or the plugin can't be found, return null.
+  static _PluginRelativeToExample? loadFromProject({
+    required XcodeBasedProject xcodeProject,
+    required FileSystem fileSystem,
+    required Logger logger,
+  }) {
+    try {
+      final FlutterProject flutterProject = xcodeProject.parent;
+      if (flutterProject.directory.path.endsWith('example') &&
```
stuartmorgan-g: This won't catch everything; it wouldn't work for [the current `google_maps_flutter_ios` examples, for instance](https://github.com/flutter/packages/tree/eb9e1dc760d49016a515575424b97bfd79718bd5/packages/google_maps_flutter/google_maps_flutter_ios/example). I would expect that's rare enough that it's not a significant issue, but we should consider whether we can document somewhere—maybe even just as a comment in the issue this fixes—how unusual examples could be fixed manually.

We could try to make this more comprehensive by walking up several levels looking for an `example` directory that is in a directory with a pubspec), but I think it's probably not worth the added complexity.

packages/flutter_tools/templates/app/ios.tmpl/Runner.xcodeproj/project.pbxproj.tmpl
```diff
@@ -54,6 +54,10 @@
 		7884E8672EC3CC0400C636F2 /* SceneDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SceneDelegate.swift; sourceTree = "<group>"; };
 		{{#withSwiftPackageManager}}
 		78E0A7A72DC9AD7400C4905E /* FlutterGeneratedPluginSwiftPackage */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = FlutterGeneratedPluginSwiftPackage; path = Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage; sourceTree = "<group>"; };
+		784666492D4C4C64000A1A5F /* FlutterFramework */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = FlutterFramework; path = Flutter/ephemeral/Packages/.packages/FlutterFramework; sourceTree = "<group>"; };
+		{{#withPlatformChannelPluginHook}}
```
stuartmorgan-g, line 58: Oops, I forgot this is the condition name we used for plugins with a plugin-style build. We'll need to rename it at some point (not directly related to this PR, mostly noting as a reminder to myself).

stuartmorgan-g (approved):
LGTM

packages/flutter_tools/lib/src/migrations/swift_package_manager_integration_migration.dart
vashworth: Right, this makes like a best effort. I plan to also add website docs: https://github.com/flutter/website/issues/12936

packages/flutter_tools/lib/src/migrations/swift_package_manager_integration_migration.dart
vashworth: Oooh cool, I didn't know you could do named fields. Done!

stuartmorgan-g: Diff LG

PR #179519, opened 2025-12-05
Implements decodeImageFromPixelsSync

### LINKED ISSUE #178488: Add `Image.decodeImageFromPixelsSync`
`Image.decodeImageFromPixels` exists already, this is a proposal to allow people to do it synchronously.  This matches the behavior of `Picture.toImageSync` but instead of using Picture, people can use raw pixels so they can generate high bitrate textures like r32float.

Use cases:
- generate an SDF and attach it to a fragment shader all in the same frame

gaaclarke:
fixes https://github.com/flutter/flutter/issues/178488

This doesn't implement the following.  They can be implemented in later PRs.
- a skia implementation (maybe won't implement?)
- a web implementation
- resizing
- target pixel format

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/ui/painting/image.cc
+++ b/engine/src/flutter/lib/ui/painting/image.cc
@@ -4,18 +4,16 @@
 
 #include "flutter/lib/ui/painting/image.h"
 
-#include <algorithm>
-#include <limits>
 #include "tonic/logging/dart_invoke.h"
 
 #if IMPELLER_SUPPORTS_RENDERING
 #include "flutter/lib/ui/painting/image_encoding_impeller.h"
+#include "flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.h"
 #endif
+#include "flutter/display_list/image/dl_image.h"
 #include "flutter/lib/ui/painting/image_encoding.h"
+#include "flutter/lib/ui/ui_dart_state.h"
 #include "third_party/tonic/converter/dart_converter.h"
-#include "third_party/tonic/dart_args.h"
-#include "third_party/tonic/dart_binding_macros.h"
-#include "third_party/tonic/dart_library_natives.h"
 
 namespace flutter {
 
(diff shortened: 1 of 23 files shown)


gaaclarke:
@flar @chinmaygarde friendly ping

engine/src/flutter/lib/ui/painting/image.cc
```diff
@@ -56,3 +54,114 @@ int CanvasImage::colorSpace() {
...
+}  // namespace
+
+void CanvasImage::decodeImageFromPixelsSync(Dart_Handle pixels_handle,
+                                            uint32_t width,
+                                            uint32_t height,
+                                            int32_t pixel_format,
+                                            Dart_Handle raw_image_handle) {
+  auto* dart_state = UIDartState::Current();
+  if (!dart_state) {
+    Dart_ThrowException(tonic::ToDart("Dart state is null."));
```
chinmaygarde: IIRC, Dart_ThrowException doesn't unwind the C++ stack. So you can't early return and reasonably expect to have RAII'ed stuff cleaned up. Instead of multiple `Dart_ThrowException`s, you could just return a `StatusOr` or an `optional` with an error message, propagate that up and just call a single Dart_ThrowException. This reason is why we have so many instances where we just log the error. I think we can do better now with absl and the new patterns.

engine/src/flutter/lib/ui/painting/image.cc
```diff
@@ -56,3 +54,114 @@ int CanvasImage::colorSpace() {
...
+    if (pixel_format == 2) {  // rgbaFloat32
+      image_info = image_info.makeAlphaType(kUnpremul_SkAlphaType);
+    } else {
+      image_info = image_info.makeAlphaType(kPremul_SkAlphaType);
+    }
+
+    sk_data = SkData::MakeWithCopy(pixels.data(), pixels.num_elements());
+    sk_image = SkImages::RasterFromData(image_info, sk_data, row_bytes);
+    if (!sk_image) {
+      Dart_ThrowException(tonic::ToDart("Failed to create image from pixels."));
```
chinmaygarde: Nothing after this ThrowException will run including RAII cleanups, so the data and image above will be leaked. Hence the recommendation to only call Dart_ThrowException once.

engine/src/flutter/lib/ui/painting/image.cc
chinmaygarde: cc @jason-simmons who might also remember more about the stack unwinding stuff (or lack therof).

engine/src/flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.cc
```diff
@@ -0,0 +1,133 @@
...
+
+PixelDeferredImageGPUImpeller::ImageWrapper::~ImageWrapper() = default;
+
+bool PixelDeferredImageGPUImpeller::ImageWrapper::isTextureBacked() const {
+  return texture_ && texture_->IsValid();
+}
+
+void PixelDeferredImageGPUImpeller::ImageWrapper::SnapshotImage(
+    sk_sp<SkImage> image) {
+  fml::TaskRunner::RunNowOrPostTask(
```
chinmaygarde, line 101: This may only be called once right? Perhaps guard against incorrect use?

chinmaygarde (commented):
The ThrowException not unwinding the stack is the big red flag. But past that, the rest looks good to me!

engine/src/flutter/lib/ui/painting/image.cc
gaaclarke: This was following the pattern in canvas.cc.  I thought it was setting some state in the dart vm and not performing any jump.

example: https://github.com/flutter/flutter/blob/2e5ed1aea77fffc7a791300b9c1ea6c1ce3ad82e/engine/src/flutter/lib/ui/painting/canvas.cc#L385

gaaclarke: Ahh, you're right.  We should fix this in canvas too.  Here's from the dart documentation:

```
 * Dart_PropagateError and Dart_ThrowException do not return.  Instead
 * they transfer control non-locally using a setjmp-like mechanism.
 * This can be inconvenient if you have resources that you need to
 * clean up before propagating the error.
 ```

Comment (gaaclarke) [reply engine/src/flutter/lib/ui/painting/image.cc]: Done, I split out a helper function that returns static strings as errors.  I also fixed a potential problem in canvas.cc where there was a c++ struct on the stack when we threw.

File: engine/src/flutter/lib/ui/painting/image.cc (hunk 2 above)
Comment (gaaclarke) [reply engine/src/flutter/lib/ui/painting/image.cc]: done

File: engine/src/flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.cc (hunk 3 above)
Comment (gaaclarke) [reply engine/src/flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.cc:101]: done

File: engine/src/flutter/lib/ui/painting/image.cc (hunk 1 above)
Comment (gaaclarke) [reply engine/src/flutter/lib/ui/painting/image.cc]: We should probably remove those return statements after the throws too, those are confusing.

File: engine/src/flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.cc [hunk 4]
```diff
@@ -0,0 +1,133 @@
...
+size_t PixelDeferredImageGPUImpeller::GetApproximateByteSize() const {
+  auto size = sizeof(PixelDeferredImageGPUImpeller);
+  if (wrapper_) {
+    if (wrapper_->texture()) {
+      size += wrapper_->texture()
+                  ->GetTextureDescriptor()
+                  .GetByteSizeOfBaseMipLevel();
+    } else {
+      size += wrapper_->size().Area() * 4;
+    }
```
flar, line 65: So the reported size might change asynchronously as the image is decoded and delivered? Is there a caller that might run into problems with that?

engine/src/flutter/testing/dart/decode_image_from_pixels_sync_test.dart
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }, skip: !impellerEnabled);
+
+  test('decodeImageFromPixelsSync throws if not Impeller', () {
+    final pixels = Uint8List(4);
+    expect(
+      () => decodeImageFromPixelsSync(pixels, 1, 1, PixelFormat.rgba8888),
+      throwsA(isA<String>()),
+    );
+  }, skip: impellerEnabled);
```
flar, line 62: Just double checking that null pixels and invalid pixel format are not possible to test due to the Dart API not allowing null and requiring a valid enum value?

flar (approved):
LGTM with a couple of verifying questions...

engine/src/flutter/lib/ui/painting/image.cc
chinmaygarde: > We should fix this in canvas too... I also fixed a potential problem in canvas.cc where there was a c++ struct on the stack when we threw... 

Yeah. I am sure we have missed a couple of other spots in review. I was also considering just banning throwing of exception from native code. Instead returning an error string to Dart code that can then throw the exception.

> remove those return statements after the throws too, those are confusing

Yeah. Perhaps we can decorate the call with `[[noreturn]]`. That way, the compiler will warn about redundant code the author didn't realize was redundant.

chinmaygarde (approved):

engine/src/flutter/lib/ui/painting/pixel_deferred_image_gpu_impeller.cc
gaaclarke: This is used by the raster cache.  The callers are all about estimating size of things.  This is also how DlDeferredImageGPUImpeller::GetApproximateByteSize works so it should be good.

engine/src/flutter/testing/dart/decode_image_from_pixels_sync_test.dart
gaaclarke: yes, sir

PR #179639, opened 2025-12-09
Restore CLI precedence for web headers and HTTPS over web_dev_config.yaml

### LINKED ISSUE #179014: Flutter 3.38 run command line arguments not working for web development config
From 3.38 it is not possible to pass in the command line the following parameters:
- web-hostname
- web-port
- web-tls-cert-path
- web-tls-cert-key-path

This issue has been introduced by this PR: https://github.com/flutter/flutter/pull/172175

The implementation document states:

> 5. Configuration Precedence
> A small but critical section to prevent user confusion, explicitly stating the override order:
> Command-Line Arguments (e.g., --web-port)
> web_dev_config.yaml settings
> Built-in default values

The flutter documentation for this feature ](https://docs.flutter.dev/platform-integration/web/web-dev-config-file) also specifies:

> Configuration precedence
> 
> Remember the order of precedence for settings:
> 
>     Command-line arguments (such as --web-hostname, --web-port)
>     web_dev_config.yaml settings
>     Built-in default values

And (some) of the command line arguments are still available in `flutter run -h`

>     --web-tls-cert-path                                      The certificate that host will use to serve using TLS connection. If not provided, the tool will use default http scheme.
>     --web-tls-cert-key-path                                  The certificate key that host will use to authenticate cert. If not provided, the tool will use default http scheme.

But regardless of the above, the implementation now ignores command line arguments and fully relies on the newly introduced `web_dev_config.yaml`.

Saqib198:
This PR restores the correct precedence behavior for web development server configuration. Since Flutter 3.38, CLI arguments were being ignored when `web_dev_config.yaml` was present — specifically for HTTPS and headers.

According to the design and documentation, the precedence should be:

1. Command-line arguments (highest priority)
2. `web_dev_config.yaml`
3. Built-in defaults (lowest priority)

## Root Cause

The regression affected:

* **HTTP headers** — merge order was reversed, causing file config to override CLI
* **HTTPS config** — CLI TLS arguments were ignored under certain conditions

Note:
`--web-port` and `--web-hostname` were already working correctly due to the `host ?? this.host` and `port ?? this.port` pattern.
The observed issue was due to headers + HTTPS only.

## Changes Made

* Fixed header merge order so CLI takes precedence
* Corrected HTTPS config resolution and precedence logic
* Unified logic across `run` and `drive` commands
* Simplified HTTPS handling using `HttpsConfig.parse()`
* Added tests for CLI precedence behavior

## Before (broken)

```
flutter run -d chrome --web-tls-cert-path cert.pem
```

❌ CLI TLS args ignored when config file existed
❌ headers overridden by file config

## After (fixed)

```
flutter run -d chrome --web-tls-cert-path cert.pem
```

✅ CLI TLS args respected
✅ headers override config as expected

## Tests

* Added tests verifying CLI takes precedence
* All existing and new tests pass
* No breaking changes introduced

## Fixes

Fixes: #179014


Saqib198:
Thanks for the review!
You're absolutely right — the previous logic in drive.dart was incorrectly depending on fileConfig != null, which caused CLI HTTPS arguments to be ignored when no config file was present.

I've updated the logic to match run.dart so that:

CLI arguments always take precedence

File config works only as a fallback

HTTPS config is correctly applied even without web_dev_config.yaml

Please let me know if anything else is needed.

packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -495,10 +495,24 @@ class RunCommand extends RunCommandBase {
...
-        certKeyPath: stringArg('web-tls-cert-key-path'),
-      );
+      // CLI arguments for TLS certificates
+      final String? cliCertPath = stringArg('web-tls-cert-path');
+      final String? cliCertKeyPath = stringArg('web-tls-cert-key-path');
+
+      // Determine HTTPS config with CLI > file precedence
+      // If CLI provides TLS certs, use them (with file config as fallback for missing values)
+      // If CLI doesn't provide TLS certs, use file config
+      final HttpsConfig? httpsConfig;
```
bkonyi: Consider doing this instead (also in `drive.dart`):

```dart
final HttpsConfig? httpsConfig = (fileConfig.https ?? HttpsConfig()).copyWith(
  certPath: stringArg('web-tls-cert-path'),
  certKeyPath: stringArg('web-tls-cert-key-path'),
);
```

bkonyi (commented):
Thanks for the PR @Saqib198!

This looks good to me in general, with a couple of minor suggestions. Also, this will need a test before it can be submitted.

packages/flutter_tools/lib/src/commands/run.dart
Saqib198: Thanks for the suggestion @bkonyi!

I tried implementing it, but HttpsConfig() cannot be instantiated without arguments since both certPath and certKeyPath are required parameters in the constructor:

const HttpsConfig({required this.certPath, required this.certKeyPath});

Instead, I've simplified the code using HttpsConfig.parse() which handles all the cases cleanly:

final HttpsConfig? httpsConfig = HttpsConfig.parse(
    stringArg('web-tls-cert-path') ?? fileConfig.https?.certPath,
    stringArg('web-tls-cert-key-path') ?? fileConfig.https?.certKeyPath,
);

  This achieves the same simplification (reduced from 15 lines to 4 lines) while:
  - CLI args take precedence over file config
  - Falls back to file config values when CLI args not provided
  - Returns null if neither provides both values
  - Throws ArgumentError if only one path is provided (existing behavior)

  I've also added tests for CLI precedence as requested. Let me know if you'd like any changes!

packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -382,6 +382,30 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
       );
     }
   }
+
+  Future<WebDevServerConfig> webDevServerConfigCore() async {
```
kevmoo, line 391: @Saqib198 – I moved things into this shared function so we have fewer places to keep this logic up-to-date!

packages/flutter_tools/lib/src/web/devfs_config.dart
```diff
@@ -163,7 +167,7 @@ class WebDevServerConfig {
       host: host ?? this.host,
```
kevmoo, line 167: Yeah, these should all be changed, too – right? @Saqib198 

RE comments from @mdebbar

mdebbar: I believe the way they are now is correct. But the issue says `web-port`'s and `web-hostname`'s precedence is incorrect, which is confusing.

kevmoo: Ah!

I get it.

By putting `headers` after `this.headers`, it overwrites.

But with `port` and `host`, the `??` has things the other way.

So it reads weird, but I think the code as it is now is correct.

kevmoo: But I agree, I see no code changes for anything but `headers`.

So was it really broken? Or are we missing something?

Saqib198: Thanks for calling this out — happy to clarify.

You’re right that web-port and web-hostname already had correct CLI > file precedence.
They use the pattern:

  host: host ?? this.host,
  port: port ?? this.port,

where `host` / `port` come from the CLI layer, so CLI values correctly override
web_dev_config.yaml today.

The regression introduced in 3.38 specifically affected:
• headers (due to merge order: `{...?headers, ...this.headers}`)
• HTTPS config (due to conditional logic that ignored CLI TLS args when file config
  was present or missing)

That’s why this PR only changes:
• header merge order
• HTTPS config resolution

The issue description mentioned web-port / web-hostname because the *symptom* users
observed was “CLI args ignored”, but the actual root cause was headers + HTTPS.
Port/hostname were already behaving correctly, so no change was needed there.

Happy to update the PR description or issue wording if that helps reduce confusion.

mdebbar: Ok, that makes more sense. Thanks for clarifying!

I would say yes let's update the issue and PR descriptions. If it confused me and @kevmoo, I'm sure it'll confuse more people later when they look at the history of this code.

mdebbar (approved):
LGTM

mdebbar:
@bkonyi does this look good to you? I think we need a 2nd approval to merge.

Saqib198:
Hi @bkonyi 👋  
All requested changes are addressed, tests are added, and the PR description is updated.
has been clarified per @mdebbar’s feedback.

Whenever you get a chance, please take a look.
Thanks a lot!

bkonyi (approved):

PR #179659, opened 2025-12-10
[ios][pv] accept/reject gesture based on hitTest (with new widget API)

hellohuanlin:
This is a follow up PR to [this original PR](https://github.com/flutter/flutter/pull/177859). 

The difference is the API - the original PR chooses Option 1 [in the design doc](https://docs.google.com/document/d/1ag4drAdJsR7y-rQZkqJWc6tOQ4qCbflQSGyoxsSC6MM/edit?tab=t.0), while this PR chooses Option 3.  

## Usage

To directly use flutter API, just pass in the policy when creating UiKitView widget. 

```
UiKitView(
  ...
  gestureBlockingPolicy: UiKitViewGestureBlockingPolicy)
  ...
)
```

For plugins, we need to update plugins to use this new API. 

```
WebView(
  ...
  gestureBlockingPolicy: UiKitViewGestureBlockingPolicy
) {
  return UiKitView(
    ..
    gestureBlockingPolicy: gestureBlockingPolicy
  )
}
```
For more information, refer to [the old PR](https://github.com/flutter/flutter/pull/177859). 

https://github.com/flutter/flutter/issues/175099
https://github.com/flutter/flutter/issues/165787

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -308,6 +308,11 @@ void _dispatchPointerDataPacket(ByteData packet) {
 }
 
 @pragma('vm:entry-point')
+bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
+  return PlatformDispatcher.instance._platformViewShouldAcceptGesture(viewId, x, y);
+}
+
+@pragma('vm:entry-point')
 void _dispatchSemanticsAction(int viewId, int nodeId, int action, ByteData? args) {
   PlatformDispatcher.instance._dispatchSemanticsAction(viewId, nodeId, action, args);
 }
@@ -414,6 +419,24 @@ void _invoke3<A1, A2, A3>(
     zone.runGuarded(() {
       callback(arg1, arg2, arg3);
     });
+  }
+}
+
+R? _invoke3WithReturn<A1, A2, A3, R>(
+  R Function(A1 a1, A2 a2, A3 a3)? callback,
+  Zone zone,
+  A1 arg1,
+  A2 arg2,
+  A3 arg3,
+) {
+  if (callback == null) {
+    return null;
+  }
+  if (identical(zone, Zone.current)) {
+    return callback(arg1, arg2, arg3);
+  } else {
+    // TODO: how to handle synchronous return? Should we assume zone is always current?
+    return null;
   }
 }
 
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -422,6 +422,12 @@ void _invoke3<A1, A2, A3>(
   }
 }
 
+/// Invokes [callback] inside the given [zone] passing it [arg1], [arg2], and [arg3],
+/// and returns a nullable result of the specified type.
+///
+/// The 3 in the name refers to the number of arguments expected by
+/// the callback (and thus passed to this function, in addition to the
+/// callback itself and the zone in which the callback is executed).
 R? _invoke3WithReturn<A1, A2, A3, R>(
   R Function(A1 a1, A2 a2, A3 a3)? callback,
   Zone zone,
@@ -435,8 +441,15 @@ R? _invoke3WithReturn<A1, A2, A3, R>(
   if (identical(zone, Zone.current)) {
     return callback(arg1, arg2, arg3);
   } else {
-    // TODO: how to handle synchronous return? Should we assume zone is always current?
-    return null;
+    // TODO(hellohuanlin): add `zone.runGuardedWithReturn` API.
+    try {
+      return zone.run(() {
+        return callback(arg1, arg2, arg3);
+      });
+    } catch (e, s) {
+      zone.handleUncaughtError(e, s);
+      return null;
+    }
   }
 }
 
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -307,11 +307,12 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+// TODO(hellohuanlin): rename function to _onHitTest.
 @pragma('vm:entry-point')
 bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
-  Offset offset = Offset(x, y);
-  // Engine must pass in a valid viewId.
-  FlutterView view = PlatformDispatcher.instance._views[viewId]!;
+  assert(PlatformDispatcher.instance._views.containsKey(viewId), 'View $viewId does not exist.');
+  final FlutterView view = PlatformDispatcher.instance._views[viewId]!;
+  final offset = Offset(x, y);
   final HitTestRequest request = HitTestRequest(view: view, offset: offset);
   final HitTestResponse response = PlatformDispatcher.instance._hitTest(request);
   return response.isPlatformView;
@@ -440,15 +441,9 @@ R? _invoke1WithReturn<A1, R>(R Function(A1 a1)? callback, Zone zone, A1 arg1) {
   if (identical(zone, Zone.current)) {
     return callback(arg1);
   } else {
-    // TODO(hellohuanlin): add `zone.runGuardedWithReturn` API.
-    try {
-      return zone.run(() {
-        return callback(arg1);
-      });
-    } catch (e, s) {
-      zone.handleUncaughtError(e, s);
-      return null;
-    }
+    return zone.runZonedGuarded(() {
+      return callback(arg1);
+    });
   }
 }
 
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -308,8 +308,8 @@ void _dispatchPointerDataPacket(ByteData packet) {
 }
 
 @pragma('vm:entry-point')
-bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
-  return PlatformDispatcher.instance._platformViewShouldAcceptGesture(viewId, x, y);
+bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
+  return PlatformDispatcher.instance._platformViewShouldAcceptTouch(viewId, x, y);
 }
 
 @pragma('vm:entry-point')
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -313,7 +313,7 @@ bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
   assert(PlatformDispatcher.instance._views.containsKey(viewId), 'View $viewId does not exist.');
   final FlutterView view = PlatformDispatcher.instance._views[viewId]!;
   final offset = Offset(x, y);
-  final HitTestRequest request = HitTestRequest(view: view, offset: offset);
+  final request = HitTestRequest(view: view, offset: offset);
   final HitTestResponse response = PlatformDispatcher.instance._hitTest(request);
   return response.isPlatformView;
 }
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -441,9 +441,14 @@ R? _invoke1WithReturn<A1, R>(R Function(A1 a1)? callback, Zone zone, A1 arg1) {
   if (identical(zone, Zone.current)) {
     return callback(arg1);
   } else {
-    return zone.runZonedGuarded(() {
-      return callback(arg1);
-    });
+    return runZonedGuarded(
+      () {
+        return callback(arg1);
+      },
+      (e, s) {
+        zone.handleUncaughtError(e, s);
+      },
+    );
   }
 }
 
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -310,7 +310,7 @@ void _dispatchPointerDataPacket(ByteData packet) {
 @pragma('vm:entry-point')
 bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
   Offset offset = Offset(x, y);
-  // TODO(hellohuanlin) investigate how to handle invalid id.
+  // Engine must pass in a valid viewId.
   FlutterView view = PlatformDispatcher.instance._views[viewId]!;
   final HitTestRequest request = HitTestRequest(view: view, offset: offset);
   final HitTestResponse response = PlatformDispatcher.instance._hitTest(request);
@@ -427,10 +427,10 @@ void _invoke3<A1, A2, A3>(
   }
 }
 
-/// Invokes [callback] inside the given [zone] passing it [arg1], [arg2], and [arg3],
+/// Invokes [callback] inside the given [zone] passing it [arg1],
 /// and returns a nullable result of the specified type.
 ///
-/// The 3 in the name refers to the number of arguments expected by
+/// The 1 in the name refers to the number of arguments expected by
 /// the callback (and thus passed to this function, in addition to the
 /// callback itself and the zone in which the callback is executed).
 R? _invoke1WithReturn<A1, R>(R Function(A1 a1)? callback, Zone zone, A1 arg1) {
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -308,6 +308,7 @@ void _dispatchPointerDataPacket(ByteData packet) {
 }
 
 // TODO(hellohuanlin): rename function to _onHitTest.
+// See: https://github.com/flutter/flutter/issues/179762.
 @pragma('vm:entry-point')
 bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
   assert(PlatformDispatcher.instance._views.containsKey(viewId), 'View $viewId does not exist.');
--- a/engine/src/flutter/lib/ui/hooks.dart
+++ b/engine/src/flutter/lib/ui/hooks.dart
@@ -309,7 +309,12 @@ void _dispatchPointerDataPacket(ByteData packet) {
 
 @pragma('vm:entry-point')
 bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
-  return PlatformDispatcher.instance._platformViewShouldAcceptTouch(viewId, x, y);
+  Offset offset = Offset(x, y);
+  // TODO(hellohuanlin) investigate how to handle invalid id.
+  FlutterView view = PlatformDispatcher.instance._views[viewId]!;
+  final HitTestRequest request = HitTestRequest(view: view, offset: offset);
+  final HitTestResponse response = PlatformDispatcher.instance._hitTest(request);
+  return response.isPlatformView;
 }
 
 @pragma('vm:entry-point')
@@ -428,23 +433,17 @@ void _invoke3<A1, A2, A3>(
 /// The 3 in the name refers to the number of arguments expected by
 /// the callback (and thus passed to this function, in addition to the
 /// callback itself and the zone in which the callback is executed).
-R? _invoke3WithReturn<A1, A2, A3, R>(
-  R Function(A1 a1, A2 a2, A3 a3)? callback,
-  Zone zone,
-  A1 arg1,
-  A2 arg2,
-  A3 arg3,
-) {
+R? _invoke1WithReturn<A1, R>(R Function(A1 a1)? callback, Zone zone, A1 arg1) {
   if (callback == null) {
     return null;
   }
   if (identical(zone, Zone.current)) {
-    return callback(arg1, arg2, arg3);
+    return callback(arg1);
   } else {
     // TODO(hellohuanlin): add `zone.runGuardedWithReturn` API.
     try {
       return zone.run(() {
-        return callback(arg1, arg2, arg3);
+        return callback(arg1);
       });
     } catch (e, s) {
       zone.handleUncaughtError(e, s);
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,6 +43,7 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptGesture].
 typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
 
 /// Signature for responses to platform messages.
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -1391,11 +1391,14 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback is invoked when a platform view performs hitTest and queries
+  /// the framework if the platform view should receive the touches.
   HitTestCallback? get onHitTest => _onHitTest;
   HitTestCallback? _onHitTest;
   Zone _onHitTestZone = Zone.root;
   set onHitTest(HitTestCallback? callback) {
     _onHitTest = callback;
+    _onHitTestZone = Zone.current;
   }
 
   // Called from the engine via hooks.dart.
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,8 +43,8 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
-/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptTouch].
-typedef PlatformViewShouldAcceptTouchCallback = bool Function(int viewId, double x, double y);
+/// Signature for [PlatformDispatcher.onHitTest].
+typedef HitTestCallback = HitTestResponse Function(HitTestRequest);
 
 /// Signature for responses to platform messages.
 ///
@@ -1391,12 +1391,11 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
-  PlatformViewShouldAcceptTouchCallback? get onPlatformViewShouldAcceptTouch =>
-      _onPlatformViewShouldAcceptTouch;
-  PlatformViewShouldAcceptTouchCallback? _onPlatformViewShouldAcceptTouch;
-  Zone _onPlatformViewShouldAcceptTouchZone = Zone.root;
-  set onPlatformViewShouldAcceptTouch(PlatformViewShouldAcceptTouchCallback? callback) {
-    _onPlatformViewShouldAcceptTouch = callback;
+  HitTestCallback? get onHitTest => _onHitTest;
+  HitTestCallback? _onHitTest;
+  Zone _onHitTestZone = Zone.root;
+  set onHitTest(HitTestCallback? callback) {
+    _onHitTest = callback;
   }
 
   // Called from the engine via hooks.dart.
@@ -1436,15 +1435,13 @@ class PlatformDispatcher {
     );
   }
 
-  bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
-    return _invoke3WithReturn<int, double, double, bool>(
-          onPlatformViewShouldAcceptTouch,
-          _onPlatformViewShouldAcceptTouchZone,
-          viewId,
-          x,
-          y,
+  HitTestResponse _hitTest(HitTestRequest request) {
+    return _invoke1WithReturn<HitTestRequest, HitTestResponse>(
+          onHitTest,
+          _onHitTestZone,
+          request,
         ) ??
-        false;
+        const HitTestResponse();
   }
 
   ErrorCallback? _onError;
@@ -3210,4 +3207,17 @@ enum ViewFocusDirection {
   ///
   /// This is typically result of the user pressing shift + tab.
   backward,
+}
+
+class HitTestRequest {
+  const HitTestRequest({required this.view, required this.offset});
+
+  final FlutterView view;
+  final Offset offset;
+}
+
+class HitTestResponse {
+  const HitTestResponse({this.isPlatformView = false});
+
+  final bool isPlatformView;
 }
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -3209,15 +3209,21 @@ enum ViewFocusDirection {
   backward,
 }
 
+/// The request containing all the information required to perform framework hitTest.
 class HitTestRequest {
   const HitTestRequest({required this.view, required this.offset});
 
+  /// The flutter view.
   final FlutterView view;
+
+  /// The touch location on screen.
   final Offset offset;
 }
 
+/// The response from the framework hitTest.
 class HitTestResponse {
   const HitTestResponse({this.isPlatformView = false});
 
+  // whether the top-most hit target is a platform view.
   final bool isPlatformView;
 }
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,6 +43,8 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
+
 /// Signature for responses to platform messages.
 ///
 /// Used as a parameter to [PlatformDispatcher.sendPlatformMessage] and
@@ -1388,6 +1390,14 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  PlatformViewShouldAcceptGestureCallback? get onPlatformViewShouldAcceptGesture =>
+      _onPlatformViewShouldAcceptGesture;
+  PlatformViewShouldAcceptGestureCallback? _onPlatformViewShouldAcceptGesture;
+  Zone _onPlatformViewShouldAcceptGestureZone = Zone.root;
+  set onPlatformViewShouldAcceptGesture(PlatformViewShouldAcceptGestureCallback? callback) {
+    _onPlatformViewShouldAcceptGesture = callback;
+  }
+
   // Called from the engine via hooks.dart.
   void _updateFrameData(int frameNumber) {
     final FrameData previous = _frameData;
@@ -1423,6 +1433,17 @@ class PlatformDispatcher {
         arguments: args,
       ),
     );
+  }
+
+  bool _platformViewShouldAcceptGesture(int viewId, double x, double y) {
+    return _invoke3WithReturn<int, double, double, bool>(
+          onPlatformViewShouldAcceptGesture,
+          _onPlatformViewShouldAcceptGestureZone,
+          viewId,
+          x,
+          y,
+        ) ??
+        false;
   }
 
   ErrorCallback? _onError;
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -1391,12 +1391,10 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
-  /// A callback invoked when platform wants to perform a hittest on a [FlutterView].
+  /// A callback invoked when platform wants to hit test a [FlutterView].
   ///
-  /// The callback are expected to return value contains the top-most hittest target that are hit at the
-  /// [HitTestRequest.offset] in the [HitTestRequest.view].
-  ///
-  /// This is typically used by iOS to determine whether a hittest will hit a [UIKitView].
+  /// For example, this is used by iOS to determine if a gesture hits a
+  /// [UIKitView].
   HitTestCallback? get onHitTest => _onHitTest;
   HitTestCallback? _onHitTest;
   Zone _onHitTestZone = Zone.root;
--- a/engine/src/flutter/lib/ui/platform_dispatcher.dart
+++ b/engine/src/flutter/lib/ui/platform_dispatcher.dart
@@ -43,8 +43,8 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
-/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptGesture].
-typedef PlatformViewShouldAcceptGestureCallback = bool Function(int viewId, double x, double y);
+/// Signature for [PlatformDispatcher.onPlatformViewShouldAcceptTouch].
+typedef PlatformViewShouldAcceptTouchCallback = bool Function(int viewId, double x, double y);
 
 /// Signature for responses to platform messages.
 ///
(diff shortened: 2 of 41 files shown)


engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,16 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+@pragma('vm:entry-point')
+bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
+  Offset offset = Offset(x, y);
+  // Engine must pass in a valid viewId.
+  FlutterView view = PlatformDispatcher.instance._views[viewId]!;
```
hellohuanlin: leftover from our discussion @chunhtai & @loic-sharma. Feel free to continue discussion here.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -22,6 +22,11 @@
```
hellohuanlin, line 24: @vashworth This file is the main diff between the 2 PRs (previous PR didn't touch it)

hellohuanlin:
I'm still fixing some tests, but this should be ready for review.

packages/flutter/lib/src/services/platform_views.dart
```diff
@@ -1574,6 +1592,30 @@ abstract class DarwinPlatformViewController {
...
+  /// decides they should be blocked.
+  ///
+  /// This is similar to FlutterPlatformViewGestureRecognizersBlockingPolicyEager. However,
+  /// internally it performs hit test rather than a blocking gesture recognizer. This addresses
+  /// a few bugs related to WKWebView being untappable. See
+  /// https://github.com/flutter/flutter/issues/175099.
+  touchBlockingOnly,
+
+  /// Fallback to use the policy set by the `registerViewFactory` engine API in FlutterPlugin.h.
+  fallbackToPluginDefault,
```
hellohuanlin, line 1616: This case is discussed in the design doc: https://docs.google.com/document/d/1ag4drAdJsR7y-rQZkqJWc6tOQ4qCbflQSGyoxsSC6MM/edit?tab=t.0#heading=h.k2qqvy11v9xi

Alternative approach would be to make the policy nullable, discussed in the doc too

hellohuanlin:
hmmmm it looks like i have to define `HitTestRequest/Response` in web version of `platform_dispatcher.dart` as well: 

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8695817196553537361/+/u/run_test.dart_for_analyze_shard_and_subshard_None/stdout

```
executing: /b/s/w/ir/x/w/flutter/bin/cache/dart-sdk/bin/dart compile wasm --packages=/b/s/w/ir/x/w/flutter/.dart_tool/package_config.json --extra-compiler-option=--platform=/b/s/w/ir/x/w/flutter/bin/cache/flutter_web_sdk/kernel/dart2wasm_platform.dill -Ddart.vm.profile=false -Ddart.vm.product=true --extra-compiler-option=--delete-tostring-package-uri=dart:ui --extra-compiler-option=--delete-tostring-package-uri=package:flutter --extra-compiler-option=--import-shared-memory --extra-compiler-option=--shared-memory-max-pages=32768 -DFLUTTER_VERSION=3.40.0-1.0.pre-126 -DFLUTTER_CHANNEL=[user-branch] -DFLUTTER_GIT_URL=unknown source -DFLUTTER_FRAMEWORK_REVISION=b14a309712 -DFLUTTER_ENGINE_REVISION=b14a309712 -DFLUTTER_DART_VERSION=3.11.0 (build 3.11.0-230.0.dev) -DFLUTTER_WEB_USE_SKIA=false -DFLUTTER_WEB_USE_SKWASM=true -DFLUTTER_WEB_CANVASKIT_URL=https://www.gstatic.com/flutter-canvaskit/b14a309712fb5843e6bc4fe922ef8973971ac096/ --extra-compiler-option=--depfile=/b/s/w/ir/x/w/flutter/dev/integration_tests/web/.dart_tool/flutter_build/c870cb0b44ebb9524086a88c540e5d18/dart2wasm.d -O2 --strip-wasm --no-source-maps --minify --extra-compiler-option=--dry-run -o /b/s/w/ir/x/w/flutter/dev/integration_tests/web/.dart_tool/flutter_build/c870cb0b44ebb9524086a88c540e5d18/main.dart.wasm /b/s/w/ir/x/w/flutter/dev/integration_tests/web/.dart_tool/flutter_build/c870cb0b44ebb9524086a88c540e5d18/main.dart
[+6628 ms] ../../../packages/flutter/lib/src/gestures/binding.dart:327:3:
           Error: Type 'ui.HitTestResponse' not found.
             ui.HitTestResponse _handleHitTest(ui.HitTestRequest request) {
             ^^^^^^^^^^^^^^^^^^
           ../../../packages/flutter/lib/src/gestures/binding.dart:327:37:
           Error: Type 'ui.HitTestRequest' not found.
             ui.HitTestResponse _handleHitTest(ui.HitTestRequest request) {
                                               ^^^^^^^^^^^^^^^^^
           ../../../packages/flutter/lib/src/gestures/binding.dart:286:9:
           Error: The setter 'onHitTest' isn't defined for the type 'PlatformDispatcher'.
            - 'PlatformDispatcher' is from 'dart:ui'.
                 ..onHitTest = _handleHitTest;
                   ^^^^^^^^^
           ../../../packages/flutter/lib/src/gestures/binding.dart:327:40:
           Error: 'HitTestRequest' isn't a type.
             ui.HitTestResponse _handleHitTest(ui.HitTestRequest request) {
                                                  ^^^^^^^^^^^^^^
           ../../../packages/flutter/lib/src/gestures/binding.dart:332:23:
           Error: Couldn't find constructor 'HitTestResponse'.
                 return const ui.HitTestResponse();
                                 ^^^^^^^^^^^^^^^
           ../../../packages/flutter/lib/src/gestures/binding.dart:335:15:
           Error: Method not found: 'HitTestResponse'.
               return ui.HitTestResponse(isPlatformView: firstHit is NativeHitTestTarget);
                         ^^^^^^^^^^^^^^^
           Error: Compilation failed.
```

I'm not familiar with this part

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -43,6 +43,9 @@ typedef KeyDataCallback = bool Function(KeyData data);
 /// Signature for [PlatformDispatcher.onSemanticsActionEvent].
 typedef SemanticsActionEventCallback = void Function(SemanticsActionEvent action);
 
+/// Signature for [PlatformDispatcher.onHitTest].
+typedef HitTestCallback = HitTestResponse Function(HitTestRequest);
```
loic-sharma:
```suggestion
typedef HitTestCallback = HitTestResponse Function(HitTestRequest request);
```

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -417,6 +427,31 @@ void _invoke3<A1, A2, A3>(
...
+/// the callback (and thus passed to this function, in addition to the
+/// callback itself and the zone in which the callback is executed).
+R? _invoke1WithReturn<A1, R>(R Function(A1 a1)? callback, Zone zone, A1 arg1) {
+  if (callback == null) {
+    return null;
+  }
+  if (identical(zone, Zone.current)) {
+    return callback(arg1);
+  } else {
+    // TODO(hellohuanlin): add `zone.runGuardedWithReturn` API.
```
loic-sharma: Could you explain why this can't use `runZonedGuarded`? That returns the callback value.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,16 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+@pragma('vm:entry-point')
```
loic-sharma, line 312: I would add a TODO to change this to a `_onHitTest` hook and to also push down this mapping logic down to `PlatformDispatcher`. We should clean this up eventually, but that can be later :)

chunhtai:
> hmmmm it looks like i have to define HitTestRequest/Response in web version of platform_dispatcher.dart as well

That is correct you will need to mirror the class declaration in the web as well.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,16 @@ void _dispatchPointerDataPacket(ByteData packet) {
...
+bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
+  Offset offset = Offset(x, y);
+  // Engine must pass in a valid viewId.
```
loic-sharma:
```suggestion
  assert(
    PlatformDispatcher.instance._views.containsKey(viewId),
    'View $viewId does not exist.',
  );
```

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,16 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+@pragma('vm:entry-point')
+bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
+  Offset offset = Offset(x, y);
```
chunhtai:
```suggestion
  final offset = Offset(x, y);
```

engine/src/flutter/lib/ui/hooks.dart
chunhtai:
```suggestion
  final FlutterView view = PlatformDispatcher.instance._views[viewId]!;
```

loic-sharma: This is indeed the correct thing for us to do. We have the same check to verify viewport metrics are sent to a valid view:

https://github.com/flutter/flutter/blob/a7e5fa1da6eaee5cbb3df709f5581cb5926b7b27/engine/src/flutter/lib/ui/platform_dispatcher.dart#L332-L336

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1388,6 +1391,16 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback is invoked when a platform view performs hitTest and queries
+  /// the framework if the platform view should receive the touches.
```
chunhtai: Since we make this more generic, the doc should also be update

```
/// A callback invoked when platform wants to perform a hittest on a [FlutterView].
///
/// The callback are expected to return value contains the top-most hittest target that are hit at the 
/// [HitTestRequest.offset] in the [HitTestRequest.view].
///
/// This is typically used by iOS to determine whether a hittest will hit a [UIKitView].

File: engine/src/flutter/lib/ui/platform_dispatcher.dart [hunk 10]
```diff
@@ -3189,3 +3211,22 @@ enum ViewFocusDirection {
...
+  final Offset offset;
+}
+
+/// The response from the framework hitTest.
+class HitTestResponse {
+  const HitTestResponse({this.isPlatformView = false});
+
+  // whether the top-most hit target is a platform view.
+  final bool isPlatformView;
+}
```
loic-sharma: @chunhtai What do you think of this?

```suggestion
/// A request to evaluate the content of a view at a specific position.
///
/// See also:
///
/// * [PlatformDispatcher.onHitTest], the callback that the platform
///   invokes to hit test a view at a specific position.
/// * [HitTestResponse], the result of a hit test request.
class HitTestRequest {
  /// Creates a hit test request.
  const HitTestRequest({required this.view, required this.offset});

  /// The view that should be hit tested.
  final FlutterView view;

  /// The position in the [view] that should be hit tested.
  final Offset offset;
}

/// The results of hit testing a view at a specific position.
///
/// See also:
///
/// * [PlatformDispatcher.onHitTest], the callback that the platform
///   invokes to hit test a view at a specific position.
/// * [HitTestRequest], the request to hit test a view at a specific position.
class HitTestResponse {
  /// Creates a hit test response.
  const HitTestResponse({required this.isPlatformView});

  /// A response with no hit entries.
  static const HitTestResponse empty = HitTestResponse(
    isPlatformView: false,
  );

  /// Whether the first hit test entry is a platform view.
  ///
  /// The first hit test entry is typically the child that is
  /// visually "on top" (i.e., paints later).
  final bool isPlatformView;
}
```

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -3189,3 +3211,22 @@ enum ViewFocusDirection {
   /// This is typically result of the user pressing shift + tab.
   backward,
 }
+
+/// The request containing all the information required to perform framework hitTest.
+class HitTestRequest {
```
loic-sharma, line 3224: @chunhtai If turns out these cannot be `@immutable`. `@immutable` comes from `package:meta`, which `dart:ui` does not have access to.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -3189,3 +3211,22 @@ enum ViewFocusDirection {
...
+  const HitTestRequest({required this.view, required this.offset});
+
+  /// The flutter view.
```
chunhtai: This comment is not useful, since the document is for implementing the callback of PlatformDispatcher.onHitTest, should talk about how this property should be used

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -3189,3 +3211,22 @@ enum ViewFocusDirection {
...
+
+  /// The touch location on screen.
+  final Offset offset;
+}
+
+/// The response from the framework hitTest.
+class HitTestResponse {
+  const HitTestResponse({this.isPlatformView = false});
+
+  // whether the top-most hit target is a platform view.
```
chunhtai: I think `top-most` is out of scope for this class

```
/// Whether the hit target is a platform view.

File: engine/src/flutter/lib/ui/platform_dispatcher.dart [hunk 14]
```diff
@@ -3189,3 +3211,22 @@ enum ViewFocusDirection {
...
+  const HitTestRequest({required this.view, required this.offset});
+
+  /// The flutter view.
+  final FlutterView view;
+
+  /// The touch location on screen.
+  final Offset offset;
+}
+
+/// The response from the framework hitTest.
```
chunhtai: this should mention this is used by the PlatformDispatcher.onHitTest

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -77,6 +77,12 @@ void PlatformConfiguration::DidCreateIsolate() {
   dispatch_pointer_data_packet_.Set(
       tonic::DartState::Current(),
       Dart_GetField(library, tonic::ToDart("_dispatchPointerDataPacket")));
+  // The embedded platform view (e.g. UIView on iOS) is called "platform view"
+  // on framework side, but "embedded view" on engine side. On the other hand,
```
chunhtai: not related to this pr. Embedded view is also not a good name though, it collide with embedder and can be confusing. It also sometimes refers to Add-to-app (for example embedded flutter app).

I think we need a better name on the engine side.

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -461,6 +462,22 @@ class PlatformConfiguration final {
...
+  /// @param[in]  view_id               The identifier of the flutter view that
+  ///                                   hosts the embedded view.
+  /// @param[in]  touch_began_location  The touch began location.
+  ///
+  /// @return     true if the embedded view should accept touch; false
+  /// otherwise.
+  ///
+  bool EmbeddedNativeViewShouldAcceptTouch(
+      int64_t view_id,
+      const flutter::PointData& touch_began_location);
```
chunhtai: not a c++ expert, but do we want to pass by reference that may have memory complication? maybe use std::move.

chunhtai: same for the entire drilling through shell

engine/src/flutter/lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -601,3 +605,22 @@ final class ViewFocusEvent {
 enum ViewFocusState { unfocused, focused }
 
 enum ViewFocusDirection { undefined, forward, backward }
+
+/// The request containing all the information required to perform framework hitTest.
```
chunhtai: we don't need to copy the doc

engine/src/flutter/lib/ui/hooks.dart
loic-sharma: Oh right. We just changed our lints to omit obvious types. My bad 😅

engine/src/flutter/runtime/runtime_controller.h
```diff
@@ -476,6 +477,22 @@ class RuntimeController : public PlatformConfigurationClient,
...
+  /// @param[in]  view_id               The identifier of the flutter view that
+  ///                                   hosts the embedded view.
+  /// @param[in]  touch_began_location  The touch began location.
+  ///
+  /// @return     true if the embedded view should accept touch; false
+  /// otherwise.
+  ///
+  bool EmbeddedNativeViewShouldAcceptTouch(
+      int64_t view_id,
+      const flutter::PointData& touch_began_location);
```
loic-sharma: Could you leave a TODO here as well? It's not a blocker for this PR, but we should make this be a `flutter::HitTestResponse HitTest(const flutter::HitTestRequest& request)` API.

engine/src/flutter/lib/ui/window/platform_configuration.h
loic-sharma: A simple solution to address this concern and get this landed ASAP would be to just pass `double offsetX` and `double offsetY` by value instead of introducing a `flutter::PointData` value. That way we don't need to worry about the lifetime of the `flutter::PointData` value.

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: makes sense, I guess we will have to wait for dart support for data class then.

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: yup looks good to me though we should make HitTestRequest constructor private, there doesn't seem to be a valid use case where customer would create a HitTestRequest.

also it should be kEmpty instead of empty

engine/src/flutter/lib/ui/window/platform_configuration.cc
vashworth: `embedded_native_view_should_accept_touch_` perhaps?

hellohuanlin: This is left over comment from the previous PR. renaming to "embedded native view"

engine/src/flutter/lib/ui/window/platform_configuration.h
hellohuanlin: I'm not a C++ expert either, but this is to follow existing convention (e.g. how we pass `PointerDataPacket` all the way up)

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: no reason except that i didn't know about this API.  updated.

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: done

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: done

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: done. i like making `isPlatformView` required

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: done

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: done

engine/src/flutter/runtime/runtime_controller.h
hellohuanlin: done

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -327,10 +341,32 @@ - (void)onCreate:(FlutterMethodCall*)call result:(FlutterResult)result {
   // Set a unique view identifier, so the platform view can be identified in unit tests.
   platformView.accessibilityIdentifier = [NSString stringWithFormat:@"platform_view[%lld]", viewId];
 
-  FlutterTouchInterceptingView* touchInterceptor = [[FlutterTouchInterceptingView alloc]
-                  initWithEmbeddedView:platformView
-               platformViewsController:self
-      gestureRecognizersBlockingPolicy:self.gestureRecognizersBlockingPolicies[viewType]];
+  NSString* gestureBlockingPolicyValue = args[@"gestureBlockingPolicy"];
```
vashworth, line 337: This is expected to be `kGestureBlockingPolicyFallbackToPluginDefault` at first since that's the default set in the `UiKitView` Widget, correct?

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -1368,6 +1368,22 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
+  @override
+  ui.HitTestCallback? get onHitTest => _onHitTest;
+  ui.HitTestCallback? _onHitTest;
+  Zone _onHitTestZone = Zone.root;
+  @override
+  set onHitTest(ui.HitTestCallback? callback) {
+    _onHitTest = callback;
+    _onHitTestZone = Zone.current;
+  }
+
```
hellohuanlin, line 1377: @chunhtai @loic-sharma would like an extra pair of eyes on these web_ui changes (since we didn't cover them yesterday)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -101,7 +108,12 @@ @interface FlutterPlatformViewsController ()
 // The FlutterPlatformViewGestureRecognizersBlockingPolicy for each type of platform view.
 @property(nonatomic, readonly)
     std::unordered_map<std::string, FlutterPlatformViewGestureRecognizersBlockingPolicy>&
-        gestureRecognizersBlockingPolicies;
+        gestureRecognizersBlockingPoliciesByType;
```
vashworth, line 111: Nit: `ByType` doesn't make a whole lot of sense. Looks like it's more like it per factory id?

```suggestion
        gestureRecognizersBlockingPoliciesByFactoryId;
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -101,7 +108,12 @@ @interface FlutterPlatformViewsController ()
...
+
+// The FlutterPlatformViewGestureRecognizersBlockingPolicy for each instance of platform view.
+@property(nonatomic, readonly)
+    std::unordered_map<int64_t, FlutterPlatformViewGestureRecognizersBlockingPolicy>&
+        gestureRecognizersBlockingPoliciesByViewId;
```
vashworth: What's the purpose of this? I can't figure out how it's used?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
vashworth: Ehh looks like it's called `viewType` in the args. Nvm

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
hellohuanlin: good catch! it's not used.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -327,10 +341,32 @@ - (void)onCreate:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+    gestureBlockingPolicy = FlutterPlatformViewGestureRecognizersBlockingPolicyTouchBlockingOnly;
+  } else if ([gestureBlockingPolicyValue isEqualToString:kGestureBlockingPolicyEagerValue]) {
+    gestureBlockingPolicy = FlutterPlatformViewGestureRecognizersBlockingPolicyEager;
+  } else if ([gestureBlockingPolicyValue
+                 isEqualToString:kGestureBlockingPolicyWaitUntilTouchesEndedValue]) {
+    gestureBlockingPolicy =
+        FlutterPlatformViewGestureRecognizersBlockingPolicyWaitUntilTouchesEnded;
+  } else if ([gestureBlockingPolicyValue
+                 isEqualToString:kGestureBlockingPolicyFallbackToPluginDefault]) {
+    gestureBlockingPolicy = self.gestureRecognizersBlockingPoliciesByType[viewType];
```
vashworth, line 349: Could this be null? What happens if it's not set in `self.gestureRecognizersBlockingPoliciesByType`?

hellohuanlin: this won't be null because `FlutterEngine` passes in `Eager` policy by default when registering the view factory

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
hellohuanlin: Yes, we will start with `fallbackToPluginDefault`, which makes the new widget API a no-op, and fallbacks to the old engine API.

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: done

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -573,8 +581,30 @@ - (BOOL)containsWebView:(UIView*)view remainingSubviewDepth:(int)remainingSubvie
...
 
+- (UIView*)hitTest:(CGPoint)point withEvent:(UIEvent*)event {
+  if (_blockingPolicy == FlutterPlatformViewGestureRecognizersBlockingPolicyTouchBlockingOnly) {
+    // In release mode, FlutterTouchInterceptingView's init is called before flutterViewController
+    // is set on platformViewsController.
+    if (self.flutterViewController == nil) {
+      _flutterViewController = self.platformViewsController.flutterViewController;
+    }
+    CGPoint pointInFlutterView = [self convertPoint:point toView:self.flutterViewController.view];
+    // Block gesture if the framework instructed so (after performing its own hitTest).
```
vashworth: This confused me at first, perhaps to make it more clear?

```suggestion
    // Consult the framework on if the touch should be handled by the platform view. 
    // If NO, the touch is handled by a Flutter widget and should be blocked (by returning self).
    // If YES, the touch should continue to the standard hit-testing (through super), allowing the touch to be delivered to the underlying native platform view or one of its subviews.
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsTest.mm
```diff
@@ -3403,6 +3438,342 @@ - (void)testFlutterPlatformViewTouchesEndedOrTouchesCancelledEventDoesNotFailThe
...
+    touchInterceptorView = touchInterceptorView.superview;
+  }
+  XCTAssertNotNil(touchInterceptorView);
+
+  XCTAssert(touchInterceptorView.gestureRecognizers.count == 1);
+  UIGestureRecognizer* forwardingRecognizer = touchInterceptorView.gestureRecognizers[0];
+  XCTAssert([forwardingRecognizer isKindOfClass:[ForwardingGestureRecognizer class]]);
+}
+
+- (void)testFlutterPlatformViewBlockGestureUnderNonHitTestPolicyShouldNotAddDelayingRecognizer {
```
vashworth:
```suggestion
- (void)testFlutterPlatformViewBlockGestureUnderNonHitTestPolicyShouldAddDelayingRecognizer {
```

vashworth (approved):
iOS side LGTM. I left a couple small comments, but nothing major. I will defer to @loic-sharma and @chunhtai on the framework and embedder side of things.

engine/src/flutter/lib/ui/window/platform_configuration.h
vashworth: According to Gemini, it's fine:

> Passing const flutter::PointData& by constant reference here is a standard and efficient C++ practice. It does not introduce memory complications and is, in fact, a good design choice.

> Here’s why:

> It Avoids a Copy: Passing by reference (&) means that instead of creating a new copy of the PointData object, the function receives a reference (essentially, the memory address) to the original object. This is more efficient, especially if the object were large, as it avoids the overhead of copying data.

> It Guarantees Immutability: The const keyword ensures that the EmbeddedNativeViewShouldAcceptTouch function cannot modify the touch_began_location object. This makes the function safer and easier to reason about, as you know it won't have side effects on the caller's data.

> No Memory Management Burden: The caller of the function is responsible for the lifetime of the PointData object. The EmbeddedNativeViewShouldAcceptTouch function simply "borrows" it for the duration of the call. Since this is a synchronous call, the original PointData object will still be valid in the caller's scope when the function returns. There are no risks of dangling references or memory leaks from this parameter passing style.

> In summary, passing by const& is a safe, efficient, and conventional way to handle this in C++. It avoids copies without creating any memory management complications.

Seems legit?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsTest.mm
hellohuanlin: Eagle eyes!

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -3189,3 +3215,42 @@ enum ViewFocusDirection {
...
+/// A request to evaluate the content of a view at a specific position.
+///
+/// See also:
+///
+/// * [PlatformDispatcher.onHitTest], the callback that the platform
+///   invokes to hit test a view at a specific position.
+/// * [HitTestResponse], the result of a hit test request.
+class HitTestRequest {
+  /// Creates a hit test request.
+  const HitTestRequest({required this.view, required this.offset});
```
chunhtai, line 3226: consider make this a private constructor and move the HitTestRequest wrapping to _hitTest method in this file.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,17 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+// TODO(hellohuanlin): rename function to _onHitTest.
+@pragma('vm:entry-point')
+bool _platformViewShouldAcceptTouch(int viewId, double x, double y) {
+  assert(PlatformDispatcher.instance._views.containsKey(viewId), 'View $viewId does not exist.');
+  final FlutterView view = PlatformDispatcher.instance._views[viewId]!;
+  final offset = Offset(x, y);
+  final HitTestRequest request = HitTestRequest(view: view, offset: offset);
```
loic-sharma: You likely need to omit the obvious type here as well.

```suggestion
  final request = HitTestRequest(view: view, offset: offset);
```

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1388,6 +1391,20 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback invoked when platform wants to perform a hittest on a [FlutterView].
```
loic-sharma:
```suggestion
  /// The callback that the platform invokes to hit test a [FlutterView] at a specific position.
```

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1388,6 +1391,20 @@ class PlatformDispatcher {
...
+  ///
+  /// The callback are expected to return value contains the top-most hittest target that are hit at the
+  /// [HitTestRequest.offset] in the [HitTestRequest.view].
```
loic-sharma: This might be true today, but we will likely expand this API later. I'd cut this.

```suggestion
```

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1388,6 +1391,20 @@ class PlatformDispatcher {
...
+  ///
+  /// This is typically used by iOS to determine whether a hittest will hit a [UIKitView].
```
loic-sharma: Consider:

```suggestion
  /// For example, this is used by iOS to determine if a
  /// gesture hits a [UIKitView].
```

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
chunhtai: since this is not going to be used by web engine, just make this a variable
```
@override
ui.HitTestCallback? onHitTest
```

engine/src/flutter/lib/ui/window/platform_configuration.h
loic-sharma: I'd prefer the `double x` and `double y` approach here. We should pass this by value to avoid corrupted memory.

TLDR: I believe this code might be correct today, but this is memory corruption waiting to happen.

We've had issues here before in split thread mode where the platform thread allocates a struct on the stack and then passes it down by reference to the UI thread. This results in corrupted memory on platforms that support split threads. The correct way to avoid this problem is by allocating on the heap, passing by value, or `std::move`ing. We've been burned by this before with event channels.

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -1368,6 +1368,22 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback invoked when platform wants to perform a hittest on a [FlutterView].
+  ///
+  /// The callback are expected to return value contains the top-most hittest target that are hit at the
+  /// [HitTestRequest.offset] in the [HitTestRequest.view].
+  ///
+  /// This is typically used by iOS to determine whether a hittest will hit a [UIKitView].
```
chunhtai: Not blocking the merge

not sure why we need doc string in this case since these won't show up in the api.flutter.dev. Looking at other comment of similar property, they are writing as framework facing doc (instead of web engine facing which would otherwise make sense). @mdebbar do you know what is the intention for these doc?

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
loic-sharma: Code looks good 👍 

Could you copy the comments from the "real" platform dispatcher callback to this one?

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: done

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -307,6 +307,17 @@ void _dispatchPointerDataPacket(ByteData packet) {
   PlatformDispatcher.instance._dispatchPointerDataPacket(packet);
 }
 
+// TODO(hellohuanlin): rename function to _onHitTest.
```
loic-sharma, line 310: Would you mind creating a tracking issue for this and dropping a link on all relevant `TODO` comments? It'll make it easier for us to remember to update the C++ as well as the Dart code.

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: done

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: what is "move the HitTestRequest wrapping"?

hellohuanlin: I still need to construct the request in unit test tho

chunhtai: oh I see, yeah if this is needed for test probable shouldn't bother to hide the constructor then

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
hellohuanlin: I will keep it since it's consistent with other parts of the code

engine/src/flutter/lib/ui/hooks.dart
hellohuanlin: [will add this link](https://github.com/flutter/flutter/issues/179762)

chunhtai (approved):
LGTM

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsTest.mm
hellohuanlin: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: done

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
hellohuanlin: what is "real" platform dispatcher.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -1388,6 +1391,18 @@ class PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback invoked when platform wants to hit test a [FlutterView].
```
loic-sharma, line 1394:
```suggestion
  /// A callback invoked when the platform wants to hit test a [FlutterView].
```

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -1368,6 +1368,13 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     _onSemanticsActionEventZone = Zone.current;
   }
 
+  /// A callback invoked when platform wants to hit test a [FlutterView].
```
loic-sharma, line 1371:
```suggestion
  /// A callback invoked when the platform wants to hit test a [FlutterView].
```

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
loic-sharma: Flutter native's is the "real" platform dispatcher: engine/src/flutter/lib/ui/platform_dispatcher.dart

This looks good to me, resolving!

engine/src/flutter/runtime/runtime_controller.h
```diff
@@ -476,6 +477,23 @@ class RuntimeController : public PlatformConfigurationClient,
   ///
   bool DispatchPointerDataPacket(const PointerDataPacket& packet);
 
+  // TODO(hellohuanlin): we should make make this a generic HitTest API.
```
loic-sharma, line 480:
```suggestion
  // TODO(hellohuanlin): we should make make this a generic HitTest API.
  // See: https://github.com/flutter/flutter/issues/179762.
```

engine/src/flutter/shell/common/platform_view.h
```diff
@@ -182,6 +182,21 @@ class PlatformView {
     virtual void OnPlatformViewDispatchPointerDataPacket(
         std::unique_ptr<PointerDataPacket> packet) = 0;
 
+    //--------------------------------------------------------------------------
+    /// @brief      Requests the delegate if an embedded native view should
+    ///             accept touch at a given touch location.
```
loic-sharma, line 187:
```suggestion
    ///             accept touch at a given touch location.
    ///             This API must be called from the UI thread.
    ///             Calling this from the platform thread causes
    ///             undefined behavior if the UI and platform threads
    ///             are not merged.
```

loic-sharma (approved):
Excellent work!

engine/src/flutter/runtime/runtime_controller.h
hellohuanlin: discussed offline - i made the changes locally, but if CI passes, i won't push it.

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
hellohuanlin: discussed offline - i made the changes locally, but if CI passes, i won't push it.

engine/src/flutter/shell/common/platform_view.h
hellohuanlin: discussed offline - i made the changes locally, but if CI passes, i won't push it.

engine/src/flutter/lib/ui/platform_dispatcher.dart
hellohuanlin: discussed offline - i made the changes locally, but if CI passes, i won't push it.

engine/src/flutter/lib/ui/window/platform_configuration.h
hellohuanlin: We synced offline and decided to pass by value (remove `&`).

Issue #156985, opened 2024-10-16
Iframed flutter web in a fixed position with a ListView in a SingleChildScrollView will share scrolling with the parent page


erik-wayhaven:
### Steps to reproduce

1. Wrap a ListView in a SingleChildScrollView
2. Embed the flutter app on another website in a fixed position
3. Hover over the flutter app and scroll

### Expected results

Just the flutter app scrolls.

### Actual results

The flutter app AND the parent website scrolls.

### Code sample

<details open><summary>Code sample</summary>

Here is the flutter app: https://dartpad.dev/?id=a2e11707d9d8b33fae7d1aa737feaf80
Here is an example of it embedded in a website: https://codepen.io/erik-wayhaven/pen/gOVWXzq
You may need to resize your window to get it to scroll.

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

https://github.com/user-attachments/assets/4fe14652-bd54-496c-9ed3-5cdc94892e5e

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.24.1, on macOS 14.6.1 23G93 darwin-x64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 16.0)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2023.2)
[✓] VS Code (version 1.94.1)
[✓] Connected device (2 available)
[✓] Network resources

• No issues found!
```

</details>

darshankawar:
Thanks for the report. Seeing the same behavior as reported.

```
Stable : 3.24.3
Master: 3.27.0-1.0.pre.88
```

yjbanov:
As a full fix for this we need to make the browser aware of the inner scrolling state, so that it knows when the iframe reaches the end of the scroll extend and starts bubbling scroll event up the chain, including outside the iframe. However, that's a big change that will take us a while to implement. For the time being, it may be possible to intercept scroll/wheel/pointer events (not sure which ones need to be intercepted) and call `preventDefault` on them so that the browser does not scroll the outer view.

muhammad-hassan-shakeel:
[...]

PR #179703, opened 2025-12-10
[web] Fix scroll event bubbling in iframes

flutter-zl:
Fixes scroll event bubbling when Flutter web is embedded in an iframe(#156985).

  ## Problem                                                                                                                                                
  When a full-page Flutter web app is embedded in an iframe, scroll chaining to                                                                             
  the host page does not work. The engine unconditionally calls `preventDefault`                                                                            
  on wheel events in `engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart`,                                                                   
  which cancels the browser's native bubbling.                                                                                                              
                                                                                                                                                            
  ## Fix                                                                                                                                                    
  `packages/flutter/lib/src/widgets/scrollable.dart` now reports back via
  `respond(allowPlatformDefault: false)` when a scrollable actually handles a                                                                               
  wheel event. For a full-page app inside an iframe, the engine skips                                                                                       
  `preventDefault` when every scrollable is at its boundary, letting the browser                                                                            
  bubble the wheel out to the host page natively. Other embeddings keep the                                                                                 
  original behavior.                                                
                                                                                                                                                            
  ## Demo                                                           
  Before: https://issue-156985-before.web.app/                                                                                                              
  After:  https://issue-156985-after.web.app/


engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -775,15 +833,40 @@ mixin _WheelEventListenerMixin on _BaseAdapter {
...
+      // Only scroll parent when ALL scrollables are at boundary.
+      // IMPORTANT: Only scroll parent if:
+      // 1. Some widget wants to allow platform default (at boundary)
+      // 2. NO widget explicitly handled the event (nested scrollables case)
+      //
+      // This fixes GitHub issue #156985
+      final bool shouldScrollParent =
+          _lastWheelEventAllowedDefault && !_lastWheelEventHandledByWidget;
+      if (shouldScrollParent) {
+        scrollParentWindow(wheelEvent.deltaX, wheelEvent.deltaY);
```
mdebbar: Does this actually do anything?

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -775,15 +833,40 @@ mixin _WheelEventListenerMixin on _BaseAdapter {
...
+    // Dispatch to framework (this triggers onRespond callbacks synchronously)
+    _callback(event, _convertWheelEventToPointerData(wheelEvent));
+
+    // Determine if we should prevent default and/or scroll parent
+    final bool isInIframe = _isEmbeddedInIframe();
+
+    if (isInIframe) {
+      // In an iframe: always prevent default to block native scroll chaining.
+      // This prevents both the iframe and parent from scrolling simultaneously.
       event.preventDefault();
```
mdebbar: So when we are in an iframe now, we always prevent default.

What's the reason we aren't always preventing default outside iframes? And why does that reason not apply to apps running inside iframes?

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Yes, this executes when all Flutter scrollables are at their boundary. For example:
 1. User has a ListView scrolled to the bottom
 2. User scrolls down again
 3. ListView calls respond(allowPlatformDefault: true) since it's at boundary
 4. No widget calls respond(allowPlatformDefault: false)
 5. shouldScrollParent becomes true
 6. scrollParentWindow() is called to scroll the parent page

This is the intended behavior, propagate scroll to parent only when Flutter can't handle it.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Outside iframes, we conditionally skip preventDefault() when a scrollable is at its boundary (allowPlatformDefault: true). This lets the browser handle the scroll.

Inside iframes, this reason doesn't apply because: 
1.The Flutter app fills the entire iframe. there's no "page" within the iframe for the browser to scroll.
2."Platform default" has a different effect: outside an iframe it means "scroll this page," but inside an iframe it means "chain the scroll to the parent page". 
3.Scroll chaining is the bug: allowing platform default would cause both iframe and parent to scroll simultaneously (#156985)

So we always block native behavior in iframes, then explicitly handle parent scrolling via postMessage when all Flutter scrollables are at their boundary.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
mdebbar: I understand that this line is executed under certain conditions. But I'm trying to figure out if it has any effect. As far as I can tell, it sends a custom `flutter-scroll` event, but what does that event do or who is handling it?

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
mdebbar: Ok, I now understand the thinking behind this, thanks for explaining!

> Inside iframes, this reason doesn't apply because:
> 1.The Flutter app fills the entire iframe. there's no "page" within the iframe for the browser to scroll.

This assumption is not valid. It's possible for an iframe to contain a flutter app that doesn't fill the entire page.

Based on what you said, I wonder if we should change the `isInIframe` condition. Instead of checking for an iframe, let's check if the flutter app is filling the entire page. And if it's filling the entire page, then always prevent default. What do you think?

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Good question! In the previous implementation, scrollParentWindow() used postMessage to send a flutter-scroll event, which required the host page to add a message listener. 
I've updated the implementation in the latest commit to use direct parent.scrollBy() instead. This works automatically for same-origin iframes without any host page code needed.
For cross-origin iframes, the browser's security restrictions prevent direct parent access, so it silently fails. This is a browser limitation with no workaround. Cross-origin iframes cannot scroll their parent page without explicit host page cooperation.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: My assumption about Flutter filling the entire iframe was too strong. 
However, I still think isInIframe is the correct condition. 
The core problem is scroll chaining across the iframe boundary, not whether Flutter fills the page. In both cases, the browser's scroll chaining causes the same problem: when the user scrolls over the Flutter content, both the iframe content and parent page scroll together. The size of the Flutter app doesn't change this behavior.

mdebbar: Let me try to break it down so that I understand exactly what's changing and what issues are being fixed (and whether we are introducing any new issues).

What we used to do:
- Conditionally `preventDefault()`
- Can't unconditionally `preventDefault()` because some apps [1] don't work correctly if we unconditionally `preventDefault()`.

What this PR is doing:
- Inside iframes: unconditionally `preventDefault()` _(new behavior)_
- Outside iframes: conditionally `preventDefault()` _(existing behavior)_

With the new behavior in this PR, my concern is that some apps [1] will have a broken scrolling experience if they are placed inside an iframe.

I'm thinking we could minimize this breakage by changing the condition to check for full-page apps instead of checking for inside vs outside iframes.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
mdebbar: Similar to my other comment, this code is based on the assumption that Flutter apps inside iframes are always full-page.

Take the following example:
```
+------------------ Parent window -------------------+
|                                                    |
|                                                    |
|                                                    |
|  +-------- iframe (non-flutter host app) --------+ |
|  |                                               | |
|  |                                               | |
|  |                                               | |
|  |          +--- Flutter embedded view ----+     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          |                              |     | |
|  |          +------------------------------+     | |
|  +-----------------------------------------------+ |
+----------------------------------------------------+
```
If the user is scrolling the flutter view and there's no more content to scroll, what happens?

### Expectation
The non-flutter host app in the iframe should scroll.

### Actual
If I'm understanding the PR correctly, the parent window will be scrolled (if it has the same origin).

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Thanks for the detailed breakdown. 
The wheel event listener is attached to [_viewTarget (Flutter's root element),](https://github.com/flutter/flutter/blob/a3cfa715b13149568399db66755c8117110495f8/engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart#L762C1-L763C1) not the document or iframe body.
So for the scenario you described (Flutter + HTML scrollable content side-by-side in an iframe), the HTML elements won't trigger Flutter's wheel handler because the listener is scoped to _viewTarget. Please check https://issue-156985-after.web.app/hybrid_demo.html. Both (left side and right side)scroll independently, proving that HTML elements alongside Flutter are not affected by the fix.
Let me know if you have a specific scenario in mind that this doesn't cover.

main demo showing the original issue fix: https://issue-156985-after.web.app/

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Great. I tested with https://flutter-demo-07-before.web.app/parent.html, it did prove you are right. I updated the logic ti check isFullPage. After the change: https://flutter-demo-07-after.web.app/parent.html.

Please let me know if I miss anything.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -606,6 +639,67 @@ abstract class _BaseAdapter {
...
+        return _cachedIsFullPageApp!;
+      }
+    } catch (e) {
+      // If we can't determine embedding type, default to full-page.
+      // This is the conservative choice: it applies the iframe scroll fix,
+      // which is the expected behavior for most Flutter web apps.
+    }
+    _cachedIsFullPageApp = true;
+    return true;
+  }
```
mdebbar: There's no need to check the DOM for this. We already have this information. You can do something like:

```suggestion
  bool get _isFullPageApp => _view.embeddingStrategy is FullPageEmbeddingStrategy;
```

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -775,15 +875,52 @@ mixin _WheelEventListenerMixin on _BaseAdapter {
...
+      // Fixes GitHub issue #156985
+      final bool shouldScrollParent =
+          _lastWheelEventAllowedDefault && !_lastWheelEventHandledByWidget;
+      if (shouldScrollParent) {
+        scrollParentWindow(wheelEvent.deltaX, wheelEvent.deltaY);
+      }
+    } else {
+      // Original behavior for:
+      // 1. Apps NOT in an iframe (normal page)
+      // 2. Custom element apps in an iframe (Flutter as a component)
```
mdebbar:
```suggestion
      // 2. Mulit-view mode inside an iframe
```

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -775,15 +875,52 @@ mixin _WheelEventListenerMixin on _BaseAdapter {
...
+      //
+      // For nested scrollables, inner widgets set _lastWheelEventHandledByWidget
+      // to true, preventing parent window scroll even if outer is at boundary.
+      //
+      // Fixes GitHub issue #156985
+      final bool shouldScrollParent =
+          _lastWheelEventAllowedDefault && !_lastWheelEventHandledByWidget;
+      if (shouldScrollParent) {
+        scrollParentWindow(wheelEvent.deltaX, wheelEvent.deltaY);
+      }
```
mdebbar: One last thought here to try to avoid the `scrollParentWindow`:

Instead of doing `event.preventDefault()` and then manually scrolling the parent, we should try something like this:
```dart
if (isInIframe && isFullPage) {
  final bool shouldScrollParent =
      _lastWheelEventAllowedDefault && !_lastWheelEventHandledByWidget;
  if (!shouldScrollParent) {
    event.preventDefault();
  }
} else {
  // ...
}
```
I'm not sure if it would work or not, but if it works let's do it!

flutter-zl: Great. I updated the logic and deploy to https://issue-156985-after.web.app/.  It works. The code is cleaner.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Great. Updated.

engine/src/flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
flutter-zl: Great. I updated the logic.

mdebbar (approved):
LGTM

Issue #157435, opened 2024-10-23
Flutter Web embed mode not scrolling the hosting page


marcelpinto:
As reported in other issues, when we embed a flutter view inside another page, either using an iframe or multi-embedding view, the touch scrolling (mainly Android/iOS browsers) is not scrolling the host page

In a recent fix it was mention that this will be a follow-up but I could not find an issue for it:

> Scrolling using the mouse and the touchpad has landed, which was in the MVP of the feature. Touch-based scrolling can be addressed as a follow-up.

_Originally posted by @yjbanov in https://github.com/flutter/flutter/issues/139263#issuecomment-2234803205_

**Expected behavior:**
If a user drags up/down inside a non-scrollable flutter view/widget the host page scrolls.

This is a major blocker to use Flutter view-embedding. 

HTML example:

```html
  <div style="display: flex; flex-direction: column; background-color: #F7F7F7; overflow-y: auto;">
    <header
      style="height: 150px; background-color: rgb(63, 50, 50); color: white; display: flex; align-items: center; justify-content: center;">
      <h2>Main header title</h2>
    </header>
    <div id="flutter-host"></div>
    <footer style="height: 300px; background-color: black; color: white; display: flex; align-items: center; justify-content: center;">
        <p>Some flutter content</p>
    </footer>
  </div>
```
flutter_bootstrap.js
```js
_flutter.loader.load({
  onEntrypointLoaded: async function(engineInitializer) {
    let engine = await engineInitializer.initializeEngine({
      multiViewEnabled: true, // Enables embedded mode.
    });
    window.flutterApp = await engine.runApp();
    
    const hostElement = document.querySelector('#flutter-host');
    if (hostElement) {
        window.viewId = flutterApp.addView({
            hostElement: hostElement,
            viewConstraints: {
                minHeight: 300,
                maxHeight: Infinity,
            }
        });
        console.log("View added");
    }
  }
});
```            

The Flutter app is then a normal flutter app with a fix size (e.g 500px height).

You can see on the attached video, how the touch in the header or footer works, but not inside the flutter-view. At the end of the video, you can also see how the mouse-wheel scroll works regardless of where it is hoovered. 

https://user-images.githubusercontent.com/2639977/379262127-6b35a706-388a-416f-b873-1f70d22c62dd.mov

marcelpinto:
@yjbanov it would be great to get an update on this based on your previous comment

darshankawar:
[...]

PR #179712, opened 2025-12-10
 [web] Fix touch scroll propagation when Flutter is embedded in iframe

flutter-zl:
Fixes touch scrolling not propagating to the host page when Flutter web is embedded in an iframe(#157435).

**Root cause**
Engine calls preventDefault() on all touch events, blocking native browser scroll.

**Fix**
Skip preventDefault() for touch events when in an iframe
Detect overscroll at scroll boundaries and send to engine via flutter/scroll channel
Engine calls parent.scrollBy() to scroll the host page

Before change
https://issue-157435-before.web.app/

After change
https://issue-157435-after.web.app/


ManexpDev:
Is there something I'm not seeing... or do those two examples work exactly the same? I tested it on mobile emulation and on a real mobile.

PR #179718, opened 2025-12-11
Remove obsolete windowing channel

robert-ancell:
This was an older method used for multi-window and has since been replaced with FFI.


mattkae (approved):
Yup!

PR #179756, opened 2025-12-11
Update Skwasm to engine style guidelines.

eyebrowsoffire:
This updates the Skwasm library to conform to the style of the rest of the engine. This includes:
* Moving the location of the skwasm library into `engine/src/flutter/skwasm` instead of `engine/src/flutter/lib/web_ui/skwasm`, since that was altogether too nested and weird.
* Changed all the file extensions to `.cc` instead of `.cpp`
* Changed all local include paths to be from the `engine/src/flutter` directory rather than relying on the source file's location.
* Changed ordering of include paths.
* Removed instances of `using namespace`
* Changed local variable names and argument names to use snake_case
* Changed class functions to use UpperCamelCase
* Changed private member variables to use trailing_ underscore instead of _leading.
* Removed some uses of `auto`


engine/src/flutter/skwasm/canvas_text.h
```diff
@@ -2,16 +2,16 @@
...
+
+#include <memory>
 
 #include "flutter/display_list/dl_text.h"
 #include "third_party/skia/include/core/SkRefCnt.h"
 
-#include <memory>
-
 namespace flutter {
 std::shared_ptr<DlText> textFromBlob(const sk_sp<SkTextBlob>& blob);
```
gaaclarke: This should be TextFromBlob too

gaaclarke (approved):
lgtm!  I don't know if we got everything but this is definitely a huge step to getting us there.  We can address anything else as they come up in code reviews.  Thanks jackson, I hope the bot was able to do a lot of the heavy lifting.

PR #179757, opened 2025-12-11
Implement popup windows in the API and test code

mattkae:
## What's new?
- Add `PopupWindowController`, `PopupWindowControllerDelegate`, and `PopupWindow`
- Implement reference logic for popup windows in the test bindings
- Add tests for the popup windows API

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -763,6 +763,141 @@ abstract class TooltipWindowController extends BaseWindowController {
...
+/// controller to release the resources associated with the window.
+///
+/// {@tool dartpad}
+/// An example usage of [PopupWindowController] looks like:
+///
+/// ** See code in examples/api/lib/widgets/windows/popup.0.dart **
+/// {@end-tool}
+///
+/// Children of a [PopupWindow] widget can access the [PopupWindowController]
+/// via the [WindowScope] inherited widget.
```
justinmc: Nit: "inherited widget" => "[InheritedWidget]" unless we're overdoing it with the links already.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -763,6 +763,141 @@ abstract class TooltipWindowController extends BaseWindowController {
...
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  bool get isActivated;
+
+  /// Requests that the window receive focus.
+  ///
+  /// The platform may also give the window input focus and bring it to the
+  /// top of the window stack. However, this behavior is platform-dependent.
+  ///
+  /// The window will be brought to the top of the window stack.
```
justinmc: Nit: This final sentence seems to duplicate or even contradict information in the previous paragraph and so it confuses me a bit. If I understand correctly, I would write:

Requests that the window receive focus.

The window will also be brought to the top of the window stack.

The platform may also give the window input focus. However, this behavior is platform-dependent.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1102,6 +1264,68 @@ class TooltipWindow extends StatelessWidget {
...
+/// the widget. The [child] widget is rendered into this newly created window.
+///
+/// When a [PopupWindow] widget is removed from the tree, the window that was created
+/// by the [controller] remains valid until the caller destroys it by calling
+/// [PopupWindowController.destroy].
+///
+/// Widgets in the same tree as the [child] widget will have access to the
+/// [PopupWindowController] via the [WindowScope] widget.
+///
+/// {@tool dartpad}
```
justinmc: This will run the example live in Dartpad on the docs website. I'm guessing this is not useful because it won't work on the web right? So I would change this to `@tool snippet`.

packages/flutter/test/widgets/windowing_test.dart
```diff
@@ -263,6 +301,24 @@ void main() {
...
+        await tester.pumpWidget(
+          wrapWithView: false,
+          PopupWindow(
+            controller: controller,
+            child: Builder(
+              builder: (BuildContext context) {
+                final BaseWindowController scope = WindowScope.of(context);
+                expect(scope, isA<PopupWindowController>());
+                return const SizedBox.shrink();
+              },
```
justinmc, line 322: Would it be better to put the `expect` after `pumpWidget` and declare `scope` before it? That would make me feel better that the expect can't be skipped if somehow the `builder` never runs.

Same for all of the similar tests below.

packages/flutter/test/widgets/windowing_test.dart
```diff
@@ -1038,6 +1330,26 @@ void main() {
...
+            child: Builder(
+              builder: (BuildContext context) {
+                final bool? isFullscreen = WindowScope.maybeIsFullscreenOf(context);
+                expect(isFullscreen, equals(false));
+                return const SizedBox.shrink();
+              },
+            ),
+          ),
+        );
+      });
```
justinmc, line 1464: Thanks for testing all of the different methods here.

justinmc (approved):
LGTM 👍

packages/flutter/lib/src/widgets/_window.dart
mattkae: Should be removed :)

PR #179768, opened 2025-12-12
[Impeller] Delete GLES framebuffer objects only on the thread where they were created

### LINKED ISSUE #178276: TextShadowCache leaks memory on OpenGL ES.
### Steps to reproduce

You can reproduce the problem by simply running the code above, but it requires certain devices, such as the Galaxy A51 I found.

### Expected results

The code above causes a continuous increase in memory usage on some Android phones. The model I've currently observed is a Galaxy A51 running Android 10, using the Impeller engine. The core issue is that the `blurRadius` property in the shadow and the loading of the GIF image, when done together, cause a bug that leads to continuous memory usage growth. These two conditions are necessary, but not every phone experiences this problem. I haven't found the same issue on iOS. I suspect that the `blurRadius` property is not supported on some models, even though the rendering is the same as for the GIF.

### Actual results

Memory continues to grow

### Code sample
```dart
Column(
      children: [
        SizedBox(height: 100,),
        Text(
          textAlign: TextAlign.center,
          'asjdhjasdh',
          style: TextStyle(
            color: Colors.white,
            fontSize: 25.fitWithScreenW,
            shadows: [
              Shadow(
                color: Color(0xff111111),
                offset: const Offset(0, 1), // 阴影的偏移量，(dx, dy)
                blurRadius: 2.0, // 模糊半径
              ),
            ],
          ),
        ),
        Image.network(
            'https://media4.giphy.com/media/v1.Y2lkPTc5MGI3NjExbzhrbXg4dDRxdjEwNHptNHRncjJ4b280amh1bnlpOGd4OThyaXFzYiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/JWnXY237vWeX3zx64V/giphy.gif',
            width: 72,
            height: 72),

      ],
    )
```

### Screenshots or Video

Image

Image

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

[!] Flutter (Channel [user-branch], 3.32.8, on macOS 14.5 23F79 darwin-arm64,
    locale zh-Hans-CN)
    ! Flutter version 3.32.8 on channel [user-branch] at
      /Users/liwei/flutter/flutter332
      Currently on an unknown channel. Run `flutter channel` to switch to an
      official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions
      at https://flutter.dev/setup.
    ! Upstream repository unknown source is not a standard remote.
      Set environment variable "FLUTTER_GIT_URL" to unknown source to dismiss
      this error.
[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor
      --android-licenses
[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
[✓] Android Studio (version 2025.2)
[✓] VS Code (version 1.105.1)
[✓] Network resources

jason-simmons:
OpenGL does not allow cross-context sharing of container objects (such as framebuffers) that hold references to other objects.  ReactorGLES must ensure that framebuffers are deleted using the thread and context where the object was created.

Fixes https://github.com/flutter/flutter/issues/178276

### DIFF
--- a/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
+++ b/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
@@ -308,6 +308,9 @@ bool ReactorGLES::ConsolidateHandles() {
     for (auto& handle : handles_) {
       // Collect dead handles.
       if (handle.second.pending_collection) {
+        // Some objects can only be deleted on a specific thread.  Collection
+        // of these objects will be deferred until the owner thead executes
+        // a reaction and triggers collection of handles.
         const auto& owner_thread = handle.first.owner_thread_;
         if (owner_thread.has_value() && *owner_thread != current_thread) {
           continue;
--- a/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
+++ b/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
@@ -296,6 +296,7 @@ bool ReactorGLES::ReactOnce() {
 bool ReactorGLES::ConsolidateHandles() {
   TRACE_EVENT0("impeller", __FUNCTION__);
   const auto& gl = GetProcTable();
+  std::thread::id current_thread = std::this_thread::get_id();
   std::vector<std::tuple<HandleGLES, std::optional<GLStorage>>>
       handles_to_delete;
   std::vector<std::tuple<DebugResourceType, GLint, std::string>>
@@ -307,6 +308,10 @@ bool ReactorGLES::ConsolidateHandles() {
     for (auto& handle : handles_) {
       // Collect dead handles.
       if (handle.second.pending_collection) {
+        const auto& owner_thread = handle.first.owner_thread_;
+        if (owner_thread.has_value() && *owner_thread != current_thread) {
+          continue;
+        }
         handles_to_delete.emplace_back(
             std::make_tuple(handle.first, handle.second.name));
         continue;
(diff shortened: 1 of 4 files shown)


engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
```diff
@@ -307,6 +308,10 @@ bool ReactorGLES::ConsolidateHandles() {
     for (auto& handle : handles_) {
       // Collect dead handles.
       if (handle.second.pending_collection) {
+        const auto& owner_thread = handle.first.owner_thread_;
+        if (owner_thread.has_value() && *owner_thread != current_thread) {
```
chinmaygarde, line 315: Couldn't this just defer collection indefinitely?

chinmaygarde (commented):
I am not able to reason about when the framebuffer object will be collected if the reactor never gets to consolidate handles on the right thread?

Perhaps we should remove support for framebuffers from the reactor altogether?

engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
jason-simmons: The framebuffer object will be deleted when the owner thread eventually calls `ReactorGLES::ConsolidateHandles`.

The owner thread called a `ReactorGLES` API to create the framebuffer handle.  Presumably that thread is also queueing `ReactorGLES` operations and will make calls to `React`, which will do a `ConsolidateHandles`.

Until then, any `ConsolidateHandles` calls that run on other threads will leave the framebuffer in the `ReactorGLES::handles_` set.

(`ReactorGLES` object deletion inherently relies on the user calling `React`.   Calling `CollectHandle` does not itself schedule a `ConsolidateHandles`)

chinmaygarde (approved):

Issue #179727, opened 2025-12-11
Invalid Unicode in TextEditingController's `text` Crashes the App on iOS


mahmuttaskiran:
### Steps to reproduce

App crashes on **iOS** when a `TextField` contains malformed Unicode (e.g., half of a surrogate pair such as `\u{DFFF}`).

### Expected results

No crash

### Actual results

Crash

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      _counter++;
    });
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text('You have pushed the button this many times:'),
            Text(
              '$_counter',
              style: Theme.of(context).textTheme.headlineMedium,
            ),
            TextField(
              controller: TextEditingController(text: '\u{DFFF}'),
              decoration: const InputDecoration(
                border: OutlineInputBorder(),
                labelText: 'Input Text',
                helperText: 'Value: \\u{DFFF}',
              ),
            ),
          ],
        ),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ), // This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}

```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[...]

PR #179775, opened 2025-12-12
Fix crash in FlutterJSONMessageCodec when decoding invalid Unicode

mahmuttaskiran:
Fix crash in FlutterJSONMessageCodec when decoding invalid Unicode

Added a safety check in `decodeMethodCall` to handle cases where JSON decoding returns nil. This prevents the app from crashing when invalid UTF-8 is passed into the codec.

Fixing https://github.com/flutter/flutter/issues/179727 

## Pre-launch Checklist

- [x] I signed the [CLA].

[talabat.com](https://www.talabat.com) [Talabat Flutter PRs](https://github.com/search?q=org%3Aflutter+talabat&type=pullrequests)


chinmaygarde:
This changes existing behavior and existing tests verify this. Seems like throwing an exception is WAI.

mahmuttaskiran:
> Hey there, I left some more details on this crash here: [#179727 (comment)](https://github.com/flutter/flutter/issues/179727#issuecomment-3663338710)
> 
> Instead of not crashing on invalid Unicode, it seems we should update the iOS embedder to accept Dart's strategy for escaping lone surrogate pairs.

Thank you for the details on the issue. I’ve raised another PR that attempts to resolve it by introducing a sanitizer. I’ve validated that both Android and iOS now behave the same.

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecs.mm
```diff
@@ -143,6 +146,10 @@ - (NSData*)encodeErrorEnvelope:(FlutterError*)error {
 
 - (FlutterMethodCall*)decodeMethodCall:(NSData*)message {
   NSDictionary* dictionary = [[FlutterJSONMessageCodec sharedInstance] decode:message];
+  if (dictionary == nil) {
+    NSLog(@"Invalid JSON method call. Couldn't decode message: %@", message);
```
LongCatIsLooong, line 151: In FluterCodecs.h there's `NS_ASSUME_NONNULL_BEGIN` so I'm expecting the return type to be non-nullable, I'm a bit surprised that the compiler isn't complaining about this.

LongCatIsLooong: The NSLog will be printing out a `NSData`, not sure what `-[NSData description]` does, but if it doesn't print out the raw data then it's basically useless, if it does then it may expose sensitive information like PII or passwords in standard output.

mahmuttaskiran: Thank you for the review @LongCatIsLooong 



> In FluterCodecs.h there's `NS_ASSUME_NONNULL_BEGIN` so I'm expecting the return type to be non-nullable, I'm a bit surprised that the compiler isn't complaining about this.

"-Wnullable-to-nonnull-conversion" is missing in compiler configs for iOS and Mac in 
`common/confic.gni`


After enabling it, compiler gives error for the following files:

```
../../flutter/shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm:40:36: error: implicit conversion from nullable pointer 'NSURL * _Nullable' to non-nullable pointer type 'NSURL * _Nonnull' [-Werror,-Wnullable-to-nonnull-conversion]

AND 

../../flutter/shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm:49:65: error: implicit conversion from nullable pointer 'NSData *__strong _Nullable' to non-nullable pointer type 'NSData * _Nonnull' [-Werror,-Wnullable-to-nonnull-conversion]

```

I'm not sure if it's disabled intentionally. So, I'm leaving it as is


> The NSLog will be printing out a `NSData`, not sure what `-[NSData description]` does, but if it doesn't print out the raw data then it's basically useless, if it does then it may expose sensitive information like PII or passwords in standard output.

True. Added debug-only preprocessor guard there.

PR #179776, opened 2025-12-12
refactor: migrate CupertinoPageTransitionsBuilder to cupertino folder

rkishan516:
Changes:

* Move CupertinoPageTransitionsBuilder from material/page_transitions_theme.dart to cupertino/route.dart

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/cupertino/route.dart
+++ b/packages/flutter/lib/src/cupertino/route.dart
@@ -1549,19 +1549,39 @@ class CupertinoDialogRoute<T> extends RawDialogRoute<T> {
   static final Tween<double> _dialogScaleTween = Tween<double>(begin: 1.3, end: 1.0);
 }
 
-/// Used by [PageTransitionsTheme] to define a horizontal [MaterialPageRoute]
-/// page transition animation that matches native iOS page transitions.
+/// A [PageTransitionsBuilder] that provides an iOS-style page transition
+/// animation.
+///
+/// The page slides in from the right and exits in reverse. It also shifts
+/// to the left in a parallax motion when another page enters to cover it.
+/// This transition is commonly seen in native iOS applications.
+///
+/// This builder can be used with [PageTransitionsTheme] to apply iOS-style
+/// transitions to [MaterialPageRoute] or any [PageRoute] that uses a
+/// [PageTransitionsTheme].
+///
+/// In a [CupertinoApp], this transition is used automatically when navigating
+/// with [CupertinoPageRoute]:
+///
+/// ```dart
+/// Navigator.push(
+///   context,
+///   CupertinoPageRoute<void>(
+///     builder: (BuildContext context) => const MyPage(),
+///   ),
+/// );
+/// ```
 ///
 /// See also:
 ///
-///  * [FadeUpwardsPageTransitionsBuilder], which defines a page transition
-///    that's similar to the one provided by Android O.
-///  * [OpenUpwardsPageTransitionsBuilder], which defines a page transition
-///    that's similar to the one provided by Android P.
-///  * [ZoomPageTransitionsBuilder], which defines the default page transition
-///    that's similar to the one provided in Android Q.
-///  * [PredictiveBackPageTransitionsBuilder], which defines a page
-///    transition that allows peeking behind the current route on Android.
+///  * [CupertinoPageRoute], which uses this transition style by default for
+///    Cupertino apps.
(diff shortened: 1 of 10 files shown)


victorsanni (commented):
Can we add a test here?

packages/flutter/lib/src/cupertino/route.dart
```diff
@@ -1548,3 +1548,45 @@ class CupertinoDialogRoute<T> extends RawDialogRoute<T> {
   // transitions.
   static final Tween<double> _dialogScaleTween = Tween<double>(begin: 1.3, end: 1.0);
 }
+
+/// Used by [PageTransitionsTheme] to define a horizontal [MaterialPageRoute]
+/// page transition animation that matches native iOS page transitions.
```
victorsanni: Can we have a more non-Material oriented + Cupertino/iOS heavy documentation here? Then add a `See also` section that mentions MaterialPageRoute and PageTransitionsTheme etc.

packages/flutter/test/cupertino/route_test.dart
```diff
@@ -3188,6 +3188,85 @@ void main() {
...
+    });
+
+    testWidgets('has delegatedTransition', (WidgetTester tester) async {
+      const builder = CupertinoPageTransitionsBuilder();
+
+      expect(builder.delegatedTransition, isNotNull);
+      expect(builder.delegatedTransition, CupertinoPageTransition.delegatedTransition);
+    });
+
+    testWidgets('works with MaterialApp and PageTransitionsTheme', (WidgetTester tester) async {
```
victorsanni: Maybe this test can live in Material instead?

packages/flutter/lib/src/cupertino/route.dart
rkishan516: Updated.

packages/flutter/test/cupertino/route_test.dart
rkishan516: I think its duplicated, we can totally remove this. Material already have this version.

packages/flutter/lib/src/cupertino/route.dart
```diff
@@ -1548,3 +1548,56 @@ class CupertinoDialogRoute<T> extends RawDialogRoute<T> {
...
+/// A [PageTransitionsBuilder] that provides an iOS-style page transition
+/// animation.
+///
+/// The page slides in from the right and exits in reverse. It also shifts
+/// to the left in a parallax motion when another page enters to cover it.
+/// This transition is commonly seen in native iOS applications.
+///
+/// This builder can be used with [PageTransitionsTheme] to apply iOS-style
+/// transitions to [MaterialPageRoute] or any [PageRoute] that uses a
+/// [PageTransitionsTheme].
```
victorsanni: I don't know that we want to keep this paragraph, saying this through the "See also" section should be enough to compartmentalize.

victorsanni (approved):

packages/flutter/lib/src/cupertino/route.dart
rkishan516: I removed this, because it makes sense to not keep anything related to material specific inside cupertino package.

victorsanni:
Google tests are failing due to missing cupertino imports.

victorsanni:
This [file](https://github.com/flutter/codelabs/blob/main/boring_to_beautiful/step_03/lib/src/shared/providers/theme.dart) will need a cupertino import if this PR lands.

Can we add a one-line dart fix here @Piinks? Considering the design libraries might be decoupling anyways. I think this PR warrants a migration guide at least.

rkishan516:
> Hmm, I am not sure if dart fix can do this, but I've reached out to see if they can. Otherwise we should be able to add it manually.
> I think a simple migration guide is a good idea. It can cover all the things we moved around wrt page transitions and explain why. :)

I will create a simple migration guide for this.

Piinks:
Thanks for the migration guide as well!

victorsanni (approved):

victorsanni:
Google testing is passing now, just need to resolve some internal hitches that are preventing it from being reflected on the Github UI.

Piinks (approved):
LGTM! This might need a rebase to land now?

victorsanni:
> Rebased 👍 I will confirm with Victor that the Google testing failures are ok before popping on the autosubmit. :)

Should be ok - I added the cupertino library import to relevant files (and a comment for the linter to ignore since it is currently unused).

PR #179817, opened 2025-12-12
Add table detailing AI rule file character limits for various tools

jwren:


docs/rules/README.md
```diff
@@ -2,3 +2,20 @@
...
+| **Gemini CLI** | Input Context | **1M+ Tokens** (Context) | Limited by model context (approx 700k words). Practical per-file limits (e.g. 20MB) may apply.<br>**Source:** Google Cloud: Gemini Models |
+| **GitHub Copilot** | **Chat** Instructions | **~2 Pages** (Soft) | "We recommend... no longer than 2 pages."<br>**Source:** GitHub Docs: Custom Instructions |
+| **GitHub Copilot** | **Code Review** Instructions | **4,000 characters** (Hard) | "Instructions... limited to 4000 characters."<br>**Source:** GitHub Docs: Custom Instructions |
+| **Goose** | `.goosehints` / `AGENTS.md` | **No Hard Limit** | Limited by model context. Auto-compacts conversation when full.<br>**Source:** Block Goose: Context Management |
+| **JetBrains AI** | `.aiassistant/rules` | **No Hard Limit** | Limited only by the model's context window (prompts are trimmed if too large).<br>**Source:** JetBrains Blog: AI Assistant Update |
+| **OpenAI** (ChatGPT) | Custom Instructions | **1,500 characters** (Hard) | "Instructions... 1500 character limit."<br>**Source:** OpenAI Help: Custom Instructions |
```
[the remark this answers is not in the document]
jakemac53: agreed - not only would it help to know where these numbers come from but also it would give us a place to go when we inevitably need to update this

docs/rules/README.md
```diff
@@ -1,4 +1,25 @@
 # AI rules for Flutter
 
-This directory contains `rules.md`, the default set of AI rules for building
-Flutter apps, following best-practices.
+This directory contains the default set of AI rules for building Flutter apps, following best practices.
+
+*   `rules.md`: The comprehensive master rule set.
+*   `rules_10k.md`: A condensed version (<10k chars) for tools with stricter context limits.
```
jakemac53, line 6: Note that we really want tokens not chars - which should give us 3-4x more to work with.

There is a gemini API to count the number of tokens in a document, we could probably write a small tool to call that API and commit it next to these so we can count tokens for these files concretely.

jakemac53: Actually nevermind, i guess these are in terms of characters, which is odd but 🤷

docs/rules/rules_10k.md
```diff
@@ -0,0 +1,174 @@
+# AI Rules for Flutter (Comprehensive 10k)
```
jakemac53: I would leave out the size info from the file, thats not useful for the agent I don't think

```suggestion
# AI Rules for Flutter
```

jwren: Removed

jwren:
URLs updated.

docs/rules/README.md
```diff
@@ -1,4 +1,25 @@
...
+*   `rules_1k.md`: An ultra-compact version (<1k chars) for very strict limits.
+
+## Device & Editor Specific Limits
+
+Different AI coding assistants and tools have varying limits for their "rules" or "custom instructions" files. *Last updated: 2026-01-05.*
+
+| Tool / Product | Rules File / Feature | **Soft / Hard Limit** | Notes & Sources |
+| Tool / Product | Limit | Source | Notes |
+| :--- | :--- | :--- | :--- |
+| Aider | No Hard Limit | [Aider Conventions](https://aider.chat/docs/usage/conventions.html) | Limited by model context window. |
```
jakemac53, line 17: What about Claude Code? I see a lot of agents I have never heard of in this list but we are missing one of the most popular ones :)

jwren: Added

docs/rules/README.md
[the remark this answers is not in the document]
jwren: URLs updated

jakemac53 (approved):
LGTM with the caveat that I don't really have any flutter expertise

johnpryan (approved):
LGTM

Issue #178894, opened 2025-11-21
[web, tool] Look into public packages blocking wasm builds and reporting them in analytics


kevmoo:
Product tracking issue...

srujzs:
> maybe at some point this included more like invalid JS interop libraries, but not anymore

Yeah, I think when we fully removed `package:js` and `dart:js_util` support from dart2wasm, we removed these from the interop checks because the CFE would already report the errors. This had the side-effect of affecting this dry-run. It's better anyways that we explicitly check for those imports in the dry run instead of relying on the JS interop checks.

srujzs:
I can add the extra errors for the two imports and it doesn't look like Flutter needs to be updated after that as it just pipes the error code directly through.

PR #179826, opened 2025-12-13
Add package info to wasm dry run events.

biggs0125:
Uses new event structure to capture package info from wasm dry run errors.

Uses the `package_info.json` to gather the set of public packages and only includes package names when the package is in that set of publicly available packages.

Prepends "-p" for when an error type is found in a private package (i.e. a `package:*` that isn't in the set of public packages), a "-h" for non-public packages (i.e. any non-"package" packages or otherwise unrecognized uris), or a "-ph" for both.

Adds a special "error" entry when the package_config can't be loaded and instead includes the findings in the same format as today (comma-separated codes).

Also adds tests for dry runs.

See https://github.com/flutter/flutter/issues/178894 for more context.


packages/flutter_tools/lib/src/build_system/targets/web.dart
```diff
@@ -438,16 +438,89 @@ class Dart2WasmTarget extends Dart2WebTarget {
...
+          ),
+        );
+        return;
+      }
+
+      final Map<String, String> hostedPackages = {};
+      final Set<String> privatePackages = {};
+      for (final Package package in packageConfigPackages.packages) {
+        final String packageName = package.name;
+        if (package.root.toString().contains('hosted/pub.dev')) {
```
srujzs, line 479: Is there no way to get the package's description and source fields instead per #178894? I also wonder if there's a way to avoid the regex search below to get the version. Maybe this is stored in `extraData`?

packages/flutter_tools/lib/src/build_system/targets/web.dart
```diff
@@ -438,16 +438,89 @@ class Dart2WasmTarget extends Dart2WebTarget {
...
+        }
+      }
+
+      final PackageConfig packageConfigPackages;
+      try {
+        packageConfigPackages = await loadPackageConfigWithLogging(
+          findPackageConfigFileOrDefault(environment.projectDir),
+          logger: environment.logger,
+        );
+      } on Exception {
```
kevmoo: Do you know a more specific `Exception` type?

kevmoo: Are you sure it's not an `Error`? I often find these things tricky

packages/flutter_tools/lib/src/build_system/targets/web.dart
```diff
@@ -438,16 +438,89 @@ class Dart2WasmTarget extends Dart2WebTarget {
         'https://docs.flutter.dev/platform-integration/web/wasm\n',
       );
       result = 'findings';
-      findingsSummary = RegExp(
-        r'\(([0-9]+)\)',
-      ).allMatches(stdout).map((RegExpMatch f) => f.group(1)).join(',');
-    } else {
-      result = 'unknown';
+      final Map<String, Set<Uri>> importUriToErrorCode = {};
```
srujzs: nit: `errorCodeToImportUris`

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: Done, made this more specific

packages/flutter_tools/lib/src/build_system/targets/web.dart
```diff
@@ -438,16 +438,89 @@ class Dart2WasmTarget extends Dart2WebTarget {
...
+              continue;
+            } else if (privatePackages.contains(packageName)) {
+              privatePackage = true;
+              continue;
+            }
+          }
+          hostApp = true;
+        }
+        final String? hpHint = switch ((hostApp, privatePackage)) {
+          (true, true) => '-hp',
```
kevmoo, line 510: Do we care about if the `host` is private? A bit duplicative, right?

srujzs, line 510: nit: I wonder if `-happ`, `-ha`, and `-pp` are slightly more descriptive or that makes the flag syntax more confusing.

packages/flutter_tools/lib/src/build_system/targets/web.dart
```diff
@@ -438,16 +438,89 @@ class Dart2WasmTarget extends Dart2WebTarget {
...
+          (true, false) => '-h',
+          (false, true) => '-p',
+          _ => null,
+        };
+
+        final String findings = [?hpHint, ...hostedPackageFindings].join(',');
```
kevmoo: Here's my crazy idea.

A `do...while` loop. While the value is > 100 length, randomly remove a item from `hostedPackageFindings` so (1) we avoid uploading part of a package name, which would be bad and (2) we avoid over sampling packages at the beginning.

packages/flutter_tools/test/general.shard/build_system/targets/web_dry_run_test.dart
```diff
@@ -0,0 +1,370 @@
...
+    () => testbed.run(() async {
+      processManager.addCommand(
+        FakeCommand(
+          command: commandArgs,
+          exitCode: 254,
+          stdout: '''
+Found incompatibilities with WebAssembly.
+
+package:foo/some/path.dart 6:1 - dart:html unsupported (0)
+package:bar/some/path.dart 8:10 - package:hs unsupported (2)
```
srujzs: nit: package:js

packages/flutter_tools/test/general.shard/build_system/targets/web_dry_run_test.dart
```diff
@@ -0,0 +1,370 @@
...
+      final Event event = fakeAnalytics.sentEvents[0];
+      expect(event.eventName, equals(DashEvent.flutterWasmDryRunPackage));
+      expect(event.eventData, hasLength(4));
+      expect(event.eventData['result'], 'findings');
+      expect(event.eventData['exitCode'], 254);
+      expect(event.eventData['error'], 'packageConfigNotLoaded');
+      expect(event.eventData['findings'], '0,2');
+    }),
+  );
+}
```
srujzs, line 480: nit: maybe test that findings are always truncated to 100 chars?

srujzs (approved):

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: Yeah technically seeing a private package is the same in terms of action we can take as seen host app code. But it could tell us something about how people are laying out their code. I don't mind the single extra character here.

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: Okay I went with some randomization and fitting of elements. Thought about it more and you're right for the data processing this is going to help.

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: There doesn't seem to be a general purpose way to get the `pubspec.lock` file from within flutter tools. There's some logic to write it but I don't see logic to read it and implementing that parser seems heavyweight for this. So I had to go with reading the `package_config.json` which instead follows this pattern.

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: Good catch. Switched the map without switching the name.

packages/flutter_tools/lib/src/build_system/targets/web.dart
biggs0125: Unless you have the context of what these mean I don't think the extra character does much. So I'd prefer to keep it shorter.

packages/flutter_tools/test/general.shard/build_system/targets/web_dry_run_test.dart
biggs0125: Done.

packages/flutter_tools/test/general.shard/build_system/targets/web_dry_run_test.dart
biggs0125: Done.

PR #179841, opened 2025-12-13
Switch to js_interop for web import configuration

srawlins:
Work towards https://github.com/dart-lang/sdk/issues/61875

All conditional web imports should now use `dart.library.js_interop`, as per https://dart.dev/tools/pub/create-packages#conditionally-importing-and-exporting-library-files and https://github.com/dart-lang/sdk/issues/61875.


Piinks:

Piinks:
This came up again in stale triage, should we leave it open?

kevmoo:
Actually, this is by design since we WANT to test for JS or not.

PR #179859, opened 2025-12-14
Allow windows to be created undecorated

### LINKED ISSUE #183557: Add API to enable undecorated windows
Add an API to Flutter to enable windows to be created undecorated.

robert-ancell:
This will be useful when we want to render all the window decorations using Flutter.

Fixes https://github.com/flutter/flutter/issues/183557


packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -673,6 +688,9 @@ class RegularWindowControllerLinux extends RegularWindowController {
     if (title != null) {
       setTitle(title);
     }
+    if (!decorated) {
+      _window.setDecorated(false);
+    }
```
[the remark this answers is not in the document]
robert-ancell: I did this to avoid the FFI call, I do think it is more readable as suggested though, so I'll change to that as the overhead should be insignificant.

engine/src/flutter/shell/platform/windows/host_window_dialog.cc
```diff
@@ -37,15 +37,16 @@ HostWindowDialog::HostWindowDialog(WindowManager* window_manager,
                                    const WindowSizeRequest& preferred_size,
                                    const BoxConstraints& constraints,
                                    LPCWSTR title,
+                                   bool decorated,
                                    std::optional<HWND> const& owner_window)
     : HostWindow(window_manager, engine) {
   InitializeFlutterView(HostWindowInitializationParams{
       .archetype = WindowArchetype::kDialog,
-      .window_style = GetWindowStyleForDialog(owner_window),
+      .window_style = decorated ? GetWindowStyleForDialog(owner_window) : 0,
```
mattkae: Can we do something like this instead? This way, we can reason about all of the decorations in one place

```suggestion
      .window_style = GetWindowStyleForDialog(owner_window, decorated),
```

mattkae (approved):
Nice stuff! I just have one thing, but it looks good other than that.

We _may_ want to be able to turn decorations on and off _after_ a window is created, but we can wait on that. I don't think that this is necessary for V1, and it would be easy to implement.

mattkae:
Going to test this  this morning

mattkae (changes requested):
I will try and fix on win32

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -362,8 +366,13 @@ class RegularWindowControllerMacOS extends RegularWindowController with _WindowC
     required Size? preferredSize,
     BoxConstraints? preferredConstraints,
     String? title,
+    bool decorated = true,
   }) : _delegate = delegate,
        super.empty() {
+    if (!decorated) {
+      throw UnimplementedError('Undecorated windows are not yet implemented on MacOS.');
```
loic-sharma: Nit: please prefer `macOS` casing

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -362,8 +366,13 @@ class RegularWindowControllerMacOS extends RegularWindowController with _WindowC
     required Size? preferredSize,
     BoxConstraints? preferredConstraints,
     String? title,
+    bool decorated = true,
   }) : _delegate = delegate,
        super.empty() {
+    if (!decorated) {
```
loic-sharma, line 386: If we keep the throws here, please add TODO comments that link to a tracking issue :)

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -362,8 +366,14 @@ class RegularWindowControllerMacOS extends RegularWindowController with _WindowC
...
+      // See https://github.com/flutter/flutter/issues/183560
```
loic-sharma, line 388: I'd make this a TODO comment. Something like:

```suggestion
      // TODO(team-macos): Support disabling window decorations.
      // See https://github.com/flutter/flutter/issues/183560
```

Same comment for Windows.

mattkae (approved):
LGTM

PR #164322, opened 2025-02-27
WIP: Only repaint views that need to be painted

knopp:
Fixes https://github.com/flutter/flutter/issues/164320

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #179874, opened 2025-12-15
Only render views that need to be rendered (with external texture & platform view support)

### LINKED ISSUE #164320: [Multiview] Flutter re-renders all views during each update
Every view is re-rendered, even those without dirty render objects. Attaching a video with quartz debug visualizing screen updates.

https://github.com/user-attachments/assets/8d3c39c6-4392-4140-94ff-98d04c9b2d5a

### Known problems

Some things we need to figure out before we can do this optimization:

1. **External textures**. When an external texture changes, the framework isn’t aware of which views are dirty. Instead, the engine schedules a new frame and lets the framework re-render all views. We’ll need to update the framework to be aware of which views are affected when an external texture changes so that it re-renders just those views.
2. **Android platform views**. This works similar to external textures.
3. More?

9AZX:
Fixes https://github.com/flutter/flutter/issues/164320
Based on @knopp work in https://github.com/flutter/flutter/pull/164322

This PR adds a dirty-view compositing optimization that skips compositing for views that don't have any dirty render objects, so it should improves performance in multi-view scenarios by avoiding unnessacery work.

When a texture has new content, the framework now gets notified via a new onTextureFrameAvailable callback from the engine. This marks the texture's render object as needing paint, so its view is included in compositing even when nothing else changed.

On Android, texture callbacks now use `markTextureFrameAvailable` instead of `scheduleEngineFrame`. This is actually more efficient since it reuses the existing layer tree rather than regenerating it. Other platforms already used the correct path.

It also forces all views to composite when returning from background, ensuring the display is properly refreshed after the app was paused.

On MacOS:

https://github.com/user-attachments/assets/9d8b21b5-0c61-40d9-aadc-8c5a4b76f20c

## Pre-launch Checklist

- [X] I signed the [CLA].


gaaclarke (commented):
I have some misgivings about adding complexity to multi-view support since I'm not sure where we are in its development cycle.  I know, for example, the dirty rect optimization had to have a flag to turn it off since it made debugging and profiling difficult in certain situations.

@loic-sharma do you have thoughts about the tradeoff of adding this extra complexity to multi-view's right now?  I think this is something we want eventually, but does it make sense to do now?  Is this unlocking real world use-cases?

loic-sharma:
> ... multi-view support since I'm not sure where we are in its development cycle

The project is very much active, Canonical has been making great progress productionizing this feature! :)

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,6 +262,13 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
     assert(_rootTransform != null);
   }
 
+  /// Whether [prepareInitialFrame] has been called on this render view.
+  ///
+  /// This is true after the first call to [prepareInitialFrame], which sets up
+  /// the initial layout and paint for the view. Before this is true, the view
+  /// cannot be composited.
+  bool get hasInitialFrameBeenPrepared => _rootTransform != null;
```
gaaclarke: no action required: This is inferring this state from a secondary property.  I wonder if it makes sense to have a bool just for this.

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -536,6 +536,23 @@ class PlatformConfiguration final {
...
+  ///             this to mark the corresponding texture render object as
+  ///             needing paint, ensuring the view containing the texture is
+  ///             recomposited even if no other render objects are dirty.
+  ///
+  ///             This method calls the `::_textureFrameAvailable` method in
+  ///             `hooks.dart`.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  void TextureFrameAvailable(int64_t texture_id);
```
gaaclarke: There is no verb in the method and it has a void return type.  Please include a verb in the name, `NotifyTextureFrameAvailable()`?

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -395,6 +395,11 @@ void _drawFrame() {
   PlatformDispatcher.instance._drawFrame();
 }
 
+@pragma('vm:entry-point')
+void _textureFrameAvailable(int textureId) {
```
gaaclarke: Same thing about naming.

engine/src/flutter/runtime/runtime_controller.h
```diff
@@ -373,6 +373,21 @@ class RuntimeController : public PlatformConfigurationClient,
...
+  ///             This is called by the shell when the platform marks a texture
+  ///             as having new content. The framework uses this to mark the
+  ///             corresponding texture render object as needing paint.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  /// @return     Returns if the notification to the running isolate was
+  ///             successful. Returns false if the root isolate is not running.
+  ///
+  bool TextureFrameAvailable(int64_t texture_id);
```
gaaclarke: Same thing on naming.

engine/src/flutter/runtime/runtime_controller.h
```diff
@@ -373,6 +373,21 @@ class RuntimeController : public PlatformConfigurationClient,
...
+  ///             available.
+  ///
+  ///             This is called by the shell when the platform marks a texture
+  ///             as having new content. The framework uses this to mark the
+  ///             corresponding texture render object as needing paint.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  /// @return     Returns if the notification to the running isolate was
+  ///             successful. Returns false if the root isolate is not running.
```
gaaclarke, line 389: I would prefer we use absl::Status instead of interpreting bool values.

engine/src/flutter/shell/common/engine.h
```diff
@@ -652,6 +652,20 @@ class Engine final : public RuntimeDelegate, PointerDataDispatcher::Delegate {
...
+  ///
+  ///             This is called when the platform marks a texture as having new
+  ///             content via `MarkTextureFrameAvailable`. The framework uses
+  ///             this to mark the corresponding texture render object as
+  ///             needing paint, ensuring the view containing the texture is
+  ///             recomposited even if no other render objects are dirty.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  void TextureFrameAvailable(int64_t texture_id);
```
gaaclarke: Same thing on naming.

engine/src/flutter/shell/common/shell.cc
```diff
@@ -1329,13 +1329,19 @@ void Shell::OnPlatformViewMarkTextureFrameAvailable(int64_t texture_id) {
...
+  // This marks the texture render object as needing paint, ensuring the view
+  // containing the texture is recomposited even if no other render objects
+  // are dirty. Also schedule a new frame without having to rebuild the layer
+  // tree.
+  fml::TaskRunner::RunNowOrPostTask(
+      task_runners_.GetUITaskRunner(),
+      [engine = engine_->GetWeakPtr(), texture_id]() {
+        if (engine) {
+          engine->TextureFrameAvailable(texture_id);
+          engine->ScheduleFrame(false);
```
gaaclarke: Please add a parameter name comment to the rvalue argument (ex `/*foo=*/`).

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/renderer/FlutterRenderer.java
```diff
@@ -889,7 +898,11 @@ public void scheduleFrame() {
         }
         lastScheduleTime = now;
       }
-      scheduleEngineFrame();
+      // Mark this specific texture as having a new frame available.
+      // This notifies the framework to mark the texture render object as needing
+      // paint, ensuring the view containing this texture is recomposited even when
+      // no other render objects are dirty.
```
gaaclarke: I don't find these comments very helpful because they aren't contextual.  It seems we can rely on the docstring for this sort of information.

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +68,49 @@ mixin RendererBinding
     rootPipelineOwner.attach(_manifold);
   }
 
+  /// Registry mapping texture IDs to their [TextureBox] render objects.
+  ///
+  /// Used to mark textures dirty when the engine notifies us of new frames
+  /// via [handleTextureFrameAvailable].
```
gaaclarke: Should this be named "dirtyTextureRegistry" then?

gaaclarke (changes requested):
Thanks, very impressive.  I've added some comments inline.  What is the story for iOS?  I see updates to the android embedder and testing, don't we need corresponding work for iOS to work?

packages/flutter/lib/src/rendering/view.dart
9AZX: Mhhh would you prefer an explicit flag ? Like a dedicated bool `_hasInitialFrameBeenPrepared` set in `prepareInitialFrame()` ?

engine/src/flutter/runtime/runtime_controller.h
9AZX: Every other method in `runtime_controller` uses the bool value pattern, just changing this one will be inconsistent imo. But I agreed that using absl::Status would be better, I can migrate everything to another PR if that works :)

packages/flutter/lib/src/rendering/binding.dart
9AZX: Oups this comment is not clear
The registry is a lookup table, it doesn't track dirtiness. I clarified the comment

9AZX:
> Thanks, very impressive. I've added some comments inline. What is the story for iOS? I see updates to the android embedder and testing, don't we need corresponding work for iOS to work?

iOS (and macOS) already call `PlatformView::MarkTextureFrameAvailable`, which routes through `Shell::OnPlatformViewMarkTextureFrameAvailable` where the new `engine->NotifyTextureFrameAvailable()` call lives. The Android embedder was the only platform that bypassed this shared C++ path by calling `scheduleFrame()` directly

engine/src/flutter/runtime/runtime_controller.h
gaaclarke: That's fine with me, thanks.

packages/flutter/lib/src/rendering/view.dart
knopp: Personally I prefer inferring state whenever possible, lowers the chance of things getting out of sync in future when somebody modifies one part but forgets about the other.

engine/src/flutter/lib/ui/hooks.dart
```diff
@@ -395,6 +395,11 @@ void _drawFrame() {
   PlatformDispatcher.instance._drawFrame();
 }
 
+@pragma('vm:entry-point')
+void _notifyTextureFrameAvailable(int textureId) {
```
knopp, line 417: This should probably get a test inside `ui_test.dart`.

knopp (commented):
Looks great.

gaaclarke (commented):
lgtm modulo @knopp 's request for a `ui_test.dart`.

packages/flutter/lib/src/rendering/view.dart
gaaclarke: Someone could switch up the initialization of _rootTransform and accidentally change the semantics of hasInitialFrameBeenPrepared.  Deriving isn't the problem so much as "is the thing you are deriving from a tautology?"

knopp (approved):

gaaclarke (dismissed):

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +68,50 @@ mixin RendererBinding
     rootPipelineOwner.attach(_manifold);
   }
 
+  /// Registry mapping texture IDs to their [TextureBox] render objects.
+  ///
+  /// When the engine notifies us of a new frame via
+  /// [handleTextureFrameAvailable], this registry is used to look up the
+  /// corresponding [TextureBox] and mark it as needing paint.
+  @visibleForTesting
+  final Map<int, TextureBox> textureRegistry = <int, TextureBox>{};
```
loic-sharma: Are we guaranteed that a texture ID can only be used by a single `TextureBox`? From a quick glance, it looks like multiple `TextureBox`es could share the same texture ID.

knopp: Good catch. Nothing prevents you from creating multiple `Texture` widgets with for same texture. Seems like this probably should be `Map<int, Set<TextureBox>>`.

knopp (changes requested):

packages/flutter/lib/src/rendering/binding.dart
9AZX: Should be better now, thanks for the catch !

loic-sharma (dismissed):
Approving only to kick off google tests again.

knopp (dismissed):

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -635,6 +640,32 @@ class PlatformDispatcher {
   @Native<Void Function(Bool)>(symbol: 'PlatformConfigurationNativeApi::SetNeedsReportTimings')
   external static void __nativeSetNeedsReportTimings(bool value);
 
+  /// A callback that is invoked when a texture has a new frame available.
+  ///
+  /// This is called by the engine when [MarkTextureFrameAvailable] is invoked
```
chunhtai: is MarkTextureFrameAvailable a valid symbol? I can't find the declaration of MarkTextureFrameAvailable.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -635,6 +640,32 @@ class PlatformDispatcher {
...
+  /// for a texture. The callback receives the texture ID that has a new frame.
+  ///
+  /// This is used by the framework to mark texture render objects as needing
+  /// paint when their backing texture has new content, without scheduling a
+  /// new frame (since the frame is already scheduled or will be scheduled
```
chunhtai: I am trying to understand the code a bit more, if this is going to schedule a new frame anyway, why would getting a signal about a certain texture available matter? Isn't it all be redrawn in the new frame?

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +68,63 @@ mixin RendererBinding
...
+  /// Registry mapping texture IDs to their [TextureBox] render objects.
+  ///
+  /// When the engine notifies us of a new frame via
+  /// [handleTextureFrameAvailable], this registry is used to look up the
+  /// corresponding [TextureBox] and mark it as needing paint.
+  ///
+  /// Multiple [TextureBox] may be registered under the same texture ID
+  /// in which case all of them are marked as needing paint.
+  @visibleForTesting
+  final Map<int, Set<TextureBox>> textureRegistry = <int, Set<TextureBox>>{};
```
chunhtai: Unless this will be a performance concern, we should define the API to be more generic and inline with other similar listener api.

Can we do something like this?
define these in RenderBinding
```dart
void addTextureFrameAvailableHandler(handler);
void removeTextureFrameAvailableHandler(handler);
```

in the TextureBox
```
void attach() {
  addTextureFrameAvailableHandler(handle);
}

void detach() {
  removeTextureFrameAvailableHandler(handle);
}

void handle(int textureID) {
  if (id == textureID) {
     markNeedsPaint();
  }
}

```

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -330,6 +389,10 @@ mixin RendererBinding
   Iterable<RenderView> get renderViews => _viewIdToRenderView.values;
   final Map<Object, RenderView> _viewIdToRenderView = <Object, RenderView>{};
 
+  // Views that have not yet been composited in a non-warm-up frame.
+  // These views are always composited regardless of dirty state.
+  final Set<RenderView> _viewsNeedingCompositing = <RenderView>{};
```
chunhtai: the information on whether a RenderView needs compositing feels like something should be stored in RenderView. Maybe we can create one property or getter that something like

`bool get needCompositing` in RenderView, which also capture renderView.hasInitialFrameBeenPrepared and renderView.owner?.needsPaint as well as keeping track whether it has gone through a reload or resume. I think both could be listened in the RenderView thorugh lifecycle observer and `void reassemble`.

engine/src/flutter/lib/ui/platform_dispatcher.dart
knopp: >  Isn't it all be redrawn in the new frame.

The point of this PR is to avoid doing exactly that. Right now all views (windows) are rendering on every frame. With this PR only views that have dirty render objects will be rendered. However there might be a texture marked dirty by the engine that the framework does not know about, which would mean missing repainting the view.

engine/src/flutter/lib/ui/platform_dispatcher.dart
knopp: It's not a dart symbol. It probably shouldn't be in brackets.

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: ah so previously when draw frame, we are force to recomposite all renderview because it is possible a backend texture has updated or not. but with this signal that correctly mark which TextureBox is dirty, we can know exactly which one need to rebuild.

If that is the case, we should remove the sentence after `, without scheduling a ...`

This is referring to the old behavior and may be confusing to the future reader

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: I would avoid mention symbol if this is referring to something internally in engine.

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: wait, won't this be a breaking change for embedding that still relies on schedule a new frame when backend texture changes?

engine/src/flutter/lib/ui/platform_dispatcher.dart
knopp: It's not internal to the engine, it's part of public embedder API. Every platform has a way to inform Flutter that texture frame is available, but the exact API is not consistent across platforms.

engine/src/flutter/lib/ui/platform_dispatcher.dart
knopp: This shouldn't break anything. MarkTextureFrame available still schedules a frame. The difference here is in case when `markTextureFrameAvaiable` is called *and* something triggered the framework route (i.e. dirty render object), the framework will know about dirty texture in views that have no other changes in them (no dirty render object).

In case when `markTextureFrameAvailable` is called and no other render objects are dirty, the rendering should go the engine-only route as usual (reusing layer tree).

chunhtai: If I understand correctly, let's say an embedder A still call scheduleEngineFrame without calling `markTextureFrameAvailable` when texture changes.

previously they are fine because all renderview will recomposite regardless.

after this change, since nothing has markNeedsPaint, all the render view will skip recomposite.

Or did I miss something?

knopp: That's a good point. This does remind of me another issue with my original PR - there are situations where we need to force-repaint everything, i.e. when the application returns from background. And this would essentially prevent that. Is this what you have in mind or are you thinking of something else?

I'm not sure if this PR handles that correctly. We may need a flag to force redrawing all views.

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: yeah, in that case, the message should be more general instead of point out an exact symbol since the reader may not have knowledge on how engine is wired up

engine/src/flutter/lib/ui/platform_dispatcher.dart
chunhtai: > when the application returns from background. And this would essentially prevent that. Is this what you have in mind or are you thinking of something else?

In this case, the handleLifeCycleChange will add all of them to the _viewNeedsComposite, and should be fine I think?

> Is this what you have in mind or are you thinking of something else?

My concern is more on third party embeddings that relies on old behavior of schedule new frame will cause all view to be recomposite.

That being said, is the external texture change the only possibility that engine want to recomposite certain render object without actually dirtying the renderobject before this change?

knopp: Right, I missed that. Apparently the text search no longer works in github PRs :-/ .  This should be fine then I think. 

> That being said, is the external texture change the only possibility that engine want to recomposite certain render object without actually dirtying the renderobject before this change?

I'm not aware of anything other than external textures and android platform views (which is handled by this PR).

engine/src/flutter/lib/ui/platform_dispatcher.dart
9AZX: Done, i updated the comments

packages/flutter/lib/src/rendering/binding.dart
9AZX: Done, i replaced the registry with `add/removeTextureFrameAvailableHandler` and moved the id-match into `TextureBox`

packages/flutter/lib/src/rendering/binding.dart
9AZX: Done, I moved the state onto RenderView as requiresComposite + markRequiresComposite()

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,8 +262,38 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
```
knopp: Consider naming it `hasPreparedInitialFrame` or `initialFrameHasBeenPrepared`.

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,8 +262,38 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
   Matrix4? _rootTransform;
 
+  /// Whether this view must be composited on the next frame regardless of
+  /// whether any of its descendant render objects are dirty.
+  ///
+  /// The binding sets this to true when the view is added, after a hot
+  /// reload, and when the app returns from a backgrounded lifecycle state.
+  /// It is cleared after a successful composite in a non-warm-up frame.
+  bool get requiresComposite => _requiresComposite;
+  bool _requiresComposite = false;
```
knopp: Consider `_requiresCompositing`.

9AZX: Done.

packages/flutter/lib/src/rendering/view.dart
9AZX: Done.

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,8 +262,38 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
+
   Matrix4? _rootTransform;
 
+  /// Whether this view must be composited on the next frame regardless of
+  /// whether any of its descendant render objects are dirty.
+  ///
+  /// The binding sets this to true when the view is added, after a hot
+  /// reload, and when the app returns from a backgrounded lifecycle state.
+  /// It is cleared after a successful composite in a non-warm-up frame.
+  bool get requiresComposite => _requiresCompositing;
```
knopp: requiresCompositing.

knopp:
I am wondering if it would be worth to have a method that would call into the framework to mark all views as dirty.  That way we could do it in `Engine::HandleLifecyclePlatformMessage` and also from `FlutterEngineScheduleFrame`. The later should take care of breaking change concerns.

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,8 +262,38 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
+  bool _requiresCompositing = false;
+
+  /// Marks this view as requiring a composite on the next frame even if no
+  /// descendant render objects are dirty.
+  void markRequiresComposite() {
```
knopp: markRequireCompositing.

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -262,8 +262,38 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
+    _requiresCompositing = true;
+  }
+
+  /// Clears the [requiresComposite] flag.
+  ///
+  /// Called by the binding once this view has been composited in a
+  /// non-warm-up frame.
+  void clearRequiresComposite() {
```
knopp: clearRequiresCompositing.

packages/flutter/lib/src/rendering/view.dart
9AZX: Oops

9AZX:
> I am wondering if it would be worth to have a method that would call into the framework to mark all views as dirty. That way we could do it in `Engine::HandleLifecyclePlatformMessage` and also from `FlutterEngineScheduleFrame`. The later should take care of breaking change concerns.

Something like a `RuntimeController::MarkAllViewsDirty()` ? But do we really need it ?

justinmc:
I re-ran a test that failed due to infra problems.

knopp:
The golden change seems a bit weird? If the platform view was not painted at all I'd assume it's a regression from the PR, but it is painted just smaller?

knopp:
@chunhtai, this should be ready to re-review. I pushed commit where framework initiated repaint will only repaint views with dirty render objects, but any engine initiated repaint will redraw all views preserving current behavior and avoiding potentially breaking change for embedder API clients.

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -549,6 +549,32 @@ class PlatformConfiguration final {
...
+  ///             this to mark the corresponding texture render object as
+  ///             needing paint, ensuring the view containing the texture is
+  ///             recomposited even if no other render objects are dirty.
+  ///
+  ///             This method calls the `::_notifyTextureFrameAvailable` method
+  ///             in `hooks.dart`.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  void NotifyTextureFrameAvailable(int64_t texture_id);
```
gaaclarke, line 569: Should this have the same nomenclature, `MarkTextureFrameDirty(int64_t)`?

engine/src/flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -368,6 +368,36 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     invoke1<List<ui.FrameTiming>>(_onReportTimings, _onReportTimingsZone, timings);
   }
 
+  /// A callback invoked when a new frame is available for a texture.
+  @override
+  ui.TextureFrameAvailableCallback? get onTextureFrameAvailable => _onTextureFrameAvailable;
+  ui.TextureFrameAvailableCallback? _onTextureFrameAvailable;
+  Zone? _onTextureFrameAvailableZone;
```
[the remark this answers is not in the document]
gaaclarke: sounds reasonable

engine/src/flutter/shell/common/engine.h
```diff
@@ -652,6 +652,20 @@ class Engine final : public RuntimeDelegate, PointerDataDispatcher::Delegate {
...
+  ///
+  ///             This is called when the platform marks a texture as having new
+  ///             content via `MarkTextureFrameAvailable`. The framework uses
+  ///             this to mark the corresponding texture render object as
+  ///             needing paint, ensuring the view containing the texture is
+  ///             recomposited even if no other render objects are dirty.
+  ///
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  void NotifyTextureFrameAvailable(int64_t texture_id);
```
gaaclarke, line 669: Worth mentioning what thread this is expected to happen on.

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +75,52 @@ mixin RendererBinding
     rootPipelineOwner.attach(_manifold);
   }
 
+  final List<TextureFrameAvailableHandler> _textureFrameAvailableHandlers =
```
gaaclarke: Should these be weak references?  Not saying they should, when I see global collections like this, it's worth considering.

packages/flutter/lib/src/rendering/texture.dart
```diff
@@ -56,6 +57,24 @@ class TextureBox extends RenderBox {
...
+  void detach() {
+    RendererBinding.instance.removeTextureFrameAvailableHandler(_handleTextureFrameAvailable);
+    super.detach();
+  }
+
+  void _handleTextureFrameAvailable(int textureId) {
+    if (textureId == _textureId) {
+      markNeedsPaint();
+    }
+  }
```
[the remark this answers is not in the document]
gaaclarke: Looks like a valid comment to me.

gaaclarke (commented):
This is looking good when looking through it bit by bit.  Do we have enough integration test coverage for this?  There is a lot of plumbing that seems tested in isolation.

packages/flutter/lib/src/rendering/binding.dart
chunhtai: I don't think dart have weak references support for complex object unless we manually counted like LayerHandle

chunhtai: for this case, I think this is fine, we do this kind of handler a lot in binding

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -642,16 +700,48 @@ mixin RendererBinding
...
   }
 
+  /// Whether the given [renderView] needs to be composited during this frame.
+  ///
+  /// A view needs compositing if:
+  ///  - It has been prepared (i.e. [RenderView.prepareInitialFrame] was called).
+  ///  - Its [RenderView.requiresCompositing] flag is set (e.g. newly added
+  ///    views or views returning from a paused lifecycle state).
+  ///  - Its [PipelineOwner] has render objects that need painting.
+  bool _needsCompositing(RenderView renderView) {
```
chunhtai: instead of a helper method here, maybe put this as a getter in the RenderView? it seems safer to keep the logic internal in RenderView.

We can also avoid exposing requiresCompositing and hasPreparedInitialFrame

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -642,16 +700,48 @@ mixin RendererBinding
...
+    ];
+
     rootPipelineOwner.flushPaint();
     if (sendFramesToEngine) {
-      for (final RenderView renderView in renderViews) {
+      for (final renderView in viewsToComposite) {
         renderView.compositeFrame(); // this sends the bits to the GPU
       }
+      // Only clear [RenderView.requiresCompositing] if this is not a warm-up
+      // frame. During warm-up the rendering surface may not be ready, so the
```
chunhtai: I am not too familiar with the engine side in the warm up stage.

Is this needed? Once the surface is ready, shouldn't it schedule a new frame that dirty all the view?

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -642,16 +700,48 @@ mixin RendererBinding
...
+  ///    views or views returning from a paused lifecycle state).
+  ///  - Its [PipelineOwner] has render objects that need painting.
+  bool _needsCompositing(RenderView renderView) {
+    if (!renderView.hasPreparedInitialFrame) {
+      return false;
+    }
+    if (renderView.requiresCompositing) {
+      return true;
+    }
+    return renderView.owner?.needsPaint ?? false;
```
chunhtai: This feels weird I saw this is to ensure those has repainted to be included in the composite list. 

We should do this the otherway around, the owner paints should add the root view to the composite list somewhere, or calls markRequiresCompositing on the view

packages/flutter/lib/src/rendering/binding.dart
knopp: The handler is added and removed on RenderBox attach/detach. We don't rely on GC to cleanup the handlers, this should not be an issue.

engine/src/flutter/lib/ui/window/platform_configuration.h
knopp: It follows the texture API naming. I don't mind either way.

packages/flutter/lib/src/rendering/binding.dart
knopp: Moved this to render view and made it a simple getter (`requiresCompositing`).

packages/flutter/lib/src/rendering/binding.dart
knopp: Agreed. I have made `requiresCompositing` a simple flag that gets set by the owner in `flushPaint` if there are any dirty render objects.

packages/flutter/lib/src/rendering/binding.dart
knopp: The warm-up frame is a bit convoluted. `SchedulerBinding.scheduleWarmUpFrame` documentation says that the warm-up frame might not be rendered and that is indeed observed behavior. The next frame is triggered by vsync, which does not dirty all the views, if it did it would render this PR pointless. So we need to preserve the `requiresCompositing` during the warm up frame. I simplified a bit though how this is done in the PR.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -973,6 +973,14 @@ class _LocalSemanticsHandle implements SemanticsHandle {
   }
 }
 
+/// Render object that can be used as a root for a [PipelineOwner].
+mixin RootRenderObject on RenderObject {
```
chunhtai: I don't think it worth adding this mixin, it is not really better than just type the root node a RenderView.

I think in this case let's just use type check to avoid the breaking change
```dart
      if (dirtyNodes.isNotEmpty && rootNode is RenderView) {
        rootNode?.markRequiresCompositing();
      }

```

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -264,6 +264,30 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
+  /// background).
+  @override
+  void markRequiresCompositing() {
+    _requiresCompositing = true;
+  }
+
+  /// Whether the given [RenderView] needs to be composited during this frame.
+  /// A [RenderView] needs to be composited of it was marked as requiring compositing,
+  /// or if the owner has any dirty render objects.
+  bool shouldBeComposited() {
```
chunhtai: this doesn't seem needed?

packages/flutter/lib/src/rendering/view.dart
chunhtai, line 366: we should add an assert here to make sure _requiresCompositing is true

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -264,6 +264,30 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
...
+  bool _requiresCompositing = false;
+
+  /// Marks this view required to be composited on the next frame.
+  ///
+  /// This happens by the [PipelineOwner] during [PipelineOwner.flushPaint] if
+  /// there are any render objects that need to be painted. It is also set for
+  /// warmup frame or any frame requested by the engine (i.e. during return from
+  /// background).
+  @override
+  void markRequiresCompositing() {
```
chunhtai: can you rename method and the getter to be markNeedsCompositing, needsCompositing to be more inline with other API?

packages/flutter/lib/src/rendering/view.dart
knopp: Good catch.

packages/flutter/lib/src/rendering/object.dart
knopp: Done.

packages/flutter/lib/src/rendering/view.dart
knopp: Done.

packages/flutter/lib/src/rendering/view.dart
knopp: There already is `RenderObject._needsCompositing`, but it has different meaning (roughly determines whether the render object need a separate layer or can paint to current picture). If you have different suggestion for a name I'll be happy to change it.

chunhtai: maybe markNeedsCompositeFrame

knopp:
@chunhtai, could you give this one another go? Thanks!

gaaclarke:
> @chunhtai, could you give this one another go? Thanks!

friendly reminder, if you want people to rereview something there is a button you should press next to their handle on the top right, its tooltip is "Re-request review".  It looks like 2 arrows chasing eachother.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1313,6 +1320,9 @@ base class PipelineOwner with DiagnosticableTreeMixin {
         return true;
       }());
       final List<RenderObject> dirtyNodes = _nodesNeedingPaint;
+      if (dirtyNodes.isNotEmpty && rootNode is RenderView) {
```
chunhtai: FYI, @loic-sharma 

I suggested the author to use `is RenderView` check here since I can't think of a better way without introducing breaking change, and introducing a breaking change for this seems a bit overkill. We can fix this if we have a better reason to directly type the rootNode as RenderView in the future

Let me know if you think otherwise

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1274,6 +1275,12 @@ base class PipelineOwner with DiagnosticableTreeMixin {
   @nonVirtual
   Iterable<RenderObject> get nodesNeedingPaint => _nodesNeedingPaint;
 
+  /// Returns true if there are any nodes that need to be painted.
+  ///
+  /// This is used by [RendererBinding.drawFrame] to determine which views
+  /// need to be composited.
+  bool get needsPaint => _nodesNeedingPaint.isNotEmpty;
```
chunhtai: is this needed? for tests, they can access nodesNeedingPaint directly

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -345,7 +361,7 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
   /// the [configuration] must be set to a non-null value, and the
   /// [prepareInitialFrame] method must have been called.
   void compositeFrame() {
-    if (!kReleaseMode) {
+    if (!kReleaseMode && _needsCompositeFrame) {
```
chunhtai: I meant an assert(_needsCompositeFrame) in the beginning of compositeFrame.

chunhtai: unless there is a case where compositeFrame is called when _needsCompositeFrame == false

packages/flutter/lib/src/rendering/object.dart
loic-sharma: I see `_RawViewElement` passes callbacks to the `PipelineOwner` to let the `PipelineOwner` interact with the `RenderView`:

https://github.com/flutter/flutter/blob/29e56679272ed26de6190550e9ffe5aaf9c41ed9/packages/flutter/lib/src/widgets/view.dart#L449-L456

https://github.com/flutter/flutter/blob/29e56679272ed26de6190550e9ffe5aaf9c41ed9/packages/flutter/lib/src/widgets/view.dart#L461-L471

What do y'all think of adding a new `onMarkNeedsCompositeFrame` callback to `PipelineOwner`? Something like:

```dart
class _RawViewElement extends RenderTreeRootElement {
  late final PipelineOwner _pipelineOwner = PipelineOwner(
    // ...
    onMarkNeedsCompositeFrame: _handleMarkNeedsCompositeFrame,
  );

  // ...

  void _handleMarkNeedsCompositeFrame() {
    (_effectivePipelineOwner.rootNode as RenderView?)?.markNeedsCompositeFrame();
  }
}
```

And then `PipelineOwner.flushPaint` would call `onMarkNeedsCompositeFrame` if there are dirty nodes. This way, `PipelineOwner` remains agnostic of the root render node.

chunhtai: agree this is the better way to structure the code.

Though the method should probably be more generic
```dart
onFlushPaint: _handleFlushPaint

void _handleFlushPaint(List<RenderObject> dirtyNode) {
   if (dirtyNode.isNotEmpty)
    (_effectivePipelineOwner.rootNode as RenderView?)?.markNeedsCompositeFrame();
  }
```

alternatively, if we don't want to introduce new API at. view.dart can create a private subclass of PipelineOwner to inject the callback

loic-sharma: A more generic method sounds good to me, but what do you think of a boolean instead of list of dirty nodes? Something like:

```dart
void _handleFlushPaint(bool isDirty) {
  if (isDirty) {
    (_effectivePipelineOwner.rootNode as RenderView?)?.markNeedsCompositeFrame();
  }
}
```

This would prevent someone from doing evil things with this callback, like modifying render objects

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -638,6 +646,46 @@ class PlatformDispatcher {
...
+  // Called from the engine, via hooks.dart
+  void _notifyTextureFrameAvailable(int textureId) {
+    _invoke1<int>(onTextureFrameAvailable, _onTextureFrameAvailableZone, textureId);
+  }
+
+  /// Called from the engine to force repaint of all views on the next frame.
+  MarkAllViewsDirtyCallback? get onMarkAllViewsDirty => _onMarkAllViewsDirty;
+  MarkAllViewsDirtyCallback? _onMarkAllViewsDirty;
+  Zone _onMarkAllViewsDirtyZone = Zone.root;
+  set onMarkAllViewsDirty(MarkAllViewsDirtyCallback? callback) {
```
loic-sharma: For some reason, I don't love this name. What do y'all think of `onMarkAllViewsNeedRender` instead? Or perhaps `onForceRender`, or `onRequestForceRender`?

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -638,6 +646,46 @@ class PlatformDispatcher {
...
+    _onTextureFrameAvailable = callback;
+    _onTextureFrameAvailableZone = Zone.current;
+  }
+
+  // Called from the engine, via hooks.dart
+  void _notifyTextureFrameAvailable(int textureId) {
+    _invoke1<int>(onTextureFrameAvailable, _onTextureFrameAvailableZone, textureId);
+  }
+
+  /// Called from the engine to force repaint of all views on the next frame.
```
loic-sharma: What do you think of this?

```suggestion
  /// A callback that is invoked when the application should re-render
  /// all views on the next frame.
  ///
  /// On the next [onBeginFrame] or [onDrawFrame], the application should
  /// invoke [FlutterView.render] on all [views].
```

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +75,52 @@ mixin RendererBinding
...
+  final List<TextureFrameAvailableHandler> _textureFrameAvailableHandlers =
+      <TextureFrameAvailableHandler>[];
+
+  /// Registers a [handler] to be notified when any texture has a new frame
+  /// available from the engine.
+  ///
+  /// The handler receives the ID of the texture that has a new frame and is
+  /// responsible for checking whether that ID matches the texture it cares
+  /// about.
+  void addTextureFrameAvailableHandler(TextureFrameAvailableHandler handler) {
```
loic-sharma: Flutter usually uses the name "callback", I'd be consistent here:

```suggestion
  void addTextureFrameAvailableCallback(TextureFrameAvailableCallback callback) {
```

packages/flutter/lib/src/scheduler/binding.dart
```diff
@@ -995,6 +995,14 @@ mixin SchedulerBinding on BindingBase {
 
   bool _warmUpFrame = false;
 
+  /// Whether the scheduler is currently producing a warm-up frame.
+  ///
+  /// A warm-up frame is scheduled by [scheduleWarmUpFrame] and may not
```
loic-sharma:
```suggestion
  /// A warm-up frame is scheduled by [scheduleWarmUpFrame] and might not
```

packages/flutter/lib/src/rendering/object.dart
knopp: Added a category to test that uses `nodesNeedingPaint` directly.

packages/flutter/lib/src/rendering/object.dart
knopp: Done.

examples/multiple_windows/.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
...
+            "type": "dart",
+            "request": "launch",
+            "program": "lib/main.dart",
+            "toolArgs": [
+                "--local-engine=host_debug_unopt_arm64",
+                "--local-engine-host=host_debug_unopt_arm64"
+            ]
+        },
+    ]
+}
```
loic-sharma: Should we remove this?

engine/src/flutter/lib/ui/platform_dispatcher.dart
loic-sharma: cc @knopp This one was missed :)

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -638,6 +646,46 @@ class PlatformDispatcher {
...
+    _onTextureFrameAvailable = callback;
+    _onTextureFrameAvailableZone = Zone.current;
+  }
+
+  // Called from the engine, via hooks.dart
+  void _notifyTextureFrameAvailable(int textureId) {
+    _invoke1<int>(onTextureFrameAvailable, _onTextureFrameAvailableZone, textureId);
+  }
+
+  /// A callback that is invoked when the application should re-render.
```
loic-sharma: Let's add a little more detail here:

```suggestion
  /// A callback that is invoked when the engine requires the application to
  /// re-render all of its views.
  ///
  /// On the next [onBeginFrame] or [onDrawFrame], the application should
  /// invoke [FlutterView.render] on all of its [views].
```

engine/src/flutter/lib/ui/window/platform_configuration.h
loic-sharma: I personally prefer this name `NotifyTextureFrameAvailable` over `MarkTextureFrameDirty`. See this other thread where I suggested renaming the other callback: https://github.com/flutter/flutter/pull/179874/changes#r3352375863

examples/multiple_windows/macos/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Flutter-related
+**/Flutter/ephemeral/
+**/Pods/
+
+# Xcode-related
+**/dgph
+**/xcuserdata/
```
loic-sharma: I'd remove this as well, it doesn't seem like other other example apps have a `macos/.gitignore` file

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1061,6 +1063,11 @@ base class PipelineOwner with DiagnosticableTreeMixin {
   /// Typical implementations will tear down the semantics tree.
   final VoidCallback? onSemanticsOwnerDisposed;
 
+  /// Called whenever this pipeline owner flushes paint.
+  ///
+  /// isDirty is true if the pipeline owner has dirty render objects that need to be painted.
+  final void Function(bool isDirty)? onFlushPaint;
```
loic-sharma: Please add a `typedef` for this callback. See the style guide here: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

examples/multiple_windows/.vscode/launch.json
knopp: Yeah. Accident.

chunhtai (dismissed):
LGTM once all comments are addressed

loic-sharma:
@knopp For my own understanding, how do platform views work here? Is the idea that the system compositor is responsible for re-compositing the platform view if its content has changed? The framework doesn't need to be notified if a platform view has changed?

engine/src/flutter/lib/ui/platform_dispatcher.dart
knopp: I think I'd go with `onMarkAllViewsNeedRender`.

knopp: Done.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1061,6 +1066,11 @@ base class PipelineOwner with DiagnosticableTreeMixin {
   /// Typical implementations will tear down the semantics tree.
   final VoidCallback? onSemanticsOwnerDisposed;
 
+  /// Called whenever this pipeline owner flushes paint.
+  ///
+  /// isDirty is true if the pipeline owner has dirty render objects that need to be painted.
+  final OnFlushViewCallback? onFlushPaint;
```
loic-sharma:
```suggestion
  /// Called when this pipeline owner starts the paint phase.
  ///
  /// isDirty is true if the pipeline owner has dirty render objects that need to be painted.
  final OnFlushPaintCallback? onFlushPaint;
```

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -973,6 +974,9 @@ class _LocalSemanticsHandle implements SemanticsHandle {
   }
 }
 
+/// Signature for a function that is called whenever the pipeline owner flushes paint.
+typedef OnFlushViewCallback = void Function(bool isDirty);
```
loic-sharma:
```suggestion
/// Signature for a function that is called when the pipeline owner starts the
/// paint phase.
typedef OnFlushPaintCallback = void Function(bool isDirty);
```

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1313,6 +1323,7 @@ base class PipelineOwner with DiagnosticableTreeMixin {
         return true;
       }());
       final List<RenderObject> dirtyNodes = _nodesNeedingPaint;
+      onFlushPaint?.call(dirtyNodes.isNotEmpty);
```
loic-sharma: I wonder if we should make this a `onFlushedPaint` callback that is invoked after this pipeline owner + all child pipeline owners have been flushed. For some reason, that feels a little cleaner to me. Currently the flow is: I start paint -> I mark the view dirty if nodes are dirty -> I actually paint. With this new callback it would be: I paint -> I mark the view dirty if there were dirty nodes.

packages/flutter/lib/src/rendering/view.dart
```diff
@@ -372,6 +388,7 @@ class RenderView extends RenderObject with RenderObjectWithChildMixin<RenderBox>
       if (!kReleaseMode) {
         FlutterTimeline.finishSync();
       }
+      _needsCompositeFrame = false;
```
loic-sharma: Should this be in the `try` branch? If `_updateSystemChrome()` throws, I don't think we want to set this to `false`.

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1775,6 +1771,10 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
       return true;
     }());
 
+    if (isWindowingEnabled) {
+      result = WindowManager(child: result);
+    }
```
loic-sharma: Can we undo this change? This seems like it should be a separate PR

knopp: Thanks for spotting this. Something weird happened when rebasing.

packages/flutter/lib/src/rendering/view.dart
knopp: Makes sense.

packages/flutter/lib/src/rendering/object.dart
knopp: That does make sense. Updated.

engine/src/flutter/lib/ui/window/platform_configuration.cc
```diff
@@ -88,6 +88,12 @@ void PlatformConfiguration::DidCreateIsolate() {
                   Dart_GetField(library, tonic::ToDart("_drawFrame")));
   report_timings_.Set(tonic::DartState::Current(),
                       Dart_GetField(library, tonic::ToDart("_reportTimings")));
+  notify_texture_frame_available_.Set(
+      tonic::DartState::Current(),
+      Dart_GetField(library, tonic::ToDart("_notifyTextureFrameAvailable")));
+  mark_all_views_dirty_.Set(
```
loic-sharma: Let's rename this field to `mark_all_views_need_render_`

knopp: I knew I missed one :)

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +75,54 @@ mixin RendererBinding
...
+    // Iterate a copy so handlers may add/remove themselves mid-dispatch.
+    for (final callback in List<TextureFrameAvailableCallback>.of(
+      _textureFrameAvailableCallbacks,
+    )) {
+      callback(textureId);
+    }
+  }
+
+  /// Marks all views to require compositing, which will force them to repaint
+  /// on the next frame.
```
loic-sharma: To mirror other handlers, let's link to the `dart:ui` API this corresponds to. Also, maybe say "re-render" instead of "repaint" since this might not actually result in a repaint.

```suggestion
  /// Marks all views to require compositing, which will force them to re-render
  /// on the next frame.
  ///
  /// See [dart:ui.PlatformDispatcher.onMarkAllViewsNeedRender].
```

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -66,6 +75,54 @@ mixin RendererBinding
...
+  @protected
+  @visibleForTesting
+  void markAllViewsNeedRender() {
```
loic-sharma: Style guide is that we prefix a methods for callbacks with `handleFoo`: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

```suggestion
  void handleMarkAllViewsNeedRender() {
```

engine/src/flutter/lib/ui/window/platform_configuration.h
```diff
@@ -549,6 +549,34 @@ class PlatformConfiguration final {
...
+  /// @param[in]  texture_id  The ID of the texture that has a new frame.
+  ///
+  /// @note       Must be called on the UI task runner.
+  ///
+  void NotifyTextureFrameAvailable(int64_t texture_id);
+
+  //----------------------------------------------------------------------------
+  /// @brief      Notifies the framework that all views should be marked dirty.
+  ///
+  ///             This is called when the engine needs to force full repaint of
```
loic-sharma:
```suggestion
  ///             This is called when the engine needs to force full re-render of
```

packages/flutter/test/rendering/texture_test.dart
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  test('TextureBox becomes dirty when its textureId fires', () {
+    final _TextureScene scene = buildScene(textureId: 42);
+    expect(scene.owner.needsPaint, isFalse);
+
+    binding.handleTextureFrameAvailable(42);
+    expect(scene.owner.needsPaint, isTrue);
+
+    tearDownScene(scene.renderView);
```
loic-sharma: Consider switching to an `addTearDown` immediately after the `scene` is created.

engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -638,6 +649,50 @@ class PlatformDispatcher {
...
+  // Called from the engine, via hooks.dart
+  void _notifyTextureFrameAvailable(int textureId) {
+    _invoke1<int>(onTextureFrameAvailable, _onTextureFrameAvailableZone, textureId);
+  }
+
+  /// A callback that is invoked when the engine requires the application to
+  /// re-render all of its views.
+  ///
+  /// On the next [onBeginFrame] or [onDrawFrame], the application should
+  /// invoke [FlutterView.render] on all of its [views].
```
loic-sharma: Consider adding (but I don't feel strongly about this!)

```suggestion
  /// On the next [onBeginFrame] or [onDrawFrame], the application should
  /// invoke [FlutterView.render] on all of its [views]. This callback
  /// can be ignored if the application always renders all views on each
  /// frame.
```

loic-sharma (dismissed):
Looks good to me!

loic-sharma:
@chunhtai This is ready for another re-review :)

loic-sharma:
FYI it looks like there are real test failures. Example:  https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8674451357878685153/+/u/run_test.dart_for_framework_tests_shard_and_subshard_slow/stdout

```
00:01 +4 -1: /b/s/w/ir/x/w/flutter/dev/tracing_tests/test/timeline_test.dart: Timeline [E]
  Expected: [
              'BUILD',
              'Placeholder',
              'CustomPaint',
              'LAYOUT',
              'UPDATING COMPOSITING BITS',
              'PAINT',
              'COMPOSITING',
              'FINALIZE TREE'
            ]
    Actual: MappedListIterable<TimelineEvent, String>:[
              'BUILD',
              'Placeholder',
              'CustomPaint',
              'LAYOUT',
              'UPDATING COMPOSITING BITS',
              'PAINT',
              'FINALIZE TREE'
            ]
     Which: at location [6] is 'FINALIZE TREE' instead of 'COMPOSITING'
  
  package:matcher                                     expect
  package:flutter_test/src/widget_tester.dart 473:18  expect
  test/timeline_test.dart 124:5                       main.<fn>
  ```

**Comment by Piinks** [comment]
Removing from the Google testing queue as there are failures here to address first.

**Review by loic-sharma — DISMISSED** [review]
Re-LGTM

File: packages/flutter/lib/src/rendering/binding.dart [hunk 54]
```diff
@@ -29,6 +29,12 @@ export 'package:flutter/gestures.dart' show HitTestResult;
 // Examples can assume:
 // late BuildContext context;
 
+/// Signature for callbacks passed to
+/// [RendererBinding.addTextureFrameAvailableCallback].
+///
+/// Called with the ID of the texture that has a new frame available.
+typedef TextureFrameAvailableCallback = void Function(int textureId);
```
chunhtai: I thought we defined this in dart ui already?

chunhtai: in the import, we should show the declaration
```
import 'dart:ui' as ui show PictureRecorder, Rect, SceneBuilder, SemanticsUpdate, TextureFrameAvailableCallback;
```

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -973,6 +974,10 @@ class _LocalSemanticsHandle implements SemanticsHandle {
   }
 }
 
+/// Signature for a function that is called when the pipeline owner and child owner
+/// finished the paint phase.
+typedef OnFlushedPaintCallback = void Function(bool isDirty);
```
chunhtai:
```suggestion
typedef FlushedPaintCallback = void Function(bool isDirty);
```
per style guide https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

chunhtai (commented):
only a few minor comments, the rest LGTM

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1061,6 +1067,11 @@ base class PipelineOwner with DiagnosticableTreeMixin {
   /// Typical implementations will tear down the semantics tree.
   final VoidCallback? onSemanticsOwnerDisposed;
 
+  /// Called when this pipeline owner and child owner have finished the paint phase.
+  ///
+  /// isDirty is true if the pipeline owner has dirty render objects that needed to be painted.
```
justinmc: Super nit: "The `isDirty` parameter is true if..."

justinmc (dismissed):
LGTM, mostly looking from the framework point of view

Correct my understanding from a high level: This optimization affects only situations where the app is multiview and one view is receiving texture updates from the platform, such as playing a video or using a platform view? Or also when a view is un-backgrounded?

knopp:
> LGTM, mostly looking from the framework point of view
> 
> Correct my understanding from a high level: This optimization affects only situations where the app is multiview and one view is receiving texture updates from the platform, such as playing a video or using a platform view? Or also when a view is un-backgrounded?

Not quite sure what you mean. Overall this PR affects situations where application has multiple views but only some of them need to be repainted. Before the PR all views are repainted all the time. The additional complexity regarding external textures in this PR is because when plugin marks external texture as changed, the framework does not know about it. And when external texture changed in view A and some render object was dirty in view B, the framework would only repaint view B. With this PR the render object for external texture will be marked dirty and view A will be repainted.

chunhtai (approved):
LGTM

loic-sharma (approved):
Re-LGTM

flar:
Reason for revert: https://github.com/flutter/flutter/issues/191251

knopp:
welp. that didn't take long.

flar:
> welp. that didn't take long.

Is there a sad emoji? The closest I found in the list was "confused" which didn't sound right. Anyway... 😢

knopp:
> > welp. that didn't take long.
> 
> Is there a sad emoji? The closest I found in the list was "confused" which didn't sound right. Anyway... 😢

I am wondering a bit if something like this should have been caught in presubmit checks...

flar:
> I am wondering a bit if something like this should have been caught in presubmit checks...

Benchmarks only run after submit because we record them in a dashboard by their true final commit. During pre-submit they don't have a commit key to report against. We could potentially run them in "don't submit/dry run" mode maybe in pre-submit, though. I'm not sure how the web benchmarks are configured or whether they have such a mode.

We also briefly looked at some other benchmark dashboards at least one of which had a nice feature that you could run your benchmarks in pre-submit and get a report on any potential regression. It would take some effort to upgrade all of our benchmarks to a new system, though. :(

jtmcdole:
We can't run benchmarks on every PR since that requires running on a phone. If we had an analogue that was just running benchmarks on a machine, I'd entertain a test in presubmit.

The only other option is using `led` to run the benchmark on a machine in the try pool.

Issue #140756, opened 2023-12-30
[RenderParagraph] - Should call markNeedsLayout in the textAlign setter


matthew-carroll:
### Steps to reproduce

1. Use the provided reproduction code
2. Tap on the text to change the alignment from left to right

### Expected results

The `LayoutBuilder` surrounding the `Text` widget should build again, because the layout of its descendant`Text` changed.

### Actual results

The `LayoutBuilder` doesn't run again because the descendant `Text` didn't report a layout change. It only reported dirty paint.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

const Color darkBlue = Color.fromARGB(255, 18, 32, 47);

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      theme: ThemeData.dark().copyWith(
        scaffoldBackgroundColor: darkBlue,
      ),
      debugShowCheckedModeBanner: false,
      home: Scaffold(
        body: MyWidget(),
      ),
    );
  }
}

class MyWidget extends StatefulWidget {
  @override
  State<MyWidget> createState() => _MyWidgetState();
}

class _MyWidgetState extends State<MyWidget> {
  @override
  Widget build(BuildContext context) {
    return LayoutBuilder(
      builder: (context, constraints) {
        print("Running LayoutBuilder");
        return SizedBox(
          width: double.infinity, 
          child: TextAlignChanger(),
        );
      }
    );
  }
}


class TextAlignChanger extends StatefulWidget {
  @override
  State<TextAlignChanger> createState() => _TextAlignChangerState();
}

class _TextAlignChangerState extends State<TextAlignChanger> {
  TextAlign _textAlign = TextAlign.left;
  
  @override
  Widget build(BuildContext context) {
    return GestureDetector(
      onTap: () => setState(() {
        print("Changing text alignment to the right");
        _textAlign = TextAlign.right;
      }),
      child: Text(
        'Hello, World!',
        textAlign: _textAlign,
        style: Theme.of(context).textTheme.headlineMedium,
      ),
    );
  }
}
```

</details>

### Flutter Doctor output

Reproduced in Dart Pad: Based on Flutter 3.16.5 Dart SDK 3.2.3

moffatman:
Just to clarify, why is this change needed, because text align change shouldn't change the size of text. Just the distribution of words within each line.

matthew-carroll:
[...]

PR #179881, opened 2025-12-15
Fix RenderParagraph.textAlign setter to call markNeedsLayout

shindonghwi:
## Problem

When you change `textAlign` on a `RenderParagraph`, the parent widget doesn't know the layout changed.

**Example:** If you use `LayoutBuilder` to position elements based on text alignment, changing from `TextAlign.left` to `TextAlign.right` won't trigger a rebuild - even though the text moved.

## Cause

The `textAlign` setter was calling `markNeedsPaint()` (just repaint) instead of `markNeedsLayout()` (recalculate layout).

## Fix

One line change: `markNeedsPaint()` → `markNeedsLayout()`

This matches how other setters like `textDirection` and `softWrap` already work.

## Fixes

https://github.com/flutter/flutter/issues/140756

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/rendering/paragraph.dart
+++ b/packages/flutter/lib/src/rendering/paragraph.dart
@@ -555,7 +555,7 @@ class RenderParagraph extends RenderBox
       return;
     }
     _textPainter.textAlign = value;
-    markNeedsPaint();
+    markNeedsLayout();
   }
 
   /// The directionality of the text.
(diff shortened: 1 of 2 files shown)


justinmc (approved):
LGTM 👍 . This makes sense to me but maybe @LongCatIsLooong can double check me here. I guess changing textAlign (or textDirection) would never change the height or width of the RenderParagraph, only the position of the text. In @shindonghwi's use case of positioning something on top of the text, that still matters for their layout. So should we consider stuff like this to require relayout then or should we expect the user to do something else if they're positioning based on the text location?

justinmc (changes requested):
After reading the issue (https://github.com/flutter/flutter/issues/140756) I'm not convinced we want to do this. @LongCatIsLooong do you still have any plans to provide a way to be notified of text layout changes?

PR #179906, opened 2025-12-15
Moved high bitrate texture tests to engine

gaaclarke:
This adds a few goldens where skia supports the file format.  This is a bit more scalable and direct than managing integration tests.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[...]

PR #179927, opened 2025-12-16
Allow setting vector uniforms by name.

walley892:
Create a few thin wrappers around `UniformFloatSlot` that allow users to get Vector uniforms by name and set all elements in one go.

Before:
```dart
shader.getUniformFloat('uColor', 0).set(color.r);
shader.getUniformFloat('uColor', 1).set(color.g);
shader.getUniformFloat('uColor', 2).set(color.b);
```

After:
```dart
shader.getUniformVec3('uColor').set(color.r, color.g, color.b);
```

This enforces that the requested vector is actually of the requested size. For example:
```dart
shader.getUniformVec2('someVec3Uniform');
```
will fail instead of allowing partial uniform access.

## Follow up
This doesn't add list or matrix datatypes, those will come later. 
Also doesn't have anything super fancy like accessors/setters for individual elements.

## Pre-launch Checklist

- [ x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ x] I signed the [CLA].
- [ x] I listed at least one issue that this PR fixes in the description above.
- [ x] I updated/added relevant documentation (doc comments with `///`).
- [ x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ x] All existing and new tests are passing.


engine/src/flutter/testing/dart/fragment_shader_test.dart
```diff
@@ -106,6 +106,49 @@ void main() async {
...
+  test('FragmentProgram getUniformVec4 wrong size', () async {
+    final FragmentProgram program = await FragmentProgram.fromAsset('uniforms.frag.iplr');
+    final FragmentShader shader = program.fragmentShader();
+    try {
+      shader.getUniformVec4('iVec3Uniform');
+      fail('Unreachable');
+    } catch (e) {
+      expect(e.toString(), contains('`iVec3Uniform` has size 3, not size 4.'));
+    }
+  });
```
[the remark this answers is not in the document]
walley892: Nice

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5439,6 +5455,45 @@ base class UniformFloatSlot {
...
+base class UniformVec2Slot {
+  UniformVec2Slot._(this._xSlot, this._ySlot);
+
+  /// Set the float value of the bound uniform.
+  void set(double x, double y) {
+    _xSlot.set(x);
+    _ySlot.set(y);
+  }
+
+  final UniformFloatSlot _xSlot, _ySlot;
```
walley892, line 5552: I'm not sure if it's nicer here to have two members of type `UniformFloatSlot` or one member that's a list of slots. The former makes very clear that this only contains two slots and doesn't require validation inside of the constructor, the latter seems simpler and *slick*. Same for the other classes.

Thoughts?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5491,6 +5546,19 @@ base class FragmentShader extends Shader {
...
 
+  List<UniformFloatSlot> _getSlotsForUniform(String name, int size) {
+    final int dataSize = _program._getUniformVectorSize(name);
+    assert(dataSize == size, 'Uniform `$name` has size $dataSize, not size $size.');
+    final int baseShaderIndex = _program._getUniformFloatIndex(name, 0);
+    final slots = List<UniformFloatSlot>.generate(
+      size,
+      (i) => UniformFloatSlot._(this, name, baseShaderIndex, i),
+    );
+    _slots.removeWhere((WeakReference<UniformFloatSlot> ref) => ref.target == null);
```
walley892, line 5668: Made a direct call to the UniformFloatSlot constructor here instead of relying on the logic in and repeatedly calling `getUniformFloat` to avoid the duplicated work in `_getUniformFloatIndex`. Also maybe `addAll` is more efficient than repeated calls to `add`?

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/painting.dart
```diff
@@ -450,3 +487,42 @@ class CkUniformFloatSlot implements ui.UniformFloatSlot {
   @override
   final int shaderIndex;
 }
+
+class CkUniformVec2Slot implements ui.UniformVec2Slot {
```
walley892: These classes feel really icky to me. I think we can maybe get away with not defining them if we make public constructors for the `UniformVecNSlot` classes, but I'm not sure if we'd want users to call those constructors directly and do things like:

```
var a = shader.getUniformFloat('uTime');
var b = shader.getUniformFloat('uLength');
var c = UniformVec2Slot.fromSlots(a, b);
```

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -403,3 +440,42 @@ class SkwasmUniformFloatSlot implements ui.UniformFloatSlot {
   @override
   final int shaderIndex;
 }
+
+class SkwasmUniformVec2Slot implements ui.UniformVec2Slot {
```
walley892: Same comment about these feeling icky.

engine/src/flutter/testing/dart/fragment_shader_test.dart
```diff
@@ -59,54 +56,92 @@ void main() async {
...
-      shader.getUniformFloat('iMat2Uniform', 1),
-      shader.getUniformFloat('iMat2Uniform', 2),
-      shader.getUniformFloat('iMat2Uniform', 3),
-    ];
-    for (var i = 0; i < slots.length; ++i) {
-      expect(slots[i].shaderIndex, equals(i));
-    }
-  });
+  group('FragmentProgram getUniform*', () {
+    late FragmentShader shader;
```
gaaclarke, line 63: I just found this yesterday, but did you notice line 43?  Can you try moving these tests up above that and seeing if they work.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: Having the named fields is better in Dart.  I don't know if Dart has a reliable way to optimize away indexing into the list, probably not given the dynamic nature of List and the checked indexing.

In languages like c/c++ the contract for indexing is very well defined and it can be optimized away by the compiler.

When doing graphics work I like to divide code between things that happen at least once a frame and things that happen less frequently.  We can be a bit more careful with the performance for things that happen each frame.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5465,6 +5455,45 @@ base class UniformFloatSlot {
   final int index;
 }
 
+base class UniformVec2Slot {
```
gaaclarke, line 5543: You're going to have to add docstrings to these classes too.

engine/src/flutter/lib/ui/painting.dart
gaaclarke: sgtm

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5517,6 +5546,19 @@ base class FragmentShader extends Shader {
     _floats = _constructor(_program, _program._uniformFloatCount, _program._samplerCount);
   }
 
+  List<UniformFloatSlot> _getSlotsForUniform(String name, int size) {
+    final int dataSize = _program._getUniformVectorSize(name);
+    assert(dataSize == size, 'Uniform `$name` has size $dataSize, not size $size.');
```
gaaclarke: Why even have the argument to this function if we can derive it correctly anyways?  I'd just check the size of the returned list and throw an error if they don't match.  We don't want to hit asserts in response to bad user interactions.  Asserts should be reserved for developer errors.

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/painting.dart
```diff
@@ -429,10 +434,38 @@ class CkFragmentShader implements ui.FragmentShader, CkShader {
...
+  }
+
   @override
   ui.ImageSamplerSlot getImageSampler(String name) {
     throw UnsupportedError('getImageSampler is not supported on the web.');
   }
+
+  List<CkUniformFloatSlot> _getUniformFloatSlots(String name, int size) {
+    final int dataSize = _program._getUniformVectorSize(name);
+    assert(dataSize == size, 'Uniform `$name` has size $dataSize, not size $size.');
```
gaaclarke: same here

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/painting.dart
gaaclarke: I think you can hide these classes, _CkUniformVec2Slot.  Only the people creating them ever need to know the actual type and the people creating them should be us.

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
gaaclarke: Same thing, make them private.

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
```diff
@@ -355,4 +355,28 @@ void testMain() {
...
+  test('Uniform by-name accessors work throw errors with incorrect datatypes.', () {
+    final Uint8List data = utf8.encode(kJsonArrayIPLR);
+    final program = CkFragmentProgram.fromBytes('test', data);
+    final shader = program.fragmentShader() as CkFragmentShader;
+    try {
+      shader.getUniformVec3('uSize');
+      fail('Unreachable');
+    } catch (e) {
+      expect(e.toString(), contains('`uSize` has size 2, not size 3.'));
+    }
```
gaaclarke: Each one of these should be their own test.  That way execution isn't interrupted if one fails and it's easier to zero in on the failure.

gaaclarke: You should use the `throwsA` matcher (or `throwsArgumentError`?):

```dart
expect(
  () => functionThatThrows(), 
  throwsA(
    isA<MyCustomException>()
      .having((e) => e.message, 'message', contains('invalid input'))
  )
);
```

gaaclarke (changes requested):
Looking good, just some minor cleanup

engine/src/flutter/testing/dart/fragment_shader_test.dart
walley892: I hit the error `Asset 'uniforms.frag.iplr' does not contain appropriate runtime stage data for current backend` without that check.

engine/src/flutter/lib/ui/painting.dart
walley892: Aye aye, captain

engine/src/flutter/lib/ui/painting.dart
walley892: Changed to throwing rather than an assert. I wanted to centralize the work of updating `_slots` - if we move this error check into each `getUniform*` function, we'd have to make sure `_slots` is correctly updated on error

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
walley892: Nice

engine/src/flutter/testing/dart/fragment_shader_test.dart
gaaclarke: yea, i ran into that too. it can be solved like this: https://github.com/flutter/flutter/pull/179906/changes#diff-7db4133a0110a4083d4d45e1d327ea06d9dfe9743ffcd3a4946ac2c306704d33R46

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
walley892: I more meant that it felt icky because we have to redefine these classes here at all. Ah well

gaaclarke: Yea, the benefit of making private is that we can tweak them later if we want.  They won't be written in stone.

eyebrowsoffire (approved):
Web changes look good!

engine/src/flutter/testing/dart/fragment_shader_test.dart
walley892: Resolving to solve this later with https://github.com/flutter/flutter/issues/179966

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5671,7 +5715,7 @@ base class FragmentShader extends Shader {
   /// ```dart
   /// void updateShader(ui.FragmentShader shader) {
   ///   shader.getUniformFloat('uScale');
-  ///   shader.getUniformVec3('uTime');
+  ///   shader.getUniformVec3('uScaledTime');
```
gaaclarke, line 5785:
```suggestion
///   shader.getUniformVec3('uTime');
````

**Review by gaaclarke — APPROVED** [review]
lgtm!

File: engine/src/flutter/lib/ui/painting.dart [hunk 12]
```diff
@@ -5373,6 +5373,22 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
     return index;
   }
 
+  int _getUniformVectorSize(String name) {
+    const sizeOfFloat = 4;
+    var sizeInFloats = 0;
+
+    for (final Object? entryDynamic in _uniformInfo) {
+      final entry = entryDynamic! as Map<String, Object>;
+      sizeInFloats = (entry['size'] as int? ?? 0) ~/ sizeOfFloat;
```
flar: Move that inside the test.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5517,6 +5614,21 @@ base class FragmentShader extends Shader {
     _floats = _constructor(_program, _program._uniformFloatCount, _program._samplerCount);
   }
 
+  List<UniformFloatSlot> _getSlotsForUniform(String name, int size) {
+    final int dataSize = _program._getUniformVectorSize(name);
```
flar: This implementation is causing 2 lookups by name, can they be combined? It's an internal function so what about having the getIndex function take the size and it gets verified before returning the index (~size of 0 or -1~ null to indicate "I don't care about the size")?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5597,6 +5709,66 @@ base class FragmentShader extends Shader {
...
+  /// ```
+  ///
+  /// ```dart
+  /// void updateShader(ui.FragmentShader shader) {
+  ///   shader.getUniformFloat('uScale');
+  ///   shader.getUniformVec2('uMagnitude');
+  /// }
+  /// ```
+  UniformVec2Slot getUniformVec2(String name) {
+    final List<UniformFloatSlot> slots = _getSlotsForUniform(name, 2);
```
flar, line 5769: What's the cost of creating a list in the _getSlots function vs. just returning the index (after verifying the size) and then constructing the slots inline here?

flar (changes requested):
Some comments about the efficiency of the implementation.

Also, I see a lot of negative tests (fails if ...), but I don't see some positive tests that it actually returns the object if you make the right request. Did I miss something?

engine/src/flutter/lib/ui/painting.dart
walley892: Which test?

engine/src/flutter/lib/ui/painting.dart
walley892: I'm not sure if there's a differing cost for that kind of thing in Dart. I have it this way for code reuse

engine/src/flutter/lib/ui/painting.dart
flar: Actually, the way to do an optional parameter in Dart is `int?` and pass null for the access that doesn't require a size constraint.

engine/src/flutter/lib/ui/painting.dart
walley892: Actually, I'm not sure if this comment is dovetailing with your comment above or if it's a comment about the efficiency of returning lists in Dart

engine/src/flutter/lib/ui/painting.dart
flar: You only need the value if the following if test triggers.

engine/src/flutter/lib/ui/painting.dart
flar: It's about unnecessary creation of lists. Maybe Dart is smart enough to optimize that out, I don't know.

engine/src/flutter/lib/ui/painting.dart
walley892: Ah I see. Function removed anyhow

engine/src/flutter/lib/ui/painting.dart
walley892: How's that looking?

flar (approved):
LGTM

PR #167762, opened 2025-04-24
Update default `selectionHeightStyle` and `selectionWidthStyle` for `EditableText`

Renzo-Olivares:
This change updates the default `selectionHeightStyle` and `selectionWidthStyle` for `EditableText` to more closely match the behavior on native platforms.

```dart
/// The default value for [selectionHeightStyle].
///
/// On web platforms, this defaults to [ui.BoxHeightStyle.max].
///
/// On native platforms, this defaults to [ui.BoxHeightStyle.includeLineSpacingMiddle] for all
/// platforms.
static ui.BoxHeightStyle get defaultSelectionHeightStyle {
  if (kIsWeb) {
    return ui.BoxHeightStyle.max;
  }
  switch (defaultTargetPlatform) {
    case TargetPlatform.android:
    case TargetPlatform.fuchsia:
    case TargetPlatform.iOS:
    case TargetPlatform.linux:
    case TargetPlatform.macOS:
    case TargetPlatform.windows:
      return ui.BoxHeightStyle.includeLineSpacingMiddle;
  }
}

/// The default value for [selectionWidthStyle].
///
/// On web platforms, this defaults to [ui.BoxWidthStyle.max] for Apple platforms and
/// [ui.BoxWidthStyle.tight] for all others.
///
/// On non-web platforms, this defaults to [ui.BoxWidthStyle.max] for all
/// platforms.
static ui.BoxWidthStyle get defaultSelectionWidthStyle {
  if (kIsWeb) {
    if (defaultTargetPlatform == TargetPlatform.iOS || WebBrowserDetection.browserIsSafari) {
      return ui.BoxWidthStyle.max;
    }
    return ui.BoxWidthStyle.tight;
  }
  switch (defaultTargetPlatform) {
    case TargetPlatform.android:
    case TargetPlatform.fuchsia:
    case TargetPlatform.iOS:
    case TargetPlatform.linux:
    case TargetPlatform.macOS:
    case TargetPlatform.windows:
      return ui.BoxWidthStyle.max;
  }
}
```

### Android Native Behavior

| Component Type | UI Framework / Component | Image |
|---|---|---|
| **Static Text** | Android TextView | native_android_static_textview |
| **Static Text** | Jetpack Compose Text | native_android_static_compose |
| **Static Text** | Chrome (Web) "p element" | android_chrome_web |
| **Text Field** | Android EditText | native_android_textfield_edittext |
| **Text Field** | Jetpack Compose TextField | native_android_textfield_compose |

[...]

PR #179928, opened 2025-12-16
fix: text selection two handles directionality.

### LINKED ISSUE #187799: Text selection handles face wrong direction in RTL app with LTR text
### Steps to reproduce

1. Run a Flutter app with ambient `TextDirection.rtl` (e.g. Arabic locale).
2. Focus a single-line text field and enter LTR text (e.g. English).
3. Select the text to show the selection handles.

### Expected results

Selection handles should align with the selected text and point toward the selection on both ends.

### Actual results

Selection handles render with the wrong orientation/position relative to the selected text when the ambient directionality is RTL and the selected text is LTR.

Screenshot showing misaligned selection handles in an RTL app with LTR text

### Environment

- Platform: Android (observed on device)
- App locale/direction: RTL (Arabic UI)
- Selected text direction: LTR (English)

### Additional context

A fix is proposed in https://github.com/flutter/flutter/pull/179928.

muhammadkamel:
Fixes #187799

This PR fixes an issue where text selection handles would render facing the wrong direction when the text directionality differed from the ambient directionality (e.g., selecting English (LTR) text within an application set to Arabic (RTL)), and also when a single selection spans both LTR and RTL runs.

Previously, `TextSelectionOverlay` determined the start and end handle types based primarily on the ambient `TextDirection` and the logical selection range. This caused handles to swap (pointing away from the selection) in mixed-directionality scenarios.

This change updates the logic in `_updateSelectionOverlay` to derive each handle's type from the **per-endpoint `TextDirection`** reported by `TextSelectionPoint.direction`, falling back to the render object's `textDirection` when an endpoint does not provide one:

* Start handle: `TextDirection.ltr` → `TextSelectionHandleType.left`, `TextDirection.rtl` → `TextSelectionHandleType.right`.
* End handle: `TextDirection.ltr` → `TextSelectionHandleType.right`, `TextDirection.rtl` → `TextSelectionHandleType.left`.

Because each endpoint is evaluated independently, selections that cross a directionality boundary (e.g. `"abc مرحبا"`) get handle types that match the local direction at each end, so both handles continue to "hug" the selection.

On iOS, UIKit conventionally keeps selection handles aligned with the field's own direction rather than the per-endpoint direction. To preserve platform behavior, when `defaultTargetPlatform == TargetPlatform.iOS` both handles use the render object's `textDirection` instead of the endpoint directions.

When a non-collapsed selection yields fewer than two endpoints (e.g. the selection is scrolled or clipped out of view), handle direction falls back to the field's `textDirection` instead of asserting.

Collapsed selections continue to use `TextSelectionHandleType.collapsed` for both handles and are unaffected by this change.

**Tests added**

Added new tests in `text_selection_test.dart`:

Android (`TargetPlatformVariant.only(TargetPlatform.android)`), driven by a table of `_DirectionalityTestCase`s that assert the endpoint directions reported by `RenderEditable` and the handle types received by a spy `TextSelectionControls`:

* ambient LTR, English text
* ambient RTL, English text
* ambient RTL, Arabic text
* ambient LTR, Arabic text
* ambient LTR, mixed `"abc مرحبا"` (English then Arabic)
* ambient RTL, mixed `"abc مرحبا"` (English then Arabic)

iOS (`TargetPlatformVariant.only(TargetPlatform.iOS)`):

* Verifies that for the mixed `"abc مرحبا"` selection under ambient LTR, both handles follow the field's `textDirection` (i.e. `left` / `right`) rather than the per-endpoint directions — exercising the UIKit-compatibility branch.

The new directionality tests set `selectAllOnFocus: false` explicitly (with an inline comment) because web CI defaults `EditableText.selectAllOnFocus` to `true`, which can select all text on focus and prevent handle rebuilds in these tests.

## Pre-launch Checklist

- [x] I signed the [CLA].

image_issue


muhammadkamel:
Hi @dkwingsmt,
Could you please assign a reviewer for this PR?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,22 +540,41 @@ class TextSelectionOverlay {
   }
 
   void _updateSelectionOverlay() {
+    final List<TextSelectionPoint> endpoints = renderObject.getEndpointsForSelection(_selection);
+    TextSelectionHandleType startHandleType = TextSelectionHandleType.collapsed;
+    TextSelectionHandleType endHandleType = TextSelectionHandleType.collapsed;
+
+    if (!_selection.isCollapsed) {
+      if (endpoints.length == 2) {
+        if (endpoints[0].point.dx <= endpoints[1].point.dx) {
```
Renzo-Olivares: I think we can avoid this additional level of nesting here if we make this case a boolean, and use a ternary operator in the handle assignments below..

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1899,6 +1901,132 @@ void main() {
...
+            ),
+          ),
+        );
+
+        final RenderEditable renderEditable = tester.allRenderObjects
+            .whereType<RenderEditable>()
+            .first;
+        expect(renderEditable.textDirection, testCase.ambientDirection);
+
+        // Focus and set selection
```
Renzo-Olivares: nit: Add a period to the end of the comments in the added tests.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1899,6 +1901,132 @@ void main() {
...
+  for (final testCase in directionalityTestCases) {
+    testWidgets(
+      'handles point correctly: ${testCase.description}',
+      (WidgetTester tester) async {
+        final customControls = DirectionalitySpyTextSelectionControls();
+        final controller = TextEditingController(text: testCase.text);
+        final focusNode = FocusNode();
+
+        await tester.pumpWidget(
+          MaterialApp(
```
Renzo-Olivares: I think we want to avoid using `Material` in the widgets library moving forward (including `TextField`). However, in this case the infrastructure (helper test widgets) are not yet in place. I am working on a test text field https://github.com/flutter/flutter/pull/180494/ but it do not have a solution for testing selection handles yet. A test `WidgetsApp` is also being worked on here https://github.com/flutter/flutter/pull/180456. fyi @loic-sharma @justinmc

Renzo-Olivares (commented):
Thank you for the contribution @muhammadkamel, this change definitely helps improve the text selection experience.

I verified that this is also the behavior on native mobile platforms.

#### Jetpack Compose (Android) with RTL set for application.




#### SwiftUI (iOS) with RTL set at application level

Screenshot 2026-01-06 at 5 32 45 PM

WRT to RTL text in general I think this also improves the experience slightly but text selection is still a bit buggy in RTL when using the text selection handles. The magnifier also seems to be broken when the application level text directionality is RTL and the textfield is LTR. That can be improved in a separate PR.

packages/flutter/lib/src/widgets/text_selection.dart
Renzo-Olivares: Giving this another thought, I don't think the logic here is robust enough to handle multi line scenarios. See the example below.

packages/flutter/test/widgets/text_selection_test.dart
muhammadkamel: Will keep it until the PR https://github.com/flutter/flutter/pull/180456 merged.

packages/flutter/lib/src/widgets/text_selection.dart
muhammadkamel: Can you please check now

Renzo-Olivares: The updated PR fixes the selection handle directionality for LTR TextField in LTR app, but still remains broken for LTR Textfield in RTL app.

Renzo-Olivares:
Hi @muhammadkamel, just wanted to follow up if there is any updates on this PR I should review? During my review last week I was still observing issues https://github.com/flutter/flutter/pull/179928#discussion_r2683636141. Let me know if I should take a look again.

muhammadkamel:
> Hi @muhammadkamel, just wanted to follow up if there is any updates on this PR I should review? During my review last week I was still observing issues [#179928 (comment)](https://github.com/flutter/flutter/pull/179928#discussion_r2683636141). Let me know if I should take a look again.

Hi @Renzo-Olivares 
I will work on fixing this behavior issue. But I have a question, should the handle (Left) or (Right) direction changed base on the text or on the app locale (Directionality)?

Renzo-Olivares:
Hi @muhammadkamel, thank you for your patience on this one. After confirming some of the native behavior on Jetpack compose and Android View API's, I think we are able to closely match the native behavior with your previous change in https://github.com/flutter/flutter/pull/179928/changes/7ff99718624081dec65a6dc9937f52fad3e69a8e . Checking the `TextDirection` of the `TextSelectionPoint` end points should be enough as you did in that commit. 

I was mistaken about the multi-line behavior before in https://github.com/flutter/flutter/pull/179928#discussion_r2683636141. Though I still think there may be a bug when moving the selection across multiple lines in the scenario where the App's directionality is RTL but the text is LTR, in the linked image the selection originally started on the first line on the word "world" but moving that selection to the line below, adds the "?" on the first line to the selection. This can be solved in a different PR though as the issue existed before this change.

I think some of the recent additional changes like passing the directionality to the `TextSelectionOverlay`, and `_SelectionHandleOverlay` are not needed. Can you explain why the changes for `getEndPointsForSelection` are needed and `EditableText`?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1952,19 +1997,16 @@ class _SelectionToolbarWrapperState extends State<_SelectionToolbarWrapper>
...
-          textDirection: Directionality.of(this.context),
-          child: FadeTransition(
-            opacity: _opacity,
-            child: CompositedTransformFollower(
-              link: widget.layerLink,
-              showWhenUnlinked: false,
-              offset: widget.offset,
-              child: widget.child,
-            ),
+    return TextFieldTapRegion(
```
Renzo-Olivares: The outer `TapRegion` here is necessary for the context menu to work properly with `SelectableRegion`.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -1125,6 +1155,21 @@ class SelectionOverlay {
         : _toolbar != null || _spellCheckToolbarController.isShown;
   }
 
+  /// The text direction of the text being selected.
+  ///
+  /// This is used to determine the direction of the selection handles.
+  ///
+  /// Changing the value while the handles are visible causes them to rebuild.
+  TextDirection? get textDirection => _textDirection;
```
Renzo-Olivares: This member and its equivalent in `TextSelectionOverlay` doesn't seem to be used by anything, is this needed?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,23 +541,49 @@ class TextSelectionOverlay {
   }
 
   void _updateSelectionOverlay() {
+    final List<TextSelectionPoint> endpoints = renderObject.getEndpointsForSelection(_selection);
```
Renzo-Olivares, line 551: I suggest something along these lines

```dart
  void _updateSelectionOverlay() {
    final List<TextSelectionPoint> endpoints = renderObject.getEndpointsForSelection(_selection);
    late final TextSelectionHandleType startHandleType;
    late final TextSelectionHandleType endHandleType;
    if (_selection.isCollapsed) {
      startHandleType = TextSelectionHandleType.collapsed;
      endHandleType = TextSelectionHandleType.collapsed;
    } else {
      assert(endpoints.length == 2);
      final TextDirection startHandleDirection =
          endpoints.first.direction ?? renderObject.textDirection;
      final TextDirection endHandleDirection =
          endpoints.last.direction ?? renderObject.textDirection;
      startHandleType = switch (startHandleDirection) {
        TextDirection.ltr => TextSelectionHandleType.left,
        TextDirection.rtl => TextSelectionHandleType.right,
      };
      endHandleType = switch (endHandleDirection) {
        TextDirection.ltr => TextSelectionHandleType.right,
        TextDirection.rtl => TextSelectionHandleType.left,
      };
    }
    _selectionOverlay
      // Update selection handle metrics.
      ..startHandleType = startHandleType
      ..lineHeightAtStart = _getStartGlyphHeight()
      ..endHandleType = endHandleType
      ..lineHeightAtEnd = _getEndGlyphHeight()
      // Update selection toolbar metrics.
      ..selectionEndpoints = endpoints
      ..toolbarLocation = renderObject.lastSecondaryTapDownPosition;
  }
```

It is also important to note that this behavior does not match the native iOS behavior. From my observation on iOS 26, when the text direction is not specified the text field will try to automatically detect the text direction from the text. It looks like it does this by checking the first character of the text, I tried a long piece of english text that ends with an arabic character, and a long piece of english text that starts with an arabic character. The former scenario resulted resulted in LTR directionality for the entire field, and the latter scenario resulted in RTL directionality for the entire field. In BiDi text the selection handle directionality does not change when the selection spreads across both LTR and RTL text, adhering to the inferred directionality. When the text direction is forced, the selection handles adhere to the forced directionality.

I don't think we have to handle the iOS behavior in this PR, but iOS should always use `renderObject.textDirection` similar to how it did before the changes in this PR so the changes are targeted for Android.

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for the quick changes, I left a few comments. I'm also still seeing the changes to `getEndPointsForSelection` and `EditableText`.

packages/flutter/lib/src/rendering/editable.dart
```diff
@@ -1748,15 +1748,45 @@ class RenderEditable extends RenderBox
       final Offset start = Offset(0.0, preferredLineHeight) + caretOffset + paintOffset;
       return <TextSelectionPoint>[TextSelectionPoint(start, null)];
     } else {
+      final int startOffset = selection.start;
```
Renzo-Olivares: Can you explain why these changes are needed. From what I can see this logic was likely changed so the endPoints have an accurate `textDirection`. Was there cases where the default logic for `getEndPointsForSelection` was not accurately reporting the `textDirection`?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2095,6 +2096,9 @@ class EditableText extends StatefulWidget {
       }
       return ui.BoxWidthStyle.tight;
     }
+    if (defaultTargetPlatform == TargetPlatform.android) {
```
Renzo-Olivares: This was recently changed in https://github.com/flutter/flutter/pull/167762 after an audit of the selection style on native platforms. Is this based on the behavior you are observing on Android?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,22 +540,45 @@ class TextSelectionOverlay {
...
+    late final TextSelectionHandleType startHandleType;
+    late final TextSelectionHandleType endHandleType;
+
+    if (_selection.isCollapsed) {
+      startHandleType = TextSelectionHandleType.collapsed;
+      endHandleType = TextSelectionHandleType.collapsed;
+    } else {
+      final TextDirection textDirection = renderObject.textDirection;
+      // UIKit keeps selection handles aligned with the field direction.
+      final bool useTextDirectionForSelectionHandles = defaultTargetPlatform == TargetPlatform.iOS;
```
Renzo-Olivares: nit: maybe `useRenderObjectDirectionForSelectionHandles` or `preferRenderObjectDirectionForSelectionHandles`.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,22 +540,45 @@ class TextSelectionOverlay {
...
+      final TextDirection startHandleDirection = useTextDirectionForSelectionHandles
```
Renzo-Olivares: I suggest the code below.

```dart
late final TextDirection startHandleDirection;
late final TextDirection endHandleDirection;
if (useTextDirectionForSelectionHandles) {
  startHandleDirection = textDirection;
  endHandleDirection = textDirection;
} else {
  startHandleDirection = endpoints.first.direction ?? textDirection;
  endHandleDirection = endpoints.last.direction ?? textDirection;
}
```

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,22 +540,45 @@ class TextSelectionOverlay {
...
+    assert(endpoints.isNotEmpty);
+
+    late final TextSelectionHandleType startHandleType;
+    late final TextSelectionHandleType endHandleType;
+
+    if (_selection.isCollapsed) {
+      startHandleType = TextSelectionHandleType.collapsed;
+      endHandleType = TextSelectionHandleType.collapsed;
+    } else {
+      final TextDirection textDirection = renderObject.textDirection;
```
Renzo-Olivares, line 560: Suggest adding an assert here at the top of this case to ensure endPoints.length == 2, `getEndPointsForSelection` should always be returning 2 end points for an uncollapsed selection so we should catch that if it's not.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -540,22 +540,45 @@ class TextSelectionOverlay {
...
+      // UIKit keeps selection handles aligned with the field direction.
+      final bool useTextDirectionForSelectionHandles = defaultTargetPlatform == TargetPlatform.iOS;
+      final TextDirection startHandleDirection = useTextDirectionForSelectionHandles
+          ? textDirection
+          : endpoints.first.direction ?? textDirection;
+      final TextSelectionPoint? endPoint = endpoints.length > 1 ? endpoints.last : null;
```
Renzo-Olivares: Instead of this fallback logic I think the assert I suggested at the top might be sufficient to catch if the endPoints being returned are less than 2.

packages/flutter/lib/src/widgets/editable_text.dart
muhammadkamel: This issue occurs with multi-line Arabic (RTL) text: selecting a single word sometimes selects the entire line instead—particularly for words in middle lines (not the last line). Updating BoxWidthStyle to tight fixes the problem.

justinmc:
@muhammadkamel FYI you have a branch conflict to resolve here.

muhammadkamel:
> @muhammadkamel FYI you have a branch conflict to resolve here.

Hi @justinmc 

Thanks!
I have resolved the conflict.

packages/flutter/lib/src/widgets/editable_text.dart
muhammadkamel: @Renzo-Olivares 
I will fix this issue in another PR

packages/flutter/lib/src/rendering/editable.dart
muhammadkamel: You're right. I reverted the RenderEditable/getEndpointsForSelection changes.

justinmc:
@muhammadkamel Looks like there is another conflict (sorry for all the maintenance work here).

muhammadkamel:
> @muhammadkamel Looks like there is another conflict (sorry for all the maintenance work here).

No problem! Done!

packages/flutter/lib/src/widgets/editable_text.dart
Renzo-Olivares: fyi, this is already being fixed in https://github.com/flutter/flutter/pull/185239.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1917,6 +1917,173 @@ void main() {
...
+      ambientDirection: TextDirection.ltr,
+      text: 'abc مرحبا',
+      selectionBase: 0,
+      selectionExtent: 9,
+      expectedStartEndpointDirection: TextDirection.ltr,
+      expectedEndEndpointDirection: TextDirection.rtl,
+      expectedStartHandleType: TextSelectionHandleType.left,
+      expectedEndHandleType: TextSelectionHandleType.left,
+    ),
+    _DirectionalityTestCase(
```
Renzo-Olivares, line 2147: I think this last case should be:

```dart
expectedStartEndpointDirection: TextDirection.rtl,
expectedEndEndpointDirection: TextDirection.ltr,
expectedStartHandleType: TextSelectionHandleType.right,
expectedEndHandleType: TextSelectionHandleType.right,
```

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -1917,6 +1917,173 @@ void main() {
...
+        controller.selection = TextSelection(
+          baseOffset: testCase.selectionBase,
+          extentOffset: testCase.selectionExtent,
+        );
+        await tester.pumpAndSettle();
+
+        final List<TextSelectionPoint> endpoints = renderEditable.getEndpointsForSelection(
+          controller.selection,
+        );
+        expect(endpoints.length, 2);
```
Renzo-Olivares: nit: consider `expect(endpoints, hasLength(2));` similar to the ios test.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -2146,6 +2313,77 @@ class FakeTextSelectionDelegate extends Fake implements TextSelectionDelegate {
...
 }
 
+class DirectionalitySpyTextSelectionControls extends TextSelectionControls {
+  final List<TextSelectionHandleType> builtHandleTypes = <TextSelectionHandleType>[];
+
+  void clearBuiltHandleTypes() {
+    builtHandleTypes.clear();
+  }
+
+  TextSelectionHandleType get lastStartHandleType => builtHandleTypes[builtHandleTypes.length - 2];
```
Renzo-Olivares: I would just check `builtHandleTypes` directly. That way you can also do `expect(builtHandleTypes, hasLength(2))` before querying `builtHandleTypes.first/last`.

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for your patience on this one! Left a few small comments. Could you also update the PR description to more accurately fit the current implementation. There are also test/analyzer failures that need to be addressed.

packages/flutter/test/widgets/text_selection_test.dart
muhammadkamel: Good catch!

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for the changes. There is still a comment that needs to be addressed https://github.com/flutter/flutter/pull/179928#discussion_r3256454929.

muhammadkamel:
> Hi @muhammadkamel, thank you for the changes. There is still a comment that needs to be addressed [#179928 (comment)](https://github.com/flutter/flutter/pull/179928#discussion_r3256454929).

Check now please

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -2146,6 +2313,73 @@ class FakeTextSelectionDelegate extends Fake implements TextSelectionDelegate {
...
+  // Wrap the handle in a widget with a Key that identifies its type.
+  @override
+  Widget buildHandle(
+    BuildContext context,
+    TextSelectionHandleType type,
+    double textLineHeight, [
+    VoidCallback? onTap,
+  ]) {
+    builtHandleTypes.add(type);
+    return SizedBox.square(key: ValueKey<TextSelectionHandleType>(type), dimension: textLineHeight);
```
Renzo-Olivares: I don't think this key is being used. I think we can just remove it.

Renzo-Olivares (dismissed):
LGTM with one final nit. Could you also update the PR description to more accurately fit the current implementation. This will need a secondary reviewer and they may be confused by the PR description since it does not match the current implementation.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -548,22 +548,52 @@ class TextSelectionOverlay {
   }
 
   void _updateSelectionOverlay() {
+    final List<TextSelectionPoint> endpoints = renderObject.getEndpointsForSelection(_selection);
+    assert(endpoints.isNotEmpty);
+
+    late final TextSelectionHandleType startHandleType;
```
LongCatIsLooong: here and below: late final -> final

LongCatIsLooong (dismissed):
LGTM with nits.

muhammadkamel:
> LGTM with one final nit. Could you also update the PR description to more accurately fit the current implementation. This will need a secondary reviewer and they may be confused by the PR description since it does not match the current implementation.

I've updated the PR description to reflect the current implementation.

Renzo-Olivares (dismissed):
LGTM, thank you for your contribution @muhammadkamel!

Renzo-Olivares:
@muhammadkamel, heads up you will need to fix the analyzer issues here. as well as any failing tests before this PR lands.

muhammadkamel:
> @muhammadkamel, heads up you will need to fix the analyzer issues here. as well as any failing tests before this PR lands.

Could you check now please.

Renzo-Olivares:
Hi @muhammadkamel, there are still some failing tests that need to be addressed before landing this PR. Thank you!

muhammadkamel:
> Hi @muhammadkamel, there are still some failing tests that need to be addressed before landing this PR. Thank you!

Hi @Renzo-Olivares 
You're welcome! :)
Please check once again.

muhammadkamel:
Hi @Renzo-Olivares 

Is there anything else needed from me to land this PR?

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -548,22 +548,47 @@ class TextSelectionOverlay {
...
+      startHandleType = TextSelectionHandleType.collapsed;
+      endHandleType = TextSelectionHandleType.collapsed;
+    } else {
+      final TextDirection textDirection = renderObject.textDirection;
+      // UIKit keeps selection handles aligned with the field direction.
+      final preferRenderObjectDirectionForSelectionHandles =
+          defaultTargetPlatform == TargetPlatform.iOS;
+      final TextDirection startHandleDirection;
+      final TextDirection endHandleDirection;
+      if (preferRenderObjectDirectionForSelectionHandles || endpoints.length < 2) {
```
Renzo-Olivares, line 577: I would add a comment here on why a non-collapsed selection may have less than two endpoints.

packages/flutter/test/widgets/editable_text_tester.dart
```diff
@@ -39,6 +39,7 @@ class TestTextField extends StatefulWidget {
     this.style,
     this.controller,
     this.onSubmitted,
+    this.selectAllOnFocus,
```
Renzo-Olivares, line 44: Is this necessary? `selectAllOnFocus` is false by default on mobile platforms.

Renzo-Olivares (commented):
Hi @muhammadkamel, thank you for your patience on this one. I had a few small questions but otherwise looks good.

packages/flutter/test/widgets/editable_text_tester.dart
muhammadkamel: You're right that it's false by default on Android/iOS. We still need it because these tests also run on web CI, where the default is true. I'll default TestTextField to false so tests don't have to pass it explicitly.

Renzo-Olivares: Looking at the test i'm curious why it failed in the first place. In the test we request the focus, after we clear the built handles and set a new selection. Shouldn't our new selection win in this case, or what was there a different point of failure?

Renzo-Olivares:
Thank you for the changes @muhammadkamel, had a small question about the failing web tests https://github.com/flutter/flutter/pull/179928#discussion_r3398751730.

packages/flutter/test/widgets/editable_text_tester.dart
muhammadkamel: Good question — the explicit selection does win, but only when it actually changes the range.

On web, `selectAllOnFocus` defaults to `true`, so focus already selects the full text. For the mixed-directionality tests (and the iOS test), our intended selection is also the full range, so assigning it again is a no-op and no handles are rebuilt after `clearBuiltHandleTypes()`.

The partial-selection Android cases pass because focus selects all, then the test sets a shorter range, which triggers a rebuild. Defaulting `TestTextField` to `selectAllOnFocus: false` avoids the focus-time select-all so the explicit assignment always changes the selection.

Renzo-Olivares: Sounds good to me, can you add a comment in the test next to `selectAllOnFocus` describing why it needs to be disabled.

Renzo-Olivares (dismissed):
LGTM modulo small comment https://github.com/flutter/flutter/pull/179928#discussion_r3438272193. This PR will need to be rebased before landing.

muhammadkamel:
Hi @Renzo-Olivares,

Could you please take another look when you have a chance?

packages/flutter/test/widgets/editable_text_tester.dart
```diff
@@ -39,6 +39,9 @@ class TestTextField extends StatefulWidget {
     this.style,
     this.controller,
     this.onSubmitted,
+    // Keep test behavior consistent across platforms, including web CI where
+    // EditableText otherwise defaults selectAllOnFocus to true.
+    this.selectAllOnFocus = false,
```
Renzo-Olivares: Apologies, what I meant before is that this should not have a default, i.e. should inherit `EditableText` behavior. And in the tests we added we should set it to `false` explicitly and add a comment next to it as to why.

muhammadkamel: Done — removed the default from TestTextField so it inherits EditableText behavior. The new directionality tests now set selectAllOnFocus: false explicitly with an inline comment explaining the web CI behavior.

packages/flutter/test/widgets/editable_text_tester.dart
muhammadkamel: Done — removed the default from TestTextField so it inherits EditableText behavior. The new directionality tests now set selectAllOnFocus: false explicitly with an inline comment explaining the web CI behavior.

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -18794,6 +18802,60 @@ void main() {
     // [intended] only applies to platforms where we supply the context menu.
     skip: kIsWeb,
   );
+
+  testWidgets(
```
Renzo-Olivares: I don't think the tests changes in `editable_text_test.dart` are related to the changes in this PR. If so we should probably revert/remove them.

packages/flutter/test/widgets/editable_text_tester.dart
```diff
@@ -60,6 +60,11 @@ class TestTextField extends StatefulWidget {
   final ValueChanged<String>? onSubmitted;
   final bool showSelectionHandles;
 
+  /// Controls whether all text is selected when the field receives focus.
+  ///
+  /// When null, [EditableText] platform defaults apply.
+  final bool? selectAllOnFocus;
```
Renzo-Olivares, line 67: This is never set in the constructor. I think we accidentally removed it, we should add it back.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -2088,6 +2088,179 @@ void main() {
...
+
+        await tester.pumpWidget(
+          TestWidgetsApp(
+            home: Directionality(
+              textDirection: testCase.ambientDirection,
+              child: TestTextField(
+                controller: controller,
+                focusNode: focusNode,
+                selectionControls: customControls,
+                // Web CI defaults EditableText.selectAllOnFocus to true, which
```
Renzo-Olivares: I suggest something along the lines of, in this test and the other test where we set `selectAllOnFocus`:

```
On the web selectAllOnFocus defaults to true, interfering with this tests programmatic select all.
```

muhammadkamel:
Hi @Renzo-Olivares 

Addressed your latest comments — please take another look when you get a chance.

Renzo-Olivares:
@muhammadkamel can you try to rebase this to the latest `master`. Tests seem to be stuck at the moment.

packages/flutter/test/widgets/text_selection_test.dart
Renzo-Olivares: I think this comment was not resolved yet, we should remove the unused key.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -587,7 +615,7 @@ class TextSelectionOverlay {
   /// See also:
   ///
   ///   * [spellCheckToolbarIsVisible], which is only whether the spell check menu
-  ///     specifically is visible.
+  ///     is visible.
```
Renzo-Olivares: We should revert this unrelated change.

packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -548,22 +548,50 @@ class TextSelectionOverlay {
...
+      startHandleType = TextSelectionHandleType.collapsed;
+      endHandleType = TextSelectionHandleType.collapsed;
+    } else {
+      final TextDirection textDirection = renderObject.textDirection;
+      // UIKit keeps selection handles aligned with the field direction.
+      final preferRenderObjectDirectionForSelectionHandles =
+          defaultTargetPlatform == TargetPlatform.iOS;
+      final TextDirection startHandleDirection;
+      final TextDirection endHandleDirection;
+      // A non-collapsed selection may still yield a single endpoint when
```
Renzo-Olivares: I wasn't able to write a test that actually reproduced this with a scrolled or clipped out of view selection.

I think the comment below would be more accurate.

```
// A non-collapsed selection might return fewer than two endpoints if the 
// text layout lacks boxes for the selected range. This typically happens when:
//
//  * Render lag: The overlay updated with a new editing value before the 
//    render object laid out the new text (selection offsets are out of bounds).
//  * Split graphemes: A selection boundary falls inside a multi-code-unit 
//    cluster (like an emoji or combining character).
//  * Degenerate layout: The layout is temporarily squashed (e.g., 
//    preferredLineHeight is 0 during a fold transition).
//
// In these cases, we fall back to the field's textDirection.
```

muhammadkamel:
Hi @Renzo-Olivares — addressed your latest comments:

Removed the unused ValueKey from DirectionalitySpyTextSelectionControls (the spy already asserts via builtHandleTypes).
Reverted the unrelated toolbarIsVisible doc change.
Updated the < 2 endpoints fallback comment to your suggested wording (render lag / split graphemes / degenerate layout).
Please take another look when you have a chance.

packages/flutter/test/widgets/text_selection_test.dart
```diff
@@ -2370,3 +2543,70 @@ class _MockTextSelectionHandleControls extends TextSelectionControls
     );
   }
 }
+
+class DirectionalitySpyTextSelectionControls extends TextSelectionControls {
```
Renzo-Olivares: nit: I'd make this class private for now until it needs to be exposed to other tests.

Renzo-Olivares (dismissed):
LGTM w/ one small nit. Thank you for your patience on this one!

packages/flutter/test/widgets/text_selection_test.dart
muhammadkamel: Check now!

Renzo-Olivares (approved):
RE-LGTM, thank you for the contribution @muhammadkamel!

muhammadkamel:
> RE-LGTM, thank you for the contribution @muhammadkamel!

You're welcome! :)

LongCatIsLooong (approved):

Issue #148915, opened 2022-03-22
Duplicate browsers start for test given on 'An introduction to integration testing' page


zpsean:
See linked issue for more context: https://github.com/flutter/flutter/issues/100447
[Web][Integration Test] Two Chrome instances launched, and the unnecessary Chrome instance breaks the foundation of the integration test [Edited]

### Page URL

https://docs.flutter.dev/cookbook/testing/integration/introduction/

### Page source

https://github.com/flutter/website/tree/main/src/cookbook/testing/integration/introduction.md

### Describe the problem

 After executing the command:

```
flutter drive \
  --driver=test_driver/integration_test.dart \
  --target=integration_test/app_test.dart \
  -d chrome
```
One Chrome is opened. Notice that it is not labeled with "chrome is controlled by automated test software".

<img width="1200" alt="image" src="https://user-images.githubusercontent.com/14213337/159384051-84080cd6-e34c-4ff6-ad71-b074d7dadd36.png">

Then try to execute the command plus arg `--no-headless`:

```
flutter drive \
  --driver=test_driver/integration_test.dart \
  --target=integration_test/app_test.dart \
  -d chrome \
  --no-headless 
```

Two Chrome instances are launched:
- One [backend one] is launched as a device [this is NOT intended for test cases execution]
- The other [frontend one] is launched via webdriver (chromedriver this case) [this is intended for test cases execution, labeled with 'chrome is controlled by automated test software']

[**Note**: Mostly, the two Chromes are overlaid, please try to drag it away (up or down), then back one would be visible.]

<img width="1641" alt="image" src="https://user-images.githubusercontent.com/14213337/159163845-67e37dfb-b21d-474d-96ee-7d4c711fa140.png">

### Expected fix

Issue code raised: https://github.com/flutter/flutter/issues/100447
[Web][Integration Test] Two Chrome instances launched, and the unnecessary Chrome instance breaks the foundation of the integration test [Edited]

Before issue 100447 is fixed. Please kindly add the above info to this page to clear the context to avoid misleading users.

PR #179935, opened 2025-12-16
[flutter_tools] Pass --chrome-binary to app hosting Chrome in flutter drive

shindonghwi:
Fixes #171504

## Problem
When running `flutter drive -d chrome --chrome-binary=/path/to/chrome`, the `--chrome-binary` flag was only being passed to WebDriver Chrome, not to the app hosting Chrome instance.

## Solution
Added `webChromeBinary` field to `DebuggingOptions` and passed it through to `ChromiumLauncher.launch()`.

## Changes
- Add `webChromeBinary` field to `DebuggingOptions`
- Add `chromeBinary` parameter to `ChromiumLauncher.launch()`
- Forward `webChromeBinary` in `WebDriverService.start()`
- Add test for custom chrome binary path

## Test
```
flutter drive -d chrome --chrome-binary="/custom/path/chrome" -v
```
Verified that log shows `Will use Chromium executable at /custom/path/chrome`.

### DIFF
--- a/packages/flutter_tools/lib/src/commands/run.dart
+++ b/packages/flutter_tools/lib/src/commands/run.dart
@@ -285,6 +285,7 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
   Future<DebuggingOptions> createDebuggingOptions({
     WebDevServerConfig? webDevServerConfig,
     String? webChromeBinary,
+    bool webNoLaunchChrome = false,
   }) async {
     final BuildInfo buildInfo = await getBuildInfo();
     final int? webBrowserDebugPort =
@@ -315,6 +316,7 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
         webBrowserFlags: webBrowserFlags,
         webCrossOriginIsolation: webCrossOriginIsolation,
         webChromeBinary: webChromeBinary,
+        webNoLaunchChrome: webNoLaunchChrome,
         webRenderer: webRenderer,
         webUseWasm: useWasm,
         enableImpeller: enableImpeller,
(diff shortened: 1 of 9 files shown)


bkonyi (dismissed):
LGTM!

However, it looks like packages/flutter_tools/test/web.shard/chrome_test.dart needs to be formatted.

mdebbar (commented):
The real bug here is that we are launching two Chrome instances (https://github.com/flutter/flutter/issues/148915), one is the correct one specified by `chrome-binary`, and the other one is unspecified (launched as a side effect of `-d chrome`). This PR is fixing the binary location of the latter, but it's not fixing the root of the issue.

It looks like we already have a mechanism to prevent `-d chrome` from launching a Chrome instance:
https://github.com/flutter/flutter/blob/d7d35fdd9f068b7346de77ee318039ed74518cfa/packages/flutter_tools/lib/src/web/web_device.dart#L138
and `flutter drive` is making an attempt to prevent the extra Chrome instance from launching:
https://github.com/flutter/flutter/blob/d7d35fdd9f068b7346de77ee318039ed74518cfa/packages/flutter_tools/lib/src/commands/drive.dart#L355
but something is failing in between.

<hr>

I recommend going after the root of the issue by fixing the `--no-launch-chrome`, what do you think @shindonghwi ?

shindonghwi:
@mdebbar yeah you're right, I traced through the code.

The platformArgs approach doesn't work because resident_web_runner.dart:363 overwrites it with `{'uri': url}`. So the `no-launch-chrome` key never reaches web_device.dart.

Fixed it by adding `webNoLaunchChrome` to DebuggingOptions and passing it through the whole chain. Also added tests for the flag behavior.

Updated commit pushed.

packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -282,7 +282,11 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
   /// Create a debugging options instance for the current `run` or `drive` invocation.
   @visibleForTesting
   @protected
-  Future<DebuggingOptions> createDebuggingOptions({WebDevServerConfig? webDevServerConfig}) async {
+  Future<DebuggingOptions> createDebuggingOptions({
+    WebDevServerConfig? webDevServerConfig,
+    String? webChromeBinary,
```
mdebbar, line 287: Is this still needed after fixing the "no launch" flag?

bkonyi:
Just checking in @shindonghwi. Do you have plans to come back to this PR? :)

shindonghwi:
@bkonyi Hey, thanks for checking in! Unfortunately, I've been quite busy lately and won't be able to get back to this PR anytime soon. Feel free to take it over or close it if needed. Sorry about that!

bkonyi:
Closing in favor of https://github.com/flutter/flutter/pull/185481

PR #180001, opened 2025-12-17
Add SnackBarTheme

### LINKED ISSUE #180000: Missing SnackBarTheme
### Use case

Customize `SnackBar` visual properties in a sub-tree through an `InheritedTheme` widget.

### Proposal

Add `SnackBarTheme` similarly to similar inherited themes.

Sample usage:

```dart
    return SnackBarTheme(
      data: const SnackBarThemeData(
        width: 300,
        behavior: SnackBarBehavior.floating,
        actionTextColor: Color(0xFF0000FF),
      ),
      child: Scaffold(
        body: Center(
          child: ElevatedButton(
            onPressed: () {
              final snackBar = SnackBar(
                content: const Text('Yay! A SnackBar!'),
                action: SnackBarAction(label: 'Undo', onPressed: () {}),
              );
              ScaffoldMessenger.of(context).showSnackBar(snackBar);
            },
            child: const Text('Show snackbar'),
          ),
        ),
      ),
    );
  }
  
```

bleroux:
## Description

This PR adds the `SnackBarTheme` subclass of InheritedTheme. Similarly to other theme classes.

Despite missing, this class was mentioned in [flutter.dev/go/material-theme-system-updates](http://flutter.dev/go/material-theme-system-updates):
"SnackBarTheme and SnackBarThemeData are conformant.. "

## Related Issue
Fixes [Missing SnackBarTheme](https://github.com/flutter/flutter/issues/180000)

## Tests

Adds 2 tests.

### DIFF
--- a/packages/flutter/test/material/snack_bar_theme_test.dart
+++ b/packages/flutter/test/material/snack_bar_theme_test.dart
@@ -286,6 +286,136 @@ void main() {
     expect(snackBarBottomRight.dx, (800 + snackBarWidth) / 2); // Device width is 800.
   });
 
+  testWidgets('Local SnackBarTheme properties are used', (WidgetTester tester) async {
+    const Color backgroundColor = Colors.purple;
+    const Color textColor = Colors.pink;
+    const elevation = 7.0;
+    const action = 'ACTION';
+    const ShapeBorder shape = RoundedRectangleBorder(
+      borderRadius: BorderRadius.all(Radius.circular(9.0)),
+    );
+    const snackBarWidth = 400.0;
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: SnackBarTheme(
+          data: const SnackBarThemeData(
+            backgroundColor: backgroundColor,
+            behavior: SnackBarBehavior.floating,
+            width: snackBarWidth,
+            elevation: elevation,
+            shape: shape,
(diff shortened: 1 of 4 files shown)


packages/flutter/test/material/snack_bar_theme_test.dart
```diff
@@ -286,6 +286,136 @@ void main() {
...
+    const elevation = 7.0;
+    const action = 'ACTION';
+    const ShapeBorder shape = RoundedRectangleBorder(
+      borderRadius: BorderRadius.all(Radius.circular(9.0)),
+    );
+    const snackBarWidth = 400.0;
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: SnackBarTheme(
```
QuncCccccc, line 301: It seems it only works when `SnackBarTheme` wraps `Scaffold`. What if it's inside Scaffold? I think it's a bit rare to use SnackBarTheme to wrap a Scaffold like this. If that's the only use case, then I don't think it's necessary to add this class. We can directly use MaterialApp.theme.

bleroux: It is common for an app to have multiple Scaffold widgets (one per route for instance), some apps even rely on nested Scaffold.

QuncCccccc (approved):
LGTM.

PR #179736, opened 2025-12-11
Add FloatingActionButtonTheme

bleroux:
## Description

This PR adds the `FloatingActionButtonTheme` subclass of InheritedTheme. Similarly to other theme classes.

This missing theme class was mentioned in [flutter.dev/go/material-theme-system-updates](http://flutter.dev/go/material-theme-system-updates):
"FloatingActionButtonThemeData is conformant but there’s no FloatingActionButtonTheme class. "

## Related Issue

Fixes [Missing FloatingActionButtonTheme](https://github.com/flutter/flutter/issues/179735) 

## Tests

Adds 2 tests.

PR #180008, opened 2025-12-17
Fix MaterialBannerTheme.data nullablity

bleroux:
## Description

This PR makes `MaterialBannerTheme.data` non nullable similarly to the data field of other themes.
(See https://github.com/flutter/flutter/pull/179736#discussion_r2624892231 for context).

## Tests

Adds 2 tests.
Those tests were missing tests for `MaterialBannerTheme`.


QuncCccccc (commented):
I don't think we need to remove the nullable type if it already exists because this may cause breakages.

Issue #107578, opened 2022-07-13
Missing default context menu buttons on each platform


justinmc:
By default, Flutter's text selection toolbar shows cut, copy, paste, and select all.  However, on native platforms, other buttons appear by default.

| Platform | Screenshot | Missing | Related Work |
| --- | --- | --- | --- |
| Android | <img width="220" src="https://user-images.githubusercontent.com/389558/178827258-e92e5d3c-a386-4a32-8312-ebaf8e96c6c0.png" /> <img width="220" src="https://user-images.githubusercontent.com/389558/178827259-299e765b-e58c-4233-9bd4-367400058abe.png" /> | Translate, Share, Web Search | Share: https://github.com/flutter/flutter/issues/138728<br>Custom buttons: https://github.com/flutter/flutter/issues/139361 |
| iOS | <img width="420" src="https://user-images.githubusercontent.com/389558/178826990-bf8fcf35-db27-4952-8c48-fc7feab18cef.jpg" /> | Select, Replace, [Translate](https://github.com/flutter/flutter/issues/150392), [unit conversion](https://github.com/flutter/flutter/issues/111672) | Look up: https://github.com/flutter/flutter/pull/130532, https://github.com/flutter/engine/pull/43308. Search web: https://github.com/flutter/engine/pull/43324, https://github.com/flutter/flutter/pull/131898, Share: https://github.com/flutter/engine/pull/44554 |
| Linux | <img width="220" src="https://user-images.githubusercontent.com/389558/178825861-089f1e6b-882b-41a3-881a-0d7dad5d9c06.png" /> | Undo, Redo, Delete, Insert Emoji, Change Case | |
| Mac | <img width="220" alt="Screen Shot 2022-08-08 at 4 00 28 PM" src="https://user-images.githubusercontent.com/389558/183528809-3cac992a-2127-4dff-be17-88e35b7e73f5.png"> | Look Up "asdf", Translate "asdf", Search With Google, Share | |
| Windows | <img width="220" src="https://user-images.githubusercontent.com/389558/178828609-43b571b8-6168-4f43-949c-ebdfa2ae4ad1.png" /> | Undo, Delete, Right to left Reading order, Show Unicode control characters, Insert Unicode control character, Open IME, Reconversion, Search with Bing | |

### Open questions
How about buttons inserted by the system, such as the Anki Card button in my Android screenshot added by the AnkiDroid app? (**edit:** -> https://github.com/flutter/flutter/issues/107603)

### Related
[...]

Issue #172799, opened 2025-07-25
Format and lint Swift files with swift-format


jmagman:
The engine now has Swift files as of https://github.com/flutter/flutter/issues/144791. There are also macOS example project Swift files. Format and lint them with `swift-format`.

See where this was already done in packages
https://github.com/flutter/flutter/issues/41129
https://github.com/flutter/packages/pull/5928

Note it is now shipped with Xcode, so we don't need the cipd packages: https://github.com/flutter/packages/pull/9460.

cbracken:
Against my better judgement, I'm assigning myself. 😂

PR #180021, opened 2025-12-17
Add Translate to iOS selection context menu

### LINKED ISSUE #150392: [iOS] Add Translate button to text fields' context menu
In WWDC24, Apple announce the new Translation API https://developer.apple.com/documentation/Translation
This would enable adding the Translate option to the context edit menu, which was previously blocked by not being able to access the Translate API
Screenshot 2024-06-17 at 3 03 47 PM

Part of https://github.com/flutter/flutter/issues/107578

WWDC video: https://www.youtube.com/watch?v=MuIFhbHNmqA

LouiseHsu:
Fixes https://github.com/flutter/flutter/issues/150392, part of https://github.com/flutter/flutter/issues/107578

This PR adds a "Translate" action to the ios selection context menu using this [swiftui translate api](https://developer.apple.com/documentation/SwiftUI/View/translationPresentation(isPresented:text:attachmentAnchor:arrowEdge:replacementAction:)) announce during WWDC 2024.
Includes the ipad implementation as well.

https://github.com/user-attachments/assets/8c508f93-2341-498a-b494-c83fafa878f4

https://github.com/user-attachments/assets/61025f9c-f937-4043-bee5-ee31761cf5c0

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,44 @@ - (void)showLookUpViewController:(NSString*)term {
                                    completion:nil];
 }
 
+- (void)showTranslateViewController:(NSString*)term {
+  UIViewController* engineViewController = [self.engine viewController];
```
hellohuanlin: uber nit: we typically call this `flutterViewController` in embedder

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,44 @@ - (void)showLookUpViewController:(NSString*)term {
...
+  TranslateController* translateController;
+
+  if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
+    // On iPad, the translate screen is presented in a popover view, and requires a rect to use as
```
hellohuanlin, line 462: please double check ios 26 behavior, we may also wanna pass the rect on iphone

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,44 @@ - (void)showLookUpViewController:(NSString*)term {
...
+    CGRect firstRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        caretRectForPosition:(FlutterTextPosition*)range.start];
+    CGRect transformedFirstRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        localRectFromFrameworkTransform:firstRect];
+    CGRect lastRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        caretRectForPosition:(FlutterTextPosition*)range.end];
+    CGRect transformedLastRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        localRectFromFrameworkTransform:lastRect];
+
+    // In case of RTL Language, get the minimum x coordinate too
```
hellohuanlin, line 478: good corner case!

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+// found in the LICENSE file.
+
+import Foundation
+import UIKit
+import SwiftUI
+import Translation
+
+@available(iOS 17.4, *)
+@objc(TranslateController)
+public class TranslateController: UIViewController {
```
hellohuanlin: nit: Translate**View**Controller

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+import UIKit
+import SwiftUI
+import Translation
+
+@available(iOS 17.4, *)
+@objc(TranslateController)
```
hellohuanlin: Nit: objc name should have a prefix

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @available(*, unavailable)
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  override public func viewDidLoad() {
+    super.viewDidLoad()
+    let swiftUIViewController = makeTranslateHostingController(termToTranslate: originalText)
```
hellohuanlin: `termToTranslate` vs `originalText` can you pick just one?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+        swiftUIViewController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+        swiftUIViewController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+        swiftUIViewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+    ])
+    swiftUIViewController.didMove(toParent: self)
+  }
+
+  @objc public func makeTranslateHostingController(termToTranslate: String) -> UIViewController {
+    var contentView = ContentView(termToTranslate: termToTranslate, ipadBounds: ipadBounds)
+    contentView.onDismiss = { [weak self] in
```
hellohuanlin, line 68: nit: can do `guard let strongSelf = self else { return }`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+  var onDismiss: (() -> Void)?
+  
+  var body: some View {
+    Color.clear
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+        .translationPresentation(
+          isPresented: $isTranslationPopoverShown,
+          text: termToTranslate,
+          attachmentAnchor: ipadBounds != nil ? .rect(.rect(ipadBounds!)) : .rect(.bounds))
```
hellohuanlin: nit: can use `??`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
hellohuanlin: Am I understanding right that the reason we have to use 2 new VCs (this and the hosting controller) is that hosting vc cannot be imported in objc? 
Maybe leave a comment if so

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,86 @@
...
+  let termToTranslate : String
+  let ipadBounds: CGRect?
+
+  var onDismiss: (() -> Void)?
+  
+  var body: some View {
+    Color.clear
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+        .translationPresentation(
+          isPresented: $isTranslationPopoverShown,
```
hellohuanlin: `isTranslationPopoverShown` is only toggled internally. I wonder you can simplify it by getting rid of this state (and the `onChange` call) and just pass a raw binding: 

```
  .translationPresentation(
        isPresented: Binding(
          get: { true },
          set: { isShown in
            if !isShown {
              onDismiss?()
            }
          }),
        text: termToTranslate,
        //...
```

hellohuanlin (commented):
This is exciting!

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: i moved this out into a computed property lol

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
LouiseHsu: it looks nice
Simulator Screenshot - new iphone 17 (26 2) - 2026-01-13 at 15 43 24

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
LouiseHsu: done

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,100 @@
...
+  let termToTranslate : String
+  let ipadBounds: CGRect?
+
+  var onDismiss: (() -> Void)?
+
+  private var anchorSource: Anchor<CGRect>.Source {
+          if let rect = ipadBounds {
+              return .rect(rect)
+          }
+          return .bounds
```
hellohuanlin: Can you add a comment on why use different anchor source for iphone/ipad? 

The format is off. @jmagman i remember you setup formatter in packages repo. Do we also have it setup for flutter repo?

hellohuanlin (dismissed):

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
jmagman: > The format is off. @jmagman i remember you setup formatter in packages repo. Do we also have it setup for flutter repo?

No, it's tracked here: https://github.com/flutter/flutter/issues/172799

hellohuanlin: @LouiseHsu can you manually format this code in xcode by using `CMD + A` then `Control + I`

jmagman:
```
dartdoc:stdout: Generating docs for package flutter...
dartdoc:stderr:   error: unresolved doc reference [WidgetsLocalizations.translationButtonLabel]
dartdoc:stderr:     from system_context_menu.IOSSystemContextMenuItemTranslate: (file:///b/s/w/ir/x/w/flutter/packages/flutter/lib/src/widgets/system_context_menu.dart:452:13)
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8692571030596401233/+/u/run_test.dart_for_docs_shard_and_subshard_None/stdout

engine/src/flutter/testing/run_tests.py
```diff
@@ -1238,7 +1238,7 @@ def main() -> int:
       '--variant',
       dest='variant',
       action='store',
-      default='host_debug_unopt_arm64',
+      default='host_debug_unopt',
```
hellohuanlin, line 1268: remember to revert this @LouiseHsu

LouiseHsu:
Ugh, the test failures are real. The semantic node order is wrong even when the translate option is not being triggered - this failure occurs for a test for just a normal disabled text field.  Note that root node in the expect has `id:0` but the actual has `id:6`, which means it must've been created, then deleted, then recreated?
I didn't touch semantics at all so I've no clue why it's happening. Digging into it now. 

```
02:13 +2500 ~14 -1: /Volumes/Work/s/w/ir/x/w/flutter/packages/flutter/test/cupertino/text_field_test.dart: when disabled does not listen to onFocus events or gain focus (variant: TargetPlatform.iOS)
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure was thrown running a test:
Expected: semantics node matching:
          TestSemantics(
            id: 0,
            rect: Rect.fromLTRB(0.0, 0.0, 2400.0, 1800.0),
            children: <TestSemantics>[
              TestSemantics(
                id: 1,
                textDirection: TextDirection.ltr,
                children: <TestSemantics>[
                  TestSemantics(
                    id: 2,
                    children: <TestSemantics>[
                      TestSemantics(
                        id: 3,
                        flags: <SemanticsFlag>[SemanticsFlag.scopesRoute],
                        children: <TestSemantics>[
                          TestSemantics(
                            id: 4,
                            flags: <SemanticsFlag>[SemanticsFlag.isTextField,
SemanticsFlag.isFocusable, SemanticsFlag.hasEnabledState, SemanticsFlag.isReadOnly],
                            inputType: SemanticsInputType.text,
                            children: <TestSemantics>[
                            ],
                          ),
                        ],
                      ),
                    ],
                  ),
                ],
              ),
            ],
          )
  Actual: SemanticsTester:<SemanticsTester for SemanticsNode#0(Rect.fromLTRB(0.0, 0.0, 2400.0,
1800.0))>
   Which: expected node id 1 but found id 6.
          Current SemanticsNode tree:
            SemanticsNode#9
             │ Rect.fromLTRB(0.0, 0.0, 2400.0, 1800.0)
             │
             └─SemanticsNode#1
               │ Rect.fromLTRB(0.0, 0.0, 800.0, 600.0) scaled by 3.0x
               │ textDirection: ltr
               │
               └─SemanticsNode#2
                 │ Rect.fromLTRB(0.0, 0.0, 800.0, 600.0)
                 │ sortKey: OrdinalSortKey#5a85b(order: 0.0)
                 │
                 └─SemanticsNode#3
                   │ Rect.fromLTRB(0.0, 0.0, 800.0, 600.0)
                   │ flags: scopesRoute
                   │
                   └─SemanticsNode#4
                       Rect.fromLTRB(0.0, 0.0, 800.0, 600.0)
                       flags: isTextField, hasEnabledState, isReadOnly, isFocusable
                       textDirection: ltr
                       inputType: text
          The semantics tree would have matched the following configuration:
            TestSemantics.root(
              children: <TestSemantics>[
                TestSemantics(
                  id: 6,
                  label: 'Test starting...',
                  textDirection: TextDirection.ltr,
                ),
              ],
            )

When the exception was thrown, this was the stack:
#4      main.<anonymous closure> (file:///Volumes/Work/s/w/ir/x/w/flutter/packages/flutter/test/cupertino/text_field_test.dart:10638:7)
<asynchronous suspension>
#5      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:192:15)
<asynchronous suspension>
#6      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1692:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

This was caught by the test expectation on the following line:
  file:///Volumes/Work/s/w/ir/x/w/flutter/packages/flutter/test/cupertino/text_field_test.dart line 10638
The test description was:
  when disabled does not listen to onFocus events or gain focus (variant: TargetPlatform.iOS)```

File: engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm [hunk 12]
```diff
@@ -451,6 +454,45 @@ - (void)showLookUpViewController:(NSString*)term {
...
+  UIViewController* flutterViewController = [self.engine viewController];
+  FLTTranslateViewController* translateViewController;
+
+  if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
+    // On iPad, the translate screen is presented in a popover view, and requires a rect to use as
+    // bounds
+    FlutterTextInputPlugin* _textInputPlugin = [self.engine textInputPlugin];
+    UITextRange* range = _textInputPlugin.textInputView.selectedTextRange;
+
+    // firstRectForRange cannot be used here as it's current implementation does
```
justinmc: "it's" => "its"

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -3080,6 +3083,39 @@ final class IOSSystemContextMenuItemDataShare extends IOSSystemContextMenuItemDa
   }
 }
 
+/// A [IOSSystemContextMenuItemData] for the system's built-in translate
```
justinmc: "A" => "An"

packages/flutter/lib/src/widgets/context_menu_button_item.dart
```diff
@@ -48,6 +48,9 @@ enum ContextMenuButtonType {
   ///  * [LiveTextInputStatusNotifier], where the status of Live Text can be listened to.
   liveTextInput,
 
+  /// A button that launches a translation popup for the current text selection
```
justinmc: Period at the end.

justinmc (commented):
Thanks for adding this button, I'm really happy that we can keep expanding functionality here. Just some ideas for tests and some grammar nits.

There are a few tests that should be added on the framework side for parity with how we did the other buttons. Some simple tests for the IOS* classes like:

https://github.com/flutter/flutter/blob/9c66d267b4d8f82e0e7c9cc6e06f4c3c3f7fe7a9/packages/flutter/test/widgets/system_context_menu_test.dart#L619-L626

and

https://github.com/flutter/flutter/blob/9c66d267b4d8f82e0e7c9cc6e06f4c3c3f7fe7a9/packages/flutter/test/widgets/system_context_menu_test.dart#L718-L725

Also test SystemContextMenuController.getDefaultItems, which now includes the new translate button. Something similar to [this test](https://github.com/flutter/flutter/blob/9c66d267b4d8f82e0e7c9cc6e06f4c3c3f7fe7a9/packages/flutter/test/services/system_context_menu_controller_test.dart#L459), but maybe you need to make a selection first in order to get it to show up. Maybe you could just add on to that test.

@chunhtai do you have any idea why the semantics id would be changing in this case?

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1267,6 +1267,9 @@ mixin TextSelectionDelegate {
   /// Whether share is enabled.
   bool get shareEnabled => true;
 
+  /// Whether translate is enabled
```
loic-sharma:
```suggestion
  /// Whether translate is enabled.
```

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2941,6 +2955,24 @@ class EditableTextState extends State<EditableText>
     }
   }
 
+  /// Launch a translation interface for the selected text.
+  ///
+  /// Currently this is only implemented for iOS.
+  ///
+  /// When 'obscureText' is true or the selection is empty,
+  /// this function will not do anything
```
loic-sharma:
```suggestion
  /// this function will not do anything.
```

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2941,6 +2955,24 @@ class EditableTextState extends State<EditableText>
...
+  Future<void> translateSelection(SelectionChangedCause cause) async {
+    assert(!widget.obscureText);
```
loic-sharma: Should this line be removed? The comment indicates this method intends to no-op on obscured text.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,45 @@ - (void)showLookUpViewController:(NSString*)term {
                                    completion:nil];
 }
 
+- (void)showTranslateViewController:(NSString*)term {
```
hellohuanlin, line 457: nit: `showTranslateViewControllerForTerm: (NSString*)term `

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    let hostingController = UIHostingController(rootView: contentView)
+    hostingController.view.backgroundColor = .clear
+    return hostingController
+  }
+}
+
+@available(iOS 17.4, *)
+struct ContentView: View {
```
hellohuanlin, line 82: nit: `private` or `fileprivate`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,121 @@
...
+  let termToTranslate: String
+  let ipadBounds: CGRect?
```
hellohuanlin, line 84: these can be private?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,121 @@
...
+  var body: some View {
+    Color.clear
+      .frame(maxWidth: .infinity, maxHeight: .infinity)
+      .translationPresentation(
+        isPresented: Binding(
+          get: { isPresented },
+          set: { newValue in
+            isPresented = newValue
+            if !newValue {
+              onDismiss?()
```
hellohuanlin, line 109: can this be just 
```
Binding(
  get: { true },
  set: { shown in 
    if !shown {
      onDismiss?()
    }
  }
)
```

and then remove the `isPresented` state (and the onAppear)?

hellohuanlin (dismissed):

packages/flutter/lib/src/widgets/localizations.dart
```diff
@@ -220,6 +220,9 @@ abstract class WidgetsLocalizations {
   /// Label for "search web" edit buttons and menu items.
   String get searchWebButtonLabel;
 
+  /// The term used for presenting a translation menu for a selection.
```
Renzo-Olivares, line 223: nit: consider changing to `Label for "translate" edit buttons and menu items.` similar to the other context menu button labels documentation.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2736,6 +2739,17 @@ class EditableTextState extends State<EditableText>
...
 
+  @override
+  bool get translateEnabled {
+    if (defaultTargetPlatform != TargetPlatform.iOS) {
+      return false;
+    }
+
+    return !widget.obscureText &&
+        !textEditingValue.selection.isCollapsed &&
+        textEditingValue.selection.textInside(textEditingValue.text).trim() != '';
```
Renzo-Olivares, line 2750: I tried on a physical iPhone and was able to get the translate button to show up on a native app (iMessage) if the only text selected was whitespace.

I would usually lean towards matching native even though I don't think translating whitespace adds value here. Definetely not a blocker.

packages/flutter/test/widgets/text_selection_toolbar_utils.dart
```diff
@@ -53,9 +76,13 @@ void expectCupertinoToolbarForPartialSelection() {
       expect(find.text('Cut'), findsOneWidget);
       expect(find.text('Copy'), findsOneWidget);
       expect(find.text('Paste'), findsOneWidget);
-      expect(find.text('Share...'), findsOneWidget);
       expect(find.text('Look Up'), findsOneWidget);
       expect(find.text('Search Web'), findsOneWidget);
+
+      // Adding Translate pushes both Share and Translate to the next page
+      // So just check that they exist on the tool bar.
```
Renzo-Olivares, line 83: nit: `tool bar` -> `toolbar`.

packages/flutter/test/widgets/text_selection_toolbar_utils.dart
```diff
@@ -92,9 +119,13 @@ void expectCupertinoToolbarForFullSelection() {
...
```
Renzo-Olivares, line 126: nit: `tool bar` -> `toolbar`.

packages/flutter/test/widgets/text_selection_toolbar_utils.dart
```diff
@@ -53,9 +76,13 @@ void expectCupertinoToolbarForPartialSelection() {
...
+      expect(cupertinoToolbarHasText('Share...'), isTrue);
```
Renzo-Olivares, line 84: Consider using `find.text('Translate', skipOffstage: false)` instead of using `cupertinoToolbarHasText`.

packages/flutter/test/cupertino/text_selection_test.dart
```diff
@@ -525,7 +527,7 @@ void main() {
         expect(findOverflowBackButton(), findsOneWidget);
         expect(findOverflowNextButton(), findsOneWidget);
 
-        // Tapping the next button again shows the last page and the Share button
+        // Tapping the next button again shows the Share button
```
Renzo-Olivares, line 530: nit: end comments with a `.`, here and in the comment below.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+    let hostingController = UIHostingController(rootView: contentView)
+    hostingController.view.backgroundColor = .clear
+    return hostingController
+  }
+}
+
+@available(iOS 17.4, *)
+fileprivate struct ContentView: View {
+  fileprivate let termToTranslate: String
+  fileprivate let ipadBounds: CGRect?
```
hellohuanlin, line 84: these 2 are not accessed outside of `ContentView` so should be `private` rather than `fileprivate`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+
+  fileprivate var onDismiss: (() -> Void)?
```
hellohuanlin, line 86: it's file for this to be default (i.e. internal), because the struct itself is already `fileprivate`

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,45 @@ - (void)showLookUpViewController:(NSString*)term {
...
+    // firstRectForRange cannot be used here as its current implementation does
+    // not always return the full rect of the range.
+    CGRect firstRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        caretRectForPosition:(FlutterTextPosition*)range.start];
+    CGRect transformedFirstRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        localRectFromFrameworkTransform:firstRect];
+    CGRect lastRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        caretRectForPosition:(FlutterTextPosition*)range.end];
+    CGRect transformedLastRect = [(FlutterTextInputView*)_textInputPlugin.textInputView
+        localRectFromFrameworkTransform:lastRect];
```
LongCatIsLooong, line 476: Is this trying to get the bounding boxes of the current selection? So the assumption here is "term" is the same as the selected text? In that case why does the framework have to send "term" over if we know it's going to be the selected text?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPlugin.mm
```diff
@@ -451,6 +454,45 @@ - (void)showLookUpViewController:(NSString*)term {
...
+
+    // In case of RTL Language, get the minimum x coordinate too
```
LongCatIsLooong, line 478: I don't think this works in bidi text (I'm not sure if there's a valid use case for selecting across bidi regions, but the UI may look out of place if the user actually does that).

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+    let swiftUIViewController = makeTranslateHostingController(
+      termToTranslate: termToTranslate
+    )
+
+    addChild(swiftUIViewController)
+    view.addSubview(swiftUIViewController.view)
+
+    swiftUIViewController.view.translatesAutoresizingMaskIntoConstraints = false
+    NSLayoutConstraint.activate([
+      swiftUIViewController.view.topAnchor.constraint(equalTo: view.topAnchor),
```
LongCatIsLooong, line 49: So `swiftUIViewController` can use a different coordinate system even from `FlutterView` (if `topAnchor` is not 0 for example), and `iPadBounds` is calculated in the textInputView's coordinates?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+    ])
+    swiftUIViewController.didMove(toParent: self)
+  }
+
+  @objc func makeTranslateHostingController(termToTranslate: String) -> UIViewController {
+    var contentView = ContentView(
+      termToTranslate: termToTranslate,
+      ipadBounds: ipadBounds
+    )
+    contentView.onDismiss = { [weak self] in
```
LongCatIsLooong, line 68: Would it be possible to make `onDismiss` a `let` variable? Looks like it's not changing.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+        equalTo: view.leadingAnchor
+      ),
+      swiftUIViewController.view.trailingAnchor.constraint(
+        equalTo: view.trailingAnchor
+      ),
+    ])
+    swiftUIViewController.didMove(toParent: self)
+  }
+
+  @objc func makeTranslateHostingController(termToTranslate: String) -> UIViewController {
```
LongCatIsLooong, line 63: Why's this method `@objc` and internal? It seems it's only called in `viewDidLoad`?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2984,6 +2998,22 @@ class EditableTextState extends State<EditableText>
     }
   }
 
+  /// Launch a translation interface for the selected text.
+  ///
+  /// Currently this is only implemented for iOS.
+  ///
+  /// When 'obscureText' is true or the selection is empty, this function will not do anything.
+  Future<void> translateSelection(SelectionChangedCause cause) async {
```
LongCatIsLooong, line 3006: Why does this take a `SelectionChangedCause`? Doesn't seem to be used anywhere.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2984,6 +2998,22 @@ class EditableTextState extends State<EditableText>
     }
   }
 
+  /// Launch a translation interface for the selected text.
```
LongCatIsLooong, line 3001: nit:
```suggestion
  /// Launches a translation interface for the selected text.
```

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2984,6 +2998,22 @@ class EditableTextState extends State<EditableText>
...
+  ///
+  /// Currently this is only implemented for iOS.
```
LongCatIsLooong, line 3003: And only for iOS 17.4+ I think?

packages/flutter/lib/src/widgets/editable_text.dart
LongCatIsLooong: Not very familiar with context menus, but it looks like this is going to be called to trigger translation in a Flutter rendered context menu while the new logic in `system_context_menu.dart` is going to be used when the app is using an iOS rendered context menu? If that's the case, how does the app developer determine whether translation is available (i.e. the app is running on iOS 17.4+) and hide the button in the `translationEnabled` implementation (to prevent a "translate" button that does nothing from showing up).

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformPluginTest.mm
```diff
@@ -5,6 +5,7 @@
 #import <OCMock/OCMock.h>
 #import <XCTest/XCTest.h>
 
+#import "flutter/shell/platform/darwin/common/InternalFlutterSwiftCommon/InternalFlutterSwiftCommon.h"
```
LongCatIsLooong, line 8: I thought this was the bridging header? Why is the import needed?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/TranslateController.swift
```diff
@@ -0,0 +1,108 @@
...
+    if let rect = ipadBounds {
+      return .rect(rect)
+    }
+    return .bounds
+  }
+
+  var body: some View {
+    Color.clear
+      .frame(maxWidth: .infinity, maxHeight: .infinity)
+      .translationPresentation(
```
LongCatIsLooong, line 101: Just curious, is this API only available in SwiftUI?

Piinks:

Piinks:

Piinks:

The portions of this PR that affect the material and cupertino libraries can be ported over to flutter/packages, although they may have to wait for the changes in the core SDK to reach stable first.

jmagman:
Closing in favor of https://github.com/flutter/flutter/pull/189878

PR #180029, opened 2025-12-17
feat: enable const Container constructor

muhammadkamel:
Refactor [Container](cci:2://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:245:0-488:1) to use `const` constructor

This PR refactors the [Container](cci:2://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:245:0-488:1) widget to allow `const` construction.

**Changes:**
*   Converted [Container](cci:2://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:245:0-488:1) constructor to `const`.
*   Promoted `width` and `height` to `final` fields.
*   Moved constraint calculation and parameter validation from the constructor to the [build](cci:1://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:384:2-457:3) method.
*   Updated [debugFillProperties](cci:1://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:99:2-109:3) to include `width` and `height`.

**Benefits:**
*   **Performance:** Enables `const` optimizations (canonicalization), reducing memory usage and GC pressure.
*   **Build Efficiency:** Allows the widget tree to skip rebuilding subtrees when `const Container` instances are reused.

**Migration Notes:**
*   Invalid parameters (e.g. negative margins) that were previously caught in the constructor for non-const values will now be caught during the [build](cci:1://file:///Users/mohammadkamel/flutter/packages/flutter/lib/src/widgets/container.dart:384:2-457:3) phase. `const` construction will still catch some errors at compile time.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/card.dart
```diff
@@ -47,16 +47,9 @@ enum _CardVariant { elevated, filled, outlined }
...
-/// single card type: the elevated card. In that mode the named constructors
-/// ([Card.filled], [Card.outlined]) behave the same as the default [Card].
-///
-/// For Material Design 3 (when [ThemeData.useMaterial3] is true), three visual
-/// variants are available: the default [Card] (elevated), [Card.filled], and
-/// [Card.outlined]. All variants share the same theme class, [CardThemeData],
-/// so theme properties (for example [CardThemeData.shape]) apply to every card
-/// variant within the theme's scope.
-///
+/// Material Design 3 introduced new types of cards. The default [Card] is the
```
Piinks: Are the changes in this file related? If not, they should probably be reverted.

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -252,37 +252,27 @@ class Container extends StatelessWidget {
...
-    BoxConstraints? constraints,
+    this.width,
+    this.height,
+    this.constraints,
     this.margin,
     this.transform,
     this.transformAlignment,
     this.child,
     this.clipBehavior = Clip.none,
-  }) : assert(margin == null || margin.isNonNegative),
```
Piinks, line 271: Why move these assertions to the build method?

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -252,37 +252,27 @@ class Container extends StatelessWidget {
...
          color == null || decoration == null,
-         'Cannot provide both a color and a decoration.\n'
-         'The color argument is just a shorthand for "decoration: BoxDecoration(color: color)".\n'
-         'To use both a color and other decoration properties, set the color in the BoxDecoration instead.',
-       ),
-       constraints = (width != null || height != null)
-           ? constraints?.tighten(width: width, height: height) ??
-                 BoxConstraints.tightFor(width: width, height: height)
-           : constraints;
+         'Cannot provide both a color and a decoration\n'
```
Piinks: Why change the error message here?

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -385,9 +381,17 @@ class Container extends StatelessWidget {
...
+    assert(padding == null || padding!.isNonNegative);
+    assert(decoration == null || decoration!.debugAssertIsValid());
+    assert(constraints == null || constraints!.debugAssertIsValid());
+    assert(decoration != null || clipBehavior == Clip.none);
     Widget? current = child;
 
-    if (child == null && (constraints == null || !constraints!.isTight)) {
+    if (child == null &&
+        (constraints == null || !constraints!.isTight) &&
+        width == null &&
```
Piinks: Can you resolve the constraints before here in the build method with width and height as they were previously done in the constructor? That way we don't have to carry forward constraints, width and height to validate.

packages/flutter/lib/src/widgets/container.dart
```diff
@@ -430,8 +434,14 @@ class Container extends StatelessWidget {
       );
     }
 
-    if (constraints != null) {
-      current = ConstrainedBox(constraints: constraints!, child: current);
+    // Effective constraints calculation (moved from constructor)
+    final BoxConstraints? effectiveConstraints = (width != null || height != null)
```
Piinks: Bingo! Can we resolve this earlier?

packages/flutter/test/widgets/container_test.dart
```diff
@@ -8,26 +8,22 @@
 library;
 
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
   group('Container control tests:', () {
-    final container = Container(
+    const container = Container(
```
Piinks, line 16: This might actually make this hard to land, making it a breaking change. Many many instances like this will fail analysis without const being there.

Piinks (commented):
Hey @muhammadkamel thanks for contributing! I noticed there isn't a linked issue, is there one you can provide that this change would resolve?
This could be a pretty disruptive change, can you share more about the motivation?

muhammadkamel:
> > I ran a small benchmark comparing the current Container implementation with a const constructor version, and the results show fewer unnecessary rebuilds when using a const constructor.
> 
> Typically in this case, we would land such a benchmark test to add it to the other performance benchmarks we have. Then once a baseline is established, we would expect to be able to tangibly measure the improvement with the change to verify it. :) Is there an existing benchmark that this change could impact? dev/benchmarks contains a lot of them, the dashboard is at https://flutter-flutter-perf.luci.app/e

Thanks for the guidance, @Piinks! I looked at the existing benchmarks and found multi_widget_construction_perf which uses Container, but the containers there use dynamic values so they can't be const.

To properly measure the benefit of const Container, I could create a new benchmark that specifically tests rebuild performance when Container instances can be canonicalized (same const Container reused across rebuilds vs creating new instances each time).

Would you prefer:

A new dedicated benchmark for const widget construction/rebuild (could be useful for other widgets too)
An extension to an existing benchmark
I can land the benchmark first to establish a baseline before the const Container change.

muhammadkamel:
Thanks for the suggestion, @Piinks! I've created a new benchmark `const_vs_non_const_container_perf` in `dev/benchmarks/macrobenchmarks` that specifically measures rebuild performance with const vs non-const Container widgets.

The benchmark creates a grid where half uses const Containers (canonicalized) and half uses non-const Containers (new instances each frame), allowing direct comparison of rebuild efficiency.

Would you prefer I land the benchmark in a separate PR first to establish a baseline, then follow up with the const Container change?

Piinks:
Thanks very much for that! Wow, appreciate the quick response. :)

I would like to give the other reviewers a chance to weigh in on this. I think the impact will still be pretty big for everyone that will need to add `const` throughout their applications, so we should be careful to consider the tradeoffs. :)

packages/flutter/lib/src/widgets/container.dart
justinmc: Most are because the right hand side is not const, so it's not allowed. However I think we can keep the one about `decoration` if I'm not mistaken?

justinmc (commented):
Thanks for the bold proposal! I have wondered why Container isn't const myself before.

I think this comes down the following tradeoff: 1. const is better for performance, versus 2. asserts in the initializers are better for avoiding leaks and bugs.

We have leak tracking disabled in some tests because of asserts in lifecycle methods: https://github.com/flutter/flutter/issues/157470. I typically prefer to put the asserts in the initializers for the architectural benefits, despite the benefits of const. Given the fact that there would be a breaking change migration to do for this PR, I definitely feel inclined to make the same recommendation and close this PR.

If we wanted to land this change, I think we'd have to show:

 1. The performance benefits of const here are significant and give real benefits to Flutter app developers, clearly outweighing the risk of leaks or other bugs.
 2. This PR passes tests and doesn't cause any leak test failures.
 3. The migration for the lint rule breakage is doable and worth it.

dkwingsmt:
Since the team is leaning against this design, I'm closing this PR for now. We can definitely still discuss it if you'd like to continue pursuing, especially with proper metrics (notably the performance gain by enabling `const` here), but I'd like us to move the discussion under an issue for better tracking. In general, the best practice is always to make sure that a change is agreed upon in an issue before turning it into a PR. Regardless of the result, I really appreciate your PR and discussion!

PR #180071, opened 2025-12-18
Fix an issue where the semantics announce event may be encoded as either an int32_t or an int64_t depending on its value

mattkae:
fixes https://github.com/flutter/flutter/issues/179563

## What's new?
- The viewId on the announce semantics event may be either an int32 or an int64, depending on what the codec perceives is best
- We should parse either one of them on the Win32 side of things

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/windows/accessibility_plugin.cc
+++ b/engine/src/flutter/shell/platform/windows/accessibility_plugin.cc
@@ -69,21 +69,14 @@ void HandleMessage(AccessibilityPlugin* plugin, const EncodableValue& message) {
     }
 
     // The viewId may be encoded as either a 32-bit or 64-bit integer.
-    FlutterViewId view_id;
-    const auto* view_id_64 = std::get_if<FlutterViewId>(&view_itr->second);
-    const auto* view_id_32 = std::get_if<int32_t>(&view_itr->second);
-
-    if (view_id_64) {
-      view_id = *view_id_64;
-    } else if (view_id_32) {
-      view_id = static_cast<FlutterViewId>(*view_id_32);
-    } else {
+    auto const view_id = view_itr->second.TryGetLongValue();
+    if (!view_id) {
       FML_LOG(ERROR)
           << "Announce message 'viewId' property must be a FlutterViewId.";
       return;
     }
 
-    plugin->Announce(view_id, *message);
+    plugin->Announce(*view_id, *message);
   } else {
     FML_LOG(WARNING) << "Accessibility message type '" << *type
                      << "' is not supported.";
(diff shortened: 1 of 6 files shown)


engine/src/flutter/shell/platform/windows/accessibility_plugin.cc
```diff
@@ -68,14 +68,22 @@ void HandleMessage(AccessibilityPlugin* plugin, const EncodableValue& message) {
...
+    // The viewId may be encoded as either a 32-bit or 64-bit integer.
+    FlutterViewId view_id;
+    const auto* view_id_64 = std::get_if<FlutterViewId>(&view_itr->second);
+    const auto* view_id_32 = std::get_if<int32_t>(&view_itr->second);
+
+    if (view_id_64) {
+      view_id = *view_id_64;
+    } else if (view_id_32) {
+      view_id = static_cast<FlutterViewId>(*view_id_32);
+    } else {
```
loic-sharma: I see `EncodableValue::LongValue` exists, but it throws:
https://github.com/flutter/flutter/blob/ca986c3900aa8fb21ea82fc293765a0888198844/engine/src/flutter/shell/platform/common/client_wrapper/include/flutter/encodable_value.h#L205-L217

What do you think of adding a `std::optional<int64_t> EncodableValue::TryGetLongValue()` method?

This would let us simplify this logic, something like:

```cpp
const auto view_id_val = view_itr->second.TryGetLongValue();
if (!view_id_val) {
  FML_LOG(ERROR)
    << "Announce message 'viewId' property must be a FlutterViewId.";
  return;
}

const auto view_id = static_cast<FlutterViewId>(*view_id_val);
plugin->Announce(view_id, *message);
```

mattkae: Agreed!

engine/src/flutter/shell/platform/windows/accessibility_plugin_unittests.cc
```diff
@@ -0,0 +1,152 @@
...
+  msg[EncodableValue{"type"}] = EncodableValue{"announce"};
+  msg[EncodableValue{"data"}] = EncodableValue{data};
+
+  auto encoded =
+      StandardMessageCodec::GetInstance().EncodeMessage(EncodableValue{msg});
+
+  messenger()->SimulateEngineMessage(
+      kAccessibilityChannelName, encoded->data(), encoded->size(),
+      [](const uint8_t* reply, size_t reply_size) {});
+}
```
loic-sharma, line 149: Thanks for going above and beyond with these tests! :)

loic-sharma (approved):
Looks great, excellent work!

PR #180098, opened 2025-12-18
Add a new flutter cli command, running-apps, using mDNS app discovery

jwren:
Includes:
- Display running apps in a formatted table with age calculation
- Implement mDNS discovery for running apps (multiple devices/interfaces)
- Deduplicate apps by WebSocket URI
- Centralize mDNS device advertisement in MDNSDeviceDiscovery
- Ensure ResidentRunner advertises correct app name and cleans up
- Add network utility functions and JSON support for running-apps
- Add comprehensive tests for discovery and list command

Lastly, this change uses mDNS to discover running Flutter apps, the multicast_dns package is used instead of mdns_dart as the discovery functionality is insufficient in the mdns_dart package, only discovering a maximum of one service.


packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,56 @@ bool isIPv6Address(String address) {
...
+          addresses.add(addr);
+        }
+      }
+    }
+  } on SocketException {
+    // Ignore, just return what we have (or defaults if empty).
+  }
+
+  if (addresses.isEmpty) {
+    return <InternetAddress>[InternetAddress.loopbackIPv4, InternetAddress.loopbackIPv6];
```
bkonyi: This doesn't seem consistent with the behavior outlined in the documentation that states that only non-loopback addresses are returned.

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,56 @@ bool isIPv6Address(String address) {
...
+
+/// Finds the primary non-loopback IPv4 address of the local machine.
+///
+/// This address is often used to tell other network devices how to reach
+/// the service running on this machine.
+Future<InternetAddress> getLocalIpAddress() async {
+  final List<InternetAddress> addresses = await getLocalInetAddresses();
+  return addresses.firstWhere(
+    (InternetAddress addr) => addr.type == InternetAddressType.IPv4,
+    orElse: () => InternetAddress.loopbackIPv4,
```
bkonyi: Same here.

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,56 @@ bool isIPv6Address(String address) {
...
+    // We bind to the specified hostname (e.g., '0.0.0.0') and let the OS
+    // dynamically assign a free port by using port 0.
+    socket = await ServerSocket.bind(hostname, 0);
+
+    final int port = socket.port;
+    return port;
+  } on SocketException {
+    // If the bind operation fails for any reason (e.g., no interfaces available),
+    // rethrow.
+    rethrow;
```
bkonyi: If we're just rethrowing, I don't think we need this block.

packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -872,9 +872,9 @@ class RunCommand extends RunCommandBase {
     final appStartedTimeRecorder = Completer<void>.sync();
 
     TerminalHandler? handler;
-    // This callback can't throw.
     unawaited(
-      appStartedTimeRecorder.future.then<void>((_) {
+      // This callback is executed once the application has successfully started (or re-started).
```
bkonyi: Shouldn't this only run once when the application started?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+@visibleForTesting
+String getProcessAge(String? epochString, SystemClock systemClock) {
+  if (epochString == null) {
+    return 'unknown age';
+  }
+  final int? epoch = int.tryParse(epochString);
+  if (epoch == null) {
+    return 'unknown age';
+  }
+  final Duration elapsed = systemClock.now().difference(DateTime.fromMillisecondsSinceEpoch(epoch));
```
bkonyi, line 180: Nit: this can be

```dart
final Duration elapsed = systemClock.ago(DateTime.fromMillisecondsSinceEpoch(epoch));
```

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+  if (elapsed.inDays > 0) {
+    return '${elapsed.inDays}d';
```
bkonyi, line 182: Would it make more sense to just use `elapsed.ago()` (from the `DurationAgo` extension in `lib/src/base/utils.dart`)?

packages/flutter_tools/test/general.shard/resident_runner_test.dart
```diff
@@ -98,6 +99,80 @@ void main() {
...
+  testUsingContext(
+    'ResidentRunner advertises correct app name via mDNS',
+    () => testbed.run(() async {
+      globals.fs.file('pubspec.yaml').writeAsStringSync('name: my_test_app\n');
+      fakeVmServiceHost = FakeVmServiceHost(
+        requests: <VmServiceExpectation>[listViews, listViews],
+        httpAddress: Uri.parse('http://localhost:12345'),
+      );
+      final futureConnectionInfo =
+          Completer<DebugConnectionInfo>.sync();
```
bkonyi: Do these completers need to be `sync`?

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,56 @@ bool isIPv6Address(String address) {
     return false;
   }
 }
+
+/// Finds all non-loopback IPv4 and IPv6 addresses of the local machine.
+Future<List<InternetAddress>> getLocalInetAddresses() async {
```
bkonyi: Nit: avoid abbreviating "internet". This should be `getLocalInternetAddresses()` instead (or even `getLocalAddresses()`).

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+  @override
+  Future<FlutterCommandResult> runCommand() async {
+    // Use mDNS to discover running Flutter apps, the multicast_dns package is
+    // used instead of mdns_dart as the discovery functionality is insufficient
+    // in the mdns_dart package, only discovering a maximum of one service.
+
+    globals.printStatus('Searching for running Flutter apps...');
+    final MDnsClient client = _mdnsClient ?? MDnsClient();
+
+    final List<Map<String, String>> apps = [];
```
bkonyi: Nit: prefer providing collection type information on the RHS.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+      // Listen for pointer records (PTR) to find services
+      await for (final PtrResourceRecord ptr
+          in client
+              .lookup<PtrResourceRecord>(ResourceRecordQuery.serverPointer('_flutter_devices._tcp'))
+              .timeout(
+                const Duration(seconds: 5),
+                onTimeout: (EventSink<PtrResourceRecord> sink) => sink.close(),
+              )) {
+        pendingLookups.add(
+          (() async {
```
bkonyi: Consider making a helper method for this instead of evaluating a large closure.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+          return 0;
+        }
+        if (epochA == null) {
+          return 1; // Put unknown age last
+        }
+        if (epochB == null) {
+          return -1; // Put unknown age last
+        }
+        return epochB.compareTo(epochA);
+      });
```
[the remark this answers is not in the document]
bkonyi: +1, these type parameters can just be `String`.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+      globals.printStatus('Found ${apps.length} running Flutter ${pluralize('app', apps.length)}:');
+      final table = <List<String>>[];
+      for (final Map<String, dynamic> app in apps) {
+        final String projectName = app['project_name'] as String? ?? 'Unknown';
+        final String mode = app['mode'] as String? ?? 'Unknown';
+        final String deviceName = app['device_name'] as String? ?? 'Unknown';
+        final String deviceId = app['device_id'] as String? ?? 'Unknown';
+        final String platform = app['target_platform'] as String? ?? 'Unknown';
+        final String vmServiceUri = app['ws_uri'] as String? ?? 'Unknown';
+        final String age = getProcessAge(app['epoch'] as String?, globals.systemClock);
```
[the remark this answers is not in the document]
bkonyi: +1

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+        table.add(<String>[
+          '$projectName ($mode)',
+          deviceString,
+          platform,
+          vmServiceUri,
+          age,
+        ]);
+      }
+
+      // Calculate column widths
```
bkonyi: We also duplicate this logic in `Emulator.descriptions` and `Device.descriptions`. We should probably create a helper utility to handle sizing and printing tables to the terminal.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+      final indices = List<int>.generate(table[0].length - 1, (int i) => i);
+      List<int> widths = indices.map<int>((int i) => 0).toList();
+      for (final row in table) {
+        widths = indices.map<int>((int i) => math.max(widths[i], row[i].length)).toList();
+      }
+
+      // Join columns into lines of text
+      for (final row in table) {
+        globals.printStatus(
+          '  ${indices.map<String>((int i) => row[i].padRight(widths[i])).followedBy(<String>[row.last]).join(' • ')}',
```
bkonyi: Nit: avoid making multiple function calls in the context of string interpolation.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+                // The 'text' property might join them or we might need to access raw data.
+                // Assuming 'text' contains the joined strings, we split by common separators if needed,
+                // but usually it's one key-value pair per string in the TXT record.
+                // For now, let's try to parse it as key=value pairs, assuming they might be separated by commas or newlines if joined.
+
+                // If txt.text is used, we need to know how it's joined.
+                // Let's try to split by common delimiters just in case, or handle it as a single pair.
+                // Given the serving side uses a list of strings, they might come as separate TXT records or joined.
+
+                final List<String> parts = txt.text.split('\n'); // Try newline first
```
bkonyi: This comment says "Try newline first", but I don't see logic trying anything else?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+        if (epochB == null) {
+          return -1; // Put unknown age last
+        }
+        return epochB.compareTo(epochA);
+      });
+
+      globals.printStatus('Found ${apps.length} running Flutter ${pluralize('app', apps.length)}:');
+      final table = <List<String>>[];
+      for (final Map<String, dynamic> app in apps) {
+        final String projectName = app['project_name'] as String? ?? 'Unknown';
```
bkonyi: We should define constants for all of these keys and use them in place of string literals.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,197 @@
...
+  @override
+  final String category = FlutterCommandCategory.tools;
+
+  @override
+  Future<FlutterCommandResult> runCommand() async {
+    // Use mDNS to discover running Flutter apps, the multicast_dns package is
+    // used instead of mdns_dart as the discovery functionality is insufficient
+    // in the mdns_dart package, only discovering a maximum of one service.
+
+    globals.printStatus('Searching for running Flutter apps...');
```
bkonyi: We're trying to avoid introducing additional uses of `globals` in new code, so any instances accessed via `globals` should be passed into the command's constructor.

bkonyi (commented):
First pass

packages/flutter_tools/lib/src/base/net.dart
jwren: Comment updated

packages/flutter_tools/lib/src/base/net.dart
jwren: Comment updated

packages/flutter_tools/lib/src/base/net.dart
jwren: Removed.

packages/flutter_tools/lib/src/commands/run.dart
jwren: Removed "(or re-started)"

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: `ago` couldn't be use: type issues/casting between `DateTime` & `Duration`

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: I considered it for this PR, but it should be a follow up, given the size of this PR already.  Created an issue, and commented in the code. -- https://github.com/flutter/flutter/issues/180949

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: fixed

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: Comments cleaned up.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/base/net.dart
jwren: renamed.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: fixed.  There should a lint for repo consistency.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: I didn't know about the utility.  The utility has a longer form.  Adding a comment: ```  // TODO(jwren): Consider using [DurationAgo] from `lib/src/base/utils.dart`.
  // We need to decide on the width and precision, possibly modifying the utility
  // to support a shorter form (e.g. "5m" versus "5 minutes ago").```

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,56 @@ bool isIPv6Address(String address) {
...
+}
+
+Future<int> findUnusedPort({String hostname = '0.0.0.0'}) async {
+  ServerSocket? socket;
+  try {
+    // We bind to the specified hostname (e.g., '0.0.0.0') and let the OS
+    // dynamically assign a free port by using port 0.
+    socket = await ServerSocket.bind(hostname, 0);
+
+    final int port = socket.port;
```
bkonyi: Nit: this can just be `return socket.port;`

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+
+/// Command to list running Flutter applications.
+class RunningAppsCommand extends FlutterCommand {
+  RunningAppsCommand({
+    this.hidden = false,
+    @visibleForTesting MDnsClient? mdnsClient,
+    @visibleForTesting SystemClock? systemClock,
+    required Logger logger,
+  }) : _mdnsClient = mdnsClient,
+       _systemClock = systemClock ?? globals.systemClock,
```
bkonyi: The `systemClock` parameter should just be required so we don't need to use `globals` here. In general, we shouldn't be importing `globals.dart` in any new files.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+  final String category = FlutterCommandCategory.tools;
+
+  @override
+  Future<FlutterCommandResult> runCommand() async {
+    // Use mDNS to discover running Flutter apps, the multicast_dns package is
+    // used instead of mdns_dart as the discovery functionality is insufficient
+    // in the mdns_dart package, only discovering a maximum of one service.
+
+    _logger.printStatus('Searching for running Flutter apps...');
+    final MDnsClient client = _mdnsClient ?? MDnsClient();
```
bkonyi: Consider just setting `_mdnsClient = mdnsClient ?? MDnsClient()` in the initialization list, then `_mdnsClient` can be non-nullable.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+    final apps = <Map<String, String>>[];
+    final seenUris = <String>{};
+    try {
+      await client.start();
+
+      final pendingLookups = <Future<void>>[];
+      // Listen for pointer records (PTR) to find services
+      await for (final PtrResourceRecord ptr
+          in client
+              .lookup<PtrResourceRecord>(ResourceRecordQuery.serverPointer('_flutter_devices._tcp'))
```
bkonyi: Nit: `'_flutter_devices._tcp'` should be defined as a constant.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+              await for (final TxtResourceRecord txt in client.lookup<TxtResourceRecord>(
+                ResourceRecordQuery.text(ptr.domainName),
+              )) {
+                final metadata = <String, String>{};
+                // The multicast_dns package joins the strings of a TXT record with newlines.
+                final List<String> parts = txt.text.split('\n');
+                for (final part in parts) {
+                  final int equalsIndex = part.indexOf('=');
+                  if (equalsIndex != -1) {
+                    metadata[part.substring(0, equalsIndex)] = part.substring(equalsIndex + 1);
```
bkonyi: It's unlikely that we'll be seeing different formats given that it's other tool instances that will be advertising, but should we consider trimming these key/value pairs to ensure there's no extra whitespace?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+                  }
+                }
+                if (metadata.isNotEmpty) {
+                  final String? uri = metadata[_kWsUri];
+                  if (uri != null) {
+                    if (seenUris.contains(uri)) {
```
bkonyi: Ubernit: consider inverting this condition and replacing `continue` with `seenUris.add(uri)`

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,205 @@
...
+            }
+          })(),
+        );
+      }
+      await Future.wait(pendingLookups);
+    } finally {
+      client.stop();
+    }
+
+    if (boolArg('machine')) {
```
bkonyi: Nit: this should be a constant somewhere.

packages/flutter_tools/lib/src/commands/running_apps.dart
bkonyi: Whoops, I missed that. Disregard!

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,143 @@
...
+
+      if (await botDetector.isRunningOnBot) {
+        logger.printTrace('Running on CI/Bot, not starting mDNS server.');
+        return;
+      }
+
+      // Silence mDNS logs unless verbose logging is enabled.
+      // mdns_dart uses the 'mdns_dart' logger.
+      log.hierarchicalLoggingEnabled = true;
+      if (logger.isVerbose) {
```
bkonyi: Nit: `log.Logger('mdns_dart').level = logger.isVerbose ? log.Level.ALL : log.Level.SEVERE;`

packages/flutter_tools/lib/src/resident_runner.dart
```diff
@@ -1373,6 +1375,24 @@ abstract class ResidentRunner extends ResidentHandlers {
...
+        logger: globals.logger,
+        platform: globals.platform,
+        flutterVersion: globals.flutterVersion,
+        systemClock: globals.systemClock,
+      );
+      _mdnsDiscoveries.add(mdnsDeviceDiscovery);
+      await mdnsDeviceDiscovery.advertise(
+        appName: FlutterProject.fromDirectory(
+          globals.fs.directory(projectRootPath),
+        ).manifest.appName,
```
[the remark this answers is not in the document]
bkonyi: This can just be `FlutterProject.current()`. It wouldn't hurt to hoist this out of the loop, although it shouldn't cause repeated I/O.

packages/flutter_tools/lib/src/resident_runner.dart
```diff
@@ -1421,12 +1442,22 @@ abstract class ResidentRunner extends ResidentHandlers {
...
+
   Future<int> waitForAppToFinish() async {
     final int exitCode = await _finished.future;
     await cleanupAtFinish();
     return exitCode;
   }
 
+  @mustCallSuper
+  Future<void> cleanupAtFinish() async {
+    for (final MDNSDeviceDiscovery discovery in _mdnsDiscoveries) {
```
bkonyi: This can just be `await Future.wait(_mdnsDiscoveries.map((e) => e.stop()));`

packages/flutter_tools/lib/src/run_hot.dart
```diff
@@ -167,6 +166,9 @@ class HotRunner extends ResidentRunner {
         );
         _sdkName = 'multiple';
         _emulator = false;
+      // Use the first device's platform as the base for building assets.
```
bkonyi: Is this comment supposed to be somewhere else?

packages/flutter_tools/lib/src/run_hot.dart
```diff
@@ -130,8 +130,9 @@ class HotRunner extends ResidentRunner {
 
   @visibleForTesting
   String? get targetPlatformName => _targetPlatformName;
+
   String? _targetPlatformName;
-  final _targetPlatforms = <TargetPlatform>{};
+  final Set<TargetPlatform> _targetPlatforms = <TargetPlatform>{};
```
bkonyi: The LHS type isn't needed.

packages/flutter_tools/lib/src/run_hot.dart
```diff
@@ -153,9 +154,7 @@ class HotRunner extends ResidentRunner {
     switch (flutterDevices.length) {
       case 1:
         final Device device = flutterDevices.first.device!;
-        final TargetPlatform targetPlatform = await device.targetPlatform;
-        _targetPlatformName = getNameForTargetPlatform(targetPlatform);
-        _targetPlatforms.add(targetPlatform);
```
bkonyi, line 158: This shouldn't be removed.

packages/flutter_tools/lib/src/resident_runner.dart
```diff
@@ -1421,12 +1442,22 @@ abstract class ResidentRunner extends ResidentHandlers {
...
+  final _mdnsDiscoveries = <MDNSDeviceDiscovery>[];
+
   Future<int> waitForAppToFinish() async {
     final int exitCode = await _finished.future;
     await cleanupAtFinish();
     return exitCode;
   }
 
+  @mustCallSuper
+  Future<void> cleanupAtFinish() async {
```
bkonyi, line 1350: Is this always called, even if we encounter an exception or SIGINT? We might want to use the `ShutdownHooks` for this cleanup.

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
```diff
@@ -0,0 +1,288 @@
...
+          ),
+        ],
+      },
+      <String, List<TxtResourceRecord>>{
+        'service.local': <TxtResourceRecord>[
+          const TxtResourceRecord(
+            'service.local',
+            100,
+            text:
+                'project_name=my_project\ndevice_name=macos\ndevice_id=macos\ntarget_platform=darwin-arm64\nmode=debug\nws_uri=ws://127.0.0.1:1234/ws',
```
bkonyi: Nit: consider writing these strings using multiline strings rather than newline characters.

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
```diff
@@ -0,0 +1,288 @@
...
+            ),
+          ],
+        },
+      );
+      final command = RunningAppsCommand(mdnsClient: fakeMDnsClient, logger: testLogger);
+      final CommandRunner<void> runner = createTestCommandRunner(command);
+      await runner.run(<String>['running-apps']);
+
+      expect(testLogger.statusText, contains('Found 1 running Flutter app:'));
+      expect(testLogger.statusText.split('app_one').length - 1, 1); // Should appear exactly once
```
bkonyi: Nit: `expect(testLogger.statusText.allMatches('app_one'), hasLength(1));`

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
```diff
@@ -0,0 +1,288 @@
...
+  final Map<String, List<SrvResourceRecord>> srvRecords;
+  final Map<String, List<TxtResourceRecord>> txtRecords;
+
+  @override
+  Future<void> start({
+    InternetAddress? listenAddress,
+    NetworkInterfacesFactory? interfacesFactory,
+    int mDnsPort = 5353,
+    InternetAddress? mDnsAddress,
+    Function? onError, // Added onError
```
bkonyi: Is this comment necessary?

jwren: removed.

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
jwren: fixed

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
jwren: Fixed here and elsewhere in this file.

packages/flutter_tools/lib/src/run_hot.dart
jwren: Done

packages/flutter_tools/lib/src/run_hot.dart
jwren: done

packages/flutter_tools/lib/src/resident_runner.dart
jwren: Done.

packages/flutter_tools/lib/src/run_hot.dart
jwren: removed.

packages/flutter_tools/lib/src/resident_runner.dart
jwren: outdated comment with other refactorings troubleshooting the bots.

jwren:
I fixed a few bot issues and addressed a bunch of comments, not all done, but getting closer.

packages/flutter_tools/lib/src/base/net.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done, test added as well

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,54 @@ bool isIPv6Address(String address) {
     return false;
   }
 }
+
+/// Finds all non-loopback IPv4 and IPv6 addresses of the local machine.
+///
+/// If no non-loopback addresses are found, returns loopback addresses.
+Future<List<InternetAddress>> getLocalInternetAddresses() async {
```
natebosch: https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

How about `get localInternetAddresses` and `get localIpAddress`

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,209 @@
...
+  final description = 'List running applications.';
+
+  @override
+  final String category = FlutterCommandCategory.tools;
+
+  @override
+  Future<FlutterCommandResult> runCommand() async {
+    // Use mDNS to discover running Flutter apps, the multicast_dns package is
+    // used instead of mdns_dart as the discovery functionality is insufficient
+    // in the mdns_dart package, only discovering a maximum of one service.
```
natebosch: [nit] this comment seems very temporally local - is this information that future readers of this code will need to be reminded of each time they read it?

How about recording this detail in the commit message?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,209 @@
...
+        _logger.printStatus('  $rowString');
+      }
+    }
+    return FlutterCommandResult.success();
+  }
+}
+
+/// Formats the elapsed time since the given epoch.
+@visibleForTesting
+String getProcessAge(String? epochString, SystemClock systemClock) {
```
natebosch: How about a noun-named method `String processAge(String? epochString, SystemClock systemClock)`

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,209 @@
...
+  // to support a shorter form (e.g. "5m" versus "5 minutes ago").
+  if (epochString == null) {
+    return 'unknown age';
+  }
+  final int? epoch = int.tryParse(epochString);
+  if (epoch == null) {
+    return 'unknown age';
+  }
+  final Duration elapsed = systemClock.now().difference(DateTime.fromMillisecondsSinceEpoch(epoch));
+  if (elapsed.inDays > 0) {
```
natebosch, line 181: [optional] Consider a switch expression.

```
  return switch (systemClock.now().difference(
    DateTime.fromMillisecondsSinceEpoch(epoch),
  )) {
    Duration(:final inDays) when inDays > 0 => '${inDays}d',
    Duration(:final inHours) when inHours > 0 => '${inHours}h',
    Duration(:final inMinutes) when inMinutes > 0 => '${inMinutes}m',
    Duration(:final inSeconds) => '${inSeconds}s',
  };
```

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,209 @@
...
+                  apps.add(metadata);
+                }
+              }
+            } on Exception {
+              // Ignore errors for individual lookups
+            }
+          })(),
+        );
+      }
+      await Future.wait(pendingLookups);
```
natebosch: Consider using `await pendingLookups.wait;`

https://api.dart.dev/dart-async/FutureIterable/wait.html

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,161 @@
...
+import 'base/net.dart';
+import 'base/platform.dart';
+import 'base/time.dart';
+import 'build_info.dart';
+import 'device.dart';
+import 'version.dart';
+
+/// Discovers devices via mDNS and advertises the current Flutter application to them.
+class MDNSDeviceDiscovery {
+  /// Creates a new [MDNSDeviceDiscovery] instance.
```
natebosch: [nit] I'm not sure the Flutter repo stance on fully redundant doc comments. In Dart repos and internally to google3 we prefer omitting docs over including docs which are redundant with the signature. Consider expanding this, and if it can't be expanded consider removing this depending on Flutter repo conventions.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,161 @@
...
+  final FlutterVersion flutterVersion;
+  final SystemClock systemClock;
+  final BotDetector botDetector;
+
+  final _servers = <MDNSServer>[];
+
+  /// Advertises the Flutter application via mDNS.
+  ///
+  /// The advertisement includes metadata about the application, device, and environment.
+  Future<void> advertise({required String appName, required Uri? vmServiceUri}) async {
```
natebosch, line 66: [nit] This method is over 80 lines and indents to at least 5 levels. Consider splitting this up into composed submethods to reduce nesting and method length.

natebosch (commented):
Are there security implications for this? Should we make responding to discovery an optional feature?

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      await stop(); // Stop any existing advertisements before starting new ones.
+
+      if (vmServiceUri == null) {
+        logger.printTrace('VM Service URI not available, not starting mDNS server.');
+        return;
+      }
+
+      if (await botDetector.isRunningOnBot ||
+          platform.environment['BOT'] == 'true' ||
```
bkonyi: Are these two cases not covered by `isRunningOnBot`?

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
```diff
@@ -0,0 +1,369 @@
...
+    expect('app_one'.allMatches(testLogger.statusText), hasLength(1));
+    expect(testLogger.statusText, contains('ws://127.0.0.1:1234/ws'));
+  });
+
+  testWithoutContext('getProcessAge', () {
+    const kNowMs = 10_000_000;
+    const kFiveSecondsMs = 5_000;
+    const kFiftyNineSecondsMs = 59_000;
+    const kOneMinuteMs = 60_000;
+    const kFiftyNineMinutesMs = 3_540_000; // 59 * 60 * 1000
```
bkonyi: Ubernit: maybe it's worth defining these constants based on each other rather than using the final integers?

Something like:

```dart
const kSecondMs = 1000;
const kMinuteMs = kSecondMs * 60;
const kHourMs = kMinuteMs * 60;
const kDayMs = kHourMs * 24;
// ...
```

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: At one point I wasn't convinced, given that the bots are in a good state now, I am removing them to see for sure.

packages/flutter_tools/test/commands.shard/hermetic/running_apps_test.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: removed this comment.  Thanks Nate.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: Holding off for now, I am going to revisit having this method (see TODO.)

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: Done, here and in the other location it was used.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/base/net.dart
jwren: done

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,54 @@ bool isIPv6Address(String address) {
...
+/// the service running on this machine.
+Future<InternetAddress> get localIpAddress async {
+  final List<InternetAddress> addresses = await localInternetAddresses;
+  return addresses.firstWhere(
+    (InternetAddress addr) => addr.type == InternetAddressType.IPv4,
+    orElse: () => InternetAddress.loopbackIPv4,
+  );
+}
+
+Future<int> findUnusedPort({String hostname = '0.0.0.0'}) async {
```
bkonyi: Is this used outside of this file? If not, I think we can remove the optional named parameter.

packages/flutter_tools/lib/src/commands/running_apps.dart
bkonyi: There's still a closure being evaluated here. Is that intended?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+        pendingLookups.add(
+          (() async {
+            try {
+              // For each PTR, look up the TXT records
+              await for (final TxtResourceRecord txt in _mdnsClient.lookup<TxtResourceRecord>(
+                ResourceRecordQuery.text(ptr.domainName),
+              )) {
+                final metadata = <String, String>{};
+                // The multicast_dns package joins the strings of a TXT record with newlines.
+                final List<String> parts = txt.text.split('\n');
```
bkonyi: Will these always be Unix style line endings? Would it be safer to use `LineSplitter.split(...)`?

packages/flutter_tools/lib/src/commands/running_apps.dart
bkonyi: Use `FlutterGlobalOptions.kMachineFlag` instead.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+          })(),
+        );
+      }
+      await pendingLookups.wait;
+    } finally {
+      _mdnsClient.stop();
+    }
+
+    if (boolArg('machine')) {
+      _logger.printStatus(json.encode(apps));
```
bkonyi, line 86: We should just return after this print and remove the `else` statement to reduce nesting.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+                onTimeout: (EventSink<PtrResourceRecord> sink) => sink.close(),
+              )) {
+        pendingLookups.add(
+          (() async {
+            try {
+              // For each PTR, look up the TXT records
+              await for (final TxtResourceRecord txt in _mdnsClient.lookup<TxtResourceRecord>(
+                ResourceRecordQuery.text(ptr.domainName),
+              )) {
+                final metadata = <String, String>{};
```
bkonyi: Consider creating a helper method `Map<String, String> txtToMetadata(TxtResourceRecord txt)` or an extension `Map<String, String> toMap()` on `TxtResourceRecord` to help make this logic a bit more compact.

packages/flutter_tools/lib/src/isolated/resident_web_runner.dart
```diff
@@ -157,6 +158,7 @@ class ResidentWebRunner extends ResidentRunner {
   // Used with the new compiler to generate a bootstrap file containing plugins
   // and platform initialization.
   Directory? _generatedEntrypointDirectory;
+  final _mdnsDiscoveries = <MDNSDeviceDiscovery>[];
```
bkonyi, line 161: What is this list for?

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,163 @@
...
+    required this.vmService,
+    required this.debuggingOptions,
+    required this.logger,
+    required this.platform,
+    required this.flutterVersion,
+    required this.systemClock,
+    required this.botDetector,
+  });
+
+  static const String _kWsUri = 'ws_uri';
```
bkonyi: We should be using the same constants as we use in the command class to ensure they don't diverge.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
bkonyi: Still needs to be addressed.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,163 @@
...
+        '$_kFlutterVersion=$frameworkVersion',
+        '$_kDartVersion=$dartVersion',
+      ];
+
+      // Advertise on all available interfaces (IPv4 and IPv6).
+      for (final ip in ips) {
+        try {
+          final MDNSService mdnsService = await MDNSService.create(
+            instance: 'Flutter Tools on $hostname',
+            service: '_flutter_devices._tcp',
```
bkonyi: We should be using a shared constant here.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  /// Stops the mDNS advertisement.
+  Future<void> stop() async {
+    // Create a copy of the list to avoid ConcurrentModificationError as the list
+    // is modified during iteration.
+    final serversToStop = List<MDNSServer>.of(_servers);
+    _servers.clear();
+    for (final server in serversToStop) {
```
bkonyi: Ubernit: maybe this instead?

```dart
  await serversToStop
      .map(
        (s) => s.stop().catchError(
          (e) => logger.printTrace('Error stopping mDNS server: $e'),
        ),
      )
      .wait;
```

That way we don't have to block waiting on each server to start individually.

packages/flutter_tools/lib/src/commands/running_apps.dart
bkonyi: Actually, maybe it's worth creating a class for this record representation that's able to handle serializing / deserializing and provides a type-safe interface?

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+                        .trim();
+                  }
+                }
+                if (metadata.isNotEmpty) {
+                  final String? uri = metadata[_kWsUri];
+                  if (uri != null) {
+                    if (seenUris.contains(uri)) {
+                      continue;
+                    }
+                    seenUris.add(uri);
```
jakemac53: `Set.add` returns a boolean you can check to see if it was new:

```suggestion
                    if (!seenUris.add(uri)) {
                      continue;
                    }
```

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+                ResourceRecordQuery.text(ptr.domainName),
+              )) {
+                final metadata = <String, String>{};
+                // The multicast_dns package joins the strings of a TXT record with newlines.
+                final List<String> parts = txt.text.split('\n');
+                for (final part in parts) {
+                  final int equalsIndex = part.indexOf('=');
+                  if (equalsIndex != -1) {
+                    // Trim to remove any whitespace that may be around the delimiters.
+                    metadata[part.substring(0, equalsIndex).trim()] = part
```
jakemac53: It seems like we only care about the `_kWsUri` metadata, so we don't need to add everything to this map right? We could just search for the first _kWsUri entry and then perform the logic below on that.

packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -0,0 +1,208 @@
...
+                }
+                if (metadata.isNotEmpty) {
+                  final String? uri = metadata[_kWsUri];
+                  if (uri != null) {
+                    if (seenUris.contains(uri)) {
+                      continue;
+                    }
+                    seenUris.add(uri);
+                  }
+                  apps.add(metadata);
```
jakemac53: Do we want to do this if the `uri` is null?

packages/flutter_tools/lib/src/base/net.dart
jwren: resolved

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: refactored into a helper

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: changed

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: Changed to `LineSplitter`

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: This was right, yes, an object is now shared for wrapping/unwrapping the data on each end.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: resolved by having a single object now, all the constants are defined in one place.

packages/flutter_tools/lib/src/isolated/resident_web_runner.dart
jwren: It is for cleanup which I didn't add, see `resident_web_runner.dart` now.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done.  Tests have been updated as well. PTAL.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: done.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: To make a cohert CLI UX experience for users.  Also, consider the idea that they may have many running apps on their machines.  Being able to know which ones come from one project versus another, or the time when it was launched, will allow users the ability to choose which URI they actually mean.

packages/flutter_tools/lib/src/commands/running_apps.dart
jwren: In my opinion if devs are running the new command, `flutter running-apps` and they get content back without a URI to use, it would be better to know that the rest of the mechanism is working --  reporting an app, just not with uri.  @bkonyi -- does this seem reasonable to you?

jwren (commented):
Made changes, rebased and pushed the branch again.  Thank you @bkonyi!

packages/flutter_tools/lib/src/base/error_handling_io.dart
```diff
@@ -21,6 +20,7 @@ import 'package:file/file.dart';
 import 'package:path/path.dart' as p; // flutter_ignore: package_path_import
 import 'package:process/process.dart';
 
+import '../convert.dart';
```
bkonyi, line 23: Is this import needed?

packages/flutter_tools/lib/src/runner/flutter_command.dart
```diff
@@ -1281,6 +1281,15 @@ abstract class FlutterCommand extends Command<void> {
     );
   }
 
+  void addEnableLocalDiscoveryFlag() {
+    argParser.addFlag(
+      'enable-local-discovery',
```
bkonyi: This should probably just be part of `run.dart` since we're not using it in any other command. Also, we should define a constant for `enable-local-discovery` and use it in the error message where it's referenced.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+  final BotDetector botDetector;
+
+  final _servers = <MDNSServer>[];
+
+  /// Advertises the Flutter application via mDNS.
+  ///
+  /// The advertisement includes metadata about the application, device, and environment.
+  Future<void> advertise({required String appName, required Uri? vmServiceUri}) async {
+    try {
+      await stop(); // Stop any existing advertisements before starting new ones.
```
bkonyi: Do we expect this to be called more than once for a given instance?

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+        return;
+      }
+
+      if (await botDetector.isRunningOnBot) {
+        logger.printTrace('Running on CI/Bot, not starting mDNS server.');
+        return;
+      }
+
+      // Silence mDNS logs unless verbose logging is enabled.
+      // mdns_dart uses the 'mdns_dart' logger.
```
bkonyi: Nit: is this logger within `package:mdns_dart`? If so, I'd reference `package:mdns_dart` instead of just `mdns_dart` here.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+        }
+      }
+    } on Exception catch (e) {
+      logger.printError('Error getting local IPs or starting mDNS: $e');
+    }
+  }
+
+  /// Stops the mDNS advertisement.
+  Future<void> stop() async {
+    // Create a copy of the list to avoid ConcurrentModificationError as the list
```
bkonyi: Is this true? I don't see `_servers` being modified in the `serversToStop.map(...)` code.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+          logger.printTrace('Error stopping mDNS server: $e');
+        }, test: (Object error) => error is Exception),
+      ),
+    );
+  }
+}
+
+/// A class representing the metadata discovered from a running Flutter application
+/// via mDNS.
+class MDnsObservation {
```
bkonyi: Nit: `MDNSObservation` for consistency.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+  static const String _kHostname = 'hostname';
+
+  final String? hostname;
+  final String? projectName;
+  final String? deviceName;
+  final String? deviceId;
+  final String? targetPlatform;
+  final String? mode;
+  final String? wsUri;
+  final String? epoch;
```
bkonyi: Nit: I think `epoch` and `pid` can still be `int` or `double` as there's implicit `toString()` calls when using string interpolation.

packages/flutter_tools/lib/src/base/net.dart
```diff
@@ -230,3 +230,41 @@ bool isIPv6Address(String address) {
     return false;
   }
 }
+
+/// Finds all non-loopback IPv4 and IPv6 addresses of the local machine.
+///
+/// If no non-loopback addresses are found, returns loopback addresses.
+Future<List<InternetAddress>> get localInternetAddresses async {
```
bkonyi: These functions don't actually seem to be used outside of tests. Are they needed?

packages/flutter_tools/lib/src/mdns_device_discovery.dart
bkonyi: Also, maybe it makes sense to move this further up in the call stack where it can only be invoked once? Maybe even behind a static flag in the constructor of this class? It feels strange that we're resetting this state each time this method is invoked, particularly since `hierarchicalLoggingEnabled` applies to all loggers and could technically be set elsewhere.

bkonyi (commented):
Almost there!

packages/flutter_tools/lib/src/commands/running_apps.dart
bkonyi: I don't think there's anything actionable on the user side if there's an app without a URI, so I don't think it would hurt to just throw the entry away. Is it even possible to advertise without a VM service URI?

bkonyi: It looks like `FlutterVmService` only has null HTTP / WS URI entries in tests, so they should never be null in practice and we should never encounter a null `uri`. We should update the type to be non-nullable to indicate that.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,258 @@
...
+  static const String _kTargetPlatform = 'target_platform';
+  static const String _kMode = 'mode';
+  static const String _kWsUri = 'ws_uri';
+  static const String _kEpoch = 'epoch';
+  static const String _kPid = 'pid';
+  static const String _kFlutterVersion = 'flutter_version';
+  static const String _kDartVersion = 'dart_version';
+  static const String _kHostname = 'hostname';
+
+  final String? hostname;
```
bkonyi: I think these types should all be non-nullable and we should reject any records that don't match the expected signature.

packages/flutter_tools/lib/src/base/error_handling_io.dart
jwren: It is needed in this class for the `Encoding` object reference.

packages/flutter_tools/lib/src/runner/flutter_command.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: The method `advertise` is called only once per `MDNSDeviceDiscovery` instance, immediately after the device connects. However, the `stop()` call at the beginning of advertise serves as a defensive measure.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: updated the comment.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

packages/flutter_tools/lib/src/base/net.dart
jwren: removed, tests removed as well.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

jwren (commented):
comments addressed

packages/flutter_tools/lib/src/runner/flutter_command.dart
```diff
@@ -1281,6 +1283,15 @@ abstract class FlutterCommand extends Command<void> {
     );
   }
 
+  void addEnableLocalDiscoveryFlag({required bool verboseHelp}) {
```
bkonyi: Since this is only used by the run command, this (and the constant) should be moved there.

packages/flutter_tools/lib/src/mdns_device_discovery.dart
```diff
@@ -0,0 +1,294 @@
...
+      if (equalsIndex != -1) {
+        // Trim to remove any whitespace that may be around the delimiters.
+        metadata[part.substring(0, equalsIndex).trim()] = part.substring(equalsIndex + 1).trim();
+      }
+    }
+    if (metadata.isEmpty) {
+      return null;
+    }
+
+    final String? hostname = metadata[_kHostname];
```
bkonyi: Consider using pattern matching here:

```dart
  final int? epoch = int.tryParse(metadata[_kEpoch] ?? '');
  final int? pid = int.tryParse(metadata[_kPid] ?? '');

  if (metadata case {
    _kHostName: final hostname,
    _kProjectName: final projectName,
    _kDeviceName: final deviceName,
    _kTargetPlatform: final targetPlatform,
    _kMode: final mode,
    _kWsUri: final wsUri,
    _kFlutterVersion: final flutterVersion,
    _kDartVersion: final dartVersion,
  } when epoch != null && pid != null) {
    return MDNSObservation(
      hostname: hostname,
      projectName: projectName,
      deviceName: deviceName,
      deviceId: deviceId,
      targetPlatform: targetPlatform,
      mode: mode,
      wsUri: wsUri,
      epoch: epoch,
      pid: pid,
      flutterVersion: flutterVersion,
      dartVersion: dartVersion,
    );
  }
  return null;
```

packages/flutter_tools/lib/src/mdns_device_discovery.dart
bkonyi: Since we never expect for `advertise` to be invoked more than once I think it'd be better to fail loudly. Being defensive here instead of raising an exception could result in unexpected behavior that could be hard to pin down.

bkonyi (commented):
Almost there! Just a couple more comments.

packages/flutter_tools/lib/src/runner/flutter_command.dart
jwren: done

packages/flutter_tools/lib/src/mdns_device_discovery.dart
jwren: done

jwren: done

jwren (commented):
rebased and updated with all comments, thanks Ben

bkonyi (approved):
LGTM once the analysis failures are resolved!

PR #159393, opened 2024-11-24
Ability to customize the curve used in the hero animation (#26150)

ercantomac:
This PR introduces a new feature that lets the developers customize the curve used in the hero animation.

Related issue: #26150

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

Issue #26150, opened 2019-01-06
Hero should allow specifying a curve and reverseCurve


marcoms:
Currently in `heroes.dart`:

```dart
  Animation<double> get animation {
    return CurvedAnimation(
      parent: (type == HeroFlightDirection.push) ? toRoute.animation : fromRoute.animation,
      curve: Curves.fastOutSlowIn,
    );
  }
```

`Curves.fastOutSlowIn` looks decent for an entry curve, however when the hero animation is played in reverse when going to the previous route, you get something like the following:

![ezgif-4-8105ac49419e](https://user-images.githubusercontent.com/4211302/50741037-f6623d00-11ef-11e9-898e-2eb1959c7a44.gif)

This could be solved by checking `HeroFlightDirection.push/pull` and flipping the curve if necessary, but IMO it would make more sense to allow the user to specify `curve` and `reverseCurve` like most of the other animation APIs we have. `curve` would naturally be flipped when going backwards if `reverseCurve` isn't specified.

When I was working on improving the animations of `CupertinoPageRoute`, the Hero animation of the title becoming the back button was the one thing I couldn't alter, since there isn't an API to specify a reverse curve. Example (watch the Colors title):

![ezgif-4-1260e8ec0452](https://user-images.githubusercontent.com/4211302/50741148-383fb300-11f1-11e9-8ca8-fdba5b056b98.gif)

sir-boformer:
I'm supporting this!

kazbeksultanov:
Me too!

dungnv2602:
This should have been supported at the very beginning. Anyone can do a PR on this, It just takes up about 5 minutes to be exposed as a param I think?

nt4f04uNd:
look like this was fixed by https://github.com/flutter/flutter/pull/39844

it is possible to customize the curve by specifying, which was already available at the time this issue was opened (see #20322)

ercantomac:
This issue is almost 6 years old... Any updates?

victorsanni:
Partial patch at https://github.com/flutter/flutter/pull/159393

PR #180100, opened 2025-12-18
Enhance Hero widget with customizable animation curves

aaazlkm:
Follow-up to #159393 (closed due to inactivity).
Implements the same feature with changes based on the previous review feedback.

Related issue: https://github.com/flutter/flutter/issues/26150

## Screenshot

Using elasticOut because the difference was hard to see with fastOutSlowIn.
<table>
  <tr>
    <th>curve=elasticOut, reverseCurve=null</th>
    <th>curve=elasticOut, reverseCurve=linear</th>
  </tr>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/c8c39fb0-e677-471e-ac2a-2a062277ae48"
             controls muted loop width="320"></video>
    </td>
    <td>
      <video src="https://github.com/user-attachments/assets/1f39d0a8-ed80-4404-b766-38844cca3909"
             controls muted loop width="320"></video>
    </td>
  </tr>
</table>

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/heroes_test.dart
```diff
@@ -620,47 +620,111 @@ Future<void> main() async {
...
+      Curves.fastOutSlowIn,
+    ];
 
-    // Expect the height of the secondKey Hero to vary from 100 to 150
-    // over duration and according to curve.
+    for (final curve in testCurvesList) {
+      final observer = TransitionDurationObserver();
+      await tester.pumpWidget(
+        MaterialApp(
+          key: UniqueKey(),
```
aaazlkm: Recreate MaterialApp with a unique key to avoid conflicts from previous curve tests.

packages/flutter/lib/src/widgets/heroes.dart
```diff
@@ -459,10 +471,23 @@ class _HeroFlightManifest {
   CurvedAnimation? _animation;
 
   Animation<double> get animation {
+    final Curve curve, reverseCurve;
+    final Animation<double> parent;
+    switch (type) {
+      case HeroFlightDirection.push:
+        parent = toRoute.animation!;
+        curve = toHero.widget.curve;
+        reverseCurve = (toHero.widget.reverseCurve ?? curve).flipped;
```
aaazlkm: > this doesn't look right
> 
> shouldn't this be
> 
> ```dart
> reverseCurve = toHero.widget.reverseCurve ?? toHero.widget.curve.flipped;
> ```
> 
> same for line 484
> https://github.com/flutter/flutter/pull/159393#discussion_r1887259053


I see the feedback in the previous PR, but I believe the current implementation is correct.
The `reverseCurve` needs to be flipped because at line [723](https://github.com/flutter/flutter/pull/180100/changes#diff-00cb27aabcd395d28ae6613c147dc619e0cfd6b3735afec0371a9fa2ec0b5b39R720), the animation is wrapped with `ReverseAnimation` for pop direction. 
When `ReverseAnimation` reverses the animation, the curve also needs to be flipped to match the reversed direction, regardless of whether `reverseCurve` is explicitly provided or defaults to `curve`.

The tests also appear to be passing correctly.

packages/flutter/test/widgets/heroes_test.dart
```diff
@@ -620,47 +620,111 @@ Future<void> main() async {
...
+                      curve: curve,
+                      child: SizedBox(height: 150.0, width: 150.0, key: secondKey),
+                    ),
+                  ),
+                ],
+              ),
+            ),
+          },
+        ),
+      );
```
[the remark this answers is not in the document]
aaazlkm: I'd prefer to keep the current implementation as-is to minimize changes.

packages/flutter/lib/src/widgets/heroes.dart
```diff
@@ -459,10 +471,23 @@ class _HeroFlightManifest {
...
+      case HeroFlightDirection.pop:
+        parent = fromRoute.animation!;
+        curve = fromHero.widget.curve;
+        reverseCurve = (fromHero.widget.reverseCurve ?? curve).flipped;
+    }
```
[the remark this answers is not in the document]
aaazlkm: discussion on https://github.com/flutter/flutter/pull/180100#discussion_r2632969761

packages/flutter/test/widgets/heroes_test.dart
```diff
@@ -3237,27 +3305,95 @@ Future<void> main() async {
     navigator.currentState!.pop();
     await tester.pump();
 
+    final Duration duration = observer.transitionDuration;
+    final Curve expectedCurve = reverseCurve ?? curve;
```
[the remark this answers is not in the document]
aaazlkm:
```dart
    final Animatable<Size?> tween = SizeTween(
      begin: const Size(200, 200),
      end: const Size(100, 100),
    ).chain(CurveTween(curve: Curves.fastOutSlowIn));
```

[The original implementation](https://github.com/flutter/flutter/pull/180100/changes/BASE..00c809b62deb89efcd9f4829ede2d1a6a2917e5c#diff-8cc285af7d78b6dbae141652c2b7d9fd7e1b9f91c22cbbfbb208d7c81a73d78aL3191-L3195) uses curves instead of flipped.

packages/flutter/lib/src/widgets/heroes.dart
```diff
@@ -261,6 +263,16 @@ class Hero extends StatefulWidget {
...
   final bool transitionOnUserGestures;
 
+  /// The curve to use in the forward direction.
+  ///
+  /// Defaults to Curves.fastOutSlowIn.
+  final Curve curve;
+
+  /// The curve to use in the reverse direction.
+  ///
+  /// If this property is null, the flipped version of [curve] is used.
```
victorsanni: Can we directly say something like `Hero.curve.flipped`?

packages/flutter/lib/src/widgets/heroes.dart
victorsanni: Can we remove the ReverseAnimation wrapping instead of having this complication? If not, this should be documented in the line above this.

packages/flutter/test/widgets/heroes_test.dart
```diff
@@ -620,47 +620,111 @@ Future<void> main() async {
     await tester.pump(const Duration(seconds: 1));
   });
 
-  testWidgets('Heroes animation is fastOutSlowIn', (WidgetTester tester) async {
```
victorsanni, line 623: I wouldn't change this test, maybe rename to "Default Hero animation is fastOutSlowIn" and add a separate test for this PR.

packages/flutter/lib/src/widgets/heroes.dart
aaazlkm: Fixed in https://github.com/flutter/flutter/pull/180100/changes/fdd9db6bb91cfe5a81a11795d69d08b35c333908

packages/flutter/test/widgets/heroes_test.dart
aaazlkm: Fixed in https://github.com/flutter/flutter/pull/180100/changes/6a568d5785c334ac017ce8cd008222caa3b4eb3a
I also reverted the pop transition test back to its default-value test.　https://github.com/flutter/flutter/pull/180100/commits/5a2c1f96357b3ba99a2ea12461c88ffd5fcf88a6

packages/flutter/lib/src/widgets/heroes.dart
aaazlkm: I’m sorry to reverse my earlier position, but I now think flipping reverseCurve inside Hero is not the right approach.
1. As you pointed out, this makes the behavior harder to reason about and increases implementation complexity.
2. In the Flutter SDK, it’s generally expected that the caller flips the reverse curve so that the easing profile remains the same in both the forward and reverse directions.
    - For example, [_CupertinoPageTransitionState](https://github.com/flutter/flutter/blob/82d232df03e35b3937ca3737ba0cf7f12777a2a8/packages/flutter/lib/src/cupertino/route.dart#L657-L660) flips the reverse curve as well.
    - Aligning Hero’s behavior with these existing APIs should make the overall behavior more predictable.

Based on that, I updated the implementation to stop flipping reverseCurve inside Hero.　
https://github.com/flutter/flutter/pull/180100/changes/9cf4b326917e8a2f8c5dc3a0377b7b4ea7e052e9

packages/flutter/lib/src/widgets/heroes.dart
```diff
@@ -261,6 +263,16 @@ class Hero extends StatefulWidget {
   /// Defaults to false.
   final bool transitionOnUserGestures;
 
+  /// The curve to use in the forward direction.
+  ///
+  /// Defaults to Curves.fastOutSlowIn.
```
victorsanni:
```suggestion
  /// Defaults to [Curves.fastOutSlowIn].
```

To link to the documentation. This might need a doc import of `package:flutter/animation.dart` to work.

packages/flutter/lib/src/widgets/heroes.dart
```diff
@@ -261,6 +263,16 @@ class Hero extends StatefulWidget {
...
+  final Curve curve;
+
+  /// The curve to use in the reverse direction.
+  ///
+  /// If this property is null, Hero.curve.flipped is used.
```
victorsanni:
```suggestion
  /// If this property is null, [Hero.curve.flipped] is used.
```

or similar.

packages/flutter/test/widgets/heroes_test.dart
```diff
@@ -3258,6 +3365,146 @@ Future<void> main() async {
     expect(heroSize, tween.transform(1.0));
   });
 
+  Future<void> verifyPoppedHeroCurve({
```
victorsanni: Our convention is to define helper methods at the very end/top of the file and outside the `main` method, not between `testWidgets` calls.

aaazlkm: Fixed in https://github.com/flutter/flutter/pull/180100/changes/bfbbe1fb4a808c9df66fff46805a37dd6e971184

packages/flutter/lib/src/widgets/heroes.dart
aaazlkm: Fixed in https://github.com/flutter/flutter/pull/180100/commits/6377019c979c8cd2dcc81bd52f80d1a5cfb743dc

packages/flutter/lib/src/widgets/heroes.dart
aaazlkm: I tried [Hero.curve.flipped], but it didn’t link, so I used [Hero.curve].flipped instead.
https://github.com/flutter/flutter/commit/6377019c979c8cd2dcc81bd52f80d1a5cfb743dc

aaazlkm:
@victorsanni Thanks! I’ve applied your latest comments and pushed an update. Could you take another look?

packages/flutter/test/widgets/heroes_test.dart
justinmc: Nit: Another thing you could do is to move the for loop outside of testWidgets altogether, so that it creates a new test for each curve. Then I think you wouldn't need the key, and the test output might be a little easier to read.

justinmc (approved):
LGTM with a nit, but I'll defer to @victorsanni.

victorsanni (approved):

chunhtai (approved):
LGTM

packages/flutter/test/widgets/heroes_test.dart
aaazlkm: Thank you for the review! I updated the code as you suggested. Both the code and the test output are now easier to read—this was really helpful!
https://github.com/flutter/flutter/pull/180100/changes/3701d0db6e9e83cdffe18bd6f48e0390a1045181

PR #180120, opened 2025-12-19
Expose the Scroll Controller of a scrollable TabBar

### LINKED ISSUE #123965: Expose the Scroll Controller of a scrollable TabBar
## Use case

Exposing the Scroll Controller of a TabBar with isScrollable: true, will let developers apply custom behavior based on the scroll.
In my case, it was necessary to apply a Gradient Mask (Container with LinearGradient in Stack) over the TabBar to indicate to the user that there are more elements out of screen.
Example:
image

There was already an issue about it, but it has been closed (https://github.com/flutter/flutter/issues/64880)

## Proposal

It can be made in many ways, two of them can be:

- As a parameter, we can provide our own ScrollController. If null, we can check the presence of a PrimaryScrollController. If null, fallback to the _TabBarScrollController already provided
- Through a key, access to the State of the TabBar, making _scrollController public

### LINKED ISSUE #124608: Wrapping a Scrollbar around a TabBar throws error `The Scrollbar's ScrollController has no ScrollPosition attached.`
Wrapping a Scrollbar around a TabBar **works** but throws a console error 

**The Scrollbar's ScrollController has no ScrollPosition attached.**
A Scrollbar cannot be painted without a ScrollPosition. 

The Scrollbar attempted to use the PrimaryScrollController. This ScrollController should be associated with the ScrollView that the Scrollbar is being applied to.A ScrollView with an Axis.vertical ScrollDirection on mobile platforms will automatically use the PrimaryScrollController if the user has not provided a ScrollController. To use the PrimaryScrollController explicitly, set ScrollView.primary to true for the Scrollable widget.

The fix described in this commend https://github.com/flutter/flutter/issues/97873#issuecomment-1086520352 does not work because a tabcontroller cannot be applied to a scrollcontroller

```
PreferredSize(
              preferredSize: const Size.fromHeight(55),
              child: RawScrollbar(
                //controller: PrimaryScrollController.of(context),
                //padding: EdgeInsets.all(5),
                mainAxisMargin: 5,
                trackVisibility: false,
                thumbColor: Theme.of(context).focusColor,
                thickness: 3,
                thumbVisibility: true,
                interactive: false,
                scrollbarOrientation: ScrollbarOrientation.top,
                child: TabBar(
                  padding: const EdgeInsets.only(top: 5),
                  controller: _tabController,
                  labelPadding: const EdgeInsets.symmetric(horizontal: 20),
                  //padding: const EdgeInsets.only(top: 5,bottom: 5, left: 5, right: 5),
                  labelColor: Theme.of(context).colorScheme.primary,
                  labelStyle: Theme.of(context).textTheme.titleMedium,
                  isScrollable: true,
                  unselectedLabelColor:
                      Theme.of(context).focusColor.withOpacity(0.7),
                  indicatorColor: Theme.of(context).colorScheme.primary,
                  tabs: __tabs,
                ),
              ),
            ),
```

Working with **Channel stable, 3.7.10** and the issue appears on all platforms

huycozy:
- Resolves https://github.com/flutter/flutter/issues/123965 and fix https://github.com/flutter/flutter/issues/124608
- See my investigation at https://github.com/flutter/flutter/issues/123965#issuecomment-3664553178
- In this PR:
    - Introduce a callback `onScrollControllerCreated` that exposes `ScrollController` for users to use.
    - Add 6 tests (5 for new callback `onScrollControllerCreated`, 1 for regression test for https://github.com/flutter/flutter/issues/124608)
    - Add 2 API examples (1 for new callback `onScrollControllerCreated`, 1 for Scrollbar integration) and tests for them

<details open>
<summary>Demo</summary>

| before | after |
| --------------- | --------------- |
 | <video src="https://github.com/user-attachments/assets/e53e29bb-c5f4-45a6-8f3c-35eb39b47db6"/>

</details>

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -938,6 +938,7 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
     this.onTap,
     this.onHover,
     this.onFocusChange,
+    this.onScrollControllerCreated,
     this.physics,
     this.splashFactory,
     this.splashBorderRadius,
@@ -993,6 +994,7 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
     this.onTap,
     this.onHover,
     this.onFocusChange,
+    this.onScrollControllerCreated,
     this.physics,
     this.splashFactory,
     this.splashBorderRadius,
@@ -1301,6 +1303,42 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
   /// {@end-tool}
   final TabValueChanged<bool>? onFocusChange;
 
+  /// Called when the custom [ScrollController] for this scrollable [TabBar] is created.
+  ///
+  /// This callback provides access to the internal [ScrollController]
+  /// so that desired scroll behaviors from outside can be applied.
+  /// The callback is only invoked when [isScrollable] is true.
+  ///
+  /// The callback is called during the build phase. Calling [State.setState] in the
+  /// callback is not recommended unless [Future.microtask] or
+  /// [SchedulerBinding.addPostFrameCallback] are used.
+  ///
+  /// The controller is being managed inside [TabBar] and will be disposed automatically.
+  /// If you add listeners to the exposed controller, you must remove them in
+  /// your widget's dispose method to prevent memory leaks.
+  ///
+  /// {@tool dartpad}
+  /// This sample showcases how to use [onScrollControllerCreated] to
+  /// programmatically control the scroll position of a scrollable [TabBar].
+  ///
+  /// ** See code in examples/api/lib/material/tabs/tab_bar.onScrollControllerCreated.dart **
+  /// {@end-tool}
+  ///
+  /// {@tool dartpad}
+  /// This example showcases how to show [Scrollbar] with [TabBar] with
+  /// the exposed controller from [TabBar.onScrollControllerCreated].
+  /// The example will demonstrate how to handle the exposed controller safely,
+  /// and use it to scroll the associated [TabBar] programmatically.
+  ///
+  /// ** See code in examples/api/lib/material/tabs/tab_bar.onScrollControllerCreated.Scrollbar.dart **
+  /// {@end-tool}
+  ///
+  /// See also:
+  ///
+  ///  * [ScrollController], which can be used to control a scrollable widget.
+  ///  * [isScrollable], which determines whether the tab bar is scrollable.
+  final void Function(ScrollController controller)? onScrollControllerCreated;
+
   /// How the [TabBar]'s scroll view should respond to user input.
   ///
   /// For example, determines how the scroll view continues to animate after the
@@ -1440,6 +1478,7 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
 
 class _TabBarState extends State<TabBar> {
   ScrollController? _scrollController;
+  bool _scrollControllerCallbackInvoked = false;
   TabController? _controller;
   _IndicatorPainter? _indicatorPainter;
   int? _currentIndex;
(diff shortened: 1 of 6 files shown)


packages/flutter/lib/src/material/tabs.dart
```diff
@@ -2016,6 +2054,13 @@ class _TabBarState extends State<TabBar> {
             ).add(widget.padding ?? EdgeInsets.zero)
           : widget.padding;
       _scrollController ??= _TabBarScrollController(this);
```
navaronbracke, line 2056: Why don't we make `_TabBarScrollController` public but `final`?

I understand that the private _TabBarState needs to stay private, but why don't we include a private setter for it?

```dart
final class TabBarScrollController extends ScrollController {
  /// The state of the attached [TabBar], if any.
  ///
  /// Is null when not attached to a [TabBar].
  _TabBarState? _tabBarState;

  void _attachToTabBar(_TabBarState tabBar) {
     _tabBarState = tabBar;
  }

  @override
  ScrollPosition createScrollPosition(
    ScrollPhysics physics,
    ScrollContext context,
    ScrollPosition? oldPosition,
  ) {
    return _TabBarScrollPosition(
      physics: physics,
      context: context,
      oldPosition: oldPosition,
      tabBar: _tabBarState!,
    );
  }
}
```

Then we only need to call the attach method using a private helper method? (which handles creating an internal controller and attaches the `this`)

For example:

```dart
TabBarScrollController _setUpTabBarController() {
   final tabBarController = widget.tabBarController ?? TabBarScrollController();
   
   tabBarController._attachToTabBar(this);
   
   return tabBarController;
}

// ...
_scrollController ??= _setUpTabBarController()
```

huycozy: Thanks for your review, @navaronbracke!
If we proceed with this approach:
1. how do we initialize a `TabBarState` to pass into the `TabBarScrollController` constructor? And we need to make TabBarState public too, for this purpose.
2. Even if this were feasible, I'm quite concerned about the integrity and encapsulation of TabBarState class since it's marked as a public class. 
3. Since marking TabBarState as public, the cost of maintenance for it (new public API) may increase, as users may encounter unexpected issues with this change in their usage?

Also, looking back the original use case from OP at https://github.com/flutter/flutter/issues/123965 reported:"Exposing the Scroll Controller of a TabBar with isScrollable: true, will let developers apply custom behavior based on the scroll.", it means they only need `offset` from scrollController which is sufficient enough?

With these reasons, I think the callback approach is sufficient and safer.

navaronbracke: 1. It would never be a constructor parameter in my proposal.

For attaching a TabBarState, I had proposed using a new function, that is private to `TabBarController` only, so it handles this automatically.

```
  void _attachToTabBar(_TabBarState tabBar) {
     _tabBarState = tabBar;
  }
```

which is called right after the line that currently does this:

`_scrollController ??= _TabBarScrollController(this);`

It would then become:

`_scrollController ??= _effectiveTabBarScrollController.._attach(this);`

2. TabBarState stays private. We should not change this.
3. The only usage change is making TabBarScrollController's constructor public. Attaching to the private TabBarState is always going to be private API. Users will be able to use the public ScrollController API though (since the tab controller inherits it), which fixes this bug.

huycozy: Thanks for your update!

> It would never be a constructor parameter in my proposal.

I thought it'd be a constructor param as I saw your snippet code above, which contains `widget.tabBarController`:
```
final tabBarController = widget.tabBarController ?? TabBarScrollController();
```

> Users will be able to use the public ScrollController API though (since the tab controller inherits it), which fixes this bug.

Could you please explain how users can access the exposed controller for their needs?

Also, would you mind submitting another pull request for your solution (which will include tests and sample codes too)? Then the team can consider which approach is better for this case. I think @Piinks is the best person to review this, just let's wait until the holidays are over.

navaronbracke: Oh, `widget.tabBarController` would be a new constructor parameter for the TabBar _widget_ (a nullable TabBarController), not a change to the TabBarController constructor.

I will make a new pull request with my proposal. I'll add you as reviewer as well.

navaronbracke: @huycozy See https://github.com/flutter/flutter/pull/180389

Piinks:
Since there are 2 proposals here, I recommend we centralize discussion here: https://github.com/flutter/flutter/issues/123965#issuecomment-3721109393 💙

PR #180164, opened 2025-12-20
docs(CupertinoSheetRoute): clarify fullscreenDialog is not applicable

### LINKED ISSUE #175198: Mistake on Flutter Documentation [CupertinoSheetRoute]
The documentation for CupertinoSheetRoute has been filled with properties that belong to CupertinoPageRoute instead.

https://api.flutter.dev/flutter/cupertino/CupertinoSheetRoute-class.html

Those properties, such as fullscreenDialog for example, does not belong there.

nathannewyen:
Override fullscreenDialog getter with documentation explaining that this inherited property from PageRoute is not applicable to CupertinoSheetRoute.

Fixes #175198

### DIFF
--- a/packages/flutter/lib/src/cupertino/sheet.dart
+++ b/packages/flutter/lib/src/cupertino/sheet.dart
@@ -641,6 +641,14 @@ class CupertinoSheetRoute<T> extends PageRoute<T> with _CupertinoSheetRouteTrans
 
   @override
   bool get opaque => false;
+
+  /// This property is not applicable to [CupertinoSheetRoute].
+  ///
+  /// Cupertino sheets always slide up from the bottom and are not fullscreen
+  /// dialogs. This property is inherited from [PageRoute] but has no effect
+  /// on the sheet's behavior or appearance.
+  @override
+  bool get fullscreenDialog => false;
 }
 
 // Internally used to see if another sheet is in the tree already.


packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -641,6 +641,14 @@ class CupertinoSheetRoute<T> extends PageRoute<T> with _CupertinoSheetRouteTrans
 
   @override
   bool get opaque => false;
+
+  /// This property is not applicable to [CupertinoSheetRoute].
+  ///
+  /// Cupertino sheets always slide up from the bottom and are not fullscreen
+  /// dialogs. This property is inherited from [PageRoute] but has no effect
+  /// on the sheet's behavior or appearance.
```
MitchellGoodwin, line 649: I think it's better to have the documentation point to the existing template we use for fullscreenDialog. If we want to point out that this currently should always be false for CupertinoSheet, then that should be moved to the end of the existing explanation.

I also think it's important to specify that it is always false instead of "is not applicable". This property does a function by staying false.

MitchellGoodwin (commented):
Thanks for putting this together. I left a comment on the documentation, but I think this change will make this behavior easier to see.

dkwingsmt:
Thanks for your contribution! Since I haven't seen a reply, I'm going to reluctantly close this pull request for now, but if you get time to address the comments, we'd be more than happy to take a look at a new patch!

PR #180246, opened 2025-12-23
Fix resonant explosion of scroll disconnect when scrolling a pv in a list

gmackall:
I don't know if there is an issue for this, but without the fix it looks like (in particular see the end of the video)

https://github.com/user-attachments/assets/0c25ffdf-1fc0-4eb2-a18e-f5b369b266d3

and with the fix it looks like

https://github.com/user-attachments/assets/ade37009-a40b-4d1e-be22-b5ca5adfe246

Basically, we were using the value passed in by the engine in `onDisplayPlatformView(2)` to try to calculate the offset for this touch. However, we aren't guaranteed a call to `onDisplayPlatformView` for each call to `FlutterMutatorView.onTouchEvent`, and in practice we don't normally get them lined up like that, so the value is stale. Also, the touches are frequently changing by sub pixel values, but the values passed to `onDisplayPlatformView` are received as ints (we could maybe fix this, if it matters elsewhere). So this is fragile, comparing whole pixel to sub pixel, and also uses stale values. We can instead just ask android where the view is, I believe (this is what the PR does, and it works AFAICT).

I don't know how to test this.


engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/mutatorsstack/FlutterMutatorView.java
```diff
@@ -197,27 +195,9 @@ public boolean onTouchEvent(MotionEvent event) {
     if (androidTouchProcessor == null) {
       return super.onTouchEvent(event);
     }
-
+    requestUnbufferedDispatch(event);
```
gmackall: This may have an impact on platform views that themselves are handling touches, as we will be sending them unbuffered inputs after we pipe them through the framework. We could alternatively make this
```java
if (event.getAction() == MotionEvent.ACTION_MOVE) {
  requestUnbufferedDispatch(event);
}
```
as currently I think the move case is the place where we get the benefit, and this would then not have an impact on taps.

gmackall:
Confirmed this works with TLHC/VD

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/mutatorsstack/FlutterMutatorView.java
gmackall: I removed this, as it is a logically separate change. Can consider in a different PR.

mboetger (approved):

PR #180325, opened 2025-12-27
Fix misleading error message for color/decoration conflict

### LINKED ISSUE #119484: Providing Container a color and decoration produces an unclear error message
Similar to #36161, but that issue was closed with the incorrect error message in place.

## Steps to Reproduce

1) Create a `Container` with any valid `color` property.
2) Add the `decoration` property to the container.
3) Give `decoration` a `BoxDecoration` with a valid `color` property.
4) An exception is caught and an error message is printed.

**Expected results:** 
An error message that clearly defines the issue and optionally provides a way to fix the error.

**Actual results:** 
```
════════ Exception caught by widgets library ═══════════════════════════════════
The following assertion was thrown building Themes(dirty, dependencies: [_InheritedTheme, _LocalizationsScope-[GlobalKey#50d61]]):
Cannot provide both a color and a decoration
To provide both, use "decoration: BoxDecoration(color: color)".
```

The error message clearly defines the issue, however the suggested fix reads: `To provide both, use "decoration: BoxDecoration(color: color)".`
This is an incorrect suggestion because `Container` has checks to verify it was NOT provided both. The error message should clearly communicate the proper condition--which is passing Container.color OR Container.decoration, not both like the error message is suggesting.

Example:
```

void main() {
  runApp(new MaterialApp(
    home: new HomePage(),
  ));
}

class HomePage extends StatefulWidget {
  @override
  _HomePageState createState() => new _HomePageState();
}

class _HomePageState extends State<HomePage> {
  @override
    return Scaffold(
      body: Container(
        width: 50.0,
        height: 50.0,
        color: Colors.blue,
        decoration: BoxDecoration(
          color: Colors.blue,
      ),
    );
  }
}
```

nathannewyen:
## Description

Fixes the contradictory error message that says "Cannot provide both... To provide both".

Updated wording to match `Ink` and `AnimatedContainer` widgets which use clearer phrasing.

## Related Issue

Fixes #119484

### DIFF
--- a/packages/flutter/lib/src/cupertino/search_field.dart
+++ b/packages/flutter/lib/src/cupertino/search_field.dart
@@ -149,14 +149,14 @@ class CupertinoSearchTextField extends StatefulWidget {
   }) : assert(
          !((decoration != null) && (backgroundColor != null)),
          'Cannot provide both a background color and a decoration\n'
-         'To provide both, use "decoration: BoxDecoration(color: '
-         'backgroundColor)"',
+         'The backgroundColor argument is just a shorthand for '
+         '"decoration: BoxDecoration(color: backgroundColor)".',
        ),
        assert(
          !((decoration != null) && (borderRadius != null)),
          'Cannot provide both a border radius and a decoration\n'
-         'To provide both, use "decoration: BoxDecoration(borderRadius: '
-         'borderRadius)"',
+         'The borderRadius argument is just a shorthand for '
+         '"decoration: BoxDecoration(borderRadius: borderRadius)".',
        );
 
   /// Controls the text being edited.
--- a/packages/flutter/lib/src/cupertino/search_field.dart
+++ b/packages/flutter/lib/src/cupertino/search_field.dart
@@ -150,13 +150,15 @@ class CupertinoSearchTextField extends StatefulWidget {
          !((decoration != null) && (backgroundColor != null)),
          'Cannot provide both a background color and a decoration\n'
          'The backgroundColor argument is just a shorthand for '
-         '"decoration: BoxDecoration(color: backgroundColor)".',
+         '"decoration: BoxDecoration(color: backgroundColor)".\n'
+         'To use both a backgroundColor and other decoration properties, set the color in the BoxDecoration instead.',',
        ),
        assert(
          !((decoration != null) && (borderRadius != null)),
          'Cannot provide both a border radius and a decoration\n'
          'The borderRadius argument is just a shorthand for '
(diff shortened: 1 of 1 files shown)


packages/flutter/lib/src/cupertino/search_field.dart
```diff
@@ -149,14 +149,14 @@ class CupertinoSearchTextField extends StatefulWidget {
   }) : assert(
          !((decoration != null) && (backgroundColor != null)),
          'Cannot provide both a background color and a decoration\n'
-         'To provide both, use "decoration: BoxDecoration(color: '
-         'backgroundColor)"',
+         'The backgroundColor argument is just a shorthand for '
+         '"decoration: BoxDecoration(color: backgroundColor)".',
```
MitchellGoodwin:
```suggestion
         'The backgroundColor argument is just a shorthand for '
         '"decoration: BoxDecoration(color: backgroundColor)".\n'
         'To use both a backgroundColor and other decoration properties, set the color in the BoxDecoration instead.',',
```

packages/flutter/lib/src/cupertino/search_field.dart
```diff
@@ -149,14 +149,14 @@ class CupertinoSearchTextField extends StatefulWidget {
...
        ),
        assert(
          !((decoration != null) && (borderRadius != null)),
          'Cannot provide both a border radius and a decoration\n'
-         'To provide both, use "decoration: BoxDecoration(borderRadius: '
-         'borderRadius)"',
+         'The borderRadius argument is just a shorthand for '
+         '"decoration: BoxDecoration(borderRadius: borderRadius)".',
```
MitchellGoodwin:
```suggestion
         'The borderRadius argument is just a shorthand for '
         '"decoration: BoxDecoration(borderRadius: borderRadius)".\n'
         'To use both a radius and other decoration properties, set the radius in the BoxDecoration instead.',
```

MitchellGoodwin (commented):
Thank you for addressing the merge conflicts. I think I really like the pattern set in #119484. Can we go with that here as well?

justinmc (commented):
@nathannewyen Could you address @MitchellGoodwin's comments? I agree with him about adding that sentence. Otherwise this looks good, thanks for improving these.

dkwingsmt:
@nathannewyen I've applied @/MitchellGoodwin 's suggestions. However due to some administrative reasons, can you open a completely new PR? And make sure to not use the same commit or rebase, but by copying the code again. My apology that such a small change went so many back and forths!

dkwingsmt:
Closing this PR in favor of https://github.com/flutter/flutter/pull/181882.

Issue #97077, opened 2022-01-22
Navigating to a new page causes the Image errorBuilder to be ignored.


MartenM:
Hey there! Ran into this issue and been stuck at it for a while now. Created a little sample project and the issue remains.
Not sure what's going on here. Any help is appreciated but I fear it's a bug with flutter itself.

Due to the nature of the images, a 404 can be returned. This is no problem, attach an errorBuilder! The problem arises when I try to push a new route. For some reason, the error now is no longer being handled properly.

Let me know if you need anything else, or if I should post this issue somewhere else!

## Steps to Reproduce

Create a new project. Add the following code:
```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      initialRoute: '/',
      routes: {
        '/': (context) => const HomePage(),
        '/test': (context) => const AnotherPage(),
      },
    );
  }
}

class HomePage extends StatelessWidget {
  const HomePage({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return Scaffold(
        appBar: AppBar(
          title: const Text('Some test app'),
        ),
        body: Column(
          children: [
            _somePicture('https://docs.flutter.dev/assets/images/shared/brand/flutter/logo/flutter-lockup.png'),
            _somePicture('https://docs.flutter.dev/assets/invalid'),
            SizedBox(height: 20),
            ElevatedButton(onPressed: () => {
              Navigator.pushNamed(context, '/test')
            }, child: Text("Another page"))
          ],
        )
    );
  }

  Widget _somePicture(String input) {
    return Image.network(input, errorBuilder: (a, b, c) {
      return const Text("Error loading image!"); 
    });
  }
}

class AnotherPage extends StatelessWidget {
  const AnotherPage({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return Scaffold(
        appBar: AppBar(
          title: const Text('Some test app'),
        ),
        body: const Center(child: Text("Second page"))
    );
  }
}
```

1. Execute `flutter run` on the code sample
2. Click on the button to go to the next page.
3. Watch the error appear in the console.

**Expected results:**
[...]

PR #180327, opened 2025-12-27
Add reportErrors to ImageStreamListener

### LINKED ISSUE #81931: Image.network uncaught exception
**Summery:**
I don't know why I can't catch `Image.network` exceptions. If I give `Image.network` an invalid URL, the errorBuilder kicks in, that's good but if I give it a valid URL with 404 response or non-image data, errorBuilder() doesn't seem to do anything. The app pauses because of the uncaught exception.

**Details:**
I made a new app with `flutter create bug`, this is my main.dart:

For first test, I give it an invalid URL (AAABBBCCC)
This one works as expected. errorBuilder() is called and 'N/A' is printed on the screen.
```

void main() {
  runApp(MyApp());
}

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

  MyHomePage({Key? key, required this.title}) : super(key: key);

  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          children: <Widget>[
            Text(
              'Network Image',
            ),
            MyImage(),
          ],
        ),
    );
  }
}

class MyImage extends StatelessWidget {
  @override
    try {
      return Image.network(
        'AAABBBCCC', //url,
        errorBuilder: (BuildContext context, Object exception, StackTrace? stackTrace) {
          return Text('N/A');
        },
      );
    } catch (e) {
      print(e.toString());
    }

    return Container();
  }
}

```

photo_2021-05-06_00-03-04

Now a valid URL but with 404 response: (https://example.com/nothing)
```
class MyImage extends StatelessWidget {
  @override
    try {
      return Image.network(
        'https://example.com/nothing', //url,
        errorBuilder: (BuildContext context, Object exception, StackTrace? stackTrace) {
          return Text('N/A');
        },
      );
    } catch (e) {
      print(e.toString());
    }

    return Container();
  }
}
```
bug01

For the last try, a valid URL with data but invalid image data: (https://google.com/)
```
class MyImage extends StatelessWidget {
  @override
    try {
      return Image.network(
        'https://google.com/', //url,
        errorBuilder: (BuildContext context, Object exception, StackTrace? stackTrace) {
          return Text('N/A');
        },
      );
    } catch (e) {
      print(e.toString());
    }

    return Container();
  }
}
```
bug02

As you can see the app is pausing due to uncaught exceptions.

I've also tried `FadeInImage()` and `CachedNetworkImage()`, they both act the same as `Image.network()`.

```
[√] Flutter (Channel beta, 2.2.0-10.2.pre, on Microsoft Windows [Version 10.0.19041.928], locale en-US)
[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[√] Android Studio (version 4.1.0)
[√] Android Studio
[√] Connected device (3 available)

• No issues found!
```

koji-1009:
fix https://github.com/flutter/flutter/issues/81931

Run the following app and quickly tap the FAB multiple times to output an exception to the console. However, `N/A` continues to be displayed on the UI, and `errorBuilder` is functioning correctly.

```dart

void main() {
  runApp(MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(home: MyHomePage());
  }
}

  @override
}

  var counter = 0;

  @override
    return Scaffold(
      appBar: AppBar(title: Text('Flutter Demo Home Page')),
      body: Center(
        child: Column(
          spacing: 16,
          children: [
            MyImage(
              url: switch (counter % 3) {
                0 => 'AAABBBCCC',
                1 => 'https://example.com/nothing',
                2 => 'https://google.com',
                _ => 'AAABBBCCC',
              },
            ),
          ],
        ),
      ),
        onPressed: () {
          setState(() {
            counter++;
          });
        },
        child: Text(counter.toString()),
      ),
    );
  }
}

class MyImage extends StatelessWidget {
  const MyImage({super.key, required this.url});

  final String url;

  @override
    return Image.network(
      url,
      errorBuilder: (context, exception, stackTrace) {
        return Text('N/A');
      },
    );
  }
}
```

### Root Cause

`ImageStreamCompleter` outlives the `Image` widget because `ImageCache` holds keepAlive handles that prevent disposal even after the widget is gone. When the widget is disposed or rebuilt with a different URL, its listener is removed, but the completer continues to live. Errors arriving after that point have no `onError` handler and fall through to `FlutterError.reportError`.

The completer has no memory of whether an error handler was ever present, so it cannot distinguish between "no one ever handled errors" and "someone handled errors but is now gone." This is the root cause.

A previous fix (`addEphemeralErrorListener` in `_stopListeningToStream`) covers the dispose path, but not the rebuild path where `_updateSourceStream` removes the listener directly. It also has a limitation when a cached image is delivered first (`_currentImage != null`), causing `addEphemeralErrorListener` to skip registration and leaving subsequent network errors unhandled. Another fix inferred error-handling intent from `onError != null`, but in debug mode `onError` is always set for `_debugBuildErrorWidget`, making it impossible to distinguish real error handlers from debug-only re-throwers.

### Fix

Added `reportErrors` flag to `ImageStreamListener`. The `Image` widget sets this to `false` only when `errorBuilder` is provided. `ImageStreamCompleter` remembers that such a listener was once registered, and skips reporting errors to `FlutterError.onError` after all listeners are removed.

This separates "error handling intent" (`reportErrors`) from "error callback presence" (`onError != null`), so `kDebugMode` behavior no longer interferes with error reporting logic.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/image_test.dart
```diff
@@ -3044,10 +3043,10 @@ void main() {
   });
 
   testWidgets(
-    'errorBuilder prevents FlutterError report only if errorBuilder is non-null when widget is disposed',
```
koji-1009, line 3065: I've updated this test to assert that errors are silenced when the widget is disposed.

If the behavior asserted by the original test satisfies the intended design, then issue #81931 should be closed as "working as intended".

packages/flutter/lib/src/painting/image_stream.dart
justinmc: These docs are no longer strictly true, could you update them?

packages/flutter/test/painting/image_provider_and_image_cache_test.dart
```diff
@@ -56,7 +56,14 @@ void main() {
...
 
-    provider.resolve(ImageConfiguration.empty);
+    final ImageStream stream = provider.resolve(ImageConfiguration.empty);
+    final listener = ImageStreamListener(
+      (ImageInfo info, bool syncCall) {},
+      onError: (Object exception, StackTrace? stackTrace) {
+        throw exception;
+      },
+    );
+    stream.addListener(listener);
```
justinmc: I guess this turns the listener into an "active" listener?

packages/flutter/test/painting/image_provider_and_image_cache_test.dart
```diff
@@ -65,6 +72,7 @@ void main() {
 
     expect(imageCache.statusForKey(provider).untracked, true);
     expect(imageCache.pendingImageCount, 0);
+    stream.removeListener(listener);
```
justinmc: I think it would be better to do this in an addTeardown just after your call to addListener, to avoid leaks.

justinmc: Same for the other tests below.

justinmc (commented):
Thanks for the clear description of the solution. I don't know the original intention of this code but I suspect your approach is right. I left a few minor comments.

packages/flutter/test/painting/image_provider_and_image_cache_test.dart
koji-1009: Thanks for the review. I fixed in 5fe7eeb.

packages/flutter/lib/src/painting/image_stream.dart
koji-1009: I updated the docs.
[1e0fe08](https://github.com/flutter/flutter/pull/180327/commits/1e0fe083096ea484d8b58cdeb9b00638798968dc) [be604b0](https://github.com/flutter/flutter/pull/180327/commits/be604b0485b7938d57741e616e3b01ae1a47052e)

justinmc (commented):
Sorry now I'm second guessing myself on this PR, because I worry about swallowing errors that someone might care about. Why is the current behavior of reporting the error with FlutterErrorDetails.silent still too much? I guess it's just showing up in the real world (though debug mode) too frequently during normal development? Would it help if we added to the error message to explain that the situation more?

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -212,14 +217,25 @@ class ImageStreamListener {
   /// be handled. An error handler can explicitly rethrow the exception reported
   /// to it to safely indicate that it did not handle the exception.
   ///
-  /// If an image stream has no listeners that handled the error when the error
-  /// was first encountered, then the error is reported using
-  /// [FlutterError.reportError], with the [FlutterErrorDetails.silent] flag set
-  /// to true.
+  /// If the image stream has no error listeners (listeners with an [onError]
+  /// callback), the error is silently ignored. This prevents unnecessary
+  /// error noise when errors occur after the widget has been disposed.
```
justinmc: Is this no longer true and should reference preventErrorReporting?

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -212,14 +217,25 @@ class ImageStreamListener {
...
+  /// after this listener is removed.
+  ///
+  /// When true, errors that arrive after this listener is removed (e.g.
+  /// because the widget is disposed) are not reported to
+  /// [FlutterError.onError]. This is useful when [FlutterError.onError] is
+  /// configured to report errors to a server.
+  ///
+  /// Defaults to false. The [Image] widget sets this to true when an
+  /// [Image.errorBuilder] is provided.
+  final bool preventErrorReporting;
```
justinmc: Nit: Maybe `reportErrors` per the styleguide? https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-double-negatives-in-apis

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -533,6 +557,10 @@ abstract class ImageStreamCompleter with Diagnosticable {
   ///    automatically removed after first image load or error.
   void addListener(ImageStreamListener listener) {
     _checkDisposed();
+    // Track that an error-preventing listener was once registered.
+    if (listener.preventErrorReporting) {
+      _hadErrorListener = true;
+    }
```
justinmc: I guess there's no valid use case for removing a preventErrorReporting listener and wanting errors to be reported? In the case of the Image widget it makes sense as-is. Not sure if anything else might use this in this way.

packages/flutter/lib/src/painting/image_stream.dart
justinmc: Or is `reportErrorsAfterRemoval` more descriptive? Though it's longer. Up to you!

justinmc (dismissed):
LGTM with some minor comments. I think this is the best approach I can think of for this problem. Sorry for the delayed review here!

dkwingsmt:
Sorry, I didn't mean to dismiss Justin's approval. Might have been a misclick. I'm working on a second review to this PR.

packages/flutter/lib/src/painting/image_stream.dart
koji-1009: I also couldn't think of a use case beyond `Image`. Since `reportErrors: false` is an opt-in flag, I don't see a scenario where the same completer would later need to opt back in.

packages/flutter/lib/src/painting/image_stream.dart
koji-1009: I renamed it to `reportErrors` and updated the docs!

[09ac5d0](https://github.com/flutter/flutter/pull/180327/commits/09ac5d09c1635457cee52a385ca26d40794b427b)

packages/flutter/lib/src/painting/image_stream.dart
koji-1009: Updated the doc to reflect the current behavior and reference `reportErrors`.

[09ac5d0](https://github.com/flutter/flutter/pull/180327/commits/09ac5d09c1635457cee52a385ca26d40794b427b)

justinmc (dismissed):
LGTM 👍 Thanks for the quick update. And thanks @dkwingsmt for being secondary reviewer.

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -215,11 +220,23 @@ class ImageStreamListener {
   /// If an image stream has no listeners that handled the error when the error
   /// was first encountered, then the error is reported using
   /// [FlutterError.reportError], with the [FlutterErrorDetails.silent] flag set
-  /// to true.
+  /// to true. However, if a listener with [reportErrors] set to false has been
+  /// registered on the completer, the error is not reported.
```
dkwingsmt: nit: The "however" makes the description a bit confusing. I suggest we simplify it as the follow and lead reader to read `[reportErrors]` for more details.
```suggestion
  /// to true. This report is suppressed if a listener whose [reportErrors] is false has ever been
  /// registered on the completer.
```

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -215,11 +220,23 @@ class ImageStreamListener {
...
   final ImageErrorListener? onError;
 
+  /// Whether to report errors to [FlutterError.onError] after this listener
+  /// is removed from the [ImageStreamCompleter].
+  ///
+  /// Defaults to true. When false, errors that arrive after removal are
+  /// silently discarded. This is useful when [FlutterError.onError] is
+  /// configured to report errors to a server.
```
dkwingsmt, line 227: Can you explain why this is the useful case? I think this bug doesn't only affect servers, but in general those who want to suppress all future errors of an image that was ever provided with a listener.

packages/flutter/lib/src/painting/image_stream.dart
```diff
@@ -754,7 +783,9 @@ abstract class ImageStreamCompleter with Diagnosticable {
   /// If no error listeners (listeners with an [ImageStreamListener.onError]
   /// specified) are attached, or if the handlers all rethrow the exception
   /// verbatim (with `throw exception`), a [FlutterError] will be reported using
-  /// [FlutterError.reportError].
+  /// [FlutterError.reportError]. However, if a listener with
+  /// [ImageStreamListener.reportErrors] set to false has been registered on
+  /// this completer, the error is not reported.
```
dkwingsmt: nit
```suggestion
  /// [FlutterError.reportError]. This report is suppressed if 
  /// [ImageStreamListener] whose [ImageStreamListener.reportErrors] is false has ever been registered on
  /// this completer.
```

dkwingsmt (commented):
Sorry for late review. I agree with this approach. Instead of my previous attempt, which  connects error suppressing with the widget, this approach essentially associates error suppressing with the image resource, which better matches the expectation of developers.

It seems to me that this approach completely covers the need for`addEphemeralErrorListener`. Do you think you can replace all usages of `addEphemeralErrorListener` to this approach?

Generally LGTM.

packages/flutter/lib/src/painting/image_stream.dart
koji-1009: The "server" example here follows [`FlutterError.onError`'s own docs](https://api.flutter.dev/flutter/foundation/FlutterError/onError.html), since `reportErrors` controls reporting to that handler.

On the broader point: this is intentionally an opt-in mechanism rather than a default behavior change. An earlier iteration silenced errors more broadly, but [review discussion](https://github.com/flutter/flutter/pull/180327#pullrequestreview-3857816264) flagged that it could swallow errors callers might care about, so we narrowed it to an explicit opt-in.

koji-1009:
Agreed — and the #97077 case (which `addEphemeralErrorListener` in `_stopListeningToStream` was added for) also appears resolved by `reportErrors: false`.

dkwingsmt (dismissed):
LGTM. Thank you!

justinmc (dismissed):
Re LGTM 👍

koji-1009:
Sorry, I fixed the code format.

dkwingsmt (approved):
Re LGTM

justinmc (approved):
Re LGTM, will autosubmit.

Issue #98350, opened 2022-02-13
Add fit, alignment properties to CircleAvatar


kirolous-nashaat:
## Use case
backgroundImageFit > Chaning the default fit for background image to values other than the default (BoxFit.cover)
backgroundImageAlignment > Chaning the default alignment for background image to values other than the default (Center)

foregroundImageFit > Chaning the default fit for foreground image to values other than the default (BoxFit.cover)
foregroundImageAlignment > Chaning the default alignment for foreground image to values other than the default (Center)

Just like on Image widgets, we need to choose the alignment in which we scale the image to fit the circle or change the fit entirely (like fit Height or fit Width).

## Proposal
Add CircleAvatar optional parameters [ Alignment, Fit ] with the default values as is now for both background/foreground images.

kirolous-nashaat:
Updated the issue description for further explaination

goderbauer:
There's another incompleted patch in https://github.com/flutter/flutter/pull/137919.

3M35:
Any progress on this? We really need fit properties with an option to set BoxFit.contain for pictures with uncommon ratio.

justinmc:
There is a PR open that implements a RawAvatar widget that could have implications here. https://github.com/flutter/flutter/pull/180338

PR #180338, opened 2025-12-27
feat: Add RawAvatar

### LINKED ISSUE #180106: Create a RawCircleAvatar, an unthemed variant of CircleAvatar
### Use case

As the Material library is decoupled from the framework, users making their own design languages will not need to import the Material library to build and customize their own CircleAvatars.

### Proposal

Share the logic in CircleAvatar from the core framework.

rkishan516:
Add RawCircleAvatar

Since CircleAvatar already have minimal coupling with Material theme

This is initial implementation, subjected to change based on initial feedback.

fixes #180106 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/circle_avatar.dart
+++ b/packages/flutter/lib/src/material/circle_avatar.dart
@@ -50,11 +50,14 @@ import 'theme.dart';
 /// ```
 /// {@end-tool}
 ///
+/// This widget is backed by [RawAvatar] and uses [ShapeDecoration] internally.
+///
 /// See also:
 ///
 ///  * [Chip], for representing users or concepts in long form.
 ///  * [ListTile], which can combine an icon (such as a [CircleAvatar]) with
 ///    some text for a fixed height list entry.
+///  * [RawAvatar], the underlying widget that powers [CircleAvatar].
 ///  * <https://material.io/design/components/chips.html#input-chips>
 class CircleAvatar extends StatelessWidget {
   /// Creates a circle that represents a user.
--- a/packages/flutter/lib/src/material/circle_avatar.dart
+++ b/packages/flutter/lib/src/material/circle_avatar.dart
@@ -59,7 +59,7 @@ import 'theme.dart';
 ///    some text for a fixed height list entry.
 ///  * [RawAvatar], the underlying widget that powers [CircleAvatar].
 ///  * <https://material.io/design/components/chips.html#input-chips>
-class CircleAvatar extends StatelessWidget {
+class CircleAvatar extends StatefulWidget {
   /// Creates a circle that represents a user.
   const CircleAvatar({
     super.key,
@@ -170,59 +170,102 @@ class CircleAvatar extends StatelessWidget {
   /// the size will snap to 40 pixels instantly.
   final double? maxRadius;
 
+  @override
+  State<CircleAvatar> createState() => _CircleAvatarState();
+}
+
+class _CircleAvatarState extends State<CircleAvatar> {
   // The duration for animating theme changes.
   static const Duration _kTextStyleChangeDuration = Duration(milliseconds: 200);
+
+  // Whether the foreground image has failed to load.
+  bool _foregroundImageFailed = false;
+
+  @override
+  void didUpdateWidget(CircleAvatar oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    // Reset the failed state if the foreground image changes.
+    if (widget.foregroundImage != oldWidget.foregroundImage) {
+      _foregroundImageFailed = false;
+    }
+  }
+
+  void _handleForegroundImageError(Object exception, StackTrace? stackTrace) {
+    setState(() {
+      _foregroundImageFailed = true;
+    });
+    widget.onForegroundImageError?.call(exception, stackTrace);
+  }
 
   @override
   Widget build(BuildContext context) {
     assert(debugCheckHasMediaQuery(context));
     final ThemeData theme = Theme.of(context);
     final Color? effectiveForegroundColor =
-        foregroundColor ?? (theme.useMaterial3 ? theme.colorScheme.onPrimaryContainer : null);
+        widget.foregroundColor ??
+        (theme.useMaterial3 ? theme.colorScheme.onPrimaryContainer : null);
     final TextStyle effectiveTextStyle = theme.useMaterial3
         ? theme.textTheme.titleMedium!
         : theme.primaryTextTheme.titleMedium!;
     TextStyle textStyle = effectiveTextStyle.copyWith(color: effectiveForegroundColor);
     Color? effectiveBackgroundColor =
-        backgroundColor ?? (theme.useMaterial3 ? theme.colorScheme.primaryContainer : null);
+        widget.backgroundColor ?? (theme.useMaterial3 ? theme.colorScheme.primaryContainer : null);
     if (effectiveBackgroundColor == null) {
       effectiveBackgroundColor = switch (ThemeData.estimateBrightnessForColor(textStyle.color!)) {
         Brightness.dark => theme.primaryColorLight,
         Brightness.light => theme.primaryColorDark,
       };
     } else if (effectiveForegroundColor == null) {
-      textStyle = switch (ThemeData.estimateBrightnessForColor(backgroundColor!)) {
+      textStyle = switch (ThemeData.estimateBrightnessForColor(widget.backgroundColor!)) {
         Brightness.dark => textStyle.copyWith(color: theme.primaryColorLight),
         Brightness.light => textStyle.copyWith(color: theme.primaryColorDark),
       };
     }
 
-    final Widget? childContent = child == null
+    final Widget? childContent = widget.child == null
         ? null
         : Center(
             child: MediaQuery.withNoTextScaling(
               child: IconTheme(
                 data: theme.iconTheme.copyWith(color: textStyle.color),
-                child: DefaultTextStyle(style: textStyle, child: child!),
+                child: DefaultTextStyle(style: textStyle, child: widget.child!),
               ),
             ),
           );
 
-    // Convert radius to size (diameter = radius * 2)
-    final double? size = radius != null ? radius! * 2.0 : null;
-    final double? minSize = minRadius != null ? minRadius! * 2.0 : null;
-    final double? maxSize = maxRadius != null ? maxRadius! * 2.0 : null;
+    // Convert radius to constraints (diameter = radius * 2)
+    final BoxConstraints? constraints;
+    if (widget.radius != null) {
+      final double diameter = widget.radius! * 2.0;
+      constraints = BoxConstraints.tightFor(width: diameter, height: diameter);
+    } else if (widget.minRadius != null || widget.maxRadius != null) {
+      constraints = BoxConstraints(
+        minWidth: widget.minRadius != null ? widget.minRadius! * 2.0 : 0.0,
+        minHeight: widget.minRadius != null ? widget.minRadius! * 2.0 : 0.0,
+        maxWidth: widget.maxRadius != null ? widget.maxRadius! * 2.0 : double.infinity,
+        maxHeight: widget.maxRadius != null ? widget.maxRadius! * 2.0 : double.infinity,
+      );
+    } else {
+      constraints = null;
+    }
+
+    // Determine which image to show: foregroundImage (if not failed) or backgroundImage as fallback.
+    final ImageProvider? effectiveImage;
+    final ImageErrorListener? effectiveOnImageError;
+    if (widget.foregroundImage != null && !_foregroundImageFailed) {
+      effectiveImage = widget.foregroundImage;
+      effectiveOnImageError = _handleForegroundImageError;
+    } else {
+      effectiveImage = widget.backgroundImage;
+      effectiveOnImageError = widget.onBackgroundImageError;
+    }
 
     return RawAvatar(
-      size: size,
-      minSize: minSize,
-      maxSize: maxSize,
+      constraints: constraints,
       shape: const CircleBorder(),
       backgroundColor: effectiveBackgroundColor,
-      backgroundImage: backgroundImage,
-      onBackgroundImageError: onBackgroundImageError,
-      foregroundImage: foregroundImage,
-      onForegroundImageError: onForegroundImageError,
+      image: effectiveImage,
+      onImageError: effectiveOnImageError,
       child: AnimatedDefaultTextStyle(
         style: textStyle,
         duration: _kTextStyleChangeDuration,
--- a/packages/flutter/lib/src/material/circle_avatar.dart
+++ b/packages/flutter/lib/src/material/circle_avatar.dart
@@ -2,14 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-/// @docImport 'chip.dart';
-/// @docImport 'color_scheme.dart';
-/// @docImport 'list_tile.dart';
-library;
-
 import 'package:flutter/widgets.dart';
 
-import 'constants.dart';
 import 'theme.dart';
 
 // Examples can assume:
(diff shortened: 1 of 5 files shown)


packages/flutter/lib/src/widgets/raw_circle_avatar.dart
```diff
@@ -0,0 +1,166 @@
...
+/// If [foregroundImage] fails then [backgroundImage] is used. If
+/// [backgroundImage] fails too, [backgroundColor] is used.
+///
+/// The [onBackgroundImageError] parameter must be null if the [backgroundImage]
+/// is null.
+/// The [onForegroundImageError] parameter must be null if the [foregroundImage]
+/// is null.
+///
+/// This widget is not themed. For a Material themed version, see [CircleAvatar].
+class RawCircleAvatar extends StatelessWidget {
```
navaronbracke: This widget should be in the widgets library?

rkishan516: Hey @navaronbracke, I asked gemini to check and update docs but it accidentally moved it to material. I have pushed the change.

justinmc (commented):
Thanks for jumping on this @rkishan516! I'm excited for us to get more Raw* widgets like this.

I'm on board with this but will defer to @victorsanni for the main approval. I picture a range for how unopinionated a raw widget is, from completely headless to drawing UI but allowing customization. This PR definitely falls on the latter side of the scale. It would be possible to make this less opinionated and more headless by leaving all image stuff up to the user, for example.

However my hunch is that for a very simple widget like this, we do want to lean to the opinionated side of the scale. I think most users will find that this API is easy to use and customizable enough for their needs.

But I could be convinced otherwise if anyone has more first hand experience with CircleAvatar.

Last thing @rkishan516, I think you should add at least a couple of tests that cover the basic use cases for RawCircleAvatar directly, like maybe one for using an image and one for using initials.

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
```diff
@@ -0,0 +1,177 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// @docImport 'package:flutter/material.dart';
+library;
+
+import 'package:flutter/widgets.dart';
```
victorsanni: circular import, need to import file by file.

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
```diff
@@ -0,0 +1,177 @@
...
+    final double minDiameter = _minDiameter;
+    final double maxDiameter = _maxDiameter;
+    return AnimatedContainer(
+      constraints: BoxConstraints(
+        minHeight: minDiameter,
+        minWidth: minDiameter,
+        maxWidth: maxDiameter,
+        maxHeight: maxDiameter,
+      ),
+      duration: _kRawCircleAvatarThemeChangeDuration,
```
victorsanni: No theme in widgets/.

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
```diff
@@ -0,0 +1,177 @@
...
+      decoration: BoxDecoration(
+        color: backgroundColor,
+        image: backgroundImage != null
+            ? DecorationImage(
+                image: backgroundImage!,
+                onError: onBackgroundImageError,
+                fit: BoxFit.cover,
+              )
+            : null,
+        shape: BoxShape.circle,
```
victorsanni: is this something that can be exposed?

victorsanni (commented):
The "Circle" in "RawCircleAvatar" feels opinionated. Maybe this is an opportunity to provide a more generic solution?

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
rkishan516: fixed

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
rkishan516: Updated.

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
rkishan516: > It would be possible to make this less opinionated and more headless by leaving all image stuff up to the user.
> The "Circle" in "RawCircleAvatar" feels opinionated. Maybe this is an opportunity to provide a more generic solution?

Based on these, I decided to make RawAvatar which is generic version for showing Avatar.

packages/flutter/lib/src/material/circle_avatar.dart
```diff
@@ -173,28 +167,8 @@ class CircleAvatar extends StatelessWidget {
...
-  }
-
-  double get _maxDiameter {
-    if (radius == null && minRadius == null && maxRadius == null) {
-      return _defaultRadius * 2.0;
-    }
-    return 2.0 * (radius ?? maxRadius ?? _defaultMaxRadius);
-  }
+  // The duration for animating theme changes.
+  static const Duration _kThemeChangeDuration = Duration(milliseconds: 200);
```
justinmc: I think @victorsanni might have mentioned this, but the name "Theme" is misleading here since Widgets has no theming. Maybe _kTextStyleChangeDuration?

packages/flutter/lib/src/widgets/raw_circle_avatar.dart
justinmc: Great idea

justinmc (approved):
LGTM :+1: . Allowing the shape to be passed in is a huge improvement here and makes this widget a lot more versatile. Also RawAvatar strikes me as a really cool name for some reason haha.

Wait for @victorsanni's final approval before merging.

packages/flutter/lib/src/material/circle_avatar.dart
rkishan516: Even though this is in material library but it make sense to update it to `_kTextStyleChangeDuration`.

packages/flutter/lib/src/widgets/raw_avatar.dart
```diff
@@ -0,0 +1,201 @@
...
+    this.backgroundColor,
+    this.foregroundColor,
+    this.backgroundImage,
+    this.foregroundImage,
+    this.onBackgroundImageError,
+    this.onForegroundImageError,
+    this.shape,
+    this.size,
+    this.minSize,
+    this.maxSize,
```
victorsanni: I don't see a reason for `minSize` and `maxSize` since they are already covered by `size`.

packages/flutter/lib/src/widgets/raw_avatar.dart
```diff
@@ -0,0 +1,201 @@
...
+  ///
+  /// Typically used as a fallback image for [foregroundImage].
+  ///
+  /// If the [RawAvatar] is to have the user's initials, use [child] instead.
+  final ImageProvider? backgroundImage;
+
+  /// The foreground image of the avatar.
+  ///
+  /// Typically used as profile image. For fallback use [backgroundImage].
+  final ImageProvider? foregroundImage;
```
victorsanni: I'm wondering why we need both a foreground and a background image. If the developer wants a child in front of an image they can provide both `child` and `backgroundImage`. If they want just an image, they can provide just a `backgroundImage`.

Is there some use case I'm missing?

packages/flutter/lib/src/widgets/raw_avatar.dart
```diff
@@ -0,0 +1,201 @@
...
+
+  /// An optional error callback for errors emitted when loading
+  /// [backgroundImage].
+  final ImageErrorListener? onBackgroundImageError;
+
+  /// An optional error callback for errors emitted when loading
+  /// [foregroundImage].
+  final ImageErrorListener? onForegroundImageError;
```
victorsanni: same here.

packages/flutter/lib/src/widgets/raw_avatar.dart
```diff
@@ -0,0 +1,201 @@
...
+import 'framework.dart';
+import 'text.dart';
+
+/// A raw avatar widget that represents a user.
+///
+/// This widget provides the core functionality for avatars with customizable
+/// shape, background/foreground colors, and images. It handles sizing
+/// constraints and builds the appropriate decorations.
+///
+/// For a Material Design styled avatar, see [CircleAvatar].
```
victorsanni: Make this the `See also` section. Instead of having it in the widget API docs description.

packages/flutter/test/material/circle_avatar_test.dart
```diff
@@ -28,7 +28,7 @@ void main() {
     final RenderConstrainedBox box = tester.renderObject(find.byType(CircleAvatar));
     expect(box.size, equals(const Size(100.0, 100.0)));
     final child = box.child! as RenderDecoratedBox;
-    final decoration = child.decoration as BoxDecoration;
+    final decoration = child.decoration as ShapeDecoration;
```
victorsanni, line 31: Is this a breaking change? Worth documenting?

victorsanni (commented):
I had some questions about reducing the API surface. I haven't played with this locally, but I'm wondering about how RawAvatar clips images to fit the different shapes that could be provided.

justinmc:
@rkishan516 I just stumbled on this issue asking for `fit` and `alignment` parameters in CircleAvatar. Would RawAvatar make this any easier? Is it something we should make sure to accommodate here? https://github.com/flutter/flutter/issues/98350

packages/flutter/lib/src/widgets/raw_avatar.dart
rkishan516: I think just like minRadius and maxRadius, minSize and maxSize will allow avatar to be flexible and resize within the range.

packages/flutter/lib/src/widgets/raw_avatar.dart
rkishan516: Its all actually coming from previous implementation of `CircleAvatar`. Based on that
It's a fallback chain:                                                                                                                                   
  1. foregroundImage - the actual profile picture (rendered on top)                                                                                           
  2. backgroundImage - fallback if foreground fails to load (shows through)                                                                                   
  3. backgroundColor - final fallback if both images fail

rkishan516:
> @rkishan516 I just stumbled on this issue asking for fit and alignment parameters in CircleAvatar. Would RawAvatar make this any easier? Is it something we should make sure to accommodate here? https://github.com/flutter/flutter/issues/98350

I think `fit` and `alignment` will be directly pass down to RawAvatar.
Yes, we need to add `fit` and `alignment` as property in RawAvatar.

packages/flutter/lib/src/widgets/raw_avatar.dart
rkishan516: Updated

packages/flutter/test/material/circle_avatar_test.dart
rkishan516: Updated.

packages/flutter/lib/src/widgets/raw_avatar.dart
victorsanni: In that case, why does `size` exist if it can already be configured by `minSize` and `maxSize`? Duplicate APIs should be avoided where possible.

packages/flutter/lib/src/widgets/raw_avatar.dart
```diff
@@ -0,0 +1,217 @@
...
+  double get _maxSize {
+    if (size == null && minSize == null && maxSize == null) {
+      return _defaultSize;
+    }
+    return size ?? maxSize ?? _defaultMaxSize;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final double minDimension = _minSize;
```
victorsanni: The fact that two lines of the build method are dedicated to renaming variables tells me these should probably be the property names. The "-size" postfix here is misleading because the properties are `double`s, not `Size`s.

packages/flutter/lib/src/widgets/raw_avatar.dart
victorsanni: I think it's important to remember here that we are creating a new widget (independent from the widget being decoupled), and so we have an opportunity here to design a more intuitive and simpler API (and avoid the pitfalls of the previous one).

I can understand an image falling back to a background color as two separate properties. But IMO the developer should handle image-to-image fallbacks if their images don't load so the API is simpler. Especially since providing an optional `child` anyways will place the `child` on top of the image.

Is there another use case for having both images I might be missing? If having both images is for backwards compatibility with `CircleAvatar`, is there a way to have logic in `CircleAvatar` that determines the appropriate image to send to `RawAvatar`?

victorsanni (commented):
Many of the google testing failures had a CircleAvatar with only a backgroundImage and a radius set. I'm wondering if making CircleAvatar stateful might be causing these failures.

victorsanni:
Hi @rkishan516, I confirmed that some tests are broken because CircleAvatar is now stateful, can this work with a Stateless CircleAvatar?

packages/flutter/lib/src/material/circle_avatar.dart
```diff
@@ -173,95 +170,107 @@ class CircleAvatar extends StatelessWidget {
...
     }
-    return 2.0 * (radius ?? minRadius ?? _defaultMinRadius);
   }
 
-  double get _maxDiameter {
-    if (radius == null && minRadius == null && maxRadius == null) {
-      return _defaultRadius * 2.0;
-    }
-    return 2.0 * (radius ?? maxRadius ?? _defaultMaxRadius);
+  void _handleForegroundImageError(Object exception, StackTrace? stackTrace) {
```
victorsanni: I'm okay with having both `foregroundImage` and `backgroundImage` in `RawAvatar`'s API if we can remove this method and make the widget stateless.

Is there another alternative?

rkishan516:
> Still a ton of google testing failures.
> 
> Example:
> 
> ```
> const CircleAvatar(
>     maxRadius: 16,
>     backgroundImage: AssetImage('assets/images/avatar1.png'),
> ),
> ```
> 
> Without the changes in this PR, the `CircleAvatar` is correctly shown with the `backgroundImage` etc. With the current changes, the `CircleAvatar` is not shown.

Could you please share the specific image, because I am able to see other images (even the cached ones)?

rkishan516:
> > Could you please share the specific image, because I am able to see other images (even the cached ones)?
> 
> I mean with an arbitrary image. Can you reproduce the bug introduced by this PR when you provide a `CircleAvatar` with a small `maxRadius` and an arbitrary `AssetImage`? This is my local repro:
> 
> Sample code

Ran the code, i am still not able to figure out whats the problem. Even ran golden test for before and after and its passing.
circle_avatar_app
image

rkishan516:
> This repro has no assets:
> 
> ```dart
> import 'package:flutter/material.dart';
> 
> void main() => runApp(const CircleAvatarApp());
> 
> class CircleAvatarApp extends StatelessWidget {
>   const CircleAvatarApp({super.key});
> 
>   @override
>   Widget build(BuildContext context) {
>     return MaterialApp(
>       home: Scaffold(
>         appBar: AppBar(
>           title: const Text('CircleAvatar Properties'),
>         ),
>         body: Center(
>           child: SingleChildScrollView(
>             child: Column(
>               mainAxisAlignment: MainAxisAlignment.center,
>               children: [
>                 // Image disappears
>                 const CircleAvatar(
>                   //minRadius: 40,  // Uncomment to make image reappear
>                   maxRadius: 66,
>                   backgroundImage: NetworkImage(
>                     'https://docs.flutter.dev/assets/images/dash/dash-fainting.gif',
>                   ),
>                 ),
>               ],
>             ),
>           ),
>         ),
>       ),
>     );
>   }
> }
> ```
> 
> Notice that with only `maxRadius` set the gif is not displayed.

I don't know how this specific case is also working for me
image

I don't know but can it be a platform specific thing?

rkishan516:
> I created a dartpad code sample to show this bug: https://dartpad.dev/100b78ab9d24c89a5e32fd1c4b49198d

Thank you, I don't know how but fix for this change has been on local branch and it never got pushed. So, I was always trying correct version and PR code was always faulty.

I have pushed the change, it should work now.

PR #180339, opened 2025-12-27
Modify the SemanticTab class to create <a> element when the node has …

### LINKED ISSUE #180321: Link does not work with Tab
In web, this code does not produce functionality of Link:

```
Link(
   uri: Uri.parse('http://google.com'),
   builder: (_, __) => const Tab(text: 'google')),
```

As result, if tabs are used for the app menu:
1. There is no ability for user to right-click or tap a menu item in order to copy its url
2. The links for menu items are not indexed by search engines

tridaxdev:
## Enable link functionality for tabs with linkUrl on Flutter Web

This PR enables proper link functionality for tabs that have a `linkUrl` set on Flutter Web. Previously, when using tabs for navigation menus with URLs (e.g., wrapping a `Tab` with the `Link` widget from `url_launcher`), the link semantics were lost because `SemanticTab` always rendered as a generic `<flt-semantics>` element with `role="tab"`. fixes #180321

### Problem

When tabs are used for app navigation menus on web:
- Users cannot right-click or tap a menu item to copy its URL
- Links for menu items are not indexed by search engines
- The native browser link behavior is completely lost

### Solution

Modified `SemanticTab` to:
1. Render as an `<a>` element when `linkUrl` is present, while maintaining `role="tab"` for accessibility
2. Set the `href` attribute based on the `linkUrl` value
3. Update the `href` attribute dynamically when `linkUrl` changes

This results in HTML like `<a role="tab" href="https://example.com">Tab Label</a>`, which provides:
- ✅ Proper tab accessibility for screen readers (via `role="tab"`)
- ✅ Right-click to copy URL functionality
- ✅ Middle-click to open in new tab
- ✅ Search engine indexing of navigation links

### Before
```html
<flt-semantics role="tab">Tab Label</flt-semantics>
```

### After (when linkUrl is set)
```html
<a role="tab" href="https://example.com" style="display: block;">Tab Label</a>
```

Fixes https://github.com/flutter/flutter/issues/XXXXX

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/tabs.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/tabs.dart
@@ -1,7 +1,7 @@
 // Copyright 2013 The Flutter Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
-
+import '../dom.dart';
 import 'label_and_value.dart';
 import 'semantics.dart';
 
@@ -20,6 +20,33 @@ class SemanticTab extends SemanticRole {
       ) {
     setAriaRole('tab');
     addTappable();
+  }
+
+  @override
+  DomElement createElement() {
+    // When the tab has a link URL, render as an <a> element to enable
+    // proper link functionality (right-click copy, SEO indexing) while
+    // still using role="tab" for accessibility.
+    if (semanticsObject.hasLinkUrl) {
+      final DomElement element = domDocument.createElement('a');
+      element.style.display = 'block';
+      return element;
+    }
+    return super.createElement();
+  }
+
+  @override
+  void update() {
+    super.update();
+
+    // Update the href attribute when the link URL changes.
+    if (semanticsObject.isLinkUrlDirty) {
+      if (semanticsObject.hasLinkUrl) {
+        element.setAttribute('href', semanticsObject.linkUrl!);
+      } else {
+        element.removeAttribute('href');
+      }
+    }
   }
 
   @override
--- a/engine/src/flutter/lib/web_ui/test/engine/semantics/semantics_test.dart
+++ b/engine/src/flutter/lib/web_ui/test/engine/semantics/semantics_test.dart
@@ -605,6 +605,91 @@ void _testEngineSemanticsOwner() {
     owner().updateSemantics(builder.build());
   }
 
+    test('tab with linkUrl renders as anchor element with href', () {
+    semantics()
+      ..debugOverrideTimestampFunction(() => _testTime)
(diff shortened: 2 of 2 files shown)


engine/src/flutter/lib/web_ui/lib/src/engine/semantics/tabs.dart
```diff
@@ -22,6 +22,33 @@ class SemanticTab extends SemanticRole {
...
+    }
+    return super.createElement();
+  }
+
+  @override
+  void update() {
+    super.update();
+
+    // Update the href attribute when the link URL changes.
+    if (semanticsObject.isLinkUrlDirty) {
```
chunhtai, line 43: cc @mdebbar If link meant to be composable, should it be a semantics behavior?

mdebbar: It makes sense to have a semantics behavior for updating the `href` attribute. Then this behavior can be used by the link and tab roles.

But behaviors can't control the creation of the DOM element. That's done by the role. Which brings up another question: what happens when a semantics tab has a link, and then it gets updated with no link? There's no mechanism to replace the DOM element of the role.

I can think of 2 ideas here:

1. Make a new role called `SemanticTabLink`.
2. Make the link a child of the tab. I'm not sure if this can be done at the framework level, but we can do it in the engine. The `SemanticTab` needs to:
    * Create a `<a>` element as a child
    * Override other methods (e.g. `append()` to make sure children are appended inside the `<a>` element).
    * In `update()`, if the node has no link anymore, the `<a>` child is removed.

chunhtai: > Make a new role called SemanticTabLink.

not a new SemanticsRole on the framework side right?  I think the engine can infer that from SemanticsRole.tab, and isLink (or linkUrl) is true?

> Make the link a child of the tab. I'm not sure if this can be done at the framework level, but we can do it in the engine. The SemanticTab needs to:

Framework now forces a tab to only create one node, I remember that is for a internal use cases.

From dom's perspective, I am not sure how screen reader treats
`<div role="tab"><a href="..">my tab</a></div>` I wonder if that will introduce an extra swipe or missing any information in this structure

mdebbar: > > Make a new role called SemanticTabLink.
> 
> not a new SemanticsRole on the framework side right? I think the engine can infer that from SemanticsRole.tab, and isLink (or linkUrl) is true?

Not a framework role. I meant an engine SemanticTabLink that's used when role is tab and `isLink` is true.

PR #180348, opened 2025-12-28
Fix a11y contrast for pre-test message in flutter_test

ashoknarayan:
## Pre-launch Checklist

- [ ] I signed the [CLA].

## Works done
- The pre-test message "Test starting..." had a contrast ratio of 2.99:1 against the background.
- This PR slightly adjusts the text color to 0xFF8F7FFF to meet the 3:1 minimum contrast ratio for accessibility.
- Tested by running flutter_test.


packages/flutter_test/lib/src/binding.dart
```diff
@@ -1265,7 +1265,7 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
     return announcements;
   }
 
-  static const TextStyle _messageStyle = TextStyle(color: Color(0xFF917FFF), fontSize: 40.0);
+  static const TextStyle _messageStyle = TextStyle(color: Color(0xFF8F7FFF), fontSize: 40.0);
```
justinmc, line 1268: What's the reason for this color change?

justinmc (commented):
@ashoknarayan Is there an existing issue that you're trying to solve? I'm missing context on this. Explaining that we need the leading "v" in version numbers makes sense.

justinmc (commented):
Ah got it. Would you be able to split this PR in two? I agree with Gemini here. One for the versioning text and one for the contrast.

PR #180375, opened 2025-12-29
Adds geometry dirty nodes

### LINKED ISSUE #166173: [A11y] Further optimize geometry calculation in semantics

chunhtai:
fixes https://github.com/flutter/flutter/issues/166173

Highlevel:

instead of clearing the entire geometry from the node that markSemanticsNeedsUpdate up to the first semantics boundary. and recalculate everything in between. This code now only clear the geometry for node that are  markSemanticsNeedsUpdate. This requires us to keep a separate list for geomtry dirty node. Thus this change.

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1489,73 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+                    !object._semantics.parentDataDirty,
+              )
+              .toList()
+            ..sort((RenderObject a, RenderObject b) => a.depth - b.depth);
+      _nodesNeedingSemanticsGeometryUpdate.clear();
+
+      // Clear geometry for nodes that needs geometry update.
+      for (final node in nodesToProcessGeometry) {
+        if (node._semantics.shouldFormSemanticsNode && node._semantics.geometryDirty) {
+          // This is node is already dirty, skip it.
```
justinmc: "This is node" => "This node"

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1489,73 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+        _RenderObjectSemantics target = node._semantics;
+        // All the parent data on the active tree should be up to date at this point
+        // except for the blocked branch.
+        //
+        // We MUST NOT update the geometry of the blocked branch since ensureGeometry short
+        // circuited when it reaches a node whose geometry is up to date. Updating
+        // the geometry of the blocked branch half-heartly will cause a gap of render
+        // object with dirty geometry where the future update may never reach.
+        //
+        // Either we update entire tree regardless of been blocked or not, or we only update
```
justinmc: "been blocked" => "whether it is blocked"

justinmc (approved):
Makes sense to me at a high level, but @LongCatIsLooong should definitely review and follow up on his old ideas (https://github.com/flutter/flutter/issues/166173#issuecomment-3404030961). Otherwise LGTM 👍

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1489,73 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+      // It is possible the updateChildren above caused some nodes to be added
+      // to _nodesNeedingSemanticsGeometryUpdate. Therefore, we need to
+      // process them here.
+      final List<RenderObject> nodesToProcessGeometry =
+          _nodesNeedingSemanticsGeometryUpdate
+              .where(
+                (RenderObject object) =>
+                    !object._needsLayout &&
+                    object.owner == this &&
+                    !object._semantics.parentDataDirty,
```
LongCatIsLooong: Is this for skipping dirty RenderObjectSemantics that will need geometry updates anyways because the parent data also needs updating? I'm thinking maybe instead of boolean flags we should be using an int/enum for indicating the dirtiness?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1489,73 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+              )
+              .toList()
+            ..sort((RenderObject a, RenderObject b) => a.depth - b.depth);
+      _nodesNeedingSemanticsGeometryUpdate.clear();
```
LongCatIsLooong, line 1515: uber nit: should we assert this list is empty in `finalizeTree` to make sure nobody is making geometry changes after the semantics flush step?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1489,73 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+      for (final node in nodesToProcessGeometry) {
+        if (node._semantics.shouldFormSemanticsNode && node._semantics.geometryDirty) {
+          // This is node is already dirty, skip it.
+          continue;
+        }
+        if (!node._semantics.contributesToSemanticsTree) {
+          // This node merely presents its subtree in the mergeup, so we need to clear
+          // the geometry for all the semantics nodes in the mergeup.
+          for (final _RenderObjectSemantics child
+              in node._semantics.mergeUp.whereType<_RenderObjectSemantics>().where(
```
LongCatIsLooong: Is `mergeUp.whereType<_RenderObjectSemantics>()` always safe even when there're WidgetSpans in the tree?

LongCatIsLooong (commented):
Hmm could you add a comment somewhere to explain when a RenderObject gets added to `_nodesNeedingSemanticsGeometryUpdate`? For a RenderObject author what should I do to make sure the semantics geometry is updated, when I skip layout and only do repaint (I think you can safely do that to re-align a child today since the child subtree's layout does not change at all). From the looks of it the RenderObject author must call `markNeedsSemanticsUpdate` when they re-align the child?

packages/flutter/lib/src/rendering/object.dart
chunhtai: so it is slightly complicated, if at this point the parent data is still dirty, it is a blocked branch. so `parentDataDirty` doesn't imply it is blocked, it depends on the phase of update as well. I will add some more comment

packages/flutter/lib/src/rendering/object.dart
chunhtai: yes, at this point, it should have been up to date

chunhtai: or are you asking about incompletefragment that generated by childConfigurationsDelegate?

the answer is still yes. These fragment won't create a semantics node so there is no geometry to calculate. They will be filtered out by the whereType

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,77 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+                (_RenderObjectSemantics e) => e.shouldFormSemanticsNode,
+              )) {
+            child.geometry = null;
+          }
+          continue;
```
LongCatIsLooong, line 1556: nit: the long expression in the for loop and the continue makes it hard to read on github since it's hard to tell where the for block and if block end.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5631,6 +5731,15 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
         in result.$1.whereType<_RenderObjectSemantics>()) {
       assert(childSemantics.contributesToSemanticsTree);
       if (childSemantics.shouldFormSemanticsNode) {
+        // In general geometry is only dirty during the first build or markNeedsLayout.
+        // In both cases, the renderobject will be added to the list of nodes needing
+        // geometry update.
+        //
+        // This specific casecan happen if a renderobject decided to conditionally include
```
LongCatIsLooong: nit: missing space between "case" and "can". By this sentence do you mean that the `visitChildrenForSemantics` of that RenderObject having a branching condition which doesn't call `markNeedsSemanticsUpdate` when changed? Is there an example?

LongCatIsLooong: Since geometry is only reset to `null` after `updateChildren` in the `ensureGeometry` step (barring `clear`), so this is for a brand new subtree that has never been incorporated into the tree?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,77 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+        // Either we update entire tree regardless of whether it is blocked or not, or we only update
+        // the tree that is not blocked. If we go with the first option, we will waste
+        // time updating something that will not be showing up in the final semantics tree.
+        //
+        // Thus, we go for the second option, and the geometry update will only be conducted
+        // once the branch is not blocked.
+        bool isBlocked = target.parentDataDirty;
+        // Find the first render object semantics that forms a semantics node
+        // and whose geometry is not dirty as the anchor point to start the geometry update.
+        while (!isBlocked &&
```
LongCatIsLooong: Would it be easier /faster if each RenderObjectSemantics knows who the parent RenderObjectSemantics that formed the semantics boundary is?

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: for a RenderObjectSemantics to find out if it is in a blocked branch, does it only have to check its own `parentDataDirty` or it has to check all the way up to the root node? I assume its the former based on the implementation here, so even for a "blocked" branch we still recursively update `parentDataDirty` for all descendant nodes that are going to be blocked?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,77 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+          for (final _RenderObjectSemantics child
+              in node._semantics.mergeUp.whereType<_RenderObjectSemantics>().where(
+                (_RenderObjectSemantics e) => e.shouldFormSemanticsNode,
+              )) {
+            child.geometry = null;
+          }
+          continue;
+        }
+        for (final _RenderObjectSemantics child in node._semantics._children) {
+          child.geometry = null;
```
LongCatIsLooong, line 1563: It looks like we're only resetting `geometry` back to `null` for nodes that will produce a semantics node? Would it be possible that a node had `contributeToSemanticsTree == true` but then it becomes false and then becomes true again? When does the `geometry` cache gets evicted in this case?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,77 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+        }
+      }
+
+      // Conduct the geometry update
+      for (final node in nodesToProcessGeometry) {
+        _RenderObjectSemantics target = node._semantics;
+        // All the parent data on the active tree should be up to date at this point
+        // except for the blocked branch.
```
LongCatIsLooong: nit: except for blocked braches? I thought there could be multiple?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,77 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+          // This node is already dirty, skip it.
           continue;
         }
-        node._semantics.ensureGeometry();
+        if (!node._semantics.contributesToSemanticsTree) {
+          // This node merely presents its subtree in the mergeup, so we need to clear
+          // the geometry for all the semantics nodes in the mergeup.
+          for (final _RenderObjectSemantics child
+              in node._semantics.mergeUp.whereType<_RenderObjectSemantics>().where(
+                (_RenderObjectSemantics e) => e.shouldFormSemanticsNode,
```
LongCatIsLooong: Hmm if `node` does not contributes to the semantics tree, but it has a child in `mergeUp` that doesn't form a semantics node, that particular child branch's geometry cache won't be evicted? That doesn't sound correct or am I missing anything? 

IOW why do we mark a RenderObjectSemantics' geometry dirty only if it forms a semantics node itself? Shouldn't we keep walking down the tree to find the closest children with semantics nodes? The `node` may very well changed the semantics clipping to that `child` but `ensureSemantics` would skip that branch because the geometry cache isn't evicted I think?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -2592,6 +2678,9 @@ abstract class RenderObject with DiagnosticableTreeMixin implements HitTestTarge
       return true;
     }());
     owner!._nodesNeedingLayout.add(this);
+    if (owner!._semanticsOwner != null) {
+      owner!._nodesNeedingSemanticsGeometryUpdate.add(this);
```
LongCatIsLooong: Why this has to be done in `scheduleInitialLayout`? Isn't `scheduleInitialSemantics` enough? When a render object does layout `markNeedsSemanticsUpdate` will be called I think?

LongCatIsLooong (commented):
So from reading this my observations are:

Every `RenderObject` will be added to `_nodesNeedingSemanticsGeometryUpdate` when it's `markedNeedsSemanticsUpdate`. But at that point we can't skip adding them to the set because the semantics tree structure has not been updated and we do not know which of them will form semantics nodes and who their parents will be after `updateChildren`.

For that reason, it is after `updateChildren`, that we traverse the `_nodesNeedingSemanticsGeometryUpdate` to figure out which branches need geometry updates and reset `geometry` to `null` to evict the geometry cache and to indicate that these branches will need to be processed for geometry update. 

After that step we "Conduct the geometry update". This is done by locating geometry update roots, from the nodes in `_nodesNeedingSemanticsGeometryUpdate`, and calling `ensureGeometry` on these "root" nodes. `ensureGeometry` walks down the tree and if one of its immediate child still has geometry cache it would skip that branch.

(Reminder to myself that I still have a question regarding sibling merge groups).

packages/flutter/lib/src/rendering/object.dart
chunhtai: I think we could, but since this is easy enough and efficient enough to find it on the spot, I try avoid adding more cache to this class unless necessary. That will probably create some corner cases.

packages/flutter/lib/src/rendering/object.dart
chunhtai: > does it only have to check its own parentDataDirty

yes

It is a delicate condition that makes this true, the node in the list should have their parentdata clear in _RenderObjectSemantics.markNeedsUpdate is called. if it is still dirty here, it means the updateChild hasn't update the the parentdata of this node, and that can only happen if it is blocked.

This is controlled by this line in _RenderObjectSemantics.markNeedsUpdate
```
    if (node != renderObject && producedSemanticsNode != null && node._semantics.parentDataDirty) {
      // If `this` node has already been added to [owner._nodesNeedingSemantics]
      // remove it as it is no longer guaranteed that its semantics
      // node will continue to be in the tree. If it still is in the tree, the
      // ancestor `node` added to [owner._nodesNeedingSemantics] at the end of
      // this block will ensure that the semantics of `this` node actually gets
      // updated.
      // (See semantics_10_test.dart for an example why this is required).
      renderObject.owner!._nodesNeedingSemantics.remove(renderObject);
    }
```
This is where the block node be excluded from updateChildren if it is blocked (if it is not blocked, the parent of this semantics boundary will call updateChildren on this subtree) 

> so even for a "blocked" branch we still recursively update parentDataDirty for all descendant nodes that are going to be blocked?

For a node in a  blocked branch that got added to this list, its parentData **must** be dirty at this point due to the reason above, so the situation you mentioned should never happen

packages/flutter/lib/src/rendering/object.dart
chunhtai: > that particular child branch's geometry cache won't be evicted? That doesn't sound correct or am I missing anything?

Because the geometry is calculated and cached on _RenderObjectSemantics that will form semantics node.

If you look the ensureSemantics, the code is walking the _RenderObjectSemantics._children which should only contain the nodes that forms semantics node.

This means the geometry in the intermediate nodes doesn't matter, so there is no point to evict them. 

> Shouldn't we keep walking down the tree to find the closest children with semantics nodes

yeah I think I missed this part, let me update the code

packages/flutter/lib/src/rendering/object.dart
chunhtai: updated the code

packages/flutter/lib/src/rendering/object.dart
chunhtai: yes in that case markNeedsSemanticsUpdate is called, and the cache is evicted when that method is called

packages/flutter/lib/src/rendering/object.dart
chunhtai: updated

packages/flutter/lib/src/rendering/object.dart
chunhtai: I suspect this is for some corner case, but I forgot, let me remove it and see how it goes

packages/flutter/lib/src/rendering/object.dart
chunhtai: I think this may be an old code that no longer needed, that me remove it and see

packages/flutter/lib/src/rendering/object.dart
chunhtai: updated

packages/flutter/lib/src/rendering/object.dart
chunhtai: ok turns out this is a corner case

I added the code comment explain the situation, I will also paste it here

        // This specific case can happen if a parent somwhere above this node decided to
        // conditionally include this node.
        //
        // For example, render object A->B->C, where A is the root and all of them
        // creates semantics nodes.
        //
        // Frame 1: B and C both mark layout dirty, and B suddenly decide not to
        // inlcude C in the semantics tree.
        //
        // In this case, C's geometry will remain dirty because it is considered as
        // blocked.
        //
        // Frame 2: B is marked dirty again and decide C should be included
        // in the semantics tree. We will run into this situation where C's geometry
        // is dirty but it is not in the _nodesNeedingSemanticsGeometryUpdate.

packages/flutter/lib/src/rendering/object.dart
chunhtai: looks like this doesn't trigger any error

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: I think this patch also removes `node._semantics.geometry = null;` from `markNeedsUpdate`? So `geometry` is only reset to `null` in the parent `for` loop here, and in `clear`?

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: Thanks for the explanation! Interestingly when I comment out that `if` block all tests still pass (I had to use `--update-goldens` tho since golden tests never pass without that on my macbook).

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,85 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+              for (final _RenderObjectSemantics nodeInSubtree in child._children) {
+                assert(nodeInSubtree.shouldFormSemanticsNode);
+                nodeInSubtree.geometry = null;
+              }
+            }
+          }
+          continue;
+        }
+        for (final _RenderObjectSemantics child in node._semantics._children) {
+          child.geometry = null;
```
LongCatIsLooong, line 1563: Also why only removing the geometry cache on child nodes? In this case if `node` also happens to be a relayout boundary that changed its size (but its parent wasn't notified because it was a relayout boundary), if we only remove `geometry` from `node`'s children, then in the "conduct the geometry update" step, `ensureGeometry` will be called on `node` but not on its semantics parent? But in this case I think it should be called on the semantics parent because `node`'s geometry will need updating right?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,85 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+          for (final _RenderObjectSemantics child
+              in node._semantics.mergeUp.whereType<_RenderObjectSemantics>()) {
+            if (child.shouldFormSemanticsNode) {
+              child.geometry = null;
+            } else {
+              // Even though this node does not form a semantics node, it still
+              // represents a group of render objects in the subtree, where node that
+              // forms a semantics node is in its _children.
+              for (final _RenderObjectSemantics nodeInSubtree in child._children) {
+                assert(nodeInSubtree.shouldFormSemanticsNode);
```
LongCatIsLooong, line 1551: hmm interesting, so the merge-up tree is guaranteed to have a max depth of 2?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,12 +1505,85 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+      _nodesNeedingSemanticsGeometryUpdate.clear();
+
+      // Clear geometry for nodes that needs geometry update.
+      for (final node in nodesToProcessGeometry) {
+        if (node._semantics.shouldFormSemanticsNode && node._semantics.geometryDirty) {
+          // This node is already dirty, skip it.
           continue;
         }
-        node._semantics.ensureGeometry();
+        if (!node._semantics.contributesToSemanticsTree) {
```
LongCatIsLooong, line 1539: So here `contributesToSemanticsTree` is checked instead of `shouldFormSemanticsNode` because `_children` is only updated in `updateChildren` if `contributesToSemanticsTree` is true? And that's less children that needs to be marked dirty for geometry?

packages/flutter/lib/src/rendering/object.dart
chunhtai: > I think this patch also removes node._semantics.geometry = null; from markNeedsUpdate? So geometry is only reset to null in the parent for loop here, and in clear?

yes, markNeedsUpdate adds node to _nodesNeedingSemanticsGeometryUpdate which is going to have their cache remove here.

for your original question

> Would it be possible that a node had contributeToSemanticsTree == true but then it becomes false and then becomes true again

its markNeedsUpdate will be called and have its cache removed here.

packages/flutter/lib/src/rendering/object.dart
chunhtai: >  in this case I think it should be called on the semantics parent because node's geometry will need updating right?

this is already the case. the `target` after the while loop must either be a root node, or shouldformsemanticsnode while geometry is not dirty

ensureGeometry() is called on the immediate parent, that form a semantics node, of the node in _nodesNeedingSemanticsGeometryUpdate.

packages/flutter/lib/src/rendering/object.dart
chunhtai: I think some of the google tests may fail, that is why I know about this blocked node shenanigan.

In the long term we should remove the block node logic as it cause state loss for live region which is not good. There is currently an effort to remove blockSemantics usage and hopefully deprecates it.

packages/flutter/lib/src/rendering/object.dart
chunhtai: right

packages/flutter/lib/src/rendering/object.dart
chunhtai: all the immediate semantics node are cached for non-semantics-node-forming renderbojectsemantics until it reach the a renderbojectsemantics that will form a semantics node.

It is this way to avoid walking the intermediate nodes to collect all renderbojectsemantics that will form semantics node

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: > ensureGeometry() is called on the immediate parent, that form a semantics node

```
      for (final node in nodesToProcessGeometry) {
        _RenderObjectSemantics target = node._semantics;
        bool isBlocked = target.parentDataDirty;
        while (!isBlocked &&Collapse
            !target.isRoot &&
            (target.geometryDirty || !target.shouldFormSemanticsNode)) {
          target = target.parent!;
          isBlocked = target.parentDataDirty;
        }

        if (!isBlocked) {
          target.ensureGeometry();
        }
      }
```

if `node.shouldFormSemanticsNode`  is true, we won't go into the `while` loop at all I think? In that case the final `target` will be `node` instead of `node.parent`? But if `node`'s sized changed, it should be `node.parent.ensureGeometry()` instead of `node.ensureGeometry`?

LongCatIsLooong (approved):
LGTM with questions regarding the RelayoutBoundary + shouldFormSemanticsNode case (see https://github.com/flutter/flutter/pull/180375#discussion_r2785119064).

packages/flutter/lib/src/rendering/object.dart
chunhtai: interesting, I think you are right that this is a corner case. IIRC before the whole renderobjectsemantics refactoring, it is already like this that only the parent semantics node would ever update the geometry of the current node. I guess we just never hit this corner case.

let me look into this and get back to you

chunhtai: fixed the corner case by clear the geometry cache if the node shouldformSemanticsnode= true.
also added tests in
semantics_layout_sized_test.dart

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1417,8 +1417,25 @@ base class PipelineOwner with DiagnosticableTreeMixin {
   }
 
   bool _debugDoingSemantics = false;
+
+  /// Nodes that need semantics update.
+  ///
+  /// When [_RenderObjectSemantics.markNeedsUpdate] are called, their closest
```
justinmc: "are called" => "is called"

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1417,8 +1417,25 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+  /// semantics boundary ancestor is added to this set.
+  ///
+  /// All nodes in this set are guaranteed to be semantics boundaries.
   final Set<RenderObject> _nodesNeedingSemantics = <RenderObject>{};
```
justinmc: Nit: Should the name be _nodesNeedingSemanticsUpdate, on order to align with _nodesNeedingSemanticsGeometryUpdate? Or I guess you could drop the "Update" from that one. Not a big deal.

packages/flutter/lib/src/rendering/object.dart
justinmc: Also,what exactly is "their" referring to? Maybe rephrase, this is my best guess but I may be misundestanding:

"
the closest semantics boundary ancestor RenderObject to the _RenderObjectSemantics is added to this set.
"

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1417,8 +1417,25 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
 
+  /// Nodes that need semantics geometry update.
+  ///
+  /// Nodes are added when their [_RenderObjectSemantics.markNeedsUpdate] are called.
```
justinmc: "are called" => "is called"

justinmc: Or I guess another option would be "[_RenderObjectSemantics.markNeedsUpdate]s are called."

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+      // It is possible the updateChildren above caused some nodes to be added
+      // to _nodesNeedingSemanticsGeometryUpdate. Therefore, we need to
+      // process them here.
+      final List<RenderObject> nodesToProcessGeometry = _nodesNeedingSemanticsGeometryUpdate
+          .where(
+            (RenderObject object) =>
+                !object._needsLayout && object.owner == this && !object._semantics.parentDataDirty,
+          )
+          .toList();
+      _nodesNeedingSemanticsGeometryUpdate.clear();
```
justinmc, line 1515: Is it ok to clear the nodes in _nodesNeedingSemanticsGeometryUpdate that are not in nodesToProcessGeometry?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+          // This node merely presents its subtree in the mergeup, so we need to clear
+          // the geometry for all the semantics nodes in the mergeup.
+          for (final _RenderObjectSemantics child
+              in node._semantics.mergeUp.whereType<_RenderObjectSemantics>()) {
+            if (child.shouldFormSemanticsNode) {
+              child.geometry = null;
+            } else {
+              // Even though this node does not form a semantics node, it still
+              // represents a group of render objects in the subtree, where node that
+              // forms a semantics node is in its _children.
```
justinmc: "where node that forms a semantics node is in its children."

If I understand this correctly I think you should just add an "a":

=>

"where a node that forms a semantics node is in its children."

Basically you're just saying that this node is not a semantics node but one of its children is, right? Without the "a", it makes me think that "node" is referring back to "this node" from the start of the sentence.

packages/flutter/test/rendering/semantics_layout_sized_test.dart
```diff
@@ -0,0 +1,315 @@
...
+    // The parent won't use child's size, so child is a relayout boundary.
+    // The child also sets itself as a semantic boundary.
+    final child = RenderTestLayoutSemanticsBoundary();
+    child.isSemanticBoundary = true;
+    final parent = RenderTestParent(child: child);
+
+    TestRenderingFlutterBinding.instance.pipelineOwner.ensureSemantics();
+    layout(parent, phase: EnginePhase.flushSemantics);
+
+    // Verify initial state
```
justinmc: Nit: Periods at the ends of these comments.

justinmc (approved):
Renewing my LGTM. I don't understand the flushSemantics algorithm as well as @LongCatIsLooong so I'm still deferring to him for final approval.

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: +1 to this, also I feel the "All nodes in this set are guaranteed to be semantics boundaries." line should suffice, and the "closest" thing might not be most accurate since IIRC we do skip-boundary for sibling merge groups?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1417,8 +1417,25 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+  /// All nodes in this set are guaranteed to be semantics boundaries.
   final Set<RenderObject> _nodesNeedingSemantics = <RenderObject>{};
 
+  /// Nodes that need semantics geometry update.
+  ///
+  /// Nodes are added when their [_RenderObjectSemantics.markNeedsUpdate] are called.
+  ///
+  /// This is used to track nodes that need their semantics geometry updated.
+  /// This is different from [_nodesNeedingSemantics] because it only tracks
+  /// nodes that need their semantics geometry updated, not nodes that need
```
LongCatIsLooong: nit: when I initially read this (before I had a rough understanding of how this works), I was confused by this sentence because the set is literally populated in `RenderObjectSemantics.markNeedsUpdate` so that felt contradicting to "not nodes that need their semantics updated". My understanding is that `_nodesNeedingSemantics` contains dirty semantics boundaries and this set contains dirty nodes whose ancestors must update the geometry for them.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
-      for (final node in nodesToProcess) {
-        if (node._semantics.parentDataDirty) {
-          // same as above.
+      // It is possible the updateChildren above caused some nodes to be added
+      // to _nodesNeedingSemanticsGeometryUpdate. Therefore, we need to
+      // process them here.
+      final List<RenderObject> nodesToProcessGeometry = _nodesNeedingSemanticsGeometryUpdate
+          .where(
+            (RenderObject object) =>
+                !object._needsLayout && object.owner == this && !object._semantics.parentDataDirty,
```
LongCatIsLooong, line 1512: nit: should the `_needsLayout` check be an assert instead? If the node is attached and not blocked in the semantics tree then it must have been laid out at this point? Otherwise we won't be able to generate accurate geometry for the node (and its subtree).

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+              // forms a semantics node is in its _children.
+              for (final _RenderObjectSemantics nodeInSubtree in child._children) {
+                assert(nodeInSubtree.shouldFormSemanticsNode);
+                nodeInSubtree.geometry = null;
+              }
+            }
+          }
+          continue;
+        }
+        // This node won't form a semantics node, but it has children that form
```
LongCatIsLooong: I thought if we reach here `node._semantics.contributesToSemanticsTree` is going to be true (so this RO may form a semantics node).

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1503,12 +1590,41 @@ base class PipelineOwner with DiagnosticableTreeMixin {
         FlutterTimeline.startSync('Semantics.ensureSemanticsNode');
       }
       for (final RenderObject node in nodesToProcess.reversed) {
-        if (node._semantics.parentDataDirty) {
-          // same as above.
-          continue;
+        node._semantics.computeAncestorInfo(treeShapeToken);
+        final targets = <_RenderObjectSemantics>[];
```
LongCatIsLooong, line 1602: Why's the list needed? It looks like the elements are immediately popped after enqueuing for processing?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5447,13 +5568,78 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
...
+  /// Updated by [computeAncestorInfo] during [PipelineOwner.flushSemantics] to
+  /// prevent redundant ancestor lookups within a single flush cycle.
+  Object _currentTreeShapeToken = const Object();
+
+  /// The nearest semantics-forming ancestor node with clean geometry.
+  ///
+  /// Cached by [computeAncestorInfo] during [PipelineOwner.flushSemantics].
+  _RenderObjectSemantics? firstAncestorNodeWithCleanGeometry;
+
+  /// Finds and caches the nearest semantics-forming ancestor with clean geometry.
```
LongCatIsLooong: nit: clean -> up-to-date?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5447,13 +5568,78 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
...
+  ///
+  /// The [treeShapeToken] parameter prevents redundant calculation by verifying
```
LongCatIsLooong, line 5611: From the looks of it, a new token is created just before ensureGeometry. So this is the same as an `int generation` which increases every frame?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -5447,13 +5568,78 @@ class _RenderObjectSemantics extends _SemanticsFragment with DiagnosticableTreeM
...
+    }
+    _RenderObjectSemantics? next;
+    if (shouldFormSemanticsNode) {
+      if (!geometryDirty) {
+        firstAncestorNodeWithCleanGeometry = this;
+      }
+      next = parentInSemanticsTree;
+    } else {
+      next = this;
+      while (!next!.parentDataDirty && !next.shouldFormSemanticsNode) {
```
LongCatIsLooong, line 5639: The root node has to form a semantics node so `next` can't be null? Could you add an `assert(next != null)` after `next = next.parent`?

packages/flutter/lib/src/rendering/object.dart
chunhtai: yes, if nodes in _nodesNeedingSemanticsGeometryUpdate but not in nodesToProcessGeometry, it means it is not on the tree(because parentData is dirty or not laid out), and will be re-added to the list when it rejoin the tree

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+
+      // Clear geometry for nodes that needs geometry update.
+      for (final node in nodesToProcessGeometry) {
+        if (node._semantics.shouldFormSemanticsNode && node._semantics.geometryDirty) {
+          // This node is already dirty, skip it.
           continue;
         }
-        node._semantics.ensureGeometry();
+
+        if (node._semantics.shouldFormSemanticsNode && (node._isRelayoutBoundary ?? false)) {
```
LongCatIsLooong, line 1532: So it looks like the overall logic for marking ROs as needing semantics geometry update becomes:
1. if !contributesToSemanticsTree => mark all node-forming children (from the mergeUp group) dirty 
2. else if `_isRelayoutBoundary` and `shouldFormSemanticsNode` => mark `this` dirty
3. else => mark all children (all of them should be node-forming) dirty

Shouldn't we just mark `this` dirty if the RenderObject is node-forming, regardless of wether it is a relayout boundary (instead of trying to mark all of its children dirty), since we had to rerun `layout` on this RO so it's size may have changed?

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -1488,13 +1504,84 @@ base class PipelineOwner with DiagnosticableTreeMixin {
...
+      }
+
+      // Used to invalidate the [_RenderObjectSemantics.firstAncestorNodeWithCleanGeometry] cache.
+      //
+      // Multiple items in [nodesToProcessGeometry] often share the same first
+      // ancestor, making redundant searches expensive. Passing this token allows
+      // nodes to reuse cached ancestor lookups.
+      //
+      // See [_RenderObjectSemantics.computeAncestorInfo] for details.
+      final treeShapeToken = Object();
```
LongCatIsLooong, line 1574: instead of using a token here can we calculate all `firstAncestorNodeWithCleanGeometry` from `nodesToProcessGeometry` and sort it? The `firstAncestorNodeWithCleanGeometry` field is only good for this frame so the cache shouldn't live past this frame / this flushSemantics call.

LongCatIsLooong (commented):
Now that we maintain `parentInSemanticsTree` I feel this becoming similar to `markNeedsLayout` + `_needsLayout` on `RenderObject`. Currently we are still setting `_geometry = null` after `updateChildren` (i.e. after the tree hierarchy is updated). But now that we have access to `parentInSemanticsTree`, I think alternatively we can invalid `_geometry` based on the old tree hierarchy (old as in, using parentInSemanticsTree before it gets updated in updateChildren), like how `markNeedsLayout` invalidates the current layout of a RO, and skips adding a node to the set if its `parentInSemanticsTree` is already marked dirty. This is not ideal if a child node is marked dirty before its parent in which case both nodes will be added as root nodes to the set, but in most cases, especially layout and build, nodes are processed in depth order.

So this basically means to move `_geometry = null` to `ROSemantics.markNeedsUpdate` (also this has to happen if a node changes parent), and only add the node as a dirty geometry root if its parent is clean. As a result we will get a (hopefully) much smaller _nodesNeedingSemanticsGeometryUpdate set to sort and call `ensureGeometry` on. It seems we can further reduce the set to include only real dirty roots by removing those whose parentInSemanticsTree is also dirty after `updateChildren`?

packages/flutter/lib/src/rendering/object.dart
chunhtai: oh this comment was updated. will update

packages/flutter/lib/src/rendering/object.dart
chunhtai: is it possible that a renderobject may skip layout because they are kept alive somehow but not on screen?

packages/flutter/lib/src/rendering/object.dart
chunhtai: this can either be one  or two depends on if else case. This is just to unify the code path instead of calling ensureSemanticsNode in various places in the if else condition

packages/flutter/lib/src/rendering/object.dart
chunhtai: yeah, do you think a counter will be better? will we run into int overflow?

packages/flutter/lib/src/rendering/object.dart
chunhtai: root node's parentData is always clean and shouldFormSemanticsNode is always true, so it won't go into this while loop.

Also `next` can be null after next = next.parent, or do you meant `assert(!isRoot || next !=null)`

packages/flutter/lib/src/rendering/object.dart
chunhtai: yeah I think so, let me try to see if we run into corner case

packages/flutter/lib/src/rendering/object.dart
chunhtai: > can we calculate all firstAncestorNodeWithCleanGeometry from nodesToProcessGeometry and sort it

I think we still need some sort of cache to avoid duplicate calculation. for example
```
A
|
B
|  \
C   D
```
and C,D is in the set and A is the firstAncestorNodeWithCleanGeometry for both. we don't want to repeatedly calculate B and A

packages/flutter/lib/src/rendering/object.dart
chunhtai: so this is for performance optimization. if this node's geometry should change if not relayout boundary, parent should do the clean up

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: Ah you mean in a scroll view? So an offscreen cell in a scroll view is not considered blocked? And it's only dropped from the semantics tree after geometry calculation and deemed invisible?

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: > if this node's geometry should change if not relayout boundary, parent should do the clean up

I'm not sure I fully understand what you mean, could you elaborate / do you have an example? 

I feel conceptually `renderObject.layout` should only invalidate its own geometry if its semantics-node-forming:
- apparently it doesn't affect its parent, unless the parent cares about its size in which case the parent needs relayout (and that would invalidate the parent's own geometry)
- it may affect its children's paint transform, but `updateGeometry` should cover that since it always traverse its children (but stop when there's no paint transform update)?

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: Hmm I guess we don't have to do this in the first place if we stored dirty roots (similar to `markNeedsLayout`) instead of storing every dirty node. But in this case B, C, D are all dirty right? Does it work if we scan the list and drop everything with a dirty parent?

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: Ah ok sounds good (but you have to call `targets.add` instead of `ensureSemanticsNode` now?)

packages/flutter/lib/src/rendering/object.dart
LongCatIsLooong: I think I was looking for something that explains why the `!` was safe

LongCatIsLooong (commented):
mostly LGTM but I feel the clean ancestor tracking can be simplified.

packages/flutter/lib/src/rendering/object.dart
chunhtai: blocked node can be here as long as someone called markNeedsSemanticsUpdate on it. in this case, if it isn't layout at this point, it should be fine to ignore it.

packages/flutter/lib/src/rendering/object.dart
chunhtai: right.

packages/flutter/lib/src/rendering/object.dart
chunhtai: oh i see yeah, this should be guaranteed, will add the assert

packages/flutter/lib/src/rendering/object.dart
chunhtai: > I'm not sure I fully understand what you mean, could you elaborate / do you have an example?

If this node is not relayout boundary, calling this node's markneedsLayout will calls the parent's markNeedsLayout which will clear the geometry of this node.

I guess at the end it doesn't matter, but it seems to be slightly more systematic to let the parent do the job

> it may affect its children's paint transform, but updateGeometry should cover that since it always traverse its children (but stop when there's no paint transform update)?

yeah that is why it still clear out child node's geometry

packages/flutter/lib/src/rendering/object.dart
chunhtai: > Does it work if we scan the list and drop everything with a dirty parent?

but we need to walk the parent chain to figure this out we still have to cache something in each node along the chain

PR #180389, opened 2025-12-30
Allow TabBar to receive a TabBarScrollController

### LINKED ISSUE #123965: Expose the Scroll Controller of a scrollable TabBar

### LINKED ISSUE #124608: Wrapping a Scrollbar around a TabBar throws error `The Scrollbar's ScrollController has no ScrollPosition attached.`

navaronbracke:
This pull request makes `TabBarScrollController` public, and allows a `TabBar` widget to receive one.

Fixes https://github.com/flutter/flutter/issues/123965
Fixes https://github.com/flutter/flutter/issues/124608
Related to https://github.com/flutter/flutter/pull/180120

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -1663,6 +1663,17 @@ class _TabBarState extends State<TabBar> {
     }
   }
 
+  void _updateScrollController() {
+    if (widget.scrollController != null) {
+      widget.scrollController?._tabBarState = this;
+
+      return;
+    }
+
+    _internalScrollController ??= TabBarScrollController();
+    _internalScrollController?._tabBarState = this;
+  }
+
   void _initIndicatorPainter() {
     final ThemeData theme = Theme.of(context);
     final TabBarThemeData tabBarTheme = TabBarTheme.of(context);
@@ -1705,10 +1716,9 @@ class _TabBarState extends State<TabBar> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
+    _updateScrollController();
     _updateTabController();
     _initIndicatorPainter();
-
-    _effectiveScrollController._tabBarState = this;
   }
 
   @override
@@ -1716,31 +1726,9 @@ class _TabBarState extends State<TabBar> {
     super.didUpdateWidget(oldWidget);
     if (widget.controller != oldWidget.controller ||
         widget.scrollController != oldWidget.scrollController) {
+      _updateScrollController();
       _updateTabController();
       _initIndicatorPainter();
-
-      if (oldWidget.scrollController == null) {
-        // The old controller was null,
-        // meaning the internal scroll controller cannot be null.
-        // Dispose of the internal scroll controller.
-        assert(_internalScrollController != null);
-        assert(widget.scrollController != null);
-
-        _internalScrollController!.detach(_internalScrollController!.position);
-        _internalScrollController!.dispose();
-        _internalScrollController = null;
-      } else {
-        // The old controller was not null, detach.
-        oldWidget.scrollController!.detach(oldWidget.scrollController!.position);
-        if (widget.scrollController == null) {
-          // If the new controller is null,
-          // set up the internal TabBarScrollController.
-          _internalScrollController = TabBarScrollController().._tabBarState = this;
-        }
-      }
-
-      // Attach the updated effective scroll controller.
-      _effectiveScrollController.attach(_effectiveScrollController.position);
 
       // Adjust scroll position.
       if (_effectiveScrollController.hasClients) {
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -884,7 +884,7 @@ final class TabBarScrollController extends ScrollController {
   /// Is null if this controller is not attached to a [TabBar].
   _TabBarState? _tabBar;
 
-  set _tabBarState(_TabBarState tabBarState) {
+  set _tabBarState(_TabBarState? tabBarState) {
     _tabBar = tabBarState;
   }
 
@@ -906,6 +906,12 @@ final class TabBarScrollController extends ScrollController {
       oldPosition: oldPosition,
       tabBar: _tabBarState,
     );
+  }
+
+  @override
+  void dispose() {
+    _tabBarState = null;
+    super.dispose();
   }
 }
 
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -884,8 +884,14 @@ final class TabBarScrollController extends ScrollController {
   /// Is null if this controller is not attached to a [TabBar].
   _TabBarState? _tabBar;
 
-  void _attachToTabBar(_TabBarState tabBarState) {
+  set _tabBarState(_TabBarState tabBarState) {
     _tabBar = tabBarState;
+  }
+
+  _TabBarState get _tabBarState {
+    assert(_tabBar != null, 'TabBarScrollController is not attached to any TabBar.');
+
+    return _tabBar!;
   }
 
   @override
@@ -898,7 +904,7 @@ final class TabBarScrollController extends ScrollController {
       physics: physics,
       context: context,
       oldPosition: oldPosition,
-      tabBar: _tabBar!,
+      tabBar: _tabBarState,
     );
   }
 }
@@ -1701,6 +1707,8 @@ class _TabBarState extends State<TabBar> {
     super.didChangeDependencies();
     _updateTabController();
     _initIndicatorPainter();
+
+    _effectiveScrollController?._tabBarState = this;
   }
 
   @override
@@ -2116,7 +2124,7 @@ class _TabBarState extends State<TabBar> {
               start: _kStartOffset,
             ).add(widget.padding ?? EdgeInsets.zero)
           : widget.padding;
-      _effectiveScrollController?._attachToTabBar(this);
+
       tabBar = ScrollConfiguration(
         // The scrolling tabs should not show an overscroll indicator.
         behavior: ScrollConfiguration.of(context).copyWith(overscroll: false),
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -1096,7 +1096,7 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
   /// will be used.
   final TabController? controller;
 
-  /// The [ScrollController] for this [TabBar].
+  /// The [TabBarScrollController] for this [TabBar].
   ///
   /// This controller can be used to manipulate the scroll position of the [TabBar].
   final TabBarScrollController? scrollController;
@@ -1558,9 +1558,9 @@ class _TabBarState extends State<TabBar> {
     }
   }
 
-  TabBarScrollController? get _effectiveScrollController {
+  TabBarScrollController get _effectiveScrollController {
     if (widget.scrollController != null) {
-      return widget.scrollController;
+      return widget.scrollController!;
     }
 
     return _internalScrollController ??= TabBarScrollController();
@@ -1708,7 +1708,7 @@ class _TabBarState extends State<TabBar> {
     _updateTabController();
     _initIndicatorPainter();
 
-    _effectiveScrollController?._tabBarState = this;
+    _effectiveScrollController._tabBarState = this;
   }
 
   @override
@@ -1719,11 +1719,9 @@ class _TabBarState extends State<TabBar> {
       _updateTabController();
       _initIndicatorPainter();
 
-      final TabBarScrollController? scrollController = _effectiveScrollController;
-
       // Adjust scroll position.
-      if (scrollController != null && scrollController.hasClients) {
-        final ScrollPosition position = scrollController.position;
+      if (_effectiveScrollController.hasClients) {
+        final ScrollPosition position = _effectiveScrollController.position;
         if (position is _TabBarScrollPosition) {
           position.markNeedsPixelsCorrection();
         }
@@ -1781,12 +1779,8 @@ class _TabBarState extends State<TabBar> {
     return clampDouble(tabCenter + paddingStart - viewportWidth / 2.0, minExtent, maxExtent);
   }
 
-  double? _tabCenteredScrollOffset(int index) {
-    final ScrollPosition? position = _effectiveScrollController?.position;
-
-    if (position == null) {
-      return null;
-    }
+  double _tabCenteredScrollOffset(int index) {
+    final ScrollPosition position = _effectiveScrollController.position;
 
     return _tabScrollOffset(
       index,
@@ -1801,27 +1795,19 @@ class _TabBarState extends State<TabBar> {
   }
 
   void _scrollToCurrentIndex() {
-    final double? offset = _tabCenteredScrollOffset(_currentIndex!);
+    final double offset = _tabCenteredScrollOffset(_currentIndex!);
 
-    if (offset == null) {
-      return;
-    }
-
-    _effectiveScrollController?.animateTo(offset, duration: kTabScrollDuration, curve: Curves.ease);
+    _effectiveScrollController.animateTo(offset, duration: kTabScrollDuration, curve: Curves.ease);
   }
 
   void _scrollToControllerValue() {
     final double? leadingPosition = _currentIndex! > 0
         ? _tabCenteredScrollOffset(_currentIndex! - 1)
         : null;
-    final double? middlePosition = _tabCenteredScrollOffset(_currentIndex!);
+    final double middlePosition = _tabCenteredScrollOffset(_currentIndex!);
     final double? trailingPosition = _currentIndex! < maxTabIndex
         ? _tabCenteredScrollOffset(_currentIndex! + 1)
         : null;
-
-    if (middlePosition == null) {
-      return;
-    }
 
     final double index = _controller!.index.toDouble();
     final double value = _controller!.animation!.value;
@@ -1839,7 +1825,7 @@ class _TabBarState extends State<TabBar> {
             : lerpDouble(middlePosition, trailingPosition, value - index)!,
     };
 
-    _effectiveScrollController?.jumpTo(offset);
+    _effectiveScrollController.jumpTo(offset);
   }
 
   void _handleTabControllerAnimationTick() {
--- a/packages/flutter/lib/src/material/tabs.dart
+++ b/packages/flutter/lib/src/material/tabs.dart
@@ -1773,8 +1773,13 @@ class _TabBarState extends State<TabBar> {
     return clampDouble(tabCenter + paddingStart - viewportWidth / 2.0, minExtent, maxExtent);
   }
 
-  double _tabCenteredScrollOffset(int index) {
-    final ScrollPosition position = _scrollController!.position;
+  double? _tabCenteredScrollOffset(int index) {
+    final ScrollPosition? position = _effectiveScrollController?.position;
+
+    if (position == null) {
+      return null;
+    }
+
     return _tabScrollOffset(
       index,
       position.viewportDimension,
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1020,6 +1026,11 @@ class TabBar extends StatefulWidget implements PreferredSizeWidget {
   /// will be used.
   final TabController? controller;
 
+  /// The [ScrollController] for this [TabBar].
```
navaronbracke: Should this be explicitly `TabBarScrollController` ?

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1623,12 +1642,16 @@ class _TabBarState extends State<TabBar> {
   @override
   void didUpdateWidget(TabBar oldWidget) {
     super.didUpdateWidget(oldWidget);
-    if (widget.controller != oldWidget.controller) {
+    if (widget.controller != oldWidget.controller ||
+        widget.scrollController != oldWidget.scrollController) {
       _updateTabController();
       _initIndicatorPainter();
+
+      final TabBarScrollController? scrollController = _effectiveScrollController;
```
navaronbracke: I'm not sure if other things are required here.

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1686,8 +1709,13 @@ class _TabBarState extends State<TabBar> {
     return clampDouble(tabCenter + paddingStart - viewportWidth / 2.0, minExtent, maxExtent);
   }
 
-  double _tabCenteredScrollOffset(int index) {
-    final ScrollPosition position = _scrollController!.position;
+  double? _tabCenteredScrollOffset(int index) {
+    final ScrollPosition? position = _effectiveScrollController?.position;
```
navaronbracke: I was trying to clean up the `!` around the internal scroll controller, so I added returns in some places. (We could also check hasClients etc, but I don't think that adds any value here?)

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1701,19 +1729,28 @@ class _TabBarState extends State<TabBar> {
...
+
+    _effectiveScrollController?.animateTo(offset, duration: kTabScrollDuration, curve: Curves.ease);
   }
 
   void _scrollToControllerValue() {
     final double? leadingPosition = _currentIndex! > 0
         ? _tabCenteredScrollOffset(_currentIndex! - 1)
         : null;
-    final double middlePosition = _tabCenteredScrollOffset(_currentIndex!);
+    final double? middlePosition = _tabCenteredScrollOffset(_currentIndex!);
```
navaronbracke: Since everything relies on the middle position, I added a return here

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -2015,14 +2052,14 @@ class _TabBarState extends State<TabBar> {
               start: _kStartOffset,
             ).add(widget.padding ?? EdgeInsets.zero)
           : widget.padding;
-      _scrollController ??= _TabBarScrollController(this);
```
navaronbracke, line 2082: Both the old and new code path attach here, in build(), which feels very side-effect-y. Should we move this to `didChangeDependencies()` ?

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -823,12 +823,22 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
...
+  ///
+  /// Is null if this controller is not attached to a [TabBar].
+  _TabBarState? _tabBar;
 
-  final _TabBarState tabBar;
+  set _tabBarState(_TabBarState tabBarState) {
+    _tabBar = tabBarState;
+  }
+
+  _TabBarState get _tabBarState {
```
navaronbracke: Since a lint was complaining about using a setter + a missing getter once I did that, I added an assert in the getter, just in case.

huycozy (commented):
Thank you for submitting a separate PR! Left a comment for tests.

These approaches seem simpler than mine. However, I haven't yet thought of any other scenarios to test the risk of changing `_TabBarScrollController` to public. Let's see comments from others!

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,69 @@ void main() {
...
+            isScrollable: true,
+            controller: tabBarController,
+            scrollController: tabScrollController,
+            tabs: tabs,
+          ),
+        ),
+      ),
+    );
+
+    expect(tester.takeException(), isNull);
```
huycozy, line 2966: Can you also check a case when hovering over the tabbar to make the scrollbar visible? I added one [here](https://github.com/flutter/flutter/pull/180120/changes#diff-4df0ac240ada8b57bf4ca7a48a0c64f3fa8efefcfb4618abd72d7b51a8fe06a6R9606)

navaronbracke:
> Thank you for submitting a separate PR! Left a comment for tests.
> 
> These approaches seem simpler than mine. However, I haven't yet thought of any other scenarios to test the risk of changing `_TabBarScrollController` to public. Let's see comments from others!

The only real difference between `TabBarScrollController` and `ScrollController`, which the former inherits from, is that the `TabBarScrollController` creates its own, private scroll position. Since that needs to stay private, I marked `TabBarScrollController` as `final` in this PR, so that people cannot extend / implement TabBarScrollController, per the class modifiers reference: https://dart.dev/language/modifier-reference#valid-combinations

So I think that making it public, _with_ the class modifier included, is the right thing to do.

packages/flutter/test/material/tabs_test.dart
navaronbracke: Done

Piinks:
Since there are 2 proposals here, I recommend we centralize discussion here: https://github.com/flutter/flutter/issues/123965#issuecomment-3721109393 💙

packages/flutter/lib/src/material/tabs.dart
Piinks: Yes!

packages/flutter/lib/src/material/tabs.dart
Piinks: There is similar logic in ScrollableState for handling when there was a user-provided scroll controller, and then not, and vice versa

packages/flutter/lib/src/material/tabs.dart
Piinks: I would keep the !, and maybe add an error message? In what cases could it be null now?

packages/flutter/lib/src/material/tabs.dart
Piinks: Why would middle be null?

Piinks (commented):
Thanks for your patience on this! Much appreciated! 🙏

packages/flutter/lib/src/material/tabs.dart
navaronbracke: Oh, the `_effectiveScrollController` never actually returns null, my mistake!

packages/flutter/lib/src/material/tabs.dart
navaronbracke: This can be reverted

packages/flutter/lib/src/material/tabs.dart
navaronbracke: Fixed

packages/flutter/lib/src/material/tabs.dart
navaronbracke: I'll take a look!

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -877,12 +877,26 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
...
+    assert(_tabBar != null, 'TabBarScrollController is not attached to any TabBar.');
+
+    return _tabBar!;
+  }
+
+  /// Whether this controller is currently attached to a [TabBar]'s [State].
+  @visibleForTesting
+  bool get debugHasTabBarState => _tabBar != null;
```
navaronbracke: @Piinks Should this go in an `assert()` ? I used it for a test (see below)

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -894,9 +908,15 @@ class _TabBarScrollController extends ScrollController {
...
       context: context,
       oldPosition: oldPosition,
-      tabBar: tabBar,
+      tabBar: _tabBarState,
     );
   }
+
+  @override
+  void dispose() {
+    _tabBarState = null;
```
navaronbracke, line 916: Just in case we forget to remove the old state reference

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,203 @@ void main() {
...
+
+    // Trigger didUpdateWidget for the change to the internal controller.
+    await tester.pumpWidget(
+      boilerplate(
+        child: TabBar(isScrollable: true, controller: tabBarController, tabs: tabs),
+      ),
+    );
+
+    // Creating the new scroll position should not throw.
+    expect(tester.takeException(), isNull);
```
navaronbracke, line 2809: Here, I can verify that the internal controller is attached to the state, by checking if I can get the pixels, otherwise we cannot create the _TabBarScrollPosition.

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,203 @@ void main() {
...
+    final ScrollableState scrollableState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+    // The pixels of the new position should be the same as before the update.
+    expect(scrollableState.position.pixels, oldPixels);
+    expect(tabScrollController.debugHasTabBarState, isTrue);
```
navaronbracke: While I could verify that the internal controller is attached in the test above, I'm not sure how to test that the internal controller is _not_ attached to the TabBarState in this test. The idea is that a given `TabBarState` is only attached to one `TabBarScrollController`.

Do I need to verify this? Or maybe not, since it is the internal controller.

If I _do_ need to verify it, do you have ideas on how to expose the `debugHasTabBarState` for the internal controller from the `TabBar` widget? (The `_TabBarState` is private)

packages/flutter/lib/src/material/tabs.dart
Piinks: Yeah that sounds good to me. 👍

packages/flutter/test/material/tabs_test.dart
Piinks: Gotcha 👍

packages/flutter/test/material/tabs_test.dart
Piinks: I don't think that needs to be verified, there is already a lot of test coverage around the internal controller I think. :)

packages/flutter/lib/src/material/tabs.dart
navaronbracke: Done

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -877,12 +877,37 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
...
+
+    return _tabBar!;
+  }
+
+  /// Asserts that this controller is currently attached to a [TabBar]'s [State].
+  ///
+  /// To invoke this function, use the following pattern:
+  ///
+  /// ```dart
+  /// assert(debugCheckHasTabBarState());
```
Piinks: This is failing analysis, I _think_ there is a pattern that will fix this, see comments with `Examples can assume:` at the top of the file. I think that would do it. We're trying to move away from `{@tool snippet}`, and the other `@tool` directives seem like overkill here.
Another alternative could be to use `assert(debugCheckHasTabBarState());` (just wrapping in backticks).

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -877,12 +877,33 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
...
+    assert(debugCheckHasTabBarState());
+
+    return _tabBar!;
+  }
+
+  /// Asserts that this controller is currently attached to a [TabBar]'s [State].
+  ///
+  /// To invoke this function, wrap it in an assert: `assert(debugCheckHasTabBarState());`
+  ///
+  /// Does nothing if asserts are disabled. Always returns true.
```
Piinks: Why does it always return true?

Piinks:
Whoops, analyzer fix inbound.

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -877,12 +877,33 @@ class _TabBarScrollPosition extends ScrollPositionWithSingleContext {
...
-// This class, and TabBarScrollPosition, only exist to handle the case
-// where a scrollable TabBar has a non-zero initialIndex.
-class _TabBarScrollController extends ScrollController {
-  _TabBarScrollController(this.tabBar);
+/// The [ScrollController] for a [TabBar] widget.
+final class TabBarScrollController extends ScrollController {
+  /// The state of the [TabBar] widget to which this controller is attached.
+  ///
+  /// Is null if this controller is not attached to a [TabBar].
+  _TabBarState? _tabBar;
```
dkwingsmt: Why do we need a setter/getter pair here? It seems to me equivalent to using `_tabBar` directly and adding the assertion to within `createScrollPosition`.

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1541,6 +1575,14 @@ class _TabBarState extends State<TabBar> {
     }
   }
 
+  TabBarScrollController get _effectiveScrollController {
+    if (widget.scrollController != null) {
+      return widget.scrollController!;
```
dkwingsmt, line 1575: Dispose `_internalScrollController` before the return here?

packages/flutter/lib/src/material/tabs.dart
```diff
@@ -1680,19 +1735,23 @@ class _TabBarState extends State<TabBar> {
...
     _initIndicatorPainter();
   }
 
   @override
   void didUpdateWidget(TabBar oldWidget) {
     super.didUpdateWidget(oldWidget);
-    if (widget.controller != oldWidget.controller) {
+    if (widget.controller != oldWidget.controller ||
+        widget.scrollController != oldWidget.scrollController) {
+      _updateScrollController(oldScrollController: oldWidget.scrollController);
```
dkwingsmt, line 1743: I think all 3 `scrollController` in the two lines should be `_effectiveScrollController`.

packages/flutter/lib/src/material/tabs.dart
Piinks: From this comment- https://github.com/flutter/flutter/pull/180389#discussion_r2653318316 
It is enforced by a lint

dkwingsmt: Yeah but what about removing both the getter and the setter? The only thing that the setter brings is the assertion, which can be added to `createScrollPosition`.

navaronbracke: True, will adjust it (we can do the assert where needed)

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,273 @@ void main() {
...
+    // The pixels of the new position should be the same as before the update.
+    expect(scrollableState.position.pixels, oldPixels);
+
+    // This should not throw, since the tab bar is attached to the external controller.
+    expect(tabScrollController.debugCheckHasTabBarState(), isTrue);
+    expect(tester.takeException(), isNull);
+  });
+
+  testWidgets('TabBar correctly detaches old external TabBarScrollController when switched to a new one', (WidgetTester tester) async {
+    final List<Tab> tabs = <Tab>[
```
Piinks:
```suggestion
    final tabs = <Tab>[
```

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,273 @@ void main() {
...
+      for (int i = 0; i < 10; i++) Tab(text: 'Tab $i'),
+    ];
+
+    final TabBarScrollController controllerA = TabBarScrollController();
+    final TabBarScrollController controllerB = TabBarScrollController();
```
Piinks:
```suggestion
    final controllerA = TabBarScrollController();
    final controllerB = TabBarScrollController();
```

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,273 @@ void main() {
...
+        ),
+      ),
+    );
+
+    expect(controllerB.debugCheckHasTabBarState(), isTrue);
+    expect(() => controllerA.debugCheckHasTabBarState(), throwsAssertionError);
+  });
+
+  testWidgets('TabBar correctly detaches external TabBarScrollController when disposed', (WidgetTester tester) async {
+    final List<Tab> tabs = <Tab>[
```
Piinks:
```suggestion
    final tabs = <Tab>[
```

packages/flutter/test/material/tabs_test.dart
```diff
@@ -2737,6 +2737,273 @@ void main() {
...
+      for (int i = 0; i < 10; i++) Tab(text: 'Tab $i'),
+    ];
+
+    final TabBarScrollController controller = TabBarScrollController();
```
Piinks:
```suggestion
    final controller = TabBarScrollController();
```

packages/flutter/lib/src/material/tabs.dart
navaronbracke: Done

packages/flutter/lib/src/material/tabs.dart
navaronbracke: _effectiveScrollController is private to the state, though?

Piinks: Yeah, @navaronbracke is right I think. This is because _effectiveScrollController is a private getter within the _TabBarState class and is not a property of the TabBar widget itself. Since oldWidget is an instance of TabBar, it does not have an _effectiveScrollController property, so using _effectiveScrollController in that way is not possible.

dkwingsmt (approved):
LGTM

dkwingsmt:
Thank you for the investigation! Let me confirm if @Piinks has more comments before landing it.

Piinks (approved):
LGTM! Woo! Let's finally get this in! 🎊

Piinks:
All good! As a member of flutter-hackers, you are welcome to apply labels as well now too! :)

PR #180411, opened 2025-12-31
Add scrollCacheExtent to PageView

### LINKED ISSUE #45632: Allow PageView to set a cacheExtent
b/260361075

It would be nice to have a `PageView` that has a `cacheExtent` specified, so that pages could be pre-built.

This is currently blocked because  the iOS accessibility bridge does not look for an `allowImplicitScrolling` flag on the parent scrollable - it only looks to see if there are additional scroll children in the tree.  Setting a cache extent would then automatically enable implicit scrolling for PageViews in Voiceover mode, which we do not want.  If we can figure out a good way to address the iOS bridge issue, it should be fairly trivial to expose a cacheExtent parameter on the PageView.

/cc @lukepighetti who has done some work on this on the framework side.

Unfortunately, the iOS bug is not trivial to fix.

koji-1009:
fix https://github.com/flutter/flutter/issues/45632

Using the Xcode developer tools, I confirmed that even when `scrollCacheExtent` is set, `allowImplicitScrolling` is always respected.

https://github.com/user-attachments/assets/4f99875d-05f9-4224-a203-5b1d9274f7c9

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      home: const PageViewDemo(),
    );
  }
}

class PageViewDemo extends StatefulWidget {
  const PageViewDemo({super.key});

  @override
  State<PageViewDemo> createState() => _PageViewDemoState();
}

class _PageViewDemoState extends State<PageViewDemo> {
  int _currentIndex = 0;

  final List<
    ({
      String title,
      Axis axis,
      bool allowImplicitScrolling,
      ScrollCacheExtent scrollCacheExtent,
      IconData icon,
      String navLabel,
    })
  >
  _scenarios = [
    (
      title: 'Horizontal (allowImplicitScrolling: false)',
      axis: Axis.horizontal,
      allowImplicitScrolling: false,
      scrollCacheExtent: ScrollCacheExtent.viewport(0),
      icon: Icons.swipe_right,
      navLabel: 'H-False',
    ),
    (
      title: 'Horizontal (allowImplicitScrolling: true)',
      axis: Axis.horizontal,
      allowImplicitScrolling: true,
      scrollCacheExtent: ScrollCacheExtent.viewport(5),
      icon: Icons.swipe_right_alt,
      navLabel: 'H-True',
    ),
    (
      title: 'Vertical (allowImplicitScrolling: false)',
      axis: Axis.vertical,
      allowImplicitScrolling: false,
      scrollCacheExtent: ScrollCacheExtent.viewport(0),
      icon: Icons.swipe_down,
      navLabel: 'V-False',
    ),
    (
      title: 'Vertical (allowImplicitScrolling: true)',
      axis: Axis.vertical,
      allowImplicitScrolling: true,
      scrollCacheExtent: ScrollCacheExtent.viewport(5),
      icon: Icons.swipe_vertical,
      navLabel: 'V-True',
    ),
  ];

  @override
    final scenario = _scenarios[_currentIndex];

    return Scaffold(
      appBar: AppBar(title: Text(scenario.title)),
      body: PageView.builder(
        key: ValueKey('$_currentIndex'),
        scrollDirection: scenario.axis,
        allowImplicitScrolling: scenario.allowImplicitScrolling,
        scrollCacheExtent: scenario.scrollCacheExtent,
        itemCount: 10,
        itemBuilder: (context, index) => Center(
          child: Column(
            mainAxisAlignment: .center,
            children: [
              Text(
                'Page ${index + 1}',
                style: Theme.of(context).textTheme.displayLarge,
              ),
              const SizedBox(height: 20),
              ElevatedButton(
                onPressed: () {},
                child: Text('Button ${index + 1}'),
              ),
            ],
          ),
        ),
      ),
      bottomNavigationBar: NavigationBar(
        onDestinationSelected: (index) => setState(() => _currentIndex = index),
        selectedIndex: _currentIndex,
        destinations: _scenarios
            .map(
              (s) =>
                  NavigationDestination(icon: Icon(s.icon), label: s.navLabel),
            )
            .toList(),
      ),
    );
  }
}
```

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -691,12 +691,18 @@ class PageView extends StatefulWidget {
...
     this.allowImplicitScrolling = false,
+    this.cacheExtent,
     this.restorationId,
     this.clipBehavior = Clip.hardEdge,
     this.hitTestBehavior = HitTestBehavior.opaque,
     this.scrollBehavior,
     this.padEnds = true,
-  }) : childrenDelegate = SliverChildListDelegate(children);
+  }) : assert(
+         !allowImplicitScrolling || (cacheExtent ?? 1.0) > 0.0,
```
Piinks: Could this assertion be a breaking change for folks who currently have allowImplicitScrolling true?

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -736,12 +742,18 @@ class PageView extends StatefulWidget {
...
     this.dragStartBehavior = DragStartBehavior.start,
     this.allowImplicitScrolling = false,
+    this.cacheExtent,
     this.restorationId,
     this.clipBehavior = Clip.hardEdge,
     this.hitTestBehavior = HitTestBehavior.opaque,
     this.scrollBehavior,
     this.padEnds = true,
-  }) : childrenDelegate = SliverChildBuilderDelegate(
+  }) : assert(
```
Piinks, line 754: Same here

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -769,12 +781,17 @@ class PageView extends StatefulWidget {
...
     this.dragStartBehavior = DragStartBehavior.start,
     this.allowImplicitScrolling = false,
+    this.cacheExtent,
     this.restorationId,
     this.clipBehavior = Clip.hardEdge,
     this.hitTestBehavior = HitTestBehavior.opaque,
     this.scrollBehavior,
     this.padEnds = true,
-  });
+  }) : assert(
```
Piinks, line 796: ++

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -789,6 +806,17 @@ class PageView extends StatefulWidget {
...
+  /// {@macro flutter.rendering.RenderViewportBase.cacheExtent}
+  ///
+  /// In [PageView], the [cacheExtent] is treated as the number of pages, not
+  /// pixels, because the [Viewport] configuration uses [CacheExtentStyle.viewport].
+  ///
+  /// If null, and [allowImplicitScrolling] is true, the [cacheExtent] defaults
+  /// to 1.0, which means one additional page before and after the visible page
+  /// is preserved in the cache. If [allowImplicitScrolling] is false, the
+  /// [cacheExtent] defaults to 0.0.
+  final double? cacheExtent;
```
Piinks: If this is referring to pages, why is it a double? What would a cacheExtent of 1.2 mean?

packages/flutter/test/widgets/page_view_test.dart
```diff
@@ -1661,4 +1661,131 @@ void main() {
     controller.animateToPage(1, duration: const Duration(milliseconds: 50), curve: Curves.bounceIn);
```
Piinks, line 1660: Can you add tests for showOnScreen to make sure this change does enable that? 
Can you also check that the scroll related semantics set by _ScrollSemantics are correct?

Piinks (commented):
I dug into this a bit, and it looks like PageView has always been a bit of a special case here. Typically, the scroll physics and scroll position are where allowImplicitScrolling comes from. PageView does have it's own private scroll position subclass, and physics, but I wonder if _ForceImplicitScrollPhysics is needed now with this update?

packages/flutter/lib/src/widgets/page_view.dart
koji-1009: The default value for `cacheExtent` is `null`, so this assertion evaluates to `(cacheExtent ?? 1.0) > 0.0`, which is `true`. Therefore, this change does not affect users who have already set `allowImplicitScrolling` to `true`.

packages/flutter/lib/src/widgets/page_view.dart
koji-1009: https://api.flutter.dev/flutter/widgets/Viewport/cacheExtentStyle.html

Since `cacheExtentStyle` is set to `CacheExtentStyle.viewport`, the scale applies to the viewport. Therefore, just as the previous 1.0 was called 1.0 page, 1.2 becomes 1.2 pages.

packages/flutter/test/widgets/page_view_test.dart
koji-1009: Thanks for the review. I'll add the test.

packages/flutter/lib/src/widgets/page_view.dart
Piinks: Looking at the source issue, how do we pre-build 1.2 pages?

koji-1009: The `cacheExtent` parameter is passed directly to [Viewport](https://api.flutter.dev/flutter/widgets/Viewport-class.html), which handles the pixel calculation using its existing mechanism. This PR simply exposes that existing parameter to `PageView`.

With `CacheExtentStyle.viewport`, a value of `1.2` means "cache widgets within 1.2 viewport-widths beyond the visible area." The [RenderViewport](https://github.com/flutter/flutter/blob/3.38.7/packages/flutter/lib/src/rendering/viewport.dart#L1643-L1646) determines which children fall within that range and pre-builds them.

The `double` type is consistent with:
- `Viewport.cacheExtent` (existing API)
- `PageController.page` (returns `double`, e.g., `0.5` when halfway between pages)

Piinks: Gotcha. Thanks for walking me through the change, as well as your patience. I was away unexpectedly.

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -789,6 +806,17 @@ class PageView extends StatefulWidget {
   /// will traverse to the next page in the page view.
   final bool allowImplicitScrolling;
 
+  /// {@macro flutter.rendering.RenderViewportBase.cacheExtent}
```
Piinks: I think there is some important documentation to add here regarding cacheExtent being "treated as the number of pages."

I think this is only strictly true if viewportFraction is 1.0. Since PageView uses CacheExtentStyle.viewport, the value actually represents "number of viewport lengths." If a user has viewportFraction: 0.5 (two pages per viewport), then cacheExtent: 1.0 will actually cache two additional pages, is that right?

packages/flutter/test/widgets/page_view_test.dart
```diff
@@ -1661,4 +1661,250 @@ void main() {
...
+      ),
+    );
+
+    expect(semantics, includesNodeWith(label: 'Page 1'));
+    expect(semantics, includesNodeWith(label: 'Page 2'));
+
+    semantics.dispose();
+  });
+
+  testWidgets('PageView cacheExtent does not affect ïhasImplicitScrolling semantics flag', (
```
Piinks:
```suggestion
  testWidgets('PageView cacheExtent does not affect hasImplicitScrolling semantics flag', (
```

Piinks (commented):
Thanks again for the detailed explanations. I am guessing the merge conflict is coming from https://github.com/flutter/flutter/pull/181092, which deprecates all cacheExtent and cacheExtentStyle to ScrollCacheExtent.

Would you like to update this PR to use ScrollCacheExtent?

packages/flutter/test/widgets/page_view_test.dart
koji-1009: Thanks for the review! I merged into the main branch and fixed it in the next commit. [f2dbc9c](https://github.com/flutter/flutter/pull/180411/commits/f2dbc9cb1d607452f323cb276f8392813852153f)

packages/flutter/lib/src/widgets/page_view.dart
koji-1009: I've updated the PR to use `ScrollCacheExtent` and updated the dartdoc accordingly.
It seems that using `ScrollCacheExtent` resolves the ambiguity regarding whether the value represents pages or pixels.
Could you take a look? Thanks! [8a47c54](https://github.com/flutter/flutter/pull/180411/commits/8a47c544fd9e84a7f862524518e3c59c81977a13)

Piinks: Oh man, thank you so much for updating this to the new API!

Piinks (approved):
This LGTM, thank you again! cc @chunhtai for a second review, who just introduced the ScrollCacheExtent API

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -691,12 +691,18 @@ class PageView extends StatefulWidget {
...
     this.allowImplicitScrolling = false,
+    this.scrollCacheExtent,
     this.restorationId,
     this.clipBehavior = Clip.hardEdge,
     this.hitTestBehavior = HitTestBehavior.opaque,
     this.scrollBehavior,
     this.padEnds = true,
-  }) : childrenDelegate = SliverChildListDelegate(children);
+  }) : assert(
+         !allowImplicitScrolling || (scrollCacheExtent?.value ?? 1.0) > 0.0,
```
chunhtai: consider give it it default in either in initializer or in constructor body, and make it non null

packages/flutter/lib/src/rendering/sliver_fill.dart
```diff
@@ -53,6 +58,43 @@ class RenderSliverFillViewport extends RenderSliverFixedExtentBoxAdaptor {
     _viewportFraction = value;
     markNeedsLayout();
   }
+
+  /// Whether the user can traverse to the hidden pages.
```
chunhtai: should mention this is for assistive technologies such as screen readers

packages/flutter/lib/src/rendering/sliver_fill.dart
```diff
@@ -53,6 +58,43 @@ class RenderSliverFillViewport extends RenderSliverFixedExtentBoxAdaptor {
...
+  ///
+  /// If set to false, the hidden pages are excluded from the semantics tree.
```
chunhtai: This sentence is implementation detail.

packages/flutter/lib/src/rendering/sliver_fill.dart
chunhtai: also consider use `{@template` to copy some doc from PageView.allowImplicitScrolling

packages/flutter/lib/src/widgets/sliver_fill.dart
```diff
@@ -64,12 +65,26 @@ class SliverFillViewport extends StatelessWidget {
   /// {@macro flutter.widgets.SliverMultiBoxAdaptorWidget.delegate}
   final SliverChildDelegate delegate;
 
+  /// Controls whether the widget's children will respond to
```
chunhtai: use `{@template` to dedup from the PageView's doc

packages/flutter/lib/src/rendering/sliver_fill.dart
```diff
@@ -53,6 +58,43 @@ class RenderSliverFillViewport extends RenderSliverFixedExtentBoxAdaptor {
     _viewportFraction = value;
     markNeedsLayout();
   }
+
+  /// {@macro flutter.widgets.PageView.allowImplicitScrolling}
+  ///
+  /// This is typically used by assistive technologies, such as screen readers.
+  bool get allowImplicitScrolling => _allowImplicitScrolling;
+  bool _allowImplicitScrolling;
+  set allowImplicitScrolling(bool value) {
```
chunhtai, line 67: I am curious why we need allowImplicitScrolling if we can control this by setting cacheextent? it feels to me allowImplicitScrolling is not too useful in the first place.

packages/flutter/lib/src/widgets/page_view.dart
```diff
@@ -787,8 +811,32 @@ class PageView extends StatefulWidget {
   /// With this flag set to true, when accessibility focus reaches the end of
   /// the current page and user attempts to move it to the next element, focus
   /// will traverse to the next page in the page view.
+  /// {@endtemplate}
```
chunhtai, line 817: if this is false, does the ScrollCacheExtent matter anymore?

packages/flutter/lib/src/rendering/sliver_fill.dart
koji-1009: Thank you for the review.

I'd like to share some points about the relationship between `allowImplicitScrolling` and `scrollCacheExtent`.

To unify them in PageView, we would need to enforce: if `allowImplicitScrolling` is `false`, `scrollCacheExtent` must be 0. And if `true`, it must be greater than 0. 
Also, `allowImplicitScrolling` affects not only semantics filtering in `RenderSliverFillViewport.visitChildrenForSemantics`, but also `showOnScreen` behavior through `_ForceImplicitScrollPhysics`.
So `scrollCacheExtent` alone cannot replace `allowImplicitScrolling`.

This PR was focused on exposing `scrollCacheExtent` to PageView.
Do you think we should also change the behavior so that `allowImplicitScrolling` is determined by `scrollCacheExtent` value?

chunhtai: I mostly worry about cases allowImplicitScrolling = false but scrollCacheExtent >0.

Or basically just why would one want to disable allowImplicitScrolling in the first place.

It is probably too late to remove it i n PageView, but I think we could let allowImplicitScrolling to determine by scrollCacheExtent here. WDYT?

koji-1009: @chunhtai 
Thank you for the comment.

I think we have two options here:

1. **Assert only**
   - Add assertion to prevent `allowImplicitScrolling = false` with `scrollCacheExtent.value > 0`.
   - But assertions are stripped in release builds, so this does not guarantee consistency in production.
   - The current implementation is close to this approach.
2. **Runtime derivation**
   - Ignore `allowImplicitScrolling` parameter and derive it from `scrollCacheExtent.value > 0` at runtime.
   - This is safe in release builds, but `scrollCacheExtent` would take priority over `allowImplicitScrolling` — setting `scrollCacheExtent > 0` would enable implicit scrolling regardless of `allowImplicitScrolling
 = false`.

Which approach do you prefer? If you think option 2 is the right direction, I can update the implementation.

chunhtai: Thinking about it a bit more, I think adding assert should probably be fine, let's go for option 1

koji-1009: @chunhtai 
Thank you for the clarification. I unified the assertions into a single one that enforces consistency between `scrollCacheExtent` and `allowImplicitScrolling` in both directions:

```dart
assert(
  scrollCacheExtent == null || (scrollCacheExtent.value > 0.0) == allowImplicitScrolling,
);
```

Updated in [74f979a](https://github.com/flutter/flutter/pull/180411/commits/74f979a4454d8b54f91cffa62b37e8a9973ab12c), [8d42d7b](https://github.com/flutter/flutter/pull/180411/commits/8d42d7bbbd22d7912f45efcd487d68b3fbd1836d) and [8031d8a](https://github.com/flutter/flutter/pull/180411/commits/8031d8abeee45ac9b98ff834313e50a5ac7ddb71).

chunhtai (commented):
LGTM, but looks like there is some foramt issue

Found 1 Dart file which was formatted incorrectly.
To fix, run `dart format packages/flutter/test/widgets/page_view_test.dart`

chunhtai (approved):
LGTM

Piinks (approved):
LGTM as well!

PR #158040, opened 2024-11-02
Scroll Stepper horizontally

victorsanni:
Fixes [Horizontal Stepper Overflows Title (with no way to allow for scrolling)](https://github.com/flutter/flutter/issues/40601) 

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #180431, opened 2026-01-02
40601 fix: Overflowing stepper

### LINKED ISSUE #40601: Horizontal Stepper Overflows Title (with no way to allow for scrolling)
## Steps to Reproduce

1. Create a stepper with 5+ steps with any title such as "Step Title"
2. Make its direction horizontal
3. The Stepper widget puts step icons into a row until it overflows the screen, with no way to make it scroll

```
Stepper(
            type: StepperType.horizontal,
            currentStep: currentStep,
            steps: <Step>[
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 0),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 1),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 2),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 3),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 4),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 5),
              Step(
                  content: Text('content'),
                  title: Text('title'),
                  isActive: currentStep == 6),
            ],
          )
```

Wrapping the stepper with a horizontal `SingleChildScrollView` does not fix the issue.

`Stepper._buildHorizontal`

```
        Material(
          elevation: 2.0,
          child: Container(
            margin: const EdgeInsets.symmetric(horizontal: 24.0),
            child: Row(
              children: children,
            ),
          ),
        ),
```

As you can see from the source code which builds the title row with steps and titles, there is no capability to scroll over overflow gracefully at all. If the `children` of the `row` don't fit on the screen, it will overflow.

Parryword:
This PR adds SingleChildScrollView with LayoutBuilder to Stepper to fix overflowing issue.

Fixes #40601 "Horizontal Stepper Overflows Title (with no way to allow for scrolling)"

We should discuss how to control the ScrollView as the widget's controller is assigned to ListView instead. In my view, it should be reassigned to ScrollView.

The test stepper_test.dart passes. Overall: 18643 tests passed, 160 tests skipped, 184 failed. I expect failures are unlikely to be linked to my changes. Let me know if I may be running tests with a wrong configuration.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/material/stepper.dart
+++ b/packages/flutter/lib/src/material/stepper.dart
@@ -914,13 +914,25 @@ class _StepperState extends State<Stepper> with TickerProviderStateMixin {
       children: <Widget>[
         Material(
           elevation: widget.elevation ?? 2,
-          child: Padding(
-            padding: const EdgeInsets.symmetric(horizontal: 24.0),
-            child: SizedBox(
-              height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
-              child: Row(children: children),
-            ),
-          ),
+          child: LayoutBuilder(builder: (context, constraints) {
+            return SingleChildScrollView(
+              scrollDirection: Axis.horizontal,
+              child: ConstrainedBox(
+                constraints: BoxConstraints(
+                  minWidth: constraints.maxWidth
+                ),
+                child: Padding(
+                  padding: const EdgeInsets.symmetric(horizontal: 24.0),
+                  child: SizedBox(
+                    height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
+                    child: IntrinsicWidth(
+                      child: Row(children: children)
+                      ),
+                  ),
+                ),
+              ),
+            );
+          })
         ),
         Expanded(
           child: ListView(
--- a/packages/flutter/lib/src/material/stepper.dart
+++ b/packages/flutter/lib/src/material/stepper.dart
@@ -914,25 +914,23 @@ class _StepperState extends State<Stepper> with TickerProviderStateMixin {
       children: <Widget>[
         Material(
           elevation: widget.elevation ?? 2,
-          child: LayoutBuilder(builder: (context, constraints) {
-            return SingleChildScrollView(
-              scrollDirection: Axis.horizontal,
-              child: ConstrainedBox(
-                constraints: BoxConstraints(
-                  minWidth: constraints.maxWidth
-                ),
-                child: Padding(
-                  padding: const EdgeInsets.symmetric(horizontal: 24.0),
-                  child: SizedBox(
-                    height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
-                    child: IntrinsicWidth(
-                      child: Row(children: children)
-                      ),
+          child: LayoutBuilder(
+            builder: (context, constraints) {
+              return SingleChildScrollView(
+                scrollDirection: Axis.horizontal,
+                child: ConstrainedBox(
+                  constraints: BoxConstraints(minWidth: constraints.maxWidth),
+                  child: Padding(
+                    padding: const EdgeInsets.symmetric(horizontal: 24.0),
+                    child: SizedBox(
+                      height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
+                      child: IntrinsicWidth(child: Row(children: children)),
+                    ),
                   ),
                 ),
-              ),
-            );
-          })
+              );
+            },
+          ),
         ),
         Expanded(
           child: ListView(
--- a/packages/flutter/test/material/stepper_test.dart
+++ b/packages/flutter/test/material/stepper_test.dart
@@ -1251,6 +1251,41 @@ void main() {
     expect(bodyMediumStyle, nextLabelTextWidget.style);
   });
 
+  testWidgets('Stepper horizontal size test with list of steps widgets', (
+    WidgetTester tester,
+  ) async {
+    tester.view.physicalSize = const Size(400, 600);
+    tester.view.devicePixelRatio = 1.0;
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Scaffold(
+          body: Stepper(
+            type: StepperType.horizontal,
+            steps: List<Step>.generate(
+              10,
+              (i) => Step(
+                title: Text('Step ${i + 1}'),
+                content: const SizedBox(width: 100.0, height: 100.0),
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+
+    // 2. Check for the overflow exception
+    // If an overflow occurs, Flutter captures it in the summary
+    final dynamic exception = tester.takeException();
+
+    // Use this to fail the test if any overflow is detected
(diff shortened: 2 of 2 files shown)


packages/flutter/test/material/stepper_test.dart
```diff
@@ -1251,6 +1251,41 @@ void main() {
     expect(bodyMediumStyle, nextLabelTextWidget.style);
   });
 
+  testWidgets('Stepper horizontal size test with list of steps widgets', (
+    WidgetTester tester,
+  ) async {
+    tester.view.physicalSize = const Size(400, 600);
```
navaronbracke: Setting the size / DPI of the TestFlutterView is not needed. You can just wrap the widget under test in a SizedBox with a specific size.

packages/flutter/test/material/stepper_test.dart
```diff
@@ -1251,6 +1251,41 @@ void main() {
...
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+
+    // 2. Check for the overflow exception
+    // If an overflow occurs, Flutter captures it in the summary
+    final dynamic exception = tester.takeException();
```
navaronbracke: You should avoid dynamic. Instead, just inline the statement. Also, the comment above is redundant, so that can be removed.

```dart
expect(tester.takeException(), isNull);
```

packages/flutter/test/material/stepper_test.dart
```diff
@@ -1251,6 +1251,41 @@ void main() {
...
+
+    // Use this to fail the test if any overflow is detected
+    expect(exception, isNull, reason: 'Stepper should handle many steps without overflowing');
+
+    // 3. Reset the view size for other tests
```
navaronbracke: This is not needed

packages/flutter/lib/src/material/stepper.dart
```diff
@@ -914,12 +914,22 @@ class _StepperState extends State<Stepper> with TickerProviderStateMixin {
...
-            child: SizedBox(
-              height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
-              child: Row(children: children),
-            ),
+          child: LayoutBuilder(
+            builder: (context, constraints) {
+              return SingleChildScrollView(
+                scrollDirection: Axis.horizontal,
+                child: ConstrainedBox(
+                  constraints: BoxConstraints(minWidth: constraints.maxWidth),
```
navaronbracke, line 922: A scrollview should have a maximum constraint in the scroll axis, so this `BoxConstraints` should define `maxWidth` instead of minWidth?

packages/flutter/test/material/stepper_test.dart
```diff
@@ -1251,6 +1251,41 @@ void main() {
     expect(bodyMediumStyle, nextLabelTextWidget.style);
   });
 
+  testWidgets('Stepper horizontal size test with list of steps widgets', (
```
navaronbracke: Since this is a regression test for a bug, this should be indicated per the style guide.
The name of the test can also be improved.

```
// Regression test for https://github.com/flutter/flutter/issues/40601
testWidgets('Horizontal stepper does not overflow with long step titles'
```

packages/flutter/lib/src/material/stepper.dart
```diff
@@ -914,12 +914,22 @@ class _StepperState extends State<Stepper> with TickerProviderStateMixin {
...
+                  child: Padding(
+                    padding: const EdgeInsets.symmetric(horizontal: 24.0),
+                    child: SizedBox(
+                      height: _stepIconHeight != null ? _stepIconHeight! * _heightFactor : null,
+                      child: IntrinsicWidth(child: Row(children: children)),
```
navaronbracke, line 927: Adding an `IntrinsicWidth` is potentially expensive, if the stepper has a lot of steps.
Instead of using a `SingleChildScrollView`, we can probably wrap the children in a CustomScrollView + SliverPadding + SliverList.builder(), with an item count equal to children.length

Parryword: Hi @navaronbracke, the problem is children are wrapped with Expanded widget (Stepper.dart, ln. 883). If I remove InstinsticWidth, Flutter will throw an error, because you cannot normally have Expanded widget inside a ScrollView. Therefore, IntrinsicWidth is needed despite the computation cost. 

Edit: I will check if your proposal will solve this issue.

packages/flutter/lib/src/material/stepper.dart
Parryword: This sets the min width of ScrollView to width of the screen. This is needed for cases where there are few Steps that don't overflow. If I remove or change minWidth, this will cause ScrollView to have the width of its children, and Step connector lines will not expand into available space. Here is an example:
image

packages/flutter/test/material/stepper_test.dart
Parryword: Done.

packages/flutter/test/material/stepper_test.dart
Parryword: Done.

packages/flutter/test/material/stepper_test.dart
Parryword: Done.

packages/flutter/test/material/stepper_test.dart
Parryword: Done.

packages/flutter/lib/src/material/stepper.dart
Parryword: ~~I have tried your solution. The steps were not rendered due to the fact that the children (Step connectors) are wrapped with Expanded inside a CustomScrollView. Unfortunately, this widget has to use IntrinsticWidth, because we have no way of knowing if the Steps will overflow before we render them.~~ 

Edit: Still working on this.

Parryword: Here's why CustomScrollView and SliverList are not applicaple in this scenario. When CustomScrollView's scroll direction is set to horizontal, CustomScrollView expands vertically, and since it does not have constraints, we get "unbounded height" error. To solve this issue, we need to know the height of children (Steps) and pass it to a SizedBox wrapping CustomScrollView, but we do not know it. Even if we knew the height of children, there is no way to set CustomScrollView's min width to the screen width. This means that CustomScrollView will have the least possible width and Step connectors will not fill the available space.
image

Parryword:
@navaronbracke I also plan to reassign widget's controller from ListView (ln. 936) to SingleChildScrollView (ln. 919), but I want to get your opinion on it.

victorsanni (commented):
Hi, thanks for the attempt.

When I tried fixing this, I stopped because I ran into some [unanswered questions](https://github.com/flutter/flutter/pull/158040#issuecomment-2457836846). Could you give an explanation to help answer some of those?

Also this would be easier to review if there was a video in the PR description showing the previous and the new behavior.

PR #180436, opened 2026-01-02
[Web] Fix IME and selection by syncing more text styles

### LINKED ISSUE #161592: Incorrect position of Japanese predictive conversion popup in TextFormField using maxLines on the Web
### Steps to reproduce

1. Prepare a Mac with macOS Sequoia 15.2 (Japanese).
2. Paste the code sample into [dartpad.dev](https://dartpad.dev/) and run it.
3. Type Japanese in TextFormField and display the predictive conversion popup.

### Expected results

The predictive conversion popup will appear without any redundant space under the text.

### Actual results

The predictive conversion popup is appeared with redundant space under the text.

### Code sample

```dart

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorSchemeSeed: Colors.blue,
      ),
    );
  }
}

  final String title;

  const MyHomePage({
    super.key,
    required this.title,
  });

  @override
}

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: TextFormField(
        maxLines: 10,
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>

</details>

### Logs

None.

</details>

### Flutter Doctor output

None. (Run on dartpad.dev with Based on Dart SDK 3.6.0 and Flutter SDK 3.27.1.)

</details>

koji-1009:
fix https://github.com/flutter/flutter/issues/161592

The current implementation does not fully reflect [letter-spacing](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/letter-spacing), [word-spacing](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-spacing), and [line-height](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-height) in the DOM. 0bc99f8c4f70973a1877c88ca35804e9bc5fabcf
And the current implementation generates an DomHTMLTextAreaElement every time the `enabled`. Therefore, it reapplies the [scrollTop](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop) value that the Element had been holding internally. 726491298d2b1f17681cab421c7c9276dde19ee6

https://github.com/user-attachments/assets/6f575366-12a0-4246-b2ab-eb2a0e85cfa5

https://github.com/user-attachments/assets/ec660d4c-5166-450c-be38-77b90fcfce76

```dart

void main() => runApp(const MainApp());

class MainApp extends StatelessWidget {
  const MainApp({super.key});

  @override
    return const MaterialApp(home: MainPage());
  }
}

class MainPage extends StatelessWidget {
  const MainPage({super.key});

  @override
    return Scaffold(
      appBar: AppBar(title: Text('Selection position')),
      body: SingleChildScrollView(
        padding: const .all(16),
        child: Column(
          children: [
            const Text('height=null'),
            TextField(maxLines: 5, style: TextStyle(height: null)),
            const Text('height=1.0'),
            TextField(maxLines: 5, style: TextStyle(height: 1.0)),
            const Text('height=2.0'),
            TextField(maxLines: 5, style: TextStyle(height: 2.0)),
          ],
        ),
      ),
    );
  }
}
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
@@ -67,6 +67,7 @@ void _setStaticStyleAttributes(DomHTMLElement domElement) {
     ..position = 'absolute'
     ..top = '0'
     ..left = '0'
+    ..margin = '0'
     ..padding = '0'
     ..opacity = '1'
     ..color = 'transparent'
@@ -107,6 +108,7 @@ void _styleAutofillElements(
   final DomCSSStyleDeclaration elementStyle = domElement.style;
   elementStyle
     ..whiteSpace = 'pre-wrap'
+    ..margin = '0'
     ..padding = '0'
     ..opacity = '1'
     ..color = 'transparent'
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
@@ -2501,6 +2501,9 @@ class EditableTextStyle {
     required this.textAlign,
     required this.fontFamily,
     required this.fontWeight,
+    required this.letterSpacing,
+    required this.wordSpacing,
+    required this.lineHeight,
   });
 
   factory EditableTextStyle.fromFrameworkMessage(Map<String, dynamic> flutterStyle) {
@@ -2527,6 +2530,9 @@ class EditableTextStyle {
       textAlign: ui.TextAlign.values[textAlignIndex],
       textDirection: ui.TextDirection.values[textDirectionIndex],
       fontWeight: fontWeight,
+      letterSpacing: flutterStyle.tryDouble('letterSpacing'),
+      wordSpacing: flutterStyle.tryDouble('wordSpacing'),
+      lineHeight: flutterStyle.tryDouble('lineHeight'),
     );
   }
 
@@ -2537,6 +2543,9 @@ class EditableTextStyle {
   final String? fontFamily;
   final ui.TextAlign textAlign;
   final ui.TextDirection textDirection;
+  final double? letterSpacing;
+  final double? wordSpacing;
+  final double? lineHeight;
 
   String? get align => textAlignToCssValue(textAlign, textDirection);
 
@@ -2545,7 +2554,10 @@ class EditableTextStyle {
   void applyToDomElement(DomHTMLElement domElement) {
     domElement.style
       ..textAlign = align!
-      ..font = cssFont;
+      ..font = cssFont
+      ..letterSpacing = letterSpacing != null ? '${letterSpacing}px' : ''
+      ..wordSpacing = wordSpacing != null ? '${wordSpacing}px' : ''
+      ..lineHeight = lineHeight != null ? '${lineHeight}px' : 'normal';
   }
 }
 
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
@@ -1299,6 +1299,9 @@ abstract class DefaultTextEditingStrategy
   /// Size and transform of the editable text on the page.
   EditableTextGeometry? geometry;
 
+  /// The scroll top of the editable text on the page.
+  final Map<String, double> _preservedScrollTops = <String, double>{};
+
   OnChangeCallback? onChange;
   OnActionCallback? onAction;
 
@@ -1454,6 +1457,11 @@ abstract class DefaultTextEditingStrategy
   @override
   void disable() {
     assert(isEnabled);
+    // Preserve the internal scroll position.
+    if (geometry != null && lastEditingState != null) {
+      final key = '${geometry!.hashCode}_${lastEditingState!.text.hashCode}';
+      _preservedScrollTops[key] = activeDomElement.scrollTop;
+    }
 
     isEnabled = false;
     lastEditingState = null;
@@ -1657,6 +1665,12 @@ abstract class DefaultTextEditingStrategy
 
     // Re-focuses after setting editing state.
     moveFocusToActiveDomElement();
+
+    // Restore the internal scroll position.
+    if (geometry != null && lastEditingState != null) {
+      final key = '${geometry!.hashCode}_${lastEditingState!.text.hashCode}';
+      activeDomElement.scrollTop = _preservedScrollTops.remove(key) ?? 0.0;
+    }
   }
 
   /// Prevent default behavior for mouse down, up and move.
@@ -2617,4 +2631,18 @@ class EditableTextGeometry {
       ..height = '${height}px'
       ..transform = cssTransform;
   }
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) {
+      return true;
+    }
+    return other is EditableTextGeometry &&
+        other.width == width &&
+        other.height == height &&
+        listEquals<double>(other.globalTransform, globalTransform);
+  }
+
+  @override
+  int get hashCode => Object.hash(width, height, Object.hashAll(globalTransform));
 }
--- a/packages/flutter/lib/src/services/text_input.dart
+++ b/packages/flutter/lib/src/services/text_input.dart
@@ -1655,6 +1655,10 @@ class TextInputConnection {
   /// This information is used by the Flutter Web Engine to change the style
   /// of the hidden native input's content. Hence, the content size will match
   /// to the size of the editable widget's content.
+  @Deprecated(
+    'Use setStyleWithMetrics instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
   void setStyle({
     required String? fontFamily,
     required double? fontSize,
@@ -1662,9 +1666,7 @@ class TextInputConnection {
     required TextDirection textDirection,
     required TextAlign textAlign,
   }) {
-    assert(attached);
-
-    TextInput._instance._setStyle(
+    setStyleWithMetrics(
       fontFamily: fontFamily,
       fontSize: fontSize,
       fontWeight: fontWeight,
@@ -1690,6 +1692,38 @@ class TextInputConnection {
   void connectionClosedReceived() {
     TextInput._instance._currentConnection = null;
     assert(!attached);
+  }
+}
+
+/// Extension to add [setStyleWithMetrics] to [TextInputConnection].
+extension TextInputConnectionExt on TextInputConnection {
+  /// Send text styling information.
+  ///
+  /// This information is used by the Flutter Web Engine to change the style
+  /// of the hidden native input's content. Hence, the content size will match
+  /// to the size of the editable widget's content.
+  void setStyleWithMetrics({
+    required String? fontFamily,
+    required double? fontSize,
+    required FontWeight? fontWeight,
+    required TextDirection textDirection,
+    required TextAlign textAlign,
+    double? letterSpacing,
+    double? wordSpacing,
+    double? lineHeight,
+  }) {
+    assert(attached);
+
+    TextInput._instance._setStyle(
+      fontFamily: fontFamily,
+      fontSize: fontSize,
+      fontWeight: fontWeight,
+      textDirection: textDirection,
+      textAlign: textAlign,
+      letterSpacing: letterSpacing,
+      wordSpacing: wordSpacing,
+      lineHeight: lineHeight,
+    );
   }
 }
 
@@ -2225,14 +2259,20 @@ class TextInput {
     required FontWeight? fontWeight,
     required TextDirection textDirection,
     required TextAlign textAlign,
+    required double? letterSpacing,
+    required double? wordSpacing,
+    required double? lineHeight,
   }) {
     for (final TextInputControl control in _inputControls) {
-      control.setStyle(
+      control.setStyleWithMetrics(
         fontFamily: fontFamily,
         fontSize: fontSize,
         fontWeight: fontWeight,
         textDirection: textDirection,
         textAlign: textAlign,
+        letterSpacing: letterSpacing,
+        wordSpacing: wordSpacing,
+        lineHeight: lineHeight,
       );
     }
   }
@@ -2424,6 +2464,10 @@ mixin TextInputControl {
   ///
   /// This method is called on the when the attached input client's text style
   /// changes.
+  @Deprecated(
+    'Use setStyleWithMetrics instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
   void setStyle({
     required String? fontFamily,
     required double? fontSize,
@@ -2431,6 +2475,29 @@ mixin TextInputControl {
     required TextDirection textDirection,
     required TextAlign textAlign,
   }) {}
+
+  /// Informs the text input control about text style changes.
+  ///
+  /// This method is called on the when the attached input client's text style
+  /// changes.
+  void setStyleWithMetrics({
+    required String? fontFamily,
+    required double? fontSize,
+    required FontWeight? fontWeight,
+    required TextDirection textDirection,
+    required TextAlign textAlign,
+    double? letterSpacing,
+    double? wordSpacing,
+    double? lineHeight,
+  }) {
+    setStyle(
+      fontFamily: fontFamily,
+      fontSize: fontSize,
+      fontWeight: fontWeight,
+      textDirection: textDirection,
+      textAlign: textAlign,
+    );
+  }
 
   /// Requests autofill from the text input control.
   ///
@@ -2557,12 +2624,35 @@ class _PlatformTextInputControl with TextInputControl {
     required TextDirection textDirection,
     required TextAlign textAlign,
   }) {
+    setStyleWithMetrics(
+      fontFamily: fontFamily,
+      fontSize: fontSize,
+      fontWeight: fontWeight,
+      textDirection: textDirection,
+      textAlign: textAlign,
+    );
+  }
+
+  @override
+  void setStyleWithMetrics({
+    required String? fontFamily,
+    required double? fontSize,
+    required FontWeight? fontWeight,
+    required TextDirection textDirection,
+    required TextAlign textAlign,
+    double? letterSpacing,
+    double? wordSpacing,
+    double? lineHeight,
+  }) {
     _channel.invokeMethod<void>('TextInput.setStyle', <String, dynamic>{
       'fontFamily': fontFamily,
       'fontSize': fontSize,
       'fontWeightIndex': fontWeight?.index,
       'textAlignIndex': textAlign.index,
       'textDirectionIndex': textDirection.index,
+      'letterSpacing': letterSpacing,
+      'wordSpacing': wordSpacing,
+      'lineHeight': lineHeight,
     });
   }
 
--- a/packages/flutter/lib/src/services/text_input.dart
+++ b/packages/flutter/lib/src/services/text_input.dart
@@ -2507,7 +2507,7 @@ mixin TextInputControl {
   /// changes.
   @Deprecated(
     'Use updateStyle instead. '
-    'This feature was deprecated after v3.40.0-0.2.pre.',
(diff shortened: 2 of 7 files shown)


engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -1450,6 +1455,11 @@ abstract class DefaultTextEditingStrategy
   @override
   void disable() {
     assert(isEnabled);
+    // Preserve the internal scroll position.
+    if (geometry != null && lastEditingState != null) {
+      final key = '${geometry!.hashCode}_${lastEditingState!.text.hashCode}';
```
mdebbar, line 1464: I'm somewhat concerned about this `key`, it doesn't seem like a robust solution. But I don't have any better ideas at the moment.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1655,16 +1655,18 @@ class TextInputConnection {
   /// This information is used by the Flutter Web Engine to change the style
   /// of the hidden native input's content. Hence, the content size will match
   /// to the size of the editable widget's content.
+  @Deprecated(
+    'Use setStyleWithMetrics instead. '
+    'This feature was deprecated after v3.40.0-0.2.pre.',
+  )
```
mdebbar, line 1769: @loic-sharma any thoughts on doing this vs adding the parameters to the existing `setStyle(...)` method without deprecating it? The latter may be considered a breaking change if users are expected to extend/implement `TextInputConnection`?

loic-sharma: It looks like people do implement `TextInputConnection` ([1](https://github.com/Flutter-Bounty-Hunters/super_editor/blob/77e8d306a5bb689706f87b9dd269249323af470c/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart#L22), [2](https://github.com/Flutter-Bounty-Hunters/super_editor/blob/77e8d306a5bb689706f87b9dd269249323af470c/super_editor/lib/src/default_editor/document_ime/ime_decoration.dart#L8)).

The current approach does avoid a breaking change, but it prevents users from overriding `setStyleWithMetrics`. I don't think we should do this approach since people do override `setStyle` ([example](https://github.com/Flutter-Bounty-Hunters/super_editor/blob/77e8d306a5bb689706f87b9dd269249323af470c/super_editor/lib/src/default_editor/document_ime/ime_decoration.dart#L37-L49)).

What do y'all think of this idea:

1. Introduce a new `TextInputStyle` class, similar to the existing [`TextInputConfiguration`](https://api.flutter.dev/flutter/services/TextInputConfiguration-class.html):

   Something like:
  
   ```dart
   final class TextInputStyle {
     const TextInputStyle(
        ...
     );
   
     final String? fontFamily;
     final double? fontSize;
     final FontWeight? fontWeight;
     final TextDirection textDirection;
     final TextAlign textAlign;
     final double? letterSpacing;
     final double? wordSpacing;
     final double? lineHeight;
   }
   ```

   _Notice `TextInputStyle` is [`final`](https://dart.dev/language/class-modifiers#final)._

2. Add a new `TextInputConnection.updateStyle(TextInputStyle style)` method that  calls `TextInput._instance._setStyle`.
3. Deprecate `TextInputConnection.setStyle`.
4. Do all the stuff in https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes

This is still a breaking change - if your package implements `TextInputConnection`, you'll need to add a new `updateStyle` method.

However, this is a one-time breaking change. In the future, we'll be able to add more members to `TextInputStyle` without it being a breaking change.

What do y'all think?

loic-sharma: cc @Renzo-Olivares too, would love your thoughts as well

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -3409,12 +3409,15 @@ class EditableTextState extends State<EditableText>
...
-        _textInputConnection!.setStyle(
+        _textInputConnection!.setStyleWithMetrics(
           fontFamily: _style.fontFamily,
           fontSize: _style.fontSize,
           fontWeight: _style.fontWeight,
           textDirection: _textDirection,
           textAlign: widget.textAlign,
+          letterSpacing: _style.letterSpacing,
+          wordSpacing: _style.wordSpacing,
+          lineHeight: renderEditable.preferredLineHeight,
```
loic-sharma: I suspect this value can be stale. I believe `RenderEditable.preferredLineHeight` isn't updated until `_Editable` rebuilds, which happens after `EditableTextState.didUpdateWidget` is called.

loic-sharma: Just like these take the `MediaQuery.boldTextOf` account, these likely need to take the following into account:

1. `MediaQuery.maybeLineHeightScaleFactorOverrideOf(context);`
2. `MediaQuery.maybeLetterSpacingOverrideOf(context);`
3. `MediaQuery.maybeWordSpacingOverrideOf(context);`

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
koji-1009: Yes, this key approach has edge cases.

I wanted to keep this PR scope small first. So I made this solution closed within the flutter engine side.

Syncing scrollTop from flutter framework when scroll occurs would solve this problem better. But that would add another breaking change discussion.

If the team prefers the more robust approach, I can work on a follow-up PR 👍

packages/flutter/lib/src/services/text_input.dart
Renzo-Olivares: Chatted offline but I agree with the `TextInputStyle` approach!

koji-1009:
Thanks for the heads up! I'll make the PR ready once I've fixed it and it passes tests excluding customer testing.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2557,12 +2663,28 @@ class _PlatformTextInputControl with TextInputControl {
...
-      'textAlignIndex': textAlign.index,
-      'textDirectionIndex': textDirection.index,
+      'fontFamily': style.fontFamily,
+      'fontSize': style.fontSize,
+      'fontWeightIndex': style.fontWeight?.index,
+      'textAlignIndex': style.textAlign.index,
+      'textDirectionIndex': style.textDirection.index,
+      'letterSpacing': style.letterSpacing,
+      'wordSpacing': style.wordSpacing,
+      'lineHeight': style.lineHeight,
```
loic-sharma: Could you move this logic to `TextStyle.toJson`, similar to `TextInputConfiguration.toJson`?

packages/flutter/test/services/text_input_test.dart
```diff
@@ -1612,6 +1608,11 @@ class FakeTextInputControl with TextInputControl {
     methodCalls.add('setStyle');
   }
 
+  @override
+  void updateStyle(TextInputStyle style) {
+    methodCalls.add('updateStyle');
+  }
```
loic-sharma, line 1699: Could we unit test the `equals`, `hashCode`, and `toJson` methods, similar to `TextInputConfiguration`'s unit tests? See:

https://github.com/flutter/flutter/blob/dfd92b7732198de288d9fb7453e8ef52b8979ec2/packages/flutter/test/services/text_input_test.dart#L319-L541

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -3409,13 +3409,29 @@ class EditableTextState extends State<EditableText>
...
+              textDirection: _textDirection,
+              textAlign: widget.textAlign,
+              letterSpacing: letterSpacingOverride ?? _style.letterSpacing,
+              wordSpacing: wordSpacingOverride ?? _style.wordSpacing,
+              // preferredLineHeight already includes lineHeightScaleFactor from
+              // _OverridingTextStyleTextSpanUtils.applyTextSpacingOverrides.
+              lineHeight: renderEditable.preferredLineHeight,
+            ),
+          );
+        }, debugLabel: 'EditableText.updateStyle');
```
loic-sharma, line 3427: I believe we also need to update `didChangeDependencies` so that when `MediaQuery.maybeLetterSpacingOverrideOf(context)` / `MediaQuery.maybeWordSpacingOverrideOf(context)` /  `MediaQuery.maybeLineHeightScaleFactorOverrideOf(context)` change, we update the connection's style as well.

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -3409,13 +3409,29 @@ class EditableTextState extends State<EditableText>
...
+              fontWeight: _style.fontWeight,
+              textDirection: _textDirection,
+              textAlign: widget.textAlign,
+              letterSpacing: letterSpacingOverride ?? _style.letterSpacing,
+              wordSpacing: wordSpacingOverride ?? _style.wordSpacing,
+              // preferredLineHeight already includes lineHeightScaleFactor from
+              // _OverridingTextStyleTextSpanUtils.applyTextSpacingOverrides.
+              lineHeight: renderEditable.preferredLineHeight,
+            ),
+          );
```
loic-sharma: Can we avoid using a post-frame callback here? A post-frame callback means we'll render a frame with a potentially incorrect text input style, and then fix it up in the next frame. That could lead to weird jumps and other unexpected behaviors.

A couple options:

* **Option 1 - Punt**: Land this PR without syncing line height text metrics. Adding a `TextInputStyle` and fixing letter/word spacing is a great improvement already. We should also fix line height metrics, but I suspect this requires non-trivial refactoring of `EditableText` to do properly.
* **Option 2 - Refactor EditableText**. Currently, the value for `RenderEditable.text` is updated in `EditableTextState.build` through `_Editable.inlineSpan`, so `renderEditable.lineHeight` is stale until after `EditableTextState.build` completes.
   
   To fix this, we update `EditableTextState.didUpdateWidget` and `EditableTextState.didChangeDependencies` such that it updates all the underlying `RenderEditable` properties needed to get the latest `RenderEditable.preferredLineHeight`. This would likely require updating `RenderEditable.text`, `RenderEditable.textScaler`, etc...

   I'm not sure if this is a good idea. I'll discuss this matter with @justinmc, @LongCatIsLooong, and @Renzo-Olivares at our next text input sync this Thursday.

loic-sharma: Actually, please ignore this comment! @Renzo-Olivares pointed out that we already use post-frame callbacks to update the style [here](https://github.com/flutter/flutter/blob/d93e2d4b82ba2cc1b730948a81f823d52374d5dc/packages/flutter/lib/src/widgets/editable_text.dart#L3382-L3390), and @LongCatIsLooong pointed out that since style updates go through a platform channel, the style update isn't applied immediately in the frame anyways. I'll close this off!

koji-1009:
[Linux linux_host_engine_test](https://github.com/flutter/flutter/pull/180436/checks?check_run_id=61714585024) is failed.

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8691441949929312513/+/u/test:_Host_Tests_for_host_release/stdout?format=raw

```
[ERROR:flutter/shell/testing/tester_main.cc(612)] Unhandled exception
Exception: Bad state: Can't call test() once tests have begun running.
When running a test as an executable directly (not as a suite by the test runner), tests must be declared in a synchronous block.
If async work is required before any tests are run use a `setUpAll` callback.
If async work cannot be avoided before declaring tests, all async events must be complete before declaring the first test.
Stack trace: #0      Declarer._checkNotBuilt (package:test_api/src/backend/declarer.dart:418:5)
#1      Declarer.test (package:test_api/src/backend/declarer.dart:210:5)
#2      test (package:test_core/src/scaffolding.dart:157:13)
#3      main (file:///b/s/w/ir/cache/builder/engine/src/flutter/testing/dart/fragment_shader_test.dart:650:5)
<asynchronous suspension>
```

https://github.com/flutter/flutter/blob/e435edfbcd26c0107eccce443756b589febb7987/engine/src/flutter/testing/dart/fragment_shader_test.dart#L643-L667

This failure seems to be caused by the order in which the test code is written. Placing `final ImageComparer comparer = await ImageComparer.create();` just after the `main()` or inside the `test()` should resolve it.

packages/flutter/lib/src/services/text_input.dart
koji-1009: I fixed by [82c5465](https://github.com/flutter/flutter/pull/180436/commits/82c546553a593a13de191817323c4944df0f65ed).

packages/flutter/test/services/text_input_test.dart
koji-1009: I fixed by [82c5465](https://github.com/flutter/flutter/pull/180436/commits/82c546553a593a13de191817323c4944df0f65ed).

loic-sharma:
> [Linux linux_host_engine_test](https://github.com/flutter/flutter/pull/180436/checks?check_run_id=61714585024) is failed.

That doesn't seem related to your changes. This _might_ be a flake or that _might_ be an issue with the commit you branched off of. Usually when I get an unrelated error like this, I rebase off the latest `master` commit :)

packages/flutter/test/widgets/editable_text_styles_test.dart
```diff
@@ -0,0 +1,274 @@
...
+    focusNode.dispose();
+  });
+
+  testWidgets('EditableText updates style when boldText changes', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      MediaQuery(
+        data: const MediaQueryData(),
+        child: Directionality(
+          textDirection: TextDirection.ltr,
+          child: EditableText(
```
Renzo-Olivares, line 93: @loic-sharma @justinmc would this be a good place to use `TestTextField`? Is there a certain type of test where we would recommend the more verbose `EditableText` approach vs using `TestTextField` and vice versa. My thought is that tests that rely on gestures should probably use `TestTextField` but outside of that the main difference is verbosity.

packages/flutter/test/widgets/editable_text_styles_test.dart
```diff
@@ -0,0 +1,274 @@
...
+      ),
+    );
+
+    await tester.tap(find.byType(EditableText));
+    await tester.pump();
+
+    expect(control.lastStyle?.letterSpacing, 1.0);
+    control.lastStyle = null;
+
+    // Update letterSpacingOverride
```
Renzo-Olivares: super nit: period after comments here and elsewhere.

packages/flutter/test/widgets/editable_text_styles_test.dart
loic-sharma: @Renzo-Olivares Excellent question. My intuition would be to:

* Prefer `EditableText` for tests specific to `EditableText` (e.g. `EditableText.rendererIgnoresPointer`)
* Prefer `TestTextField` if the test needs a fully functioning text field (e.g. gesture support)
* Prefer `TestTextFIeld` if using `EditableText` would result in large amounts of boilerplate
* Otherwise if none of the guidelines apply, prefer `EditableText`.

----

A drawback of using `TestTextField` for _all_ tests is that it would reduce our visibility on the usability of the `EditableText`'s API. Ideally, `EditableText` would be easy to use (per your recommendations in https://flutter.dev/go/decoupling-design-from-text). If we introduce additional bloat to `EditableText`'s API, using `TestTextField` _everywhere_ might "hide" the usability impact. If we use `EditableText` in _some_ of our tests, we'll have better sense of the usability impact of `EditableText` API changes.

(However, a benefit of using `TestTextField` for _all_ tests is that it would make it easier for us to migrate to `EditableTextField2` if/when we introduce that)

----

What do you think of this reasoning? Is there anything you disagree with, or other cases you'd add?

Assuming y'all agree with this reasoning, I'd _slightly_ lean towards using `EditableText` for these tests: the behavior isn't specific to `EditableText`, however, the boilerplate if you use `EditableText` is very reasonable.

Renzo-Olivares: I agree with the reasoning above! Definitely see the concern of reducing visibility over `EditableText`.

I'd also lean towards `EditableText` for these tests since there is logic that is being added to `EditableText` specifically and it does not rely on gestures.

packages/flutter/test/services/text_input_test.dart
loic-sharma: Thanks!

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -5783,6 +5789,9 @@ void main() {
           'fontWeightIndex': 6,
           'textAlignIndex': 4,
           'textDirectionIndex': 1,
+          'letterSpacing': null,
+          'wordSpacing': null,
```
loic-sharma: For this test, could we add letter spacing and word spacing to `textStyle2` so these values are non-null?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -3438,6 +3447,24 @@ class EditableTextState extends State<EditableText>
     }
   }
 
+  TextInputStyle _getTextInputStyle() {
```
loic-sharma: Instead of using the `State.context` property, I'd consider making this method accept a `BuildContext` argument. That would make it clearer from the call site that this method requires the widget is mounted.

I don't feel strongly about this though, feel free to skip.

loic-sharma (approved):
This looks good to me, but before landing this please also get an approval from @mdebbar for the Flutter Web side.

packages/flutter/test/widgets/editable_text_test.dart
koji-1009: I've fixed it in [4624815](https://github.com/flutter/flutter/pull/180436/commits/46248156d766ada2b71a6ec0c3f33db30840d905). Thank you.

packages/flutter/lib/src/widgets/editable_text.dart
koji-1009: Agreed! I've confirmed other classes follow this pattern too, so I've updated the code accordingly. Thanks for the suggestion! [552d627](https://github.com/flutter/flutter/pull/180436/commits/552d627d9b81674f3ce1a40226bf021d3c818368)

packages/flutter/test/widgets/editable_text_test.dart
loic-sharma: Sweet, thanks!

packages/flutter/lib/src/widgets/editable_text.dart
loic-sharma: Sounds good, thanks!

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1518,6 +1518,114 @@ mixin DeltaTextInputClient implements TextInputClient {
...
+      'textAlignIndex': textAlign.index,
+      'textDirectionIndex': textDirection.index,
+      'letterSpacing': letterSpacing,
+      'wordSpacing': wordSpacing,
+      'lineHeight': lineHeight,
+    };
+  }
+
+  @override
+  String toString() {
```
Renzo-Olivares: Consider mixing in `Diagnosticable` and using `debugFillProperties` here. https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#override-tostring

Renzo-Olivares (commented):
framework LGTM w/ one small comment

packages/flutter/lib/src/services/text_input.dart
koji-1009: I've added `Diagnosticable` mixin to `TextInputStyle` and replaced `toString` with `debugFillProperties`. [632d088](https://github.com/flutter/flutter/pull/180436/commits/632d08861678f25750aa16df4e827bd3890f04d2)

Note: I kept this change scoped to the new `TextInputStyle` class only, as modifying existing classes like `TextInputConfiguration` would expand the scope of this PR.

mdebbar (approved):
LGTM!

PR #180452, opened 2026-01-03
fix: clear git environment variables in version.dart

kumarprashant94395:
This PR clears git environment variables before running git commands in version.dart. This ensures that the git commands run correctly even when alternate object directories are present.


kumarprashant94395:
Hi @bkonyi, thank you for the feedback. I have been trying my best to fix
the issue, but I am currently struggling to get the CI checks to pass due
to some syntax errors. I am working on resolving these now, but if you see
any specific improvements needed in the code I've submitted, please feel
free to suggest or guide me. Once the tests are green, I will create the
separate issue and add the required tests as you suggested. Thanks for your
patience!"

Issue #177415, opened 2025-10-22
(Test cross-imports) Don't import Material in tests for convenience


justinmc:
Umbrella issue: https://github.com/flutter/flutter/issues/177028
Design doc: https://docs.google.com/document/d/1UHxALQqCbmgjnM1RNV9xE2pK3IGyx-UktGX1D7hYCjs/edit?tab=t.0

Tests shouldn't import the Material library simply out of convenience (such as to use MaterialApp instead of WidgetsApp). This will help with our general test organization and also help remove a devDependency after Material and Cupertino are decoupled from the framework.

### How to fix
Replace the "convenient" widget imported from another library with something in the current library. If needed, create a helper to make it easier to accomplish the same thing in tests.

Widget | What to do
 --- | ---
Material | This is probably used to wrap widgets that require a Material ancestor. If possible, just remove it, or replace it with something like Container. If not possible, consider replacing or removing the widget that depends on its presence.
MaterialApp | Replace with WidgetsApp. We may want to consider a TestWidgetsApp in a utils file for convenience.
Scaffold | Remove it if possible.
Colors | Replace with Color and a hex value. Per offline discussion with @victorsanni, consider creating a utils file with convenient color constants, but only in the tests. Use prefixed names like debugRed, etc., to avoid collisions if used with static extensions and dot shorthands in the future.
ListTile | Remove, or replace with something like Container.
Icons | Remove if possible.
ElevatedButton | We should create a TestButton.
Checkbox | If it can't be removed, replace it with buildToggleable, and/or we should create a TestCheckbox or RawCheckbox.
MaterialPageRoute | Use PageRouteBuilder.
Divider | Can it just be removed? Or possibly we should create a RawDivider in Widgets if there is a real need.
TextField | Replace with EditableText. @Renzo-Olivares is investigating creating a test-only EditableText wrapper to make this easier.

#### Located in Widgets, imported Material
 - [ ] [test/widgets/slivers_evil_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/slivers_evil_test.dart)
   - Located in Widgets, imported Material.
   - Category: Convenience 
 - [ ] [test/widgets/opacity_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/opacity_test.dart)
   - Located in Widgets, imported Material.
   - Category: Convenience 
[...]

Issue #177028, opened 2025-10-14
☔ Framework unit tests shouldn't cross-import libraries


justinmc:
Currently, unit tests in the Widgets, Material, and Cupertino libraries freely import between each other, usually without a good reason. This is confusing for maintainers trying to determine test coverage and for contributors trying to decide where to place their tests.

For example, cupertino/magnifier_test.dart uses MaterialApp simply out of convenience or habit:

https://github.com/flutter/flutter/blob/56ce192dc679a059c1f5c566e373669c96157123/packages/flutter/test/cupertino/magnifier_test.dart#L48

Instead, I propose that we should enforce the following structure:

 - The Widgets library unit tests should not import Material or Cupertino.
 - The Cupertino library should not import Material.
 - The Material library should contain all tests that require imports of multiple libraries.

Design doc: https://docs.google.com/document/d/1UHxALQqCbmgjnM1RNV9xE2pK3IGyx-UktGX1D7hYCjs/edit?tab=t.0

### Sub-issues

 * https://github.com/flutter/flutter/pull/178693
 * https://github.com/flutter/flutter/issues/177415
 * https://github.com/flutter/flutter/issues/177414
 * https://github.com/flutter/flutter/issues/177412
 * https://github.com/flutter/flutter/issues/179429

### Data

Reasons | Totals | Percentage
-- | -- | --
Legacy | 56 | 21.37%
Convenience | 127 | 48.47%
Misplacement | 57 | 21.76%
Legitimate cross-platform testing | 22 | 8.40%

<img width="736" height="450" alt="Image" src="https://github.com/user-attachments/assets/68aabb04-725b-41d9-908c-4d14a2dd9e02" />

<details>

<summary>Confirmed acceptable cross-imports:</summary>

#### Located in Material, imported Cupertino
 - [x] [test/material/progress_indicator_test.dart](test/material/progress_indicator_test.dart)
   - Located in Material, imported Cupertino.
   - Category: Legitimate cross-platform testing
 - [x] [test/material/adaptive_text_selection_toolbar_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/material/adaptive_text_selection_toolbar_test.dart)
   - Located in Material, imported Cupertino.
   - Category: Legitimate cross-platform testing  
 - [x] [test/material/slider_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/material/slider_test.dart)
   - Located in Material, imported Cupertino
[...]

PR #180456, opened 2026-01-03
Add TestWidgetsApp utility and refactor widget tests to use WidgetsApp

rkishan516:
This PR introduces a TestWidgetsApp utility class that provides a minimal WidgetsApp wrapper for widget tests, replacing direct usage of Directionality widgets.

part of: #177415 

## Changes
- test_widgets_app.dart - A reusable TestWidgetsApp widget that wraps WidgetsApp with a builder pattern, providing Directionality, MediaQuery, and other app-level widgets that WidgetsApp provides.
- gesture_detector_test.dart - Replaced 9 Directionality usages
- mouse_region_test.dart - Replaced 13 Directionality usages
- opacity_test.dart - Replaced 2 Directionality usages
- slivers_evil_test.dart - Replaced 3 Directionality usages and removed redundant MediaQuery wrappers

## Note
- This PR updates golden test baselines for opacity_test.dart.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/widgets_app_tester.dart
```diff
@@ -0,0 +1,49 @@
...
+  final Widget child;
+
+  /// The primary color for the application.
+  ///
+  /// Defaults to white.
+  final Color color;
+
+  @override
+  Widget build(BuildContext context) {
+    return WidgetsApp(
```
victorsanni, line 50: I ran into an issue writing tests with WidgetsApp whenever a descendant widget tries to insert an OverlayEntry into the root overlay because (I think) WidgetsApp doesn't create an Overlay by default unlike Material/CupertinoApp. This was the workaround:

```dart
WidgetsApp(
  ...
  pageRouteBuilder: <T>(RouteSettings settings, WidgetBuilder builder) {
    return PageRouteBuilder<T>(
      pageBuilder:
          (
            BuildContext context,
            Animation<double> animation,
            Animation<double> secondaryAnimation,
          ) => builder(context),
    );
  },
);
```

Maybe we can add this here as well so others don't run into this problem.

packages/flutter/test/widgets/gesture_detector_test.dart
```diff
@@ -3,10 +3,12 @@
 // found in the LICENSE file.
 
 import 'package:flutter/gestures.dart';
-import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
 
+import 'test_widgets_app.dart';
```
victorsanni: Wondering if this class can be exported from the `flutter_test` package instead of from `widgets.dart`. Feels awkward that a supposedly test-only class exists in source code.

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: Sure, I have added it.

packages/flutter/test/widgets/gesture_detector_test.dart
rkishan516: Makes sense. I have moved to flutter_test.

packages/flutter/test/widgets/gesture_detector_test.dart
```diff
@@ -177,8 +177,7 @@ void main() {
 
       Future<void> pumpWidgetTree(HitTestBehavior? behavior) {
         return tester.pumpWidget(
-          Directionality(
-            textDirection: TextDirection.ltr,
+          TestWidgetsApp(
```
justinmc: Good catch that we can get rid of cases where Directionality is used instead of a WidgetsApp, I didn't think of that.

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -814,7 +811,7 @@ void main() {
               },
               child: Container(
                 key: key,
-                color: Colors.blue,
+                color: const Color(0xFF2196F3),
```
justinmc: We may consider adding a `TestColorBlue` or something like that in the future, but this is fine for now. I'm not yet convinced we want to do that.

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -2006,13 +2001,26 @@ void main() {
...
+              builder: (BuildContext context) => Center(
+                child: DefaultTextStyle(
+                  style: const TextStyle(fontSize: 14),
+                  child: Draggable<int>(
+                    feedback: Container(width: 20, height: 20, color: const Color(0xFF2196F3)),
+                    childWhenDragging: Container(
+                      width: 20,
+                      height: 20,
+                      color: const Color(0xFFFFEB3B),
+                    ),
```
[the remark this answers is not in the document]
justinmc: Not true Gemini!

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -2006,13 +1999,26 @@ void main() {
...
     await tester.pumpWidget(
-      MaterialApp(
-        home: Center(
-          child: Draggable<int>(
-            feedback: Container(width: 20, height: 20, color: Colors.blue),
-            childWhenDragging: Container(width: 20, height: 20, color: Colors.yellow),
-            child: ElevatedButton(child: const Text('Drag me'), onPressed: () {}),
-          ),
+      TestWidgetsApp(
+        child: Overlay(
```
justinmc: Should we consider adding Overlay to TestWidgetsApp or do you think that's overkill when it's not needed?

packages/flutter/test/widgets/opacity_test.dart
```diff
@@ -151,22 +152,20 @@ void main() {
...
+        child: SingleChildScrollView(
+          child: RepaintBoundary(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: List<Widget>.generate(10, (int index) {
+                return const Opacity(
+                  opacity: 0.5,
+                  child: Padding(
+                    padding: EdgeInsets.all(5.0),
+                    child: ColoredBox(color: Color(0xFF2196F3), child: SizedBox(height: 50)),
```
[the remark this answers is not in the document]
justinmc: Not necessary because Opacity is already const.

packages/flutter/test/widgets/widgets_app_tester.dart
justinmc: Good catch. I also noticed that there is a test above that includes an Overlay now. Does this change here help? Or should we add an Overlay here so we don't need to include it in the test?

justinmc (commented):
Thanks for jumping on this! This is basically what I had in mind for TestWidgetsApp.

Looks good minus a few nits, but I see that there is a leak in text_field_test strangely. @rkishan516 any idea what would trigger that?

packages/flutter/test/widgets/widgets_app_tester.dart
navaronbracke: In https://github.com/flutter/flutter/pull/180990 I also wanted to add a WidgetsApp for testing, which used a `pageBuilder` (not a `pageRouteBuilder`), but I suppose the code sample here is equivalent? (since it uses a stub `PageRouteBuilder()` )
Also, the `child` can be renamed to `home` and passed to `WidgetsApp(home: home)` ?

rkishan516: > Good catch. I also noticed that there is a test above that includes an Overlay now. Does this change here help? Or should we add an Overlay here so we don't need to include it in the test?

@justinmc I think we can add Overlay here, which will help in multiple instances.

rkishan516: > In https://github.com/flutter/flutter/pull/180990 I also wanted to add a WidgetsApp for testing, which used a pageBuilder (not a pageRouteBuilder), but I suppose the code sample here is equivalent? (since it uses a stub PageRouteBuilder() )

Yes , its equivalent.

> Also, the child can be renamed to home and passed to WidgetsApp(home: home) ?

Sure, can do.

packages/flutter/test/widgets/gesture_detector_test.dart
rkishan516: Yup, it will make widgets/ test lookwise more consistent.

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: It has it's pro and con, for now I have added. I feel its not adding any problems to other tests(currently) + reducing boilerplate.

rkishan516:
> Looks good minus a few nits, but I see that there is a leak in text_field_test strangely. @rkishan516 any idea what would trigger that?

I will look at it. There is no change in text_field_test specific file.

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: Nvm, its complicating TestWidgetApp and tests both. So, keeping both separated.

packages/flutter/test/widgets/mouse_region_test.dart
navaronbracke: Drive by comment: what about `const Color(0xFF0000FF)` ? I don't think the r,g channels matter here.

packages/flutter/test/widgets/widgets_app_tester.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// @docImport 'package:flutter/cupertino.dart';
```
navaronbracke, line 5: @justinmc Are doc imports allowed? When Material and Cupertino are split off, this will cause `comment_references` warnings.

Apparently that lint is not yet enabled, see https://github.com/flutter/flutter/blob/master/analysis_options.yaml#L95

but once the linked issue is fixed and the lint is enabled, this will be an issue.

packages/flutter_test/lib/src/test_widgets_app.dart
```diff
@@ -0,0 +1,61 @@
...
+  /// The primary color for the application.
+  ///
+  /// Defaults to white.
+  final Color color;
+
+  @override
+  Widget build(BuildContext context) {
+    return WidgetsApp(
+      color: color,
+      builder: (BuildContext context, Widget? navigatorChild) {
```
navaronbracke: @rkishan516 The `home` parameter should be passed to `WidgetsApp.home` in the constructor, to match what WidgetsApp does.

I'm also not sure if `builder: (BuildContext context, Widget? navigatorChild) {...}` is needed for anything specific in tests? Can we remove `builder` ?

rkishan516: No, builder is not need in current version. I will update to use WidgetApp.home rather than builder.

packages/flutter/test/widgets/widgets_app_tester.dart
justinmc, line 1: I just realized this is in flutter_test. We've been putting utils like this in utils files in test/widgets ([example](https://github.com/flutter/flutter/pull/180494/files#diff-ac14371a4efcf19679d81fa776791c0d1e73830ecdaaa5eb7e12e8ea59fad419)). But I see the advantage here, by putting it in flutter_test you allow our users to use this in their tests too.

@Renzo-Olivares I wonder if you thought about this at all when you wrote TestTextField?

packages/flutter/test/widgets/widgets_app_tester.dart
justinmc: Good catch. It looks like there are a couple other similar docImports around flutter_test ([example](https://github.com/flutter/flutter/blob/8cbe1be5e6cf8dce28229d9cb1a7dbe1335c22e4/packages/flutter_test/lib/src/accessibility.dart#L5)). I say we keep this as-is for now. If we need to migrate it in when we decouple, we can migrate this one too.

I will write this down as something to figure out before decoupling.

justinmc (commented):
Looks good but I want to be sure about putting this in flutter_test (comment below).

packages/flutter/test/widgets/widgets_app_tester.dart
justinmc: I talked to @Renzo-Olivares offline, and this same discussion came up in that PR: https://github.com/flutter/flutter/pull/180494/files#r2677591418

I think I agree with that argument. So this should go in the tests/ directory for now, but we should document it very well (as you have already done) and consider moving it to flutter_test in the future when the API is more proven.

packages/flutter/test/widgets/widgets_app_tester.dart
navaronbracke: Maybe we should add a sub issue to the umbrella of the cross imports project where we
1) update the check you wrote for regular dependencies, to handle `@docImport` prefixes as well
2) update the items list with the results of checking doc imports
3) Either clean up doc imports in the same issue as for the regular ones, or split that off (since this is only an issue once the lint is enabled, which could happen post split anyway?)

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: Cool, I will move it to `tests/`. Should we create a issue for moving these to `flutter_test`?

justinmc: Good idea, here you go: https://github.com/flutter/flutter/issues/181283

Maybe add it as a TODO next to the declaration of TestWidgetsApp.

packages/flutter/test/widgets/widgets_app_tester.dart
justinmc: Good call, this is more complicated than I thought. I wrote up an issue: https://github.com/flutter/flutter/issues/181288

And I added it as "related" on https://github.com/flutter/flutter/issues/177028, though not as a strict subissue.

justinmc (approved):
LGTM if the tests are passing.

packages/flutter/test/widgets/widgets_app_tester.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
victorsanni, line 1: I wonder if we've considered potential names for this file that make it easier to find/group with related files. For example, this library has a `TestSemantics` class in `semantics_tester.dart`, so one such name could be `widgets_app_tester.dart` containing `TestWidgetsApp`.  

Another option is appending `utils` or `helpers` like many other files in this lib do (e.g sensitive_content_utils) to create something like `widgets_app_utils.dart`. Although I prefer this option less because `TestWidgetsApp` is not a helper/utils class for `WidgetsApp`, but its own separate entity serving a different purpose.

packages/flutter/test/widgets/opacity_test.dart
```diff
@@ -196,16 +195,18 @@ void main() {
     await tester.pumpWidget(
       RepaintBoundary(
         key: key,
-        child: Directionality(
```
victorsanni, line 199: I see a ton of these `Directionality` -> `TestWidgetsApp` replacements, do we gain anything here asides from potential new bugs as `TestWidgetsApp` gets bigger? IMO keeping to the primitives here >>> an all-encompassing utility class that will break a million tests if it ever gets buggy.

victorsanni (commented):
Thanks for the work here, I left some more comments.

I think test_widgets_app.dart will need its own test (`test_widgets_app_test`)? To verify its properties are passed in and resolved appropriately. Also to verify edge cases like overlays etc work.

packages/flutter/test/widgets/opacity_test.dart
rkishan516: Reason behind converting Directionality to TestWidgetsApp was Consistency across tests + end user will always use some form of WidgetApp at top + easy to use API (combination of Directionality + MediaQuery + other). Also discussed same thing at https://github.com/flutter/flutter/pull/180456#discussion_r2692383733.

If you still feel we should change back to directionality, I can do that.

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: As this will go to `flutter_test`, I think we can continue with `test_widgets_app` because all other are suffixed with `_test.dart`, so its distinguishable.
Or
We can create another folder inside `tests/widget/` which will have all these widgets. WDYT?

packages/flutter/test/widgets/gesture_detector_test.dart
victorsanni: I'm worried that replacing Directionality means bugs with TestWidgetsApp can propagate across the codebase. Or am I looking at this the wrong way? i.e If there is a bug with TestWidgetsApp (or its underlying widgets) that hasn't already been caught, using TestWidgetsApp everywhere stress tests it, making any bugs more likely to be caught?

packages/flutter/test/widgets/mouse_region_test.dart
victorsanni: I personally prefer using easier colors (e.g colors with only one channel, completely transparent etc). Using colors like above can be confusing to someone reading the code who might then think the complexity of the color is an indicator that it serves a non-trivial purpose.

packages/flutter/test/widgets/gesture_detector_test.dart
rkishan516: Since we have many test already, any change in TestWidgetsApp will be stress tested. But something that is not covered in these tests and newly added can be a reason to introduce bug. So, IMO its 50-50 on both side.

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: Updated to single channel colors.

packages/flutter/test/widgets/opacity_test.dart
navaronbracke: @rkishan516 I have to agree with @victorsanni here, some tests do the inverse: use `MaterialApp` where `Directionality` would have been sufficient. So I would keep the `Directionality` if it does not affect the test.

It might not make a test "consistent" with other tests, but the test will be easier to read, since then it is _just_ Directionality as root.

rkishan516: I'm okay with both approaches, just waiting for @victorsanni's go ahead.

packages/flutter/test/widgets/widgets_app_tester.dart
victorsanni: > I think we can continue with test_widgets_app because all other are suffixed with _test.dart, so its distinguishable

Being distinguishable from other tests is not sufficient IMO. File names should make it easy for a first time contributor to this library to tell what groups of tests exist, and have a perception of what to expect when opening a file just from its name.

I personally am leaning towards `widgets_app_tester.dart` because of the grouping with `semantics_tester.dart` which makes it easy to interpret the `-tester` suffix as "files containing test variant classes of another class", i.e `TestWidgetsApp` for `WidgetsApp` and `TestSemantics` for `Semantics`. But open to better suggestions that achieve the goal of easy understandability.

packages/flutter/test/widgets/opacity_test.dart
victorsanni: I prefer `Directionality`. The simpler and more primitive, the better IMO.

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: Updated.

packages/flutter/test/widgets/opacity_test.dart
rkishan516: Updated.

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -1241,6 +1247,8 @@ void main() {
         },
       ),
     );
+    // Let WidgetsApp settle before checking frame scheduling.
+    await tester.pumpAndSettle();
```
justinmc: Why is this needed now?

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -1299,9 +1307,8 @@ void main() {
         );
       }
 
-      return Directionality(
-        textDirection: TextDirection.ltr,
-        child: Align(
+      return TestWidgetsApp(
+        home: Align(
```
justinmc: Should this remain Directionality per https://github.com/flutter/flutter/pull/180456#discussion_r2729585292?

packages/flutter/test/widgets/widgets_app_tester.dart
```diff
@@ -0,0 +1,60 @@
...
+///     TestWidgetsApp(
+///       home: GestureDetector(
+///         onTap: () {},
+///         child: Container(),
+///       ),
+///     ),
+///   );
+/// });
+/// ```
+// TODO(rkishan516): Move this to flutter_test package.
```
justinmc, line 34: Add a link to the issue on the next line: https://github.com/flutter/flutter/issues/181283

packages/flutter/test/widgets/widgets_app_tester_test.dart
justinmc, line 1: Who tests the tester? The tester test tests the tester.

In all seriousness though thank you for adding tests for this.

justinmc (approved):
LGTM with nits :+1: .

packages/flutter/test/widgets/slivers_evil_test.dart
```diff
@@ -255,15 +257,15 @@ void main() {
...
+      TestWidgetsApp(
         home: CustomScrollView(
           cacheExtent: 0.0,
           slivers: <Widget>[
             SliverFixedExtentList(
               itemExtent: 100.0,
               delegate: SliverChildBuilderDelegate((BuildContext context, int index) {
                 if (index > 3) {
-                  return ColoredBox(color: Colors.blue, child: Text(index.toString()));
+                  return ColoredBox(color: const Color(0xFF2196F3), child: Text(index.toString()));
```
victorsanni: Is there a non-trivial reason for a multi-channeled/non-trivial hex value?

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -2005,17 +2012,25 @@ void main() {
   testWidgets('Handle mouse events should ignore the detached MouseTrackerAnnotation', (
     WidgetTester tester,
   ) async {
-    await tester.pumpWidget(
-      MaterialApp(
-        home: Center(
+    final entry = OverlayEntry(
```
victorsanni: Can we abstract this `Overlay`/`OverlayEntry` insertion in `TestWidgetsApp`?

packages/flutter/test/widgets/slivers_evil_test.dart
justinmc: Because it's pretty! Is beauty trivial? :laughing:

victorsanni: It does look fancy, but the extra 20 seconds I spent trying to understand its fanciness might have been better spent fixing the number 1 most upvoted issue in flutter/flutter 😆

victorsanni: I do agree this is an uber nit though, feel free to discard.

justinmc: I'm also fine either way haha. It's true I can't really tell that `2196F3` is blue at a glance, but maybe that doesn't matter.

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: Removed

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: Updated

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: Added

packages/flutter/test/widgets/slivers_evil_test.dart
rkishan516: Updated to single channel, I missed it.

packages/flutter/test/widgets/mouse_region_test.dart
rkishan516: I tried this before, but it was complicating things on both tester and tests side. But again giving a try.

packages/flutter/test/widgets/mouse_region_test.dart
```diff
@@ -2006,14 +2012,20 @@ void main() {
...
     await tester.pumpWidget(
-      MaterialApp(
-        home: Center(
-          child: Draggable<int>(
-            feedback: Container(width: 20, height: 20, color: Colors.blue),
-            childWhenDragging: Container(width: 20, height: 20, color: Colors.yellow),
-            child: ElevatedButton(child: const Text('Drag me'), onPressed: () {}),
+      TestWidgetsApp(
+        home: const SizedBox(),
+        overlayEntries: <WidgetBuilder>[
```
victorsanni: Why do we have this `overlayEntries` API? Since `TestWidgetsApp` has a default `pageRouteBuilder`, can't we directly put this entry in the internal `Overlay` using `OverlayPortal`?

I may have misunderstood the situation here when leaving my previous comment. My understanding of the initial problem was that trying to insert entries using `OverlayEntry` would fail with `WidgetsApp`/`TestWidgetsApp` because of the absence of an `Overlay`. Providing `pageRouteBuilder` to `TestWidgetsApp` forces the widget to create a `Navigator` (which creates an `Overlay`), so its children become eligible ancestors for `OverlayEntry`s.

The reason `Overlay` is relevant here is because internally `Draggable` creates an `OverlayEntry` and tries to insert it into the nearest ancestor `Overlay`. Since the `TestWidgetsApp.pageRouteBuilder` workaround forces the creation of an `Overlay` to insert entries into,  we don't need to create a new `OverlayEntry` for `Draggable`, so the `Center -> DefaultTextStyle -> Draggable...` can just be `TestWidgetsApp.home`.

packages/flutter/test/widgets/widgets_app_tester.dart
```diff
@@ -0,0 +1,119 @@
...
+// Tracking issue: https://github.com/flutter/flutter/issues/181283
+class TestWidgetsApp extends StatefulWidget {
+  /// Creates a minimal [WidgetsApp] for testing.
+  const TestWidgetsApp({
+    super.key,
+    required this.home,
+    this.color = const Color(0xFFFFFFFF),
+    this.overlayEntries,
```
victorsanni: As above, unnecessary API, sorry for the misunderstanding here.

packages/flutter/test/widgets/mouse_region_test.dart
victorsanni: EDIT: Discard the first paragraph, we also don't need `OverlayPortal` here since `Draggable` does that internally.

rkishan516: I have updated test, please have a look.

packages/flutter/test/widgets/widgets_app_tester.dart
rkishan516: Actually my bad, Removed this API.

victorsanni (approved):

PR #180467, opened 2026-01-03
Fix deprecation validator message to require leading v

### LINKED ISSUE #178246: [tool] Deprecations validator truncates its error message right before the required "v" prefix of the deprecation version tag
For https://github.com/flutter/flutter/pull/178242 I was looking to add a new deprecation.
I looked at the docs in https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#deprecations to see how the deprecation message is structured and started working on my PR.

However, when running `./dev/bots/analyze.dart` I noticed that it truncates its error message just before the `v` of the version, which made it less obvious in the error message. I had taken the template from the docs as verbatim, but overlooked a typo.

### Proposal

1) In https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#deprecations
 change the template to the following, since I presume that more people migh take the template as verbatim?

```
@Deprecated(
  '[description of how to migrate] '
  '[brief motivation for why we are breaking the API] '
  'This feature was deprecated after v[beta version at time of deprecation].'
)
```

2) In https://github.com/flutter/flutter/blob/master/dev/bots/analyze.dart#L762-L764 don't truncate the message before the leading "v", as the regex in https://github.com/flutter/flutter/blob/master/dev/bots/analyze.dart#L716-L718 specifically requires it. Perhaps the message should be updated to use a version pattern, instead of three periods?

ashoknarayan:
Fixes https://github.com/flutter/flutter/issues/178246

This PR updates the deprecation validator error message to clarify that
the version must include a leading `v` (for example, `v1.2.3`).

It also updates the deprecation template documentation to match the
validator requirement.

No functional behavior is changed; this only improves clarity and
consistency between the validator and documentation.


Piinks (commented):
Thanks for sending another PR! It looks like this is failing some checks, can you take a look?

Piinks (approved):
LGTM! Thanks for the updates here!

dev/bots/analyze.dart
```diff
@@ -766,8 +766,9 @@ class _DeprecationMessagesVisitor extends RecursiveAstVisitor<void> {
       _addErrorWithLineInfo(
         versionLiteral,
         error:
-            'Deprecation notice must end with a line saying "This feature was deprecated after...".',
+            'Deprecation notice must end with a line saying "This feature was deprecated after v<version> ...".',
```
navaronbracke: This is going to need a test, to verify that the version is not cut off.

See https://github.com/flutter/flutter/blob/master/dev/bots/test/analyze_test.dart#L70-L85
which is the existing test. You will have to update the expectation, to also check if it contains the line

`Deprecation notice must end with a line saying "This feature was deprecated after v<version> ..."`

dev/bots/analyze.dart
```diff
@@ -766,8 +766,9 @@ class _DeprecationMessagesVisitor extends RecursiveAstVisitor<void> {
...
       );
+
```
navaronbracke: Nit: this newline can be removed?

navaronbracke (changes requested):
Just a suggestion on how to test this.

dev/bots/analyze.dart
ashoknarayan: I updated the expectation in this [commit](https://github.com/flutter/flutter/pull/180467/changes/d74755575110c4f7b22b5a59b016dce9a3e33eb2)

dev/bots/analyze.dart
ashoknarayan: I removed the newline in this [commit](https://github.com/flutter/flutter/pull/180467/changes/36d909d1b39a28dadb8dce650cf69ac38c738240)

dev/bots/test/analyze_test.dart
```diff
@@ -78,6 +78,7 @@ void main() {
       matchesErrorsInFile(
         fixture,
         endsWith: <String>[
+          'Deprecation notice must end with a line saying "This feature was deprecated after v<version> ...".',
```
navaronbracke: Apologies, it looks like this is not the right way to test it, because I didn't yet understand the test :(

This test makes use of a fixture file to define the deprecation message tests. The `endsWith` argument here _builds on top of that_. So instead of modifying this file, you need to add a test case in the fixture file at `dev/bots/test/analyze-test-input/root/packages/foo/deprecation.dart`.

You can apply the following git patch to update the fixture file. It updates the `test()` functions, so that the indexes are incremented, and adds some test cases for the fix here. (including the adjustment for the period at the end)

```
diff --git a/dev/bots/test/analyze-test-input/root/packages/foo/deprecation.dart b/dev/bots/test/analyze-test-input/root/packages/foo/deprecation.dart
index cf20a1a6465..252d512c8d6 100644
--- a/dev/bots/test/analyze-test-input/root/packages/foo/deprecation.dart
+++ b/dev/bots/test/analyze-test-input/root/packages/foo/deprecation.dart
@@ -59,39 +59,51 @@ void test9() { }
 )
 void test10() { }
 
+@Deprecated(
+  'Version number test (should fail). '
+  'This feature was deprecated after 1.0.0.' // ERROR: Deprecation notice must end with a line saying "This feature was deprecated after v<version>.".
+)
+void test11() {}
+
+@Deprecated(
+  'Version number test (should fail). '
+  'This feature was deprecated after v1.0.0' // ERROR: Deprecation notice must end with a line saying "This feature was deprecated after v<version>.".
+)
+void test12() {}
+
 @Deprecated(
   'Version number test (special beta should pass). '
   'This feature was deprecated after v3.1.0.'
 )
-void test11() { }
+void test13() { }
 
 @Deprecated(
   'Version number test (should be fine). '
   'This feature was deprecated after v0.1.0.'
 )
-void test12() { }
+void test14() { }
 
 @Deprecated(
   'Version number test (should be fine). '
   'This feature was deprecated after v1.20.0-1.0.pre.'
 )
-void test13() { }
+void test15() { }
 
 @Deprecated(
   "Double quotes' test (should fail). " // ERROR: Deprecation notice does not match required pattern. You might have used double quotes (") for the string instead of single quotes (').
   'This feature was deprecated after v2.1.0-11.0.pre.'
 )
-void test14() { }
+void test16() { }
 
 @Deprecated( // flutter_ignore: deprecation_syntax, https://github.com/flutter/flutter/issues/000000
   'Missing the version line. '
 )
-void test15() { }
+void test17() { }
 // dart format on
 
 // flutter_ignore: deprecation_syntax, https://github.com/flutter/flutter/issues/000000
 @Deprecated('Missing the version line. ')
-void test16() {}
+void test18() {}
 
 class TestClass1 {
   // flutter_ignore: deprecation_syntax, https://github.com/flutter/flutter/issues/000000
```

dev/bots/analyze.dart
```diff
@@ -766,7 +766,7 @@ class _DeprecationMessagesVisitor extends RecursiveAstVisitor<void> {
       _addErrorWithLineInfo(
         versionLiteral,
         error:
-            'Deprecation notice must end with a line saying "This feature was deprecated after...".',
+            'Deprecation notice must end with a line saying "This feature was deprecated after v<version> ...".',
```
navaronbracke: Just a suggestion, but I see that the deprecation lines all end with the version, followed by a single period. Maybe we can verify that, instead of the ellipsis.

```suggestion
            'Deprecation notice must end with a line saying "This feature was deprecated after v<version>.".',
```

navaronbracke (changes requested):

dev/bots/test/analyze_test.dart
ashoknarayan: Have updated the deprecator.dart file

navaronbracke (approved):

navaronbracke:
The tree is currently red, but this is good to land as-is, once it becomes green again.

PR #180469, opened 2026-01-03
Improve accessibility contrast for pre-test message

ashoknarayan:
Fixes https://github.com/flutter/flutter/issues/179181

The pre-test message "Test starting..." previously had a contrast ratio
of 2.99:1 against the background, which is below the recommended 3:1
minimum for accessibility.

This PR slightly adjusts the text color from 0xFF917FFF to 0xFF8F7FFF
to meet the minimum contrast requirement while preserving the visual
appearance.


Piinks (commented):
Hey @ashoknarayan thanks for contributing! This kind of change should probably have a [meetsGuideline](https://api.flutter.dev/flutter/flutter_test/meetsGuideline.html) test and use the [text contrast guideline](https://api.flutter.dev/flutter/flutter_test/textContrastGuideline-constant.html) to ensure we do not regress this in a future change.

But first, can you provide more context? Is there an open issue? I am also curious about what background you are referring to. Thanks!

packages/flutter_test/lib/src/binding.dart
```diff
@@ -1265,7 +1265,7 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
     return announcements;
   }
 
-  static const TextStyle _messageStyle = TextStyle(color: Color(0xFF917FFF), fontSize: 40.0);
+  static const TextStyle _messageStyle = TextStyle(color: Color(0xFF8F7FFF), fontSize: 40.0);
```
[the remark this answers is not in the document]
ashoknarayan: Proceed

packages/flutter_test/test/pre_test_message_accessibility_test.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:flutter/widgets.dart';
```
Piinks: This file requires the license header.

ashoknarayan: I have added the comments that I found in the start of every test file.

packages/flutter_test/test/pre_test_message_accessibility_test.dart
```diff
@@ -0,0 +1,28 @@
...
+
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  /// Verifies that the pre-test message shown by flutter_test
+  /// meets the minimum WCAG text contrast accessibility guideline.
+  testWidgets('pre-test message meets text contrast guideline', (WidgetTester tester) async {
+    // Pump the pre-test message widget.
+    await tester.pumpWidget(
```
chunhtai: I think you don't need to pump widget to see the pre test text, it is already attached to the binding before this test body is run

ashoknarayan: The pre-test message is already attached to the render tree by `TestWidgetsFlutterBinding` before the test body runs. I’ve removed the manual `pumpWidget` and now validate the existing render tree directly using `meetsGuideline(textContrastGuideline)`.

This should ensure we’re testing the real pre-test message and guarding against regressions.

Piinks (approved):
LGTM

chunhtai (approved):
LGTM

PR #180488, opened 2026-01-04
Draft: Multiview iOS

littleGnAl:
Design doc: http://flutter.dev/go/multi-view-on-android-ios

https://github.com/user-attachments/assets/d8162c18-fd3d-4b5a-8f02-c1e39c1e6429

## Pre-launch Checklist

- [ ] I signed the [CLA].


engine/src/flutter/lib/ui/platform_dispatcher.dart
```diff
@@ -280,7 +280,6 @@ class PlatformDispatcher {
   // The implicit view must be added before [implicitView] is first called,
   // which is typically the main function.
   void _addView(int id, _ViewConfiguration viewConfiguration) {
-    assert(!_views.containsKey(id), 'View ID $id already exists.');
```
dkwingsmt, line 289: Why did you make these changes in the `ui/*.dart`? None of them seems reasonable to me. The Dart API should not need any changes.

engine/src/flutter/shell/common/shell.cc
```diff
@@ -1828,9 +1828,10 @@ fml::TimePoint Shell::GetLatestFrameTargetTime() const {
 bool Shell::ShouldDiscardLayerTree(int64_t view_id,
                                    const flutter::LayerTree& tree) {
   std::scoped_lock<std::mutex> lock(resize_mutex_);
+  bool has_rendering_surface = platform_view_->HasRenderingSurface(view_id);
```
dkwingsmt: What's this for? 

The source of truth of list of views are in the engine, and therefore I don't think we should ever query the Dart side about the list of active rendering surfaces.

engine/src/flutter/shell/common/shell.cc
```diff
@@ -2271,9 +2272,6 @@ void Shell::OnPlatformViewRemoveView(int64_t view_id,
   TRACE_EVENT0("flutter", "Shell::RemoveView");
   FML_DCHECK(is_set_up_);
   FML_DCHECK(task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread());
-  FML_DCHECK(view_id != kFlutterImplicitViewId)
```
dkwingsmt, line 2285: I think this change is not reasonable. It's intentional that implicit views are never removed (or added).

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -163,6 +163,8 @@ FLUTTER_DARWIN_EXPORT
       allowHeadlessExecution:(BOOL)allowHeadlessExecution
           restorationEnabled:(BOOL)restorationEnabled NS_DESIGNATED_INITIALIZER;
 
+- (void)enableMultiView;
```
dkwingsmt: Changing public APIs in this PR takes a longer reviewing process and it might be easier for you to do it in a separate API. If you're hoping to keep it for demonstration purposes, a compromise is to create a new category interface like the below in a private header. (I'm not sure how reviewers from the iOS team will see this, although it is something that I'd be willing to try if I don't want to remove this part of code.)
```
@interface FlutterEngine (MultiView)
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -109,6 +109,7 @@ @interface FlutterEnginePluginRegistrar : FlutterEngineBaseRegistrar <FlutterPlu
 
 @interface FlutterEngine () <FlutterIndirectScribbleDelegate,
                              FlutterUndoManagerDelegate,
+                             FlutterTextInputPluginDelegate,
```
dkwingsmt, line 112: Can you explain this?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -198,6 +219,13 @@ @implementation FlutterEngine {
 
   FlutterBinaryMessengerRelay* _binaryMessenger;
   FlutterTextureRegistryRelay* _textureRegistry;
+
+  // It can't use NSDictionary, because the values need to be weak references.
```
dkwingsmt:
```suggestion
  // Maps from view IDs to view controllers registered to this engine.
  //
  // The view controllers are kept as weak references, since they are managed by the user app.
  // Therefore this table must be a `NSMapTable`.
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -511,50 +552,148 @@ - (void)ensureSemanticsEnabled {
...
+    // From non-nil to nil.
+    [self deregisterViewControllerForIdentifier:flutter::kFlutterImplicitViewId];
   } else {
-    self.flutterViewControllerWillDeallocObserver = nil;
-    [self notifyLowMemory];
+    // From non-nil to a different non-nil view controller.
+    if (currentController) {
+      NSString* errorMessage =
+          [NSString stringWithFormat:
+                        @"The supplied FlutterEngine %@ is already used with FlutterViewController "
```
dkwingsmt: So this is why I recommend delaying the public API changes to another PR. This part of the changes is not compatible with the current behavior. The current version allows swapping view controllers without setting to null in between, which is probably used in many apps already and must not be changed without thorough research and discussion.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -175,7 +179,8 @@ - (void)performSubmit:(const LayersMap&)platformViewLayers
                 unusedLayers:
                     (const std::vector<std::shared_ptr<flutter::OverlayLayer>>&)unusedLayers
                surfaceFrames:
-                   (const std::vector<std::unique_ptr<flutter::SurfaceFrame>>&)surfaceFrames;
+                   (const std::vector<std::unique_ptr<flutter::SurfaceFrame>>&)surfaceFrames
+        currentFlutterViewId:(int64_t)currentFlutterViewId;
```
dkwingsmt, line 275: Does a `FlutterPlatformViewsController` only interact with one Flutter view (the `currentFlutterView`) at a time? If so, can we either extract the state related to the current flutter view to another class (maybe called `FlutterComposingSession`) or add a `Prepare` method that records the current view?

dkwingsmt (commented):
I've only reviewed the first half of the PR.

I strongly recommend delaying the changes related to the iOS public API `Headers/*.h`. There are enough changes for a PR to adapt the engine implementation to multiview. The public API changes would require a much longer discussion with complicated research.

engine/src/flutter/shell/common/shell.cc
littleGnAl: The implicit view (`kFlutterImplicitViewId`) is now assumed to never be removed. However, this is invalid in an add-to-app scenario on Android/iOS, as the user can open the `FlutterActivity` or `FlutterViewController` with `kFlutterImplicitViewId` and then close it, the implicit view in framework side should be removed too, or it would be leaked.

engine/src/flutter/shell/common/shell.cc
littleGnAl: This is intended to skip rendering for an inactive view.

For example, when we push two full-screen `FlutterViewController`s (fvc1, fvc2), fvc1 moves to the background. However, on the framework side there are still two active views, which results in two `LayerTree`s (layerTree1, layerTree2) being generated.
We do not expect layerTree1 to continue rendering to the inactive `FlutterViewController` fvc1.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
littleGnAl: Agree, I will create a design doc for this part.

engine/src/flutter/lib/ui/platform_dispatcher.dart
littleGnAl: The implicit view (`kFlutterImplicitViewId`) is now assumed to never be removed. However, this is invalid in an add-to-app scenario on Android/iOS, as the user can open the `FlutterActivity` or `FlutterViewController` with `kFlutterImplicitViewId` and then close it, the implicit view in framework side should be removed too, or it would be leaked.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
littleGnAl: This is borrowed from the macOS implementation, which lets the `FlutterTextInputPlugin` get the current `FlutterViewController` by view ID. You can find the protocol declaration in `FlutterTextInputPlugin.h`.

engine/src/flutter/lib/ui/platform_dispatcher.dart
littleGnAl: @dkwingsmt  I reverted this part to keep it the same as the current implementation.

engine/src/flutter/shell/common/shell.cc
littleGnAl: @dkwingsmt  I reverted this part to keep it the same as the current implementation.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
littleGnAl: @dkwingsmt I created a design doc to discuss this part [flutter.dev/go/multi-view-on-android-ios-runner-api](https://flutter.dev/go/multi-view-on-android-ios-runner-api)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -131,7 +133,8 @@ @interface FlutterEngine () <FlutterIndirectScribbleDelegate,
 
 @property(nonatomic, readwrite, copy) NSString* isolateId;
 @property(nonatomic, copy) NSString* initialRoute;
-@property(nonatomic, strong) id<NSObject> flutterViewControllerWillDeallocObserver;
+// @property(nonatomic, strong) id<NSObject> flutterViewControllerWillDeallocObserver;
```
dkwingsmt:
```suggestion
```

engine/src/flutter/shell/common/shell.cc
dkwingsmt: I think we talked about trusting the list of active layer trees, and so this should no longer be needed?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/accessibility_bridge.mm
```diff
@@ -54,7 +54,7 @@ void PostAccessibilityNotification(UIAccessibilityNotifications notification,
       weak_factory_(this) {
   accessibility_channel_ = [[FlutterBasicMessageChannel alloc]
          initWithName:@"flutter/accessibility"
-      binaryMessenger:platform_view->GetOwnerViewController().engine.binaryMessenger
+      binaryMessenger:view_controller_.engine.binaryMessenger
```
vashworth, line 57: Why is this changed? Will the `view_controller` and `platform_view->GetOwnerViewController()` be different at any point?

engine/src/flutter/shell/common/platform_view.h
```diff
@@ -1029,6 +1029,8 @@ class PlatformView {
   ///             Impeller context creation can be moved off the startup path.
   virtual void SetupImpellerContext() {}
 
+  virtual bool HasRenderingSurface(int64_t flutter_view_id) { return true; }
```
vashworth: Please add a doc comment

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.h
```diff
@@ -131,6 +133,15 @@ NS_ASSUME_NONNULL_BEGIN
 /// view
 - (void)pushClipPathToVisitedPlatformViews:(const flutter::DlPath&)clipPath;
 
+- (void)collectView:(int64_t)flutterViewId;
```
vashworth, line 136: Please add doc comments

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -519,52 +566,139 @@ - (void)ensureSemanticsEnabled {
...
-      [[NSNotificationCenter defaultCenter]
-          removeObserver:_flutterViewControllerWillDeallocObserver];
+- (void)registerViewController:(FlutterViewController*)controller
+                 forIdentifier:(FlutterViewIdentifier)viewIdentifier {
+  id<NSObject> existingObserver =
+      [self.flutterViewControllerWillDeallocObservers objectForKey:@(viewIdentifier)];
+  if (existingObserver) {
+    [[NSNotificationCenter defaultCenter] removeObserver:existingObserver];
+    [self.flutterViewControllerWillDeallocObservers removeObjectForKey:@(viewIdentifier)];
+  }
```
vashworth: This should be moved to a method

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth: Could this be combined with the existing `FlutterTextInputDelegate` delegate? Both delegates are essentially the same `FlutterEngine` and I don't think there's a situation where we would want them to be different engines. It just seems redundant to pass in two delegates, which are actually the same object. Also, the naming is so similar it's confusing on what's the purpose of the `FlutterTextInputPluginDelegate`.

```

self.textInputPlugin = [[FlutterTextInputPlugin alloc] initWithDelegate:self
                                                        textInputDelegate:self];
...
- (instancetype)initWithDelegate:(id<FlutterTextInputPluginDelegate>)delegate
               textInputDelegate:(id<FlutterTextInputDelegate>)textInputDelegate {
```

engine/src/flutter/shell/platform/darwin/ios/ios_external_view_embedder.mm
```diff
@@ -13,18 +21,19 @@
...
 }
 
 IOSExternalViewEmbedder::~IOSExternalViewEmbedder() = default;
 
 // |ExternalViewEmbedder|
 DlCanvas* IOSExternalViewEmbedder::GetRootCanvas() {
-  // On iOS, the root surface is created from the on-screen render target. Only the surfaces for the
-  // various overlays are controlled by this class.
-  return nullptr;
+  return pending_frame_->Canvas();
```
vashworth, line 36: Can you explain why this changed?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -244,6 +276,8 @@ - (instancetype)initWithName:(NSString*)labelPrefix
   _allowHeadlessExecution = allowHeadlessExecution;
   _labelPrefix = [labelPrefix copy];
   _dartProject = project ?: [[FlutterDartProject alloc] init];
+  _viewControllers = [NSMapTable weakToWeakObjectsMapTable];
```
vashworth, line 282: `weakToWeakObjectsMapTable`'s [documentation](https://developer.apple.com/documentation/foundation/nsmaptable/weaktoweakobjects()?changes=latest_ma_10_8_8&language=objc) states:

> A new map table object which has weak references to the keys and values.

When adding an object like this, wouldn't the key get garbage collected since it's a weak reference?
```
[_viewControllers setObject:controller forKey:@(viewIdentifier)];
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
vashworth, line 815: What's the ramifications on AppLifecycleState when using multiple views?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
```diff
@@ -1173,14 +1319,79 @@ - (void)pushClipPathToVisitedPlatformViews:(const flutter::DlPath&)clipPath {
...
+
+- (UIView*)flutterViewForIdentifier:(int64_t)flutterViewId {
+  FlutterViewController* flutterViewController =
+      [self.flutterViewControllers objectForKey:@(flutterViewId)];
+  UIView* flutterView = flutterViewController.view;
+  if (flutterView != nil) {
+    return flutterView;
+  }
+  if (flutterViewId == flutter::kFlutterImplicitViewId) {
+    return self.flutterView;
```
vashworth: Where does this get set? Looks like it's always nil?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
vashworth, line 1841: What is the plan for plugins that use the viewController?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/accessibility_bridge.mm
littleGnAl: In multi-view implementation, the `AccessibilityBridge` is now created per `FlutterViewController`, while `platform_view->GetOwnerViewController()` still represents the implicit owner controller.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
littleGnAl: The `FlutterTextInputPluginDelegate` only exposes a method to get a view controller by view ID. Combining it with the existing `FlutterTextInputDelegate` seems like a really good idea; I’ll try to do that.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
littleGnAl: I copied this from the macOS engine implementation: https://github.com/flutter/flutter/blob/ad80825c24d770a19e33f67800fc0338a3b89ec7/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm#L555

The reason for using a weak map here is that the engine shouldn’t keep the view controllers alive. The keys are just boxed view ids, and with the current small monotonically assigned ids they end up as tagged `NSNumber`s, so this shouldn’t disappear in practice. That said, I agree the ownership would be clearer with strong keys and weak values. I’m fine changing it if that reads better.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
littleGnAl: Makes sense. This cleanup is used by both registration and deregistration paths, so I’ll move it into a small helper method and call that from both places.

engine/src/flutter/shell/platform/darwin/ios/ios_external_view_embedder.mm
littleGnAl: In the multi-view path, `PrepareFlutterView` now creates the pending `SurfaceFrame` for the target platform surface. `GetRootCanvas` should return that frame’s canvas so the `LayerTree` content is rendered into the correct platform surface. The pending frame is then submitted via `submitFrame:...withFlutterViewId:`.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
littleGnAl: I agree passing two delegates here is confusing, especially since they are both the engine.

I’d rather not add `viewControllerForIdentifier:` directly to `FlutterTextInputDelegate` though, since that protocol is mostly for text input callbacks. I’m going to add a small lookup protocol instead:

```objc
@protocol FlutterViewControllerLookupDelegate <NSObject>
- (nullable FlutterViewController*)viewControllerForIdentifier:
    (FlutterViewIdentifier)viewIdentifier;
@end
```

Then `FlutterTextInputPlugin` can just take one delegate that conforms to both protocols:

```objc
- (instancetype)initWithDelegate:
    (id<FlutterTextInputDelegate, FlutterViewControllerLookupDelegate>)delegate;
```

and the engine still passes itself once:

```objc
self.textInputPlugin = [[FlutterTextInputPlugin alloc] initWithDelegate:self];
```

That should remove the redundant delegate parameter without mixing view-controller lookup into the text input callback protocol. Does that sound reasonable to you?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsController.mm
littleGnAl: Good catch. This was leftover from the old single-view path. `flutterView` is no longer set by production code after switching to the controller map, so the fallback could never provide a non-nil view. I removed the property and now resolve the FlutterView only through the registered FlutterViewController.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
littleGnAl: The lifecycle channel is still engine-scoped, not view-scoped. With a single engine used by multiple `FlutterViewController`s, each controller can currently send lifecycle messages as it appears or disappears, so the framework may receive multiple `AppLifecycleState` updates for the same engine.

That is an existing limitation of the single-engine add-to-app scenario, and this PR preserves that behavior. Fixing it properly would require defining how lifecycle state should be aggregated across all attached views, which would significantly expand the scope of this multi-view change. I’d prefer to track that separately rather than solve it in this PR.

dkwingsmt:
I misclicked Ready for review. Sorry about that.

Update: I'm drafting a design doc for this PR.

PR #180494, opened 2026-01-05
Add `TestTextField` and migrate tests

Renzo-Olivares:
This PR introduces the convenience widget `TestTextField` in the framework widget tests. This new widget should be used in place of convenient `TextField` or `CupertinoTextField` usages in the widgets library tests.

part of: #177415

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/text_selection.dart
+++ b/packages/flutter/lib/src/widgets/text_selection.dart
@@ -2056,16 +2056,21 @@ class _SelectionHandleOverlayState extends State<_SelectionHandleOverlay>
   Widget build(BuildContext context) {
     final Rect handleRect = _getHandleRect(widget.type, widget.preferredLineHeight);
 
-    // Make sure the GestureDetector is big enough to be easily interactive.
-    final Rect interactiveRect = handleRect.expandToInclude(
-      Rect.fromCircle(center: handleRect.center, radius: kMinInteractiveDimension / 2),
-    );
-    final padding = RelativeRect.fromLTRB(
-      math.max((interactiveRect.width - handleRect.width) / 2, 0),
-      math.max((interactiveRect.height - handleRect.height) / 2, 0),
-      math.max((interactiveRect.width - handleRect.width) / 2, 0),
-      math.max((interactiveRect.height - handleRect.height) / 2, 0),
-    );
+    // Make sure the GestureDetector is big enough to be easily interactive if
+    // the handleRect is not empty.
+    final Rect interactiveRect = handleRect.isEmpty
+        ? handleRect
+        : handleRect.expandToInclude(
+            Rect.fromCircle(center: handleRect.center, radius: kMinInteractiveDimension / 2),
+          );
+    final RelativeRect padding = interactiveRect.isEmpty
+        ? RelativeRect.fill
+        : RelativeRect.fromLTRB(
+            math.max((interactiveRect.width - handleRect.width) / 2, 0),
+            math.max((interactiveRect.height - handleRect.height) / 2, 0),
+            math.max((interactiveRect.width - handleRect.width) / 2, 0),
+            math.max((interactiveRect.height - handleRect.height) / 2, 0),
+          );
 
     final Offset handleAnchor = widget.selectionControls.getHandleAnchor(
       widget.type,
--- a/packages/flutter/test/material/text_field_test.dart
+++ b/packages/flutter/test/material/text_field_test.dart
@@ -19113,6 +19113,36 @@ void main() {
 
     expect(controller.text, 'H');
   });
+
+  testWidgets('Cursor does not show when not focused', (WidgetTester tester) async {
+    // Regression test for https://github.com/flutter/flutter/issues/106512 .
+    final focusNode = FocusNode();
+    addTearDown(focusNode.dispose);
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Material(child: TextField(focusNode: focusNode, autofocus: true)),
+      ),
+    );
+    assert(focusNode.hasFocus);
+    final EditableTextState editableTextState = tester.firstState(find.byType(EditableText));
+    final RenderEditable renderEditable = editableTextState.renderEditable;
+
+    focusNode.unfocus();
+    await tester.pump();
+
+    for (var i = 0; i < 10; i += 10) {
+      // Make sure it does not paint for a period of time.
+      expect(renderEditable, paintsExactlyCountTimes(#drawRect, 0));
+      expect(tester.hasRunningAnimations, isFalse);
+      await tester.pump(const Duration(milliseconds: 29));
+    }
+
+    // Refocus and it should paint the caret.
+    focusNode.requestFocus();
+    await tester.pump();
+    await tester.pump(const Duration(milliseconds: 100));
+    expect(renderEditable, isNot(paintsExactlyCountTimes(#drawRect, 0)));
+  });
 }
 
 /// A Simple widget for testing the obscure text.
--- a/packages/flutter/test/material/text_field_test.dart
+++ b/packages/flutter/test/material/text_field_test.dart
@@ -19069,6 +19069,50 @@ void main() {
     // [intended] only applies to platforms where we supply the context menu.
     skip: kIsWeb,
   );
+
+  testWidgets('entering text does not scroll a surrounding PageView', (WidgetTester tester) async {
+    // regression test for https://github.com/flutter/flutter/issues/19523
+    final pageController = PageController(initialPage: 1);
+    final controller = TextEditingController();
+    addTearDown(pageController.dispose);
+    addTearDown(controller.dispose);
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: MediaQuery(
+          data: const MediaQueryData(),
+          child: Directionality(
+            textDirection: TextDirection.ltr,
+            child: Material(
+              child: PageView(
+                controller: pageController,
+                children: <Widget>[
+                  Container(color: Colors.red),
+                  ColoredBox(
+                    color: Colors.green,
+                    child: TextField(controller: controller),
+                  ),
+                  Container(color: Colors.red),
+                ],
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+
+    await tester.showKeyboard(find.byType(EditableText));
+    await tester.pumpAndSettle();
+    expect(controller.text, '');
+    tester.testTextInput.enterText('H');
+    final int frames = await tester.pumpAndSettle();
+
+    // The text input should not trigger any animations, which would indicate
+    // that the surrounding PageView is incorrectly scrolling back-and-forth.
+    expect(frames, 1);
+
+    expect(controller.text, 'H');
+  });
 }
 
 /// A Simple widget for testing the obscure text.
--- a/packages/flutter/test/material/text_field_test.dart
+++ b/packages/flutter/test/material/text_field_test.dart
@@ -19143,6 +19143,174 @@ void main() {
     await tester.pump(const Duration(milliseconds: 100));
     expect(renderEditable, isNot(paintsExactlyCountTimes(#drawRect, 0)));
   });
+
+  // Regression test for https://github.com/flutter/flutter/issues/175511.
+  testWidgets('Radio group does not intercept key events when no radio is focused', (
+    WidgetTester tester,
+  ) async {
+    final log = <String>[];
+    late final shortcuts = <ShortcutActivator, Intent>{
+      const SingleActivator(LogicalKeyboardKey.arrowLeft): VoidCallbackIntent(() => log.add('←')),
+      const SingleActivator(LogicalKeyboardKey.arrowRight): VoidCallbackIntent(() => log.add('→')),
+      const SingleActivator(LogicalKeyboardKey.arrowDown): VoidCallbackIntent(() => log.add('↓')),
+      const SingleActivator(LogicalKeyboardKey.arrowUp): VoidCallbackIntent(() => log.add('↑')),
+      const SingleActivator(LogicalKeyboardKey.space): VoidCallbackIntent(() => log.add('_')),
+    };
+
+    final firstRadioFocusNode = FocusNode();
+    addTearDown(firstRadioFocusNode.dispose);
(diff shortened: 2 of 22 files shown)


packages/flutter/lib/src/widgets/text_selection.dart
```diff
@@ -2056,16 +2056,21 @@ class _SelectionHandleOverlayState extends State<_SelectionHandleOverlay>
...
-    );
-    final padding = RelativeRect.fromLTRB(
-      math.max((interactiveRect.width - handleRect.width) / 2, 0),
-      math.max((interactiveRect.height - handleRect.height) / 2, 0),
-      math.max((interactiveRect.width - handleRect.width) / 2, 0),
-      math.max((interactiveRect.height - handleRect.height) / 2, 0),
-    );
+    // Make sure the GestureDetector is big enough to be easily interactive if
+    // the handleRect is not empty.
+    final Rect interactiveRect = handleRect.isEmpty
```
Renzo-Olivares, line 2061: This change was needed because https://github.com/flutter/flutter/blob/96403e0fa5704dacd8fee2509b9333a86b3c7fed/packages/flutter/test/widgets/text_selection_test.dart#L1232 was failing due to the hit rect on the selection handle always expanding even when the rect of the handle is empty.

packages/flutter/test/widgets/editable_text_utils.dart
```diff
@@ -140,3 +140,153 @@ class OverflowWidgetTextEditingController extends TextEditingController {
...
 }
+
+/// A minimal wrapper around [EditableText] for use in widget tests.
+///
+/// This wrapper provides robust text selection gestures through
+/// [TextSelectionGestureDetector].
+///
+/// This wrapper does not provide selection handles or context menus
+/// out of the box.
+class TestTextField extends StatefulWidget {
```
Renzo-Olivares, line 161: Should this live here/a separate file/or in flutter_test? I mention `flutter_test` because I don't think we generate dart docs for any classes in our widget tests and those may be helpful.

packages/flutter/test/widgets/editable_text_cursor_test.dart
```diff
@@ -185,9 +185,9 @@ void main() {
   testWidgets(
     'Cursor animates on iOS',
     (WidgetTester tester) async {
-      await tester.pumpWidget(const MaterialApp(home: Material(child: TextField())));
```
Renzo-Olivares, line 188: I wonder if tests like these (lots in this specific file) are good candidates to be duplicated in design libraries since they test the defaults set internally by `TextField` and not `EditableText`. or is it enough that we are setting the same defaults in `TestTextField`. What do others think?

Renzo-Olivares:
This is ready for an initial review. The way I went about this is I went through all the usages of `TextField` in the widgets library and lazily added parameters to `TestTextField` as needed. I also added robust text selection gestures through `TextSelectionGestureDetector`, matching the same way `TextField` and `CupertinoTextField` implement their gestures (note, there weren't many tests actually using these gestures).

The main indicator to whether I should duplicate a test was whether it was a regression test. There was only a handful I duplicated and added to `text_field_test.dart`. All of the bugs the regression tests were targeting where reported on `TextField` specifically so I did not duplicate the the tests for `CupertinoTextField` but can do so if we think that is necessary. There are still certain tests I have noted that might deserve a closer look with regards to duplicating across design libraries, would like to hear others thoughts on these.

Another question is whether we'd this `TestTextField` and other similar helper test widgets to be discoverable through our online docs. If so, maybe we want to consider placing them a different location, maybe `flutter_test`?

There are still ~5 tests that have not been migrated to use `TestTextField` in this PR, that will require a closer look.

packages/flutter/test/widgets/editable_text_utils.dart
Renzo-Olivares: Are doc macros usable here?

loic-sharma: I think your current location is a very good starting point. It lets us prove the API, solves our immediate need, and gives us maximum flexibility.

My suspicion is `TestTextField` in its current shape is most helpful for the widget library's tests, which are weird in that they cannot depend on a design system library. Other design system libraries can use their own text fields in tests directly, and might not need `TestTextField` at all.

Nonetheless, moving this to `flutter_test` seems very reasonable, but once we do so the API is locked, and our ability to do breaking changes is limited. With your current approach, we're free to move this to `flutter_test` whenever we'd like.

What do you think of this: design this as if it's in `flutter_test`, but for now keep this here?

(I don't a strong preference on whether to split this into its own file, though the current location seems very reasonable to me 👍)

packages/flutter/test/material/text_field_test.dart
```diff
@@ -19069,6 +19069,248 @@ void main() {
     // [intended] only applies to platforms where we supply the context menu.
     skip: kIsWeb,
   );
+
+  testWidgets('entering text does not scroll a surrounding PageView', (WidgetTester tester) async {
```
loic-sharma, line 19073: It's possible this test will drift over time with the corresponding one in the widgets layer. Should we add comments to keep these tests in sync? I don't feel strongly, just a thought.

packages/flutter/test/widgets/autocomplete_test.dart
```diff
@@ -2704,7 +2706,11 @@ void main() {
 
       // Shrink the screen further so that the options become smaller than
       // kMinInteractiveDimension and move to overlap the field.
-      const shortWindowSize = Size(1920.0, 90.0);
+      const shortWindowSize = Size(
+        1920.0,
+        48.0,
+      ); //TODO(Renzo-Olivares): Should a test like this be in material,cupertino, and widgets? The shortWindowSize varies depending
+      // on the input field because of additional padding by decorators.
```
loic-sharma: Don't forget to remove this TODO depending on the result of this conversation: https://github.com/flutter/flutter/pull/180494#discussion_r2677516714

packages/flutter/test/material/text_field_test.dart
loic-sharma: Mulling on this some more. Considering these tests will eventually move to packages in another repository, keeping these in sync will be painful. Perhaps it's not even worth attempting (and thus not worth adding these comments), but I'll leave this open in case others have more thoughts.

packages/flutter/test/widgets/editable_text_cursor_test.dart
loic-sharma: My immediate though is to lean towards duplicating the tests.

My gut feeling is that we should assume `TestTextField` might not have high fidelity with `TextField`/`CupertinoTextField`, even if it does today. Perhaps it's best to treat `TestTextField` as if it's a text field from a completely different design system?

I'll defer to @justinmc here though for when they're back :)

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -17176,18 +17177,73 @@ void main() {
...
     );
+
     await tester.pumpWidget(widget);
+    expect(find.byType(EditableText), findsNothing);
     await tester.showKeyboard(find.byType(EditableText, skipOffstage: false));
     await tester.pumpAndSettle();
-    expect(scrollController.offset, 75.0);
+    expect(find.byType(EditableText), findsOneWidget);
+    // TODO(Renzo-Olivares): Decide if the below calculation is worthwhile to verify
+    // or does simply verifying that EditableText is on screen suffice.
```
loic-sharma: I'm good with the keeping the calculations below if that's what you choose to do 👍

packages/flutter/test/widgets/editable_text_utils.dart
```diff
@@ -140,3 +140,180 @@ class OverflowWidgetTextEditingController extends TextEditingController {
...
+    return Offset.zero;
+  }
+
+  @override
+  Size getHandleSize(double textLineHeight) {
+    return Size.zero;
+  }
+}
+
+final TextSelectionControls testTextSelectionHandleControls = TestTextSelectionHandleControls();
```
loic-sharma, line 298: Just curious, did you consider a static field named `instance` on `TestTextSelectionHandleControls`? That might be a little cleaner than a top-level global, but I don't feel strongly.

loic-sharma (approved):
This looks great, excellent work!

packages/flutter/test/widgets/editable_text_utils.dart
Renzo-Olivares: Good point, I also think `TestTextField` will be most useful in our own widget tests. Designing as if it's in `flutter_test` sgtm, these helper widgets should definitely be clearly documented so users are clear what to expect when writing their own tests.

packages/flutter/test/material/text_field_test.dart
justinmc: I think in general we shouldn't need duplicate tests like this at all. If someone reports a bug in TextField, and it reproduces in EditableText, I think it's ok to fix it and test it only in EditableText. If the fix requires a change in TextField too, or if there's some other specific concern with TextField, only then we should test that in TextField too, IMO.

In this case I'm fine either way... It kind of makes sense to duplicate this test here because without it, we are losing test coverage that may have had some effect over the years. Keeping it in sync with its Widgets counterpart is probably not necessary, though.

packages/flutter/test/widgets/editable_text_cursor_test.dart
justinmc: Long term vision: Stuff like this is system UI, and it should be in Widgets where possible. Then it could be tested in Widgets only.

If the long term vision isn't possible right now, then I would move this test to Material, because the logic that it's testing is in Material (for now at least). Any duplication of that logic in TestTextField is coincidental and doesn't need to be tested anyway.

packages/flutter/test/widgets/autocomplete_test.dart
justinmc: Expanding from that conversation, I think here it's probably fine just in Widgets. The slight differences in shortWindowSize are probably irrelevant to the overall test. If there were specific concerns about, say, TextField's behavior in a short window due to its padding, then we could test that specific case in Material too.

packages/flutter/test/widgets/editable_text_test.dart
justinmc: I appreciate all the thorough math here! :+1:

justinmc (approved):
LGTM :+1:. Thank you @Renzo-Olivares for doing this, it's really a big help for these tests!

After writing this PR, do you have any more thoughts on whether or not some kind of widget like TestTextField might be useful for developers using only the Widgets library? Talking about outside of tests.

Also just double checking, does this PR remove all uses of TextField from Widgets?

packages/flutter/test/widgets/autocomplete_test.dart
Renzo-Olivares: Agreed, this specific test is probably okay only being in the widgets library. The tests in this file are more about the behavior of the autocomplete menu and not the input field it is attached to.

packages/flutter/test/widgets/editable_text_cursor_test.dart
Renzo-Olivares: Yeah I would rope these tests into the bucket of "Tests that rely on defaults set in `TextField` or `CupertinoTextField`". For example "cursor animates on iOS", this is not an existing default behavior of `EditableText` only `TextField` and `CupertinoTextField`.

packages/flutter/test/widgets/editable_text_utils.dart
Renzo-Olivares: I did not consider it, I mainly followed the pattern we use in the framework (top-level global) for other text selection controls like `materialTextSelectionHandleControls`. It eliminates some verbosity, see below.

`TestTextSelectionHandleControls.testTextSelectionHandleControls` vs `testTextSelectionHandleControls`.

loic-sharma: SGTM! 👍

packages/flutter/test/widgets/editable_text_cursor_test.dart
Renzo-Olivares: I ended up creating a `text_field_cursor_test.dart` in material and cupertino with tests that verify their internal defaults respectively. I also refactored some of the cursor tests in `editable_text_cursor_test.dart` that were simply testing defaults in `TextField`, in the widgets layer we now only test that the cursor can animate, and that even when `cursorOpacityAnimates` is set to `true`, `EditableText.debugDeterministicCursor` still takes precedence.

commit: [`454838f` (#180494)](https://github.com/flutter/flutter/pull/180494/commits/454838fd04bde8abb94525bd16f4c6af782af664)

Renzo-Olivares:
> After writing this PR, do you have any more thoughts on whether or not some kind of widget like TestTextField might be useful for developers using only the Widgets library? Talking about outside of tests.

I found that in the Widgets library convenient usages of `TextField` did not often make use of gestures provided by `TextSelectionGestureDetector`. And in the same realm I also found that most of the tests verifying gestures are in the design library tests `text_field_test.dart`. While `EditableText` does not make use of `TextSelectionGestureDetector` by default, it still might be useful to have some integration tests to verify both components work well together. `TestTextField` can likely be used for these tests. I don't think the lack of gesture usages indicates that these gestures would not be useful addition to `EditableText`, rather the given tests simply did not require them.

A few tests fell into the category of testing defaults set by `TextField` but not `EditableText`. I think for these category of tests we should move them to the design library and if possible write a test better catered to `EditableText`. For example there were various tests that verified cursor defaults set by text field, "Cursor animates on iOS", this is not a default set by `EditableText`, so I moved this to test to `text_field_cursor_test.dart` and created a new test for editable text called "Cursor animates when cursorOpacityAnimates is true".

In general I didn't find any additional evidence from the convenient usages of `TextField` I went through in our tests to support a `TestTextField` outside of tests.

> Also just double checking, does this PR remove all uses of TextField from Widgets?

There are still around ~5 tests that need a deeper look when migrating.

justinmc (approved):
LGTM :+1:

loic-sharma (approved):
LGTM!

justinmc (approved):
Changes LGTM :+1:

loic-sharma (approved):
Re-LGTM!

Issue #180503, opened 2026-01-05
Many outdated dependencies - is the pub autoroller working correctly?


goderbauer:
`flutter pub outdated` reports many outdated dependencies in the flutter repo that could be resolved to a higher version:

```
$ flutter pub outdated 
Showing outdated packages.
[*] indicates versions that are not the latest available.

Package Name             Current   Upgradable  Resolvable  Latest   

direct dependencies:    
_fe_analyzer_shared      *89.0.0   *89.0.0     92.0.0      92.0.0   
analyzer                 *8.2.0    *8.2.0      9.0.0       9.0.0    
archive                  *3.6.1    *3.6.1      4.0.7       4.0.7    
code_assets              *0.19.10  *0.19.10    1.0.0       1.0.0    
ffigen                   *18.1.0   *18.1.0     20.1.1      20.1.1   
gcloud                   *0.8.19   *0.8.19     *0.8.19     0.9.0    
google_mobile_ads        *5.1.0    *5.1.0      7.0.0       7.0.0    
googleapis               *12.0.0   *12.0.0     *14.0.0     15.0.0   
googleapis_auth          *1.6.0    *1.6.0      2.0.0       2.0.0    
hooks                    *0.20.5   *0.20.5     1.0.0       1.0.0    
metrics_center           *1.0.13   *1.0.13     1.0.14      1.0.14   
native_toolchain_c       *0.17.2   *0.17.2     0.17.4      0.17.4   
petitparser              *6.1.0    *6.1.0      7.0.1       7.0.1    
shelf_web_socket         *2.0.1    *2.0.1      3.0.0       3.0.0    
webview_flutter          *4.9.0    *4.9.0      4.13.0      4.13.0   
webview_flutter_android  *3.16.9   *3.16.9     4.10.11     4.10.11  
xml                      *6.5.0    *6.5.0      6.6.1       6.6.1    
```

I would have expected that the pub autoroller picks up these new versions, but it appears to not be doing that. The pin list (https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/update_packages_pins.dart) also doesn't seem to explain why these are not updated...

Piinks:
Last pub roll landed yesterday: https://github.com/flutter/flutter/pull/180510
I wonder if it is missing some steps? I have been manually rolling it semi-regularly because it will not generate lock files sometimes.

jtmcdole:
[...]

PR #180504, opened 2026-01-05
Bump deps of hooks,code_assets,native_toolchain_c

goderbauer:
It's unclear why these aren't getting picked up by the autoroller, see https://github.com/flutter/flutter/issues/180503.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -126,10 +126,10 @@ packages:
     dependency: "direct main"
     description:
       name: code_assets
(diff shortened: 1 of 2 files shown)


dcharkes (approved):
Thanks @goderbauer!

(It seems that the flutter_tools pubspec.yaml is being updated automatically after unpinning, the hooks_runner package automatically got updated from 1.0.0 to 1.0.1.)

PR #180507, opened 2026-01-05
Bump ffigen

goderbauer:
It's unclear why it isn't getting picked up by the autoroller, see https://github.com/flutter/flutter/issues/180503.


liamappelbe:
Typically when you update ffigen you should also regen whatever bindings it generated. From a quick search, it looks like the only places that ffigen is used (aside from the templates, which hard code a different version) is in the [link_hook integration test](dev/integration_tests/link_hook/ffigen.yaml) and some [Fuchsia zircon bindings](engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon_ffi/pubspec.yaml). So I think to do this roll properly you'd also have to regen those, otherwise there's not really much point updating the ffigen version.

```shell
dart run ffigen --config dev/integration_tests/link_hook/ffigen.yaml
dart run ffigen --config engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon_ffi/pubspec.yaml
```

It's also a little odd to have ffigen as a dependency rather than a dev_dependency, but it looks like this top level pubspec doesn't have a dev_dependency section, so I guess that's WAI.

goderbauer:
We should have some tests that fail if the generated bindings become outdated (e.g. through bumping the ffigen version). I added one, but I think the problem is now that the bots are not set up with the right header files? I'll have to look into that...

(The Fuchsia zircon bindings use an independent ffigen version specified [here](https://github.com/flutter/flutter/blob/d5ee675b2fcb73761562f57036b71e6bcf97a72e/engine/src/flutter/shell/platform/fuchsia/dart-pkg/zircon_ffi/pubspec.yaml#L10), that's not affected by this bump).

dev/integration_tests/link_hook/ffigen.yaml
```diff
@@ -12,9 +12,9 @@ headers:
     - 'src/link_hook.h'
 ffi-native:
 preamble: |
-  // ignore_for_file: always_specify_types
-  // ignore_for_file: camel_case_types
-  // ignore_for_file: non_constant_identifier_names
+  // Copyright 2014 The Flutter Authors. All rights reserved.
+  // Use of this source code is governed by a BSD-style license that can be
+  // found in the LICENSE file. 123
```
goderbauer: note to self: once verified that the test fails on CI with an outdated binding error remove the 123.

liamappelbe (approved):

Issue #180496, opened 2026-01-05
[web] Form validation errors not linked to inputs via aria-describedby (WCAG 2.2 compliance)


flutter-zl:
## Description

Form validation error messages in Flutter Web are not semantically linked to their corresponding input fields via `aria-describedby`, which is required for WCAG 2.2 compliance (Success Criteria 3.3.1 and 3.3.3).

## Current Behavior

When a `TextFormField` displays a validation error:

1. ✅ The input element correctly has `aria-invalid="true"`
2. ❌ The input element is **missing** `aria-describedby` linking to the error message
3. ❌ Screen readers announce "invalid" but don't read the error message with the input

**Current DOM structure:**
<input aria-invalid="true" 
       aria-label="Email" 
       type="text">

<flt-semantics id="flt-semantic-node-15">
  <span>Email is required</span>
</flt-semantics>**Screen reader experience:**
- Focus on Email field → "Email, text field, invalid"
- Error message is announced separately (if at all)
- User cannot programmatically discover what the error is

## Expected Behavior

The input should be semantically linked to its error message:

<input aria-invalid="true" 
       aria-label="Email"
       aria-describedby="flt-semantic-node-15"
       type="text">

<flt-semantics id="flt-semantic-node-15">
  <span>Email is required</span>
</flt-semantics>**Expected screen reader experience:**
- Focus on Email field → "Email, text field, invalid. Email is required"
- Error is announced together with the input context
- User immediately understands what needs to be corrected

## Live Demo

**Before (current behavior):** https://flutter-demo-04-before.web.app

### Steps to Reproduce

1. Visit the demo URL
2. Click "Submit" to trigger validation errors
3. Open browser DevTools (F12)
4. Inspect the email or password `<input>` element
5. Notice: `aria-invalid="true"` is present, but `aria-describedby` is missing

## WCAG 2.2 Requirements

This issue affects compliance with:

- **SC 3.3.1 Error Identification (Level A):** Error messages must be programmatically associated with the relevant form control
- **SC 3.3.3 Error Suggestion (Level AA):** When an input error is detected, suggestions for correction should be provided and linked to the control

Reference: [WCAG 2.2 Understanding Error Identification](https://www.w3.org/WAI/WCAG22/Understanding/error-identification.html)

## Proposed Solution

This would require coordinated changes across the framework and engine:

### 1. Framework: Add `describedBy` Property to Semantics

**File:** `packages/flutter/lib/src/semantics/semantics.dart`

[...]

PR #169157, opened 2025-05-20
Split hint from label and expose it via aria-description or aria-describedby

flutter-zl:
**Summary**
This PR improves web accessibility in Flutter by separating hint from label in the semantics engine and exposing them via ARIA attributes. The hint is set using aria-description (or aria-describedby as a fallback for browsers that do not support aria-description).

**Details**
Hint separation:
The hint is exposed via aria-description if supported, or via aria-describedby with a hidden node as a fallback.
Browser compatibility:
Uses feature detection to choose between aria-description and aria-describedby.
Test coverage:
Added/updated tests to verify correct ARIA attribute behavior and fallback logic.

**How to verify**
All relevant tests pass (semantics_test.dart, semantics_text_test.dart).

**Motivation**
This change brings Flutter web closer to accessibility best practices and ARIA standards, improving the experience for users of assistive technologies.

**Before/After Change**
before: https://before-change-hint.web.app/
after: https://after-change-hint.web.app/ 
after(fallback to aria-describedby): https://after-change-hint-fallback.web.app/

**Issues fixed**
https://github.com/flutter/flutter/issues/162140

**Note**
Some focus-related tests (e.g., incrementable sends focus events) are failing, but these failures are also present on the main branch and are unrelated to the ARIA label/hint changes in this PR.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

PR #180556, opened 2026-01-06
[web] Pass form validation errors to screen readers via aria-description

flutter-zl:
Fixes part of #180496

## Summary
Pass form validation errors to screen readers via `aria-description` on text fields.

**Before:** 
https://flutter-demo-04-before.web.app/

**After:** 
https://flutter-demo-04-after.web.app/

## Limitations & Future Work

This handles **string-based errors** (`errorText`). For more complex cases(brought in https://github.com/flutter/flutter/issues/180496#issuecomment-3713178684), a follow-up implementation using `aria-describedby` with element IDs is tracked in #180496:

- Custom error widgets (`InputDecoration.error`)
- Errors outside `InputDecoration`
- Custom announcement ordering

The `aria-description` approach (current) and `aria-describedby` approach (future) can coexist per ARIA specifications.

## Related
- Related discussion: #169157 comments


packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1792,10 +1792,19 @@ class _TextFieldState extends State<TextField>
           child: AnimatedBuilder(
             animation: controller, // changes the _currentLength
             builder: (BuildContext context, Widget? child) {
+              // Pass error or hint text to semantics so screen readers announce it
+              // along with the input field (via aria-description on web).
+              // Error takes priority over hint since it requires user action.
+              // Note: The visual hintText is already shown as placeholder, so we
+              // only need to expose it via semantics when there's no error.
+              final String? errorText =
+                  widget.decoration?.errorText ?? widget.decoration?.error?.toString();
```
chunhtai: widget toString() contains debug message and can be potentially nonsensical string in release mode. we can't use it as hint. In case they provide an error widget, we should consider let developer provide one string that represent the error

packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1792,10 +1792,19 @@ class _TextFieldState extends State<TextField>
...
+              final String? semanticsHint = errorText ?? widget.decoration?.hintText;
```
chunhtai: This ignores hinttext if error text is presented. I would guess they should concatenate together. Can you double check the native material textfield's behavior if both are present?

flutter-zl: according to t[his link](https://m1.material.io/components/text-fields.html#text-fields-layout:~:text=When%20input%20isn%E2%80%99t%20accepted%2C%20text%20fields%20can%20display%20an%20error%20message%20below%20the%20input%20line%2C%20with%20instructions%20on%20how%20to%20fix%20the%20error.%20Until%20the%20error%20is%20fixed%2C%20the%20error%20replaces%20the%20helper%20text.), When an error is set, it replaces the helper text visually. The current logic errorText ?? hintText matches how native Material handles this : errors override helper/hint text rather than being combined with it.

packages/flutter/lib/src/material/text_field.dart
flutter-zl: good catch. updated.

packages/flutter/lib/src/material/text_field.dart
chunhtai: sounds good

chunhtai (commented):
Think a bit more, setting the semantics of hint text should probably be in InputDecorator widget since that is where the UI are built.

packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1792,10 +1792,18 @@ class _TextFieldState extends State<TextField>
           child: AnimatedBuilder(
             animation: controller, // changes the _currentLength
             builder: (BuildContext context, Widget? child) {
+              // Pass error or hint text to semantics so screen readers announce it
+              // along with the input field (via aria-description on web).
+              // Error takes priority over hint since it requires user action.
+              // Note: The visual hintText is already shown as placeholder, so we
+              // only need to expose it via semantics when there's no error.
+              final String? semanticsHint =
+                  widget.decoration?.errorText ?? widget.decoration?.hintText;
```
chunhtai: Do we still want to use `widget.decoration?.hintText` if the widget.decoration?.error widget is present? I imagine the hintText won't be visible in this case?

flutter-zl:
> Think a bit more, setting the semantics of hint text should probably be in InputDecorator widget since that is where the UI are built.

[InputDecorator is the visual decoration layer inside](https://api.flutter.dev/flutter/material/InputDecorator-class.html#:~:text=TextField%20uses%20this%20widget%20to%20decorate%20its%20EditableText%20child.).  It doesn't have access to the Semantics wrapper that TextField adds around it. The hint property we're setting becomes aria-description on the <input> element, and that wrapping is done by TextField.

packages/flutter/lib/src/material/text_field.dart
flutter-zl: Good catch! Updated the logic.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -342,9 +342,24 @@ class SemanticTextField extends SemanticRole {
     } else {
       editableElement.removeAttribute('aria-required');
     }
+
+    // Apply hint as aria-description on the editable element so screen readers
+    // announce it along with the input field. This enables form validation
+    // errors to be announced when the error text is passed via the hint property.
+    _updateHintDescription();
```
chunhtai, line 349: Is this only specific for textfield, or should we apply to all semantics roles.

chunhtai (approved):
Framework side LGTM. left a question on engine side. should wait for @mdebbar's review before merge

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -2678,11 +2678,21 @@ class _InputDecoratorState extends State<InputDecorator> with TickerProviderStat
       expands: widget.expands,
     );
 
+    // Pass error or hint text to semantics so screen readers announce it
+    // along with the input field (via aria-description on web).
+    // Error takes priority over hint since it requires user action.
+    // Note: When an error widget is displayed (even without errorText),
+    // we don't announce hintText since the visual state shows an error.
+    final bool hasErrorWidget = decoration.error != null || decoration.errorText != null;
+    final String? semanticsHint = hasErrorWidget ? decoration.errorText : decoration.hintText;
```
mdebbar: If there's an error in `decoration.error` but `decoration.errorText` is null, then `semanticsHint` becomes null. Is that a desired result?

I think this may be the "complex case" mentioned in the PR description, is it? If so, let's leave a TODO here explaining what needs to happen.

engine/src/flutter/lib/web_ui/lib/src/engine/semantics/text_field.dart
mdebbar: Looks like we already have similar code elsewhere:
https://github.com/flutter/flutter/blob/f916dd6887bf0242132ab6e86bd0a77bd6ff2089/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/label_and_value.dart#L507

That `LabelAndValue` class is handling two things right now: (1) aria-label, (2) aria-description. We may want to split "HintDescription" into its own behavior that can be used by text field and other roles.

That said, I'm okay with keeping the scope of this PR limited to text fields since that's the issue being addressed. But if we want to make things more complex in the future (e.g. adding support for aria-describedby), we should do the refactoring first.

mdebbar (approved):
I left some comments for improvement but overall LGTM!

packages/flutter/lib/src/material/input_decorator.dart
flutter-zl: 1. Yes, that's intentional. If the UI is visually showing an error (via the error widget), announcing the hintText (e.g., "Enter your email") would be misleading since it suggests everything is normal. Comment already explains the scenario.

2.  Good idea. I added the TODO.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -2678,11 +2678,24 @@ class _InputDecoratorState extends State<InputDecorator> with TickerProviderStat
...
+    // Pass error text to semantics so screen readers announce it along with
+    // the input field (via aria-description on web). The hintText is already
+    // in the semantics tree via the hint Text widget that merges up through
+    // _RenderDecoration's childSemanticsConfigurationDelegate, so we only
+    // need to add errorText here. Adding hintText would cause duplication.
+    // TODO(flutter-zl): A follow-up using aria-describedby with element IDs
+    // will address complex cases (custom error widgets, errors outside
+    // InputDecoration, custom announcement ordering). See
+    // https://github.com/flutter/flutter/issues/180496#issuecomment-3713178684.
+    final String? semanticsHint = decoration.errorText;
```
flutter-zl, line 2690: @chunhtai and @mdebbar , I updated the previous logic to this.

**Previous logic**
final bool hasErrorWidget = decoration.error != null || decoration.errorText != null;
final String? semanticsHint = hasErrorWidget ? decoration.errorText : decoration.hintText;

**Issue with previous logic**
hintText was entering the semantics tree twice:
1. Automatically — [the Text(hintText) widget inside InputDecorator](https://github.com/flutter/flutter/blob/f8a1b34d600fd0a7fd72de482c563fb8c4ea10d3/packages/flutter/lib/src/material/input_decorator.dart#L2333) gets merged up into the text field's semantics via _RenderDecoration's markAsMergeUp.
2. Explicitly — our Semantics(hint: hintText) wrapper also set it on the same node.
The semantics framework concatenates both with \n, producing "Search Google Pay\nSearch Google Pay" error in google internal app.

**Fix**
Only pass errorText (not hintText) in the explicit wrapper, since hintText already has the automatic path.

mdebbar: I don't fully understand the semantics merging logic, but from what you said, does that mean `decoration.errorText` and `hintText` will be concatenated now?

There are tests for `labelText+hintText` and for `labelText+errorText`. Can we try a test for `errorText+hintText`?

I'll defer to @chunhtai who knows more about this. Everything else looks good to me!

mdebbar (approved):

packages/flutter/lib/src/material/input_decorator.dart
flutter-zl: No, errorText and hintText do not concatenate. They end up in separate semantics properties:

1. hintText → merges into label ([via the hint Text widget's RenderParagraph setting config.attributedLabel](https://github.com/flutter/flutter/blob/da086ba7436da5afe261d551d5f4748a21d2bb82/packages/flutter/lib/src/rendering/paragraph.dart#L1197C1-L1198C1) then [markAsMergeUp](https://github.com/flutter/flutter/blob/da086ba7436da5afe261d551d5f4748a21d2bb82/packages/flutter/lib/src/material/input_decorator.dart#L1726C1-L1727C1)) 

2. errorText → goes into hint (via our explicit Semantics(hint:) wrapper).

Added a test for errorText+hintText to verify this.

chunhtai: The change sounds good to me. I will take another review

chunhtai (approved):
LGTM

PR #180577, opened 2026-01-06
Add bottom navigation bar item semantics label

### LINKED ISSUE #180621: Add a semanticsLabel for the BottomNavigationBarItem
### Use case

Currently, BottomNavigationBarItem uses the label property for both visual display and accessibility (screen readers) in the BottomNavigationBar widget. This creates a problem when we need to provide additional context for screen reader users.

### Proposal

I already have an [associated PR](https://github.com/flutter/flutter/pull/180577/files) that adds a String semanticsLabel property to BottomNavigationBarItem

An alternative approach could be to replace the current String label with a Widget label, giving developers full control over both the visual content and its accessibility semantics. However, this would be a breaking change.

Happy to adjust the implementation based on your feedback.

Lilian-C:
Fix [#180621](https://github.com/flutter/flutter/issues/180621)

This PR adds a semanticsLabel for the BottomNavigationBarItem.

While this PR addresses the need for a semanticsLabel, it's worth noting that replacing the current "String label" parameter with a "Widget label" could be a more flexible long-term solution in the BottomNavigationBarItem. This approach would allow full control over the label's content and appearance.

However, this would be a breaking change.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/bottom_navigation_bar_item.dart
```diff
@@ -104,4 +105,17 @@ class BottomNavigationBarItem {
   ///
   /// Defaults to null, in which case the tooltip is not shown.
   final String? tooltip;
+
+  /// The semantic label for this [BottomNavigationBarItem].
+  ///
+  /// This is used by accessibility tools to describe the item. When provided,
+  /// it overrides the default semantic label which is typically composed of
+  /// the [label] and the tab position information.
```
Piinks:
```suggestion
  /// it overrides the default [label] String when read by accessibility tools.
```
-ish, feel free to tweak.

I think, and @chunhtai can tell me if I am wrong :), the tab information will still be provided in the default semantics here. 
The BottomNav..Item ends up getting wrapped in a Semantics container with the tab index, and then further down the tree from there is where the label text is inserted. The docs should be clear about this in case folks think they need to figure out tab indexing themselves.

packages/flutter/test/material/bottom_navigation_bar_test.dart
```diff
@@ -3103,6 +3103,39 @@ void main() {
...
+              label: 'B',
+            ),
+          ],
+        ),
+      ),
+    );
+
+    expect(
+      tester.getSemantics(find.text('A')),
+      containsSemantics(label: 'Custom A label\nTab 1 of 2'),
```
Piinks: Yep, the test actually confirms this, great thanks!

packages/flutter/lib/src/widgets/bottom_navigation_bar_item.dart
Lilian-C: Fixed :)

Piinks (approved):
LGTM thank you!

hannah-hyj (approved):
LGTM!

PR #180584, opened 2026-01-06
Add resetOnBlur property to DropdownMenu

### LINKED ISSUE #177983: In DropdownMenu add an option to reset text to the selected value on blur
### Use case

Searchable select components in many UI libraries reset the input text to the currently selected option when the field loses focus, unless a new value is actually selected. This avoids leaving temporary search text in place when the user did not confirm it as their selection.

In Flutter's `DropdownMenu`, when the search input loses focus, whatever text the user typed remains in the field even if it does not correspond to a selected option. This behavior can be confusing because users expect the text to revert to the selected item unless they explicitly choose a new one.

---

#### Case 1: Search text does not match any option label

1. "A1" is selected
  
Image

2. User opens the dropdown

Image

3. User types "foo"

Image

4. Dropdown loses focus without selection, and Input resets to "A1" because "foo" was never selected

Image

#### Case 2: Search text matches an option label but is not selected

1. "A1" is selected

Image

2. User opens the dropdown

Image

3. User types "A2" but does not select it

Image

4. Dropdown loses focus and Input resets to "A1" because "A2" was not selected

Image

This follows behavior seen in many select widgets on the web including some Material Design implementations.

### Proposal

Add a parameter to `DropdownMenu` that resets the text field to the currently selected value when the widget loses focus and no new selection has been made.

Suggested parameter name:

* `clearOnBlur`

This provides a more intuitive UX for searchable dropdowns while avoiding extra code to manage focus and manually restore text.

---

### Current workaround

Developers can currently work around this behavior by adding a FocusNode listener and manually restoring the text when losing focus. However, this is verbose and not intuitive for new users. 

* Flutter 3.35.7

```dart

class MyDropdownMenu extends StatefulWidget {
  const MyDropdownMenu({super.key});

  @override
  State<MyDropdownMenu> createState() => _MyDropdownMenuState();
}

class _MyDropdownMenuState extends State<MyDropdownMenu> {
  final TextEditingController _textEditingController = TextEditingController();
  final FocusNode _focusNode = FocusNode();

  void _dropdownFocusNodeListener() {
    if (!_focusNode.hasFocus) {
      final String currentSelectedOptionLabel = getOptionLabel(_selectedOption);
      if (_textEditingController.text != currentSelectedOptionLabel) {
        _textEditingController.text = currentSelectedOptionLabel;
      }
    }
  }

  @override
    return DropdownMenu<Option>(
      controller: _textEditingController,
      focusNode: _focusNode,
      enableSearch: true,
      enableFilter: true,
      initialSelection: _selectedOption,
      dropdownMenuEntries: Option.values
          .map(
            (option) =>
                DropdownMenuEntry(value: option, label: getOptionLabel(option)),
          )
          .toList(),
      onSelected: (Option? option) {
        setState(() => _selectedOption = option);
      },
    );
  }

  Option? _selectedOption = Option.a1;

  String getOptionLabel(Option? option) => switch (option) {
    Option.a1 => 'A1',
    Option.a2 => 'A2',
    null => '',
  };

  @override
  void initState() {
    super.initState();

    _focusNode.addListener(_dropdownFocusNodeListener);
  }

  @override
  dispose() {
    _textEditingController.dispose();
    _focusNode.removeListener(_dropdownFocusNodeListener);
    _focusNode.dispose();

    super.dispose();
  }
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: Container(
            padding: EdgeInsets.all(24.0),
            constraints: BoxConstraints(maxWidth: 300),
            child: Column(
              spacing: 24,
              children: [MyDropdownMenu()],
            ),
          ),
        ),
      ),
    );
  }
}

enum Option { a1, a2 }

```

Koichi5:
This PR adds a `clearOnBlur` property to `DropdownMenu`.

Hoping to fix https://github.com/flutter/flutter/issues/177983

## Pre-launch Checklist

- [x] I signed the [CLA].


QuncCccccc (commented):
Hi, thanks for your contribution! I haven't taken a deep dive into the PR yet, but I noticed that some tests have failed. Could you help fix them?

QuncCccccc (commented):
Thanks for fixing all the tests! I'll take a look ASAP!

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -369,6 +370,25 @@ class DropdownMenu<T> extends StatefulWidget {
...
+  /// * The user submits the form (e.g., presses Enter).
+  ///
+  /// Example scenarios:
+  ///
+  /// * If "A1" is selected and the user types "foo" but doesn't select it,
+  ///   the field will reset to "A1" when focus is lost or when Enter is pressed.
+  /// * If "A1" is selected and the user types "A2" but doesn't select it,
+  ///   the field will reset to "A1" when focus is lost or when Enter is pressed.
+  ///
+  /// Defaults to false.
```
dkwingsmt: I think the current doc is quite repetitive. I consulted AI and rephrased it into follows:
```dart
  /// Whether to revert the text field to the currently selected value when the 
  /// widget loses focus or is submitted (e.g., pressing Enter).
  ///
  /// If true, any unselected manual input will be discarded and replaced by 
  /// the last selected label when the user navigates away or submits the field.
  ///
  /// Defaults to false.
```

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -369,6 +370,25 @@ class DropdownMenu<T> extends StatefulWidget {
...
+  final bool clearOnBlur;
```
dkwingsmt: Instead of `clearOnBlur`, I think this flag is more proper named `resetOnBlur`. What do you think?

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -1225,6 +1309,18 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
       controller: _controller,
       menuChildren: menu,
       crossAxisUnconstrained: false,
+      onClose: () {
+        if (widget.clearOnBlur) {
+          WidgetsBinding.instance.addPostFrameCallback((_) {
```
dkwingsmt: Can you explain why this is needed? This PR doesn't change how focus is managed, but only uses the focus change as a result. It shouldn't need adding extra focus management.

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -748,13 +769,20 @@ class _DropdownMenuState<T> extends State<DropdownMenu<T>> {
         text: filteredEntries[index].label,
         selection: TextSelection.collapsed(offset: filteredEntries[index].label.length),
       );
+      _lastSelectedValue = widget.initialSelection;
     }
     refreshLeadingPadding();
     _controller = widget.menuController ?? MenuController();
+    if (widget.focusNode != null) {
```
dkwingsmt, line 767: This might be a problem because `clearOnBlur` is only useful if `focusNode` is set. 

(if `focusNode` is not set, `TextField` manages its own focus node.)

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: Thanks for the suggestion! Applied the rephrased version.
https://github.com/flutter/flutter/pull/180584/commits/68b8df642c768adc7e7a40b48da67ddd77370015

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: The name suggested in this issue was `clearOnBlur`, but I believe `resetOnBlur` is clearer, so I renamed it as follows:
https://github.com/flutter/flutter/pull/180584/commits/78cff7d3422f32568f999410f458b68aefc6686c

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: This code was left behind due to my oversight. I have removed it below.
https://github.com/flutter/flutter/pull/180584/commits/4cb690f140b9f028aac24e4603bf3e5dc3af20ab

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: Assertions have been added below.
https://github.com/flutter/flutter/pull/180584/commits/eef4d9714a6eef548b2518ff7652534d751b8962

packages/flutter/lib/src/material/dropdown_menu.dart
```diff
@@ -224,7 +225,8 @@ class DropdownMenu<T> extends StatefulWidget {
     this.cursorHeight,
     this.restorationId,
     this.menuController,
-  }) : assert(filterCallback == null || enableFilter),
+  }) : assert(!resetOnBlur || focusNode != null, 'resetOnBlur requires a focusNode to detect focus changes.'),
```
dkwingsmt, line 228: While this technically works, it puts a pretty strong (and from developers' perspective, unreasonable) limitation to this API: To use `resetOnBlur` you must provide a `focusNode`.

A more ideal implementation would be that, if `focusNode` is null, `DropdownMenu` manages its own internal focus node. In this way `DropdownMenu` will always have a focus node to listen to whether `widget.focusNode` is null. This is a common pattern used in many widgets, which you can find by searching for `_effectiveFocusNode`. It would be great if you can add this first, preferably in a separate PR. What do you think?

packages/flutter/test/material/dropdown_menu_test.dart
```diff
@@ -5334,6 +5334,611 @@ void main() {
...
+      await tester.pump();
+      // Type the same text as the selected value
+      await tester.enterText(find.byType(TextField).first, 'Item 0');
+      await tester.pump();
+      expect(controller.text, 'Item 0');
+
+      otherFocusNode.requestFocus();
+      await tester.pumpAndSettle();
+
+      // No reset needed since text already matches the selected value
```
dkwingsmt: I don't think this is actively testing anything. Whether it resets or not, `controller.text` is gonna be `'Item 0'`. Maybe you'd want to test that `onSelected` is not called at all.

packages/flutter/test/material/dropdown_menu_test.dart
```diff
@@ -5334,6 +5334,611 @@ void main() {
...
+
+      // After a valid selection, losing focus should not reset the text
+      otherFocusNode.requestFocus();
+      await tester.pumpAndSettle();
+
+      expect(controller.text, 'Item 2');
+      expect(selectedValue, TestMenu.mainMenu2);
+    });
+
+    testWidgets('resetOnBlur does not reset on Enter when false', (WidgetTester tester) async {
```
dkwingsmt, line 5696: What if you
* Set `resetOnBlur` to true
* Input something that's not an option
* press Enter

Does it reset or does it prevent submission? Do we have a test?

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: I agree that requiring a focusNode for resetOnBlur is too restrictive. I'll create a separate PR to introduce the `_effectiveFocusNode` pattern in DropdownMenu first, then rebase this PR on top of it.

packages/flutter/test/material/dropdown_menu_test.dart
Koichi5: I've updated the test to track `onSelectedCount` and verify that `onSelected` is not called when the text already matches the selected value. I also updated the implementation in `_handleFocusChange` to only call onSelected when the text is actually changed by the reset.
https://github.com/flutter/flutter/pull/180584/changes/3c649061ddf581d33e9098d375c9a0f3f4f804ca

packages/flutter/test/material/dropdown_menu_test.dart
Koichi5: When resetOnBlur is true and the user inputs something that's not an option, pressing Enter resets the text to the last selected value (via `_handleSubmitted` → `_restoreSelectionFromText`).
I've added a test for this case: 'resetOnBlur resets on Enter when true and text does not match any option'. 
https://github.com/flutter/flutter/pull/180584/changes/ff5085d6ffed725e8fc18c3c479266880218bd5a

packages/flutter/lib/src/material/dropdown_menu.dart
Koichi5: I've created #183684 to introduce the `_effectiveFocusNode` pattern. Once that is merged, I'll rebase this PR to use it and remove the assertion. Would you mind taking a look when you get a chance?

dkwingsmt: Awesome! I'll definitely review it asap.

Piinks:

Piinks:
This PR is now ready to port over to flutter/packages!

See instructions and guidance in https://github.com/flutter/flutter/issues/188444 Thank you!

QuncCccccc (commented):
I just left a similar [comment](https://github.com/flutter/flutter/pull/183684#issuecomment-5517080763) in https://github.com/flutter/flutter/pull/183684 which seems can unblock this PR. Once that PR is ported, feel free to port this PR change to material_ui and we can proceed with code review:) Thanks again for your contribution!

PR #180586, opened 2026-01-06
[ Widget Preview ] Add support for `dart:ffi` imports

### LINKED ISSUE #166431: Support for previews that include calls to unsupported web APIs (e.g., native plugins, dart:io, etc.)
Since the widget previewer is a Flutter web application, native plugins and `dart:io` won't be supported. Since it's still possible to import `dart:io` and invoke Dart code for native plugins, the widget previewer should catch these classes of exceptions and inform developers that their preview can't be displayed.

Ideally, an error widget will be shown in the place of the user's preview explaining what caused the crash with links to documentation on possible solutions to resolve the issue (e.g., refactoring widgets to separate business logic from UI logic).

bkonyi:
Adds logic to pass `--include-unsupported-platform-library-stubs` to the CFE when launched via `flutter widget-preview start`, as well as logic to ensure the flag can't be passed as an extra frontend option to bypass compile time errors due to imports of unsupported libraries.

Fixes https://github.com/flutter/flutter/issues/166431


jyameo (approved):
lgtm!

PR #180597, opened 2026-01-06
Added type annotations and removed lints for run_tests.py

gaaclarke:
This passes mypy and pylint.  I looked into introducing mypy into ci, but it didn't make sense considering the small number of python scripts.

Test exempt: this is a test

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/testing/run_tests.py
+++ b/engine/src/flutter/testing/run_tests.py
@@ -63,7 +63,7 @@ def print_divider(char: str = '=') -> None:
 
 
 def is_asan(build_dir: str) -> bool:
-  with open(os.path.join(build_dir, 'args.gn')) as args:
+  with open(os.path.join(build_dir, 'args.gn'), encoding='utf-8') as args:
     if 'is_asan = true' in args.read():
       return True
 
@@ -91,7 +91,7 @@ def run_cmd( # pylint: disable=too-many-arguments
 
   start_time = time.time()
 
-  process = subprocess.Popen(
+  with subprocess.Popen(
       cmd,
       cwd=cwd,
       stdout=subprocess.PIPE,
@@ -99,14 +99,14 @@ def run_cmd( # pylint: disable=too-many-arguments
       env=env,
       universal_newlines=True,
       **kwargs
-  )
-  output = ''
+  ) as process:
+    output = ''
 
-  for line in iter(process.stdout.readline, ''):
-    output += line
-    logger.info(line.rstrip())
+    for line in iter(process.stdout.readline, ''):
+      output += line
+      logger.info(line.rstrip())
 
-  process.wait()
+    process.wait()
   end_time = time.time()
 
   if process.returncode != 0 and not expect_failure:
@@ -126,7 +126,7 @@ def run_cmd( # pylint: disable=too-many-arguments
 
     if not allowed_failure:
       raise RuntimeError(
-          'Command "%s" (in %s) exited with code %s.' % (command_string, cwd, process.returncode)
+          f'Command "{command_string}" (in {cwd}) exited with code {process.returncode}.'
       )
 
   for forbidden_string in forbidden_output:
@@ -183,7 +183,7 @@ def find_executable_path(path: str) -> str:
     if os.path.exists(bat_path):
       return bat_path
 
-  raise Exception('Executable %s does not exist!' % path)
+  raise Exception(f'Executable {path} does not exist!')
 
 
 def vulkan_validation_env(build_dir: str) -> typing.Dict[str, str]:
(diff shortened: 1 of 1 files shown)


engine/src/flutter/testing/run_tests.py
```diff
@@ -691,43 +698,46 @@ def gather_dart_test(build_dir, dart_file, options):
...
   ios_test_lib = os.path.join(tmp_out_dir, 'libios_test_flutter.dylib')
   message = []
   message.append('gn --ios --unoptimized --runtime-mode=debug --no-lto --simulator')
-  message.append('ninja -C %s ios_test_flutter' % ios_out_dir)
-  final_message = "%s or %s doesn't exist. Please run the following commands: \n%s" % (
-      ios_out_dir, ios_test_lib, '\n'.join(message)
+  message.append(f'ninja -C {ios_out_dir} ios_test_flutter')
+  final_message = (
+      f"{ios_out_dir} or {ios_test_lib} doesn't exist. "
+      f'Please run the following commands: \n{chr(10).join(message)}'
```
b-luk: Is `chr(10).join()` preferred over `"\n".join()`?

engine/src/flutter/testing/run_tests.py
```diff
@@ -91,22 +91,16 @@ def run_cmd( # pylint: disable=too-many-arguments
...
-  for line in iter(process.stdout.readline, ''):
-    output += line
-    logger.info(line.rstrip())
+    if process.stdout:
+      for line in iter(process.stdout.readline, ''):
+        output += line
+        logger.info(line.rstrip())
 
-  process.wait()
+    process.wait()
```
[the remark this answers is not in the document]
b-luk: This suggestion makes sense to me.

engine/src/flutter/testing/run_tests.py
```diff
@@ -691,43 +698,46 @@ def gather_dart_test(build_dir, dart_file, options):
...
+def assert_expected_xcode_version() -> None:
   """Checks that the user has a version of Xcode installed"""
   version_output = subprocess.check_output(['xcodebuild', '-version'])
   # TODO ricardoamador: remove this check when python 2 is deprecated.
-  version_output = version_output if isinstance(version_output,
-                                                str) else version_output.decode(ENCODING)
-  version_output = version_output.strip()
-  match = re.match(r'Xcode (\d+)', version_output)
+  version_output_str = version_output if isinstance(version_output,
+                                                    str) else version_output.decode(ENCODING)
```
b-luk, line 728: Is Python2 still used with this? I think `subprocess.Popen` with a `with` block isn't supported in Python2, so maybe this file is supposed to be Python3-only.

If this is only for Python3, I think the TODO is suggesting to simplify this to always do `.decode(ENCODING)`.

b-luk (approved):
LGTM, with the caveat that I haven't written any Python code for many years (and never Python3 or Python with type annotations). So if you want someone with actual Python experience to review it, make sure to include another reviewer.

engine/src/flutter/testing/run_tests.py
b-luk: Maybe just get rid of this whole message-list/final_message/'\n'-join thing and have one single message f-string. I'm not seeing why this was made to be this slightly convoluted message/final_message/join thing thing.

gaaclarke: lol no, gemini did it automatically and I missed it.  thanks.  i'll fix it

engine/src/flutter/testing/run_tests.py
gaaclarke: You're right.  It looks like it was introduced in 3.2.  We must be using that on ci.  We use the 2.7 version of pylint, I would have expected it to flag this, but doesn't look like it can.

engine/src/flutter/testing/run_tests.py
gaaclarke: Oh, it looks like f-strings doesn't handle having string literals in the {} expressions.

engine/src/flutter/testing/run_tests.py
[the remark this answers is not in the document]
gaaclarke: redundant now that we can't use popen with with statements

engine/src/flutter/testing/run_tests.py
gaaclarke: done

engine/src/flutter/testing/run_tests.py
gaaclarke: Removed them, based on the fact we are using the 2.7 version of pylint i'll assume we are targeting that.

engine/src/flutter/testing/run_tests.py
b-luk: > Oh, it looks like f-strings doesn't handle having string literals in the {} expressions.

That's surprising. We do have `{" ".join(flags)}']` and `{",".join(all_types)}` in f-strings elsewhere in this file.

gaaclarke: ~~I think it may be some weird thing with the fact there are 2 fstrings back to back?  The linter will complain about the previous line actually: `testing/run_tests.py:717:7: E0001: f-string expression part cannot include a backslash (<unknown>, line 717) (syntax-error)`~~

Looked at it a bit more, I'm actually not sure why it freaks out here.

walley892 (approved):
LGTM

PR #180647, opened 2026-01-07
Add support for fetching array uniforms by name

walley892:
Adds support for fetching objects representing array Uniforms of type float, vec2, vec3, and vec4.  Happy new year!

Before this change, setting an array of uniforms was verbose.
```glsl
uniform vec3[2] uColors;
```

```dart
shader.setFloat(0, 1.0);
shader.setFloat(1, 0.0);
shader.setFloat(2, 1.0);
shader.setFloat(3, 0.0);
shader.setFloat(4, 1.0);
shader.setFloat(5, 0.0);
```

After this change, we have:

```glsl
uniform vec3[2] uColors;
```

```dart
final colors = shader.getUniformVec3Array("uColors");
colors[0].set(1.0, 0.0, 1.0);
colors[1].set(0.0, 1.0, 0.0);
```

- Adds a class UniformArray to represent uniform arrays in Dart.
- Adds a bunch of methods that construct these arrays for various datatypes
- Adds tests
- Also removes some of the awkwardness in the _getUniformFloatIndex function by replacing that function entirely

## Pre-launch Checklist

- [x ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ x] I signed the [CLA].
- [ x] I listed at least one issue that this PR fixes in the description above.
- [ x] I updated/added relevant documentation (doc comments with `///`).
- [ x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ x] All existing and new tests are passing.


engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -477,3 +538,14 @@ class _SkwasmUniformVec4Slot implements ui.UniformVec4Slot {
...
+class _SkwasmUniformArray<T extends ui.UniformType> implements ui.UniformArray<T> {
+  _SkwasmUniformArray._(this._elements);
+
+  @override
+  T operator [](int index) {
+    return _elements[index];
+  }
+
+  final List<T> _elements;
+}
```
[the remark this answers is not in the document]
walley892: Good bot

engine/src/flutter/lib/web_ui/lib/painting.dart
```diff
@@ -1022,18 +1024,22 @@ abstract class UniformFloatSlot {
...
 }
 
-abstract class UniformVec4Slot {
+abstract class UniformVec4Slot extends UniformType {
   void set(double x, double y, double z, double w);
 }
 
+abstract class UniformArray<T extends UniformType> {
+  T operator [](int index);
+}
```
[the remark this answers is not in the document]
walley892: Good bot

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5832,6 +5874,130 @@ base class FragmentShader extends Shader {
...
+  ///
+  /// ```glsl
+  /// uniform vec3[10] uColors;
+  /// ```
+  ///
+  /// ```dart
+  /// void updateShader(ui.FragmentShader shader) {
+  ///   final UniformArray<UniformVec3Slot> colors = shader.getUniformVec3Array('uColors');
+  ///   colors[0].set(1.0, 0.0, 1.0, 0.5);
+  /// }
```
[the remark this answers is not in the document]
walley892: Good bot

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
```diff
@@ -376,6 +376,13 @@ void testMain() {
     shader.getUniformVec4('u_color').set(0.8, 0.1, 0.3, 1.0);
   });
 
+  test('getUniformFloatArray works with correct datatype', () {
+    final Uint8List data = utf8.encode(kJsonIPLR);
+    final program = CkFragmentProgram.fromBytes('test', data);
+    final shader = program.fragmentShader() as CkFragmentShader;
+    shader.getUniformFloatArray('uFloats')[0].set(0.8);
+  });
```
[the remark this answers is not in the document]
walley892: Pushing back on this one Gem gem. These tests are on the way out, and adding a new shader with the requisite uniforms is painful. Added some tests for getUniformVec2Array though

engine/src/flutter/lib/web_ui/lib/src/engine/canvaskit/painting.dart
```diff
@@ -524,3 +591,17 @@ class _CkUniformVec4Slot implements ui.UniformVec4Slot {
...
+class _CkUniformFloatArray<T extends ui.UniformType> implements ui.UniformArray<T> {
+  _CkUniformFloatArray._(this._elements);
+
+  @override
+  T operator [](int index) {
+    return _elements[index];
+  }
+
+  @override
+  int get length => _elements.length;
```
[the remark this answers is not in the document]
walley892: Good bot

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5508,6 +5503,10 @@ base class FragmentProgram extends NativeFieldWrapperClass1 {
   }
 }
 
+// A binding into a uniform defined in a shader. Used now to restrict the types
+// of UniformArrays that can be created.
+abstract base class UniformType {}
```
gaaclarke:
```suggestion
sealed base class UniformType {}
```

There is no reason to let users subclass from this.

gaaclarke: I think you can probably make this private too?  _UniformType

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5765,8 +5797,18 @@ base class FragmentShader extends Shader {
   /// ```
   UniformFloatSlot getUniformFloat(String name, [int? index]) {
     index ??= 0;
-    final int shaderIndex = _program._getUniformFloatIndex(name, index);
-    final result = UniformFloatSlot._(this, name, index, shaderIndex);
+
+    if (index < 0) {
+      throw ArgumentError('Index `$index` out of bounds for `$name`.');
```
gaaclarke: There is a special error for these, IndexError.  It should be safe to change.

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5832,6 +5874,130 @@ base class FragmentShader extends Shader {
...
+
+    if (info.size % elementSize != 0) {
+      throw ArgumentError(
+        'Uniform size (${info.size}) for "$name" is not a multiple of $elementSize.',
+      );
+    }
+    final int numElements = info.size ~/ elementSize;
+    final allAddedSlots = List<UniformFloatSlot>.empty(growable: true);
+
+    final elements = List<T>.empty(growable: true);
```
gaaclarke: nit: We know the size of the List.  Please switch this to a fixed sized List.  It's more efficient on memory.

engine/src/flutter/lib/web_ui/lib/painting.dart
```diff
@@ -1010,7 +1010,9 @@ abstract class FragmentProgram {
   FragmentShader fragmentShader();
 }
 
-abstract class UniformFloatSlot {
+abstract class UniformType {}
```
gaaclarke: Same notes here.

engine/src/flutter/testing/dart/fragment_shader_test.dart
```diff
@@ -160,6 +160,77 @@ void main() async {
...
+      slots[1].set(1.0);
+    });
+
+    test('FragmentProgram getUniformArray not found', () async {
+      try {
+        shader.getUniformFloatArray('unknown');
+        fail('Unreachable');
+      } catch (e) {
+        expect(e.toString(), contains('No uniform named "unknown".'));
+      }
```
gaaclarke: Use `throwA`, there's examples of it in this file.

engine/src/flutter/testing/dart/fragment_shader_test.dart
```diff
@@ -160,6 +160,77 @@ void main() async {
         expect(e.toString(), contains('`iVec3Uniform` has size 3, not size 4.'));
       }
     });
+
+    test('FragmentProgram getUniformArray float', () async {
+      final UniformArray<UniformFloatSlot> slots = shader.getUniformFloatArray(
+        'iFloatArrayUniform',
+      );
+      slots[0].set(1.0);
+      slots[1].set(1.0);
```
gaaclarke, line 199: add assert for the size of slots here and below

gaaclarke (changes requested):

engine/src/flutter/lib/ui/painting.dart
walley892: `sealed` is neat. I don't think you can have a class defined as `sealed base`, though.

I tried that initially, but that makes the linter mad at the declaration `UniformArray<T extends _UniformType` since `UniformArray` is public

engine/src/flutter/lib/ui/painting.dart
walley892: Nice, cleaned up the logic a bit too

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
```diff
@@ -355,25 +355,31 @@ void testMain() {
     expect(program.uniforms, hasLength(7));
     expect(program.name, 'test');
   });
+  group('Uniform by-name accessors happy path.', () {
```
harryterkelsen: could you rework these tests so they can be UI-agnostic (i.e. work in both CanvasKit and Skwasm) and move the test under `test/ui/` instead of `test/canvaskit/`? That way both the CanvasKit and Skwasm implementation will get tested and you can test against the `dart:ui` API instead of the engine-specific classes

engine/src/flutter/lib/ui/painting.dart
walley892: Since these aren't nullable, the only way to create a fixed-size list is with .generate(). I changed the implementation to use .generate() for both of these, but it requires a allAddedSlots.SubList() - which I think creates a new list? I don't know enough about the underlying list implementation to know which is more efficient.

engine/src/flutter/lib/web_ui/lib/painting.dart
walley892: Gottem

engine/src/flutter/testing/dart/fragment_shader_test.dart
walley892: SGTM

engine/src/flutter/testing/dart/fragment_shader_test.dart
walley892: Aye aye, captain!

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
walley892: That's a good idea. This PR is already getting quite large, I filed https://github.com/flutter/flutter/issues/180707 to work on moving the tests over. SGTY?

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -5887,21 +5887,20 @@ base class FragmentShader extends Shader {
...
+
+    final elements = List<T>.generate(
+      numElements,
+      (i) => elementFactory(allFloatSlots.sublist(i * elementSize, i * elementSize + elementSize)),
+    );
 
     // Clean up expired weak references and add the new ones.
     _slots.removeWhere((WeakReference<UniformFloatSlot> ref) => ref.target == null);
-    _slots.addAll(allAddedSlots.map((slot) => WeakReference<UniformFloatSlot>(slot)));
+    _slots.addAll(allFloatSlots.map((slot) => WeakReference<UniformFloatSlot>(slot)));
```
gaaclarke: nit: If you nest your generations you don't have to create that extra list.

```suggestion
    _slots.removeWhere((WeakReference<UniformFloatSlot> ref) => ref.target == null);

    final elements = List<T>.generate(
      numElements,
      (i) {
        final slots = List<UniformFloatSlot>.generate(info.size, (j) => UniformFloatSlot._(...));
        _slots.addAll(slots);
        return elementFactory(slots);
      }
    );
```

gaaclarke (commented):
lgtm once @harryterkelsen is happy!

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -383,8 +372,13 @@ class SkwasmFragmentShader implements SkwasmShader, ui.FragmentShader {
   @override
   ui.UniformFloatSlot getUniformFloat(String name, [int? index]) {
     index ??= 0;
-    final int shaderIndex = _program._getShaderIndex(name, index);
-    return SkwasmUniformFloatSlot._(this, index, name, shaderIndex);
+    final UniformData info = _program._getUniformFloatInfo(name);
+
+    if (index + 1 > info.floatCount) {
```
harryterkelsen: In the native engine you do this check with

```dart
    if (index < 0 || index + 1 > info.size) {
      throw IndexError.withLength(
        index,
        info.size,
        message: 'Index `$index` out of bounds for `$name`.',
      );
    }
```

I suggest using the `IndexError.check` method instead: https://api.flutter.dev/flutter/dart-core/IndexError/check.html

Use the same logic for the index check in all three implementations (native, CanvasKit, and Skwasm)

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
harryterkelsen: Since I don't have much familiarity with the shader code, I can't feel confident in letting this PR land without tests for both CanvasKit and Skwasm. Moving the changes to `test/ui` shouldn't add more code to this PR since we're just moving it from one place to another. The tests will be 99% the same but use the generic `dart:ui` FragmentShader class instead of the Ck-specific class.

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
```diff
@@ -355,25 +355,31 @@ void testMain() {
...
+    late CkFragmentShader shader;
+    setUp(() {
+      final Uint8List data = utf8.encode(kJsonArrayIPLR);
+      final program = CkFragmentProgram.fromBytes('test', data);
+      shader = program.fragmentShader() as CkFragmentShader;
+    });
+    test('getUniformVec2', () {
```
harryterkelsen: These tests all check that setting values on the by-name float arrays doesn't crash. I think we should also check that after setting the values, if we get the float arrays again, the values are set to what we just set them to. Otherwise, these tests will pass even if `set()` is a no-op.

harryterkelsen (commented):
Thanks for your patience and for contributing code to the Flutter engine. I have some comments

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
walley892: ctrl-c+ctrl-v'd the shader definitions and moved the existing getUniform* tests into the other location

engine/src/flutter/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
walley892: Nice

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
walley892: We don't have a mechanism to do this (read uniform values) in the current fragmentshader API. 

It wouldn't be too hard to write a `getFloat(int index)` API and expand the UniformXSlot API based on that, but. I'd want to check with the rest of the engine team before moving forward with that kind of change.

@gaaclarke for thoughts.

engine/src/flutter/lib/ui/painting.dart
walley892: I like this one because it's *clean* but I'm still not sure which is the best for memory allocations. This one creates a new list for each element and updates _slots each iteration rather than all at once.

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
gaaclarke: Checking the result isn't all that helpful because what really matters is that that value is propagated to the GPU.  That would require golden tests.  These classes were designed to be things you create in the widget tree declaratively, not storage classes so I think adding getters to them is contrary to how we want them used.

It would be nice to have a golden tests that runs through this.  I'm not sure if there is one off the top of my mind.  Eventually as the framework adopts this we'll get some.  You can file an issue to follow up since this meets the standard set by the alternative API.

engine/src/flutter/lib/web_ui/test/ui/fragment_shader_test.dart
```diff
@@ -70,6 +70,260 @@ const kTextureShaderSksl = r'''
 }}
 ''';
 
+/// This is generated by impellerc, and included here as a source file to
+/// simplify the build.
+const String kJsonIPLR = r'''
```
gaaclarke: Oye, this is fragile.  At some point we added version numbers to our output.  I don't see that here, maybe it didn't make it to the json output?  I'd prefer this was generated as part of the build system or at the least has a version number so that we get a nice error if it ever becomes out of date.

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
gaaclarke: Oh, actually we do have tests like golden tests, check out `FragmentShader simple shader renders correctly`.  You can do something like that.

walley892:
@gaaclarke @harryterkelsen

I created a PR to move the existing tests to web_ui and committed the existing test shaders to github to make things less bad here. Will rebase on top of that https://github.com/flutter/flutter/pull/180910

engine/src/flutter/lib/web_ui/test/canvaskit/fragment_program_test.dart
walley892: I'm not sure if goldens are the best solution for testing initial implementation here, but they should serve as kind of a regression test?

engine/src/flutter/lib/web_ui/test/ui/fragment_shader_test.dart
walley892: Reverse engineered existing shaders and committed to github in https://github.com/flutter/flutter/pull/180910. Added additional shaders here.

This situation is still really brittle. If we ever change the impellerc format, we'll need to manually update all of these dang strings.

harryterkelsen (approved):
LGTM. Thanks!

gaaclarke (approved):

PR #180651, opened 2026-01-07
De-interleaves engine dart test output

gaaclarke:
This includes the following changes:
1) Logging output from dart tests is no longer interleaved
1) The backend that is running the test is now printed out
1) `TestingContext` abstraction was added in order to support more rendering backends in the future

test exempt: it is tests

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
+++ b/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
@@ -8,8 +8,6 @@
 
 #include <vulkan/vulkan.h>
 
-#include <iostream>
-#include <string>
 #include <vector>
 
 #include "flutter/fml/mapping.h"
--- a/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
+++ b/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
@@ -23,7 +23,8 @@
 
 namespace flutter {
 
-static std::vector<std::shared_ptr<fml::Mapping>> ShaderLibraryMappings() {
+namespace {
+std::vector<std::shared_ptr<fml::Mapping>> ShaderLibraryMappings() {
   return {
       std::make_shared<fml::NonOwnedMapping>(impeller_entity_shaders_vk_data,
                                              impeller_entity_shaders_vk_length),
@@ -104,6 +105,8 @@ class TesterContextVK : public TesterContext {
   std::shared_ptr<impeller::ContextVK> context_;
   std::shared_ptr<impeller::SurfaceContextVK> surface_context_;
 };
+
+}  // namespace
 
 std::unique_ptr<TesterContext> TesterContextVKFactory::Create(
     bool enable_validation) {
--- a/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
+++ b/engine/src/flutter/shell/testing/tester_context_vk_factory.cc
@@ -4,7 +4,7 @@
 
 #define FML_USED_ON_EMBEDDER
 
-#include "flutter/shell/testing/tester_vk.h"
+#include "flutter/shell/testing/tester_context_vk_factory.h"
 
 #include <vulkan/vulkan.h>
 
--- a/engine/src/flutter/shell/testing/BUILD.gn
+++ b/engine/src/flutter/shell/testing/BUILD.gn
@@ -24,7 +24,11 @@ executable("testing") {
     "//flutter:export_dynamic_test_symbols",
   ]
 
-  sources = [ "tester_main.cc" ]
+  sources = [
+    "tester_main.cc",
+    "tester_vk.cc",
+    "tester_vk.h",
+  ]
   libs = []
   if (is_win) {
     libs += [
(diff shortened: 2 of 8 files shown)


engine/src/flutter/shell/testing/tester_context_vk_factory.cc
```diff
@@ -0,0 +1,117 @@
...
+    FML_DCHECK(surface->IsValid());
+    return surface;
+  }
+
+ private:
+  std::shared_ptr<impeller::ContextVK> context_;
+  std::shared_ptr<impeller::SurfaceContextVK> surface_context_;
+};
+
+std::unique_ptr<TesterContext> TesterContextVKFactory::Create(
```
walley892, line 111: Why have a separate factory class here? This could be a member of the TesterContextVk

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
```diff
@@ -0,0 +1,117 @@
...
+          impeller_framebuffer_blend_shaders_vk_data,
+          impeller_framebuffer_blend_shaders_vk_length),
+      std::make_shared<fml::NonOwnedMapping>(
+          impeller_compute_shaders_vk_data, impeller_compute_shaders_vk_length),
+  };
+}
+
+class TesterContextVK : public TesterContext {
+ public:
+  TesterContextVK() = default;
```
walley892, line 43: Should this constructor be private so instances can't be obtained other than by the `Create` function?

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
```diff
@@ -0,0 +1,117 @@
...
+
+  ~TesterContextVK() override {
+    if (context_) {
+      context_->Shutdown();
+    }
+  }
+
+  bool Initialize(bool enable_validation) {
```
walley892, line 51: Instead of having a separate method here, can this just take the place of `Create`? It looks like Create now just calls this, checks the output, and returns a nullptr if the init didn't work.

walley892 (commented):
Couple of comments on structure

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
gaaclarke: This allows us to put all of the vulkan specific stuff inside the .cc file.  A file we only compile if vulkan is supported.  If we put create in the TesterContextVk we'd have to move that to the .h file which would require a lot of preprocessor macros.  Look at the code before I edited it.  There was preprocessor guards all over the place, sloppy.  Now I've factored everything so there is only 1 preprocessor guard.

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
gaaclarke: That's not necessary since the class isn't visible in any headers, therefore it's private to this file already.  This file was missing an anonymous namespace though to enforce that it doesn't collide with other symbols so I've added that.  That locks it down from any trickery to get access to it.

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
gaaclarke: You could, it would be a bit messy though because the factory doesn't have access to the private members of `TesterContextVK`.

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
walley892: I'm not sure what you mean. If we moved `Create` to be a member of `TesterContextVk`, the files would be `tester_context_vk.(h/cc)`, the header would contain signatures for all the methods here, and the cc file would have the implementations. I don't see anything in the method signatures that have anything Vulkan specific that would require an import of vulkan stuff into the header file.

I think there would still be only one point where you have to use the preprocessor guards (right where you have it now, but inside is `TesterContextVk::Create` instead of `TesterContextFactoryVk::Create`.

I may be missing something.

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
walley892: Yeah, moving the factory method to this class would be necessary for that.

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
walley892: This is also in conjunction with the below comment

engine/src/flutter/shell/testing/tester_context_vk_factory.cc
gaaclarke: Yea, check out the instance variables for TesterContextVk, they are both vulkan specific.  We'd need to put preprocessor guards around them.  By putting everything but the factory function in the header we don't have to mess around with that.

walley892: homer-simpson-doh.gif

you rite, LGTM!

walley892 (approved):
LGTM!

PR #180657, opened 2026-01-07
draggable_scrollable_sheet_test/focus_traversal_test/linked_scroll_view_test/list_view_test Material imports clean up + remove Divider from scroll_view_test

### LINKED ISSUE #180501: (Test cross-imports) About Divider material widget import for tests in `test/widgets`
### Use case

As mentioned in https://github.com/flutter/flutter/issues/177415#issue-3542356791:

> Divider | Can it just be removed? Or possibly we should create a RawDivider in Widgets if there is a real need.

1. There are 12 places using `Divider` material widget in widget tests:

Image

2. Divider  concept exists in both Material and iOS design specs/docs:
- Material: https://m3.material.io/components/divider/overview
- iOS design docs: https://developer.apple.com/documentation/SwiftUI/Divider

### Proposal

We can go with one of these options:
- **A. building a new RawDivider widget in widgets**:
    - Pros:
        - the import will be neat and tidy. Maintainers will only need to import RawDivider widget in widgets and corresponding design libraries on either widget usage or test.
        - allow other design systems to build their own divider based on RawDivider widget
        - follow existing patterns as other widgets: RawAutocomplete/Autocomplete,  RawRadio/Radio...
    - Cons:                                                       
        - Increase API maintenance cost (design docs needed?, documentation for itself and update Material one too, bug fixes and reviews, etc...)
        - Does this seem to be over-engineering? As the current Divider widget is simply a sized Container with border and theming
-  **B. adding a helper widget or something like that in a utils file in `test/widgets`**
    - Pros:
        - No risk and lower API maintenance cost: no need design docs, minimum documentation (only class and method docs are needed in new utils file in test directory only). If there are new issues introduced, the fix for them will be simpler than RawDivider widget 
        - Quickly solve the main goal of cleaning up cross-imports issue without getting sidetracked into other time-consuming tasks.
        - The test helper can be modified or removed at any time without considering breaking changes like RawDivider widget
    - Cons:                                                       
        - The helper widget may duplicate the "drawing a line" logic in the test utils, which is already implemented in the Divider material widget.

Based on the above analysis, I lean towards option B, as it helps achieve the goal of cleaning up cross-imports effectively at a lower/minimal cost. cc @justinmc for thoughts. Regardless of which option is chosen, I assign myself to this work.

navaronbracke:
This PR cleans up the following test files in flutter/widgets:
- draggable_scrollable_sheet_test.dart (Migrated to widgets)
- focus_traversal_test.dart (Migrated to widgets)
- linked_scroll_view_test.dart (Migrated to widgets)
- list_view_test.dart (Migrated to widgets)
- scroll_view_test.dart (Divider cleanup only)

The main purpose is to clean up usages of `Divider`, but I also cleaned up `Colors.primaries` and a bunch of MaterialApp / Scaffold/ElevatedButton/TextButton usages as well.

It also adds some test fixtures for future usage.

Fixes https://github.com/flutter/flutter/issues/180501
Part of https://github.com/flutter/flutter/issues/177415

## Pre-launch Checklist

- [x] I signed the [CLA].


navaronbracke:
@justinmc Do I update the dev/bots list entries for the cross import check in this PR?

packages/flutter/test/widgets/widget_test_fixtures.dart
```diff
@@ -0,0 +1,77 @@
...
+  Color(0xFF2196F3),
+  Color(0xFF4CAF50),
+  Color(0xFFFFEB3B),
+  Color(0xFFFF9800),
+];
+
+/// A basic divider widget, which draws a black horizontal line
+/// and is surrounded by some padding.
+@visibleForTesting
+class BasicDivider extends StatelessWidget {
```
justinmc, line 35: Since @Renzo-Olivares named his thing TestTextField in https://github.com/flutter/flutter/pull/180494, I think this should be TestDivider for consistency.

dev/bots/check_tests_cross_imports.dart
justinmc, line 1: When you land this, can you also check the boxes on https://github.com/flutter/flutter/issues/177415 for the test files that you removed here?

packages/flutter/test/widgets/widget_test_fixtures.dart
```diff
@@ -0,0 +1,77 @@
...
+/// A small set of colors for widget tests, inspired by the original Material 2 primary colors.
+@visibleForTesting
+const List<Color> genericWidgetTestColors = [
+  Color(0xFFF44336),
+  Color(0xFF9C27B0),
+  Color(0xFF2196F3),
+  Color(0xFF4CAF50),
+  Color(0xFFFFEB3B),
+  Color(0xFFFF9800),
+];
```
justinmc, line 30: We should get @victorsanni to check off on this. We want to make sure we make a decision on how to expose global test colors that will apply across Widgets/Material/Cupertino.

packages/flutter/test/widgets/widget_test_fixtures.dart
```diff
@@ -0,0 +1,77 @@
...
+
+Widget buildWidgetsApp(Widget child) {
+  return WidgetsApp(
+    color: const Color(0xFFFFFFFF),
+    home: child,
+    pageRouteBuilder: <T>(RouteSettings settings, WidgetBuilder builder) {
+      return TestPageRoute<T>(settings: settings, builder: builder);
+    },
+  );
+}
```
justinmc, line 19: Would this be better as TestWidgetsApp for consistency?

packages/flutter/test/widgets/widget_test_fixtures.dart
```diff
@@ -0,0 +1,77 @@
...
+    return Padding(
+      padding: const EdgeInsets.all(4),
+      child: Container(color: const Color(0xFF000000), height: 4, width: double.infinity),
+    );
+  }
+}
+
+/// A very basic widget that acts as a focusable button.
+@visibleForTesting
+class BasicButton extends StatelessWidget {
```
justinmc, line 49: Same here: TestButton. Long term we may want to introduce a RawButton in the Widgets library, but it should be fine to create a test one here for now.

packages/flutter/test/widgets/widget_test_fixtures.dart
justinmc, line 1: For consistency I would name this something ending in `_utils.dart`. Or, maybe this file could be split up into multiple files, but I guess it's pretty short.

packages/flutter/test/widgets/widget_test_route_fixtures.dart
justinmc, line 1: This should also be something `_utils.dart`. Maybe it should be combined with TestWidgetsApp?

packages/flutter/test/widgets/widget_test_route_fixtures.dart
```diff
@@ -0,0 +1,124 @@
...
+class TestPageRouteBuilder extends PageRouteBuilder<void> {
+  TestPageRouteBuilder({required Widget child})
+    : super(
+        pageBuilder: (BuildContext _, Animation<double> _, Animation<double> _) {
+          return child;
+        },
+      );
+}
+
+@visibleForTesting
```
justinmc, line 60: I think visibleForTesting isn't necessary for a file inside of `test/`. Or are you aware of anything that this is doing?

packages/flutter/test/widgets/focus_traversal_test.dart
```diff
@@ -3039,7 +3035,7 @@ void main() {
         final focusNodeLowerRight = FocusNode(debugLabel: 'lowerRight');
         addTearDown(focusNodeLowerRight.dispose);
 
-        Widget generatetestWidgets(bool ignoreTextFields) {
+        Widget generateTestWidgets(bool ignoreTextFields) {
```
justinmc, line 3038: Thanks for fixing the capitalization typo.

justinmc (commented):
It might be best to break this PR up into several smaller PRs. As it is, it solves https://github.com/flutter/flutter/issues/180501 plus partially solves a couple of other things, and it will probably require some deep thought and checkoffs from multiple people for the various utils that you added. Do you think it would be easy for you to break it up?

Otherwise though, thank you for jumping on helping out with these tests and getting the ball rolling on some of the utils classes.

packages/flutter/test/widgets/linked_scroll_view_test.dart
```diff
@@ -341,7 +342,7 @@ class _TestState extends State<Test> {
               ],
             ),
           ),
-          const Divider(),
+          const BasicDivider(),
```
justinmc, line 345: Is Divider/BasicDivider really necessary here, or could we get rid of it, or replace it with SizedBox or something? And same question for other uses of BasicDivider.

packages/flutter/test/widgets/widget_test_fixtures.dart
victorsanni: What's the rationale for choosing these specific colors? I searched and saw the list was used to replace Colors.primaries, should that be called out somewhere?

I see the list is used in a function to generate random colors, why isn't the function exposed instead?

packages/flutter/test/widgets/widget_test_fixtures.dart
huycozy: @navaronbracke Once `TestButton`created here, you can go for for `ElevatedButton` item as mentioned in https://github.com/flutter/flutter/issues/177415#issue-3542356791 as well.

huycozy (approved):
Thank you for this work! LGTM except for @justinmc's comments.

packages/flutter/test/widgets/widget_test_fixtures.dart
huycozy: I recommend splitting BasicButton change into a separate PR  for `ElevatedButton` item

dev/bots/check_tests_cross_imports.dart
navaronbracke: I don't have the right access rights to check those checkboxes in the list. So you'll have to check them for me, once these fixes land.

packages/flutter/test/widgets/widget_test_fixtures.dart
navaronbracke: The original tests used `Colors.primaries`, although this could probably be any kind of colors. I'll update the docs for the generate random colors method and probably use standard red/green/blue etc.

dev/bots/check_tests_cross_imports.dart
justinmc: Ah that makes sense. I'll try to be on top of that 👍

PR #180667, opened 2026-01-07
Add leadingIndex to CarouselController and onItemChanged to CarouselView

Mairramer:
Adds a `leadingIndex` property to CarouselController and an `onItemChanged` callback to CarouselView.
Closes #169207

## Pre-launch Checklist

- [x] I signed the [CLA].


Piinks:
Self-ping to my inbox for review. Thank you! 🙏

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -636,21 +682,35 @@ class _CarouselViewState extends State<CarouselView> {
...
-          viewportBuilder: (BuildContext context, ViewportOffset position) {
-            return Viewport(
-              cacheExtent: 0.0,
-              cacheExtentStyle: CacheExtentStyle.viewport,
-              axisDirection: axisDirection,
-              offset: position,
-              clipBehavior: Clip.antiAlias,
-              slivers: <Widget>[_buildSliverCarousel(theme)],
-            );
+        return NotificationListener<ScrollNotification>(
```
Piinks:
```suggestion
        return NotificationListener<ScrollUpdateNotification>(
```

It looks like we only care about ScrollUpdateNotification, so you could filter for only those and remove it from the conditional inside. However, why use a notification listener when you could use the controller?

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -518,6 +551,19 @@ class _CarouselViewState extends State<CarouselView> {
     super.dispose();
   }
 
+  // The initialItem means the index of the item to occupy the first maximum weight
```
Piinks: Does this contradict with the documentation above?

> /// For both [CarouselView] and [CarouselView.weighted], the leading item is
> /// the first visible item in the carousel view.

packages/flutter/test/material/carousel_test.dart
```diff
@@ -1977,6 +1977,214 @@ void main() {
     });
```
Piinks, line 1977: Can you add test coverage for when reverse is used?

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -448,6 +452,33 @@ class CarouselView extends StatefulWidget {
   /// The child widgets for the carousel.
   final List<Widget> children;
 
+  /// {@template flutter.material.CarouselView.onIndexChanged}
+  /// A callback invoked when the leading item changes.
+  ///
+  /// For both [CarouselView] and [CarouselView.weighted], the leading item is
```
QuncCccccc: Maybe we can simplify it to
```suggestion
  /// The leading item is
```

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -448,6 +452,33 @@ class CarouselView extends StatefulWidget {
...
+  /// the first visible item in the carousel view.
+  ///
+  /// The callback fires only when the leading index actually changes,
+  /// whether due to user interaction or programmatic scrolling.
```
QuncCccccc:
```suggestion
  /// The callback fires only when the leading item is completely out of view,
  /// whether due to user interaction or programmatic scrolling. If the leading item
  /// remains partially visible, the leading index will not change and the callback will 
  /// not be invoked.
```

packages/flutter/lib/src/material/carousel.dart
```diff
@@ -518,6 +551,19 @@ class _CarouselViewState extends State<CarouselView> {
     super.dispose();
   }
 
+  // The initialItem means the index of the item to occupy the first maximum weight
+  // when flexWeights is not null. So it might be negative when initialItem value
+  // is small but the first max weight index is large. In that case, the initial
+  // leading item should be 0.
```
QuncCccccc:
```suggestion
  // For weighted carousel, the initialItem means the index of the item to occupy the first maximum weight
  // in flexWeights. To get the initial leading item, it should be initialItem - index of the first max weight in flexWeights. So it might be negative when initialItem value
  // is small but the first max weight index is large. In that case, the initial
  // leading item should be 0.
```

QuncCccccc (approved):
LGTM with some doc comments:)

packages/flutter/lib/src/material/carousel.dart
Mairramer: I initially followed the same pattern used by `PageView`. But you’re right, in this case it’s better to rely on the controller instead of coupling the widget to a `NotificationListener`.

packages/flutter/lib/src/material/carousel.dart
Mairramer: Solved.

Piinks (approved):
LGTM

PR #180674, opened 2026-01-08
Make sure that a Focus doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the Focus widget.


victorsanni (commented):
Does Focus have its own renderbox? I took a look at the source code for _FocusState, in its build method it just wraps the child in an InheritedNotifier + Semantics.

dkwingsmt (approved):
LGTM. I think it doesn't hurt to add one.

victorsanni (approved):

PR #180692, opened 2026-01-08
Fix issue where web embedder is synthesizing key up events too eagerly

### LINKED ISSUE #162305: [BUG] [WEB] `KeyboardListener`'s and `Focus.onKeyEvent`'s `KeyUpEvent` do not work correctly in JS and WASM builds
### Steps to reproduce

The `KeyboardListener` and `Focus.onKeyEvent` that replace `RawKeyboardListener` and `Focus.onKey` have a bug in triggering the `KeyUpEvent` correctly in the JS and WASM builds. The `KeyUpEvent` is not triggered when a key is released when pressing multiple keys in sequence and releasing them in reverse order. This bug is not present in VM builds.

A reproduction issue sample app is provided in the code sample section.

This issue reproduces on all current Flutter channels.

## Background

We discovered this bug after starting to more actively pushing and using WEB builds of the game platform [HypeHype](https://app.hypehype.com/).
We had already migrated from the deprecated `RawKeyboardListener` to the new `KeyboardListener` and its related APIs, but due to this issue we had to revert the migration and continue to use the deprecated `RawKeyboardListener` and its related APIs.

When making games, having correct n-key rollover function and n-key release sequence is critical, as game-play typically depends on being able to press multiple keys at the same time and having the game react correctly to their triggered key down and key up release order.

## Request to hold removal of deprecated `RawKeyboardListener` and related APIs until issue is fixed

We kindly ask if you can hold the removal of the deprecated `RawKeyboardListener` and its related APIs, until this issue is resolved and its fix has landed in the Flutter stable channel.

The  `RawKeyboardListener` and its related APIs are scheduled for removal here https://github.com/flutter/flutter/issues/136419 by gspencergoog.

### Expected results

When pressing multiple keys in sequence and releasing them in reverse order, the `KeyUpEvent` should be triggered for each key in the correct release order.

Using the reproduction keyboard event listener demo app, we can see that in a **VM build** the `KeyUpEvent` is triggered correctly for each key in the correct order.

**Using the VM build with `RawKeyboardListener` or `Focus.onKey` APIs**
1. Press and hold `[1]` then, press and hold `[2]` then press and hold `[3]`.
2. Release `[3]`, keep holding `[2]` and `[1]`.
3. Release `[2]`, keep holding `[1]`, the `RawKeyUpEvent` for `[2]` is triggered.
4. Release `[1]`, the `RawKeyUpEvent` for `[1]` is triggered.

This is CORRECT and **expected** behaviour.

**Using the VM build with `KeyboardListener` or `Focus.onKeyEvent` APIs**
1. Press and hold `[1]` then, press and hold `2` then press and hold `[3]`.
2. Release `[3]`, keep holding `[2]` and `[1]`.
3. Release `[2]`, keep holding `[1]`, the `KeyUpEvent` for `[2]` is triggered.
4. Release `[1]`, the `KeyUpEvent` for `[1]` is triggered.

This is CORRECT and **expected** behavior.

This is demonstrated in the video recording below:

https://github.com/user-attachments/assets/7a449594-746c-4528-ace3-cbf0e0658907

### Actual results

When pressing multiple keys in sequence and releasing them in reverse order, the `KeyUpEvent` is not triggered for each key in the correct order on WEB builds. The same incorrect result is present in both JS and WASM builds. The previous key that were pressed are triggered automatically when last key is released, slightly after it, or not at all.

**Using WEB VM build with `RawKeyboardListener` or `Focus.onKey` APIs**
1. Press and hold `[1]` then, press and hold `[2]` then press and hold `[3]`.
2. Release `[3]`, keep holding `[2]` and `[1]`.
3. Release `[2]`, keep holding `[1]`, the `RawKeyUpEvent` for `[2]` is triggered.
4. Release `[1]`, the `RawKeyUpEvent` for `[1]` is triggered.

This is CORRECT and **expected** behavior.

**Using the VM build with `KeyboardListener` or `Focus.onKeyEvent` APIs**
1. Press and hold `[1]` then, press and hold `[2]` then press and hold `[3]`.
2. Release `[3]`, keep holding `[2]` and `[1]`, the `KeyUpEvent` for `[2]` is triggered shortly after, despite `[2]` still being pressed.
3. Release `[2]`, keep holding `[1]`, the actual `KeyUpEvent` for `[2]` is not triggered.
4. Release `[1]`, the `KeyUpEvent` for `[1]` is not triggered.

This is incorrect and **not expected** behaviour when using the `KeyboardListener` and related APIs on WEB builds.

This is demonstrated in the video recording below:

https://github.com/user-attachments/assets/5ff9655c-e516-4be7-86e7-a989a61b606d

The recording shows a Web WASM build, the result is the same in a Web JS build.

### Code sample

```dart

void main() => runApp(const KeyboardListenerComparisonApp());

// Check if this is a Web-WASM build, Web-JS build or native VM build.
const bool isRunningWithWasm = bool.fromEnvironment('dart.tool.dart2wasm');
const String buildType = isRunningWithWasm
    ? '(WASM build)'
    : kIsWeb
        ? '(JS build)'
        : '(VM build)';

class KeyboardListenerComparisonApp extends StatelessWidget {
  const KeyboardListenerComparisonApp({super.key});

  @override
    return MaterialApp(
      home: Scaffold(
        body: const KeyboardListenerDemo(),
      ),
    );
  }
}

class KeyboardListenerDemo extends StatefulWidget {
  const KeyboardListenerDemo({super.key});

  @override
  State<KeyboardListenerDemo> createState() => _KeyboardListenerDemoState();
}

class _KeyboardListenerDemoState extends State<KeyboardListenerDemo> {
  final List<String> _rawEvents = [];
  final List<String> _newEvents = [];
  late final FocusNode _rawFocusNode;
  late final FocusNode _newFocusNode;
  bool _useNewListener = false;

  @override
  void initState() {
    super.initState();
    _rawFocusNode = FocusNode(debugLabel: 'RawKeyboardListener');
    _newFocusNode = FocusNode(debugLabel: 'KeyboardListener');
    _rawFocusNode.addListener(_handleFocusChange);
    _newFocusNode.addListener(_handleFocusChange);
    _rawFocusNode.requestFocus();
  }

  @override
  void dispose() {
    _rawFocusNode.removeListener(_handleFocusChange);
    _newFocusNode.removeListener(_handleFocusChange);
    _rawFocusNode.dispose();
    _newFocusNode.dispose();
    super.dispose();
  }

  void _handleFocusChange() => setState(() {});

  void _clearEvents() {
    setState(() {
      _rawEvents.clear();
      _newEvents.clear();
    });
  }

  void _switchListener(bool useNew) {
    setState(() => _useNewListener = useNew);
    if (useNew) {
      _newFocusNode.requestFocus();
      _rawFocusNode.unfocus();
    } else {
      _rawFocusNode.requestFocus();
      _newFocusNode.unfocus();
    }
  }

  @override
    return Column(
      children: [
        AppBar(
          title: const Text('Keyboard Listeners Comparison $buildType'),
          actions: [
            IconButton(
              icon: const Icon(Icons.delete),
              onPressed: _clearEvents,
              tooltip: 'Clear all events',
            ),
          ],
        ),
        Expanded(
          child: Row(
            children: [
              _buildListenerColumn(
                context: context,
                title: 'RawKeyboardListener',
                events: _rawEvents,
                isActive: _rawFocusNode.hasFocus,
                onTap: () => _switchListener(false),
              ),
              _buildListenerColumn(
                context: context,
                title: 'KeyboardListener',
                events: _newEvents,
                isActive: _newFocusNode.hasFocus,
                onTap: () => _switchListener(true),
              ),
            ],
          ),
        ),
        // Raw Keyboard Events (Legacy)
        if (!_useNewListener)
          Focus(
            focusNode: _rawFocusNode,
            onKey: (FocusNode node, RawKeyEvent event) {
              if (event is RawKeyDownEvent || event is RawKeyUpEvent) {
                _handleKeyEvent(event, isNew: false);
              }
              return KeyEventResult.handled; // Suppress system sounds
            },
            child: const SizedBox.shrink(),
          ),
        // New Keyboard Events
        if (_useNewListener)
          Focus(
            focusNode: _newFocusNode,
            onKeyEvent: (FocusNode node, KeyEvent event) {
              _handleKeyEvent(event, isNew: true);
              return KeyEventResult.handled; // Suppress system sounds
            },
            child: const SizedBox.shrink(),
          ),
      ],
    );
  }

  void _handleKeyEvent(dynamic event, {required bool isNew}) {
    final timeStamp = DateTime.now().toIso8601String().substring(11, 23);
    final entry = '[${event.runtimeType}] ${event.logicalKey} ($timeStamp)';

    setState(() {
      if (isNew) {
        _newEvents.add(entry);
      } else {
        _rawEvents.add(entry);
      }
    });
  }

  Widget _buildListenerColumn({
    required BuildContext context,
    required String title,
    required List<String> events,
    required bool isActive,
    required VoidCallback onTap,
  }) {
    final colorScheme = Theme.of(context).colorScheme;

    return Expanded(
      child: Padding(
        child: MouseRegion(
          cursor: SystemMouseCursors.click,
          child: GestureDetector(
            onTap: onTap,
            behavior: HitTestBehavior.opaque,
            child: Container(
              decoration: BoxDecoration(
                border: Border.all(
                  color: isActive
                      ? colorScheme.primary
                      : colorScheme.outlineVariant,
                  width: 4,
                ),
                borderRadius: BorderRadius.circular(8),
              ),
              padding: const EdgeInsets.all(8),
              child: Column(
                children: [
                  Text(
                    title,
                    style: TextStyle(
                      color: isActive
                          ? colorScheme.primary
                          : colorScheme.onSurface,
                      fontWeight: FontWeight.bold,
                      fontSize: 16,
                    ),
                  ),
                  Expanded(
                    child: ListView.builder(
                      reverse: true,
                      itemCount: events.length,
                      itemBuilder: (context, index) => Padding(
                        padding: const EdgeInsets.symmetric(vertical: 4),
                        child: Text(
                          events.reversed.elementAt(index),
                          style: TextStyle(
                            color: colorScheme.onSurfaceVariant,
                            fontSize: 12,
                          ),
                        ),
                      ),
                    ),
                  ),
                ],
              ),
            ),
          ),
        ),
      ),
    );
  }
}

```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

## Flutter version

This issue reproduces on all current Flutter channels, the latest tested master version is `3.29.0-1.0.pre.127`.

```console
flutter doctor -v
[✓] Flutter (Channel master, 3.29.0-1.0.pre.127, on macOS 15.2 24C101 darwin-arm64, locale en-US) [1,489ms]
    • Flutter version 3.29.0-1.0.pre.127 on channel master at /Users/rydmike/fvm/versions/master
    • Framework revision 9e273d5e6e (6 hours ago), 2025-01-27 19:43:46 -0800
    • Engine revision 9e273d5e6e
    • Dart version 3.8.0 (build 3.8.0-24.0.dev)
    • DevTools version 2.42.0

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0) [902ms]
    • Android SDK at /Users/rydmike/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [517ms]
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [68ms]

[✓] Android Studio (version 2023.2) [68ms]
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)

[✓] IntelliJ IDEA Community Edition (version 2024.3.2) [67ms]
    • IntelliJ at /Applications/IntelliJ IDEA CE.app
    • Flutter plugin version 83.0.4
    • Dart plugin version 243.23654.44

[✓] VS Code (version 1.96.4) [10ms]
    • Flutter extension version 3.102.0

[✓] Connected device (2 available) [5.6s]
    • macOS (desktop) • macos  • darwin-arm64   • macOS 15.2 24C101 darwin-arm64
    • Chrome (web)    • chrome • web-javascript • Google Chrome 132.0.6834.110

[✓] Network resources [618ms]
```

</details>

O-Hannonen:
This PR fixes #162305

The reason this issue was visible in `Focus.onKeyEvent` but not in the legacy `Focus.onKey` is that the focus manager silently ignores the problematic synthesized messages for `FocusNode.onKey` as they don't contain a raw event:
image

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/lib/web_ui/lib/src/engine/keyboard_binding.dart
```diff
@@ -266,18 +277,31 @@ class KeyboardConverter {
...
+  /// state because it thinks the key was never released.
+  ///
+  /// To avoid this, we rely on the fact that browsers send repeat events
+  /// while the key is held down by the user. If we don't receive a repeat
+  /// event within a specific duration [_kKeydownCancelDurationMac] we assume
+  /// the user has released the key and we synthesize a keyup event. A limitation
+  /// in this approach is that if multiple non-modifier keys are pressed at the same time,
+  /// only the last pressed non-modifier key will send repeated events. This means that
+  /// the guard will eventually synthesize keyup events for the other keys, even if they
+  /// are still pressed.
```
O-Hannonen: The same issue fixed in this PR is also present in key combos with one or more modifier keys and multiple non-modifier keys (eg. cmd+i+j). In practice this means eg.
1. Press and hold cmd
2. Press and hold a
3. Press and hold b
4. Wait for over 2 seconds (>_kKeydownCancelDurationMac)
Expected events: cmd down, a down, b down
Actual events: cmd down, a down, b down, a up (synthesized)

This is a bit shame but at least I could not think of ways to work around this, any ideas are welcome. One option would be to delay the synthesizing of those non-modifier keys until the modifier is released (so the debouncing could be fully removed). But not sure if that would be any better as with that option too the framework might be out of sync for the non-modifier keys until the modifier is released.

mdebbar: Here's an idea:

Right now, we have a key guard for each physical, and we track repeat events for each physical key individually like you said. What if we make it a single key guard for all of the pressed keys. If any of the pressed keys is repeated, we keep restarting the single key guard. Do you think this could work?

O-Hannonen: That could work and would be slightly better than my original suggestion in this thread. In your suggestion releasing all other keys than the modifier key will synthesize the release of the non-modifier keys, whereas in my suggestion they'd be stuck as pressed until modifier is released. 🤔

However, both of the approaches have some drawbacks:
- The real state of keyboard might be out of sync with the state in the framework. In my suggestion all non-modifier keys could be released and the framework would think they're pressed until modifier is released. In your suggestion all but one non-modifier keys could be released and the framework would think they're pressed until the final non-modifier is released.
- The real release order is lost. We have no idea in which order the keys were actually released, so the synthesized events might (and most probably will) be fired in the wrong order. Not sure if we can do anything about this drawback as we're limited by the browsers events, which are pretty bad to begin with when modifier key is involved.

Should I open another issue for this sub-issue when the modifier keys are involved? I think we could still go forward with this PR as it fixes the issue perfectly when no modifier is involved (might already offer the necessary improvement for many game-devs). Whatever the solution for modifier key-combos will be, it will be a workaround. The browser already sends the key events perfectly when no modifier keys are involved, so we should handle those separately from the workaround (as we do in this PR).

mdebbar: > In your suggestion all but one non-modifier keys could be released and the framework would think they're pressed until the final non-modifier is released.

Is this true?

Assume keys "R", "S" and "T" are pressed. The browser keeps sending repeat events for one of them, say "T".

When "R", for example, is released, the browser sends a "keyup" event, the engine will translate that to a `ui.KeyEventType.up`, the framework will know that "R" is no longer pressed.

I'm not sure why you say "the framework would think they're pressed until the final non-modifier is released". Could you clarify please?

O-Hannonen: > Assume keys "R", "S" and "T" are pressed. The browser keeps sending repeat events for one of them, say "T".
>
> When "R", for example, is released, the browser sends a "keyup" event, the engine will translate that to a ui.KeyEventType.up, the framework will know that "R" is no longer pressed.

Yes, this is true when no modifier key is involved. But its not true if one or more modifier is pressed.

How I got to this conclusion?

1. Print the key up and key down dom events

image

2. Press these keys in this order in the sample app

- cmd down
- j down
- k down
- j up
- k up
- cmd up

3. See the logs, only key up event logged is the one for cmd

image

With your suggested solution (if I understood it correctly) releasing the last non-modifier key will synthesize the key up event for all non-modifier keys. This means that as long as one or more non-modifier key is pressed, no key-up events will be synthesized (even if some keys are released) -> thus "the framework would think they're pressed until the final non-modifier is released".

mdebbar: Oh wow, that's weird. Looks like the behavior with Meta is different from other modifier keys. When Meta is pressed, the browser doesn't send `keyup` events for other keys.

We should make sure that the case of Meta + one key is handled correctly.

Other cases that involve Meta + multiple keys are impossible to handle 100% correctly. Any solution we pick is going to have downsides, and I don't have a strong opinion on which solution we should go with. I'll accept your PR as it's simpler than what I suggested earlier.

mdebbar (approved):

engine/src/flutter/lib/web_ui/lib/src/engine/keyboard_binding.dart
O-Hannonen: ### Implementation improvements

Oh, did some more debugging and that seems to be right, its *only* meta that behaves differently, not all modifiers. The changes in this PR were done under the assumption that its all modifiers. I'll modify the implementation to match with the results of this debugging. Good catch 🎣, don't know how I ended up thinking its all modifiers 😅

### The meta limitations

```dart
  /// A limitation in this approach is that if meta key + multiple non-modifier keys are
  /// pressed at the same time, only the last pressed non-modifier key will send repeated
  /// events. This means that the guard will eventually synthesize keyup events for the other
  /// keys, even if they are still pressed.
```
Just to be clear, the changes in this PR don't yet do anything discussed in this thread for the meta + 2 or more non-modifier keys. The changes simply skip the key-up event debouncing if meta is not pressed, since in that case we can trust the events browser is sending. So with the changes in this PR, the same issue reported in the linked issue and documented in the comment in code still exist if meta-key is involved. So this will still happen, even though it probably should not:
1. cmd down
2. j down
3. k down
4. wait for 1500ms
5. j up is synthesized, even though its still pressed

Should I try to fix the meta-key too in this PR, or open a separate issue for it? I think separate issue might be good, since its impossible to handle 100% correctly -> might be worth asking around opinions on which of the ways to handle it is the least bad. The workaround for the meta-case anyways has to be kept separate from the changes in this PR (the non-meta case) as that we can handle 100% correctly by trusting the browser events.

mdebbar: I think this PR is good as is. It fixes the general issue with multiple key presses.

Pressing Meta + multiple keys while holding and releasing randomly is not a common pattern. Usually Meta is used for activating shortcuts, which typically involve pressing Meta + key 1 + key 2 then quickly releasing everything.

The issue of holding multiple keys and releasing some of them and then pressing others, sounds like a gaming thing. I imagine those keys usually don't include Meta.

I say let's merge this PR as is, and create a new issue for the Meta case with detailed explanation from this thread. We can fix it separately later. What do you think?

O-Hannonen: Sounds good, I'll open another issue! Thanks for the review 🎉

engine/src/flutter/lib/web_ui/test_output.txt
mdebbar, line 1: This file should be deleted.

kevmoo (commented):
Love to see the tests!

kevmoo (approved):
LGTM

PR #180699, opened 2026-01-08
[web] Don't serve files outside of project

mdebbar:
Fixes internal bug: b/463611148

### DIFF
--- a/packages/flutter_tools/test/general.shard/web/devfs_web_test.dart
+++ b/packages/flutter_tools/test/general.shard/web/devfs_web_test.dart
@@ -1455,6 +1455,64 @@ void main() {
   );
 
   test(
+    'ReleaseAssetServer does not serve files outside of the project',
+    () => testbed.run(() async {
+      const testHomePath = '/home/user';
+      const testFlutterRoot = '/home/user/flutter';
+      const testProjectPath = '/home/user/project';
+      const testWebBuildPath = '/home/user/project/build/web';
+
+      final platform = FakePlatform();
+      platform.environment = <String, String>{'HOME': testHomePath};
+
+      // The secret file that should not be accessible.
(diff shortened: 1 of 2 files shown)


packages/flutter_tools/test/general.shard/web/devfs_web_test.dart
```diff
@@ -1454,6 +1454,64 @@ void main() {
...
+        '../../user/secret.txt',
+        // An absolute path.
+        '/home/user/secret.txt',
+      ];
+
+      for (final path in possibleSecretPaths) {
+        final Response response = await server.handle(
+          Request('GET', Uri.parse('http://foobar/$path')),
+        );
+        expect(response.statusCode, 200, reason: 'Path "$path" should return 200 and index.html');
```
harryterkelsen, line 1505: So the expected behavior is for any URL that doesn't point to an existing file under the project directory to return index.html? Is it possible to 404?

mdebbar: We can and in fact, that's what we used to do in the past. But then `PathUrlStrategy` doesn't work.

If the app defines routes such as `/settings`, `/users/123`, etc, you want to be able to hit refresh at any of those paths and be served with the `index.html` file.

With the `HashUrlStrategy`, the url in the address bar becomes `http://localhost:xxxx/#/users/123`. Hitting refresh will request `/` from the server, which correctly serves `index.html`.

But with `PathUrlStrategy`, the url in the address bar becomes `http://localhost:xxxx/users/123`. Hitting refresh will request `/users/123` from the server, which returns 404 since there's no file called `users/123` in the file system.

harryterkelsen: Ah, that makes sense. Thanks

harryterkelsen (approved):
LGTM

PR #180715, opened 2026-01-09
Make sure that a FocusScope doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the FocusScope widget.


victorsanni (commented):
Does FocusScope have its own renderbox? I took a look at the source code and it just wraps its child in an InheritedNotifier + Semantics. cc @dkwingsmt

dkwingsmt (approved):
LGTM. I think it's fine. It's to guarantee that we don't make any significant changes that involve layout in the future.

victorsanni (approved):

PR #180752, opened 2026-01-09
Add `clearError` API to Form and FormFieldState

### LINKED ISSUE #178805: Add a way to clear the _errorText.value in FormFieldState without having to call .reset()
### Use case

For a custom Form, if we have a field showing an error after calling `validate()`, it could be required that modifying a different field changes the state of the Form, including the first invalid one.
If we use the auto validate mode, the field will be still invalid, but we may not want to show that to the user until they make a different action (like pressing a button).

This can be somewhat accomplished by calling `reset()`  and then `didChange(previousValue)` but it seems more like a workaround than the right way of just clearing the error.

### Proposal

To add a new method `clearError()` which would just set the `_errorText.value` to `null`, similar to the `reset()` method without modifying the field value.

Mairramer:
This PR introduces a `clearError` API for `FormFieldState` and `FormState`, allowing
validation errors to be cleared without resetting field values or re-running
validation.

Closes #178805

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/form.dart
```diff
@@ -649,13 +657,23 @@ class FormFieldState<T> extends State<FormField<T>> with RestorationMixin {
...
+      _clearErrorInternal();
     });
     widget.onReset?.call();
     Form.maybeOf(context)?._fieldDidChange();
   }
 
+  /// Clears any visible validation error for this field without resetting the field's value.
+  ///
+  /// Resets [errorText] to `null` and marks [hasInteractedByUser] as false,
+  /// effectively hiding any validation messages until the next validation or interaction.
```
justinmc: I think this will depend on the AutovalidateMode, is that right? Maybe just change these docs to acknowledge that?

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -333,6 +333,14 @@ class FormState extends State<Form> {
     _fieldDidChange();
   }
 
+  /// Clears the validation errors for all [FormField]s in this [Form]
+  /// without resetting their values.
+  void clearError() {
+    for (final FormFieldState<dynamic> field in _fields) {
+      field.clearError();
+    }
```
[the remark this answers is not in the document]
justinmc: I think Gemini may be right here, do you think it would be possible and useful to only call fieldDidChange once?

justinmc (commented):
I see how this is useful and not doable without an API change, I'm on board. Some questions below though.

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -333,6 +333,15 @@ class FormState extends State<Form> {
     _fieldDidChange();
   }
 
+  /// Clears the validation errors for all [FormField]s in this [Form]
+  /// without resetting their values.
```
justinmc, line 337: Nit: Maybe add a "See also" section with a reference to FormField.clearError.

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -649,13 +658,25 @@ class FormFieldState<T> extends State<FormField<T>> with RestorationMixin {
...
     widget.onReset?.call();
     Form.maybeOf(context)?._fieldDidChange();
   }
 
+  /// Clears any visible validation error for this field without resetting the field's value.
+  ///
+  /// This sets [errorText] to null and [hasInteractedByUser] to false.
+  ///
+  /// If [AutovalidateMode.always] is used, the error may reappear immediately
+  /// because the field will trigger a new validation cycle during the next build.
```
justinmc, line 676: Here you should probably also link back to Form.clearError.

justinmc (approved):
LGTM with nits :+1:

packages/flutter/lib/src/widgets/form.dart
Piinks: I think this is still outstanding to add the 'See also' here.

Piinks (approved):
LGTM with the doc nit remaining to land.

packages/flutter/lib/src/widgets/form.dart
Mairramer: Done.

packages/flutter_tools/lib/src/test/test_compiler.dart
```diff
@@ -102,7 +102,7 @@ class TestCompiler {
   ///
   /// If [testTimeRecorder] is passed, times will be recorded in it.
   TestCompiler(
-    this.buildInfo,
+    BuildInfo buildInfo,
```
Piinks, line 105: Hmm, it looks like this got pulled in by accident?

Piinks (changes requested):

packages/flutter_tools/lib/src/test/test_compiler.dart
Mairramer: I just did a rebase. It came from this PR: https://github.com/flutter/flutter/pull/181826

Mairramer: It's really strange that this showed up here.

Mairramer: Fixed it. That was weird, but it's all good now.

Piinks (approved):
Great thanks!

PR #180815, opened 2026-01-11
Add fields getter to FormState

### LINKED ISSUE #67283: Make "fields" in FormState not a private member
## Use case

I want to implement feature that when form validated fail scroll to that FormField.

currently I implement with copy `form.dart` file  and change `_field` to `field` to make it public

and do some validation like [document](https://flutter.dev/docs/cookbook/forms/validation) said

and scroll to specify field with 

```dart
formKey.currentState.validate();
Scrollable.ensureVisible(formKey.currentState.fields.first.context,
        duration: Duration(milliseconds: 500));
```

## Proposal

form.dart
```dart
...
class FormState extends State<Form> {
  int _generation = 0;
  final Set<FormFieldState<dynamic>> fields = <FormFieldState<dynamic>>{};
...
```

What do you think ?

Thank you.

### LINKED ISSUE #107528: [proposal] Provide a way to retrieve the data of all textfields within a `Form` widget
## Use case

the Flutter `Form` widget can be extremely powerful and has a lot of potential to simplify form implementations. Sadly I'm finding myself having to create my own Form widgets because there is not simple build-in way to aggregate all the form data in one place. Just like the form can interact with all its `FormField` children through validation, saving, ... it should also have an option to submit it's childrens data. Similar to how HTML handles form submission.

## Proposal

since the `FormState` already has all its `FormFieldState` children registered it would be trivial to collect all their values. the implementation difficulties come from returning the form data in a manageable way.

I would propose adding a `submissionKey` string to the `FormField<T>` base class. this allows us to aggregate all the `FormField` data under a `Map<String, dynamic>` where the entries would be shaped as: `submissionKey:  value`. 

this would require all `FormField` to have a non-null submissionKey. This can be hande by either throwing an exception when calling a form submit on a null submissionKey, or not adding `FormField` values to the end `Form` submit data if the submissionKey is null.

This would be a minor addition to the framework with no breaking or changing code, that would still have enormous benefits.

comparing the current closest implementation of this:
```dart
Form(
  child: Column(
    children: [
      TextField( 
        onSave: (value) {
          data['name'] = value;
        },
      ),
      TextField(
        onSave: (value) {
          data['email'] = value;
        },
      ),
      TextField(
         onSave: (value) {
          data['message'] = value;
        },
      ),
      MaterialButton(
        child: Text('submit'),
        onPressed: () {
          final form = Form.of(context);
          if(!form.validate()) return;

          form.save();
          /// handle data in stateful widget...
        }
      ),
    ],
  ),    
);
```

would be turned into something like this: 
```dart
Form(
  child: Column(
    children: [
      TextField(
        submissionKey: 'name',
        ...
      ),
      TextField(
        submissionKey: 'email',
        ...
      ),
      TextField(
        submissionKey: 'message',
        ...
      ),
      MaterialButton(
        child: Text('submit'),
        onPressed: () {
          final form = Form.of(context);
          if(!form.validate()) return;
          
          final data = form.submit();
          // final object = FormObject.fromJson(data); ...
        }
      );
    ],
  ),    
);

```

### LINKED ISSUE #108110: [Proposal]Add `isValid` method to FormState
This feature was already requested in https://github.com/flutter/flutter/issues/48874 , and was partially developed in https://github.com/flutter/flutter/pull/48948
However, the part about having the `isValid` method on the FormState was rejected, because if we use it to disable the "submit" button, the user might not know what the errors are. https://github.com/flutter/flutter/pull/48948#issuecomment-584234140
I believe this was a good choice back then, but now we have `autovalidateMode: AutovalidateMode.onUserInteraction` where the error message would show up at the correct time. 
Therefore I'm asking you to re-evaluate this feature request.
@tjonganthony I copied your original Use case and Proposal which were perfect descriptions. I hope it's fine.

## Use case

`FormState.validate` causes all field to be validated even though the field has never been touched. There are cases when I want to only check whether the current form is valid without triggering UI changes on all the field.

For example:
I have a `Form` with 15 `FormField` and a submit button. I want to enable/disable the submit button based on whether there's a field that is notValid.
so I added `onChanged: _updateSubmitButton()` and `_updateSubmitButton()` will use the value from `formKey.currentState.validate()` to determine whether to enable or disable the submit button.

This will cause changing field 1 to make all the 15 fields to be validated and maybe show error just in order for me to disable the submit button.

## Proposal

Add an `isValid()` method to `FormState` that will check whether all the field is valid.
This will also require `isValid()` to be added to `FormFieldState` in order not to rely on `FormFieldState.validate()` which will cause the error text to appear.

### LINKED ISSUE #110760: exposing `FormState` `_fields` value
## Use case

the form widget can be a powerful tool as it caches and can interact with all its child `FormField` without having to manually track them through keys. this is useful for a couple build in features like submitting & saving. but blocks the user from creating it's own implementations like for example https://github.com/flutter/flutter/pull/107530

## Proposal

make the FormState `_fields` value into a public `field` so we can use it to extend its functionality

```dart
/// State associated with a [Form] widget.
///
/// A [FormState] object can be used to [save], [reset], and [validate] every
/// [FormField] that is a descendant of the associated [Form].
///
/// Typically obtained via [Form.of].
class FormState extends State<Form> {

  final Set<FormFieldState<dynamic>> fields = <FormFieldState<dynamic>>{};
  
   ...

}
```

this issue would allow us to implement our own extensions that would, for example, make https://github.com/flutter/flutter/issues/107528 doable by the user.

Mairramer:
This PR exposes a public `FormState.fields` getter that returns all registered
`FormFieldState` instances associated with a `Form`.

Fixes flutter/flutter#110760
Fixes  flutter/flutter#67283
Partially fixes flutter/flutter#107528
Partially fixes flutter/flutter#108110

## Pre-launch Checklist

- [x] I signed the [CLA].


justinmc (commented):
@Mairramer could you explain your concerns you posted in https://github.com/flutter/flutter/issues/110760#issuecomment-3730089050 and if they apply here?

Five years ago I decided not to make this change (https://github.com/flutter/flutter/issues/67283#issuecomment-712220425), but in the meantime I've become less strict about the Form API, as I hope to move to a replacement at some point.

Do you have any reason to make this change that my [workaround](https://dartpad.dev/?id=120893372689ce66bbf89e5178848834) I posted in that issue wouldn't solve?

packages/flutter/test/widgets/form_test.dart
```diff
@@ -1830,6 +1830,152 @@ void main() {
...
+
+    formKey.currentState?.validate();
+    await tester.pump();
+
+    expect(find.text(errorText('foo')), findsOneWidget);
+    expect(find.text(errorText('bar')), findsOneWidget);
+
+    final List<FormFieldState<dynamic>> fields = formKey.currentState!.fields.toList();
+
+    expect(fields.every((field) => field.isValid), isFalse);
```
loic-sharma: Should this be `fields.any(...)` instead of `fields.every(...)`? With `.every`, this will pass if at least one it not valid. With `.any`, this would pass only if all are not valid.

packages/flutter/test/widgets/form_test.dart
```diff
@@ -1830,6 +1830,152 @@ void main() {
...
+    );
+
+    Map<String, dynamic> collectData() {
+      return {
+        for (final field in formKey.currentState!.fields)
+          if (field.widget.key case ValueKey<String>(:final value)) value: field.value,
+      };
+    }
+
+    expect(collectData(), {'name': 'Name', 'email': 'Email', 'age': 18, 'animal': 'cat'});
```
loic-sharma: Per Flutter's style guide, please explicitly type map literals: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-var-and-dynamic

Something like:

```suggestion
    expect(collectData(), <String, Object>{'name': 'Name', 'email': 'Email', 'age': 18, 'animal': 'cat'});
```

Same comment below for line 1972.

packages/flutter/test/widgets/form_test.dart
Mairramer: The use of `every` is correct, but the way I wrote it introduces a slight semantic inconsistency.

loic-sharma (approved):
The implementation + tests look good to me. However, I'll defer to @justinmc on whether we should move forward with the API. Please also get an approval from @justinmc before landing this :)

justinmc (approved):
LGTM 👍 

@Mairramer I really appreciate your persistence in this PR and all your work answering my questions, including the Dartpad example. You've clearly explained the upsides and I can't think of any concrete downsides besides the maintenance burden, which is small. So let's land it.

PR #156708, opened 2024-10-14
Fixes initial validation with AutovalidateMode.always on first build

Mairramer:
Fixes #142701
This PR fixes an issue where on the first build `AutovalidateMode.always` was not called.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #180822, opened 2026-01-11
Add useStrictAutovalidateMode to respect per-field autovalidateMode

### LINKED ISSUE #125766: `TextFormField` do not disabled validation when set AutovalidateMode.disabled and Form parent has `autovalidateMode` != AutovalidateMode.disabled

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)

### Steps to reproduce

1. Run code sample
2. Write text into `TextFormField`
3. Error: Show error even `autovalidateMode` on TextFormField is disabled

### Expected results

When `TextFormField` has the property `autovalidateMode` with value **AutovalidateMode.disabled**, the field won't show the error, event with property `autovalidateMode` with **AutovalidateMode.always** or **AutovalidateMode.onUserInteraction** on `Form` parent widget

This would be _follow the Flutter's inherited widget convention were the innermost takes priority_ (https://github.com/flutter/flutter/issues/107350#issuecomment-1295694555)

### Actual results

When `Form` parent widget has property `autovalidateMode` with **AutovalidateMode.always** or **AutovalidateMode.onUserInteraction** and `TextFormField` has the property `autovalidateMode` with value **AutovalidateMode.disabled**; the `TextFormField` validate after the first user validation, like **AutovalidateMode.onUserInteraction**

### Code sample

<details><summary>Code sample</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
    );
  }
}

  final String title;

  @override
}

  final _formKey = GlobalKey<FormState>();

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Center(
        child: Padding(
          padding: const EdgeInsets.all(24.0),
          child: Column(
            children: <Widget>[
              Form(
                autovalidateMode: AutovalidateMode.always,
                child: TextFormField(
                  autovalidateMode: AutovalidateMode.disabled,
                  validator: (value) =>
                      value == null || value.length < 4 ? 'Not valid' : null,
                ),
              ),
            ],
          ),
        ),
      ),
        onPressed: () => _formKey.currentState!.validate(),
        tooltip: 'Validate',
        child: const Icon(Icons.check),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details>

[FORM-ISSUE.webm](https://user-images.githubusercontent.com/21011641/235349942-02b677cf-7de5-4eb6-925e-b6059b45c3dd.webm)

</details>

### Logs

<details><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.7.12, on Ubuntu 22.04.2 LTS 5.19.0-41-generic, locale en_US.UTF-8)
    • Flutter version 3.7.12 on channel stable at /home/matias/develop/flutter/stable
    • Framework revision 4d9e56e694 (12 days ago), 2023-04-17 21:47:46 -0400
    • Engine revision 1a65d409c7
    • Dart version 2.19.6
    • DevTools version 2.20.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /home/matias/Android/Sdk
    • Platform android-33, build-tools 33.0.0
    • Java binary at: /snap/android-studio/current/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.13+0-b1751.21-8125866)

    • CHROME_EXECUTABLE = /snap/bin/firefox

[✓] Linux toolchain - develop for Linux desktop
    • Ubuntu clang version 14.0.0-1ubuntu1
    • cmake version 3.22.1
    • ninja version 1.10.1
    • pkg-config version 0.29.2

[✓] Android Studio (version 2021.3)
    • Android Studio at /snap/android-studio/125/android-studio
    • Flutter plugin version 72.1.1
    • Dart plugin version 213.7433
    • Java version OpenJDK Runtime Environment (build 11.0.13+0-b1751.21-8125866)

[✓] Android Studio
    • Android Studio at /snap/android-studio/current/android-studio
    • android-studio-dir = /snap/android-studio/current/android-studio
    • Java version OpenJDK Runtime Environment (build 11.0.13+0-b1751.21-8125866)

    • sdk gphone64 x86 64 (mobile) • emulator-5554 • android-x64    • Android 13 (API 33) (emulator)
    • Linux (desktop)              • linux         • linux-x64      • Ubuntu 22.04.2 LTS 5.19.0-41-generic
    • Chrome (web)                 • chrome        • web-javascript • Mozilla Firefox 112.0.1

[✓] HTTP Host Availability
    • All required HTTP hosts are available

• No issues found!
```

</details>

### Additional information

1. Maybe related with #107350 and the linked PR #120730
2. I think that this error is because if not set `autovalidateMode` or set **AutovalidateMode.disabled**, is the same behavior. Can see in this line https://github.com/flutter/flutter/blob/9241329b4669a468e7e11ef1e2f6cbd23cd2098b/packages/flutter/lib/src/widgets/form.dart#L341
3. With currently behavior and definition, if by default `autovalidateMode` on FormField is disabled, the `autovalidateMode` property on `Form`parent never would be works because `FormField` property has priority with respect to `Form`

Mairramer:
This PR introduces the `useStrictAutovalidateMode` property to the Form widget. When enabled, each descendant FormField respects its own `autovalidateMode`, allowing fields to opt out of automatic validation even if the parent Form has a global `AutovalidateMode` set.

Fixes flutter/flutter#125766

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/form.dart
```diff
@@ -220,6 +233,21 @@ class Form extends StatefulWidget {
...
+  /// [FormField.autovalidateMode], ignoring the [Form]'s setting.
+  ///
+  /// When false, the [Form]'s [AutovalidateMode] takes precedence and determines
+  /// the automatic validation behavior of all descendant fields.
+  ///
+  /// Calling [FormState.validate] always triggers validation for all fields,
+  /// regardless of this setting.
+  ///
+  /// Defaults to false.
+  final bool useStrictAutovalidateMode;
```
justinmc, line 247: Could this be done by seeing if autovalidateMode is null, or why wouldn't that work?

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -695,7 +728,17 @@ class FormFieldState<T> extends State<FormField<T>> with RestorationMixin {
       _value = value;
       _hasInteractedByUser.value = true;
     });
-    Form.maybeOf(context)?._fieldDidChange();
+    _formState?._fieldDidChange();
+  }
+
+  bool get _shouldValidate {
```
justinmc: This is a nice way to encapsulate this important logic :+1:

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -803,7 +842,8 @@ class FormFieldState<T> extends State<FormField<T>> with RestorationMixin {
       child: widget.builder(this),
     );
 
-    if (Form.maybeOf(context)?.widget.autovalidateMode == AutovalidateMode.onUnfocus &&
+    // TODO(Mairramer): Refactor this to use switch case instead.
```
justinmc: Heads up you have a TODO here.

packages/flutter/lib/src/widgets/form.dart
```diff
@@ -29,6 +29,18 @@ import 'will_pop_scope.dart';
...
+/// Defines how a [Form] validation is triggered.
+enum _FormValidation {
+  /// Automatic validation driven by the [Form]'s [AutovalidateMode].
+  auto,
+
+  /// Explicit validation via [FormState.validate] or
+  /// [FormState.validateGranularly].
+  manual;
+
+  bool get isAuto => this == _FormValidation.auto;
```
justinmc: Is this necessary? Is it just to save a few lines?

justinmc (commented):
I think validation should have been handled programmatically instead of via an enum, but that's a major change that will have to come in another iteration of Form. For now I agree that this is a problem we should solve along these lines. See my comment asking about using the null value of autovalidateMode.

packages/flutter/lib/src/widgets/form.dart
Mairramer: Regarding the null value of autovalidateMode, to be honest, I didn’t give it much thought. I used useStrictAutovalidateMode to make the validation behavior explicit: it ensures that each FormField either follows its own autovalidateMode or inherits the Form’s, depending on what I want.

Mairramer: One thing I’ve been thinking about is moving this to the `FormField` level instead of the `Form`. That might make it easier to manage which fields should not be validated. I’m not sure why someone would want that, but it could still be a good option, especially since we currently don’t have programmatic validation.

justinmc: Would something like this give you enough control and be doable?

 * When FormField.autovalidateMode is non-null
    - Always use it and ignore Form.autovalidateMode.
 * When FormField.autovalidateMode is null
    - Defer to Form.autovalidateMode.
    - If Form.autovalidateMode is also null, default to .disabled.

Or maybe that's a potential breaking change for existing users?

justinmc (commented):
I want to be sure of the useStrictAutovalidateMode API before we add it, see my comment.

packages/flutter/lib/src/widgets/form.dart
Mairramer: In my opinion, since both currently default to `AutovalidateMode.disabled`, and because `null` is effectively treated the same as `AutovalidateMode.disabled`, changing this behavior would indeed be a breaking change.

Introducing `useStrictAutovalidateMode` addresses this by preserving existing behavior and avoiding regressions. If we were to adopt the alternative approach you suggested, it would result in a breaking change, albeit one that could be justified as correcting an unintuitive or inconsistent behavior rather than introducing a new feature.

justinmc: @Mairramer Are you currently facing the problem described in https://github.com/flutter/flutter/issues/125766 in your own work? If so could you describe what you want to achieve there and how you expected autovalidateMode to work? Just trying to make sure I'm considering what devs want here.

Current behavior: The value on Form takes precedence, except if it's `disabled` (or explicitly `null`).

My interpretation of https://github.com/flutter/flutter/issues/125766 is that users find the current behavior unintuitive (I do too). They expect FormField.autovalidateMode to override Form.autovalidateMode. In the long run, that's the direction that I think Flutter should go, but as I've said elsewhere, Form needs a rewrite and maybe this isn't the time to make a big breaking change.

So that leaves us with the option of introducing a sub-optimal new API for the short term. My question is, does that solve a significant user problem today that makes it worth adding the complexity?

Mairramer: > @Mairramer Are you currently facing the problem described in https://github.com/flutter/flutter/issues/125766 in your own work? If so could you describe what you want to achieve there and how you expected autovalidateMode to work? Just trying to make sure I'm considering what devs want here.

At the moment, this isn’t a major issue for me. Since I work at a company with a design system, it helps mitigate these problems to some extent. However, introducing this would still benefit many of the wrappers we currently maintain.

> Current behavior: The value on Form takes precedence, except if it's disabled (or explicitly null).
My interpretation of https://github.com/flutter/flutter/issues/125766 is that users find the current behavior unintuitive (I do too). They expect FormField.autovalidateMode to override Form.autovalidateMode. In the long run, that's the direction that I think Flutter should go, but as I've said elsewhere, Form needs a rewrite and maybe this isn't the time to make a big breaking change.

> So that leaves us with the option of introducing a sub-optimal new API for the short term. My question is, does that solve a significant user problem today that makes it worth adding the complexity?

I implemented this following your suggestion https://github.com/flutter/flutter/pull/180822#discussion_r2710224689. However, it introduces a breaking change: `widget.validator` cannot call `setState`. Please see this commit for context: https://github.com/flutter/flutter/pull/180822/changes/1182874248c97590efe7a5d8ea12e376de54d8f0, similar to this https://github.com/flutter/flutter/pull/156708/changes/a3d01814d9a38409560dc2cd937274962bbd1813 .

I’d like to confirm the best way to avoid this breaking change during the first build.

My proposal to introduce `useStrictAutovalidateMode` is intended to preserve the current behavior without introducing breaking changes at this time.

Mairramer:
Closing this in favor of https://github.com/flutter/flutter/pull/182752

PR #180856, opened 2026-01-12
Fix false positives for hit test misses in flutter_test

### LINKED ISSUE #99302: AnimatedScale hit test miss
Having an animated scale with an inkwell makes hit test miss

Repro repository:

https://github.com/cedvdb/flutter_repros/tree/animated_scale_hit_test   <= branch animated_scale_hit_test

Then run:

`flutter test`

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({Key? key}) : super(key: key);

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const MyHomePage(),
    );
  }
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({Key? key}) : super(key: key);

  @override
    return const ScaledCard(
      key: ValueKey('tap-me'),
    );
  }
}

class ScaledCard extends StatelessWidget {
  const ScaledCard({
    Key? key,
  }) : super(key: key);

  @override
    return AnimatedScale(
      scale: 1.05,
      duration: const Duration(seconds: 1),
      child: Card(
        child: InkWell(
          onTap: () {},
          child: const Text('tap me'),
        ),
      ),
    );
  }
}
```

test

```dart

import 'package:flutter_repros/main.dart';

void main() {
  testWidgets('Counter increments smoke test', (WidgetTester tester) async {
    // Build our app and trigger a frame.
    await tester.pumpWidget(const MyApp());

    await tester.pumpAndSettle();
    await tester.tap(find.byKey(const ValueKey('tap-me')));
    await tester.pumpAndSettle();
  });
}

```

### LINKED ISSUE #100758: False positives for "missed hit test" warning for RenderObjects that don't add themselves to the HitTestResult
When the `RenderObject` "of" a `StatelessWidget` (the `RenderObject` of its first child `Element` that has one) does not add itself to the `HitTestResult` when hit, tapping that `StatelessWidget` in a test results in a "missed hit test" warning. Examples of such `RenderObject`s are those created by `Transform`, `Opacity`, and `AbsorbPointer`.

I think there should be no warning in these cases as the `StatelessWidget` in clearly hit.

Two examples:

```dart

void main() {
  testWidgets('with $Transform', (tester) async {
    await tester.pumpWidget(Center(
      child: Button(
        onTap: () => print("tap 1"),
      ),
    ));

    await tester.tap(find.byType(Button));
  });

  testWidgets('with $AbsorbPointer', (tester) async {
    await tester.pumpWidget(Center(
      child: GestureDetector(
        onTap: () => print("tap 2"),
        child: const Dummy(),
      ),
    ));

    await tester.tap(find.byType(Dummy));
  });
}

class Button extends StatelessWidget {
  const Button({Key? key, required this.onTap}) : super(key: key);

  final VoidCallback onTap;

  @override
  Widget build(BuildContext context) => Transform.scale(
        scale: 1.1,
        child: GestureDetector(
          onTap: onTap,
          child: Container(
            width: 40,
            height: 40,
            color: const Color(0xffff0000),
          ),
        ),
      );
}

class Dummy extends StatelessWidget {
  const Dummy({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) => const AbsorbPointer(
        child: Padding(
          padding: EdgeInsets.all(10),
          child: SizedBox(
            width: 20,
            height: 20,
          ),
        ),
      );
}
```

Output:
```
Warning: A call to tap() with finder "exactly one widget with type "Button" (ignoring offstage widgets): Button" derived an Offset (Offset(400.0, 300.0)) that would not hit test on the specified widget.
Maybe the widget is actually off-screen, or another widget is obscuring it, or the widget cannot receive pointer events.
The finder corresponds to this RenderBox: RenderTransform#bdd63 relayoutBoundary=up1
The hit test result at that offset is: HitTestResult(_RenderColoredBox#1700a@Offset(20.0, 20.0), RenderConstrainedBox#5cc34@Offset(20.0, 20.0), RenderPointerListener#3ffda@Offset(20.0, 20.0), RenderSemanticsGestureHandler#f5878@Offset(20.0, 20.0), RenderPositionedBox#29595@Offset(400.0, 300.0), HitTestEntry#5c7be(RenderView#0c197), HitTestEntry#cedf5(<AutomatedTestWidgetsFlutterBinding>))
#0      WidgetController._getElementPoint (package:flutter_test/src/controller.dart:956:25)
#1      WidgetController.getCenter (package:flutter_test/src/controller.dart:839:12)
#2      WidgetController.tap (package:flutter_test/src/controller.dart:273:18)
#3      main.<anonymous closure> (file:///Users/sander/Development/flutter_bug/test/widget_test.dart:12:18)
<asynchronous suspension>
#4      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:125:47)
<asynchronous suspension>
To silence this warning, pass "warnIfMissed: false" to "tap()".
To make this warning fatal, set WidgetController.hitTestWarningShouldBeFatal to true.
```
and the same for the other test.

Expected output: no warning.

```
[✓] Flutter (Channel unknown, 2.10.2, on macOS 12.3 21E230 darwin-arm, locale en-NL)
    • Flutter version 2.10.2 at /Users/.../flutter
    • Upstream repository unknown
    • Framework revision 097d3313d8 (5 weeks ago), 2022-02-18 19:33:08 -0600
    • Engine revision a83ed0e5e3
    • Dart version 2.16.1
    • DevTools version 2.9.2
```

O-Hannonen:
## Thise PR
- Closes #100758
- Closes #99302

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_test/lib/src/finders.dart
```diff
@@ -28,6 +29,26 @@ typedef SemanticsNodePredicate = bool Function(SemanticsNode node);
 /// Signature for [FinderBase.describeMatch].
 typedef DescribeMatchCallback = String Function(Plurality plurality);
 
+/// Returns true if [renderObject] is equal to [target] or is an ancestor of
+/// [target] in the render tree.
+///
+/// This is useful for hit testing because some render objects (like
+/// [RenderTransform]) don't add themselves to the hit test path but instead
+/// just transforms the hit test point and pass it to their children.
+bool isRenderObjectAncestorOfTarget(RenderObject renderObject, HitTestTarget target) {
```
O-Hannonen, line 38: Not sure if this is the right place for this helper, any suggestions where to place it?

packages/flutter_test/lib/src/controller.dart
```diff
@@ -2102,7 +2102,9 @@ abstract class WidgetController {
       final FlutterView view = _viewOf(finder);
       final result = HitTestResult();
       binding.hitTestInView(result, location, view.viewId);
-      final bool found = result.path.any((HitTestEntry entry) => entry.target == box);
+      final bool found = result.path.any(
+        (HitTestEntry entry) => finders.isRenderObjectAncestorOfTarget(box, entry.target),
+      );
```
justinmc, line 2121: Can you give me an example of something that will still trigger the warnIfMissed error despite this change?

justinmc (commented):
I'm having trouble reconciling in my head the problem in the issues with the change in this PR. Why is it correct to always check for the ancestors of the element too? What is warnIfMissed really trying to catch, and why is it better with this change?

I think I could be convinced that something like `Transform.scale(scale: 1.1, ...)` should  not trigger this error when its children are hit. Meanwhile, something like `IgnorePointer` should indeed trigger the error, because it is not hit.

We should agree on strict criteria for when the error should be triggered and then make the code match that.

By the way, could you fix the analyzer error that's failing now? I really want to make sure that this PR doesn't cause any new failures or anything.

packages/flutter_test/lib/src/controller.dart
O-Hannonen: Added some tests that highlight where the warnIfMissed still works 👍

packages/flutter_test/lib/src/finders.dart
```diff
@@ -28,6 +29,26 @@ typedef SemanticsNodePredicate = bool Function(SemanticsNode node);
 /// Signature for [FinderBase.describeMatch].
 typedef DescribeMatchCallback = String Function(Plurality plurality);
 
+/// Returns true if [renderObject] is equal to [target] or is an ancestor of
+/// [target] in the render tree.
+///
+/// This is useful for hit testing because some render objects (like
+/// [RenderTransform]) don't add themselves to the hit test path but instead
+/// just transforms the hit test point and pass it to their children.
```
justinmc: "transforms" => "transform"

packages/flutter_test/lib/src/finders.dart
justinmc: Given that it's used in this file I think it's good here. I think the other main option would be duplicating the code, but it seems worth the addition to the API.

packages/flutter_test/lib/src/controller.dart
justinmc: Perfect, thank you!

justinmc (approved):
LGTM with a nit. Thanks for thoroughly arguing your case here. I agree with your "strict criteria" for warnIfMissed and the code looks good.

AbdeMohlbi (approved):
LGTM

PR #180857, opened 2026-01-12
[ Tool ] Ensure `flutter.version.json` is regenerated on upgrade

bkonyi:
The tool performs various checks to determine whether or not to regenerate version information as part of a `flutter upgrade` instead of just regenerating `flutter.version.json` as part of each `flutter upgrade`. This has lead to some situations where the actual framework version doesn't match that reported by the `flutter.version.json`.

This change updates `flutter upgrade` to always regenerate `flutter.version.json` after checking out the new version of the framework to ensure that it contains the latest version information.

Fixes https://github.com/flutter/flutter/issues/178926


jtmcdole (approved):
LGTM.

PR #176992, opened 2025-10-14
Enable misc leak tracking

dkwingsmt:
This PR adds a new subshard that runs leak tracking tests on flutter/misc.

This is used to address a problem faced by PR https://github.com/flutter/flutter/pull/176519, which fixes a leak in `WidgetTester` but is unable to write a unit test.

I'm not certain this is the best approach so I'm open to other suggestions.

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[...]

PR #180879, opened 2026-01-13
Fix leaks in `flutter_test` tests

dkwingsmt:
Following https://github.com/flutter/flutter/pull/176992, this PR fixes all leaks in `flutter_test` tests.

A new method is added to reliably dispose the root layer of render views. Feel free to let me know if you have suggestions for the name!

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter_test/test/live_binding_test.dart
```diff
@@ -94,6 +95,7 @@ void main() {
     widgetCenter = tester.getRect(find.byType(Text)).center;
     expect(widgetCenter.dx, windowCenterX);
     expect(widgetCenter.dy, windowCenterY);
+    addTearDown(binding.resetLayers);
```
justinmc: Should this teardown be added earlier? If the test fails before it reaches here, then resetLayers won't be called.

justinmc (approved):
LGTM :+1: (but @chunhtai probably should have a look). The resetLayers name works for me.

Why were these leaks not caught by our leak tracking?

packages/flutter_test/lib/src/binding.dart
```diff
@@ -1838,6 +1838,34 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
...
+  /// triggers a clean disposal of the old layer tree.
+  ///
+  /// See also:
+  ///
+  ///  * [setSurfaceSize], which often necessitates calling this method during
+  ///    tear down.
+  Future<void> resetLayers() async {
+    await setSurfaceSize(null);
+    for (final RenderView renderView in renderViews) {
+      // Trigger a layer replacement by changing the device pixel ratio.
```
chunhtai: I tried to understand the code here. why do we assign twice in line  1862 and 1865?

packages/flutter_test/lib/src/binding.dart
```diff
@@ -1838,6 +1838,34 @@ abstract class TestWidgetsFlutterBinding extends BindingBase
...
+  /// not be automatically disposed, potentially leading to resource leaks or
+  /// interference with subsequent tests. This method forces a layer
+  /// replacement by temporarily toggling the device pixel ratio, which
+  /// triggers a clean disposal of the old layer tree.
+  ///
+  /// See also:
+  ///
+  ///  * [setSurfaceSize], which often necessitates calling this method during
+  ///    tear down.
+  Future<void> resetLayers() async {
```
chunhtai, line 1860: maybe this should be called resetViewConfiguration?

dkwingsmt:
@jmagman Yes, removing the `bringup` of `framework_tests_misc_leak_tracking` is the goal.

I haven't had time to address the comments, but plan to do it asap.

packages/flutter_test/lib/src/binding.dart
loic-sharma: See the comment above:

>This method forces a layer replacement by temporarily toggling the device pixel ratio, which triggers a clean disposal of the old layer tree.

I'd consider copying that "which triggers..." bit down to the implementation to make this more obvious :)

packages/flutter_test/test/widget_tester_test.dart
```diff
@@ -792,19 +797,22 @@ void main() {
...
-        (FlutterError e) => e.message,
-        'message',
-        contains('consider setting the "wrapWithView" parameter of that method to false'),
-      ),
-    );
-  });
+  testWidgets(
+    'passing a view to pumpWidget with wrapWithView: true throws',
+    experimentalLeakTesting: LeakTesting.settings
+        .withIgnoredAll(), // leaking by design because of exception
```
loic-sharma, line 803: Just double-checking - we can't use something like a `catch` to dispose? Ideally we don't leak even with exceptions.

However, I'm OK ignoring here since this seems like an uncommon scenario.

loic-sharma (approved):
Very nice!

packages/flutter_test/test/live_binding_test.dart
dkwingsmt: Good catch. I've moved all such teardowns to the beginning of test cases.

packages/flutter_test/lib/src/binding.dart
dkwingsmt: I get why you would think so. But the goal of this method is really to replace the root layers with new, unused ones, so that the current used ones can be disposed. Resetting view configuration is a side effect that has to be done anyway. Calling it `resetViewConfiguration` might lead to improper usage. I prefer the current name over `resetViewConfiguration`.

packages/flutter_test/lib/src/binding.dart
dkwingsmt: I've rewritten the doc of this method and the inline comments. Can you check if it's clearer now?

packages/flutter_test/test/widget_tester_test.dart
dkwingsmt: I actually tried to trace the state for this test (before I knew that exception is leaking by design), and from what I understand, the interruption caused by the exception leaves the system at an inconsistent state and breaks expectations of how to cleanup.

For example, class A and B keep reference counting of each other for resource management. If an exception occurs at the middle of `A.dispose`, B might not be notified and still keeps A's resources. And there would be no way to make B clean up in a public way later, because B's cleanup will call A's, who has recorded itself as `disposed` and, for example, throw assertions.

In the end, exceptions are not expected for an app, and memory cleanup after exceptions shouldn't be that important.

PR #180895, opened 2026-01-13
macOS: Implement tooltip window controller

knopp:
This PR implements tooltip windows on macOS.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -1173,7 +1173,7 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
            @"The provided view controller is not attached to this engine.");
   FlutterView* view = viewController.flutterView;
   CGRect scaledBounds = [view convertRectToBacking:view.bounds];
-  CGSize scaledSize = view.sizedToContents ? NSZeroSize : scaledBounds.size;
+  CGSize scaledSize = scaledBounds.size;
   double pixelRatio = view.layer.contentsScale;
   auto displayId = [view.window.screen.deviceDescription[@"NSScreenNumber"] integerValue];
   FlutterWindowMetricsEvent windowMetricsEvent = {
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -1189,6 +1189,7 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
       .top = static_cast<size_t>(scaledBounds.origin.y),
       .display_id = static_cast<uint64_t>(displayId),
       .view_id = viewController.viewIdentifier,
+      .has_constraints = view.sizedToContents,
       .min_width_constraint = static_cast<size_t>(minimumContentSize.width),
       .min_height_constraint = static_cast<size_t>(minimumContentSize.height),
       .max_width_constraint = static_cast<size_t>(maximumContentSize.width),
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -1171,10 +1171,14 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
   }
   NSAssert([self viewControllerForIdentifier:viewController.viewIdentifier] == viewController,
            @"The provided view controller is not attached to this engine.");
-  NSView* view = viewController.flutterView;
+  FlutterView* view = viewController.flutterView;
   CGRect scaledBounds = [view convertRectToBacking:view.bounds];
-  CGSize scaledSize = scaledBounds.size;
-  double pixelRatio = view.bounds.size.width == 0 ? 1 : scaledSize.width / view.bounds.size.width;
+  CGSize scaledSize = view.sizedToContents ? NSZeroSize : scaledBounds.size;
+  double pixelRatio = view.layer.contentsScale;
+  CGSize maximumContentSize =
+      view.sizedToContents ? [view convertSizeToBacking:view.maximumContentSize] : NSZeroSize;
+  CGSize minimumContentSize =
+      view.sizedToContents ? [view convertSizeToBacking:view.minimumContentSize] : NSZeroSize;
   auto displayId = [view.window.screen.deviceDescription[@"NSScreenNumber"] integerValue];
   const FlutterWindowMetricsEvent windowMetricsEvent = {
       .struct_size = sizeof(windowMetricsEvent),
@@ -1185,6 +1189,10 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
       .top = static_cast<size_t>(scaledBounds.origin.y),
       .display_id = static_cast<uint64_t>(displayId),
       .view_id = viewController.viewIdentifier,
+      .min_width_constraint = static_cast<size_t>(minimumContentSize.width),
+      .min_height_constraint = static_cast<size_t>(minimumContentSize.height),
+      .max_width_constraint = static_cast<size_t>(maximumContentSize.width),
+      .max_height_constraint = static_cast<size_t>(maximumContentSize.height),
   };
   _embedderAPI.SendWindowMetricsEvent(_engine, &windowMetricsEvent);
 }
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -1194,6 +1194,11 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
     windowMetricsEvent.min_height_constraint = static_cast<size_t>(minimumContentSize.height);
     windowMetricsEvent.max_width_constraint = static_cast<size_t>(maximumContentSize.width);
     windowMetricsEvent.max_height_constraint = static_cast<size_t>(maximumContentSize.height);
+  } else {
+    windowMetricsEvent.min_width_constraint = static_cast<size_t>(scaledSize.width);
+    windowMetricsEvent.min_height_constraint = static_cast<size_t>(scaledSize.height);
+    windowMetricsEvent.max_width_constraint = static_cast<size_t>(scaledSize.width);
+    windowMetricsEvent.max_height_constraint = static_cast<size_t>(scaledSize.height);
   }
   _embedderAPI.SendWindowMetricsEvent(_engine, &windowMetricsEvent);
 }
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -1175,12 +1175,8 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
   CGRect scaledBounds = [view convertRectToBacking:view.bounds];
   CGSize scaledSize = view.sizedToContents ? NSZeroSize : scaledBounds.size;
   double pixelRatio = view.layer.contentsScale;
-  CGSize maximumContentSize =
-      view.sizedToContents ? [view convertSizeToBacking:view.maximumContentSize] : NSZeroSize;
-  CGSize minimumContentSize =
-      view.sizedToContents ? [view convertSizeToBacking:view.minimumContentSize] : NSZeroSize;
   auto displayId = [view.window.screen.deviceDescription[@"NSScreenNumber"] integerValue];
-  const FlutterWindowMetricsEvent windowMetricsEvent = {
+  FlutterWindowMetricsEvent windowMetricsEvent = {
       .struct_size = sizeof(windowMetricsEvent),
       .width = static_cast<size_t>(scaledSize.width),
       .height = static_cast<size_t>(scaledSize.height),
@@ -1189,12 +1185,16 @@ static void SetThreadPriority(FlutterThreadPriority priority) {
       .top = static_cast<size_t>(scaledBounds.origin.y),
       .display_id = static_cast<uint64_t>(displayId),
       .view_id = viewController.viewIdentifier,
-      .has_constraints = view.sizedToContents,
-      .min_width_constraint = static_cast<size_t>(minimumContentSize.width),
-      .min_height_constraint = static_cast<size_t>(minimumContentSize.height),
-      .max_width_constraint = static_cast<size_t>(maximumContentSize.width),
-      .max_height_constraint = static_cast<size_t>(maximumContentSize.height),
   };
+  if (view.sizedToContents) {
+    CGSize maximumContentSize = [view convertSizeToBacking:view.maximumContentSize];
+    CGSize minimumContentSize = [view convertSizeToBacking:view.minimumContentSize];
+    windowMetricsEvent.has_constraints = true;
+    windowMetricsEvent.min_width_constraint = static_cast<size_t>(minimumContentSize.width);
+    windowMetricsEvent.min_height_constraint = static_cast<size_t>(minimumContentSize.height);
+    windowMetricsEvent.max_width_constraint = static_cast<size_t>(maximumContentSize.width);
+    windowMetricsEvent.max_height_constraint = static_cast<size_t>(maximumContentSize.height);
+  }
   _embedderAPI.SendWindowMetricsEvent(_engine, &windowMetricsEvent);
 }
 
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterRunLoop.swift
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterRunLoop.swift
@@ -13,6 +13,11 @@ import Foundation
 /// (`pollFlutterMessagesOnce()`).
 @objc public final class FlutterRunLoop: NSObject {
   private static let flutterRunLoopMode = CFRunLoopMode("FlutterRunLoopMode" as CFString)
+  /// Displaying sheet animation (beginSheet) uses a private run loop mode for the animation
+  /// duration. FlutterRunLoop needs to work in this mode in order to render the window content
+  /// correctly during the sheet animation.
+  private static let moveTimerRunLoopMode = CFRunLoopMode("_NSMoveTimerRunLoopMode" as CFString)
+
   private static var _mainRunLoop: FlutterRunLoop?
 
   private let runLoop: CFRunLoop = CFRunLoopGetCurrent()
@@ -50,6 +55,7 @@ import Foundation
     source = createdSource
     CFRunLoopAddSource(runLoop, source, .commonModes)
     CFRunLoopAddSource(runLoop, source, Self.flutterRunLoopMode)
+    CFRunLoopAddSource(runLoop, source, Self.moveTimerRunLoopMode)
 
     var timerContext = CFRunLoopTimerContext(
       version: 0,
@@ -77,15 +83,18 @@ import Foundation
     timer = createdTimer
     CFRunLoopAddTimer(runLoop, timer, .commonModes)
     CFRunLoopAddTimer(runLoop, timer, Self.flutterRunLoopMode)
+    CFRunLoopAddTimer(runLoop, timer, Self.moveTimerRunLoopMode)
   }
 
   deinit {
     CFRunLoopTimerInvalidate(timer)
     CFRunLoopRemoveTimer(runLoop, timer, .commonModes)
     CFRunLoopRemoveTimer(runLoop, timer, Self.flutterRunLoopMode)
+    CFRunLoopRemoveTimer(runLoop, timer, Self.moveTimerRunLoopMode)
     CFRunLoopSourceInvalidate(source)
     CFRunLoopRemoveSource(runLoop, source, .commonModes)
     CFRunLoopRemoveSource(runLoop, source, Self.flutterRunLoopMode)
+    CFRunLoopRemoveSource(runLoop, source, Self.moveTimerRunLoopMode)
   }
 
   // Ensures that the `FlutterRunLoop` for main thread is initialized. Only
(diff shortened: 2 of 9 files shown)


engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -1170,12 +1170,12 @@ - (void)updateWindowMetricsForViewController:(FlutterViewController*)viewControl
   }
   NSAssert([self viewControllerForIdentifier:viewController.viewIdentifier] == viewController,
            @"The provided view controller is not attached to this engine.");
-  NSView* view = viewController.flutterView;
+  FlutterView* view = viewController.flutterView;
   CGRect scaledBounds = [view convertRectToBacking:view.bounds];
-  CGSize scaledSize = scaledBounds.size;
-  double pixelRatio = view.bounds.size.width == 0 ? 1 : scaledSize.width / view.bounds.size.width;
+  CGSize scaledSize = view.sizedToContents ? NSZeroSize : scaledBounds.size;
```
loic-sharma: Could you expand on the `NSZeroSize` here? This size becomes the `FlutterView.physicalSize`, which should be the view's _current_ size: https://api.flutter.dev/flutter/dart-ui/FlutterView/physicalSize.html

> The current dimensions of the rectangle as last reported by the platform into which scenes rendered in this view are drawn.

Would this cause a content-sized view to have a 0x0 `FlutterView.physicalSize`? Can we use the `scaledBounds.size` instead?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.mm
```diff
@@ -158,4 +167,21 @@ - (NSString*)accessibilityLabel {
   return applicationName;
 }
 
+- (BOOL)sizedToContents {
+  return _sizingDelegate != nil &&
+         !NSEqualSizes([_sizingDelegate maximumViewSize:self], NSZeroSize);
+}
+
+- (NSSize)minimumContentSize {
+  return _sizingDelegate != nil ? [_sizingDelegate minimumViewSize:self] : NSZeroSize;
```
loic-sharma: Instead of returning `NSZeroSize` if `_sizingDelegate` is `nil`, could we return the current size of the view? Same comment for `maximumContentSize` below.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -127,10 +140,108 @@ - (void)windowWillExitFullScreen:(NSNotification*)notification {
...
+                                                _creationRequest.constraints.min_height);
+}
 
-@interface FlutterWindowController () {
-  NSMutableArray<FlutterWindowOwner*>* _windows;
+- (NSSize)maximumViewSize:(FlutterView*)view {
+  // Maximum window server size is limited to 10000
+  // https://developer.apple.com/documentation/appkit/nswindow/setcontentsize(_:)
+  const double maxWindowSize = 10000.0;
+  if (_creationRequest.has_size) {
```
loic-sharma: Should this be `has_constraints`?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -127,10 +140,108 @@ - (void)windowWillExitFullScreen:(NSNotification*)notification {
   _creationRequest.notify_listeners();
 }
 
-@end
+- (NSSize)minimumViewSize:(FlutterView*)view {
+  return _creationRequest.has_size ? NSZeroSize
```
loic-sharma: Should this be `has_constraints`?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
knopp: I think this a vestige from my initial sized to content implementation. I'll change it back to actual size.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: It could be `!has_constraints`. I'm still not completely clear what the semantics of specifying size for views sized to contents are, but it seems that `! has_constraints` is better fit here.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.mm
knopp: Done.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -1185,6 +1185,15 @@ - (void)updateWindowMetricsForViewController:(FlutterViewController*)viewControl
...
   };
+  if (view.sizedToContents) {
+    CGSize maximumContentSize = [view convertSizeToBacking:view.maximumContentSize];
+    CGSize minimumContentSize = [view convertSizeToBacking:view.minimumContentSize];
+    windowMetricsEvent.has_constraints = true;
+    windowMetricsEvent.min_width_constraint = static_cast<size_t>(minimumContentSize.width);
+    windowMetricsEvent.min_height_constraint = static_cast<size_t>(minimumContentSize.height);
+    windowMetricsEvent.max_width_constraint = static_cast<size_t>(maximumContentSize.width);
+    windowMetricsEvent.max_height_constraint = static_cast<size_t>(maximumContentSize.height);
+  }
```
mattkae, line 1202: If we're not sized to content, should we still set the constraints as "tight" constraints?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h
```diff
@@ -12,6 +12,38 @@
...
+ * be resized (and repositioned) accordingly.
+ */
+@protocol FlutterViewSizingDelegate <NSObject>
+
+/**
+ * When view should be sized to content, this method should return the minimum
+ * logical size of the view.
+ * For views that are not sized to content, this method is not called.
+ */
+- (NSSize)minimumViewSize:(nonnull FlutterView*)view;
```
mattkae: Just to let you know: I moved these fields to the `FlutterView` in the Win32 API, as we thought that the size constraints best fit their alongside other information about view sizing. Not a blocker, but food for thought

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -127,10 +140,103 @@ - (void)windowWillExitFullScreen:(NSNotification*)notification {
...
+                                                       _creationRequest.constraints.min_height)
+                                          : NSZeroSize;
+}
 
-@interface FlutterWindowController () {
-  NSMutableArray<FlutterWindowOwner*>* _windows;
+- (NSSize)maximumViewSize:(FlutterView*)view {
+  if (!_creationRequest.has_constraints) {
+    // Window is not sized to contents.
+    return NSZeroSize;
```
mattkae: Should this be max ints?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowControllerTest.mm
```diff
@@ -80,6 +80,50 @@ void TearDown() {
...
+      .parent_view_id = parentViewId,
+      .on_should_close = [] {},
+      .on_will_close = [] {},
+      .notify_listeners = [] {},
+      .on_get_window_position = position_callback,
+  };
+
+  const int64_t tooltipViewId =
+      InternalFlutter_WindowController_CreateTooltipWindow(engineId, &request);
+  EXPECT_NE(tooltipViewId, 0);
```
mattkae, line 126: Can we also write any tests about the size that we expect? e.g. whether the metrics are right and we're getting constrained or not?

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -135,28 +144,216 @@ class WindowingOwnerMacOS extends WindowingOwner {
...
+  late final NativeCallable<Void Function()> _onShouldClose;
+  late final NativeCallable<Void Function()> _onWillClose;
+  late final NativeCallable<Void Function()> _onResize;
+  late final NativeCallable<
+    Pointer<_Rect> Function(
+      Pointer<_Size> childSize,
+      Pointer<_Rect> parentRect,
+      Pointer<_Rect> outputRect,
+    )
+  >
```
mattkae, line 246: Can we `typedef` this function type?

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -634,6 +801,48 @@ class _MacOSPlatformInterface {
...
 
+  @Native<Int64 Function(Int64, Pointer<_WindowCreationRequest>)>(
+    symbol: 'InternalFlutter_WindowController_CreateTooltipWindow',
+  )
+  external static int _createTooltipWindow(int engineId, Pointer<_WindowCreationRequest> request);
+
+  /// Creates a new window and returns the viewId of the created FlutterView.
+  static int createTooltipWindow({
+    required BoxConstraints preferredConstraints,
+    int? parentViewId,
```
mattkae: Should this be required instead of implicitly `0`?

mattkae (changes requested):
A few notes, but looks good overall 👍

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h
knopp: I think it's cleaner this way. There is no threading issue here and I don't think I want to pollute FlutterView with additional logic.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: When it is NSZeroSize FlutterView assumes iti s not sized to content. Maybe it would be better to have explicit signal for that rather than using "magical" size value.

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: I think so.

packages/flutter/lib/src/widgets/_window_macos.dart
knopp: IT doesn't seem to be used more than once don't quite see the reason for typedef?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowControllerTest.mm
knopp: Done.

packages/flutter/lib/src/widgets/_window_macos.dart
mattkae: Whoops, thought I saw it elsewhere. Ok then :+1:

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
mattkae: Max size would be a less magical in my opinion, but I get your point. I'm fine either way

knopp: NSSize is CGFloat. That makes it rather awkward to check for max size.

mattkae: But do you even have to check for it on the other hand? Would it not just work out because it's very large?

knopp: Yes, resizing synchronization is turned off when sized to content.

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -634,6 +801,48 @@ class _MacOSPlatformInterface {
...
+      >
+    >
+    onGetWindowPosition,
+  }) {
+    final Pointer<_WindowCreationRequest> request = _allocator<_WindowCreationRequest>()
+      ..ref.onShouldClose = onShouldClose
+      ..ref.onWillClose = onWillClose
+      ..ref.onNotifyListeners = onNotifyListeners
+      ..ref.onGetWindowPosition = onGetWindowPosition
+      ..ref.parentViewId = parentViewId ?? 0;
```
mattkae: This null check is useless now?

mattkae (approved):
One nit, but looks good otherwise 🚀 Maybe contact Justin + Loic if they want input though

LouiseHsu:
@knopp from triage: do you think you could add a description for future reference? :)

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h
loic-sharma: Yes I do agree this approach works much better for macOS than it did for Windows 👍

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -127,10 +140,103 @@ - (void)windowWillExitFullScreen:(NSNotification*)notification {
   _creationRequest.notify_listeners();
 }
 
-@end
+- (NSSize)minimumViewSize:(FlutterView*)view {
+  return _creationRequest.has_constraints ? NSMakeSize(_creationRequest.constraints.min_width,
+                                                       _creationRequest.constraints.min_height)
+                                          : NSZeroSize;
```
loic-sharma: For `minimumViewSize` and `maximumViewSize`, I don't like that these these return `NSZeroSize`. I worry that this magic value could cause weird bugs in the future if one uses these methods on a view that might not be sized to content.

Could we instead return the current size of the view if it's not sized to content? If not, could we have an assertion that this method is only called if the view is sized to content?

loic-sharma (approved):
LGTM modulo one nitpick!

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -1185,6 +1185,20 @@ - (void)updateWindowMetricsForViewController:(FlutterViewController*)viewControl
       .display_id = static_cast<uint64_t>(displayId),
       .view_id = viewController.viewIdentifier,
   };
+  if (view.sizedToContents) {
+    CGSize maximumContentSize = [view convertSizeToBacking:view.maximumContentSize];
+    CGSize minimumContentSize = [view convertSizeToBacking:view.minimumContentSize];
```
hellohuanlin, line 1191: what is "backing"?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterRunLoop.swift
```diff
@@ -13,6 +13,11 @@ import Foundation
 /// (`pollFlutterMessagesOnce()`).
 @objc public final class FlutterRunLoop: NSObject {
   private static let flutterRunLoopMode = CFRunLoopMode("FlutterRunLoopMode" as CFString)
+  /// Displaying sheet animation (beginSheet) uses a private run loop mode for the animation
+  /// duration. FlutterRunLoop needs to work in this mode in order to render the window content
+  /// correctly during the sheet animation.
+  private static let moveTimerRunLoopMode = CFRunLoopMode("_NSMoveTimerRunLoopMode" as CFString)
```
hellohuanlin: this private api usage could result in app rejections.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.mm
```diff
@@ -43,7 +43,14 @@ - (instancetype)initWithMTLDevice:(id<MTLDevice>)device
 }
 
 - (void)onPresent:(CGSize)frameSize withBlock:(dispatch_block_t)block delay:(NSTimeInterval)delay {
-  [_resizeSynchronizer performCommitForSize:frameSize afterDelay:delay notify:block];
+  // This block will be called in main thread same run loop turn as the layer content
+  // update.
+  auto notifyBlock = ^{
+    NSSize scaledSize = [self convertSizeFromBacking:frameSize];
+    [self.sizingDelegate viewDidUpdateContents:self withSize:scaledSize];
```
hellohuanlin, line 52: weak

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -198,6 +309,55 @@ - (FlutterViewIdentifier)createDialogWindow:(const FlutterWindowCreationRequest*
   return c.viewIdentifier;
 }
 
+- (FlutterViewIdentifier)createTooltipWindow:(const FlutterWindowCreationRequest*)request {
+  FlutterViewController* c = [[FlutterViewController alloc] initWithEngine:_engine
```
hellohuanlin: nit: `flutterVC` or `flutterViewController`

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -198,6 +309,55 @@ - (FlutterViewIdentifier)createDialogWindow:(const FlutterWindowCreationRequest*
...
+  // using ARC.
+  [window setReleasedWhenClosed:NO];
+
+  window.contentViewController = c;
+  window.styleMask = NSWindowStyleMaskBorderless;
+  window.hasShadow = NO;
+  window.opaque = NO;
+  window.backgroundColor = [NSColor clearColor];
+
+  FlutterWindowOwner* w = [[FlutterWindowOwner alloc] initWithWindow:window
```
hellohuanlin: same here. avoid single letter variable name if possible

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -198,6 +309,55 @@ - (FlutterViewIdentifier)createDialogWindow:(const FlutterWindowCreationRequest*
...
+                                               flutterViewController:c
+                                                     creationRequest:*request];
+
+  c.flutterView.sizingDelegate = w;
+  [c.flutterView setBackgroundColor:[NSColor clearColor]];
```
hellohuanlin: does it have a property `.backgroundColor`?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -198,6 +309,55 @@ - (FlutterViewIdentifier)createDialogWindow:(const FlutterWindowCreationRequest*
...
+      }
+    }
+  }
+
+  NSAssert(parent != nil, @"Tooltip window must have a parent window.");
+
+  window.ignoresMouseEvents = YES;
+  window.collectionBehavior = NSWindowCollectionBehaviorAuxiliary;
+  [parent addChildWindow:window ordered:NSWindowAbove];
+  // window.alphaValue = 0.0;
```
hellohuanlin: revert this

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -265,6 +423,27 @@ - (void)closeAllWindows {
   [_windows removeAllObjects];
 }
 
+static BOOL IsChildParent(NSWindow* child, NSWindow* parent) {
```
hellohuanlin: this isn't checking child/parent, but rather descendent relationship

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: nice catch

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: Fair point.

knopp: Renamed to IsChildAncestor.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: It's not formally declared but it should compile. I'll change to `.backgroundColor` for consistency.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
knopp: [Backing store](https://developer.apple.com/documentation/appkit/nsview/converttobacking(_:)-4ra9y?language=objc) coordinate system.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.mm
knopp: I don't think that makes any difference here. The view is not meant to get deallocated before the block executes.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterRunLoop.swift
knopp: I have removed this. It shouldn't have been in this PR anyway as it is unrelated to tooltips. 

I've found a workaround where I can trigger a very quick invisible `NSMoveHelper` animation which uses the run loop mode, which can then be queried through `CFRunLoopCopyAllModes` and does not need specified explicitly.

hellohuanlin (approved):
LGTM

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: I will change it to `std::optional<NSSize>` instead.

Issue #180787, opened 2026-01-10
Migrate iOS plugin tests from XCTest to Swift Testing


jmagman:
[Swift Testing](https://developer.apple.com/xcode/swift-testing/) has some interesting features, like parameterized tests.
```swift
@Test("Continents mentioned in videos", arguments: [
    "A Beach",
    "By the Lake",
    "Camping in the Woods"
])
func mentionedContinents(videoName: String) async throws {
    let videoLibrary = try await VideoLibrary()
    let video = try #require(await videoLibrary.video(named: videoName))
    #expect(video.mentionedContinents.count <= 3)
}
```

Migrate our plugin iOS and macOS tests from XCTest to Swift Testing, where possible. This is just to do the mechanical migration, not to actually adopt new Swift Testing features.

https://developer.apple.com/videos/play/wwdc2024/10179/
https://developer.apple.com/videos/play/wwdc2024/10195/

Available in Xcode 16. The test targets must use Swift 6.

Note: this does not require the plugin itself to adopt Swift 6, which is tracked here https://github.com/flutter/flutter/issues/150388

* [ ] camera_avfoundation
* [ ] file_selector_ios
* [ ] file_selector_macos
* [ ] google_maps_flutter_ios
* [x] https://github.com/flutter/packages/pull/10787
* [ ] image_picker_ios
* [ ] in_app_purchase_storekit
* [ ] https://github.com/flutter/packages/pull/10786
* [ ] https://github.com/flutter/packages/pull/10802
* [x] https://github.com/flutter/packages/pull/10761
* [ ] pigeon
* [x] https://github.com/flutter/packages/pull/10762/
* [ ] https://github.com/flutter/packages/pull/10767
* [ ] https://github.com/flutter/packages/pull/10766
* [x] https://github.com/flutter/packages/pull/10764
* [ ] https://github.com/flutter/packages/pull/10780
* [x] https://github.com/flutter/packages/pull/10763
* [ ] video_player_avfoundation

stuartmorgan-g:
I did google_maps_flutter, file_selector_*, and webview_flutter almost entirely with Gemini, with minimal intervention; it worked quite well. Some manual cleanup was necessary, but definitely less work than the earlier ones I did by hand.

Two notes for future conversions that try to go that route:
- I tried `camera_avfoundation`, and it did manage to do a conversion, but there's some pretty weird handling of continuations. I expect that one to take non-trivial human work to do correctly.
[...]

PR #180898, opened 2026-01-13
Update discussion of iOS/macOS testing

stuartmorgan-g:
- Adds references to Swift Testing to the sections discussion running XCTests.
- Updates the outdated guidance to use Obj-C for new iOS test setup
- Removes the instructions for adding XCTest test targets to a plugin, as we have added it to all existing plugins, and it's part of the template now so will automatically exist in new plugins.
- Remove discussion of OCMock, because we should not be adding new usage of OCMock at this point.

See https://github.com/flutter/flutter/issues/180787


jmagman (approved):
Thank you! I forgot about these docs.

PR #181004, opened 2026-01-14
Linux build hooks: Read compiler config from CMakeCache.txt

### LINKED ISSUE #180770: Native assets: Take GNU linkers into account when building Linux apps
This is a follow-up to https://github.com/flutter/flutter/issues/178659. When building Flutter apps using code assets on Linux systems that don't have LLVM's linker installed, the build process fails with `Failed to find any of [ld.lld, ld] in LocalDirectory: /usr/lib/llvm-18/bin`.

To reproduce this, start on a fresh Linux installation that has all the [required tools](https://docs.flutter.dev/platform-integration/linux/setup) but no full LLVM toolchain, check out an app using code assets (such as [this one](https://github.com/simolus3/sqlite3.dart/tree/main/examples/flutter_integration_tests)) and run `flutter run -d linux`.

To fix https://github.com/flutter/flutter/issues/178659, we've made the search in `cCompilerConfigLinux` optional when running unit tests. My understanding is that we fix the compiler and linker for actual app builds to we'll be able to support static linking with consistent toolchains in the future.
However, that method is still incorrect. It claims it

https://github.com/flutter/flutter/blob/0cfc04170d834ca1868454f2cd6598f335e87dd6/packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart#L12-L13

which is just not true. Default linux setups that have clang but no `lld` will use the GNU linker. I'm actually pretty sure that even if `lld` was installed, cmake would still prefer to use GNU ld. And for cross-compilation, that tool should probably use `aarch64-linux-gnu-ld` or whatever it's called. But trying to figure out compilers and linkers from the target platform alone is fundamentally flawed since users might customize their cmake logic or pass build flags to customize the linker.

To me at least, it seems like the only way to actually do what the documentation comment promises is to parse `CMakeCache.txt` in the build directory and extract values for `CMAKE_CXX_COMPILER`, `CMAKE_CXX_COMPILER_AR` and `CMAKE_LINKER`. That seems a bit fragile, but everything else is just wrong.

cc @dcharkes

simolus3:
When running native assets as part of an app build, the Flutter tool tries to infer clang, archiver and linker executables from `PATH`. That logic doesn't handle builds where clang is used with a GNU linker.

To fix this, this PR changes the logic to read `CMakeCache.txt` from the Linux build directory, which contains all tools to be used in the build.
While Windows also uses CMake, we don't need to adopt this for Windows since MSVC is the only supported toolchain for that target anyway.

Closes https://github.com/flutter/flutter/issues/180770. cc @dcharkes 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
+++ b/packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
@@ -2,82 +2,92 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:code_assets/code_assets.dart';
 
 import '../../../base/common.dart';
 import '../../../base/file_system.dart';
-import '../../../base/io.dart';
 import '../../../globals.dart' as globals;
 
-/// Returns a [CCompilerConfig] matching a toolchain that would be used to compile the main app with
-/// CMake on Linux.
+/// Returns a [CCompilerConfig] suitable for compiling code assets for Linux apps.
 ///
-/// Flutter expects `clang++` to be on the path on Linux hosts, which this uses to search for the
-/// accompanying `clang`, `ar`, and `ld`.
+/// For app builds, [cmakeDirectory] must be given and point to the CMake build root for the app,
+/// e.g. `build/linux/x64/debug`. The compiler configuration is resolved by reading the
+/// `CMakeCache.txt` file in that directory to ensure we use the same compiler as the main app.
 ///
-/// If [throwIfNotFound] is false, this is allowed to fail (in which case `null`) is returned. This
-/// is used for `flutter test` setups, where no main app is compiled and we thus don't want a
-/// `clang` toolchain to be a requirement.
-Future<CCompilerConfig?> cCompilerConfigLinux({required bool throwIfNotFound}) async {
-  const kClangPlusPlusBinary = 'clang++';
-  // NOTE: these binaries sometimes have different names depending on the installation;
-  // thus, we check for a few possible options (in order of preference).
-  const kClangBinaryOptions = ['clang'];
-  const kArBinaryOptions = ['llvm-ar', 'ar'];
-  const kLdBinaryOptions = ['ld.lld', 'ld'];
-
-  final ProcessResult whichResult = await globals.processManager.run(<String>[
-    'which',
-    kClangPlusPlusBinary,
-  ]);
-  if (whichResult.exitCode != 0) {
-    if (throwIfNotFound) {
-      throwToolExit('Failed to find $kClangPlusPlusBinary on PATH.');
-    } else {
-      return null;
-    }
-  }
-  File clangPpFile = globals.fs.file((whichResult.stdout as String).trim());
-  clangPpFile = globals.fs.file(await clangPpFile.resolveSymbolicLinks());
-
-  final Directory clangDir = clangPpFile.parent;
-  Uri? findExecutable({required List<String> possibleExecutableNames, required Directory path}) {
-    final Uri? found = _findExecutableIfExists(
-      possibleExecutableNames: possibleExecutableNames,
-      path: path,
-    );
-
-    if (found == null && throwIfNotFound) {
-      throwToolExit('Failed to find any of $possibleExecutableNames in $path');
-    }
-
-    return found;
-  }
-
-  final Uri? linker = findExecutable(path: clangDir, possibleExecutableNames: kLdBinaryOptions);
-  final Uri? compiler = findExecutable(
-    path: clangDir,
-    possibleExecutableNames: kClangBinaryOptions,
-  );
-  final Uri? archiver = findExecutable(path: clangDir, possibleExecutableNames: kArBinaryOptions);
-
-  if (linker == null || compiler == null || archiver == null) {
-    assert(!throwIfNotFound); // otherwise, findExecutable would have thrown
+/// Flutter also builds code assets for widget tests. Since there is no app build in that context,
+/// [cmakeDirectory] should be set to null for those builds.
+Future<CCompilerConfig?> cCompilerConfigLinux({Directory? cmakeDirectory}) async {
+  if (cmakeDirectory == null) {
+    // No CMake reference (e.g. for a widget test). Hooks can resolve to any
+    // compiler.
     return null;
   }
-  return CCompilerConfig(linker: linker, compiler: compiler, archiver: archiver);
+
+  // For app builds, use the same compiler as the native/GTK parts of the app.
+  final File cmakeCacheTxt = cmakeDirectory.childFile('CMakeCache.txt');
+  if (!cmakeCacheTxt.existsSync()) {
+    throwToolExit(
+      'Could not read compiler configurations for build hooks, expected ${cmakeCacheTxt.path} to exist.',
+    );
+  }
+
+  const archiverVariable = 'CMAKE_AR';
+  const compilerVariable = 'CMAKE_CXX_COMPILER';
+  const linkerVariable = 'CMAKE_LINKER';
+
+  String? archiver;
+  String? cxxCompiler;
+  String? linker;
+
+  final String cmakeCacheContents = await cmakeCacheTxt.readAsString();
+  for (final String line in const LineSplitter().convert(cmakeCacheContents)) {
+    final RegExpMatch? match = _cmakeCacheEntry.firstMatch(line);
+    if (match != null) {
+      final String variable = match.group(1)!;
+      final String value = match.group(2)!;
+
+      switch (variable) {
+        case archiverVariable:
+          archiver = value;
+        case compilerVariable:
+          cxxCompiler = value;
+        case linkerVariable:
+          linker = value;
+      }
+    }
+  }
+
+  Uri requireTool(String? found, String variableName) {
+    if (found == null) {
+      throwToolExit('Expected ${cmakeCacheTxt.path} to contain an entry for $variableName');
+    }
+
+    final File file = globals.fs.file(found);
+    if (!file.existsSync()) {
+      throwToolExit(
+        'Expected ${file.path} (read from $variableName in ${cmakeCacheTxt.path}) to exist.',
+      );
+    }
+
+    return file.uri;
+  }
+
+  // Find clang next to the clang++ we use in CMake
+  File clangPpFile = globals.fs.file(requireTool(cxxCompiler, compilerVariable));
+  clangPpFile = globals.fs.file(await clangPpFile.resolveSymbolicLinks());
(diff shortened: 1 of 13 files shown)


packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
```diff
@@ -6,78 +6,87 @@ import 'package:code_assets/code_assets.dart';
...
-  const kClangBinaryOptions = ['clang'];
-  const kArBinaryOptions = ['llvm-ar', 'ar'];
-  const kLdBinaryOptions = ['ld.lld', 'ld'];
+/// Flutter also builds code assets for widget tests. Since there is no app build in that context,
+/// [cmakeDirectory] should be set to null for those builds.
+Future<CCompilerConfig?> cCompilerConfigLinux({Directory? cmakeDirectory}) async {
+  if (cmakeDirectory == null) {
+    // No CMake reference (e.g. for a widget test). Hooks can resolve to any
+    // compiler.
+    return null;
```
dcharkes, line 24: This is somewhat unfortunate, because it means that between your widget test and your app build the hook invocations aren't shared. Which means that a hook respecting `CCompilerConfig` and otherwise finding its own compiler would compile the assets twice unless it realizes the compilers are equal and it manually uses `input.outputDirectoryShared` instead of `input.outputDirectory` (which is a hash influenced by the c compiler config).

Maybe we should try to default to what CMake would do with the defaults that Flutter passes to it:

https://github.com/flutter/flutter/blob/590666c456aa3f105d8435619439d0abcb298204/packages/flutter_tools/lib/src/linux/build_linux.dart#L172-L192

Find `clang` on the `PATH`.

What does `CMake` default to for `ar` and `ld`, could we reasonably try to duplicate that?

If we do this, we need to pass the target `Architecture` to this method to avoid setting `clang` for the host for cross compiling (assuming we'd would try to call this method for some other cross compilation that doesn't involve a cmake build in some future).

If not, maybe we just need to accept that your widgets are compiled twice if you run widget tests and build your app. If you happened to `dart run` something in your project the `CCompilerConfig` is also `null` right now, so at least the assets should be shared with that run. (In that case, please just add a comment that the assets will _not_ be shared with the main build that that they are shared with any `dart test` or `dart run` commands.)

I'd be happy with either of the two solutions here. We just need to make a conscious choice.

packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
```diff
@@ -6,78 +6,87 @@ import 'package:code_assets/code_assets.dart';
...
-Future<CCompilerConfig?> cCompilerConfigLinux({required bool throwIfNotFound}) async {
-  const kClangPlusPlusBinary = 'clang++';
-  // NOTE: these binaries sometimes have different names depending on the installation;
-  // thus, we check for a few possible options (in order of preference).
-  const kClangBinaryOptions = ['clang'];
-  const kArBinaryOptions = ['llvm-ar', 'ar'];
-  const kLdBinaryOptions = ['ld.lld', 'ld'];
+/// Flutter also builds code assets for widget tests. Since there is no app build in that context,
+/// [cmakeDirectory] should be set to null for those builds.
+Future<CCompilerConfig?> cCompilerConfigLinux({Directory? cmakeDirectory}) async {
```
dcharkes, line 20: Wait, before this PR we were actually passing the wrong compiler in case we were cross compiling?! And we never caught that? 🙈

packages/flutter_tools/test/general.shard/isolated/linux/native_assets_test.dart
```diff
@@ -93,60 +90,123 @@ void main() {
       await fileSystem.file('/some/path/to/llvm-ar').create();
       await fileSystem.file('/some/path/to/ld.lld').create();
 
-      final CCompilerConfig result = (await cCompilerConfigLinux(throwIfNotFound: true))!;
+      await environment.outputDir.childFile('CMakeCache.txt').writeAsString('''
+//CXX compiler
+CMAKE_CXX_COMPILER:FILEPATH=/some/path/to/clang++
```
dcharkes, line 96: Are these always absolute paths? The hooks process is not executed in the same working directory necessarily as the cmake build.

dcharkes (commented):
❤️  Thanks @simolus3!

I believe the data assets failure is because it's the `flutter run` that runs some of the data assets hooks instead of `flutter assemble`. Flutter run invokes cmake, which invokes flutter assemble. On the first run, there is no cmake build directory yet. I believe we should make it so that we don't try to look up the ccompiler config at all if we're not trying to run the build for the code assets. When looking at the stack trace we might actually be running the build for the code assets twice now? Once in `flutter run` and once in `flutter assemble`? Code assets should only be built in `flutter assemble`. 

cc @mosuem who last touched this code. Actually, doing data assets during `flutter run` also means data assets won't work at all if the flutter app is embedded in another app? Because then `flutter run` isn't run. It would just be another xcodebuild invoking another xcodebuild, or another cmake invoking the flutter cmake.

Edit: relevant part of the stack trace:

```
[ 18.2s] stdout:            #3      FlutterNativeAssetsBuildRunnerImpl.setCCompilerConfig (package:flutter_tools/src/isolated/native_assets/native_assets.dart:282:75)
-> this is followed by `_runDartHooks` which _also_ builds the code assets.

[ 18.2s] stdout:            #4      runFlutterSpecificHooks (package:flutter_tools/src/isolated/native_assets/native_assets.dart:106:23)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #5      DartBuild.build (package:flutter_tools/src/build_system/targets/native_assets.dart:76:14)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #6      _BuildInstance._invokeInternal (package:flutter_tools/src/build_system/build_system.dart:937:9)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #7      FlutterBuildSystem.build (package:flutter_tools/src/build_system/build_system.dart:684:16)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #8      FlutterHookRunnerNative.runHooks (package:flutter_tools/src/build_system/targets/hook_runner_native.dart:45:35)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #9      HotRunner._updateDevFS (package:flutter_tools/src/run_hot.dart:502:30)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #10     HotRunner._attach (package:flutter_tools/src/run_hot.dart:309:40)
[ 18.2s] stdout:            <asynchronous suspension>
[ 18.2s] stdout:            #11     RunCommand.runCommand (package:flutter_tools/src/commands/run.dart:904:26)
-> we are in `flutter run`, not `flutter assemble`
```

dcharkes:
And also, maybe add to the PR description we don't need the same for Windows, because Flutter mandates MSVC and we don't support the cmake on Windows targeting clang.

packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
simolus3: I think it was fine because `clang` and `lld` can cross-compile natively? And then `native_toolchain_c` would have added the correct target triple based on the target ABI.

packages/flutter_tools/test/general.shard/isolated/linux/native_assets_test.dart
simolus3: I've never seen a relative path in there, but I'll add an explicit check for this to improve the error message in case someone does encounter a relative path.

packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
simolus3: > In that case, please just add a comment that the assets will _not_ be shared with the main build that that they are shared with any `dart test` or `dart run` commands.)

Done 👍

dcharkes:
> I've added the option passing the build directory as a type for `buildCodeAssets` because it's only relevant when building code assets. Not configuring compilers on code targets when we're not building code assets seems to work.

In https://github.com/flutter/flutter/pull/181507 I'm also changing the build output directory and making it only be used in `installCodeAssets`. So maybe we should wait on another land to main and then rebase again. 🙈  That PR turned out to be a huge refactor (see PR description).

packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
```diff
@@ -6,78 +6,87 @@ import 'package:code_assets/code_assets.dart';
...
-    'which',
-    kClangPlusPlusBinary,
-  ]);
-  if (whichResult.exitCode != 0) {
-    if (throwIfNotFound) {
-      throwToolExit('Failed to find $kClangPlusPlusBinary on PATH.');
-    } else {
-      return null;
+  const archiverVariable = 'CMAKE_AR';
+  const compilerVariable = 'CMAKE_CXX_COMPILER';
```
dcharkes, line 39: Should we also use `CMAKE_C_COMPILER`? Is there a reason we don't read `CMAKE_C_COMPILER` directly from the cache if available, rather than assuming `clang` is next to `clang++`?

packages/flutter_tools/lib/src/isolated/native_assets/native_assets.dart
```diff
@@ -52,34 +67,56 @@ Future<DartHooksResult> runFlutterSpecificHooks({
...
+  final BuildMode buildMode = _getBuildMode(
+    environmentDefines,
+    targetPlatform == TargetPlatform.tester,
+  );
+
+  // When building for Linux, we need access to the native build directory to
+  // read compiler options from CMakeLists.txt that are then forwarded to build
+  // hooks.
+  Directory? nativeBuildDirectory;
+  if (buildCodeAssets?.appBuildDirectory case final appBuildDirectory?) {
```
dcharkes: This is target OS specific, can this live in
packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart ?

Can we maybe just move it into `targetsFor` and `_linuxTarget`?

dcharkes (approved):
Sorry! This fell off my radar! LGTM with some small comments.

packages/flutter_tools/lib/src/isolated/native_assets/linux/native_assets.dart
simolus3: The reason is that Flutter projects configure cmake with only the `CXX` language on the project, at least in my testing the `CMAKE_C_COMPILER` variable seemed to be unset.

I'll add a comment explaining this behavior 👍

robert-ancell (approved):
The changes make sense to me and I don't see any issues, thanks!

Issue #180873, opened 2026-01-12
enable 'FragmentShader shader with array uniforms renders correctly' for impeller


gaaclarke:
https://github.com/flutter/flutter/blob/82d232df03e35b3937ca3737ba0cf7f12777a2a8/engine/src/flutter/testing/dart/fragment_shader_test.dart#L480 fails on impeller.

It turns out that the `mat2Array` is partially initialized on impeller.  For example mat2Array[0][0][0] is its expected value, but mat2Array[0][1][0] is not, but mat2Array[1][0][0] is.

I've placed a print statement in the FragmentShader constructor for the size of the floats to intialize.  On skia it correctly reports 20, on impeller it reports 40.

gaaclarke:
This seems to be related to the "_RESERVED_IDENTIFIER_FIXUP_gl_DefaultUniformBlock" struct.  The 40 floats comes from taking the size of that struct and dividing it by sizeof(float), which gives 40.

Another issue is tracked with that struct in https://github.com/flutter/flutter/issues/180595.

gaaclarke:
https://github.com/flutter/flutter/blob/d7965e1e136a3f8b7889f556a75a078c986a0ed7/engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc#L33 appears to be in the code where we are supposed to be accounting for padding.  So making a buffer of 40 floats isn't wrong as long as we place them correctly when we create the buffer for the gpu.

PR #181013, opened 2026-01-15
Enables fragment shader test for impeller (fixes mat2 on vulkan)

gaaclarke:
This rearranges the packing of mat2 fields in structs which was causing an existing fragment shader test to fail.

The test still doesn't work on Metal, but this PR fixes it for Vulkan.  I suspect the fix for metal may be similar but since the uniforms aren't in a virtual struct the code isn't hitting this fix.

issue: https://github.com/flutter/flutter/issues/180873

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/impeller/compiler/reflector.cc
```diff
@@ -869,19 +870,40 @@ std::vector<StructMember> Reflector::ReadStructMembers(
...
+      result.emplace_back(StructMember{
+          /*p_type=*/"Mat2",
+          /*p_base_type=*/spirv_cross::SPIRType::BaseType::Float,
+          /*p_name=*/GetMemberNameAtIndex(struct_type, i),
+          /*p_offset=*/struct_member_offset,
+          /*p_size=*/sizeof(Point),
+          /*p_byte_length=*/total_length,
+          /*p_array_elements=*/count,
+          /*p_element_padding=*/8,
+      });
```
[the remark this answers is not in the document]
gaaclarke: Gemini, don't you ever have anything nice to say?  Would it kill you to say "bang up job!"?

engine/src/flutter/impeller/fixtures/mat2_test.frag
```diff
@@ -0,0 +1,14 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+layout(std140) uniform Params {
```
walley892, line 5: If you get rid of the explicit std140 layout directive, will this fail to compile on vulkan?

I'm not sure what the proper boundaries for this testing layer should be, but I think since the original issue was with plain mat2s (outside of structs) that should be what we test against (and rely on the compiler to convert things to structs for us).

walley892: Oh, I was assuming that this was used in the fragment shader tests. Could you add a test there?

engine/src/flutter/impeller/compiler/reflector.cc
```diff
@@ -869,19 +870,40 @@ std::vector<StructMember> Reflector::ReadStructMembers(
...
 
+    // Mat2
+    if (member.basetype == spirv_cross::SPIRType::BaseType::Float &&
+        member.width == 32 && member.columns == 2 && member.vecsize == 2) {
+      uint32_t count = array_elements.value_or(1) * 2;
+      uint32_t stride = 16;
+      uint32_t total_length = stride * count;
+
+      result.emplace_back(StructMember{
+          /*p_type=*/"Mat2",
```
walley892, line 896: Where is p_type used? Is it just something that gets written to the json output?

engine/src/flutter/impeller/compiler/reflector.cc
```diff
@@ -869,19 +870,40 @@ std::vector<StructMember> Reflector::ReadStructMembers(
...
+          /*p_base_type=*/spirv_cross::SPIRType::BaseType::Float,
+          /*p_name=*/GetMemberNameAtIndex(struct_type, i),
+          /*p_offset=*/struct_member_offset,
+          /*p_size=*/sizeof(Point),
+          /*p_byte_length=*/total_length,
+          /*p_array_elements=*/count,
+          /*p_element_padding=*/8,
```
walley892, line 903: What does p_size here mean? Is it the size of the entire matrix?

walley892: Oh wait, you're treating the matrix as an array of two vec2s, with two floats worth of padding between them? That's gnarly. That does conform to what the vulkan shader wants.

Could you add a comment above this explaining the layout?

walley892 (changes requested):
Wild. Main thing is it would be nice to have another test with the fragment shader stuff.

engine/src/flutter/impeller/compiler/reflector.cc
gaaclarke: I'm not sure who ingests it.  I've just asserted that that is in the output in the reflection tests to match what other sections are doing.

engine/src/flutter/impeller/fixtures/mat2_test.frag
gaaclarke: The test has to have the mat2 in a struct because this matches the behavior when we compile for vulkan (everything is in a virtual struct).  I think these tests only run on metal right now.  We'll sort it out as we continue to improve things.

engine/src/flutter/impeller/compiler/reflector.cc
gaaclarke: done

engine/src/flutter/impeller/fixtures/mat2_test.frag
gaaclarke: added mat2 integration test

engine/src/flutter/impeller/compiler/reflector.cc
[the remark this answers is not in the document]
walley892: I wonder if she's programed to leave at least one comment

engine/src/flutter/impeller/fixtures/mat2_test.frag
walley892: >I think these tests only run on metal right now

:(

>added mat2 integration test

Nice

engine/src/flutter/impeller/compiler/reflector.cc
walley892: Ah, okay

walley892 (approved):
Very slick programming by steamboat willie

PR #181024, opened 2026-01-15
Make HCPP upgrading work for vd/tlhc

gmackall:
Before the change we get a NPE when trying to cast the return of the create message as an int.

There are also some more changes needed on top of the change to the cast, making the `TextureAndroidViewController` aware of if it is using HCPP and exposing that bool for use in `packages/flutter/lib/src/rendering/platform_view.dart`

This is untested because it currently isn't possible to use HCPP without uncommenting the code block here: https://github.com/flutter/flutter/blob/2beceb6b1c0701ce0d3289dec41c7fa7f8678b9e/engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformViewsChannel.java#L95

It will be tested in the follow up pr which adds the cli flag, uncomments the block, and adds golden tests that we can upgrade each of the 3 legacy modes.

Fixes https://github.com/flutter/flutter/issues/181000


packages/flutter/lib/src/rendering/platform_view.dart
```diff
@@ -211,7 +211,8 @@ class RenderAndroidView extends PlatformViewRenderBox {
 
   @override
   void paint(PaintingContext context, Offset offset) {
-    if (_viewController.textureId == null || _currentTextureSize == null) {
+    if ((_viewController.textureId == null && !_viewController.requiresViewComposition) ||
```
reidbaker, line 214: Double check on this logic change is your intent 
(Null texture id and does not require view composition) or (null texture size)?

packages/flutter/lib/src/services/platform_views.dart
```diff
@@ -1256,21 +1255,29 @@ class TextureAndroidViewController extends AndroidViewController {
...
+      (_internals as _TextureAndroidViewControllerInternals).textureId = response;
+    } else {
+      // Fallback to HCPP.
+      _internals = _Hybrid2AndroidViewControllerInternals();
+    }
   }
 
   @override
   int? get textureId {
+    // If we've fallen back to HCPP, we don't have a texture ID.
```
reidbaker: In what situation is a HCPP a fallback?

packages/flutter/lib/src/services/platform_views.dart
```diff
@@ -1286,11 +1293,17 @@ class TextureAndroidViewController extends AndroidViewController {
 
   @override
   Future<Size> _sendResizeMessage(Size size) {
+    if (_internals.requiresViewComposition) {
```
reidbaker, line 1294: Can you add a comment explaining why this does not need to send the resize message?

packages/flutter/lib/src/services/platform_views.dart
```diff
@@ -1286,11 +1293,17 @@ class TextureAndroidViewController extends AndroidViewController {
...
+      return Future<Size>.value(size);
+    }
     return _internals.setSize(size, viewId: viewId, viewState: _state);
   }
 
   @override
   Future<void> setOffset(Offset off) {
+    if (_internals.requiresViewComposition) {
```
reidbaker, line 1302: same here for offset

reidbaker (approved):

packages/flutter/lib/src/rendering/platform_view.dart
gmackall: Yes, otherwise we would return early from paint in all cases in the hcpp flow, which is not what we would want (because texture id is always null in that case).

packages/flutter/lib/src/services/platform_views.dart
gmackall: This is just a bad comment, it doesn't say anything more than the simple if statement itself. Removed. But also yeah should not have said fallback

Issue #180382, opened 2025-12-30
Running Gradle task 'assembleDebug' keeps re-downloading huge amounts of files


nav9:
### Steps to reproduce

1. Create a new Flutter project, build it to create an APK file and make it run on the emulator. This causes Gradle task 'assembleDebug' to run, and it downloads Android SDK Build-Tools 35 v.35.0.0 or whatever it needs and installs it. This can sometimes take upto 500MB or almost 1GB of downloads. I make changes to the code and try to run it again and the Gradle task 'assembleDebug' does some 200MB or 300MB of downloads again just to be able to build the APK.
2. Create a new Flutter project and various similar files that were already downloaded get downloaded all over again for this new project too.

### Expected results

Not everyone has the luxury of living in a first world country that has super-fast internet and unlimited downloads. Even if you are in a first world country, as Engineers, your job is to optimize code and infrastructure. This repeated downloading of the same files  is a very poorly implemented bug (yes I'm calling it a bug) that needs to be fixed. 

### Actual results

Even if the actual issue is with the people who built and maintain Gradle, this issue can be fixed by programming Flutter in such a way that it can search for already downloaded files in nearby folders (by maintaining such info about prior builds/downloads in a cache) and avoid having to re-download everything every time a build is initiated.

### Code sample

<details open><summary>It could be any project. Not necessarily this code</summary>

```dart
import 'package:flutter/material.dart';
import 'package:flutter/services.dart';

void main() {
  runApp(const TimerApp());
}

class TimerApp extends StatelessWidget {
  const TimerApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: TimerScreen(),
      debugShowCheckedModeBanner: false,
    );
  }
}

class TimerScreen extends StatefulWidget {
  @override
  State<TimerScreen> createState() => _TimerScreenState();
}

class _TimerScreenState extends State<TimerScreen> {
  static const channel = MethodChannel('timer_service');

  int hours = 0;
  int minutes = 20;
  int seconds = 0;
  bool running = false;

  Future<void> startTimer() async {
    await channel.invokeMethod('startTimer', {
      "seconds": hours * 3600 + minutes * 60 + seconds,
    });
    setState(() => running = true);
  }

  Future<void> pauseTimer() async {
    await channel.invokeMethod('pauseTimer');
    setState(() => running = false);
  }

  Future<void> stopTimer() async {
    await channel.invokeMethod('stopTimer');
    setState(() => running = false);
[...]

PR #181039, opened 2026-01-15
Enable Gradle caching in gradle.properties template #180382

nav9:
Solving issue #180382 and related issues where users complain of repeated gradle assembleDebug downloads.

Users typically use `flutter create <appName>` to create their Flutter project. Many users have complained of gradle AssembleDebug doing repeated downloads and being unable to build an APK unless they were connected to the internet. The problem was that caching was not enabled in the gradle properties file. Enabling it allows Gradle to use the files that are already downloaded, to do the build without triggering any fresh downloads which can be taxing for users with limited daily download limits or slow/flaky internet connections

Solves issue[ #180382](https://github.com/flutter/flutter/issues/180382) and hopefully Gradle [issue #36269](https://github.com/gradle/gradle/issues/36269) too.

## Pre-launch Checklist

- [ yes ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ yes, but when creating the PR a notification said that submitting a change will write it to a new branch so I believe it does not affect tree hygiene ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ N/A ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ yes ] I signed the [CLA].
- [ yes ] I listed at least one issue that this PR fixes in the description above.
- [ N/A ] I updated/added relevant documentation (doc comments with `///`).
- [ N/A ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ N/A ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ N/A ] All existing and new tests are passing.

### DIFF
--- a/packages/flutter_tools/templates/app/android.tmpl/gradle.properties.tmpl
+++ b/packages/flutter_tools/templates/app/android.tmpl/gradle.properties.tmpl
@@ -1,2 +1,3 @@
 org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=4G -XX:ReservedCodeCacheSize=512m -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
+org.gradle.caching=true


gmackall (changes requested):
> The problem was that caching was not enabled in the gradle properties file. Enabling it allows Gradle to use the files that are already downloaded, to do the build without triggering any fresh downloads which can be taxing for users with limited daily download limits or slow/flaky internet connections

I believe this is not correct, it seems you are conflating Gradle's concept of the build cache
https://docs.gradle.org/current/userguide/build_cache.html
with its concept of dependency caching
https://docs.gradle.org/current/userguide/dependency_caching.html
(your flag affects the former, but your comment seems to indicate you believe it affects the latter (which is always enabled by default)).

Also re build caching, note that Gradle is already avoiding re-building that which was built in the last invocation through its use of "up-to-date" checks
https://docs.gradle.org/current/userguide/incremental_build.html#incremental_build

So if the problem is, as you say in your issue:
> I make changes to the code and try to run it again and the Gradle task 'assembleDebug' does some 200MB or 300MB of downloads again just to be able to build the APK.

Note that the cache you are enabling here, according to the Gradle docs:
> ... task output caching leverages the same intelligence as [up-to-date checks](https://docs.gradle.org/current/userguide/incremental_build.html#incremental_build) that Gradle uses to avoid work when a previous local build has already produced a set of task outputs.

So this would not have any effect for your specific case.

Also, this PR would not be able to progress without you signing the CLA

gmackall:
> @gmackall - I believe there may be something here however - specifically when switching between variants:
> 
> https://docs.gradle.org/current/userguide/build_cache_use_cases.html#speed_up_developer_builds_with_the_local_cache
> 
> > The local cache can also be useful when working with a project that has multiple variants, as in the case of Android projects. Each variant has a number of tasks associated with it, and some of those task variant dimensions, despite having different names, can end up producing the same output. With the local cache enabled, reuse between task variants will happen automatically when applicable.

Yeah I still think the gradle build cache is a feature that developers should consider, and we could consider turning it on by default. However, as this affects the build and not dependency cache I think this would still not have any effect on downloading, unless we are talking about downloading that is performed directly by the Flutter Gradle plugin (not AGP/Gradle) as part of the build, and is not properly cached.

nav9:
No problem. Adding `org.gradle.caching=true` has significantly reduced the massive downloads that used to happen earlier, so yes; it will surely help. Thank you for considering it.

PR #181040, opened 2026-01-15
Fix DropdownButtonFormField underline alignment at bottom

### LINKED ISSUE #162350: DropdownButtonFormField does not scale correctly to fit its parent's height
### Steps to reproduce

Stretch a DropdownButtonFormField to fill its parent widget's height

### Expected results

The UI should scale according to the widget height. This works when using DropdownButton instead of DropdownButtonFormFeild as shown in the screenshot, where the bottom border is moved lower to fill the remaining height.

### Actual results

The actual visible elements including the border are only rendered at the top of the field instead of expanded to the full height. The widget scales to full height and is clickable across the whole height, but the border and decoration doesn't scale accordingly.

### Code sample

<details open><summary>DropdownButton UI scales to fit its parents hide, wheares DropdownButtonFormField does not</summary>

```dart

void main() {
  runApp(
    MaterialApp(
      home: const TestView(),
    )
  );
}

class TestView extends StatelessWidget {
  const TestView({Key? key}) : super(key: key);

  @override
    List<DropdownMenuItem<int>> debugItems = [
      DropdownMenuItem(
        value: 1,
        child: Text("Option 1"),
      ),
      DropdownMenuItem(
        value: 2,
        child: Text("Option 2"),
      ),
    ];
    
    return Scaffold(
      body: SizedBox(
        height: 200,
        child: Row(
          children: [
            Expanded(
              child: DropdownButton<int> ( // rendered correctly
                isExpanded: true,
                value: 1,
                items: debugItems,
                onChanged: (int? value) {},
              ),
            ),
            Expanded(
              child: DropdownButtonFormField<int>( // looks weird
                isExpanded: true,
                value: 1,
                items: debugItems,
                onChanged: (int? value) {},
              ),
            ),
          ],
        ),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>
<summary>DropdownButton on the left, DropdownButtonFormField on the right</summary>

Image

</details>

### Flutter Doctor output

```console
[√] Flutter (Channel stable, 3.22.2, on Microsoft Windows [Version 10.0.19045.5247], locale de-DE)
    • Flutter version 3.22.2 on channel stable at D:\Workspace\flutter
    • Framework revision 761747bfc5 (8 months ago), 2024-06-05 22:15:13 +0200
    • Engine revision edd8546116
    • Dart version 3.4.3
    • DevTools version 2.34.3

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at D:\Workspace\Android\Sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: D:\Programme\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-b2043.56-10550314)

    • CHROME_EXECUTABLE = D:\Programme\Google\Chrome\Application\chrome.exe      

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.10.5)
    • Visual Studio at D:\Programme\Visual Studio\2022
    • Visual Studio Community 2022 version 17.10.35122.118
    • Windows 10 SDK version 10.0.22621.0

[√] Android Studio (version 2023.1)
    • Android Studio at D:\Programme\Android\Android Studio
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-b2043.56-10550314)

[√] Connected device (3 available)
    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [Version 10.0.19045.5247]
    • Chrome (web)      • chrome  • web-javascript • Google Chrome 132.0.6834.111
    • Edge (web)        • edge    • web-javascript • Microsoft Edge 131.0.2903.146

[√] Network resources

• No issues found!
```

</details>

Mairramer:
This PR fixes the visual issue where the underline of a `DropdownButtonFormField` did not align at the bottom of its expanded container.

Fixes flutter/flutter#162350

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1223,11 +1223,14 @@ class DropdownButton<T> extends StatefulWidget {
   /// its own decorations, like [InputDecorator].
   final bool isDense;
 
-  /// Set the dropdown's inner contents to horizontally fill its parent.
-  ///
-  /// By default this button's inner width is the minimum size of its contents.
-  /// If [isExpanded] is true, the inner width is expanded to fill its
-  /// surrounding container.
+  /// Controls whether the dropdown's inner contents expand to fill the
```
QuncCccccc: We changed the definition of isExpanded here, which might cause breaking changes. Originally, the documentation stated it was only used to fill the parent horizontally. I'm a little hesitant to approve it just yet—let me ask around for suggestions!

dkwingsmt: Yeah, I agree. Although it's kind of unfortunate that such a general name as `isExpanded` is used for horizontal only, it's only worse if we change the behavior and cause breakage. That being said, I can't think of a better name than `isVerticallyExpanded`. Feel free to suggest other names.

Mairramer: This only applies  when it doesn’t have a parent with a defined height. If a height is specified, it directly affects how `isExpanded` behaves.

```dart
SizedBox(
  height: 200,
  child: DropdownButton<int>(
    isExpanded: true,
    value: 1,
    items: debugItems,
    onChanged: (int? value) {},
  ),
),
```
That’s why I made this change to the documentation. However, on second thought, this could potentially cause a breaking change. The chances are minimal, but it could still happen.


So I believe we can improve the description itself. Do you agree?

QuncCccccc: hmm I don't think the breaking change is minimum actually. Originally, it keeps the minimum height itself but with the change, the height can be expanded in many cases. But I do agree that this API name and its behavior don't match each other. We should either deprecate the existing one and introduce a new API to expand both vertical and horizontal, or introduce a new API for vertical expansion only.

Mairramer: I agree. Maybe we can follow @dkwingsmt’s suggestion and introduce an `isVerticallyExpanded` parameter (or a better name) to address this.

I also considered adding an `expandHeight` parameter, but that would introduce something into the API that merely masks a bug or an unintended behavior rather than fixing it.

I’m unsure now.

QuncCccccc: Introducing a `isVerticallyExpanded` sounds good to me:)! We should make sure that when it's true, only height is expanded.

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1019,6 +1019,7 @@ class DropdownButton<T> extends StatefulWidget {
     this.barrierDismissible = true,
     this.mouseCursor,
     this.dropdownMenuItemMouseCursor,
+    this.isVerticallyExpanded = true,
```
Mairramer: I left this set to `true` because if it were `false`, it would break many tests that currently rely on the unexpected behavior of the height being expanded.

QuncCccccc: Sorry for the late reply. I'm a little confused about why it will cause breaking change. I thought originally the height will not be expanded, is that correct?

Mairramer: Consider the following examples:

 ```dart
    return Scaffold(
      body: SizedBox(
        height: 200,
        child: DropdownButton<int>(
          isExpanded: true,
          value: 1,
          items: debugItems,
          onChanged: (int? value) {},
        ),
      ),
    );
 ```

 or

 ```dart
 return Scaffold(
   body: ListView(
     children: <Widget>[
       DropdownButton<String>(
         value: value,
         items: <DropdownMenuItem<String>>[
           DropdownMenuItem<String>(
             key: itemKey,
             value: value,
             child: const Text(value),
           ),
         ],
         onChanged: (_) {},
       ),
     ],
   ),
 );
 ```

If we set `isVerticallyExpanded` to `false`, this breaks some adjacent behaviors. In my opinion, those behaviors are incorrect (or at least questionable), but changing them would currently break a large number of tests that rely on the existing behavior.

Actual behavior https://dartpad.dev/703b12049925021469f2fe01812db978

Mairramer: I have a slight feeling that I might be overlooking something.

Mairramer: After reviewing it more carefully, I found the cause. That said, setting this to false would introduce a minor breaking change affecting three tests.

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1334,6 +1336,27 @@ class DropdownButton<T> extends StatefulWidget {
...
+  ///
+  /// When true, the inner height of the dropdown is expanded to match its
+  /// parent container. When false, the inner height is determined by the
+  /// height of the selected item (its intrinsic height).
+  ///
+  /// This is particularly useful when the dropdown is used inside a fixed-height
+  /// container, ensuring that the underline and items align properly within
+  /// the allocated space.
+  ///
+  /// Defaults to true for [DropdownButton] to maintain its standard behavior,
```
QuncCccccc: The default value for DropdownButton is not aligned with the doc here. Seems we should keep the default value to `true` for DropdownButton to maintain its existing behavior.

packages/flutter/test/material/dropdown_test.dart
```diff
@@ -3934,6 +3934,7 @@ void main() {
         home: Material(
           child: DropdownButton<String>(
             key: menuKey,
+            isVerticallyExpanded: true,
```
QuncCccccc: If we have to add this to make the tests pass, it means we might also break a lot of tests for developers' apps, which is not desired.

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1019,6 +1019,7 @@ class DropdownButton<T> extends StatefulWidget {
     this.barrierDismissible = true,
     this.mouseCursor,
     this.dropdownMenuItemMouseCursor,
+    this.isVerticallyExpanded = false,
```
QuncCccccc: If we are only fixing the expansion issue on DropdownButtonFormField, we don't need to expose this here in the DropdownButton constructor. Instead, we can add isVerticallyExpanded = true; after line 1040.

packages/flutter/lib/src/material/dropdown.dart
QuncCccccc: We can remove the doc related to DropdownButton if we don't expose the property to the DropdownButton constructor. But in implementation, we should make sure no breaking changes are introduced to DropdownButton.

Mairramer: > The default value for DropdownButton is not aligned with the doc here. Seems we should keep the default value to `true` for DropdownButton to maintain its existing behavior.

Oops, that was my mistake. I forgot to update it.

Mairramer: I intend to keep it in both locations to preserve the following behaviors:

isVerticallyExpanded: true
Captura de Tela 2026-02-26 às 18 37 08

isVerticallyExpanded: false
Captura de Tela 2026-02-26 às 18 36 49

packages/flutter/test/material/dropdown_test.dart
Mairramer: I’ll remove these.

packages/flutter/lib/src/material/dropdown.dart
QuncCccccc: Follow up on this [comment](https://github.com/flutter/flutter/pull/181040#discussion_r2861312133) just in case it's missed:) I think if we are only fixing the issue on DropdownButtonFormField, maybe it's not necessary to expose isVerticallyExpanded in the DropdownButton constructor here. We can do something like below to keep its default behavior and not add an extra field in the constructor to maintain.
DropdownButton(
  ...
) :  ...isVerticallyExpanded = true;

QuncCccccc (commented):
Sorry for the late response. I'll make sure we made progress this week:)

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1334,6 +1336,27 @@ class DropdownButton<T> extends StatefulWidget {
...
+  /// When true, the inner height of the dropdown is expanded to match its
+  /// parent container. When false, the inner height is determined by the
+  /// height of the selected item (its intrinsic height).
+  ///
+  /// This is particularly useful when the dropdown is used inside a fixed-height
+  /// container, ensuring that the underline and items align properly within
+  /// the allocated space.
+  ///
+  /// Defaults to true for [DropdownButton] to maintain its standard behavior,
+  /// and false for [DropdownButtonFormField] to ensure the input decorator
```
QuncCccccc:
```suggestion
  /// Defaults to false for [DropdownButtonFormField] to ensure the input decorator
```

If we don't expose this field to DropdownButton constructor, I think there's no need to mention it here.

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1687,6 +1710,7 @@ class _DropdownButtonState<T> extends State<DropdownButton<T>> with WidgetsBindi
           minWidth: widget.iconSize + suffixIconEndMargin,
           minHeight: widget.iconSize,
         ),
+
```
QuncCccccc:
```suggestion
```

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1751,6 +1776,19 @@ class _DropdownButtonState<T> extends State<DropdownButton<T>> with WidgetsBindi
...
+    // // available height. Only the horizontal axis remains constrained,
+    // // preserving the width defined by the parent.
+    if (!widget.isVerticallyExpanded) {
+      result = UnconstrainedBox(
+        constrainedAxis: Axis.horizontal, // Keep width constrained by parent
+        alignment: Alignment.topCenter,
+        child: result,
+      );
+    }
+
```
QuncCccccc: Seems this blocks of code is not necessary?

packages/flutter/test/material/dropdown_form_field_test.dart
```diff
@@ -1056,7 +1056,12 @@ void main() {
     expect(find.text(currentValue), findsOneWidget);
 
     // Tap the DropdownButtonFormField widget
-    await tester.tap(find.byType(DropdownButton<String>));
+    await tester.tap(
```
QuncCccccc: Why do we make this change here?

QuncCccccc (commented):
Overall looks good to me! Just left some comments.

packages/flutter/lib/src/material/dropdown.dart
Mairramer: Actually, it is necessary, since the decorator area would otherwise become clickable.

packages/flutter/test/material/dropdown_form_field_test.dart
Mairramer: It’s due to the `UnconstrainedBox`, which now correctly constrains the height.

packages/flutter/lib/src/material/dropdown.dart
QuncCccccc: Can you give me an example? I'm still a bit confused.

Mairramer: https://github.com/user-attachments/assets/c0ac43d4-d89f-4f77-b685-b953a017d0f2

QuncCccccc: I see. Then can we replace `UnconstrainedBox` with `Align`? One benefit of Align that I can think of is when the parent height is smaller than the DropdownButtonFormField, it is still safe and the field is clipped. But UnconstrainedBox will throw error.

Mairramer: It will have the same behavior.

packages/flutter/test/material/dropdown_form_field_test.dart
Mairramer: The tap always internally uses `getCenter`, but since the height is now constrained, it will no longer be able to click the actual center. By default, windows are 800×600, but with this change the `DropdownButtonFormField` will only occupy the required size (400×300), so the tap will no longer work.

Mairramer: I didn’t want to change the test, but I think this is the only one that will actually require manual intervention due to the fix of an “incorrect” behavior.

packages/flutter/lib/src/material/dropdown.dart
QuncCccccc: I don't think they have the same behavior. Please try this:
```dart
Center(
        child: SizedBox(
          width: 200,
          height: 20,  // Smaller parent height than DropdownButtonFormField.
          child: ColoredBox(
            color: Colors.blue,
            child: DropdownButtonFormField<int>(
              isVerticallyExpanded: false,
              items: debugItems,
              onChanged: (int? value) {},
            ),
          ),
        ),
      )
```

We won't get any error with Align
Screenshot 2026-03-05 at 10 22 02 AM

But bottom overflow with UnconstrainedBox
Screenshot 2026-03-05 at 10 23 05 AM

Mairramer: humm, I see.

packages/flutter/test/material/dropdown_form_field_test.dart
QuncCccccc: Yeah, it makes sense!

QuncCccccc (dismissed):
LGTM! Once all the tests have passed, I'll ask for a second reviewer:)

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1744,13 +1766,21 @@ class _DropdownButtonState<T> extends State<DropdownButton<T>> with WidgetsBindi
...
             ),
           ),
         ),
       );
+
+      // When vertical expansion is disabled, wrap with Align
+      // to prevent the widget from stretching to fill the parent's
+      // available height.
+      if (!widget.isVerticallyExpanded) {
+        result = Align(alignment: Alignment.topCenter, heightFactor: 1.0, child: result);
```
QuncCccccc:
```suggestion
        result = Align(alignment: AlignmentDirectional.topStart, heightFactor: 1.0, child: result);
```

Changing it to topStart should fix the failed google testing. Some customer tests are failed because the dropdownButton will be located in top center when parent width is large.

QuncCccccc (dismissed):
LGTM

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1340,6 +1342,26 @@ class DropdownButton<T> extends StatefulWidget {
...
+  /// container, ensuring that the underline and items align properly within
+  /// the allocated space.
+  ///
+  /// Defaults to false for [DropdownButtonFormField] to ensure the input decorator
+  /// handles the vertical constraints correctly by default.
+  ///
+  /// See also:
+  ///
+  ///  * [DropdownButton.isExpanded], which expands the inner width of the
+  ///    dropdown to fill the available horizontal space.
```
justinmc, line 1366: Could you also add a "See also" linking back to isVerticallyExpanded from the isExpanded docs?

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1750,13 +1772,21 @@ class _DropdownButtonState<T> extends State<DropdownButton<T>> with WidgetsBindi
...
           ),
         ),
       );
+
+      // When vertical expansion is disabled, wrap with Align
+      // to prevent the widget from stretching to fill the parent's
+      // available height.
+      if (!widget.isVerticallyExpanded) {
+        result = Align(alignment: AlignmentDirectional.topStart, heightFactor: 1.0, child: result);
+      }
```
justinmc, line 1793: I haven't looked at the Google failures, but this looks like a slight default behavior change that could cause a failure for the _formField case.

justinmc (dismissed):
LGTM if the Google tests pass👍

packages/flutter/lib/src/material/dropdown.dart
Mairramer: Done.

QuncCccccc (approved):
LGTM.

dkwingsmt:
Status: g3fix has been approved, and Google Testing has passed with the g3fix. I'm deflaking the other presubmit checks. Once everything is green, we can mark this PR as complete and archive it for unfreezing.

QuncCccccc:
Seems we need a roller to approve the CL to pass google testing.

Screenshot 2026-04-22 at 5 10 33 PM

dkwingsmt:
It's ok. The Google test has passed at least once, indicating that this is probably ok. I'll mark that this PR has reached completion.

At the moment we're not able to merge this PR for the moment due to the Material/Cupertino code freeze (https://github.com/flutter/flutter/issues/184093). As soon as the decoupling is complete, we will post a way to port this PR to the new packages to land. In the meantime, I'll convert this PR to a draft and apply a label so that we can quickly find it when the freeze is lifted.

Thanks for your contribution!

Piinks:

Piinks:

This PR can be ported over to flutter/packages!

Mairramer:
@QuncCccccc @dkwingsmt Ported to https://github.com/flutter/packages/pull/12551.

PR #181047, opened 2026-01-15
Remove use of SkDeserialProcs::fTypefaceProc

bungeman:
SkDeserialProcs::fTypefaceProc took a void* and size_t, but the expectation was that the void* was a SkStream** and the size_t was always sizeof(SkStream*). Users should now use the less confusing replacement SkDeserialProcs::fTypefaceStreamProc which takes an SkStream&.


jason-simmons (approved):

flar (approved):
LGTM

PR #181092, opened 2026-01-16
Introduce ScrollCacheExtent and also fixes unbound shrinkwrap cache ex…

chunhtai:
…tent NaN problem

besides the general piping through parameters and breaking change. Fixing an issue when the shrinkwrap: true the mainAxisExetent will be double.infinity, which cause the cacheExtent to be double.Nan after the calculation and crash the app.

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter/lib/src/rendering/viewport.dart
```diff
@@ -23,6 +23,99 @@ import 'object.dart';
 import 'sliver.dart';
 import 'viewport_offset.dart';
 
+/// The amount of additional content to cache around the viewport.
+///
+/// See also:
+///
+///  * [Viewport.scrollCacheExtent], which uses this class to define the cache area.
+@immutable
+sealed class ScrollCacheExtent {
```
Piinks, line 32: This is so nice. Chef's kiss.

packages/flutter/lib/src/rendering/viewport.dart
```diff
@@ -23,6 +23,99 @@ import 'object.dart';
...
+  const ScrollCacheExtent._();
+
+  /// Creates a cache extent in logical pixels.
+  const factory ScrollCacheExtent.pixels(double pixels) = _PixelScrollCacheExtent;
```
Piinks, line 36: Should we have a dart fix that transforms a previously set cacheExtent to this with the value provided as `pixels`?

packages/flutter/lib/src/rendering/viewport.dart
Piinks: WDYT about a ScrollCacheExtent that is for lazy loaders like RenderSliverList that allows the user to specify the number of children to lay out beyond the viewport bounds that are not visible?
Maybe something for later, not blocking.

Piinks: Meh, this might be harder. Here currently, both methods boil down to a pixel value.
Maybe for the other child-based idea, we should consider it a feature on lazy loader to ensure there is always one fully out of bounds child available to progress scroll forward or backward to. Just musing.

packages/flutter/lib/src/rendering/viewport.dart
```diff
@@ -421,23 +528,48 @@ abstract class RenderViewportBase<ParentDataClass extends ContainerParentDataMix
     markNeedsLayout();
   }
 
-  // TODO(ianh): cacheExtent/cacheExtentStyle should be a single
```
Piinks, line 424: Hooray for resolving old TODOs!

packages/flutter/lib/src/rendering/viewport.dart
```diff
@@ -1618,6 +1751,7 @@ class RenderViewport extends RenderViewportBase<SliverPhysicalContainerParentDat
   double _attemptLayout(double mainAxisExtent, double crossAxisExtent, double correctedOffset) {
     assert(!mainAxisExtent.isNaN);
     assert(mainAxisExtent >= 0.0);
+    assert(mainAxisExtent.isFinite);
```
Piinks, line 1754: Is this safe to add? Does the infinite scroll view case take a different path?

Piinks: How does this differ from the shrink wrapping viewport below?

packages/flutter/test/widgets/scroll_cache_extent_test.dart
```diff
@@ -0,0 +1,201 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
Piinks, line 1: These mostly check the property is set 👍  Can you add a test that verifies the child layout actually respects this? e.g. One layout pass with scrollCacheExtent lays out 3 children, update value, and 5 children are laid out or some such?

Piinks (commented):
This is awesome. Thank you. 🙏

packages/flutter/lib/src/rendering/viewport.dart
chunhtai: can a viewport to be in a unbounded container? I thought that will throw an layout error. but i am not too sure

> How does this differ from the shrink wrapping viewport below?

I think only shrink wrap viewport is allowed to be layout in unbound situation

packages/flutter/test/widgets/scroll_cache_extent_test.dart
chunhtai: I think that is covered in existing test case, the renderobject layer all use ScrollCacheExtent directly now

chunhtai: actually i should probably add a test case for the based scrollview

packages/flutter/lib/src/rendering/viewport.dart
Piinks: Ah yeah I see now. Normal viewport checks debugCheckHasBoundedAxis during computeDryLayout. Do you want to call that method instead of the assertion?
Shrinkwrapping does in fact allow infinite main axis extent. 👍

packages/flutter/test/widgets/scroll_cache_extent_test.dart
Piinks: > I think that is covered in existing test case, the renderobject layer all use ScrollCacheExtent directly now

Totally agree, I just meant for like the new one that is a multiple of the viewport extent.

packages/flutter/lib/src/rendering/viewport.dart
chunhtai: by child count will be easier to implement with the new class. but the main problem with child count is that whether it is renderobject child count or semantics child count. By the use case of cache extent, it makes more sense to be by semantics child count. Such information is only available after semantics tree build, which is at the very end of the draw frame. That means we may need multiple run of layout to properly implement cache by semantics child count

packages/flutter/lib/src/rendering/viewport.dart
chunhtai: this will depends on cacheExtenType, is it possible to conditionally use different constructor based on whether the other parameter is set or not?

packages/flutter/lib/src/rendering/viewport.dart
chunhtai: debugCheckHasBoundedAxis checks against constraints which this method has already teared down into axis extent. I will keep this assert as is. let me know otherwise

packages/flutter/test/widgets/scroll_cache_extent_test.dart
chunhtai: done

packages/flutter/lib/src/rendering/viewport.dart
Piinks: Right right! Gotcha. This assert should be ok then. 👍

packages/flutter/lib/src/rendering/viewport.dart
Piinks: Yes, in the fix, you should be able to check cacheExtent, cachExtentStyle, and then add the new parameter and provide the right constructor based on the values. Them you can use remove parameter to remove the old ones.

Piinks: Why don't we do the dart fix in a separate change? There are many classes to account for, and it might be a bit complicated.

Piinks (approved):
This LGTM. We can do the dart fix in a separate change 👍 
I'll file an issue to follow up on the 2D scrolling package to migrate those when this gets to stable.

chunhtai:
I will include dart fix since gemini finds it easy to add

Piinks (approved):
This LGTM!
Do you think we need a migration guide on the website?

PR #181093, opened 2026-01-17
Broaden the applicability of `Color` fix_data rules from flutter/painting.dart

b-luk:
See the conversation at https://github.com/dart-lang/sdk/issues/59764#issuecomment-3599823329 for more context.

Prior to this, these `Color` fix_data rules would only be visible to users that explicitly import `package: flutter/painting.dart`. This PR broadens the applicability of these rules so they are visible for users that import `dart:ui`, or import any one of the `flutter/` packages that expose `Color`.

#### Enabling `Color` fix_data fixes for `dart:ui`:
- Adds a `fix_data.yaml` file for `'dart:ui'` fixes to `engine/src/flutter/lib/ui/`. This contains `Color` fixes copied from `packages/flutter/lib/fix_data/fix_painting.yaml`.
- Typically `fix_data` files go in a Dart package's `lib/` directory. But `dart:ui` is not a standard Dart package and doesn't have a typical `lib` directory. This directory is the closest thing to a `lib` directory for the `dart:ui` package that I can find. It contains the Dart code that implements the `dart:ui` package.
- The actual Dart package for `dart:ui` is generated by `engine/src/flutter/sky/packages/sky_engine/BUILD.gn` with the package name  `sky_engine`. I updated the build rule to copy the new `fix_data.yaml` file into the generated `sky_engine/lib/` directory.

#### Enabling `Color` fix_data fixes for `package:flutter/` libraries:
- The `Color` fixes from `packages/flutter/lib/fix_data/fix_painting.yaml` are remove, and copied to over to a new `packages/flutter/lib/fix_data/fix_dart_ui.yaml` file.
- This fix_data file specifies all the `flutter/` libraries which export `Color`. With this change, importing any of these libraries will surface the `Color` fixes, rather than only surfacing the fixes only when `package:flutter/painting.dart` is specifically imported.

#### Other related changes
- Updates some entries in `fix_cupertino.yaml` and `fix_material.yaml` which used the old `Color.withOpacity` method.
- Updates existing flutter package `test_fixes` tests.
- The color tests from the `painting` `test_fixes` test are moved to a new `dart_ui` test. Note that this only runs the test with only one of the `flutter/` libraries specified in `fix_dart_ui.yaml` (`animation.dart`). Ideally I think we should run this test while independently importing each one of the `flutter/` libraries specified in `fix_dart_ui.yaml`. But AFAICT there's no easy way to do this without creating a lot of duplicate code.
- There is no `fix_data` test for `dart:ui`. Because this isn't a standard Dart library, I don't think there's an easy way to test its `fix_data.yaml` in a similar way to the `flutter/` library fixes.

Somewhat fixes #180933 (makes the fixes more broadly available).
Somewhat addresses https://github.com/dart-lang/sdk/issues/59764#issuecomment-3599823329 (establishes a way to use `fix_data.yaml` files for `dart:X` libraries).

## Pre-launch Checklist

- [x] I signed the [CLA].


bwilkerson (approved):
The change lgtm, but I don't have the ability to approve changes in the flutter repo.

packages/flutter/lib/fix_data/fix_dart_ui.yaml
```diff
@@ -0,0 +1,101 @@
...
+# requiring the fix was made.
+
+# Every fix must be tested. See the flutter/packages/flutter/test_fixes/README.md
+# file for instructions on testing these data driven fixes.
+
+# For documentation about this file format, see
+# https://dart.dev/go/data-driven-fixes.
+
+# * Fixes in this file are for classes exported from the 'dart:ui' library. *
+
```
gaaclarke, line 23: I know there is a lot of text up here already, but we should at least have a link to the other file since this is duplicated.  We should make sure we don't change one without the other.

engine/src/flutter/sky/packages/sky_engine/BUILD.gn
```diff
@@ -201,6 +201,12 @@ copy("copy_dart_ui") {
   outputs = [ "$root_gen_dir/dart-pkg/sky_engine/lib/ui/{{source_file_part}}" ]
 }
 
+copy("copy_dart_ui_fix_data") {
+  sources = [ dart_ui_fix_data_file ]
```
gaaclarke, line 205: Could the source of this rule just be the other file?  They look exactly the same.

engine/src/flutter/lib/ui/fix_data.yaml
gaaclarke, line 1: It would be helpful if the name of the source filename matches the destination of the copy.

gaaclarke (commented):
Mostly looks good to me, i just have a few organizational questions.

packages/flutter/lib/fix_data/fix_dart_ui.yaml
b-luk: Done. I updated both this packages/flutter fix file and also the engine's fix file to have a bit more context and to have them reference each other.

engine/src/flutter/sky/packages/sky_engine/BUILD.gn
b-luk: This source uses the engine's fix file. We shouldn't use the packages/flutter fix file here.

The engine's file uses 'dart:ui' as the URI for its fixes.

The packages/flutter file specifies 'package:flutter/*' libraries as the URIs for its fixes.

engine/src/flutter/lib/ui/fix_data.yaml
b-luk: Done

gaaclarke (approved):
lgtm, thanks for digging into this.  @loic-sharma is going to be excited

b-luk:
> Created [fzyzcjy/flutter_portal#127](https://github.com/fzyzcjy/flutter_portal/pull/127) to fix the broken customer_testing tests.

Ended up disabling the tests instead (https://github.com/flutter/tests/pull/479). Updated this PR to update the customer testing commit version to include disabling those tests.

Issue #180827, opened 2026-01-12
[Google3 Bug]: [Crash Report] `anonymous namespace'::PolygonInfo::AppendVertex-dead1d17


calebmah:
### Help us understand the severity of this issue

- [x] causing severe production issues e.g. malfunctions or data loss
- [x] blocking next binary release
- [ ] blocking a client feature launch within a quarter
- [ ] nice-to-have but does not block a launch within the next quarter

### Steps to reproduce

Unable to verify detailed reproduction steps, see [b/475054280](https://b.corp.google.com/issues/475054280) for full crash report

This crash happens when using Impeller on iOS. From crash logs, appears to happen when rendering `DrawRoundRect` with blurred shadow. 

Likely an issue in the `PolygonInfo::AppendVertex` function within [impeller/entity/geometry/shadow_path_geometry.cc](https://github.com/flutter/flutter/blame/master/engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc). This possibly happens because the code attempts to access the .back() element of an empty vector on [line 1362](https://github.com/flutter/flutter/blame/master/engine/src/flutter/impeller/entity/geometry/shadow_path_geometry.cc#L1362).

### Expected results

The application should render the UI element and its associated shadow correctly without crashing.

### Actual results

The application crashes with an `EXC_BAD_ACCESS` in impeller::PolygonInfo::AppendVertex (See [b/475054280](https://b.corp.google.com/issues/475054280) for full crash report).

### Code sample

<details open><summary>Code sample</summary>

```dart
// shadow_path_geometry.cc:1362
uint16_t PolygonInfo::AppendVertex(const Point& vertex, Scalar gaussian) {
  FML_DCHECK(gaussian >= 0.0f && gaussian <= 1.0f);
  uint16_t index = vertices_.size();
  FML_DCHECK(index == gaussians_.size());
  // ...
  if (gaussian == gaussians_.back() && vertex == vertices_.back()) { // <--- CRASH HERE
    return index - 1;
  }
  // ...
}
```

Potential fix:

```dart
  if (!vertices_.empty() && gaussian == gaussians_.back() && vertex == vertices_.back()) {
```

</details>

### Logs

<details open><summary>Logs</summary>

```console
0x00000001074255cc (Flutter - shadow_path_geometry.cc: 1362)	(anonymous namespace)::PolygonInfo::AppendVertex(impeller::TPoint<float> const&, float)
[...]

PR #180920, opened 2026-01-13
[Impeller] Fix crash trying to check for duplicate vertices in shadow_path code

flar:
Fixes https://github.com/flutter/flutter/issues/180827

The new general convex path shadow code was checking for duplicate vertices without actually checking if the vectors contained any vertices. Thus, <vector>.back() was being called on empty vectors, which is bad. This led to a crash in G3 as their code is being run with a vector implementation that protects against this, but apparently we do not have such a vector implementation in our local building and testing.

No tests because this covered by existing test cases and a new FML_DCHECK, however I'm not sure if we have DCHECKs enabled in CI...?

PR #181095, opened 2026-01-17
[beta] [Impeller] Fix crash trying to check for duplicate vertices in shadow_path code (#180920)

flar:
The new general convex path shadow code was checking for duplicate vertices without actually checking if the vectors contained any vertices. Thus, <vector>.back() was being called on empty vectors, which is bad. This led to a crash in G3 as their code is being run with a vector implementation that protects against this, but apparently we do not have such a vector implementation in our local building and testing.

No tests because this covered by existing test cases and a new FML_DCHECK.

Fixes https://github.com/flutter/flutter/issues/180827

Impacted users: Any user who uses shadows (pretty much every Flutter app) on a platform that has address sanitization built into its `std::vector` implementation, if they use Impeller (always used on iOS and the default on Android)

Impact Description: Every shadow render operation in Impeller backend will access `back()` on an empty `std::vector` which will crash immediately if the vector is using sanitization on that platform.

Workaround: Potentially look into disabling address sanitization? On platforms that still support Skia, it can be used as a workaround, but that is not an option on iOS

Risk: The fix was obvious and non-controversial.

Test Coverage: Existing tests cover this case really well and there is now an FML_DCHECK (engine assert) to check if we trigger the condition under testing.

Validation: The original issue indicates an app that will trigger the sanitization crash, but only in G3. None of the platforms we support directly through flutter perform this sanitization.

## Pre-launch Checklist

- [x] I signed the [CLA].


camsim99 (approved):
This LGTM!

jtmcdole (approved):
Needed change

flar:
> I reached out to the author this morning and let them know I could not merge it because of failing tests. I tried to land it but yes it has missed the cut for the next release.

The failing tests were already in the tree so I needed to wait for their fix to go in. I've merged to ToT to see if they pass now.

PR #181099, opened 2026-01-17
Make sure that an ImageIcon doesn't crash in 0x0 environment

ahmedsameha1:
This is my attempt to handle https://github.com/flutter/flutter/issues/6537 for the ImageIcon widget.


packages/flutter/test/widgets/image_icon_test.dart
```diff
@@ -97,4 +97,14 @@ void main() {
     );
     handle.dispose();
   });
+
+  testWidgets('ImageIcon does not crash at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      const Directionality(
+        textDirection: TextDirection.ltr,
+        child: Center(child: SizedBox.shrink(child: ImageIcon(null))),
+      ),
```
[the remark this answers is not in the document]
victorsanni: Agree, can we test with non-null image?

packages/flutter/test/widgets/image_icon_test.dart
```diff
@@ -97,4 +97,14 @@ void main() {
     );
     handle.dispose();
   });
+
+  testWidgets('ImageIcon does not crash at zero area', (WidgetTester tester) async {
+    await tester.pumpWidget(
+      Directionality(
+        textDirection: TextDirection.ltr,
+        child: Center(child: SizedBox.shrink(child: ImageIcon(image))),
```
victorsanni, line 105: There are leaks, should `image` be disposed or similar?

dkwingsmt: I guess we need to add a `    imageCache.clear();` at the end.

victorsanni (approved):

dkwingsmt (approved):

Issue #150611, opened 2024-06-21
showDatePicker(): Allowing locale override for InputDatePickerFormField only


JeromeGsq:
### Steps to reproduce

1. Set locale of your app to `Locale('ar')`
2. call `showDatePicker();`
3. The date in the textfield is show arab oriental numbers and not arab numbers.

<img width="499" alt="image" src="https://github.com/flutter/flutter/assets/6270529/4f59099c-608a-4a28-8509-5b47c0e7ae14">

### Expected results

Allowing override local only for TextField.
I want to show "2020/10/10" date for example

### Actual results

The date in the textfield is show arab oriental numbers and not arab numbers.

### Code sample

```dart
await showDatePicker(
        context: context,
        initialDate: selectedDate,
        firstDate: DateTime(2000),
        lastDate: DateTime(2100));
}
  ```

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.22.1, on macOS 14.5 23F79
    darwin-arm64, locale fr-FR)
    • Flutter version 3.22.1 on channel stable at
      /Users/jeromegsq/Documents/Software/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision a14f74ff3a (4 weeks ago), 2024-05-22
      11:08:21 -0500
    • Engine revision 55eae6864b
    • Dart version 3.4.1
    • DevTools version 2.34.3

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/jeromegsq/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/jeromegsq/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 15.0)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15A240d
    • CocoaPods version 1.15.2

[✓] Chrome - develop for the web
    • CHROME_EXECUTABLE = /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2023.2)
    • Android Studio at /Applications/Android Studio.app/Contents
[...]

PR #181105, opened 2026-01-17
Support custom date input formatters via CalendarDelegate

### LINKED ISSUE #62401: showDatePicker has date format issues when DatePickerEntryMode.input is selected
## Steps to Reproduce

1. Call showDatePicker method and choose edit mode in dialog. (Tested with 'tr' locale)
2. Try to change the TextField manually. It changes from dot(.) format to slashes(/). Then when you enter, you get invalid format error. If you delete them all, then you can only enter digits because you don't have control over keyboard.

**Expected results:** 
What I expect from this feature(DatePickerEntryMode.input) is to be able to disable/enable it. If I can't disable it, I want to control it. I want to be able to specify date format, control keyboard etc.

**Actual results:** 
As specified at step 2, when you try to change the TextField manually, It changes from dot(.) format to slashes(/). Then when you enter, you get invalid format error. If you delete them all, then you can only enter digits because you don't have control over keyboard.

<details>
  <summary>Logs</summary>

```
```

```
```

```
[✓] Flutter (Channel stable, v1.17.5, on Mac OS X 10.15.5 19F101, locale en-TR)
    • Flutter version 1.17.5 at /Users/figengungor/Documents/development/flutter
    • Framework revision 8af6b2f038 (4 weeks ago), 2020-06-30 12:53:55 -0700
    • Engine revision ee76268252
    • Dart version 2.8.4

 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
    • Android SDK at /Users/figengungor/Library/Android/sdk
    • Platform android-29, build-tools 29.0.2
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] Xcode - develop for iOS and macOS (Xcode 11.5)
    • Xcode 11.5, Build version 11E608c
    • CocoaPods version 1.8.4

[✓] Android Studio (version 4.0)
    • Flutter plugin version 47.1.2
    • Dart plugin version 193.7361
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[!] Connected device
    ! No devices available

```

</details>

### LINKED ISSUE #83302: Support providing DateFormat for showDatePicker and showDateRangePicker
## Use case

When entering a Date in input mode rather than calendar mode, applications are unable to specify a format to parse other than en_US. [This can be seen here.](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/material_localizations.dart#L711) This results in applications being forced to use the default en_US parsing of the input text even when that is not the DateFormat.

The date_range_form_field package wraps the showDateRangePicker function into a FormField widget. It displays the selected date in a DateFormat specified by the application. However, as stated above, in the showDateRangePicker, the input format will always be en_US, regardless of the DateFormat being used in the date_range_form_field package or elsewhere in the app. In order for date_range_form_field to do the right thing, it needs to have a change in showDateRangePicker. For consistency, showDatePicker should be updated as well.

## Proposal

Add an optional DateFormat parameter to showDatePicker and showDateRangePicker to specify what DateFormat should be displayed and parsed.

### LINKED ISSUE #97134: Allow custom date formats for the showDatePicker input mode
## Use case

I am trying to get the text-based date input in the `showDatePicker` dialog to be based on specific date format (in my case DD/MM/YY). Currently, showDatePicker only takes a `Locale` parameter, which means that I am stuck with the choices that the available locales offer me. To my knowledge, there is no locale which would fit my use-case. As a result, I now display the date in DD/MM/YY everywhere except in the text-based input entry mode of the date-picker where my locale dictates MM/DD/YYYY, which is obviously not ideal.

## Proposal

The `showDatePicker` method should take an optional parameter `dateFormat` which directly controls how the date may be entered when the user switches to `DatePickerEntryMode.input`. This format should override the locale's setting if necessary.

### LINKED ISSUE #115151: Add inputFormatters option in DatePickerDialog and InputDatePickerFormField
## Use case

When using `DatePickerDialog` or `InputDatePickerFormField` in some Android devices(E.g. Samsung)  slash symbol(/) is not available for `TextInputType.datetime`, therefore the only way to use `DatePickerEntryMode.input` mode is to use a different `keyboardType` which makes it less efficient.

## Proposal

Add `inputFormatters` option which will allow developers to use something like `mask_text_input_formatter` to autocomplete slashes and improve the overall usability of the `DatePickerEntryMode.input` and `DatePickerEntryMode.inputOnly` modes.

### LINKED ISSUE #181043: showDatePicker manual input does not support inputFormatters
### Use case

# Allow customization of the manual input field in showDatePicker

## Problem

When using `showDatePicker` with `DatePickerEntryMode.input` or `inputOnly`, the manual date entry field does not expose any way to customize text input behavior.

Specifically, it is not possible to provide:
- `TextInputFormatter` (e.g. input masks like `dd/MM/yyyy`)
- A custom `TextEditingController`
- Any hook or builder to customize the internal `TextField`

This makes it impossible to enforce locale-specific date formats or guide users while typing.

---

## Why this is an issue

- Many locales rely on strict date formats (e.g. `dd/MM/yyyy` in most of the world)
- Free text input easily leads to invalid or ambiguous dates
- Developers are forced to disable manual input entirely (`calendarOnly`) or reimplement a custom date picker
- This breaks consistency with `TextField`, which already supports formatters and controllers

This is not a bug, but a limitation in the current API that affects extensibility and usability.

---

## Current workaround

- Disable manual input using `DatePickerEntryMode.calendarOnly`, or
- Build a fully custom date picker, losing native Material UX and accessibility benefits

---

## Flutter version

```text
Flutter 3.38.7 • channel stable
Tools • Dart 3.10.7 • DevTools 2.51.1
```

## Image of the Issue

The screenshot below shows the manual date input mode in `showDatePicker`, where the text field does not apply any input mask or formatting, allowing free-form input that can lead to invalid or ambiguous dates.

Image

### Proposal

Provide a way to add/customize the manual input field used by `showDatePicker`, for example:
- Allow passing custom `TextInputFormatter` instances.
- Allow providing a `TextEditingController`.
- Or expose a builder or callback to customize the internal input field.

Optionally, the framework could offer a locale-aware date input formatter that developers may choose to use.

This would allow developers to keep the native Material date picker while enforcing proper input formats.

Mairramer:
This PR adds the `keyboardInputFormatters` method to `CalendarDelegate`.

Previously, date input was strictly coupled to `MaterialLocalizations`. This change allows developers to provide a list of `TextInputFormatter`s via a custom `CalendarDelegate` to enforce specific input masks (e.g., dd/MM/yyyy) or restrict character entry in date pickers (`showDatePicker`, `showDateRangePicker`, and `InputDatePickerFormField`).

Fixes flutter/flutter#83302
Fixes flutter/flutter#97134 
Fixes flutter/flutter#62401
Fixes flutter/flutter#181043
Fixes flutter/flutter#115151
Fixes partially flutter/flutter#150611

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/date.dart
```diff
@@ -243,6 +248,19 @@ class GregorianCalendarDelegate extends CalendarDelegate<DateTime> {
   }
 }
 
+/// A [GregorianCalendarDelegate] that synchronizes calendar operations with
+/// text input requirements, such as [inputFormatters].
+///
+/// Subclasses must provide the appropriate [inputFormatters] to ensure
+/// that user entry matches the parsing logic in [parseCompactDate].
+abstract class DateInputCalendarDelegate extends GregorianCalendarDelegate {
```
juampiq6: should this be DateInputGregorianCalendarDelegate?

dkwingsmt (commented):
Sorry for the late feedback.

I have questions about the API design. I think it would be a better idea so that the `inputFormatters` be a property of `DatePicker` instead of the delegate. A couple of reasons:
* Runtime type check like this signals bad class design, which I'd prefer avoiding if possible.
* Semantically, non-gregorian calendars will also want to format their text.
* It also feels unnecessary that people have to extend the entire delegate to define a new formatter. The two aspects feel independent.

Otherwise, great code quality, and fantastic example app. Thank you!

packages/flutter/lib/src/material/date.dart
```diff
@@ -477,3 +480,44 @@ class DateTimeRange<T extends DateTime> {
...
+/// for date text fields within a date picker.
+///
+/// This delegate decouples the text input logic from the calendar system,
+/// allowing custom date formats, input masks (via [inputFormatters]), and
+/// parsing rules independent of the [CalendarDelegate].
+///
+/// See also:
+///
+///  * [CalendarDelegate], which handles the underlying calendar math and navigation.
+abstract class DateInputDelegate {
```
dkwingsmt: I propose that this class should be `abstract base class`. Adding `base` forces apps to `extend` this class instead of using `implements`, which makes the API more future-proof. When we add more methods to this class in the future, apps using `implements` will always be broken, while apps using `extend` can remain valid as long as we provide a default method body.

dkwingsmt (commented):
I like the current API in terms of design, but it might not have provided enough guidance to developers. If people just want to customize input formatter, which I wonder if will be the more common case, it'll be hard for them to figure out what to fill in for the other methods to keep the current behavior. I suggest either providing a default behavior in the document of these methods, or it might be a good idea to write a `StandardGregorianDateInputDelegate` for people to extend. (I'm not an expert on `DatePicker` and localization so feel free to correct me.)

Otherwise looking good to me!

packages/flutter/lib/src/material/date_picker.dart
```diff
@@ -3446,10 +3492,16 @@ class _InputDateRangePickerState extends State<_InputDateRangePicker> {
...
     final InputDecorationThemeData inputTheme = InputDecorationTheme.of(context);
+    final DateInputDelegate? config = widget.dateInputDelegate;
     final InputBorder inputBorder =
         inputTheme.border ??
         (useMaterial3 ? const OutlineInputBorder() : const UnderlineInputBorder());
 
+    final List<TextInputFormatter> effectiveFormatters =
+        config?.inputFormatters ?? <TextInputFormatter>[];
+    final String effectiveHelpText =
+        config?.helpText(localizations) ?? widget.calendarDelegate.dateHelpText(localizations);
```
dkwingsmt: Since we already have `config.helpText`, is this design overlapping by defining another `calendarDelegate.dateHelpText`?

packages/flutter/lib/src/material/input_date_picker_form_field.dart
```diff
@@ -195,7 +201,10 @@ class _InputDatePickerFormFieldState extends State<InputDatePickerFormField> {
   void _updateValueForSelectedDate() {
     if (_selectedDate != null) {
       final MaterialLocalizations localizations = MaterialLocalizations.of(context);
-      _inputText = widget.calendarDelegate.formatCompactDate(_selectedDate!, localizations);
+      final DateInputDelegate? dateInputDelegate = widget.dateInputDelegate;
+      _inputText = dateInputDelegate != null
+          ? dateInputDelegate.format(_selectedDate!, localizations)
+          : widget.calendarDelegate.formatCompactDate(_selectedDate!, localizations);
```
dkwingsmt: Likewise: why the overlapping API here?

dkwingsmt (commented):
It seems like most of the methods are overlapping with `CalendarDelegate`. I find it confusing, especially since I don't think we're deprecating `CalendarDelegate` at all.

packages/flutter/lib/src/material/date_picker.dart
Mairramer: This happens because everything depends on the `CalendarDelegate` for formatting. Simply adding inputFormatters is not enough.

That’s why I introduced a `DateInputDelegate`, focusing only on the three methods needed to properly format, parse, and validate dates.

I’ll check if I can abstract this further.

packages/flutter/lib/src/material/date.dart
```diff
@@ -477,3 +480,44 @@ class DateTimeRange<T extends DateTime> {
...
+/// parsing rules independent of the [CalendarDelegate].
+///
+/// See also:
+///
+///  * [CalendarDelegate], which handles the underlying calendar math and navigation.
+abstract base class DateInputDelegate {
+  /// Abstract const constructor. This constructor enables subclasses to provide
+  /// const constructors so that they can be used in const expressions.
+  const DateInputDelegate();
+
```
nate-thegrate: This class doesn't implement any fields, so the `interface` modifier might make more sense here.

```suggestion
abstract interface class DateInputDelegate {
```

nate-thegrate: If we end up going with this change over #184283, I agree with @dkwingsmt that the DateInputDelegate class should come with some sort of implementation.

Possibly:

```dart
abstract interface class DateInputDelegate {
  const factory DateInputDelegate.standardGregorian() = _StandardGregorianDelegate;

  // ...
}

class _StandardGregorianDelegate implements DateInputDelegate { ... }
```

Mairramer: I’d prefer to remove this. Since all formatting depends on the `CalendarDelegate`, it’s clearer to just override the methods and document that they must be overridden if the user wants behavior different from the default.

The example makes this more explicit.

examples/api/lib/material/date_picker/show_date_picker.2.dart
```diff
@@ -0,0 +1,292 @@
...
+
+  @override
+  State<DatePickerSample> createState() => _DatePickerSampleState();
+}
+
+class _DatePickerSampleState extends State<DatePickerSample> {
+  DateTime? _selectedDate;
+
+  DateInputFormat _formatType = DateInputFormat.dayMonthYear;
+  DateSeparator _separator = DateSeparator.dot;
```
nate-thegrate: Dart's new [dot shorthands](https://dart.dev/language/dot-shorthands) could be fun to use here!

```suggestion
  DateInputFormat _formatType = .dayMonthYear;
  DateSeparator _separator = .dot;
```

examples/api/lib/material/date_picker/show_date_picker.2.dart
```diff
@@ -0,0 +1,292 @@
...
+              initialValue: _formatType,
+              decoration: const InputDecoration(labelText: 'Date format'),
+              items: DateInputFormat.values
+                  .map(
+                    (DateInputFormat format) => DropdownMenuItem(
+                      value: format,
+                      child: Text(format.patternLabel),
+                    ),
+                  )
+                  .toList(),
```
nate-thegrate: We might be able to improve readability here by pulling the `children` out into a variable and using [for elements](https://dart.dev/language/collections#for-element) in place of `values.map(...).toList()`:

```dart
    final content = <Widget>[
      Text(label),
      const SizedBox(height: 24),
      DropdownButtonFormField<DateInputFormat>(
        initialValue: _formatType,
        decoration: const InputDecoration(labelText: 'Date format'),
        items: [
          for (final format in DateInputFormat.values)
            DropdownMenuItem(value: format, child: Text(format.patternLabel)),
        ],
        onChanged: (DateInputFormat? value) {
          if (value != null) {
            setState(() => _formatType = value);
          }
        },
      ),
      const SizedBox(height: 16),
      DropdownButtonFormField<DateSeparator>(
        initialValue: _separator,
        decoration: const InputDecoration(labelText: 'Separator'),
        items: [
          for (final separator in DateSeparator.values)
            DropdownMenuItem(value: separator, child: Text(separator.value)),
        ],
        onChanged: (DateSeparator? value) {
          if (value != null) {
            setState(() => _separator = value);
          }
        },
      ),
      const SizedBox(height: 24),
      OutlinedButton(onPressed: _showPicker, child: const Text('Select date')),
    ];
```

packages/flutter/test/material/date_picker_test.dart
```diff
@@ -2765,6 +2766,118 @@ void main() {
...
+
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: Builder(
+              builder: (BuildContext context) {
+                return ElevatedButton(
+                  child: const Text('Open'),
+                  onPressed: () async {
+                    result = await showDatePicker(
```
nate-thegrate: Looks like the convenient `buildApp()` function wasn't called here since the `onPressed` callback is setting the `result` value.

We probably could use it, though, with a bit of restructuring:

```dart
    Future<DateTime?> showPicker() {
      return showDatePicker(
        context: find.text('Open').evaluate().first,
        initialDate: initialDate,
        firstDate: firstDate,
        lastDate: lastDate,
        initialEntryMode: DatePickerEntryMode.input,
        calendarDelegate: const _TestDateInputDelegate(),
        inputFormatters: inputFormatters,
      );
    }

    Widget buildApp({VoidCallback? onPressed}) {
      final button = ElevatedButton(
        onPressed: onPressed ?? showPicker,
        child: const Text('Open'),
      );
      return MaterialApp(home: Material(child: button));
    }

    // ...

    Future<void> buttonPress() async {
      result = await showPicker();
    }

    await tester.pumpWidget(buildApp(onPressed: buttonPress));
```

I think this would be more readable, but if you disagree that's fine 🙂

packages/flutter/test/material/date_picker_test.dart
```diff
@@ -2765,6 +2766,118 @@ void main() {
...
+            },
+          ),
+        ),
+      );
+    }
+
+    testWidgets('applies inputFormatters and allows text entry parsing', (
+      WidgetTester tester,
+    ) async {
+      late DateTime? result;
```
nate-thegrate: Nullable fields are initialized to `null` by default, so we can get rid of `late` here.

```suggestion
      DateTime? result;
```

nate-thegrate (commented):
This looks really good!

<br>

Quick heads-up: I currently don't have commit access in this repo so I won't be able to give an approval.

If you're interested in making progress more quickly, I have a suggestion—if you have a chance, would you be able to look through #184283? It'd be super helpful if you could describe what's different between this PR and that one, so we can decide which PR to move forward with.

examples/api/lib/material/date_picker/show_date_picker.2.dart
```diff
@@ -0,0 +1,282 @@
...
+    );
+
+    if (digits.length > maxLength) {
+      return oldValue;
+    }
+
+    final String formatted = _applyMask(digits);
+    return TextEditingValue(
+      text: formatted,
+      selection: TextSelection.collapsed(offset: formatted.length),
```
dkwingsmt: This `offset` will cause the text cursor to always be moved to the end of the text, which isn't expected if the user is editing at the middle of the string, which is reproducible in my local testing.

The following video recording includes two segments:
* The first half shows pressing Delete in the middle of a string that's not immediately after a separator.
* The second half shows pressing Delete in the middle of a string that's immediately after a separator.

https://github.com/user-attachments/assets/b3f799df-7499-4aa3-a547-34fc7437b42f

packages/flutter/lib/src/material/input_date_picker_form_field.dart
```diff
@@ -150,6 +152,31 @@ class InputDatePickerFormField extends StatefulWidget {
...
+  /// Optional list of [TextInputFormatter]s applied to the date input field.
+  ///
+  /// These formatters control and restrict how the user enters text, such as
+  /// enforcing a specific date mask (e.g., `MM/dd/yyyy`), limiting allowed
+  /// characters, or automatically inserting date separators.
+  ///
+  /// For date input and validation to function correctly, these formatters
+  /// must be consistent with the [CalendarDelegate.dateHelpText],
+  /// [CalendarDelegate.formatCompactDate], and [CalendarDelegate.parseCompactDate]
+  /// methods. While [CalendarDelegate.dateHelpText] provides the hint text
```
dkwingsmt: The first sentence contains a "must", which is discouraged in the style guide for being prescriptive. I consulted AI, who suggested the following rephrase in a more descriptive way:
```
/// The functionality of date input and validation relies on consistency between these formatters and the [CalendarDelegate] methods: [CalendarDelegate.dateHelpText],
/// [CalendarDelegate.formatCompactDate], and [CalendarDelegate.parseCompactDate].
```

dkwingsmt (commented):
The implementation looks good to me, but I found two problems while local testing the example (I understand that the logic here is really delicate and complicated). 

Overall really good quality. Thank you very much for your contribution!

dkwingsmt (dismissed):
LGTM

victorsanni:
Thanks for your contribution! This PR has reached completion, however we're not able to merge this PR at the moment due to the Material/Cupertino code freeze (https://github.com/flutter/flutter/issues/184093). As soon as the decoupling is complete, we will post a way to port this PR to the new packages to land. In the meantime, I'll convert this PR to a draft and apply a label so that we can quickly find it when the freeze is lifted.

packages/flutter/lib/src/material/date.dart
```diff
@@ -10,6 +10,8 @@ library;
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 
+import 'date_picker.dart';
+import 'input_date_picker_form_field.dart';
```
victorsanni: Do we need these imports? If they are only for documentation, they should be docImports instead.

Mairramer: Done

Piinks:

Piinks:

This PR can now be ported over to flutter/packages!

Mairramer:
@dkwingsmt Ported to https://github.com/flutter/packages/pull/12608. It will stay in draft due to the 2 PR limit.

PR #181110, opened 2026-01-17
refactor: Centralize copying of drag events when handling scrollbar scrolling

luanpotter:
This PR changes the signatures of `handleThumbPressStart`, `handleThumbPressUpdate`, and `handleThumbPressEnd`, related to scrollbar handling, to take in a copy of the original event with the `localPosition` added inside, instead of separate fields.

This centralizes the copying of these events by the generic platform code (while still allowing some of them to re-copy based on implementation details), ensures that additional fields are propagated (such as `kind` when it exists, `buttons` when it is eventually added, and others) and simplifies the method signature significantly, future proofing it to future additions.

This was extracted from https://github.com/flutter/flutter/pull/176968 in an attempt to simplify it, and it will make it possible to add the missing `buttons` property to these events more easily.

This also adds a `copyWith` method to these events in an attempt to ensure that the copies have all non-specified fields propagated by centralizing the "exhaustiveness" of the copy into the event class itself; that way if a new field is added in the future it can be trivially added to the `copyWith` method and not require figuring out all "copiers" out there (specially if it has a default value which could go un-noticed elsewhere).

Note: this is a public signature change but only if people are creating their own custom scrollbar implementations (?) without using either the material or cupertino ones; and on top of that should be very trivial to fix. But please advise if there are concerns.

Note: this has no test changes because it is a refactor which should have zero change in behaviour; the existing tests passing is proof of it working. So I am disregarding this bot, but please advise if I am mistaken.

Note: this PR assumes that the `renderBox.localToGlobal(localPosition)` computation that was previously done was essentially the _reverse_ of the global -> local computation, so essentially the flow was `global --(global to local)--> local --(local to global)--> global` (the second step being there just because we didn't provide the original global to the method). So this is saving a computation step and should not change the behaviour. I will let the tests be the judge of this assumption but please let me know if there are concerns.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/cupertino/scrollbar.dart
+++ b/packages/flutter/lib/src/cupertino/scrollbar.dart
@@ -177,15 +177,15 @@ class _CupertinoScrollbarState extends RawScrollbarState<CupertinoScrollbar> {
   // on the scrollbar thumb and then drags the scrollbar without releasing.
 
   @override
-  void handleThumbPressStart(Offset localPosition) {
-    super.handleThumbPressStart(localPosition);
+  void handleThumbPressStart(DragStartDetails details) {
+    super.handleThumbPressStart(details);
     final Axis? direction = getScrollbarDirection();
     if (direction == null) {
       return;
     }
     _pressStartAxisPosition = switch (direction) {
-      Axis.vertical => localPosition.dy,
-      Axis.horizontal => localPosition.dx,
+      Axis.vertical => details.localPosition.dy,
+      Axis.horizontal => details.localPosition.dx,
     };
   }
 
@@ -199,16 +199,16 @@ class _CupertinoScrollbarState extends RawScrollbarState<CupertinoScrollbar> {
   }
 
   @override
-  void handleThumbPressEnd(Offset localPosition, Velocity velocity) {
+  void handleThumbPressEnd(DragEndDetails details) {
     final Axis? direction = getScrollbarDirection();
     if (direction == null) {
       return;
     }
     _thicknessAnimationController.reverse();
-    super.handleThumbPressEnd(localPosition, velocity);
+    super.handleThumbPressEnd(details);
     final (double axisPosition, double axisVelocity) = switch (direction) {
-      Axis.horizontal => (localPosition.dx, velocity.pixelsPerSecond.dx),
-      Axis.vertical => (localPosition.dy, velocity.pixelsPerSecond.dy),
+      Axis.horizontal => (details.localPosition.dx, details.velocity.pixelsPerSecond.dx),
+      Axis.vertical => (details.localPosition.dy, details.velocity.pixelsPerSecond.dy),
     };
     if (axisPosition != _pressStartAxisPosition && axisVelocity.abs() < 10) {
       HapticFeedback.mediumImpact();
--- a/packages/flutter/lib/src/gestures/drag_details.dart
+++ b/packages/flutter/lib/src/gestures/drag_details.dart
@@ -93,6 +93,22 @@ class DragStartDetails with Diagnosticable implements PositionedGestureDetails {
     properties.add(DiagnosticsProperty<Duration?>('sourceTimeStamp', sourceTimeStamp));
     properties.add(EnumProperty<PointerDeviceKind?>('kind', kind));
   }
+
+  /// Creates a copy of this event but replacing the provided fields, while
+  /// maintaining the others.
+  DragStartDetails copyWith({
+    Offset? globalPosition,
+    Offset? localPosition,
+    Duration? sourceTimeStamp,
+    PointerDeviceKind? kind,
+  }) {
+    return DragStartDetails(
+      globalPosition: globalPosition ?? this.globalPosition,
+      localPosition: localPosition ?? this.localPosition,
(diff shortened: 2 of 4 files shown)


packages/flutter/lib/src/widgets/scrollbar.dart
```diff
@@ -1780,22 +1777,26 @@ class RawScrollbarState<T extends RawScrollbar> extends State<T> with TickerProv
...
-      delta: delta,
+      delta: delta, // adds the computed delta
       primaryDelta: primaryDelta,
-      globalPosition: renderBox.localToGlobal(localPosition),
-      localPosition: localPosition,
+      globalPosition: renderBox.localToGlobal(details.localPosition),
+      localPosition: details.localPosition,
+      sourceTimeStamp: details.sourceTimeStamp,
+      kind: details.kind,
     );
```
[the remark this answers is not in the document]
luanpotter: oops

victorsanni (commented):
I agree with this change, but it would be nice if we don't break developers who create their own classes that extend RawScrollbar (e.g design language developers making custom scrollbars).

Is this something a [data-driven fix](https://github.com/flutter/flutter/wiki/Data-driven-Fixes/c971f7c97fb18c7375acc2955c355bbe1238866c) might be able to help with?

packages/flutter/lib/src/cupertino/scrollbar.dart
```diff
@@ -177,15 +177,15 @@ class _CupertinoScrollbarState extends RawScrollbarState<CupertinoScrollbar> {
   // on the scrollbar thumb and then drags the scrollbar without releasing.
 
   @override
-  void handleThumbPressStart(Offset localPosition) {
```
Renzo-Olivares, line 180: Another option here would be to deprecate `handleThumbPressStart` and add a `handleThumbPressStartWithDetails` or similar. That would give users some time to migrate off the old API.

packages/flutter/lib/src/gestures/drag_details.dart
```diff
@@ -93,6 +93,22 @@ class DragStartDetails with Diagnosticable implements PositionedGestureDetails {
     properties.add(DiagnosticsProperty<Duration?>('sourceTimeStamp', sourceTimeStamp));
     properties.add(EnumProperty<PointerDeviceKind?>('kind', kind));
   }
+
+  /// Creates a copy of this event but replacing the provided fields, while
+  /// maintaining the others.
+  DragStartDetails copyWith({
```
Renzo-Olivares, line 99: Is a `copyWith` necessary here? The only reason I ask is that `sourceTimeStamp` is nullable and when null this means that it may have come proxied events. However, `copyWith` here cannot return `sourceTimeStamp` back to `null`. This is potentially error-prone since a user might use `details.copyWith(sourceTimeStamp: null)` expecting to reset `sourceTimeStamp` but this will not be the case.

Renzo-Olivares (commented):
Hi @luanpotter, thank you for the contribution! I left a few comments.

packages/flutter/lib/src/gestures/drag_details.dart
luanpotter: it is definitely not needed - my goal here was to "centralize" in one place everything that needs to be copied.
imagine in the future someone adds yet another new field to the events - they wouldn't know to go back to the scrollbar code and add a `foo = foo` line when creating the copy. the attempt here was to keep this code closer to the field list.

that being said the sourceTimeStamp does pose a problem if someone wants to "clear it". however I don't think flutter can make distinction here between undefined and null (probably for the best). are there other parts of the SDK were similar "copy" methods are used / this issue is solved?

I am ok with removing it entirely if you think my rationale for it is not worth the effort.

packages/flutter/lib/src/cupertino/scrollbar.dart
luanpotter: I can certainly give that idea a try - would that be instead of the "data-driven fix" ?

dkwingsmt: I agree that it should be a new method. I didn't realize that it was a public method via `RawScrollbarState`. 

I don't think we will need a data-driven fix (of course, good to have one if possible!). We can provide both methods at the same time and let `_handleThumbDragStart` call both. Just document clearly that only one should be used.

packages/flutter/lib/src/gestures/drag_details.dart
Renzo-Olivares: Yeah dart unfortunately does not provide a good solution to this problem. I think if a nullable parameter on `copyWith` has an equivalent non-null representation for "null" then a `copyWith` can be alright in that case, or if there are no nullable members at all. I understand your rationale and also like the convenience of `copyWith` though if the `copyWith` usage will be minimal i'd lean towards not including it. An `apply` method that requires every parameter has been done in the framework as another option but it is not too far off from just using the constructor to recreate the entire object manually.

luanpotter:
@dkwingsmt no worries at all, this ends up being messier than I wished!

I have re-created the PR here addressing @Renzo-Olivares 's insight about the copyWith footgun:
https://github.com/flutter/flutter/pull/186677
The PR is much simpler now, no centralized copyWith but still unblocking the `buttons` addition on followups.

PR #181151, opened 2026-01-19
fix: boundary check for restored scroll offsets in ScrollPosition

Danilospano00:
# Fix boundary check for restored scroll offsets in ScrollPosition

When a `ScrollPosition` restores its offset from `PageStorage`, the restored value might be out of bounds if the scrollable content has changed dimensions (e.g., a list that was previously scrolled down becomes empty). With `ClampingScrollPhysics`, the position should be automatically corrected to valid bounds during the first layout, but this wasn't happening.

Previously, `ScrollPosition.applyContentDimensions` only performed boundary correction via `correctForNewDimensions` when `haveDimensions` was `true`. On the first layout frame after restoration, `haveDimensions` is still `false`, so no correction was applied. This resulted in the viewport remaining stuck at an invalid offset (e.g., 1000.0 on an empty list with `maxScrollExtent` of 0.0).

This PR adds an `else if (hasPixels)` branch that applies boundary conditions even when `haveDimensions` is `false`, ensuring that restored offsets are immediately clamped to valid bounds during the first layout.

Fixes: #[issue_number]

## Pre-launch Checklist

- [x] I signed the [CLA].


Piinks (commented):
Hi @Danilospano00 thanks for contributing! It looks like this is failing a number of tests and will need the cla signed, can you take a look?

PR #181232, opened 2026-01-21
Implemented SliverExpansionPanelList

zawhtetnaing10:
Implemented SliverExpansionPanelList which can be used inside CustomScrollView for better scrolling performance. SliverAnimatedList is used under the hood for the animation for expansion and collapsing.

Issue #106521

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
```diff
@@ -0,0 +1,585 @@
+import 'package:flutter/material.dart';
+
+const double _kPanelHeaderCollapsedHeight = kMinInteractiveDimension;
```
navaronbracke: These values are probably Material specific? Can we allow users to customize these?

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
```diff
@@ -0,0 +1,585 @@
...
+      } else {
+        _addItem(newItems, oldItems);
+      }
+      // Sync data
+      _list = ListModel(
+        listKey: _listKey,
+        removedItemBuilder: _buildRemovedItem,
+        initialItems: newItems,
+      );
+    }
```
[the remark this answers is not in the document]
zawhtetnaing10: The logic inside didUpdateWidget has been updated to handle multiple insertion, multiple removal and reordering of items.

Enderjua:
@dkwingsmt @Piinks

i apologize for mention, but i think the PR worth looking into.

packages/flutter/test/material/sliver_expansion_panel_list_test.dart
```diff
@@ -0,0 +1,73 @@
...
+          ),
+        ),
+      ),
+    );
+    await tester.pump(const Duration(milliseconds: 200));
+
+    expect(find.text('A'), findsNothing);
+    expect(find.text('B'), findsOneWidget);
+  });
+}
```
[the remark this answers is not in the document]
victorsanni: +1

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
```diff
@@ -0,0 +1,609 @@
...
+  }
+
+  Widget _buildRemovedItem(
+    _SliverExpansionPanelItem item,
+    BuildContext context,
+    Animation<double> animation,
+  ) {
+    return SizeTransition(
+      sizeFactor: animation,
+      child: widget.expansionPanels[item.panelIndex].body,
```
Piinks: Accessing widget.expansionPanels[item.panelIndex] might be risky because the widget list may have already changed. WDYT about capturing the Widget body in _SliverExpansionPanelItem or the _ListModel.removeAt call to ensure the 'ghost' widget remains valid during the animation?

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
```diff
@@ -0,0 +1,609 @@
...
+        _list.insert(i, movedItem);
+      } else {
+        // New item. Add the new item at the index.
+        // This includes expanding items programmatically.
+        _list.insert(i, expectedItem);
+      }
+    }
+
+    // Sync the items.
+    _list = _ListModel(
```
Piinks, line 446: Is the re-allocation of _list at the end of didUpdateWidget necessary? The loops above seem to already sync the _list._items using insert and removeAt. If we re-allocate, we might be doing redundant work.

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
```diff
@@ -0,0 +1,609 @@
...
+}
+
+/// A clipper used to prevent shadow bleeding at the top of a body.
+///
+/// When a [SliverExpansionPanel] is expanded, the body is elevated. Without
+/// clipping, the elevation shadow can "bleed" upwards onto the header above.
+///
+/// This clipper creates a bounding box that is strictly constrained at the top
+/// (0.0), while allowing the shadow to spread freely on the left, right, and
+/// bottom by providing a 10.0 pixel buffer. The shadow is
```
Piinks, line 260: Instead of hardcoding a 10px buffer, should this consider a buffer size relative to widget.elevation to ensure high-elevation shadows aren't clipped unexpectedly? Or maybe default to 10 and expose a way for the user to customize it?

packages/flutter/test/material/sliver_expansion_panel_list_test.dart
```diff
@@ -0,0 +1,77 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
Piinks, line 1: Since the the recent updates to didUpdateWidget logic handles complex reordering and multiple insertions, could we add a test case in that specifically verifies a reorder and a multi-item programmatic update?

Piinks (commented):
Thank you very much for the contribution! 
I am curious, should this be in the widgets library? Usually material widgets implement the material design specification. I see it's making use of ExpandIcon, Inkwell, Theme, and MaterialLocalizations (very thorough!), but I wonder if this is coming from the material specification? If we move this to widgets, I would consider something more generic and less stylized. No need to move now, curious for your thoughts. 🙂 🙏

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
zawhtetnaing10: I've captured the headerBuilder and body inside _SliverExpansionPanelItem  and returned it instead of using the panelIndex.

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
zawhtetnaing10: I've looked into removing the final re-allocate logic, but I found that the panelIndex property on existing items becomes stale when the list changes. Since the expansionCallback relies on panelIndex to identify the correct item, I think we must sync this property to prevent logic errors.

packages/flutter/lib/src/material/sliver_expansion_panel_list.dart
zawhtetnaing10: I've added an additional property called shadowMargin which defaults to 10px and is exposed for the user to customize it.

packages/flutter/test/material/sliver_expansion_panel_list_test.dart
zawhtetnaing10: I've added more test cases to verify re-order, insertion, deletion and programmatic expansion and collapsing of items.

Piinks:
Thank you for your contribution. I'm going to close this PR for now sine we need to discuss further in the issues as suggested above. Please don't hesitate to re-open or submit a new PR if you have the time to contribute. Thanks!

PR #181238, opened 2026-01-21
Add WidgetStatesController support to ExpansionTile

iamvikashtiwari:
This PR adds support for providing a `WidgetStatesController` to `ExpansionTile`
and forwards it to the backing `ListTile`. This allows callers to observe and
control hovered, pressed, and focused states of the tile header.

The change includes updated documentation with a snippet example and adds
test coverage for both provided and null `statesController` cases.

Fixes #180161

## Pre-launch Checklist

- [x] I read the Contributor Guide and followed the process outlined there for submitting PRs.
- [x] I read the Tree Hygiene wiki page, which explains my responsibilities.
- [x] I read and followed the Flutter Style Guide, including Features we expect every widget to implement.
- [x] I signed the CLA.
- [x] I added new tests to check the change I am making.
- [ ] I followed the breaking change policy and added Data Driven Fixes where supported.

### DIFF
--- a/packages/flutter/lib/src/material/expansion_tile.dart
+++ b/packages/flutter/lib/src/material/expansion_tile.dart
@@ -152,6 +152,7 @@ class ExpansionTile extends StatefulWidget {
     this.enabled = true,
     this.expansionAnimationStyle,
     this.internalAddSemanticForOnTap = false,
+    this.statesController,
   }) : assert(
          expandedCrossAxisAlignment != CrossAxisAlignment.baseline,
          'CrossAxisAlignment.baseline is not supported since the expanded children '
@@ -484,6 +485,40 @@ class ExpansionTile extends StatefulWidget {
   // the default value to true.
   final bool internalAddSemanticForOnTap;
 
+  /// {@tool snippet}
+  /// This example listens for hover and press states on an [ExpansionTile].
+  ///
+  /// ```dart
+  /// class ExpansionTileStatesExample extends StatelessWidget {
+  ///   ExpansionTileStatesExample({super.key});
+  ///
+  ///   final WidgetStatesController controller = WidgetStatesController();
+  ///
+  ///   @override
+  ///   Widget build(BuildContext context) {
+  ///     controller.addListener(() {
+  ///       if (controller.value.contains(WidgetState.hovered)) {
+  ///         debugPrint('Tile is hovered');
+  ///       }
+  ///     });
+  ///
+  ///     return MaterialApp(
+  ///       home: Scaffold(
+  ///         body: ExpansionTile(
+  ///           title: const Text('Settings'),
+  ///           statesController: controller,
+  ///         ),
+  ///       ),
+  ///     );
+  ///   }
+  /// }
+  /// ```
+  /// {@end-tool}
+  ///
+  /// If null, the backing [ListTile] will create and manage its own
+  /// [WidgetStatesController].
+  final WidgetStatesController? statesController;
+
   @override
   State<ExpansionTile> createState() => _ExpansionTileState();
 }
@@ -626,6 +661,7 @@ class _ExpansionTileState extends State<ExpansionTile> {
             : null,
         minTileHeight: widget.minTileHeight,
         internalAddSemanticForOnTap: widget.internalAddSemanticForOnTap,
+        statesController: widget.statesController,
       ),
     );
 
--- a/packages/flutter/lib/src/material/expansion_tile.dart
+++ b/packages/flutter/lib/src/material/expansion_tile.dart
@@ -485,6 +485,12 @@ class ExpansionTile extends StatefulWidget {
   // the default value to true.
   final bool internalAddSemanticForOnTap;
 
+  /// The controller that notifies when the widget's [WidgetState]s change.
+  ///
+  /// This allows listening to and controlling states such as
+  /// [WidgetState.hovered], [WidgetState.focused], [WidgetState.pressed],
+  /// and [WidgetState.disabled] for the tile's header.
+  ///
   /// {@tool snippet}
   /// This example listens for hover and press states on an [ExpansionTile].
   ///
--- a/packages/flutter/lib/src/material/expansion_tile.dart
+++ b/packages/flutter/lib/src/material/expansion_tile.dart
@@ -491,36 +491,6 @@ class ExpansionTile extends StatefulWidget {
   /// [WidgetState.hovered], [WidgetState.focused], [WidgetState.pressed],
   /// and [WidgetState.disabled] for the tile's header.
   ///
-  /// {@tool snippet}
-  /// This example listens for hover and press states on an [ExpansionTile].
-  ///
-  /// ```dart
-  /// class ExpansionTileStatesExample extends StatelessWidget {
-  ///   ExpansionTileStatesExample({super.key});
-  ///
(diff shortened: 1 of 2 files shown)


packages/flutter/test/material/expansion_tile_test.dart
```diff
@@ -2073,4 +2073,36 @@ void main() {
       variant: const TargetPlatformVariant(<TargetPlatform>{TargetPlatform.android}),
     );
   });
+
+  testWidgets('ExpansionTile forwards statesController to ListTile', (tester) async {
```
QuncCccccc, line 2162: Run `dart format packages/flutter/test/material/expansion_tile_test.dart` to fix linux analyzer.

QuncCccccc (commented):
We might need to fix the failed tests and also the comments from the closed PR:) Thanks!

packages/flutter/test/material/expansion_tile_test.dart
iamvikashtiwari: Thanks for pointing this out!

I’ve fixed the formatting issue in
`packages/flutter/test/material/expansion_tile_test.dart`
by running `dart format` and pushed the update.

packages/flutter/test/material/expansion_tile_test.dart
```diff
@@ -2073,4 +2073,30 @@ void main() {
...
+    final controller = WidgetStatesController();
```
QuncCccccc, line 2163: We need to dispose controller when we finish testing. Adding `addTearDown(controller.dispose);` after this line should fix the failed "Windows framework_tests_libraries_leak_tracking" test.

iamvikashtiwari: Thanks for the suggestion!

Added `addTearDown(controller.dispose);` to properly dispose the
`WidgetStatesController` in the test. This should resolve the
Windows leak tracking failure.

packages/flutter/test/material/expansion_tile_test.dart
```diff
@@ -2073,4 +2073,38 @@ void main() {
...
+          child: ExpansionTile(title: const Text('Tile'), statesController: controller),
+        ),
+      ),
+    );
+
+    final ListTile listTile = tester.widget<ListTile>(find.byType(ListTile));
+    expect(listTile.statesController, controller);
+  });
+
+  testWidgets('ExpansionTile forwards statesController to ListTile', (tester) async {
```
QuncCccccc: We have 2 identical "ExpansionTile forwards statesController to ListTile" tests now. I think originally this is "ExpansionTile forwards null statesController to ListTile".

packages/flutter/test/material/expansion_tile_test.dart
```diff
@@ -2073,4 +2073,38 @@ void main() {
       variant: const TargetPlatformVariant(<TargetPlatform>{TargetPlatform.android}),
     );
   });
+
+  testWidgets('ExpansionTile forwards statesController to ListTile', (tester) async {
+    final controller = WidgetStatesController();
```
QuncCccccc, line 2163:
```suggestion
    final controller = WidgetStatesController();
    addTearDown(controller.dispose);
```

packages/flutter/lib/src/material/expansion_tile.dart
```diff
@@ -484,6 +485,40 @@ class ExpansionTile extends StatefulWidget {
...
+  ///     return MaterialApp(
+  ///       home: Scaffold(
+  ///         body: ExpansionTile(
+  ///           title: const Text('Settings'),
+  ///           statesController: controller,
+  ///         ),
+  ///       ),
+  ///     );
+  ///   }
+  /// }
```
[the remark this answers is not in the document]
loic-sharma: @iamvikashtiwari This looks like a real problem that should be addressed

packages/flutter/lib/src/material/expansion_tile.dart
```diff
@@ -484,6 +485,40 @@ class ExpansionTile extends StatefulWidget {
   // the default value to true.
   final bool internalAddSemanticForOnTap;
 
+  /// {@tool snippet}
```
loic-sharma: We should add a summary sentence so that IDEs like VS Code show useful code completion information if you type `myExpansionTile.statesController`. Consider something like:

```suggestion
  /// The controller that notifies when the widget's [WidgetState]s change.
  ///
  /// {@tool snippet}
```

packages/flutter/test/material/expansion_tile_test.dart
```diff
@@ -2073,4 +2073,31 @@ void main() {
...
+  testWidgets('ExpansionTile forwards null statesController to ListTile', (tester) async {
+    await tester.pumpWidget(
+      const MaterialApp(
+        home: Material(child: ExpansionTile(title: Text('Tile'))),
+      ),
+    );
+
+    final ListTile listTile = tester.widget<ListTile>(find.byType(ListTile));
+    expect(listTile.statesController, isNull);
+  });
```
loic-sharma, line 2187: Should we add a test that interacts with the `ListTile` and then checks the `listTile.statesController.value` is correct? For example, check that changing `ListTile.enabled` from `true` to `false` adds `WidgetState.disabled` to `listTile.statesController.value`.

packages/flutter/lib/src/material/expansion_tile.dart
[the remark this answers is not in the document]
QuncCccccc: Oh I missed this. For code snippet, maybe we should make it shorter? I didn't see a whole class example in code snippet previously. Maybe check the other use cases as examples.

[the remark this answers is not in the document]
loic-sharma: It looks like other `WidgetStatesController` fields don't provide a code example. However, it seems useful to have at least _one_ example of this pattern.

I'd consider:

1. Move this code example to be a full sample: `examples/api/lib/widgets/widget_state/widget_states_controller.0.dart`
2. Add unit tests for this sample
3. Add this code example to `WidgetStatesController`'s docs directly using a `{@tool dartpad}` (instead of a `{@tool snippet}`).

packages/flutter/test/material/expansion_tile_test.dart
iamvikashtiwari: Thanks for the suggestion!

For this PR, I intentionally limited the tests to verifying that ExpansionTile correctly forwards the statesController to its backing ListTile (including the null case), since that’s the new API contract being introduced here.

The behavior of how ListTile updates WidgetStatesController.value in response to state changes (e.g. enabled, hover, press) is already covered by existing ListTile tests, so I avoided duplicating that logic here.

Happy to add an interaction-based test if you feel it’s necessary for this change — please let me know 👍

loic-sharma: I chatted with @QuncCccccc and she confirmed this is OK

packages/flutter/lib/src/material/expansion_tile.dart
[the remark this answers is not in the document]
QuncCccccc: Gentle ping just in case this comment is missed:)!

QuncCccccc (approved):
LGTM. Thanks!

justinmc (approved):
LGTM 👍 . Always good to have more WidgetStatesController support.

PR #181240, opened 2026-01-21
Expose computeLineMetrics on RenderParagraph

### LINKED ISSUE #44834: Expose TextPainter.computeLineMetrics on RenderParagraph
Hi there,

I have a need to access the line metrics of a rendered `Text` widget, and while the information I need is available on `TextPainter`, `RenderParagraph` doesn't expose it. Can we get a `computeLineMetrics` method on RenderParagraph?

Thanks!

cc @GaryQian

crackedhandle:
This change exposes TextPainter.computeLineMetrics() through RenderParagraph by adding a small forwarding method.

This allows callers that only have access to RenderParagraph to retrieve line metrics without duplicating layout logic or accessing internal state.

The implementation follows the same pattern as existing query methods such as getBoxesForSelection and getPositionForOffset.
 
fixes #44834


packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -1135,6 +1135,17 @@ class RenderParagraph extends RenderBox
...
+  ///
+  /// This exposes [TextPainter.computeLineMetrics].
+  ///
+  /// Valid only after [layout].
+  List<ui.LineMetrics> computeLineMetrics() {
+    assert(!debugNeedsLayout);
+    _layoutTextWithConstraints(constraints);
+    return _textPainter.computeLineMetrics();
+  }
+
```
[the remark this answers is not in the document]
crackedhandle: Fixed — removed the extra blank line. Thanks!

crackedhandle:
> @crackedhandle This will need a test if we do decide to expose this.

Thanks for the review! That makes sense. I’ll add a test covering the new computeLineMetrics forwarding on RenderParagraph and update the PR shortly.

crackedhandle:
@Renzo-Olivares  can you review this now?
@justinmc  i posted the test

crackedhandle:
> Hi @crackedhandle, thank you for your contribution! What is your use-case for exposing this? We don't currently have a mechanism in the framework to listen to text layout changes so it would be difficult to monitor and update any saved line metrics.

Hi @Renzo-Olivares, thank you for the thoughtful question.

The primary use case for exposing `computeLineMetrics` on `RenderParagraph` is to support advanced text layout scenarios where developers need precise per-line information directly from the render object layer — for example, implementing custom text selection behavior, drawing line-based decorations, aligning widgets to specific text lines, or building rich text editing/annotation tools.

While similar information can be accessed via `TextPainter`, that requires re-creating layout logic outside the render tree. Exposing this on `RenderParagraph` allows access to the already-laid-out paragraph, avoiding duplication and ensuring consistency with the actual rendered layout.

Regarding layout changes, I agree that line metrics can become stale if cached. The expectation would be that consumers query metrics during paint/layout phases rather than persist them across frames, similar to other layout-dependent properties.

Happy to refine the API further if needed.

packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -1136,6 +1136,17 @@ class RenderParagraph extends RenderBox
     return _textPainter.size;
   }
 
+  /// Returns the line metrics for the laid out text.
+  ///
+  /// This exposes [TextPainter.computeLineMetrics].
+  ///
+  /// Valid only after [layout].
```
Renzo-Olivares: nit: for the docs we should probably use the same documentation from `TextPainter.computeLineMetrics`. I recommend making it a doc template and using that template here.

packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -1136,6 +1136,17 @@ class RenderParagraph extends RenderBox
...
+  List<ui.LineMetrics> computeLineMetrics() {
+    assert(!debugNeedsLayout);
+    _layoutTextWithConstraints(constraints);
+    return _textPainter.computeLineMetrics();
```
Renzo-Olivares, line 1178: I wonder if it would be beneficial here to expose other methods like `getLineMetricsAt` which may have less overhead then calling `computeLineMetrics` if someone only wants to access metrics for specific lines. cc @LongCatIsLooong since that seems to be the approach you were taking in https://github.com/flutter/flutter/pull/145190/changes#diff-8869f4e2a7291974a4badaa8af9df4e31da61cba8a0beb83948a258486166fdc .

packages/flutter/lib/src/rendering/paragraph.dart
crackedhandle: > nit: for the docs we should probably use the same documentation from `TextPainter.computeLineMetrics`. I recommend making it a doc template and using that template here.

Thanks for pointing that out!

I’ll update the docs to reuse the same template as `TextPainter.computeLineMetrics` so we avoid duplication and keep everything consistent.

packages/flutter/lib/src/rendering/paragraph.dart
crackedhandle: > I wonder if it would be beneficial here to expose other methods like `getLineMetricsAt` which may have less overhead then calling `computeLineMetrics` if someone only wants to access metrics for specific lines. cc @LongCatIsLooong since that seems to be the approach you were taking in https://github.com/flutter/flutter/pull/145190/changes#diff-8869f4e2a7291974a4badaa8af9df4e31da61cba8a0beb83948a258486166fdc .

That’s a really good suggestion.

For this PR, I was aiming to keep the change minimal and just expose the existing `computeLineMetrics` through `RenderParagraph`.

I’m happy to open a follow-up PR to explore exposing `getLineMetricsAt` (or other more granular APIs) if that would be useful.

examples/api/lib/painting/line_metrics_example.dart
```diff
@@ -0,0 +1,133 @@
...
+    super.didUpdateWidget(oldWidget);
+
+    if (oldWidget.text != widget.text ||
+        oldWidget.style != widget.style ||
+        oldWidget.maxWidth != widget.maxWidth) {
+      _compute();
+    }
+  }
+
+  void _compute() {
```
LongCatIsLooong: This doesn't seem to be using `RenderParagraph` or the API you exposed.

Another problem with this example is that I don't think this recomputes the line metrics (but it should) when available fonts change. That kinda speaks volumes for how difficult it will be to make sure `computeLineMetrics` is called when there's a text layout change.

crackedhandle: @LongCatIsLooong please have a look on updated example

examples/api/lib/painting/line_metrics_example.dart
```diff
@@ -0,0 +1,97 @@
...
+  RenderObject createRenderObject(BuildContext context) {
+    return RenderLineMetricsDemo();
+  }
+}
+
+class RenderLineMetricsDemo extends RenderBox {
+  late RenderParagraph _paragraph;
+
+  RenderLineMetricsDemo() {
+  _paragraph = RenderParagraph(
```
Renzo-Olivares: Instead of managing a `RenderParagraph` like this we should either extend `RenderParagraph` directly or extend `RenderProxyBox` and create our own widget around it that wraps a `Text` widget and intercepts paint.

Renzo-Olivares: The `RenderProxyBox` approach only works if there are no repaint boundaries between the `RenderParagraph` and the proxy.

crackedhandle: Thanks for the review, @Renzo-Olivares! 🙏

I've refactored the implementation based on your suggestion:

packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -1164,6 +1164,20 @@ class RenderParagraph extends RenderBox
     return _textPainter.size;
   }
 
+  /// Returns a list of [ui.LineMetrics] objects describing the
```
Renzo-Olivares: I recommend making the documentation in `TextPainter.computeLineMetrics` a doc template and using that as a macro here in `RenderParagraph`. An example of this is `TextPainter.strutStyle` which puts its documentation in a template that can be reused.

Renzo-Olivares: We should also link to the example you created in the dart docs see the example below:

https://github.com/flutter/flutter/blob/2f461fe9071178be5773650e04785272d364e857/packages/flutter/lib/src/widgets/selectable_region.dart#L3513-L3518

crackedhandle: Updated the documentation to use a template in `TextPainter.computeLineMetrics` and reused it via a macro in `RenderParagraph`, and added a dartpad-style example reference.

Renzo-Olivares:
Hi @crackedhandle, thank you for the contribution but after some more though I think for now we are going to hold off on this change and close the pull request. In the past (and present) we have chosen not to expose some of these layout metric APIs because we do not have enough data on real use-cases that this API could be used to solve. It also cannot reliably be used at the widgets level because the layout may become stale without any signal or notification. We want to move towards a more robust solution for exposing text layout where a user could reliably use it at the widgets level to decorate their text and be notified of layout changes.

While I'm closing this PR as out of scope, thank you again for putting together this patch and working with us on reviews. The team would gladly welcome future contributions from you!

crackedhandle:
> Hi @crackedhandle, thank you for the contribution but after some more though I think for now we are going to hold off on this change and close the pull request. In the past (and present) we have chosen not to expose some of these layout metric APIs because we do not have enough data on real use-cases that this API could be used to solve. It also cannot reliably be used at the widgets level because the layout may become stale without any signal or notification. We want to move towards a more robust solution for exposing text layout where a user could reliably use it at the widgets level to decorate their text and be notified of layout changes.
> 
> While I'm closing this PR as out of scope, thank you again for putting together this patch and working with us on reviews. The team would gladly welcome future contributions from you!

Thanks for the detailed explanation, @Renzo-Olivares! I completely understand the reasoning. I look forward to seeing the more robust solution you're working towards, and happy to contribute again when the time is right. and it was a great experience to work under your guidance.

Issue #177414, opened 2025-10-22
(Test cross-imports) Tests should go in the library that they're testing


justinmc:
Umbrella issue: https://github.com/flutter/flutter/issues/177028
Design doc: https://docs.google.com/document/d/1UHxALQqCbmgjnM1RNV9xE2pK3IGyx-UktGX1D7hYCjs/edit?tab=t.0

Tests should be located in the library whose code they are testing. If a tests needs to test multiple libraries (say, a platform adaptive constructor), then put it in Material. This will help with our general test organization and also help remove a devDependency after Material and Cupertino are decoupled from the framework.

### How to fix
Move the test into the correct library. It may not be the entire test file that needs to move, maybe only one or more tests within the file.

#### Located in Widgets, imported Material
 - [ ] [test/widgets/semantics_debugger_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/semantics_debugger_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement 
 - [ ] [test/widgets/routes_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/routes_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement and Convenience
 - [ ] [test/widgets/sliver_main_axis_group_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/sliver_main_axis_group_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement and Convenience
   - Button
 - [ ] [test/widgets/sliver_appbar_opacity_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/sliver_appbar_opacity_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement 
 - [ ] [test/widgets/semantics_tester_generate_test_semantics_expression_for_current_semantics_tree_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/semantics_tester_generate_test_semantics_expression_for_current_semantics_tree_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement
   - Could be a bigger refactor could fix it.
 - [ ] [test/widgets/page_route_builder_test.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/page_route_builder_test.dart)
   - Located in Widgets, imported Material.
   - Category: Misplacement 
[...]

PR #181245, opened 2026-01-21
Move decorated_sliver_test.dart from widgets to material

brahim-guaali:
## Description

Moves `decorated_sliver_test.dart` from `test/widgets/` to `test/material/` since it imports and uses Material widgets extensively (MaterialApp, Scaffold, etc.).

This is part of the effort to organize tests in the library they're testing.

Part of https://github.com/flutter/flutter/issues/177414

## Pre-launch Checklist

[talabat.com contributions](https://www.talabat.com) [Talabat Flutter PRs](https://github.com/search?q=org%3Aflutter+talabat&type=pullrequests)


justinmc (commented):
Looking at the test again, it might be possible to migrate this to not rely on Material at all (in which case I put this test file in the wrong issue and it should probably have gone in https://github.com/flutter/flutter/issues/177415, sorry about that). Could you try to migrate it? Maybe use TestWidgetsApp from https://github.com/flutter/flutter/pull/180456.

Either way, it looks like many tests from this file do not depend on Material at all, and they should stay where they are.

PR #162464, opened 2025-01-30
Don't strip symbols from `libapp.so` on android by default

gmackall:
Fixes https://github.com/flutter/flutter/issues/170664

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #181275, opened 2026-01-21
[Reland] Don't strip symbols from `libapp.so` on android by default

### LINKED ISSUE #170664: [android] Generate symbol files for AOT snapshots even without split-debug-info
Currently when building for AOT snapshots we usually [strip](https://github.com/flutter/flutter/blob/d9c63168c8d232e3232a3c1af3393b45db4c6cf8/packages/flutter_tools/lib/src/base/build.dart#L179-L183) them by default loosing all native symbolic information - because we don't expect this information to be useful. Dart runtime symbolizes stack traces it produces using information embedded into snapshot itself.

However native symbols can still come useful: consider for example a case when user hits an ANR due to long running Dart code (see https://github.com/flutter/flutter/issues/169906 as example). In this case they will be left with a stack trace in Android native format which they can't read[^1]. 

There is however a way to improve this experience: Android implementation should follow iOS one (which makes an effort to produce `dSYM`). Instead of stripping snapshots right away at generation time, it should produce unstripped snapshot and then split symbolic information out (e.g. using `llvm-objcopy`). The symbolic information could be then included in the app bundle - in which case it will be [automatically picked up by Play Store](https://developer.android.com/build/include-native-symbols#agp-4.1). When building APKs native symbols should be split into a separate archive to simplify subsequent upload to Play Store.

@jmagman @reidbaker 

[^1]: We could actually provide tooling to symbolize them using `analyze_snapshot`, but I think a better solution is to have Play Console symbolize them.

gmackall:
Relands https://github.com/flutter/flutter/pull/162464
Fixes https://github.com/flutter/flutter/issues/170664

The only test that failed last time was `Linux_pixel_7pro android_obfuscate_test`, and I ran it locally (`flutter test test/integration.shard/android_obfuscate_test.dart `) to both repro the failure and verify it passes on this branch with the additional changes.

Also does some stuff in the FGP:

The reason we failed the tests last time was we were bundling of the libapp.so code into a jar in the aar case, instead of simply including it as its own file in the source sets. I let gemini go on that part, modifying it to no longer pack as a jar,  but it looks correct to me. Now that the `libapp.so` file is simply included in this case, AGP is able to strip it (it's not hidden inside a jar).

Because this changes the FGP build process for add to app, I manually verified that the add to app flow isn't broken for including a flutter module both as source and as an aar.

<details>
  <summary>Before and after logs from the gradle task output</summary>
  
Before:
```
                     > Task :flutter:stripReleaseDebugSymbols NO-SOURCE
                     Skipping task ':flutter:stripReleaseDebugSymbols' as it has no source files and no previous output files.
```

After:

```
                    Task ':flutter:stripReleaseDebugSymbols' is not up-to-date because:
                       No history is available.
                     The input changes require a full rebuild for incremental task ':flutter:stripReleaseDebugSymbols'.
                     C/C++: android.ndkVersion from module build.gradle is [28.2.13676358]
                     C/C++: android.ndkVersion from module build.gradle is [28.2.13676358]
                     C/C++: android.ndkPath from module build.gradle is not set
                     C/C++: android.ndkPath from module build.gradle is not set
                     C/C++: ndk.dir in local.properties is not set
                     C/C++: ndk.dir in local.properties is not set
                     C/C++: Not considering ANDROID_NDK_HOME because support was removed after deprecation period.
                     C/C++: Not considering ANDROID_NDK_HOME because support was removed after deprecation period.
                     C/C++: android.ndkVersion from module build.gradle is [28.2.13676358]
                     C/C++: android.ndkPath from module build.gradle is not set
                     C/C++: sdkFolder is /Users/mackall/Library/Android/sdk
                     C/C++: ndk.dir in local.properties is not set
                     C/C++: Not considering ANDROID_NDK_HOME because support was removed after deprecation period.
                     C/C++: sdkFolder is /Users/mackall/Library/Android/sdk
                     C/C++: sdkFolder is /Users/mackall/Library/Android/sdk
                     Starting process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''. Working directory:
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter Command: /Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip
                     --strip-unneeded -o
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/stripped_native_libs/release/stripReleaseDebugSymbols/out/lib/armeabi-v7a/libapp.so
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/merged_native_libs/release/mergeReleaseNativeLibs/out/lib/armeabi-v7a/libapp.so
                     Starting process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''. Working directory:
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter Command: /Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip
                     --strip-unneeded -o
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/stripped_native_libs/release/stripReleaseDebugSymbols/out/lib/x86_64/libapp.so
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/merged_native_libs/release/mergeReleaseNativeLibs/out/lib/x86_64/libapp.so
                     Starting process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''. Working directory:
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter Command: /Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip
                     --strip-unneeded -o
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/stripped_native_libs/release/stripReleaseDebugSymbols/out/lib/arm64-v8a/libapp.so
                     /Users/mackall/development/BugTesting/mblahm/.android/Flutter/build/intermediates/merged_native_libs/release/mergeReleaseNativeLibs/out/lib/arm64-v8a/libapp.so
                     Successfully started process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''
                     Successfully started process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''
                     Successfully started process 'command '/Users/mackall/Library/Android/sdk/ndk/28.2.13676358/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-strip''
```

</details>

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -435,16 +445,6 @@ class FlutterPlugin : Plugin<Project> {
                     }
                 }
             }
-            // Copy the native assets created by build.dart and placed here by flutter assemble.
```
gmackall, line 438: Native assets are now copied by copyJniLibsTask to the jniLibs folder, which is registered as a source set.

gmackall:
re requesting your review @mraleph, as you were on the original pr (many months ago)

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
mraleph: I know it is somewhat orthogonal question, but did we make sure that native assets are also unstripped on Android by default (so that they could also be correctly picked up when publishing)?

cc @dcharkes

mraleph (approved):
LGTM from my side.

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
dcharkes: Currently we don't do any stripping for native assets, but maybe we should? I've filed:

* https://github.com/flutter/flutter/issues/181377

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -700,45 +700,41 @@ class FlutterPlugin : Plugin<Project> {
...
-                        // The `$project.layout.buildDirectory` is '.android/Flutter/build/' instead of 'build/'.
                         val buildDir =
                             "${FlutterPluginUtils.getFlutterSourceDirectory(project)}/build"
                         val nativeAssetsDir =
                             "$buildDir/native_assets/android/jniLibs/lib"
                         from("$nativeAssetsDir/$abi") {
                             include("*.so")
-                            rename { filename: String -> "lib/$abi/$filename" }
+                            into(abi!!)
                         }
```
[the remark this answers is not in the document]
reidbaker: +1 to !! being risky to use. Are you sure you want to crash here?

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -700,45 +700,41 @@ class FlutterPlugin : Plugin<Project> {
...
                 }
-            val packJniLibsTask: Task = packJniLibsTaskProvider.get()
-            FlutterPluginUtils.addApiDependencies(
-                project,
-                variant.name,
-                project.files({
-                    packJniLibsTask
-                })
-            )
+            val mergeJniLibsTaskName = "merge${FlutterPluginUtils.capitalize(variant.name)}JniLibFolders"
```
reidbaker, line 732: Non blocking but this task will need to migrated to the new style with inputs and outputs for agp 9.

packages/flutter_tools/lib/src/base/build.dart
```diff
@@ -170,6 +177,9 @@ class AOTSnapshotter {
       if (stripAfterBuild) {
         _logger.printTrace('Will strip AOT snapshot manually after build and dSYM generation.');
       }
+    } else if (targetingAndroidPlatform) {
+      stripAfterBuild = false;
```
reidbaker, line 181: uber nit: comment before setting the variable. (non blocking)

reidbaker (approved):

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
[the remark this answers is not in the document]
gmackall: Hmm I had assumed the line that existed before would crash, but string interpolation allows null in kotlin (it just inserts the string "null") so I was wrong about that.

I made both of these cases behave similarly, where we use the string "null" if abi is null (the `into()` method in the closure doesn't accept a `String?`)

Issue #180932, opened 2026-01-13
Re-enable fragment_shader_tests.dart 'Shader Compiler appropriately pads vec3 uniform arrays' test


b-luk:

b-luk:
This test was made obsolete by [#181822](https://github.com/flutter/flutter/pull/181822)

PR #181292, opened 2026-01-21
Re-enable fragment_shader_tests.dart 'Shader Compiler appropriately pads vec3 uniform arrays' test on Vulkan

b-luk:
- The [shader](https://github.com/flutter/flutter/blob/3b206173454ec192938126322c58b5a1f090652f/engine/src/flutter/lib/ui/fixtures/shaders/general_shaders/vec3_uniform.frag) has `uniform vec3[4] color_array;` and sets `fragColor = vec4(color_array[3].xyz, 1);`. Update the test to use the correct indexes for `color_array[3]`: 9, 10, 11 instead of 12, 13, 14.
- Skip the test for metal. Metal seems like it has some weird index padding issue. From testing it out:
  - `color_array[0].xyz` is set when setting indexes 0, 1, 2, as expected.
  - `color_array[1].xyz` should use indexes 3, 4, 5. But it seems to use indexes 4, 5, 6. Maybe there's some incorrect math making it skip index 3.
  - `color_array[2].xyz` should use indexes 6, 7, 8. But it seems to use indexes 8, 9, 10. Maybe there's some incorrect math making it skip indexes 3 and 7.
  - `color_array[3].xyz` should use indexes 9, 10, 11. But it seems to be impossible to set. Maybe it skips indexes 3, 7, and 11 and then tries to look at out-of-bounds indexes 12, 13, 14?
- Fix incorrect logic in the `_expectShaderRendersColor()` method.
  - The 4 bytes for each pixel of an image are in the format `0xRRGGBBAA`.
  - This method used to use `renderedBytes.buffer.asUint32List()`. This reads 4 bytes to a Uint32, but uses `Endian.host` endianness. Maybe this differs on different platforms, but on my Mac this is little endian. So the image's `0xRRGGBBAA` bytes are read into a Uint32 as `0xAABBGGRR`.
  - This is then compared to the target `Color.value`, which is in ARGB32 format: `0xAARRGGBB`.
  - So it's comparing the each image's pixel as `0xAABBGGRR` to the target color as `0xAARRGGBB`. It switches the RR and BB channels for every comparison.
  - This did not result in any test failures because `_expectShaderRendersColor()` is only ever used as a helper function for `_expectShaderRendersGreen()` and `_expectShaderRendersBlack()`. And (un)luckily, those perfectly ignore the RR and BB channels.
  - Fix the logic by reading each pixel's bytes with separate `getUint8()` calls and parsing them into a `Color` with  `fromARGB()`, using the correct byte offsets for each channel.

Partially addresses #180932. Re-enables the test for Vulkan, but it's still disabled for Metal.

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke (approved):
LGTM, thanks!

Issue #173104, opened 2025-08-01
ListView under a Transform breaks for some perspective values


FufferKS:
### Steps to reproduce

Hi, I'm experiencing render issues when playing around with perspective in Flutter.

I have a "Road Like" widget that is basically a ListView under perspective, but for certain matrix values it breakes.

Repro steps:
Create a `ListView`, wrap it in a `Transform` and `setEntry(3,1, highEnoughValueLike0point005)`, see it disappear.

I've raised similar issues in the past:
https://github.com/flutter/flutter/issues/155450
https://github.com/flutter/flutter/issues/154104
And got some help from @jason-simmons, so 🤞 its resolvable this time around as well.

I've provided minimal repro and a video, where it's visible for which values the ListView starts breaking (-0.0003 on the device I was testing) (iPhone 16 Pro Simulator).
- Using y scale makes the situation worse.
- Changing the height of the Container does not seem to matter.
- I am confident this is not device, platform or debug mode specific, as our actual app experiences this for both ios and android + release and debug.

### Expected results

The transform should work.

### Actual results

It does not.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Perspective Issues',
      theme: ThemeData(primarySwatch: Colors.blue),
      home: const MyHomePage(),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key});

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  double _yScale = 1.0;
  double _perspective = 0.000;

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(title: const Text('Perspective Issues')),
      body: Stack(
        alignment: Alignment.bottomCenter,
        children: [
          Transform(
            transform: Matrix4.identity()
              ..setEntry(3, 1, _perspective)
              ..scale(1.0, _yScale, 1.0),
            alignment: Alignment.center,

            child: ListView.builder(
              reverse: true,
              itemBuilder: (context, index) {
                final hue = (index * 137.5) % 360;
                final color = HSLColor.fromAHSL(1.0, hue, 0.6, 0.5).toColor();
                return Container(
                  height: 100.0,
[...]

PR #181312, opened 2026-01-22
Fix interpolation error in Rect::TransformAndClipBounds

flar:
We had been interpolating the points backwards in the `Rect::TransformAndClipBounds` code.

When we transform a rectangle under a perspective transform, some of the points may generate a homogenous bias (W) that is less than 0. This is a special condition that means that the point was transformed "behind the camera" (this is one way of looking at it and helps explain why we avoid such values). In order for it to contribute properly to the bounds in front of the "camera", it needs to be moved back into the positive (W>0) plane. We do this by interpolating from that out of bounds coordinate to either of its adjacent neighbors which are in bounds (i.e. their W>0) to move the point to a bias value that is just slightly in the "in bounds" half-plane - we choose W=(1/2^14) as our "in bounds" bias value.

Unfortunately, the code to interpolate the 2 coordinates was written with conflicting definitions of which direction we computed the `t` value and which direction we applied it to the 2 coordinates - thus generating nonsensical values.

This turned up while debugging a fix for https://github.com/flutter/flutter/issues/173104 where we fixed the transform logic in the clip code, but the rectangle was still partially clipped due to a bad result from the `GetCoverage` method of the rect geometry code. Note that this fix alone does not correct the problem in that Issue, it merely fixes a needed utility method that the fix will rely on.

We create a new method to compute a Lerp `t` value and then use the existing `Vector3::Lerp` method to apply it consistently.


engine/src/flutter/impeller/geometry/scalar.h
```diff
@@ -39,6 +39,20 @@ constexpr inline bool ScalarNearlyEqual(Scalar x,
...
+///          values being interploted.
+///
+/// For example, if you want to find the color between 2 other colors
+/// where the alpha value is A, you could compute:
+///
+///    Scalar t = GetLerpTValue(A, color1.alpha, color2.alpha);
+///    Color result = Color::Lerp(color1, color2, t);
+///    FML_DCHECK(ScalarNearlyEqual(A, result.alpha));
+constexpr inline Scalar GetLerpTValue(Scalar v, Scalar min, Scalar max) {
+  return (v - min) / (max - min);
```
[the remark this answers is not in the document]
flar: Done. Only capturing 0.0

engine/src/flutter/impeller/geometry/geometry_unittests.cc
```diff
@@ -37,6 +37,35 @@ TEST(GeometryTest, ScalarNearlyEqual) {
...
+
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(19.0f, 30.0f, 20.0f), 1.1f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(20.0f, 30.0f, 20.0f), 1.0f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(21.0f, 30.0f, 20.0f), 0.9f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(22.0f, 30.0f, 20.0f), 0.8f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(28.0f, 30.0f, 20.0f), 0.2f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(29.0f, 30.0f, 20.0f), 0.1f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(30.0f, 30.0f, 20.0f), 0.0f));
+  EXPECT_TRUE(ScalarNearlyEqual(GetLerpTValue(31.0f, 30.0f, 20.0f), -0.1f));
+}
```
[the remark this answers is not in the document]
flar: Done.

engine/src/flutter/impeller/geometry/scalar.h
```diff
@@ -39,6 +40,102 @@ constexpr inline bool ScalarNearlyEqual(Scalar x,
...
+/// @param a               The first instance to interpolate between.
+/// @param b               The second instance to interpolate between.
+/// @param fieldPtr        The field on the instances to interpolate to the
+///                        desired_value.
+/// @param desired_value   The value you want the result to contain in
+///                        the specified field.
+/// @return   The instance of T which exists on a linear space defined by
+///           a and b, but whose indicated field contains the desired_value.
+template <typename T>
+std::optional<T> LerpToFieldValue(const T& a,
```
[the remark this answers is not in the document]
flar: Done

engine/src/flutter/impeller/geometry/geometry_unittests.cc
gaaclarke, line 1: Instead of having a large geometry_unittests.cc file I would like us to organize this such that the tests are named after the file being tested.  So for example, we should have a scalar_unittests.cc with the tests on GetLerpTValue.

engine/src/flutter/impeller/geometry/rect_unittests.cc
gaaclarke, line 3116: Instead of having a test that tests multiple setups, I'd like to break these down into their own separate tests.

gaaclarke (changes requested):
Math looks good, tests contents looks good, nice find.  Can you please organize the tests to match conventions:
1) tests on functions in scalar.h should be in scalar_unittests.cc
1) each test should test one thing, that way if it fails there is a more brief context about what went wrong.

engine/src/flutter/impeller/geometry/rect.h
```diff
@@ -775,11 +775,14 @@ struct TRect {
...
     if (neighbor.z >= kMinimumHomogenous) {
-      auto t = (kMinimumHomogenous - p.z) / (neighbor.z - p.z);
-      clipped[index++] = {
-          (t * p.x + (1.0f - t) * neighbor.x) / kMinimumHomogenous,
-          (t * p.y + (1.0f - t) * neighbor.y) / kMinimumHomogenous,
-      };
+      auto lerp =
+          LerpToFieldValue(p, neighbor, &Vector3::z, kMinimumHomogenous);
+      // p and neighbor z values should have been distinct as one is out
+      // of the half-plane W>0 and the other is in the half-plane.
```
b-luk, line 781: They're in different half-spaces rather than different half-planes, right?

And the homogenous coordinate is z in this case rather than w, right?

So I think p and neighbor are on opposite sides of the two half-spaces that are split by the plane z=kMinimumHomogenous

engine/src/flutter/impeller/geometry/scalar.h
```diff
@@ -39,6 +40,102 @@ constexpr inline bool ScalarNearlyEqual(Scalar x,
...
+///      FML_DCHECK(ScalarNearlyEqual(A, result.alpha));
+///    }
+///
+/// @param val1            The first value to interpolate from.
+/// @param val2            The second value to interpolate to.
+/// @param desired_value   The value you want the result to interpolate.
+/// @return   The "t" value used to interpolate between val1 and val2 to
+///           produce the indicated desired_value, or std::nullopt if such
+///           interpolation is not possible.
+constexpr inline std::optional<Scalar> GetLerpTValue(Scalar val1,
```
b-luk, line 74: Would this function ever be used other than directly preceding a call to `LerpToFieldValue`? I can't imagine a useful case off the top of my head.

Maybe this shouldn't be a public API function and its logic should just be rolled in to `LerpToFieldValue`?

engine/src/flutter/impeller/geometry/rect.h
flar: Right half-space is a better term.

In this case we are using a method that homogenously transforms a 2D coordinate and returns a Vector3 with the homogenous weight in the z spot. Originally we probably should have created a 2D homogenous struct with x,y,w.

engine/src/flutter/impeller/geometry/scalar.h
flar: No, it would be preceding calls to Lerp, not LerpToFieldValue. LTFV uses it, but there are cases in the code where we also compute a T value that I haven't converted in this PR. For example, the gradient code has places where it interpolates colors based on relative index values that aren't represented by any field of the Color object. That could be encapsulated into some sort of "Lerp between these objects based on these independent relative measurements", but in the general case, LerpTValue would be used by multiple mechanisms.

engine/src/flutter/impeller/geometry/rect.h
flar: I could clarify this in the comments.

engine/src/flutter/impeller/geometry/geometry_unittests.cc
flar: That work was done at times when some of the more heavily tested examples were broken out in various PRs. I filed https://github.com/flutter/flutter/issues/181408 to finish the job for all of these cases. We might consider beefing up the testing of these utility objects as we discover their true API coverage by breaking them out.

This is outside the scope of this PR which is focused on fixing `TransformAndClipBounds`.

engine/src/flutter/impeller/geometry/rect_unittests.cc
flar: This test does seem out of hand.

engine/src/flutter/impeller/geometry/scalar.h
b-luk: Makes sense. If there are other cases outside of `LerpToFieldValue` where this will be useful, then this LGTM

engine/src/flutter/impeller/geometry/geometry_unittests.cc
gaaclarke: Yea, no need to reorganize all existing tests.  Just put the ones that are being added here in the correct place.

flar: The tests added here would drag along a bunch of other tests if I started refactoring here which would be a non-trivial refactoring...

PR #181317, opened 2026-01-22
Replace `pub run` mentions with `dart run`

sigurdm:
Following up on https://github.com/dart-lang/pub/issues/4737, this PR replaces deprecated `pub run` commands with `dart run`.


parlough (approved):
Thanks for updating these!

PR #181326, opened 2026-01-22
Add displayCornerRadii support to predictive back transitions.

### LINKED ISSUE #178463: Android's predictive back feature doesn't use the device's corner radius

ksokolovskyi:
Closes https://github.com/flutter/flutter/issues/178463

- Adds support of `MediaQuery.displayCornerRadii` to `PredictiveBackPageTransitionsBuilder`
- Adds support of `MediaQuery.displayCornerRadii` to `PredictiveBackFullscreenPageTransitionsBuilder`

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/flutter/test/material/predictive_back_page_transitions_builder_test.dart
```diff
@@ -601,6 +601,92 @@ void main() {
...
+
+        expect(find.text('push'), findsNothing);
+        expect(find.text('page b'), findsOneWidget);
+        expect(_findPredictiveBackPageTransition(pageTransitionsBuilder), findsNothing);
+        expect(_findFallbackPageTransition(pageTransitionsBuilder), findsOneWidget);
+        expect(find.byType(ClipRRect), findsNothing);
+
+        // Only Android supports backGesture channel methods. Other platforms will
+        // do nothing.
+        if (defaultTargetPlatform != TargetPlatform.android) {
```
navaronbracke: Can we instead use the variant of the test to tweak this? I.e. have an Android only test and a test for the other platforms

justinmc: Or we could only run this test on Android. I think that above this line, the test is not testing anything relevant to the displayCornerRadii feature, it's testing stuff that is already tested elsewhere.

justinmc (approved):
Surprisingly straightforward change now that https://github.com/flutter/flutter/pull/179219 is done. Thank you @ksokolovskyi for following up here! I'm really excited to fix this. LGTM :+1:

packages/flutter/test/material/predictive_back_page_transitions_builder_test.dart
navaronbracke: True, my point was that we can leverage the variant, instead of putting a return in the middle of the test :p

ksokolovskyi:
@navaronbracke @justinmc thanks a lot for your reviews! 

I've removed the `variant: TargetPlatformVariant.all()` and `defaultTargetPlatform != TargetPlatform.android` from the test, as I don't see any point in their usage.

In the test, we set the `PredictiveBackPageTransition` for all platforms and mock the platform channel, so there is no need to actually check for `defaultTargetPlatform`, as with that setup, `PredictiveBackPageTransition` will work for any `targetPlatform`.

navaronbracke (approved):

justinmc (approved):
LGTM 👍

PR #162903, opened 2025-02-07
Delete un-tested `hybrid_android_views`, add layering to `android_engine_test`.

matanlurey:
Another large test suite that has never been running.

The underlying thing I _think_ it wanted tested, which was hybrid composition, is now tested by `android_engine_test`. I also added a red square on top of each platform view so that, like `hybrid_android_views`, we can make sure layering is working as expected.

Any other ideas? Can also be follow-up PRs?

PR #181336, opened 2026-01-22
Reland hybrid views test

gmackall:
This test was deleted in https://github.com/flutter/flutter/pull/162903 under the claim that it wasn't running, but it was actually (just only in bringup). In fact the [glue code](https://github.com/flutter/flutter/blob/dbed04aa32e5f0319898d29c8996919ee6ae8e2e/dev/devicelab/lib/tasks/integration_tests.dart#L75) and [ci.yaml](https://github.com/flutter/flutter/blob/master/.ci.yaml#L2874) target are still around. It tested stuff that isn't covered by the current testing landscape, like alert dialogue popup and event types being preserved.

The original pr was not a pure deletion so I maintained the good changes (mostly a change to use shared code for the app under test).

Had to make some changes (described in comments below) but the test is passing in its current state.


dev/integration_tests/hybrid_android_views/lib/nested_view_event_page.dart
```diff
@@ -117,11 +117,10 @@ class NestedViewEventBodyState extends State<NestedViewEventBody> {
...
             ),
-            if (nestedViewClickCount > 0)
-              Text(
-                'Click count: $nestedViewClickCount',
-                key: const ValueKey<String>('NestedViewClickCount'),
-              ),
+            Text(
+              'Click count: $nestedViewClickCount',
+              key: const ValueKey<String>('NestedViewClickCount'),
+            ),
```
gmackall, line 123: Before this change the test would hang locally 100% of the time, due to a combination of
1. the tapChildView button not working
2. the fact that the next line in `main_test.dart` tries to find the widget with the text

So because the tap failed, the widget didn't exist, so we would hang forever.

I fixed this by changing the test to just tap on the platform view instead of tapping on a button that then taps on the platform view. But I also figured we should change the above so that the test just fails instead of hanging.

dev/integration_tests/hybrid_android_views/test_driver/main_test.dart
```diff
@@ -50,8 +50,10 @@ Future<void> main() async {
       final SerializableFinder addChildView = find.byValueKey('AddChildView');
       await driver.waitFor(addChildView);
       await driver.tap(addChildView);
-      final SerializableFinder tapChildView = find.byValueKey('TapChildView');
-      await driver.tap(tapChildView);
+      final SerializableFinder childView = find.byValueKey('PlatformView');
+      await driver.tap(childView);
+      // delay of 1000ms to allow widget to update
```
gmackall, line 55: In my defense this is what the original test was doing (the tapChildView button inserts a delay of 1000ms). 

I don't have a ton of experience writing this type of test, my understanding is that:
1. Because this is a driver test we don't have access to `tester.settle/pumpAndSettle`, which is what we really want here.
2. gemini suggested we use `driver.waitUntilNoTransientCallbacks()`, which is allegedly a driver equivalent, but it only worked some of the time. The delay seems to always work

If you know of a better solution let me know.

reidbaker (approved):
RSLGTM

PR #181340, opened 2026-01-22
Fixes metal vec3 uniform padding

### LINKED ISSUE #180873: enable 'FragmentShader shader with array uniforms renders correctly' for impeller

gaaclarke:
fixes https://github.com/flutter/flutter/issues/180873

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/impeller/compiler/compiler_test.cc
+++ b/engine/src/flutter/impeller/compiler/compiler_test.cc
@@ -106,6 +106,7 @@ bool CompilerTestBase::CanCompileAndReflect(
       entry_point_name);
 
   Reflector::Options reflector_options;
+  reflector_options.target_platform = GetParam();
   reflector_options.header_file_name = ReflectionHeaderName(fixture_name);
   reflector_options.shader_name = "shader_name";
 
--- a/engine/src/flutter/impeller/compiler/reflector.cc
+++ b/engine/src/flutter/impeller/compiler/reflector.cc
@@ -26,6 +26,7 @@
 #include "impeller/geometry/matrix.h"
 #include "impeller/geometry/scalar.h"
 #include "impeller/runtime_stage/runtime_stage.h"
+#include "runtime_stage_types_flatbuffers.h"
 #include "spirv_common.hpp"
 
 namespace impeller {
@@ -384,9 +385,11 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
           (spir_type.columns == 1 || spir_type.columns == 3)) {
         for (size_t c = 0; c < spir_type.columns; c++) {
           for (size_t v = 0; v < 3; v++) {
-            uniform_description.padding_layout.push_back(1);
+            uniform_description.padding_layout.push_back(
+                fb::StructByteType::kFloat);
           }
-          uniform_description.padding_layout.push_back(0);
+          uniform_description.padding_layout.push_back(
+              fb::StructByteType::kPadding);
         }
       }
     }
@@ -415,7 +418,7 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
     size_t binding =
         compiler_->get_decoration(ubo.id, spv::Decoration::DecorationBinding);
     auto members = ReadStructMembers(ubo.type_id);
-    std::vector<uint8_t> padding_layout;
+    std::vector<fb::StructByteType> padding_layout;
     size_t float_count = 0;
 
     for (size_t i = 0; i < members.size(); i += 1) {
@@ -426,7 +429,7 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
           size_t padding_count =
               (member.size + sizeof(float) - 1) / sizeof(float);
           while (padding_count > 0) {
-            padding_layout.push_back(0);
+            padding_layout.push_back(fb::StructByteType::kPadding);
             padding_count--;
           }
           break;
@@ -437,18 +440,18 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
             // and 0 layout property per byte of padding
             for (auto i = 0; i < member.array_elements; i++) {
               for (auto j = 0u; j < member.size / sizeof(float); j++) {
-                padding_layout.push_back(1);
+                padding_layout.push_back(fb::StructByteType::kFloat);
               }
               for (auto j = 0u; j < member.element_padding / sizeof(float);
                    j++) {
-                padding_layout.push_back(0);
+                padding_layout.push_back(fb::StructByteType::kPadding);
               }
             }
           } else {
             size_t member_float_count = member.byte_length / sizeof(float);
             float_count += member_float_count;
             while (member_float_count > 0) {
-              padding_layout.push_back(1);
+              padding_layout.push_back(fb::StructByteType::kFloat);
               member_float_count--;
             }
           }
--- a/engine/src/flutter/impeller/compiler/reflector.cc
+++ b/engine/src/flutter/impeller/compiler/reflector.cc
@@ -377,11 +377,6 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
         uniform_description.type == spirv_cross::SPIRType::BaseType::Float) {
       // Metal aligns float3 to 16 bytes.
       // Metal aligns float3x3 COLUMNS to 16 bytes.
-      // However, GetSize() calculates tight packing (12 bytes for float3,
-      // 36 bytes for float3x3).
-      // We must add padding to struct_layout to force GetSize() to match
-      // MSL alignment.
-      //
       // For float3: Size 12. Padding 4. Stride 16.
       // For float3x3: Size 36. Padding 12 (4 per col). Stride 48.
       size_t floats_per_element = 0;
@@ -393,14 +388,6 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
       } else if (spir_type.vecsize == 3 && spir_type.columns == 3) {
         // float3x3
         floats_per_element = 9;
-        // 3 columns * 1 pad per column = 3 pads.
-        // But here we just dump 1s and 0s linearly.
-        // Col0: 1,1,1,0. Col1: 1,1,1,0. Col2: 1,1,1,0.
-        // So we need to interleave?
-        // EmplaceUniform iterates struct_layout linearly and copies
-        // floats. If we push 1,1,1,0,1,1,1,0... it works if input is
-        // 1,1,1,1,1,1,1,1,1. Yes. But 'padding_floats' logic above was 'total
-        // padding'. We should just construct the full layout loop.
       }
 
       if (spir_type.vecsize == 3 &&
--- a/engine/src/flutter/impeller/compiler/reflector.cc
+++ b/engine/src/flutter/impeller/compiler/reflector.cc
@@ -386,10 +386,10 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
         for (size_t c = 0; c < spir_type.columns; c++) {
           for (size_t v = 0; v < 3; v++) {
             uniform_description.padding_layout.push_back(
-                fb::StructByteType::kFloat);
+                fb::PaddingType::kFloat);
           }
           uniform_description.padding_layout.push_back(
-              fb::StructByteType::kPadding);
+              fb::PaddingType::kPadding);
         }
       }
     }
@@ -418,7 +418,7 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
     size_t binding =
         compiler_->get_decoration(ubo.id, spv::Decoration::DecorationBinding);
     auto members = ReadStructMembers(ubo.type_id);
-    std::vector<fb::StructByteType> padding_layout;
+    std::vector<fb::PaddingType> padding_layout;
     size_t float_count = 0;
 
     for (size_t i = 0; i < members.size(); i += 1) {
@@ -429,7 +429,7 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
           size_t padding_count =
               (member.size + sizeof(float) - 1) / sizeof(float);
           while (padding_count > 0) {
-            padding_layout.push_back(fb::StructByteType::kPadding);
+            padding_layout.push_back(fb::PaddingType::kPadding);
             padding_count--;
           }
           break;
@@ -440,18 +440,18 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
             // and 0 layout property per byte of padding
             for (auto i = 0; i < member.array_elements; i++) {
               for (auto j = 0u; j < member.size / sizeof(float); j++) {
-                padding_layout.push_back(fb::StructByteType::kFloat);
+                padding_layout.push_back(fb::PaddingType::kFloat);
               }
               for (auto j = 0u; j < member.element_padding / sizeof(float);
                    j++) {
-                padding_layout.push_back(fb::StructByteType::kPadding);
+                padding_layout.push_back(fb::PaddingType::kPadding);
               }
             }
           } else {
             size_t member_float_count = member.byte_length / sizeof(float);
             float_count += member_float_count;
             while (member_float_count > 0) {
-              padding_layout.push_back(fb::StructByteType::kFloat);
+              padding_layout.push_back(fb::PaddingType::kFloat);
               member_float_count--;
             }
           }
--- a/engine/src/flutter/impeller/compiler/reflector.cc
+++ b/engine/src/flutter/impeller/compiler/reflector.cc
@@ -397,7 +397,7 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
         // But here we just dump 1s and 0s linearly.
         // Col0: 1,1,1,0. Col1: 1,1,1,0. Col2: 1,1,1,0.
         // So we need to interleave?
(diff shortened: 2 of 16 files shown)


engine/src/flutter/impeller/compiler/compiler_test.cc
```diff
@@ -106,6 +106,7 @@ bool CompilerTestBase::CanCompileAndReflect(
       entry_point_name);
 
   Reflector::Options reflector_options;
+  reflector_options.target_platform = GetParam();
```
gaaclarke, line 109: This is just a residual change from when gemini was fiddling with the compiler tests.  I kept it since it's a good change.

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
```diff
@@ -30,29 +30,47 @@ constexpr char kFloatType = 1;
...
-  size_t uniform_byte_index = 0u;
-  for (char byte_type : uniform.struct_layout) {
-    if (byte_type == kPaddingType) {
-      uniform_buffer.push_back(0.f);
-    } else {
-      FML_DCHECK(byte_type == kFloatType);
-      uniform_buffer.push_back(reinterpret_cast<const float*>(
-          input_data->data())[uniform_byte_index++]);
-    }
+  size_t count = uniform.array_elements.value_or(1);
```
b-luk: Is there a reason `count` is declared up here instead of within the callback? From a readability perspective, it makes sense to declare it right above the for-loop where it is used.

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
```diff
@@ -30,29 +30,47 @@ constexpr char kFloatType = 1;
...
+  if (count == 0) {
```
b-luk: In what situation would count be 0?

engine/src/flutter/impeller/compiler/reflector.cc
```diff
@@ -372,6 +372,25 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
...
     uniform_description.array_elements = GetArrayElements(spir_type);
+
+    if (TargetPlatformIsMetal(options_.target_platform) &&
+        uniform_description.type == spirv_cross::SPIRType::BaseType::Float) {
+      // Metal aligns float3 to 16 bytes.
+      // Metal aligns float3x3 COLUMNS to 16 bytes.
+      // For float3: Size 12. Padding 4. Stride 16.
+      // For float3x3: Size 36. Padding 12 (4 per col). Stride 48.
+
+      if (spir_type.vecsize == 3 &&
```
b-luk, line 384: So we need this specifically for vec3 and mat3, but not vec2/vec4/mat2/mat4? And specifically for floats but not ints or other types? Maybe add a comment to explain when it is/isn't needed.

engine/src/flutter/impeller/compiler/types.h
```diff
@@ -55,7 +55,7 @@ struct UniformDescription {
   size_t columns = 0u;
   size_t bit_width = 0u;
   std::optional<size_t> array_elements = std::nullopt;
-  std::vector<uint8_t> struct_layout = {};
+  std::vector<uint8_t> padding_layout = {};
```
b-luk: Should we try to use `fb::StructByteType` or some other shared enum as the type everywhere that padding_layout is used?

The 0s and 1s used in most places aren't very explanatory.  `runtime_effect_contents.cc` declares its own constants for 0 and 1 values, which is a little more informative but it's not great that they're siloed file-level constants.

I think it would be better if everything just consistently used the same enum. Ideally directly from the source of truth (`fb::StructByteType`) unless there is some dependency issue or another problem preventing that. If not `fb::StructByteType`, then maybe a shared enum declared in this file.

b-luk: I think it's not very intuitive how `padding_layout` is structured and used. An explanatory comment (maybe with examples) would be helpful.

b-luk (commented):
I'm not very familiar with this code, so I just have a bunch of nits for you and don't have any higher level thoughts on the overall approach. Maybe walley892 can probably be helpful for that.

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
gaaclarke: Yes, in c++ it's much more straightforward to capture an integer than a pointer.  We know that the callback is going to be called synchronously but there is no guarantee that will happen.

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
gaaclarke: Fields that are not arrays have `array_elements` as zero.

engine/src/flutter/impeller/compiler/types.h
gaaclarke: Done

engine/src/flutter/impeller/compiler/reflector.cc
gaaclarke: I do have a comment above that explains this case.  I'm letting the combination of integration tests and comments like this document when they are necessary.

engine/src/flutter/impeller/compiler/reflector.cc
```diff
@@ -372,6 +372,25 @@ std::shared_ptr<RuntimeStageData::Shader> Reflector::GenerateRuntimeStageData()
...
+          (spir_type.columns == 1 || spir_type.columns == 3)) {
```
walley892, line 385: This is kind of confusing. It may be more maintainable/readable if we had a function spir->array<uint> and explicitly made the call: `padding_layout = getPaddingLayout(type)` and define `getPaddingLayout` as something like

```{
...
// Vec3s are packed as Vec4
case spir_type.vecsize==3 && spir_type.columns == 1:
    return [1,1,1,0];
// Mat3s are packed as 3 vec4s
case spir_type.vecsize==3 && spir_type.columns == 3:
    return [1,1,1,0,
        1, 1, 1, 0,
        1, 1, 1, 0];
...
}
```
We may need to add more types in the future, or support for more platforms, or change the padding layout if we change spirv to use some different format. it would be ugly but more maintainable IMO.

engine/src/flutter/impeller/compiler/types.h
gaaclarke: Yea, I agree.  I was just following what was already here.  I fixed it, we use the shared enum now with flatbuffer when we can and explicitly convert it to the runtime enum.  The fb generated file wasn't available to me in the runtime_types so there is still a conversion, but it's explicit now.

engine/src/flutter/impeller/compiler/reflector.cc
walley892: I'm also skeptical about maintaining our own mappings of uniform types to their padding layout. Have you tried to coax this information out of spirv directly?

gaaclarke: Yea, those are good points.  Now that i've moved us to enums from bluks feedback it's not going to be that pretty anymore.  This level of refinement and refactoring will be better to consider after we have a complete test suite that runs.  Otherwise, refactoring may be premature.  I'm not quite sure how this is going to look when we are done.

> I'm also skeptical about maintaining our own mappings of uniform types to their padding layout. Have you tried to coax this information out of spirv directly?

No, it's not bad right now but something we could look into in the future if it makes sense.

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
b-luk: I don't see why `count` needs to be captured at all. The callback captures `uniform`, which `count` depends on. So can't we just declare `count` as a local variable inside the callback?

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
b-luk: I see. I would have expected fields that are not arrays to use the `optional` "no value" case. But ¯\\_(ツ)_/¯

engine/src/flutter/impeller/compiler/reflector.cc
b-luk: So just to be clear, ints behave differently from floats, so an `ivec3` doesn't need padding like a `vec3`?

engine/src/flutter/impeller/compiler/types.h
```diff
@@ -55,7 +56,10 @@ struct UniformDescription {
   size_t columns = 0u;
   size_t bit_width = 0u;
   std::optional<size_t> array_elements = std::nullopt;
-  std::vector<uint8_t> struct_layout = {};
+  /// The layout of padding bytes in the uniform buffer.
```
walley892, line 59: Could you add a comment about how this is handled differently for struct and array types (vulkan vs metal). In vulkan, the padding_layout includes padding for the array type fields. For Metal, the padding_layout doesn't, and we rely on the `array_elements`.

It's consistent to think of it this way, but a little confusing at first.

engine/src/flutter/impeller/core/runtime_types.cc
```diff
@@ -13,7 +13,9 @@ size_t RuntimeUniformDescription::GetSize() const {
     // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
     size *= array_elements.value();
   }
-  size += sizeof(float) * struct_layout.size();
+  if (type == kStruct) {
+    size += sizeof(float) * padding_layout.size();
+  }
```
walley892: Can you document that `GetSize` has two different meanings?

`myStruct::GetSize` is the total physical size of the thing. I.e, what we send to the GPU including padding

`myArrayUniform::GetSize` is the logical size of the thing, not what we actually send over to the GPU.

engine/src/flutter/impeller/core/runtime_types.h
```diff
@@ -38,6 +38,11 @@ struct RuntimeUniformDimensions {
   size_t cols = 0;
 };
 
+enum class RuntimeStructByteType : uint8_t {
```
walley892: Nit: `Struct` here is a misnomer, since it describes the byte type (padding or data) for non structs

engine/src/flutter/impeller/compiler/types.h
```diff
@@ -12,6 +12,7 @@
 #include <optional>
 #include <string>
 
+#include "runtime_stage_types_flatbuffers.h"
```
walley892, line 15: Is this import safe? I remember seeing a PR from a while ago about detangling the dependencies between types, runtime_types, and flatbuffers.

engine/src/flutter/impeller/runtime_stage/runtime_stage_types.fbs
```diff
@@ -43,7 +43,7 @@ table UniformDescription {
   rows: uint64;
   columns: uint64;
   array_elements: uint64;
-  struct_layout: [StructByteType];
+  padding_layout: [StructByteType];
```
walley892: Same comment about renaming the type from `StructByteType`

engine/src/flutter/impeller/runtime_stage/runtime_stage_unittests.cc
walley892, line 1: There's a `case` right above the one you modified that checks the format on Metal, can you modify that to verify the field you've added?

walley892 (commented):
I'm not sure of the overall approach in the long term, but this looks good for now. Could of comments

engine/src/flutter/impeller/compiler/reflector.cc
gaaclarke: I don't know if that's the case.  It may just be a case that we don't have ivec3 tests.  We may not even have ivec3 support.

engine/src/flutter/impeller/compiler/types.h
gaaclarke: Safe meaning it won't cause runtime errors or build errors?  Yes it's safe and doesn't change the dependency graph at all.

This is an improvement from what we had previously and I'm open to further refactoring if someone has a cohesive idea about what should change.

engine/src/flutter/impeller/compiler/reflector.cc
walley892: We don't support integer or boolean vec types

engine/src/flutter/impeller/compiler/types.h
gaaclarke: I think you are thinking about it a bit off.  padding_layout is agnostic to how its used.  On vulkan we use it to layout struct padding, on metal we use it for array layout.  It means the same thing in both cases though.  Adding comments to something about how it is used is fragile versus documenting what it is.

I've added a comment linking to where it is used.

engine/src/flutter/impeller/compiler/types.h
walley892: https://github.com/flutter/engine/pull/35360

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
gaaclarke: This comes from flatbuffers fwiw

engine/src/flutter/impeller/runtime_stage/runtime_stage_unittests.cc
gaaclarke: done

engine/src/flutter/impeller/runtime_stage/runtime_stage_types.fbs
gaaclarke: done

engine/src/flutter/impeller/core/runtime_types.h
gaaclarke: done

engine/src/flutter/impeller/core/runtime_types.cc
gaaclarke: Yea, I'm not a huge fan of this either.   I'll spend some time trying to detangle this.  Unfortunately the docstring is kind of ambiguous.  This is what I'd expect the implementation to be:

```c++
size_t RuntimeUniformDescription::GetSize() const {
  size_t size = 0;
  if (padding_layout.empty()) {
    size = dimensions.rows * dimensions.cols * bit_width / 8u;
  } else {
    size = sizeof(float) * padding_layout.size();
  }
  if (array_elements.value_or(0) > 0) {
    // Covered by check on the line above.
    // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
    size *= array_elements.value();
  }
  return size;
}
```

engine/src/flutter/impeller/entity/contents/runtime_effect_contents.cc
gaaclarke: Yea, missed we are already grabbing the uniform.  Done.

engine/src/flutter/impeller/compiler/reflector.cc
gaaclarke: Note to future self: spv::DecorationArrayStride and spv::DecorationMatrixStride might be apropos to trying to tease the info

engine/src/flutter/impeller/core/runtime_types.cc
walley892: If it works it works for now I guess

b-luk (approved):
lgtm as long as walley approves

engine/src/flutter/impeller/compiler/reflector.cc
walley892: I tried grabbing `spv::DecorationArrayStride` and `spv::DecorationMatrixStride` locally earlier today, both reported zero :(.

There's a TODO comment in the spirv source about surfacing the physical size of uniforms. Maybe we could follow up on that? [third_party/vulkan-deps/spirv-cross/src/spirv_reflect.cpp::358](https://github.com/KhronosGroup/SPIRV-Cross/blob/main/spirv_reflect.cpp#L358)

walley892 (approved):
Nice

engine/src/flutter/impeller/core/runtime_types.cc
gaaclarke: Ugh, this is bothering me.  For the vec3array in the test the current GetSize returns 24 bytes, but the proposed one returns 32 bytes.  This causes the test to fail.  This is not working as intended and shouldn't be landed until this is addressed.  My brain is spent for now.  I'll look monday =(

gaaclarke: Okay, got it sorted out.  There is a difference between the size a uniform takes up in the float buffer from dart and the size it will take up on the GPU because there is no padding in the dart float buffer.  I've gotten rid of the oddity and split the names apart.  This should clear everything up.  I'm not 100% happy with it but this gives the next person looking at the code an understanding of what is actually going on.

engine/src/flutter/impeller/core/runtime_types.cc
```diff
@@ -4,17 +4,34 @@
 
 #include "impeller/core/runtime_types.h"
 
+#include "flutter/fml/logging.h"
+
 namespace impeller {
 
-size_t RuntimeUniformDescription::GetSize() const {
+size_t RuntimeUniformDescription::GetDartSize() const {
```
walley892, line 11: Instead of calling these `GPU` and `Dart` size, what about `Physical` and `Logical`?

gaaclarke: I think those are probably equally valid names.  We're going to have to lean into the docstring on this one since there is only so much semantic information you can pack into these names.  Logical vs Physical is sort of loaded with different meanings too already that aren't a 100% match.

PR #181362, opened 2026-01-23
Fix nested selection area menu

### LINKED ISSUE #181231: [SelectionArea] An incorrect context menu popped up when SelectionArea was nested
### Steps to reproduce

Right-clicking within a nested SelectionArea area may sometimes pop up the ContextMenu of the parent Selection.

### Expected results

Right-clicking within a nested SelectionArea area always pop up the ContextMenu of current SelectionArea.

### Actual results

Right-clicking within a nested SelectionArea area may sometimes pop up the ContextMenu of the parent Selection.

### Code sample

```dart

void main() {
  runApp(const MainApp());
}

  const MainApp({super.key});

  @override
}

class _MainAppState extends State<MainApp> {
  @override
    return MaterialApp(home: HomeScreen());
  }
}

class HomeScreen extends StatelessWidget {
  const HomeScreen({super.key});

  @override
    return Scaffold(
      appBar: AppBar(title: Text('Home')),
      body: SelectionArea(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.spaceEvenly,
          children: [
            Text('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'),
            SelectionArea(
              child: Container(
                height: 300,
                color: Colors.yellow,
                alignment: Alignment.center,
                child: Text('bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'),
              ),
            ),
            Text('ccccccccccccccccccccccccccccc'),
          ],
        ),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>

Image

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[√] Flutter (Channel stable, 3.38.6, on Microsoft Windows [版本 10.0.26100.7462], locale zh-CN) [231ms]
    • Flutter version 3.38.6 on channel stable at D:\Flutter\flutter_windows_latest\flutter
    • Framework revision 8b87286849 (12 days ago), 2026-01-08 10:49:17 -0800
    • Engine revision 78fc3012e4
    • Dart version 3.10.7
    • DevTools version 2.51.1
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android,
      enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging

[√] Windows Version (Windows 11 or higher, 24H2, 2009) [717ms]

[√] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [1,340ms]
    • Android SDK at D:\Android\Sdk
    • Emulator version 36.3.10.0 (build_id 14472402) (CL:N/A)
    • ANDROID_HOME = D:\Android\Sdk
    • Java binary at: D:\Android\Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14196175-b1038.72)

[√] Chrome - develop for the web [66ms]

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.14.12) [65ms]
    • Visual Studio at D:\VisualStudio\Community

[√] Connected device (3 available) [257ms]
    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [版本 10.0.26100.7462]
    • Chrome (web)      • chrome  • web-javascript • Google Chrome 142.0.7444.176
    • Edge (web)        • edge    • web-javascript • Microsoft Edge 144.0.3719.82

[√] Network resources [832ms]

• No issues found!
```

</details>

zulkaif121:
Fixes #181231

This PR fixes an issue where right-clicking inside a nested SelectionArea would incorrectly trigger the context menu of the parent SelectionArea.

Description
Previously, right-click events in nested SelectionArea widgets were not properly scoped, causing the event to propagate up and be handled by parent SelectionArea widgets as well. This resulted in the parent stealing focus or showing its own context menu instead of the child's.

The Fix: I introduced a 
_SelectableRegionSecondaryTapNotification
 to coordinate right-click handling:

When 
_handleRightClickDown
 is triggered, it dispatches this notification.
The 
SelectableRegion
 listens for this notification from its descendants.
If a parent receives the notification (meaning a child has already processed the right-click), it sets a temporary flag (_ignoreOneRightClick) to ignore the subsequent right-click event handled by its own gesture recognizer.
Tests
Added a regression test in 
packages/flutter/test/material/selection_area_test.dart
 verifying that:
Interacting with a child SelectionArea (right-click) does not select the parent text.
The child content is correctly selected.

## Pre-launch Checklist

- [✅ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [✅ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [✅ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [✅ ] I signed the [CLA].
- [✅ ] I listed at least one issue that this PR fixes in the description above.
- [✅ ] I updated/added relevant documentation (doc comments with `///`).
- [✅ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [✅ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [✅ ] All existing and new tests are passing.


packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -233,6 +234,15 @@ const double _kSelectableVerticalComparingThreshold = 3.0;
 ///    and provides api to dispatch selection event to the collected widget.
 ///  * [SelectionListener], which enables accessing the [SelectionDetails] of
 ///    the selectable subtree it wraps.
+/// A notification that indicates a secondary tap (right click) has occurred in a
```
navaronbracke: This block is in the wrong spot? The docs above belong to `SelectableRegion`, not `_SelectableRegionSecondaryTapNotification`.

zulkaif121: Fixed can u check now

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -233,6 +234,7 @@ const double _kSelectableVerticalComparingThreshold = 3.0;
 ///    and provides api to dispatch selection event to the collected widget.
 ///  * [SelectionListener], which enables accessing the [SelectionDetails] of
 ///    the selectable subtree it wraps.
+
```
navaronbracke: This newline should be removed?

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -397,6 +399,7 @@ class SelectableRegionState extends State<SelectableRegion>
 
   final Map<Type, GestureRecognizerFactory> _gestureRecognizers =
       <Type, GestureRecognizerFactory>{};
+  bool _ignoreOneRightClick = false;
```
navaronbracke: Should this be `_ignoreOnRightClick` ? I find the name a bit confusing.

packages/flutter/lib/src/widgets/selectable_region.dart
```diff
@@ -1946,31 +1955,43 @@ class SelectableRegionState extends State<SelectableRegion>
...
-      onTapOutside: (PointerDownEvent event) {
-        // To match the native web behavior, this selectable region is
-        // unfocused when tapping outside of it causing the selection to
-        // be dismissed.
-        //
-        // Tapping outside the selectable region does not unfocus
-        // the region on non-web platforms.
-        if (kIsWeb) {
-          _focusNode.unfocus();
+    return NotificationListener<_SelectableRegionSecondaryTapNotification>(
```
navaronbracke, line 1963: I have a feeling that this notification listener, which uses a private notification class defined in this file, is maybe not the right solution?

In https://github.com/flutter/flutter/issues/181231#issuecomment-3775915102 someone mentioned using `SelectionContainer.disabled`, but that did not help them out.

Maybe we can fix the behavior of `SelectionContainer.disabled` instead.

What if we do the following:

Given the pointer gesture happens within a `SelectionContainer.disabled` widget's region, consume the pointer event and prevent it from bubbling up to any parent `SelectionContainer`.

I'm not sure about the implications of fixing it like that, regarding overlapping `SelectionContainer`s. (can those even overlap visually?)

packages/flutter/test/material/selection_area_test.dart
```diff
@@ -706,4 +706,84 @@ void main() {
...
+      final TestGesture gesture = await tester.startGesture(
+        tester.getCenter(childTextFinder),
+        kind: PointerDeviceKind.mouse,
+        buttons: kSecondaryMouseButton,
+      );
+      await gesture.up();
+      await tester.pumpAndSettle();
+
+      // Verify context menu appears (skip on macOS and web where toolbars don't show in tests)
+      if (defaultTargetPlatform != TargetPlatform.macOS && !kIsWeb) {
```
navaronbracke: Can we move this to a different test and use the test `variant` instead? So one test for MacOS, one test for web and one for the other platforms

packages/flutter/test/material/selection_area_test.dart
```diff
@@ -706,4 +706,84 @@ void main() {
     );
     expect(tester.getSize(find.byType(SelectionArea)), Size.zero);
   });
+
+  testWidgets(
```
navaronbracke: The fix is for `SelectableRegion`, but this tests `SelectionArea` ? You should be testing `SelectableRegion` instead. That should go in `test/widgets/selectable_region_test.dart`

packages/flutter/test/material/selection_area_test.dart
```diff
@@ -706,4 +706,84 @@ void main() {
...
+      }
+
+      // Tap Select All if available
+      if (defaultTargetPlatform != TargetPlatform.macOS && !kIsWeb) {
+        await tester.tap(find.text('Select all'));
+        await tester.pumpAndSettle();
+      }
+
+      expect(
+        parentSelection?.plainText,
```
navaronbracke: Nit: `expect(parentSelection, isNull);`

packages/flutter/test/material/selection_area_test.dart
```diff
@@ -706,4 +706,84 @@ void main() {
...
+        isNull,
+        reason: 'Parent should NOT be selected when interacting with child context menu',
+      );
+
+      if (defaultTargetPlatform != TargetPlatform.macOS && !kIsWeb) {
+        expect(
+          childSelection?.plainText,
```
navaronbracke: Nit: `expect(childSelection?.plainText, 'Child Text', reason: 'Child content SHOULD be selected');`

packages/flutter/lib/src/widgets/selectable_region.dart
zulkaif121: Thanks for the suggestion! I did look into the SelectionContainer.disabled approach, but unfortunately it wouldn’t solve this specific issue because the problem occurs at the gesture handling level rather than the selection registration level.

SelectionContainer.disabled prevents child selectables from registering with a parent SelectionRegistrar, but it does not affect how RawGestureDetector routes pointer events. Both parent and child TapGestureRecognizers still enter the gesture arena, and both end up firing handleTapDown after the 100ms kPressTimeout deadline. The original issue reporter also confirmed that wrapping with SelectionContainer.disabled did not resolve the behavior.

The Notification pattern aligns with how Flutter already coordinates child-to-parent communication in similar cases (for example ScrollNotification and KeepAliveNotification). It allows the fix to remain narrowly scoped to right-click handling without altering gesture routing or the broader selection registration system.

I also discovered an issue in my original test. It performed startGesture followed immediately by gesture.up(), which meant the 100ms kPressTimeout never elapsed and the parent’s _handleRightClickDown was never triggered. As a result, the test passed regardless of the fix. I’ve updated the test to include await tester.pump(const Duration(milliseconds: 200)) between the down and up events to better simulate real user interaction. With this change, the test now fails without the fix (parent selects Parent TextOther Parent Text) and passes with it.

packages/flutter/lib/src/widgets/selectable_region.dart
zulkaif121: Removed.

packages/flutter/lib/src/widgets/selectable_region.dart
zulkaif121: Renamed to _childHandledSecondaryTap for clarity , it better describes what the flag         
 represents (a child already processed the secondary tap).

packages/flutter/test/material/selection_area_test.dart
zulkaif121: Moved to test/widgets/selectable_region_test.dart.

zulkaif121: Split into two separate tests with explicit variant parameters, one for Linux/Windows (where
context menu + Select All is tested) and one for macOS.

packages/flutter/test/material/selection_area_test.dart
zulkaif121: Moved to test/widgets/selectable_region_test.dart.

packages/flutter/test/material/selection_area_test.dart
zulkaif121: Fixed.

packages/flutter/test/material/selection_area_test.dart
zulkaif121: Fixed.

packages/flutter/lib/src/widgets/selectable_region.dart
Renzo-Olivares: Thank you for the great discussion here!

I think there are a few questions we have to ask about nested `SelectableRegion`s. For example:

* What are user expectations of nested `SelectableRegion`s?
	* Should nested `SelectableRegion`s work as one or should a selection starting in an inner `SelectableRegion` not be able to escape that region?

The existing behavior is that nested `SelectableRegion` selections cannot escape the region they were initiated in, i.e. if the selection starts on the inner region it cannot escape this region. If the selection starts in the outer region then it cannot enter the inner region. This is because each `SelectableRegion` has its own `SelectionRegistrar`. If nested `SelectableRegion`s were to work as one then  they wouldn't be constrained by the bounds of their region.

Option 1: Make `SelectableRegion` smart enough to know when it is nested in another `SelectableRegion` and if so do not install gesture infrastructure to prevent gesture conflicts.

One scenario users will likely run into is the case where they may be using some third-party widget or library that uses `SelectableRegion` as part of their implementation detail. For example take a List wrapped with a `SelectableRegion`, where one of the children of the list is from a third party library that uses `SelectableRegion` in its implementation. In this scenario we may hit the wrong context menu issue that this change is trying to fix. In this case I think `SelectableRegion` should be smart enough to know that it is already in a `SelectableRegion` so there is no need to include any gesture handling.

The above scenario describes the third party library as the child of the user's `SelectableRegion`, but we can also run into the scenario where the third party library is the parent. In this case `SelectableRegion` should also be smart enough to know that it is already in a `SelectableRegion`.

This option puts the responsibility on the framework to check if there is an ancestor `SelectableRegion` and if so do not install gesture handling. This way framework users, users of third party libraries, and developers of third party libraries do not have to worry about nesting `SelectableRegion`s.

With this option we also have to consider how a user might accomplish the previous behavior of constrained regions. This should still be possible through the use of `SelectionContainer` and setting up your own registrar.

Option 2: Say let's not allow nested `SelectableRegion`s at all, and add some assertion to check this. Then users would not run into gesture conflicts imposed by nested `SelectableRegion`s. 

This option puts the responsibility on the third-party library developer to check if there is an ancestor `SelectableRegion` and if so do not install your own `SelectableRegion`. If they do not do this then their users will hit assertions/ the package may be completely incompatible with their layout. Take the List wrapped in a `SelectableRegion` example, if a child of this list is a third-party library that wraps itself with a `SelectableRegion` indiscriminately then the assertion will always be hit, and the user of the third-party library has no workaround because they need a `SelectableRegion` around their list to be able to select other items in the list.

I'm leaning towards Option 1 because of the possible incompatibilities that Option 2 introduces for third-party libraries. Essentially users of the described scenario would be broken until the third-party library author updated their use of `SelectableRegion` to consider nesting.

Curious to here any thoughts on these options.

cc @chunhtai if you had any thoughts.

Some additional notes:

When looking at the `GestureArena` when this failure happens i'm observing the inner `SelectableRegion` win the gesture but both `SelectableRegion` right click handlers are called because of `didExceedDeadline`. The tap gesture recognizer has a timer it sets and when the timer goes off it decides to trigger its gesture handlers regardless of whether it won the arena. What is likely happening here in some cases is that both tap recognizers from both `SelectableRegion`s are competing for a long enough time that `didExceedDeadline` is triggered for both regions triggering the erroneous context menu.

Renzo-Olivares (commented):
Hi @zulkaif121, thank you for the contribution! I left a few thoughts let me know what you think.

packages/flutter/lib/src/widgets/selectable_region.dart
zulkaif121: I checked Option 1, but I noticed it doesn't satisfy the original issue's expected behavior, the
  reporter expects right-clicking in the inner SelectionArea to always show the inner's context menu. With
  Option 1, the inner never handles gestures, so only the outer's menu appears. Is that the intended
  tradeoff, or should the fix preserve the inner's ability to show its own context menu?

justinmc:
Making it a draft to remove from the review queue but we can undraft it when the conflict is fixed.

Piinks:

PR #181074, opened 2026-01-16
Revert DropdownMenu non-nullable breaking change

Gustl22:
As explained in #180121 the PR #176711 is a breaking change, which must be properly documented.

Closes #180121

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[...]

Issue #180121, opened 2025-12-19
Non-nullable DropdownMenu is breaking change


Gustl22:
### Steps to reproduce

#175090 and PR #175090 does break current implementation, since: [3.39.0-0.1.pre](https://github.com/flutter/flutter/releases/tag/3.39.0-0.1.pre)

An example can be found here:
https://github.com/flutter/flutter/pull/176711#issuecomment-3490055344

Further the type safe part of #175090 (in my opinion) does raise more problems than it actually solves. I personally don't see removing the ambiguity has any real world scenarios. In contrast the need of a null value is actually used. I don't see a concrete issue with getting the same value can result from two different events does have a negative impact (why would someone want to differentiate that, and also with this change there is not even a built-in option to differentiate. 

### Expected results

To either:

- Revert
- Not break the current implementations, e.g. https://github.com/flutter/flutter/pull/176711#issuecomment-3674266049
- Write a migration guide or in the docs how to circumvent the lack of using null as value (https://github.com/flutter/flutter/pull/176711#issuecomment-3580201716) (least preferred)

### Actual results

The changes are breaking the below example and are a valid approach to reflect empty selection: https://material-web.dev/components/select/#usage

There are already three cases which were mentioned to break (myself included).

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() => runApp(const MyApp());

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      title: 'Flutter Demo',
      home: Card(
        child: DropdownMenu<String?>(
          dropdownMenuEntries: [
            DropdownMenuEntry(value: null, label: 'Label'),
            DropdownMenuEntry(value: '1', label: '1'),
            DropdownMenuEntry(value: '2', label: '2'),
          ],
        ),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[Paste your output here]
```

</details>

tirth-patel-nc:
Thanks for the report. Seeing the same behaviour with master SDK version. It is not throwing this error on 3.38.5 hence marking it as a regression.

```
master : 3.40.0-1.0.pre-307
```

[...]

PR #176711, opened 2025-10-08
Make DropdownMenu generic type non nullable

ValentinVignal:
Fixes https://github.com/flutter/flutter/issues/175090

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #181367, opened 2026-01-23
[beta] Cherry pick https://github.com/flutter/flutter/pull/181074

dkwingsmt:
This PR cherrypicks https://github.com/flutter/flutter/pull/181074 into beta, which is mostly a revert of https://github.com/flutter/flutter/pull/176711 (except that some docs are left unreverted and a few new tests are added.) This is needed to address a breakage as described in https://github.com/flutter/flutter/issues/180121.

**Impacted Users:** Apps that use `DropdownMenu` or `DropdownMenuFormField` for a nullable class, such as `DropdownMenu<int?>`.

**Impact Description:** Can not compile.

**Workaround:** Yes, through non-trivial changes. For example, the developer can replace  `DropdownMenu<int?>` with `DropdownMenu<Maybe<int>>`, as long as they import `Maybe` or create their own equivalence. 

**Risk:** Low. This PR is basically a revert.

**Test Coverage:** Yes. The widget has high test coverage.

**Validation Steps:** The PR compiles and tests pass.

### DIFF
--- a/packages/flutter/test/material/dropdown_menu_form_field_test.dart
+++ b/packages/flutter/test/material/dropdown_menu_form_field_test.dart
@@ -32,6 +32,14 @@ void main() {
     return find.widgetWithText(MenuItemButton, menuItem.label).last;
   }
 
+  Finder findMenuItemButton(String label) {
+    // For each menu items there are two MenuItemButton widgets.
+    // The last one is the real button item in the menu.
+    // The first one is not visible, it is part of _DropdownMenuBody
+    // which is used to compute the dropdown width.
+    return find.widgetWithText(MenuItemButton, label).last;
+  }
+
   testWidgets('Creates an underlying DropdownMenu', (WidgetTester tester) async {
     await tester.pumpWidget(
       MaterialApp(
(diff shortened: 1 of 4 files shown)


packages/flutter/test/material/dropdown_menu_form_field_test.dart
```diff
@@ -32,6 +32,14 @@ void main() {
     return find.widgetWithText(MenuItemButton, menuItem.label).last;
   }
 
+  Finder findMenuItemButton(String label) {
+    // For each menu items there are two MenuItemButton widgets.
+    // The last one is the real button item in the menu.
+    // The first one is not visible, it is part of _DropdownMenuBody
+    // which is used to compute the dropdown width.
+    return find.widgetWithText(MenuItemButton, label).last;
+  }
```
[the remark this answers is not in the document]
Gustl22: I originally thought about this, but decided to keep both methods to not need a refactor in a "revert" PR and keep the method in sync with the conventions in `dropdown_menu_test.dart`. We could do a follow-up refactor, if desired.

packages/flutter/test/material/dropdown_menu_test.dart
```diff
@@ -5333,6 +5333,49 @@ void main() {
...
+
+    // Open the menu.
+    await tester.tap(find.byType(DropdownMenu<String?>));
+    await tester.pump();
+
+    // Select the 'None' item.
+    await tester.tap(findMenuItemButton(selectNoneLabel));
+    await tester.pumpAndSettle();
+
+    expect(controller.text, selectNoneLabel);
```
LongCatIsLooong, line 5377: Huh I thought the text should be empty or something since  the widget uses `null` for "no selection"?

LongCatIsLooong (commented):
The revert LGTM. But I don't think the original commit has reached stable: https://github.com/flutter/flutter/commit/0205d4e8cc24511918e0348410f34605102682c9. So can you confirm that this is a CP to beta?

justinmc (approved):
LGTM :+1: but I'm unsure about @LongCatIsLooong's comment https://github.com/flutter/flutter/pull/181367/files#r2733486116.

packages/flutter/test/material/dropdown_menu_test.dart
Gustl22: What exactly is the difference? Only the label? Is that something we can integrate in https://github.com/flutter/flutter/pull/181369, but still merge this cherry pick?

Issue #116525, opened 2022-12-05
Support Material 3 motion - Easing and duration


guidezpl:
https://m3.material.io/styles/motion/easing-and-duration/applying-easing-and-duration

- [ ] Add support for all curves
- [ ] Add support for all durations

Existing material curves are at https://github.com/flutter/flutter/blob/main/packages/flutter/lib/src/material/curves.dart#L26. We should probably create `motion.dart` and put everything there. Some thought needs to be put into the naming for legacy curves.

PR #181403, opened 2026-01-23
Use AnimationStyle curve and reverseCurve in ModalBottomSheet animation

### LINKED ISSUE #38958: How do I configure BottomSheet opening and closing animation curves?

### LINKED ISSUE #174857: AnimationStyle curve and reverseCurve are not being used in showModalBottomSheet
### Steps to reproduce

1. run `showModalBottomSheet` onTap

### Expected results

`Curves.elasticOut` animation

### Actual results

`Curves.linear` animation

### Code sample

```dart
showModalBottomSheet(
  context: context,
  sheetAnimationStyle: AnimationStyle(
    curve: Curves.elasticOut,
    duration: Duration(seconds: 2),
    reverseCurve: Curves.elasticOut,
    reverseDuration: Duration(seconds: 2),
  ),
);
```

</details>

### Screenshots or Video

_No response_

### Logs

_No response_

### Flutter Doctor output

```console
[√] Flutter (Channel stable, 3.35.2, on Microsoft Windows [versÆo 10.0.26100.4946], locale pt-BR) [844ms]
    • Flutter version 3.35.2 on channel stable at D:\flutter
    • Framework revision 05db968908 (8 days ago), 2025-08-25 10:21:35 -0700
    • Engine revision a8bfdfc394
    • Dart version 3.9.0
    • DevTools version 2.48.0

[√] Windows Version (Windows 11 or higher, 24H2, 2009) [2,6s]

[√] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [5,1s]
    • Android SDK at C:\Users\lucas\AppData\Local\Android\sdk
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java

[√] Chrome - develop for the web [537ms]

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.14.12) [536ms]
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community

[√] Android Studio (version 2025.1.1) [21ms]

[√] VS Code (version 1.103.2) [18ms]
    • VS Code at C:\Users\lucas\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.118.0

[√] Connected device (4 available) [631ms]
    • SM M346B (mobile) • 192.168.1.171:46641 • android-arm64  • Android 15 (API 35)
    • Windows (desktop) • windows             • windows-x64    • Microsoft Windows [versÆo 10.0.26100.4946]
    • Chrome (web)      • chrome              • web-javascript • Google Chrome 138.0.7204.184
    • Edge (web)        • edge                • web-javascript • Microsoft Edge 139.0.3405.125

[√] Network resources [1.125ms]

• No issues found!
```

</details>

Mairramer:
Fixes flutter/flutter#38958
Fixes flutter/flutter#174857

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -1080,6 +1080,17 @@ class ModalBottomSheetRoute<T> extends PopupRoute<T> {
     return _animationController!;
   }
 
+  @override
+  Animation<double> createAnimation() {
+    final Animation<double> parent = super.createAnimation();
+
+    return CurvedAnimation(
```
victorsanni: This is causing memory leaks, instead create a state instance variable to store and eventually dispose this `CurvedAnimation`. See [CupertinoContextMenu](https://github.com/flutter/flutter/blob/72321c82f99852685c5f9a25a44adbadb7d020ff/packages/flutter/lib/src/cupertino/context_menu.dart#L959) and [showMenu](https://github.com/flutter/flutter/blob/72321c82f99852685c5f9a25a44adbadb7d020ff/packages/flutter/lib/src/material/popup_menu.dart#L982) for inspiration.

packages/flutter/test/material/bottom_sheet_test.dart
```diff
@@ -3047,6 +3047,47 @@ void main() {
...
+    );
+
+    await tester.tap(find.text('X'));
+    await tester.pump();
+
+    final route = observer.lastRoute! as ModalBottomSheetRoute;
+
+    final proxy = route.animation! as ProxyAnimation;
+
+    final curved = proxy.parent! as CurvedAnimation;
```
victorsanni: Not a big fan of using type casts to regression test instead of testing the actual widget behavior (e.g that the animation follows the passed-in curve rather than a default, etc.).

Mairramer: I find it a bit confusing how tests for these scenarios are written. Is there an easier way to do this?

victorsanni: I wonder if we can monitor the details of the animation. Maybe use a simple curve like Curves.linear and verify that the sheet is at a certain non-default height halfway through the animation (if the height of the sheet is what is animating) etc.

Mairramer: done

packages/flutter/test/material/bottom_sheet_test.dart
```diff
@@ -3047,6 +3047,74 @@ void main() {
...
+        ),
+      ),
+    );
+
+    await tester.tap(find.text('X'));
+    await tester.pump();
+
+    // Advance the animation by 1/3 of the duration.
+    await tester.pump(const Duration(milliseconds: 100));
+    expect(tester.getTopLeft(find.byKey(sheetKey)).dy, closeTo(547.3, 0.1));
```
victorsanni: I'm scared of these magic numbers because they make the test prone to break (and difficult to debug) with the tiniest changes to the underlying implementation. I wonder if we can use a predictable curve such as `Curves.linear` and somehow express the sheet extent as a function of both the animation duration and the sheet's initial extent. 

For example, at 100ms it's at x, at 200ms it's at 2x, etc. That way even if the sheet's internal implementation changes, the test still reflects the relationship between the provided curve and the sheet's extension value.

Mairramer: Yes, exactly, that’s why I don’t like testing animations 😄 . I just followed other existing examples.

victorsanni: This is a test I got working for `curve`:

<details>

```dart

  testWidgets('ModalBottomSheet uses AnimationStyle curve', (WidgetTester tester) async {
    final Key sheetKey = UniqueKey();

    await tester.pumpWidget(
      MaterialApp(
        home: Scaffold(
          body: Builder(
            builder: (BuildContext context) {
              return GestureDetector(
                onTap: () {
                  showModalBottomSheet<void>(
                    context: context,
                    sheetAnimationStyle: const AnimationStyle(curve: Curves.linear),
                    builder: (BuildContext context) {
                      return SizedBox.expand(
                        key: sheetKey,
                        child: FilledButton(
                          onPressed: () {
                            Navigator.pop(context);
                          },
                          child: const Text('Close'),
                        ),
                      );
                    },
                  );
                },
                child: const Text('X'),
              );
            },
          ),
        ),
      ),
    );

    await tester.tap(find.text('X'));
    await tester.pump();

    // Advance the animation by 50ms.
    await tester.pump(const Duration(milliseconds: 50));
    final double openExtent1 = tester.getTopLeft(find.byKey(sheetKey)).dy;

    // Advance the animation by an additional 50ms.
    await tester.pump(const Duration(milliseconds: 50));
    final double openExtent2 = tester.getTopLeft(find.byKey(sheetKey)).dy;

    // Advance the animation by an additional 50ms.
    await tester.pump(const Duration(milliseconds: 50));
    final double openExtent3 = tester.getTopLeft(find.byKey(sheetKey)).dy;

    // For the linear curve, the distance covered in each time interval should
    // be the same.
    expect(openExtent1 - openExtent2, closeTo(openExtent2 - openExtent3, 0.1));
    await tester.pumpAndSettle();

    // Dismiss the bottom sheet.
    await tester.tap(find.widgetWithText(FilledButton, 'Close'));
    await tester.pumpAndSettle();
  });

```

</details>

Can you add a test for `reverseCurve` as well?

dkwingsmt: I recommend using `Curves.decelerate` instead, which is predicatble for being simply a `t^2` and also being able to distinguish between `t` and output.

dkwingsmt (approved):
LGTM except for Victor's comment.

victorsanni (approved):

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -1080,6 +1083,18 @@ class ModalBottomSheetRoute<T> extends PopupRoute<T> {
     return _animationController!;
   }
 
+  @override
+  Animation<double> createAnimation() {
+    if (sheetAnimationStyle != AnimationStyle.noAnimation) {
```
victorsanni: Google testing failure seems legit, the `Scaffold` background color is different from its previous value while the bottom sheet is animating. This happens when `showModalBottomSheet` doesn't set a `sheetAnimationStyle`.

Is previous behavior respected if no `sheetAnimationStyle` is provided?

Mairramer: Perhaps this is due to the removal of `.chain(CurveTween(curve: barrierCurve))`. I'll add it again.

Mairramer: This is probably because, after the removal, the resulting colors were:
```
Color.fromRGBO(0, 0, 0, 0.4059)
Color.fromRGBO(0, 0, 0, 0.2706)
Color.fromRGBO(0, 0, 0, 0.1353)
```

Without the removal, the colors were:
```
Color.fromRGBO(0, 0, 0, 0.5198)
Color.fromRGBO(0, 0, 0, 0.4340)
Color.fromRGBO(0, 0, 0, 0.2216)
```

victorsanni: Can we add a test for this?

Mairramer: I reverted to `.chain(CurveTween(curve: barrierCurve))` and validated it locally. Google Testing should probably pass.

victorsanni: Not a blocker for this PR, but it is scary that we don't have a test for this behavior.

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -1023,8 +1023,11 @@ class ModalBottomSheetRoute<T> extends PopupRoute<T> {
 
   final ValueNotifier<EdgeInsets> _clipDetailsNotifier = ValueNotifier<EdgeInsets>(EdgeInsets.zero);
 
+  CurvedAnimation? _animation;
```
victorsanni: `_ModalBottomSheetState.handleDragEnd` uses the const default curve, should it use the provided curves?

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -1080,6 +1083,18 @@ class ModalBottomSheetRoute<T> extends PopupRoute<T> {
     return _animationController!;
   }
 
+  @override
+  Animation<double> createAnimation() {
+    if (sheetAnimationStyle != AnimationStyle.noAnimation) {
+      return _animation ??= CurvedAnimation(
+        parent: super.createAnimation(),
+        curve: sheetAnimationStyle?.curve ?? _modalBottomSheetCurve,
+        reverseCurve: sheetAnimationStyle?.reverseCurve ?? _modalBottomSheetCurve,
```
victorsanni: Should the fallback here be `_modalBottomSheetCurve` or `_modalBottomSheetCurve.reverse`?

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -1080,6 +1083,18 @@ class ModalBottomSheetRoute<T> extends PopupRoute<T> {
     return _animationController!;
   }
 
+  @override
+  Animation<double> createAnimation() {
+    if (sheetAnimationStyle != AnimationStyle.noAnimation) {
+      return _animation ??= CurvedAnimation(
+        parent: super.createAnimation(),
+        curve: sheetAnimationStyle?.curve ?? _modalBottomSheetCurve,
```
victorsanni: Nit: Conventionally our constants have a `_k` prefix, so this should be `_kModalBottomSheetCurve`. But I see this is the convention in this file so no pressure.

victorsanni (commented):
Google testing is still failing, I will rerun again. The images look like only the barrier color is changing, not the vertical extent of the bottom sheet.

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: I think so.

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: IMHO, the behavior feels a bit odd. Unless someone explicitly wants this in the future, I don’t see a strong reason to change it.

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: The default animation is `Curves.linear`, right? I think we can use that.

packages/flutter/lib/src/material/bottom_sheet.dart
victorsanni: > the behavior feels a bit odd.

Can you explain further here?

My assumption is the sheet should animate with the provided animation when the drag ends. Currently it just uses its default curve for all these animations, but now that this PR overrides the default with a user-provided curve I'm wondering if that can cause discrepancies with the sheet animating one way in a certain state and a different way in another state.

packages/flutter/lib/src/material/bottom_sheet.dart
victorsanni: > The default animation is Curves.linear

I can't find this default, can you point out where this is set?

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: I understand what you mean. Should it be reverseCurve in this case?

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: I might be mistaken, but the Material documentation mentions Emphasized decelerate. So we could use `Easing.legacyDecelerate` as the default.

victorsanni: I don't think we want to change the defaults here, so it would be nice to confirm within the code what the current default is for both forward and reverse animations and ensure that this PR falls back to those defaults in all states.

Mairramer: Based on the Material documentation:
https://m3.material.io/styles/motion/easing-and-duration/applying-easing-and-duration
 and
https://m3.material.io/styles/motion/easing-and-duration/tokens-specs

The token for Enter the screen is `md.sys.motion.easing.emphasized.decelerate`. However, Flutter currently uses `md.sys.motion.easing.legacy.decelerate`.

From what I can see, these tokens were introduced in Flutter in https://github.com/flutter/flutter/issues/116525.

In Material 2, the documentation describes `Decelerated easing (cubic-bezier(0.0, 0.0, 0.2, 1)),` which corresponds to what Flutter currently exposes as `legacyDecelerate`.

Given that, using `Easing.legacyDecelerate `as the default is correct.

Mairramer: Let me know if I missed anything.

victorsanni: Out of scope of this PR, but we should change to the M3 spec in the future.

Mairramer: > Out of scope of this PR, but we should change to the M3 spec in the future.

Should we file an issue for this?

victorsanni (commented):
Hi @Mairramer, sorry for the delay here. I have dug deeper into the `showModalBottomSheet` implementation and I think we can land this change by localizing the application of the `sheetAnimationStyle.curve/reverseCurve`.

Before this PR, `ModalBottomSheetRoute` used the default `super.createAnimation`, which creates an un-curved basic `AnimationController` with only the `duration/reverseDuration` specified (from `sheetAnimationStyle.duration/reverseDuration`). Making this animation curved will affect other children animations (e.g barrier color animations) since Flutter will now stack the child curves on top of the `createAnimation` curve. 

What we want to do here instead is to apply the curves provided by `sheetAnimationStyle.curve/reverseCurve` only to the `ModalBottomSheet` (and not the barrier, etc). We can do this by supplying `sheetAnimationStyle` into the `_ModalBottomSheet` returned in `ModalBottomSheetRoute.buildPage`. Then, in `_ModalBottomSheet`, we can just replace instances of `_modalBottomSheetCurve` with `sheetAnimationStyle?.curve ?? _modalBottomSheetCurve`.

victorsanni (changes requested):
Requesting changes.

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -702,12 +718,17 @@ class _ModalBottomSheetState<T> extends State<_ModalBottomSheet<T>> {
...
   void handleDragStart(DragStartDetails details) {
     // Allow the bottom sheet to track the user's finger accurately.
-    animationCurve = Curves.linear;
+    _sheetAnimation.parent = widget.route.animation;
   }
 
   void handleDragEnd(DragEndDetails details, {bool? isClosing}) {
-    // Allow the bottom sheet to animate smoothly from its current position.
-    animationCurve = Split(widget.route.animation!.value, endCurve: _modalBottomSheetCurve);
+    _curvedSheetAnimation = CurvedAnimation(
```
victorsanni: Can we abstract away this repeating logic in a function?

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -675,13 +676,28 @@ class _ModalBottomSheet<T> extends StatefulWidget {
...
+  final AnimationStyle? animationStyle;
 
   @override
   _ModalBottomSheetState<T> createState() => _ModalBottomSheetState<T>();
 }
 
 class _ModalBottomSheetState<T> extends State<_ModalBottomSheet<T>> {
-  ParametricCurve<double> animationCurve = _modalBottomSheetCurve;
+  late final ProxyAnimation _sheetAnimation;
+  late Animation<double> _curvedSheetAnimation;
```
victorsanni: Do we need to dispose any of the above?

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -702,12 +718,17 @@ class _ModalBottomSheetState<T> extends State<_ModalBottomSheet<T>> {
...
   void handleDragEnd(DragEndDetails details, {bool? isClosing}) {
-    // Allow the bottom sheet to animate smoothly from its current position.
-    animationCurve = Split(widget.route.animation!.value, endCurve: _modalBottomSheetCurve);
+    _curvedSheetAnimation = CurvedAnimation(
+      parent: widget.route.animation!,
+      curve: widget.animationStyle?.curve ?? _kModalBottomSheetCurve,
+      reverseCurve: widget.animationStyle?.reverseCurve ?? _kModalBottomSheetCurve,
+    );
+
+    _sheetAnimation.parent = _curvedSheetAnimation;
```
victorsanni, line 742: We are no longer using a `Split` curve? Can we still track the user's finger and animate smoothly?

Mairramer: Yes, since I removed the `transform` from inside the `builder`.

victorsanni (approved):
Thanks for the work and the patience here!

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -675,13 +676,34 @@ class _ModalBottomSheet<T> extends StatefulWidget {
...
   _ModalBottomSheetState<T> createState() => _ModalBottomSheetState<T>();
 }
 
 class _ModalBottomSheetState<T> extends State<_ModalBottomSheet<T>> {
-  ParametricCurve<double> animationCurve = _modalBottomSheetCurve;
+  late final ProxyAnimation _sheetAnimation;
+  late final CurvedAnimation _curvedSheetAnimation;
+
+  @override
+  void initState() {
```
chunhtai, line 690: didUpdateWidget may also need to be updated

packages/flutter/test/material/bottom_sheet_test.dart
```diff
@@ -3047,6 +3047,168 @@ void main() {
 
     semantics.dispose();
   });
+
+  testWidgets('ModalBottomSheet uses AnimationStyle curve', (WidgetTester tester) async {
```
chunhtai, line 3051: we should also add a test that change the animation curve during the life time of modalBottomSheet

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: I’m not sure whether `didUpdateWidget` is necessary in this case.
As far as I know, a `ModalBottomSheet` does not receive configuration updates once the route is shown, either during its presentation or afterward. Please let me know if I’m missing something.

chunhtai: oh didn't realize this is a private class, in that case can you add an assert just to be safe?
```dart
didUpdateWidget(oldWidget) {
  super.didUpdateWidget(oldWidget);
  assert(oldWidget.route == widget.route);
  assert(_curvedSheetAnimation.curve == widget.animationStyle?.curve ?? _kModalBottomSheetCurve);
  ...
}
```

In case some one change the code in the future and the assumption suddenly not held

packages/flutter/test/material/bottom_sheet_test.dart
chunhtai: nvm, this is no logically possible

packages/flutter/lib/src/material/bottom_sheet.dart
```diff
@@ -697,6 +697,22 @@ class _ModalBottomSheetState<T> extends State<_ModalBottomSheet<T>> {
     _sheetAnimation = ProxyAnimation(_curvedSheetAnimation);
   }
 
+  @override
+  void didUpdateWidget(_ModalBottomSheet<T> oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    super.didUpdateWidget(oldWidget);
```
chunhtai, line 702: duplicated call

chunhtai (commented):
LGTM, except for the last comment

packages/flutter/lib/src/material/bottom_sheet.dart
Mairramer: ops

chunhtai (approved):

chunhtai:
looks like there is something wrong in the google test ci currently. wait a bit before retrying

PR #181420, opened 2026-01-23
[Impeller] Fix interpolation error in Rect::TransformAndClipBounds

flar:
We had been interpolating the points backwards in the `Rect::TransformAndClipBounds` code.

When we transform a rectangle under a perspective transform, some of the points may generate a homogenous bias (W) that is less than 0. This is a special condition that means that the point was transformed "behind the camera" (this is one way of looking at it and helps explain why we avoid such values). In order for it to contribute properly to the bounds in front of the "camera", it needs to be moved back into the positive (W>0) space. We do this by interpolating from that out of bounds coordinate to either of its adjacent neighbors which are in bounds (i.e. their W>0) to move the point to a bias value that is just slightly in the "in bounds" half-space - we choose W=(1/2^14) as our "in bounds" bias value.

Unfortunately, the code to interpolate the 2 coordinates was written with conflicting definitions of which direction we computed the `t` value and which direction we applied it to the 2 coordinates - thus generating nonsensical values.

This turned up while debugging a fix for https://github.com/flutter/flutter/issues/173104 where we fixed the transform logic in the clip code, but the rectangle was still partially clipped due to a bad result from the `GetCoverage` method of the rect geometry code. Note that this fix alone does not correct the problem in that Issue, it merely fixes a needed utility method that the fix will rely on.

### DIFF
--- a/engine/src/flutter/impeller/geometry/rect.h
+++ b/engine/src/flutter/impeller/geometry/rect.h
@@ -777,8 +777,8 @@ struct TRect {
     if (neighbor.z >= kMinimumHomogenous) {
       auto t = (kMinimumHomogenous - p.z) / (neighbor.z - p.z);
       clipped[index++] = {
-          (t * p.x + (1.0f - t) * neighbor.x) / kMinimumHomogenous,
-          (t * p.y + (1.0f - t) * neighbor.y) / kMinimumHomogenous,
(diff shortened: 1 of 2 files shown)


gaaclarke (approved):
lgtm

b-luk (approved):

PR #181421, opened 2026-01-23
[ Tool ] Cleanup `ResidentCompiler` initialization logic

bkonyi:
Moves the initialization logic for web and non-web compilers into the `ResidentCompiler` factory and reduces code duplication.

Also bumps the flutter_tools Dart SDK version to ^3.10.0 to enable dot shorthands.

### DIFF
--- a/packages/flutter_tools/lib/src/compile.dart
+++ b/packages/flutter_tools/lib/src/compile.dart
@@ -532,6 +532,7 @@ abstract interface class ResidentCompiler {
     String? librariesSpec;
     TargetModel targetModel = targetModelOverride ?? .flutter;
 
+    // Configure the compiler to target the DDC runtime.
     if (targetPlatform case .web_javascript) {
       sdkRoot = artifacts.getHostArtifact(HostArtifact.flutterWebSdk).path;
       targetModel = .dartdevc;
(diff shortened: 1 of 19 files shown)


packages/flutter_tools/lib/src/compile.dart
```diff
@@ -867,52 +932,48 @@ class DefaultResidentCompiler implements ResidentCompiler {
...
-          if (nativeAssetsUri != null) ...<String>['--native-assets', nativeAssetsUri],
-          if (platformDill != null) ...<String>['--platform', platformDill!],
-          if (unsafePackageSerialization) '--unsafe-package-serialization',
-          // See: https://github.com/flutter/flutter/issues/103994
-          '--verbosity=error',
-          ...?_filterExtraFrontEndOptions(extraFrontEndOptions),
-        ];
+    final command = <String>[
+      ...commandToStartFrontendServer,
+      ...<String>[
```
nshahan: Can this inner list and spread just be removed now?

packages/flutter_tools/lib/src/compile.dart
```diff
@@ -507,32 +509,89 @@ class _RejectRequest extends _CompilationRequest {
...
+    String sdkRoot = artifacts.getArtifactPath(
+      Artifact.flutterPatchedSdkPath,
+      platform: targetPlatform,
+      mode: buildInfo.mode,
+    );
+    String? platformDillPath;
+    String? librariesSpec;
+    TargetModel targetModel = targetModelOverride ?? .flutter;
+
+    if (targetPlatform case .web_javascript) {
```
nshahan, line 536: Dart2js compiles never come through here? We should comment that here or maybe introduce another platform to make it obvious?

packages/flutter_tools/test/general.shard/compile_test.dart
```diff
@@ -224,10 +219,14 @@ void main() {
...
             '--output-dill',
             'foo.dill',
+            '--packages',
+            '.dart_tool/package_config.json',
             '-Ddart.vm.profile=false',
             '-Ddart.vm.product=false',
             '--enable-asserts',
             '--track-widget-creation',
+            '--initialize-from-dill',
+            'build/72879a0a1c6924a9ff6430cb690ed03a.cache.dill.track.dill',
```
nshahan, line 229: Where did this come from?

nshahan (approved):

packages/flutter_tools/lib/src/compile.dart
bkonyi: Yeah, it's strange, and the target platform / target architecture usage throughout the tool is a bit of a mess. We have `AndroidArch` and `DarwinArch` to specify architectures for Android and iOS/MacOS and only have simple `android` and `ios` `TargetPlatform` enum values, but for desktop targets we specify multiple `TargetPlatform` values to cover all the architectures covered by a single platform.

I'll add a comment stating that this is for targeting the DDC runtime.

packages/flutter_tools/lib/src/compile.dart
bkonyi: Good call! I'll do that.

packages/flutter_tools/test/general.shard/compile_test.dart
bkonyi: We use this cached dill file when we try to launch the CFE with extra frontend options or Dart defines. If we don't provide them, we generate `build/cache.dill` or `build/cache.dill.track.dill` (if track widget creation is enabled) and pass them as `--initialize-from-dill` on subsequent runs. Otherwise, we prepend a hash of the options and defines to the generated dill file and use it on subsequent runs.

This needed to be added here due to some of the refactoring I did, since now `initializeFromDill` is being set in a different place and isn't `null` for the purposes of this test. The hash is stable though and doesn't rely on any non-deterministic state, so this is safe.

jyameo (approved):
lgtm!

bkonyi:
> @bkonyi there seems to be a performance regression as a result of this PR, but we aren't quite sure what it means. Maybe you have an idea? Can you take a look please:
> 
> https://flutter-flutter-perf.luci.app/e/?begin=1769790672&end=1770053728&keys=X9c797768dc02345bd8f1893b4796af29&num_commits=50&request_type=0&xbaroffset=2806316

Do you have steps to reproduce those results? I'm not quite sure how these results are obtained.

bkonyi:
Thanks for sending this my way @gaaclarke! Fix is up for review here: https://github.com/flutter/flutter/pull/181826.

Issue #179812, opened 2025-12-12
[Impeller] Re-enable Vulkan on Pixel 10.


chinmaygarde:
Pixel 10 devices do not use Vulkan. Instead, they fall back to using the OpenGL ES backend.

The Pixel 10 was the first device in the Pixel series to use an ImgTec GPU and some bugs were encountered. To work around this, some versions of the Pixel 10 Vulkan GPU driver would query the [self-identified Impeller Vulkan application version](https://github.com/flutter/flutter/blob/43f3f920ad1b4e990161b3def3a5765f551dc040/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc#L209) and reject the creation of a Vulkan context. Impeller would then fallback to OpenGL.

Per updates from the IHV, the version that the driver reads is the **engine** version (as opposed to the application version in the struct above) which is 1.0 today. The driver should give Impeller a Vulkan context if the version is anything higher.

This issue tracks the re-enabling of Vulkan on Pixel 10 by bumping the Vulkan engine version reported by Impeller (and linked above). The following pre-requisites need to be performed:
* On a pixel 10, the Vulkan engine version must be bumped up past 1.0 and any other additional checks gating Vulkan removed. It must be confirmed that Impeller is able to create a Vulkan context.
* The usual sample applications must render accurately.
* The reduced test cases in the internal doc must be tested.
* If testing is successful, the engine version update may be committed. This is necessary for enabling Vulkan on Pixel 10 but not sufficient to ensure stability of newer Vulkan enabled Impeller applications on Pixel 10 devices that have not yet been updated.
* Impeller must detect the driver version and self-reject Vulkan on device older than the current Pixel 10 GPU driver version. The IHV reports that a safe conservative driver version is >= 25.1 and any verification must happen on versions past this. Impeller should still base its checks on the driver version that we verify correctness on. What we have today works well and it is not advised to introduce instability for the sake of enabling Vulkan on this device.

[More context for Googlers here](https://goto.google.com/347b4dd5-4db5-4879-96fb-0bcf3ca1034b).

PR #181432, opened 2026-01-24
fix: swap app and engine version in vk::ApplicationInfo

jtmcdole:
The driver on PIxel 10 requires Flutter to set the `engine` version to something higher than 1.0. We had previous set the application version without effect. With this change, I see we can get a valid vulkan context.

- ran some testing against reported errors (saw none)
- compiled and ran wonderous with validation layers (runs)
- still researching 

towards: #179812

> On a pixel 10, the Vulkan engine version must be bumped up past 1.0 and any other additional checks gating Vulkan removed. It must be confirmed that Impeller is able to create a Vulkan context.

Pixel 10 gets a Vulkan context with just the engine version bump.

> The usual sample applications must render accurately.

wonderous and failing code passes and renders correctly

> The reduced test cases in the internal doc must be tested.

it does.

> If testing is successful, the engine version update may be committed. This is necessary for enabling Vulkan on Pixel 10 but not sufficient to ensure stability of newer Vulkan enabled Impeller applications on Pixel 10 devices that have not yet been updated.

This PR does that.

> Impeller must detect the driver version and self-reject Vulkan on device older than the current Pixel 10 GPU driver version. The IHV reports that a safe conservative driver version is >= 25.1 and any verification must happen on versions past this. Impeller should still base its checks on the driver version that we verify correctness on. What we have today works well and it is not advised to introduce instability for the sake of enabling Vulkan on this device.

Done - driver_version appears to be a build number, so monotonically increasing.

### DIFF
--- a/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc
@@ -141,6 +141,10 @@ Context::BackendType ContextVK::GetBackendType() const {
   return Context::BackendType::kVulkan;
 }
 
+/* version 2.0.0 */
+static constexpr uint32_t kImpellerEngineVersion =
+    VK_MAKE_API_VERSION(0, 2, 0, 0);
+
 void ContextVK::Setup(Settings settings) {
   TRACE_EVENT0("impeller", "ContextVK::Setup");
 
@@ -217,8 +221,7 @@ void ContextVK::Setup(Settings settings) {
   // variant, major, minor, patch
   application_info.setApplicationVersion(VK_API_VERSION_1_0);
   application_info.setApiVersion(VK_API_VERSION_1_1);
-  application_info.setEngineVersion(
-      VK_MAKE_API_VERSION(0, 2, 0, 0) /*version 2.0.0*/);
+  application_info.setEngineVersion(kImpellerEngineVersion);
   application_info.setPEngineName("Impeller");
   application_info.setPApplicationName("Impeller");
 
--- a/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc
@@ -206,19 +206,19 @@ void ContextVK::Setup(Settings settings) {
 
   vk::ApplicationInfo application_info;
 
-  // Use the same encoding macro as vulkan versions, but otherwise application
+  // Use the same encoding macro as vulkan versions, but otherwise engine
   // version is intended to be the version of the Impeller engine. This version
   // information, along with the application name below is provided to allow
   // IHVs to make optimizations and/or disable functionality based on knowledge
   // of the engine version (for example, to work around bugs). We don't tie this
   // to the overall Flutter version as that version is not yet defined when the
-  // engine is compiled. Instead we can manually bump it occassionally.
+  // engine is compiled. Instead we can manually bump it occasionally.
   //
   // variant, major, minor, patch
-  application_info.setApplicationVersion(
-      VK_MAKE_API_VERSION(0, 2, 0, 0) /*version 2.0.0*/);
+  application_info.setApplicationVersion(VK_API_VERSION_1_0);
   application_info.setApiVersion(VK_API_VERSION_1_1);
-  application_info.setEngineVersion(VK_API_VERSION_1_0);
+  application_info.setEngineVersion(
+      VK_MAKE_API_VERSION(0, 2, 0, 0) /*version 2.0.0*/);
   application_info.setPEngineName("Impeller");
   application_info.setPApplicationName("Impeller");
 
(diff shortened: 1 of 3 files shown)


engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -206,19 +206,19 @@ void ContextVK::Setup(Settings settings) {
...
+  // engine is compiled. Instead we can manually bump it occasionally.
   //
   // variant, major, minor, patch
-  application_info.setApplicationVersion(
-      VK_MAKE_API_VERSION(0, 2, 0, 0) /*version 2.0.0*/);
+  application_info.setApplicationVersion(VK_API_VERSION_1_0);
   application_info.setApiVersion(VK_API_VERSION_1_1);
-  application_info.setEngineVersion(VK_API_VERSION_1_0);
+  application_info.setEngineVersion(
+      VK_MAKE_API_VERSION(0, 2, 0, 0) /*version 2.0.0*/);
```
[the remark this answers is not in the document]
jtmcdole: fine, you got me.

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
```diff
@@ -117,6 +121,9 @@ constexpr std::array<std::pair<std::string_view, PowerVRGPU>, 6> kGpuSeriesMap =
     }};
 }  // namespace
 
+const uint32_t kPixel10DeviceID = 0x71061212;
```
gaaclarke, line 127: Add comment where this comes from.

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
```diff
@@ -117,6 +121,9 @@ constexpr std::array<std::pair<std::string_view, PowerVRGPU>, 6> kGpuSeriesMap =
...
+const uint32_t kPixel10MinDriverVersion = 6794074;  // Corresponds to 25.1
```
gaaclarke, line 132: How does that correspond to 25.1?

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
```diff
@@ -358,6 +368,19 @@ bool DriverInfoVK::IsEmulator() const {
 }
 
 bool DriverInfoVK::IsKnownBadDriver() const {
+#if FML_OS_ANDROID
+  // Pixel 10 is identified by the PowerVR vendor and device ID 0x71061212.
+  // The driver version 25.1 or greater is required - which is device specific
+  // to 6794074 or greater.
+  if (vendor_ == VendorVK::kPowerVR && device_id_ == kPixel10DeviceID &&
+      driver_version_ < kPixel10MinDriverVersion) {
```
gaaclarke, line 383: Are these number sequential?  Can we do a less than on them?

gaaclarke (commented):
lgtm, just a few documentation questions

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
jtmcdole: done

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
jtmcdole: This looks like its an always increasing number from CI/CD.  The raw numbers are vendor specific, but they look increasing on khronos vulkan db. There are no public verifications that this is a good value to test - nor are Chrome, ANGEL, Dawn, and others doing more than device id / vendor id testing.

I am more comfortable removing the driver version test since Android is the one blocking us from getting a Vulkan context - but I believe Chinmay was concerned that they added that before they had all the fixes in.

I am not comfortable parsing "25.1@6794074" from the Vulkan call `vkGetPhysicalDeviceProperties2` `driverInfo` - especially since there is no assurance they will always follow x.y format.

https://vulkan.gpuinfo.org/listreports.php?devicename=Google+Pixel+10+Pro+XL&platform=android
https://vulkan.gpuinfo.org/listreports.php?devicename=Google+Pixel+10&platform=android
https://vulkan.gpuinfo.org/listreports.php?devicename=Google+Pixel+10+Pro+Fold&platform=android
https://vulkan.gpuinfo.org/listreports.php?devicename=Google+Pixel+10+Pro&platform=android

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
jtmcdole: Added a link

engine/src/flutter/impeller/renderer/backend/vulkan/driver_info_vk.cc
jtmcdole: doen

gaaclarke (approved):
lgtm, this is kind of funky but I think this is as good as we can get, thanks for updating the documentation to make it more clear.

PR #139513, opened 2023-12-04
Overlay supports unconstrained environments

goderbauer:
Fixes https://github.com/flutter/flutter/issues/137875.

Unfortunately, we cannot auto-detect which OverlayEntry should be sizing the Overlay in unconstrained environment. So, this PR adds a special flag to annotate the Overlay Entry that should be used.

PR #181490, opened 2026-01-26
Fix overlay in size to content scenario with loose constraints

knopp:
Right now overlay assumes that when sizing to contents the constraints are always unconstrained. That might not be true however, for example `constraints.biggest` may be limited to usable desktop size. If that happens the current behavior is that `Overlay` sizes itself to `constraints.biggest`, which defeats the purpose of views sized to contents.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1337,7 +1337,8 @@ class _RenderTheater extends RenderBox
 
   @override
   double? computeDryBaseline(BoxConstraints constraints, TextBaseline baseline) {
-    final Size size = constraints.isTight
+    final Size size =
+        (constraints.isTight || (!_hasSizeDeterminingChild() && constraints.biggest.isFinite))
         ? constraints.biggest
         : _findSizeDeterminingChild().getDryLayout(constraints);
     final nonPositionedChildConstraints = BoxConstraints.tight(size);
@@ -1362,7 +1363,7 @@ class _RenderTheater extends RenderBox
 
   @override
   Size computeDryLayout(BoxConstraints constraints) {
-    if (constraints.isTight) {
+    if (constraints.isTight || (!_hasSizeDeterminingChild() && constraints.biggest.isFinite)) {
       return constraints.biggest;
     }
     return _findSizeDeterminingChild().getDryLayout(constraints);
@@ -1412,7 +1413,7 @@ class _RenderTheater extends RenderBox
   @override
   void performLayout() {
     RenderBox? sizeDeterminingChild;
-    if (constraints.isTight) {
+    if (constraints.isTight || (!_hasSizeDeterminingChild() && constraints.biggest.isFinite)) {
       size = constraints.biggest;
     } else {
       sizeDeterminingChild = _findSizeDeterminingChild();
@@ -1429,6 +1430,19 @@ class _RenderTheater extends RenderBox
         layoutChild(child, nonPositionedChildConstraints);
       }
     }
+  }
+
+  bool _hasSizeDeterminingChild() {
+    RenderBox? child = _lastOnstageChild;
+    while (child != null) {
+      final childParentData = child.parentData! as _TheaterParentData;
+      if ((childParentData.overlayEntry?.canSizeOverlay ?? false) &&
+          !childParentData.isPositioned) {
+        return true;
+      }
+      child = childParentData.previousSibling;
+    }
+    return false;
   }
 
   RenderBox _findSizeDeterminingChild() {
@@ -2668,7 +2682,9 @@ class _RenderLayoutSurrogateProxyBox extends RenderProxyBox {
     // needed.
     if (!theater._layingOutSizeDeterminingChild) {
       final BoxConstraints theaterConstraints = theater.constraints;
-      final Size boxSize = theaterConstraints.isTight ? theaterConstraints.biggest : theater.size;
+      final Size boxSize = theaterConstraints.biggest.isFinite
+          ? theaterConstraints.biggest
+          : theater.size;
       deferredChild._doLayoutFrom(this, constraints: BoxConstraints.tight(boxSize));
     }
   }
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -1337,7 +1337,7 @@ class _RenderTheater extends RenderBox
 
   @override
   double? computeDryBaseline(BoxConstraints constraints, TextBaseline baseline) {
-    final Size size = constraints.biggest.isFinite
+    final Size size = constraints.isTight
         ? constraints.biggest
         : _findSizeDeterminingChild().getDryLayout(constraints);
     final nonPositionedChildConstraints = BoxConstraints.tight(size);
@@ -1362,7 +1362,7 @@ class _RenderTheater extends RenderBox
 
   @override
   Size computeDryLayout(BoxConstraints constraints) {
-    if (constraints.biggest.isFinite) {
+    if (constraints.isTight) {
       return constraints.biggest;
     }
     return _findSizeDeterminingChild().getDryLayout(constraints);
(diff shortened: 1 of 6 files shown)


packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1337,7 +1337,7 @@ class _RenderTheater extends RenderBox
 
   @override
   double? computeDryBaseline(BoxConstraints constraints, TextBaseline baseline) {
-    final Size size = constraints.biggest.isFinite
+    final Size size = constraints.isTight
```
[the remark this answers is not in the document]
knopp: That's kinda the whole point of this PR...

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1412,7 +1412,7 @@ class _RenderTheater extends RenderBox
   @override
   void performLayout() {
     RenderBox? sizeDeterminingChild;
-    if (constraints.biggest.isFinite) {
+    if (constraints.isTight) {
```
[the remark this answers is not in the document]
knopp: Yes, the original logic would do it and it breaks content sized views. So the logic needs to change. Will it break something else that assumes current behavior? Possibly. That's what we need to find out.

packages/flutter/test/material/scaffold_test.dart
```diff
@@ -1645,27 +1645,22 @@ void main() {
...
+            child: Overlay.wrap(child: child),
+          ),
         );
       }
 
       final Widget child = boilerplate(
         Scaffold(
           resizeToAvoidBottomInset: false,
           body: const Placeholder(),
-          bottomNavigationBar: Navigator(
```
knopp, line 1656: I assume originally Navigator was inserted here because the test failed because BottomNavigationBar requires overlay in hierarchy (for tooltips), but that seems to be wrong way to fix it. Instead the overlay was added in the `boilderplate`.

packages/flutter/test/material/scaffold_test.dart
```diff
@@ -1688,7 +1683,9 @@ void main() {
         ),
       );
       final Offset finalPoint = tester.getCenter(find.byType(Placeholder));
-      expect(initialPoint, finalPoint);
+      // BottomNavigationBar adds 20 pixel padding so the center point moves
+      // by 10 pixels.
+      expect(initialPoint, finalPoint.translate(0, 10));
```
knopp, line 1688: Because there is no `Navigator` now above `BottomNavigationBar`, the `BottomNavigationBar` always inserts bottom paddings from MediaQuery and we need to account for that.

packages/flutter/test/material/search_anchor_test.dart
```diff
@@ -2622,24 +2622,24 @@ void main() {
     await tester.pumpWidget(
       MaterialApp(
         builder: (BuildContext context, Widget? child) {
-          return Scaffold(
-            body: Padding(padding: const EdgeInsets.all(rootSpacing), child: child),
-          );
+          return Padding(padding: const EdgeInsets.all(rootSpacing), child: child);
```
knopp, line 2625: Putting `Scaffold` in the `builder` and thus above `Navigator` seems like a bad idea. I'm not sure why it was done, but it causes issues because `Scaffold` makes the constraints loose, which means the overlay in `Navigator` will try to size itself to child.

packages/flutter/test/material/search_anchor_test.dart
```diff
@@ -2622,24 +2622,24 @@ void main() {
...
-                icon: const Icon(Icons.search),
-                onPressed: () {
-                  controller.openView();
-                },
-              );
-            },
-            suggestionsBuilder: (BuildContext context, SearchController controller) {
-              return <Widget>[];
-            },
+        home: Scaffold(
```
knopp, line 2627: I've put the `Scaffold` in the `home` instead of above `Navigator` which seems right the place for it.

packages/flutter/test/material/time_picker_test.dart
```diff
@@ -2718,33 +2718,31 @@ Future<void> mediaQueryBoilerplate(
             size: tester.view.physicalSize / tester.view.devicePixelRatio,
           ),
           child: Material(
-            child: Center(
```
knopp, line 2721: Putting center above `Navigator` will cause the `Navigator` and thus the `Overlay` to size to content, which is not what the tests expect.

packages/flutter/test/widgets/text_golden_test.dart
```diff
@@ -519,32 +519,27 @@ void main() {
 
   testWidgets('Text Inline widget textfield', (WidgetTester tester) async {
     await tester.pumpWidget(
-      Center(
```
knopp, line 522: This Center widget doesn't seem to be doing anything useful. It just loosens the constraints causing the rest of the tree to size to content.

packages/flutter/test/widgets/text_golden_test.dart
```diff
@@ -561,104 +556,102 @@ void main() {
   // This tests if multiple Text.rich widgets are able to inline nest within each other.
   testWidgets('Text Inline widget nesting', (WidgetTester tester) async {
     await tester.pumpWidget(
-      Center(
```
knopp, line 564: Again, unecessary Center widget on top loosens constraints causing the rest to size to contents.

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1337,7 +1337,8 @@ class _RenderTheater extends RenderBox
 
   @override
   double? computeDryBaseline(BoxConstraints constraints, TextBaseline baseline) {
-    final Size size = constraints.biggest.isFinite
+    final Size size =
+        (constraints.isTight || (!_hasSizeDeterminingChild() && constraints.biggest.isFinite))
```
mattkae, line 1341: I would wrap this check into a method since it is reused in multiple places, e.g.:

```dart
bool shouldUseBiggestConstraints(BoxConstraints constraints)
```

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -1431,6 +1432,19 @@ class _RenderTheater extends RenderBox
     }
   }
 
+  bool _hasSizeDeterminingChild() {
+    RenderBox? child = _lastOnstageChild;
+    while (child != null) {
+      final childParentData = child.parentData! as _TheaterParentData;
```
mattkae, line 1438: Could this ever meaningfully fail? If so, we may want to check for it.

mattkae (approved):
Approving just to see what effect this has on Google testing

packages/flutter/lib/src/widgets/overlay.dart
knopp: The logic is lifted from `_findSizeDeterminingChild`.  I don't think it can fail.

justinmc (commented):
Quickly looking through a few different Google test failures:

 * Error message: `Looking up a deactivated widget's ancestor is unsafe.` 
 https://github.com/flutter/flutter/blob/2c74ab162e6e6604acabcda314ac8c9491e26023/packages/flutter/lib/src/widgets/framework.dart#L5050

 * Golden failure: Subtle gradient color changes, as if for antialiasing or something. Weird.
 * Golden failure: Constraint changes causing stuff to get huge. I specifically see things like images and icons ending up being huge and broken. Perhaps they are inside of Overlays but it's hard to tell without digging into the code more.

LongCatIsLooong (commented):
Could you provide a concrete use case for this PR? 

IIRC https://github.com/flutter/flutter/pull/139513 wasn't meant for making Overlay "sized by child", the goal was really making sure the widget doesn't blow up when given unbound constraints (and the developer has to make sure the "size determining child" has an intrinsic size). Without more context it's hard to determine whether using a size-determining child is still the best approach.

Also as you mentioned the current implementation changes the behavior if the incoming constraints are loose. It's not uncommon for people to add Overlays to the widget tree so I think the failures might be from these custom Overlays.

knopp:
Closing in favor of https://github.com/flutter/flutter/pull/182009.

PR #181507, opened 2026-01-26
[native_assets] Fix `flutter build ios-framework`

### LINKED ISSUE #181724: [native assets] Changing build modes bundles old assets
Also, the stripping does not change correctly when switching between build modes. If you do

`flutter build ios --debug` 
`otool -l /path/to/my_app/build/native_assets/ios/my_ffi_plugin.framework/my_ffi_plugin` -> LC_DYSYMTAB nlocalsym > 1 as expected

then do

`flutter build ios --release` 
`otool -l /path/to/my_app/build/native_assets/ios/my_ffi_plugin.framework/my_ffi_plugin` -> LC_DYSYMTAB nlocalsym == 1 as expected

then do

`flutter build ios --debug` 
`otool -l /path/to/my_app/build/native_assets/ios/my_ffi_plugin.framework/my_ffi_plugin` -> LC_DYSYMTAB nlocalsym is NOT > 1 as expected ❌

_Originally posted by @vashworth in https://github.com/flutter/flutter/issues/181533#issuecomment-3819778409_

The root cause is a `Target` writing to a directory outside `.dart_tool/flutter_build/<hash>/`, while the `Target`s assume that if their inputs don't change their outputs don't have to be rewritten. This is most likely also an issue for code assets for all OSes.

```
// The second `flutter build ios --debug -v`:
           [        ] Skipping target: dart_build
           [        ] Skipping target: install_code_assets
```

And, Android has the same issue:

```
flutter build apk --profile -v
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 13:55 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/.dart_tool/hooks_runner/shared/my_package/build/16a46ad652892aace39acf5234912532/libmy_package.so
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 13:55 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/build/native_assets/android/jniLibs/lib/arm64-v8a/libmy_package.so

flutter build apk --debug -v
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 14:00 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/.dart_tool/hooks_runner/shared/my_package/build/48df710ce179958291528f1ec25cad14/libmy_package.so
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 14:00 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/build/native_assets/android/jniLibs/lib/arm64-v8a/libmy_package.so

flutter build apk --profile -v
[        ] [   +2 ms] Skipping target: install_code_assets
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 13:55 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/.dart_tool/hooks_runner/shared/my_package/build/16a46ad652892aace39acf5234912532/libmy_package.so
-rwxr-xr-x@ 1 dacoharkes  primarygroup   5.2K Jan 30 14:00 /Users/dacoharkes/src/dacoharkes/playground/my_package/example/build/native_assets/android/jniLibs/lib/arm64-v8a/libmy_package.so
```

Initial analysis on how the flutter framework is correctly shipped for the various build-modes (and target SDKs):

* MacOS/Android - has debug/release/profile in the directory structure.
* iOS - has debug/release/profile and device/simulator in the directory structure.
* Linux/Windows - the `Target` is forcably executed overwriting the files.

=> apparently we care less about caching for Linux/Windows. The caching behavior is worse.

We can model the code assets in the same way by using nested directories for iOS/MacOS/Linux and force rerun for Linux/Windows.

dcharkes:
This PR fixes two issues. Accidental reuse of code assets between build modes and SDKs (https://github.com/flutter/flutter/issues/181724), and the bundling in ios-frameworks (https://github.com/flutter/flutter/issues/181382).

To fix the accidental caching, the `Target`s related to build hooks and code assets now output their files to `environment.outputDir` instead of `$projectDir/$buildDir/native_assets`.

* `xcode_backend` is updated to deal with this.
* `Flutter.kt` has been updated to deal with this.
* Because the `Target`s are responsible for caching, the code has been refactored to provide the target directories from there. The "global-ish" function `nativeAssetsBuildUri` that was calculating the directory before has been removed.
* `runFlutterSpecificHooks` has nothing to do with that directory, it's access to it has been removed.
* To avoid another cmakefile migration, the Linux and Windows implementation use the same directory. (Note that output dir and build dir overlap for Linux and Windows, while they do not for MacOS, iOS, and Android.)
* This also means that we don't have to read `NativeAssetsManifest.json` in `xcode_backend` anymore. Instead the `Target` clears the output directory, so we should not have any stale frameworks.
* Refactored `installCodeAssets` and its platform-specific implementations to return a list of all produced files. These are now added to the `Target`'s depfile. This fixes an issue where the build system would skip re-installing native assets after an Xcode "Clean Build Folder because it wasn't tracking the frameworks/dylibs as outputs.

Closes: https://github.com/flutter/flutter/issues/181724

Other `Target`s related tweaks:

* Added proper `Source.pattern('{BUILD_DIR}/${DartBuild.dartHookResultFilename}'),` for all `Target`s that depend on that file. These were missing. (The build system uses `dependencies` for ordering of `Target`s, but relies on `inputs` and `outputs` for caching.)
* Removed code assets from `CopyAssets`. That target is supposed to make an asset-bundle that is OS-independent if I understand correctly.

This PR changes the way code assets are bundled in `flutter build ios-framework`.

* This PR now packages in an `.xcframework`, which is necessary to be able to package both device and simulator.
* Run through the frameworks of both device and simulator and give errors on inconsistencies.

Closes: https://github.com/flutter/flutter/issues/181382

Other iOS related tweaks:

* Use `xcrun` for invoking all the commands. (This is used for producing the app framework, but was not for code assets frameworks.)
* Make sure all commands are added to the traces when running verbose. (Also to bring it in line for with the other `xcrun` commands.)

Testing:

* The integration test is updated to inspect the `xcframework`s.
* Added a test that simulates Xcode "Product > Clean Build Folder...", to check that it now correctly triggers a rebuild.
* We do _not_ have an integration test that _runs_ the frameworks output from `flutter build ios-framework` inside a host app at all as far as I'm aware.
  * dev/devicelab/bin/tasks/build_ios_framework_module_test.dart builds a framework, but doesn't run it in a host app
  * dev/integration_tests/ios_add2app_life_cycle/build_and_test.sh runs, but does so via `flutter build ios` not as a framework.
* Does not add an integration test for caching behavior between switching build modes. However, the proper functioning of `flutter build ios-framework` depends on the `Target`s for different not using overlapping directories.

Architectural approaches tried but didn't work:

* Subclass `InstallCodeAssets` per OS to be more precise in the `output`s on what files are output. This doesn't work because other OS-independent targets on the `InstallCodeAssets` target.

### DIFF
--- a/packages/flutter_tools/bin/xcode_backend.dart
+++ b/packages/flutter_tools/bin/xcode_backend.dart
@@ -2,7 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:convert';
 import 'dart:io';
 
 void main(List<String> arguments) {
@@ -102,8 +101,6 @@ class Context {
   }
 
   Directory directoryFromPath(String path) => Directory(path);
-
-  File fileFromPath(String path) => File(path);
 
   /// Run given command ([bin]) in a synchronous subprocess.
   ///
@@ -345,72 +342,30 @@ class Context {
     // current build.
     final String builtProductsDir = environment['BUILT_PRODUCTS_DIR']!;
     final nativeAssetsPath = '$builtProductsDir/native_assets/';
+    final Directory nativeAssetsDir = directoryFromPath(nativeAssetsPath);
     final bool verbose = (environment['VERBOSE_SCRIPT_LOGGING'] ?? '').isNotEmpty;
-
-    final Set<String> referencedFrameworks = {};
-    final appResourcesDir = platform == TargetPlatform.macos ? 'Resources/' : '';
-    final File nativeAssetsJson = fileFromPath(
-      '$xcodeFrameworksDir/App.framework/${appResourcesDir}flutter_assets/NativeAssetsManifest.json',
-    );
-    if (!nativeAssetsJson.existsSync()) {
+    if (!nativeAssetsDir.existsSync()) {
       if (verbose) {
-        print("♦ No native assets to bundle. ${nativeAssetsJson.path} doesn't exist.");
+        print("♦ No native assets to bundle. $nativeAssetsPath doesn't exist.");
       }
       return;
     }
-    // NativeAssetsManifest.json looks like this: {
-    //   "format-version":[1,0,0],
-    //   "native-assets":{
-    //     "ios_arm64":{
-    //       "package:sqlite3/src/ffi/libsqlite3.g.dart":[
-    //         "absolute",
-    //         "sqlite3arm64ios.framework/sqlite3arm64ios"
-    //       ]
-    //     }
-    //   }
-    // }
-    //
-    // Note that this format is also parsed and expected in
-    // engine/src/flutter/assets/native_assets.cc
-    try {
-      final nativeAssetsSpec = json.decode(nativeAssetsJson.readAsStringSync()) as Map;
-      for (final Object? perPlatform
-          in (nativeAssetsSpec['native-assets'] as Map<String, Object?>).values) {
-        for (final Object? asset in (perPlatform! as Map<String, Object?>).values) {
-          if (asset case ['absolute', final String frameworkPath]) {
-            // frameworkPath is usually something like sqlite3arm64ios.framework/sqlite3arm64ios
-            final [String directory, String name] = frameworkPath.split('/');
-            if (directory != '$name.framework') {
-              throw Exception(
-                'Unexpected framework path: $frameworkPath. Should be $name.framework/$name',
-              );
-            }
 
-            referencedFrameworks.add(name);
-          }
-        }
-      }
-    } on Object catch (e, stackTrace) {
-      echo(e.toString());
-      echo(stackTrace.toString());
-      echoXcodeError('Failed to embed native assets: $e');
-      exitApp(-1);
-    }
+    final Iterable<String> frameworks = nativeAssetsDir
+        .listSync()
+        .whereType<Directory>()
+        .where((Directory d) => !d.path.endsWith('.dSYM'))
+        .map(_parseFrameworkNameFromDirectory)
+        .whereType<String>();
 
     if (verbose) {
-      print('♦ Copying native assets ${referencedFrameworks.join(', ')} from $nativeAssetsPath.');
+      print('♦ Copying native assets ${frameworks.join(', ')} from $nativeAssetsPath.');
     }
 
-    for (final framework in referencedFrameworks) {
+    for (final framework in frameworks) {
       final Directory frameworkDirectory = directoryFromPath(
         '$nativeAssetsPath$framework.framework',
       );
-      if (!frameworkDirectory.existsSync()) {
-        throw Exception(
-          'The native assets specification at ${nativeAssetsJson.path} references $framework, '
-          'which was not found in $nativeAssetsPath.',
-        );
-      }
 
       runRsync(frameworkDirectory.path, xcodeFrameworksDir);
       if (codesign && expandedCodeSignIdentity != null) {
@@ -728,3 +683,21 @@ class Context {
 }
 
 enum TargetPlatform { ios, macos }
+
+String? _parseFrameworkNameFromDirectory(Directory dir) {
+  final List<String> pathSegments = dir.uri.pathSegments;
+  if (pathSegments.isEmpty) {
+    return null;
+  }
+  final String basename;
+  if (pathSegments.last.isEmpty && pathSegments.length > 1) {
+    basename = pathSegments[pathSegments.length - 2];
+  } else {
+    basename = pathSegments.last;
+  }
+  final int extensionIndex = basename.indexOf('.framework');
+  if (extensionIndex == -1) {
+    return null;
+  }
+  return basename.substring(0, extensionIndex);
+}
--- a/packages/flutter_tools/bin/xcode_backend.dart
+++ b/packages/flutter_tools/bin/xcode_backend.dart
@@ -343,22 +343,8 @@ class Context {
   }) {
     // Copy native assets referenced in the native_assets.json file for the
     // current build.
-    final String sourceRoot = environment['SOURCE_ROOT'] ?? '';
-    var projectPath = '$sourceRoot/..';
-    if (environment['FLUTTER_APPLICATION_PATH'] != null) {
-      projectPath = environment['FLUTTER_APPLICATION_PATH']!;
-    }
-    final String flutterBuildDir = environment['FLUTTER_BUILD_DIR']!;
-    var nativeAssetsPath = '$projectPath/$flutterBuildDir/native_assets/${platform.name}/';
-    final String? builtProductsDir = environment['BUILT_PRODUCTS_DIR'];
-    if (builtProductsDir != null) {
-      final Directory nativeAssetsInBuiltProducts = directoryFromPath(
-        '$builtProductsDir/native_assets/',
-      );
-      if (nativeAssetsInBuiltProducts.existsSync()) {
-        nativeAssetsPath = '${nativeAssetsInBuiltProducts.path}/';
-      }
-    }
+    final String builtProductsDir = environment['BUILT_PRODUCTS_DIR']!;
+    final nativeAssetsPath = '$builtProductsDir/native_assets/';
     final bool verbose = (environment['VERBOSE_SCRIPT_LOGGING'] ?? '').isNotEmpty;
 
     final Set<String> referencedFrameworks = {};
--- a/packages/flutter_tools/bin/xcode_backend.dart
+++ b/packages/flutter_tools/bin/xcode_backend.dart
@@ -349,7 +349,16 @@ class Context {
       projectPath = environment['FLUTTER_APPLICATION_PATH']!;
     }
     final String flutterBuildDir = environment['FLUTTER_BUILD_DIR']!;
-    final nativeAssetsPath = '$projectPath/$flutterBuildDir/native_assets/${platform.name}/';
+    var nativeAssetsPath = '$projectPath/$flutterBuildDir/native_assets/${platform.name}/';
+    final String? builtProductsDir = environment['BUILT_PRODUCTS_DIR'];
+    if (builtProductsDir != null) {
+      final Directory nativeAssetsInBuiltProducts = directoryFromPath(
+        '$builtProductsDir/native_assets/',
+      );
+      if (nativeAssetsInBuiltProducts.existsSync()) {
+        nativeAssetsPath = '${nativeAssetsInBuiltProducts.path}/';
+      }
+    }
     final bool verbose = (environment['VERBOSE_SCRIPT_LOGGING'] ?? '').isNotEmpty;
 
     final Set<String> referencedFrameworks = {};
--- a/packages/flutter_tools/lib/src/build_system/targets/native_assets.dart
+++ b/packages/flutter_tools/lib/src/build_system/targets/native_assets.dart
@@ -80,7 +80,8 @@ class DartBuild extends Target {
     Source.pattern(
       '{FLUTTER_ROOT}/packages/flutter_tools/lib/src/build_system/targets/native_assets.dart',
     ),
-    // If different packages are resolved, different native assets might need to be built.
+    // If different packages are resolved, different native assets might need to
+    // be built.
     Source.pattern('{WORKSPACE_DIR}/.dart_tool/package_config.json'),
     // TODO(mosuem): Should consume resources.json. https://github.com/flutter/flutter/issues/146263
   ];
@@ -140,6 +141,14 @@ class InstallCodeAssets extends Target {
     // And install/copy the code assets to the right place and create a
     // native_asset.yaml that can be used by the final AOT compilation.
     final Uri nativeAssetsFileUri = environment.buildDir.childFile(nativeAssetsFilename).uri;
+
+    Uri targetUri = environment.outputDir.childDirectory('native_assets').uri;
+    final String osName = targetPlatform.osName;
+    if (osName == 'linux' || osName == 'windows') {
+      // Avoid needing migration for CMake files, keep old directory structure.
+      targetUri = targetUri.resolve('$osName/');
+    }
+
     await installCodeAssets(
       dartHookResult: dartHookResult,
       environmentDefines: environment.defines,
@@ -147,6 +156,7 @@ class InstallCodeAssets extends Target {
       projectUri: projectUri,
       fileSystem: fileSystem,
       nativeAssetsFileUri: nativeAssetsFileUri,
+      targetUri: targetUri,
     );
     assert(await fileSystem.file(nativeAssetsFileUri).exists());
 
@@ -172,7 +182,10 @@ class InstallCodeAssets extends Target {
     Source.pattern(
       '{FLUTTER_ROOT}/packages/flutter_tools/lib/src/build_system/targets/native_assets.dart',
     ),
-    // If different packages are resolved, different native assets might need to be built.
+    Source.pattern('{BUILD_DIR}/${DartBuild.dartHookResultFilename}'),
+    // If different packages are resolved, different native assets might need to
+    // be built. We can't depend on the exact outputs from `DartBuild`, so
+    // depend on all the same inputs.
     Source.pattern('{WORKSPACE_DIR}/.dart_tool/package_config.json'),
   ];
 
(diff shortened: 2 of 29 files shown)


bkonyi (approved):
LGTM, but let's wait for input from @vashworth to confirm the iOS project logic.

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -318,27 +318,46 @@ class BuildIOSFrameworkCommand extends BuildFrameworkCommand {
...
+          .whereType<Directory>();
+      if (frameworkDirectoriesDevice.length != frameworkDirectoriesSimulator.length) {
+        throwToolExit(
+          'Did not find the same frameworks for device and simulator.\n'
+          '$frameworkDirectoriesDevice $frameworkDirectoriesSimulator',
+        );
+      }
+      for (final frameworkDirectorySimulator in frameworkDirectoriesSimulator) {
+        final Directory frameworkDirectoryDevice = nativeAssetDeviceDirectory.childDirectory(
+          frameworkDirectorySimulator.basename,
```
vashworth: I don't think they are guaranteed to have the same name. For example, I've seen one that has them named as `sqlite3arm64ios_sim.framework` and `sqlite3arm64ios.framework` [using build hooks](https://github.com/simolus3/sqlite3.dart/blob/a6ebb6eb471a49bb84e1077092e648f4a631f143/sqlite3/lib/src/hook/assets.dart#L96).

vashworth: You might consider getting the `NativeAssetsManifest.json` from the build App.framework and then parsing the assets out of it. https://github.com/flutter/flutter/pull/179251 did something similar.

That's what I was planning to do for a new SwiftPM version of this command. See prototype:
https://github.com/vashworth/flutter/blob/2a62d6647f8917938d0bd3150c5ae32f20a984bb/packages/flutter_tools/lib/src/commands/build_swift_packages.dart#L671
https://github.com/vashworth/flutter/blob/2a62d6647f8917938d0bd3150c5ae32f20a984bb/packages/flutter_tools/lib/src/commands/build_swift_packages.dart#L725-L771

vashworth (commented):
You should consider waiting until https://github.com/flutter/flutter/pull/181533 is complete, since the dSYM should also be included and would be good to verify it's correctly being placed in the generated xcframeworks.

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
HyperlinkKishanr: @dcharkes any plan for merge this PR? I didn't a get chance to verify the fix.

dcharkes: @vashworth The fix for https://github.com/flutter/flutter/issues/181724 means we will only have frameworks for the right build mode and SDK in the directory.

@HyperlinkKishanr Yes, but I uncovered more bugs and refactorings I had to land first.

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -318,27 +318,42 @@ class BuildIOSFrameworkCommand extends BuildFrameworkCommand {
...
-          '--filter',
-          '- native_assets.json',
-          nativeAssetsDirectory.path,
-          modeDirectory.path,
-        ]);
-        if (rsyncResult.exitCode != 0) {
-          throwToolExit('Failed to copy native assets:\n${rsyncResult.stderr}');
-        }
+      // Package native assets.
+      final Directory nativeAssetSimulatorDirectory = simulatorBuildOutput.childDirectory(
```
bkonyi: It feels like we're duplicating a lot of the logic here, with the main difference being the directory we're working on.

Maybe we can do something like this instead?

```dart
List<String> findFrameworkNames(Directory root) => root
    .childDirectory('native_assets')
    .listSync()
    .whereType<Directory>()
    .where((d) => !d.basename.endsWith('.dSYM'))
    .map((d) => d.basename);

final frameworkNames = <String>[
  ...findFrameworkNames(simulatorBuildOutput),
  ...findFrameworkNames(iPhoneBuildOutput),
];
```

bkonyi: And maybe this should be somewhere shared, since we use this logic in `build_macos_framework.dart` as well.

bkonyi (approved):
LGTM once verified by @vashworth

packages/flutter_tools/bin/xcode_backend.dart
vashworth: Can this be simplified to now just copy from the `nativeAssetsInBuiltProducts`?

packages/flutter_tools/bin/xcode_backend.dart
```diff
@@ -482,7 +482,16 @@ class Context {
       projectPath = environment['FLUTTER_APPLICATION_PATH']!;
     }
     final String flutterBuildDir = environment['FLUTTER_BUILD_DIR']!;
-    final nativeAssetsPath = '$projectPath/$flutterBuildDir/native_assets/${platform.name}/';
+    var nativeAssetsPath = '$projectPath/$flutterBuildDir/native_assets/${platform.name}/';
```
vashworth: In what situation would this be the correct path? Seems like we're falling back to it, but why?

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -146,6 +146,18 @@ abstract class BuildFrameworkCommand extends BuildSubCommand {
     }
   }
 
+  static Iterable<String> findFrameworkNames(Directory outputDirectory) {
```
vashworth:
```suggestion
  static Iterable<String> findNativeAssetFrameworkNames(Directory outputDirectory) {
```

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -318,27 +330,28 @@ class BuildIOSFrameworkCommand extends BuildFrameworkCommand {
...
-        ]);
-        if (rsyncResult.exitCode != 0) {
-          throwToolExit('Failed to copy native assets:\n${rsyncResult.stderr}');
-        }
+      // Package native assets.
+      final Iterable<String> frameworkNames = <String>{
+        ...BuildFrameworkCommand.findFrameworkNames(simulatorBuildOutput),
+        ...BuildFrameworkCommand.findFrameworkNames(iPhoneBuildOutput),
+      };
+      for (final frameworkName in frameworkNames) {
```
vashworth, line 424: This still is a problem when it doesn't have the same name per build mode/arch/device. See original comment: https://github.com/flutter/flutter/pull/181507#discussion_r2739601663

dcharkes: I don't think so. Below we check `existsSync`, and otherwise we don't add it.

And the native assets mapping in the `NativeAssetsManifest.json` is different
for device and simulator:

* my_package/example/build/ios/framework/Release/App.xcframework/ios-arm64/App.framework/flutter_assets/NativeAssetsManifest.json
* my_package/example/build/ios/framework/Release/App.xcframework/ios-arm64_x86_64-simulator/App.framework/flutter_assets/NativeAssetsManifest.json

So the result would be a different framework for device and simulator and then
the different framework being loaded when run on the different one.

And, different names per architecture are overwritten to the first name found in fatAssetTargetLocationsIOS. (And then also the dylib names are rewritten as well with `setInstallNamesDylib`.)

Different OSes are different builds, so they shouldn't influence this.

(It could be worth it to add an integration test that does all kinds of wonky stuff instead of relying on `flutter create --template package_ffi` which is very regular.)

packages/flutter_tools/bin/xcode_backend.dart
dcharkes: Oh nice catch, that was while I was migrating targets one by one. Cleaned up!

packages/flutter_tools/bin/xcode_backend.dart
dcharkes: Yes, provided we empty the directory in the `Target` so we don't see any frameworks that were removed. Done!

dcharkes: FYI @simolus3

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
vashworth: So this is a still a problem, though, because if you have a separate xcframework for simulators vs physical phones and you embed it in your app (following our instructions https://docs.flutter.dev/add-to-app/ios/project-setup), it will error.

For example, if I use the `sqlite3` plugin, it'll build `sqlite3arm64ios.xcframework` and `sqlite3x64ios_sim.xcframework`. If I embed both in my iOS app and build for a physical iOS device, it will fail with this error because the `sqlite3x64ios_sim.xcframework` does not support physical iOS devices:
Screenshot 2026-02-08 at 10 41 02 PM

vashworth: Really it'd be best if they had the same name. However, we might be able to wrap them in the same xcframework and name the xcframework with the physical device framework's name, since that's the one that will actually be included in an archive for the App Store. 

For example,

- sqlite3arm64ios.xcframework
  - ios-arm64
    - sqlite3arm64ios.framework
  - ios-arm64_x86_64-simulator
    - sqlite3x64ios_sim.framework
    
 I tested it against App Store validation and it passed so I think it's okay

dcharkes: Thanks @vashworth. I didn't get it the first time that this was an Apple constraint.

> I tested it against App Store validation and it passed so I think it's okay

I am afraid Apple will tighten their validation later, and things will start failing. So, I believe it'd be better to force package authors to use the same names.

(Since we do two `flutter assemble` calls and those the renaming and code signing, and _then_ combine them into a framework we can't easily rename the simulator dylibs/frameworks. Unless we start passing in some new environment variable with the code assets manifest of for the device build.)

* I've added a hard error on this now. This will trigger bug reports for package authors to start supporting `flutter build ios-framework`

Also, we've been silently rewriting the dylib names if the different architectures have different dylib names.

* I've added a warning on this. It will still work, but it will nudge package authors to do a cleaner naming solution for all ios and macos `build`/`run` commands.

This definitely requires some documentation, so I've filed:

* https://github.com/flutter/website/pull/13047

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -146,6 +147,95 @@ abstract class BuildFrameworkCommand extends BuildSubCommand {
...
+      if (deviceAssetPath != simulatorAssetPath) {
+        throwToolExit(
+          'Consistent code asset framework names are required for '
+          'XCFramework creation.\n'
+          'The asset "$assetId" has different framework paths across '
+          'platforms:\n'
+          '  - iphoneos: $deviceAssetPath\n'
+          '  - iphonesimulator: $simulatorAssetPath\n\n'
+          'Ensure the "build.dart" hook produces consistent filenames for '
+          'all targets.',
```
vashworth: This doesn't seem actionable for the app developer. Doesn't the plugin author need to make the change?

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -146,6 +147,95 @@ abstract class BuildFrameworkCommand extends BuildSubCommand {
...
+    Directory simulatorBuildOutput,
+  ) {
+    final Map<String, String> deviceAssets = BuildFrameworkCommand.parseNativeAssetsManifest(
+      iPhoneBuildOutput,
+    );
+    final Map<String, String> simulatorAssets = BuildFrameworkCommand.parseNativeAssetsManifest(
+      simulatorBuildOutput,
+    );
+
+    for (final String assetId in simulatorAssets.keys) {
```
vashworth: If the device build is the source of truth, shouldn't we iterate through that one instead?

packages/flutter_tools/lib/src/isolated/native_assets/macos/native_assets_host.dart
```diff
@@ -330,3 +331,46 @@ Map<Architecture?, List<String>> parseOtoolArchitectureSections(String output) {
...
+    final KernelAssetPath currentPath = targetLocationCallback(asset, alreadyTakenNames).path;
+
+    if (existingPath != null && existingPath != currentPath) {
+      final String existingName = (existingPath as KernelAssetAbsolutePath).uri.pathSegments.first;
+      final String currentName = (currentPath as KernelAssetAbsolutePath).uri.pathSegments.first;
+      globals.logger.printWarning(
+        'Code asset "$assetId" has different framework names for '
+        'different architectures. Picking "$existingName" and '
+        'ignoring "$currentName". Ensuring consistent filenames in '
+        '"build.dart" is recommended.',
```
vashworth: Again, I'd make this message more actionable to the app developer. Like "Please file an issue with plugin $pluginName..."

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
```diff
@@ -146,6 +147,95 @@ abstract class BuildFrameworkCommand extends BuildSubCommand {
...
+      final String? deviceAssetPath = deviceAssets[assetId];
+      final String simulatorAssetPath = simulatorAssets[assetId]!;
+      if (deviceAssetPath == null) {
+        throwToolExit(
+          'The simulator build contains a code asset "$assetId" that is '
+          'not present in the physical device build. \n'
+          'The device build is the source of truth for distributed '
+          'frameworks. \n'
+          'Ensure "$assetId" is also built for physical devices.',
```
vashworth: This doesn't seem actionable for the app developer. Doesn't the plugin author need to make the change? Perhaps something like "Please file an issue with the $pluginName plugin."

packages/flutter_tools/lib/src/commands/build_ios_framework.dart
vashworth: Same as above, make the message actionable for the app developer

packages/flutter_tools/lib/src/build_system/targets/native_assets.dart
```diff
@@ -172,7 +182,10 @@ class InstallCodeAssets extends Target {
     Source.pattern(
       '{FLUTTER_ROOT}/packages/flutter_tools/lib/src/build_system/targets/native_assets.dart',
     ),
-    // If different packages are resolved, different native assets might need to be built.
+    Source.pattern('{BUILD_DIR}/${DartBuild.dartHookResultFilename}'),
```
vashworth, line 193: So I checked this out again. When I build an app through Xcode, then do a clean "Product > Clean Build Folder...", native assets are deleted from the build folder as I would expect. However, when I rebuild using Xcode, the native assets are not re-generated. I'm guessing because the input did not change and we don't track what this outputs really so the target gets skipped.

We likely need to update this target's outputs to have something in the `OUTPUT_DIR`, so that if it's deleted, the target will re-run. It'd be best for it to output each generated framework, but that may not be possible. Easiest would be to output like the native_assets.json to the `OUTPUT_DIR` too

packages/flutter_tools/lib/src/isolated/native_assets/macos/native_assets_host.dart
```diff
@@ -330,3 +331,46 @@ Map<Architecture?, List<String>> parseOtoolArchitectureSections(String output) {
...
+      );
```
vashworth, line 370: Warnings within `flutter assemble` don't show in Xcode unless they are prefixed with "warning: " and are sent to stderr. Perhaps use `printXcodeWarning`?

https://github.com/flutter/flutter/blob/7fa78a68f472270c6ec96ff51e380e38728f68a2/packages/flutter_tools/lib/src/build_system/targets/darwin.dart#L125

Also, make this warning more actionable for the app developer.

vashworth (approved):
LGTM once tests are passing! Sorry for all the back and forth!

aleufms:
Is this already shipped on stable version?

dcharkes:
> Is this already shipped on stable version?

It didn't make it into Flutter stable 3.41, it will be in the next Flutter stable 3.44. It should be in the Flutter beta from 3.43.

PR #181510, opened 2026-01-26
feature: implementation of tooltips in the `_TestWindowingOwner` and minor bugfixes to the multiple windows example app

mattkae:
## What's new?
- Implemented `_TestTooltipWindowController` in the `TestWindowingOwner`
- Fixed how child windows of the main window are rendered in the application
- Introduced `WindowManager.getChildrenOf` to simplify filtering and rendering of the children of a parent window
- Added a test for the tooltip window
- Added a test for modal dialogs of the main window

## Pre-launch Checklist

- [X] I signed the [CLA].


examples/multiple_windows/lib/app/models.dart
```diff
@@ -44,6 +55,10 @@ class WindowManager extends ChangeNotifier {
     _windows.removeWhere((KeyedWindow window) => window.key == key);
     notifyListeners();
   }
+
+  Iterable<KeyedWindow> getChildrenOf(BaseWindowController? parent) {
```
loic-sharma: We often use the `Of` suffix to signify a method depends on an inherited widget and can add new sources of rebuilds. I'd consider avoiding the `Of` suffix here.

Also, I see you use `getChildrenOf(null)` to get the root windows. That seems not ideal since these are top-level windows - not child windows - and it's not clear what `null` means.

What do you think of switching this `getWindows({required BaseWindowController? parent})`?

```
windowManager.getWindows(parent: null); // Get top-level windows
windowManager.getWindows(parent: myWindow); // Get child windows
```

packages/flutter_test/lib/src/binding.dart
```diff
@@ -529,6 +552,77 @@ class _TestDialogWindowController extends DialogWindowController with _ChildWind
...
+  }
+
+  final TooltipWindowControllerDelegate _delegate;
+  final _TestWindowingOwner windowingOwner;
+  BoxConstraints _constraints;
+  ui.Rect _anchorRect;
+  WindowPositioner _positioner;
+  final BaseWindowController _parent;
+  // ignore: unused_field
+  final bool _isSizedToContent;
```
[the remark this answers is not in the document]
loic-sharma: This looks like valid feedback

loic-sharma (approved):
LGTM but please also get approval from @justinmc

robert-ancell (approved):
Other than the comments from @loic-sharma this looks good, thanks!

examples/multiple_windows/lib/app/dialog_window_content.dart
```diff
@@ -82,17 +82,17 @@ class DialogWindowContent extends StatelessWidget {
...
-                  controller: window.controller,
-                  windowKey: window.key,
-                  onDestroyed: () => windowManager.remove(window.key),
-                  onError: () => windowManager.remove(window.key),
-                ),
-              );
-            }
+          for (final KeyedWindow childWindow in windowManager.getWindows(
+            parent: window,
+          )) {
```
justinmc: Nit: I would assign the result of getWindows to a variable and use that here, just to make this easier to read. Fine either way though.

justinmc (approved):
LGTM :+1:

Issue #168555, opened 2025-05-08
Use `@awaitNotRequired` in Flutter SDK


srawlins:
In a number of discussions (https://github.com/dart-lang/sdk/issues/46218, https://github.com/dart-lang/sdk/issues/58348), Flutter team members have indicated that they would use an `@awaitNotRequired` annotation if it were available, to annotate Future-returning functions that do not need to be awaited or wrapped in `unawaited()`.

In the next release of the meta package, 1.17.0, we'll be releasing `@awaitNotRequired`.

I'm just creating this issue to track adding `@awaitNotRequired` annotations to everywhere in the Flutter SDK that we wanted them. In https://github.com/dart-lang/core/issues/742, @goderbauer mentioned https://api.flutter.dev/flutter/animation/AnimationController/forward.html.

CC @Piinks

RepliedSage11:
What do you think about calling it `@unawaited` to match the function name?

srawlins:
@RepliedSage11 sorry it is called `@awaitNotRequired`. See https://github.com/dart-lang/sdk/issues/46218 for discussion.

justinmc:
I'm on board with this 👍 . If I think of any other places that could use it I'll leave a comment.

srawlins:
The annotation is now released with meta version 1.17.0.

And there is an `unnecessary_unawaited` lint rule also available in recent Dart commits that lets users know they don't need to `unawaited()` a function call for a function annotated with `@awaitNotRequired`.

victorsanni:
Hi @srawlins, some dart libraries export Future-returning functions that aren't called with `await`. If we turn on the `unawaited_futures` lint in the sdk, these callsites will get flagged. Is there a plan to add this annotation to dart libraries?

PR #160484, opened 2024-12-18
Implement golden-file matching for `integration_test` on Android and iOS devices

matanlurey:
Work towards https://github.com/flutter/flutter/issues/143299.
Work towards https://github.com/flutter/flutter/issues/160043.

---

This PR implements, end-to-end, support for `matchesGoldenFile` when (a) running with `package:integration_test` (b) on a device, such as an Android emulator, Android device, iOS simulator, or iOS device, where the _runner_ of a test file does not have process and local-file system access.

There are multiple parts to this PR; I could make it smaller than 1K lines, but the bulk of that is tests, and it would mean landing PRs that are incomplete and unused, which does not seem useful - so instead here is a quick overview of the PR's contents - questions/feedback welcome, and I am willing to break code out or land incremental refactors if requested.

1. Augmented `flutter_platform.dart` (used for iOS and Android), similar to [`flutter_web_platform.dart`](https://github.com/flutter/flutter/blob/1398dc7eecb696d302e4edb19ad79901e615ed56/packages/flutter_tools/lib/src/test/flutter_web_platform.dart#L117-L128), now creates and uses [`test_golden_comparator.dart`](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/test/test_golden_comparator.dart) to proxy calls (coming from the VM service protocol) for golden-file updates and comparisons to a `flutter_tester` process. A full explanation of how (or why) it works this way is too hard to include here, but see https://github.com/flutter/flutter/pull/160215 for more details.
1. Added `VmServiceProxyGoldenFileComparator`, which is a currently unused (outside of a single e2e test) comparator that forwards calls to `compare` and `update` to the VM service protocol (of which, the other side of this is implemented above, in `flutter_platform.dart`. The idea is that this comparator would be used automatically when running in an integration test on a device that requires it (similar to how web works today), but that is **not** wired up yet and requires additional work in `flutter_tools`.
1. Added two unit tests (of both the client and server), and a full e2e-test using it to run `matchesGoldenFile`.

PR #100657, opened 2022-03-23
Fix crashes when current route parsing transactions are discarded

chunhtai:
There doesn't seem to be a way to create generic future that can be downward cast to any type. This pr refactor the logic around discarding parsing transaction

fixes https://github.com/flutter/flutter/issues/100225

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #59803, opened 2020-06-18
Add benchmark for Mouse region (web)

dkwingsmt:
## Description

This PR adds two benchmarks for web, testing its performance on handling a large number of mouse regions:
- `bench_mouse_region_grid_scroll`, where a large number of non-trivial mouse regions are rendered and scrolled (https://github.com/flutter/flutter/issues/41194)
- `bench_mouse_region_grid_hover`, where heavily nested mouse regions are rendered and hovered.

## Related Issues

- Tests https://github.com/flutter/flutter/issues/41194

## Tests

I added the following tests:

- N/A
## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I signed the [CLA].
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] All existing and new tests are passing.
- [x] The analyzer (`flutter analyze --flutter-repo`) does not report any problems on my PR.
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Did any tests fail when you ran them? Please read [Handling breaking changes].

- [x] No, no existing tests failed, so this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Test Coverage]: https://github.com/flutter/flutter/wiki/Test-coverage-for-package%3Aflutter
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Handling breaking changes]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes

PR #181513, opened 2026-01-26
Add @awaitNotRequired annotation to flutter sdk

victorsanni:
Part of [Use @awaitNotRequired in Flutter SDK](https://github.com/flutter/flutter/issues/168555)

Turned on the `unawaited_futures` lint and added the annotation to every function declared in the flutter SDK and flagged by the linter. This means the annotation is only added to `Future`-returning functions called at least once without the `await` keyword preceding it. So if there exists a `Future`-returning function in the Flutter SDK that does not need to be `await`ed, but all call sites in the SDK have the `await` keyword, it is not annotated. 

**EDIT:** Opened PRs to add `await`s or `ignore`s to callsites instead of adding the annotation. Some functions may still require this annotation, but this set is much smaller than originally anticipated.


packages/flutter/lib/src/widgets/router.dart
```diff
@@ -821,6 +821,7 @@ class _RouterState<T> extends State<Router<T>> with RestorationMixin {
     };
   }
 
+  @awaitNotRequired
   Future<void> _rebuild([void value]) {
```
dkwingsmt: It seems like the returned future of this private method is never used. Why can't it simply return void?

packages/flutter/test/material/elevated_button_test.dart
```diff
@@ -1994,6 +1994,7 @@ void main() {
     expect(material.textStyle!.color, Colors.white);
   });
 
+  @awaitNotRequired
   Future<void> testStatesController(Widget? icon, WidgetTester tester) async {
```
dkwingsmt, line 1994: I suspect that this should not have `@awaitNotRequired`, but instead should have their usages prefixed with `await`, since those tests _should_ wait for this function to finish.

Similarly for other instances of `testStatesController`.

packages/flutter/lib/src/painting/image_provider.dart
```diff
@@ -481,7 +481,8 @@ abstract class ImageProvider<T extends Object> {
       }
       if (!didError) {
         didError = true;
-        errorCallback(obtainedKey, exception, stack);
+        // TODO(victorsanni): Annotate errorCallback typedef with @awaitNotRequired.
+        errorCallback.call(obtainedKey, exception, stack);
```
dkwingsmt: await here and mark `handleError` as `@awaitNotRequired`. I also propose the following doc to `handleError`:
```dart
    // `handleError` does not need to be awaited because `errorCallback` will
    // always eventually set the stream.completer, and the only thing external
    // callers care about is when the stream completes.
```

packages/flutter/lib/src/services/platform_channel.dart
```diff
@@ -237,6 +237,7 @@ class BasicMessageChannel<T> {
   ///
   /// Returns a [Future] which completes to the received response, which may
   /// be null.
+  @awaitNotRequired
   Future<T?> send(T message) async {
```
dkwingsmt, line 240: Should we really mark `send` and `invokeMethod` as `awaitNotRequired`? IMO these foundational methods should require await and let individual usages suppress the warning explicitly.

packages/flutter/lib/src/services/platform_views.dart
```diff
@@ -1611,6 +1612,7 @@ abstract class PlatformViewController {
   bool get awaitingCreation => false;
 
   /// Dispatches the `event` to the platform view.
+  @awaitNotRequired
```
dkwingsmt: Why is this needed? I can't find a usage without await in the repo.

packages/flutter_driver/lib/src/common/handler_factory.dart
```diff
@@ -500,6 +500,7 @@ mixin CommandHandlerFactory {
   }
 
   /// Runs `finder` repeatedly until it finds one or more [Element]s.
+  @awaitNotRequired
```
dkwingsmt: It feels like exactly where `await` should be enforced.

packages/flutter_driver/lib/src/driver/driver.dart
```diff
@@ -213,6 +214,7 @@ abstract class FlutterDriver {
   }
 
   /// Taps at the center of the widget located by [finder].
+  @awaitNotRequired
```
dkwingsmt: I think await should definitely be enforced here.

packages/flutter_driver/lib/src/driver/driver.dart
```diff
@@ -226,6 +228,7 @@ abstract class FlutterDriver {
   ///
   ///  * [FlutterDriver.runUnsynchronized], which will execute an action
   ///    with frame sync disabled even while frames are pending.
+  @awaitNotRequired
```
dkwingsmt: I think await should definitely be enforced here.

packages/flutter_driver/lib/src/driver/timeline_summary.dart
```diff
@@ -390,6 +391,7 @@ class TimelineSummary {
   /// See also:
   ///
   /// * [Timeline.fromJson], which explains detail about the timeline data.
+  @awaitNotRequired
   Future<void> writeTimelineToFile(
```
dkwingsmt, line 393: I think await should probably be enforced here, otherwise the test might end before the file write is complete.

packages/flutter_test/lib/src/goldens.dart
```diff
@@ -82,6 +82,7 @@ abstract class GoldenFileComparator {
   ///
   /// The method by which [golden] is located and by which its bytes are written
   /// is left up to the implementation class.
+  @awaitNotRequired
   Future<void> update(Uri golden, Uint8List imageBytes);
```
dkwingsmt, line 85: I feel like this should enforce `await`. Can you check where it is used?

packages/flutter_test/lib/src/test_pointer.dart
```diff
@@ -521,6 +522,7 @@ class TestGesture {
   }
 
   /// In a test, send a pointer remove event for this pointer.
+  @awaitNotRequired
   Future<void> removePointer({Duration timeStamp = Duration.zero, Offset? location}) {
```
dkwingsmt, line 524: This should definitely require an await.

packages/flutter_test/lib/src/test_pointer.dart
```diff
@@ -576,6 +578,7 @@ class TestGesture {
 
   /// End the gesture by releasing the pointer. For trackpad pointers this
   /// will send a panZoomEnd event instead of an up event.
+  @awaitNotRequired
   Future<void> up({Duration timeStamp = Duration.zero, FlutterView? view}) {
```
dkwingsmt, line 579: This should definitely require an await.

packages/flutter_test/lib/src/test_text_input.dart
```diff
@@ -228,6 +228,7 @@ class TestTextInput {
   /// If this is used to inject an action when there is a real IME connection,
   /// for example when using the [integration_test] library, there is a risk
   /// that the real IME will become confused as to the current state of input.
+  @awaitNotRequired
```
dkwingsmt: I see only one place where this isn't awaited. Can you check if this is really needed to be optional?

packages/flutter_test/lib/src/widget_tester.dart
```diff
@@ -828,6 +828,7 @@ class WidgetTester extends WidgetController implements HitTestDispatcher, Ticker
   ///
   /// * Expose a [Future] in your application code that signals the readiness of
   ///   your widget tree, then await that future inside [callback].
+  @awaitNotRequired
```
dkwingsmt: I think this probably should enforce `await`.

victorsanni:
Thanks for the review @dkwingsmt. I didn't think there are situations in the framework where some `Future`-returning methods do not have a preceding `await` but should. This is a valid point and makes sense to me.

Is this review exhaustive? i.e are the methods you've pointed out the only ones that should have `await` added instead of the annotation? Asking because I want to go ahead and open PRs to add `await` to the ones you have pointed out.

packages/flutter/lib/src/painting/image_provider.dart
victorsanni: Got it, will do in separate PR.

packages/flutter/lib/src/services/platform_channel.dart
victorsanni: I agree, there are other callsites in the framework with `await`. Should we suppress the warning with an `ignore` or add `await`s?

packages/flutter/lib/src/services/platform_views.dart
victorsanni: There is in a test: https://github.com/flutter/flutter/blob/6ac6d3ec0f4fdac87f2c783e2d603d960ec832e8/packages/flutter/test/services/platform_views_test.dart#L481

Should we just add `await` to this test?

packages/flutter/lib/src/widgets/router.dart
victorsanni: True, i think it was initially added as a `Future<void>` to satisfy the `.then` API, but no longer needed to after https://github.com/flutter/flutter/pull/100657.

packages/flutter/test/material/elevated_button_test.dart
victorsanni: removing annotation, will add `await`s in separate PR.

packages/flutter_driver/lib/src/common/handler_factory.dart
victorsanni: Removing annotation, will add `await`s in separate PR.

There are also callsites with `await`. Probably a sign it should be enforced (or intentionally ignored) as needed.

packages/flutter_driver/lib/src/driver/driver.dart
victorsanni: Removing annotation, will add awaits in separate PR.

packages/flutter_driver/lib/src/driver/driver.dart
victorsanni: Removing annotation, will add awaits in separate PR.

packages/flutter_driver/lib/src/driver/timeline_summary.dart
victorsanni: Removing annotation, will add awaits in separate PR.

packages/flutter_test/lib/src/goldens.dart
victorsanni: Agree, most of its callsites use `await`, its non-`await` callsites are primarily from this test file in this PR: https://github.com/flutter/flutter/pull/160484.

I can go ahead and add `await` to these callsites.

packages/flutter_test/lib/src/test_pointer.dart
victorsanni: Removing annotation, will add `await`s in separate PR.

packages/flutter_test/lib/src/test_pointer.dart
victorsanni: Removing annotation, will add `await`s in separate PR.

packages/flutter_test/lib/src/test_text_input.dart
victorsanni: It is not, I agree. Removing annotation and adding `await` to that callsite.

packages/flutter_test/lib/src/widget_tester.dart
victorsanni: Agreed, removing annotation and adding `await` to callsites.

packages/flutter/lib/src/services/platform_channel.dart
dkwingsmt: I suspect many such usages are intentional, probably because those `send` do not wait for responses. For such cases it's ok to explicitly `ignore` while commenting that it's intentional to not wait for response. LMK if there are other cases that you're unsure of.

packages/flutter/lib/src/services/platform_views.dart
dkwingsmt: Likely yes. Can you try adding `await` here and see if it works?

dev/benchmarks/macrobenchmarks/lib/src/web/bench_mouse_region_grid_hover.dart
```diff
@@ -161,6 +162,7 @@ class _Tester {
     await _UntilNextFrame.wait();
   }
 
+  @awaitNotRequired
   Future<void> start() async {
```
victorsanni, line 163: Hi @dkwingsmt, I see you added this method (and its other clones below) in https://github.com/flutter/flutter/pull/59803, did you intend to use it in the postFrameCallback without `await`?

dev/benchmarks/macrobenchmarks/lib/src/animation_with_microtasks.dart
```diff
@@ -50,6 +51,7 @@ class _ChunkedWork {
     _canceled = true;
   }
 
+  @awaitNotRequired
```
victorsanni: I think we should keep the annotation here. This PR added it: https://github.com/flutter/flutter/pull/79163? WDYT @dkwingsmt

dev/benchmarks/macrobenchmarks/lib/src/web/bench_mouse_region_grid_hover.dart
dkwingsmt: I think `awaitNotRequired` is ok here, since this is only used in a fully managed test that I expect no errors to occur, and I'm ok with errors being thrown.

dev/benchmarks/macrobenchmarks/lib/src/animation_with_microtasks.dart
dkwingsmt: Yeah, I think having no await is the intention here.

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -461,6 +461,7 @@ class AnimationController extends Animation<double>
   /// During the animation, [status] is reported as [AnimationStatus.forward],
   /// which switches to [AnimationStatus.completed] when [upperBound] is
   /// reached at the end of the animation.
+  @awaitNotRequired
   TickerFuture forward({double? from}) {
```
victorsanni, line 465: Add comment to TickerFuture impl:

```A TickerFuture should probably NOT be awaited at all, since it is risky to hang forever. Instead, people who wish to know when/whether it completes will explicitly attach a callback anyway. This should make it reasonable to make mark all TickerFuture as awaitNotRequired.```

dkwingsmt (dismissed):
LGTM. Great work!

dkwingsmt:
This PR is affected by code freeze. I've suggested Victor to split this PR into two parts, landing the non-material/cupertino part while holding the remaining until the freeze is lifted.

victorsanni:
I moved the code-freezed changes to a separate PR. So this can be merged cleanly.

srawlins (dismissed):

dev/benchmarks/macrobenchmarks/lib/src/web/bench_mouse_region_grid_hover.dart
chunhtai: If this is not public facing API, we should just add unawaited at the call site

dev/benchmarks/macrobenchmarks/lib/src/animation_with_microtasks.dart
chunhtai: or apply unawaited to the call site

chunhtai (dismissed):
LGTM, but for internal facing api, we should use unawaited at the call site

dkwingsmt (approved):

Issue #178529, opened 2025-11-14
Flutter 3.38.1 - AppLocalizations are getting deleted every time on web in debug mode


vasilevzhivko:
Hey,
I have generate: true in my pubspec.yaml and everything was working perfectly until I updated my flutter version today.
For some reason Applocalizations are getting deleted once my app starts.
I am able to later generate them again, but it doesnt matter, they keep disappearing.

flutter doctor
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.38.1, on macOS 15.6.1 24G90 darwin-arm64, locale en-BG)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 16.0)
[✓] Connected device (2 available)
[✓] Network resources
```

```
arb-dir: lib/l10n
template-arb-file: app_en.arb
output-localization-file: app_localizations.dart
output-class: AppLocalizations
nullable-getter: false
output-dir: lib/l10n/gen_l10n
```

flutter gen-l10n -v
```
[  +48 ms] executing: sysctl hw.optional.arm64
[   +4 ms] Exit code 0 from: sysctl hw.optional.arm64
[        ] hw.optional.arm64: 1
[   +6 ms] Artifact Instance of 'MaterialFonts' is not required, skipping update.
[        ] Artifact Instance of 'GradleWrapper' is not required, skipping update.
[        ] Artifact Instance of 'AndroidGenSnapshotArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'AndroidInternalBuildArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'IOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[        ] Artifact Instance of 'LegacyCanvasKitRemover' is not required, skipping update.
[        ] Artifact Instance of 'FlutterSdk' is not required, skipping update.
[        ] Artifact Instance of 'WindowsEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'MacOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'LinuxEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'LinuxFuchsiaSDKArtifacts' is not required, skipping update.
[...]

PR #174685, opened 2025-08-29
Add data assets

mosuem:
Refiling of #169273 (reverted in https://github.com/flutter/flutter/pull/170034), which is a refiling of #164094, which itself is a rebase of #159675.

This PR adds bundling support for the experimental dart data asset feature: Dart packages with hooks can now emit data assets which the flutter tool will bundle.

It relies on flutter's existing asset bundling mechanism (e.g. entries in AssetManifest.json, DevFS syncing in reload/restart, ...).

The support is added under an experimental flag (similar to the existing native assets experimental flag).

Also, kNativeAssets is removed to also bundle data assets on flutter build bundle.

The chrome sandbox is disabled as per #165664.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[...]

PR #181542, opened 2026-01-27
[hooks] Don't run build hooks for code assets in `flutter run`

dcharkes:
Since https://github.com/flutter/flutter/pull/174685, the code assets are wrongly invoked on in the `flutter run` process (in addition to rightly in the `flutter assemble` process). This should not be the case: We don't know the target architectures we want to build for, neither do we know which native compiler is set by the native build system that is invoking us (https://github.com/flutter/flutter/pull/181004#pullrequestreview-3664863551).

This PR changes the way the hooks are invoked:

* From `flutter run` only run for data assets. (Unblocks https://github.com/flutter/flutter/pull/181004)
* All other remaining calls, run for both. There might be locations where data assets could be disabled, but they are needed from the `DartBuild` target at least in some cases.

The architecture becomes as follows:

* `FlutterNativeAssetsBuildRunner` this is basically the wrapper around the `NativeAssetsBuildRunner` and there should be only one in a `flutter_tools` instance, unchanged.
* `FlutterHookRunner` seems to be an interface to be able to supply fakes, unchanged.
* `runFlutterSpecificHooks` get bool arguments whether they should build code assets and data assets.
  * The callers of these APIs know what asset types are needed in that context. The invocations added in https://github.com/flutter/flutter/pull/174685 should be data assets only.
* Simplification: `FlutterHookRunnerNative.runHooks` does no longer use `globals.buildSystem.build`. Instead, it directly calls `runFlutterSpecificHooks`.
  * This completely avoids writing to the flutter build directory, which was the cause of https://github.com/flutter/flutter/issues/178529.
* The `ProtocolExtension`s (which determine which asset types are built) are taken from `AssetBuildTarget`s, and take into account what asset types to build, unchanged.

Tests:

* Code assets covered by existing integration tests such as packages/flutter_tools/test/integration.shard/isolated/native_assets_test.dart
* Data assets covered by existing integration tests such as packages/flutter_tools/test/integration.shard/isolated/dart_data_asset_test.dart
* This PR completely deletes packages/flutter_tools/test/general.shard/build_system/targets/hook_runner_native_test.dart. It was a unit test that testsed the specific workaround, not the effect of the workaround.
  * I've manually tested the steps in https://github.com/flutter/flutter/issues/178529#issuecomment-3542651915, the issue does not come back.
  
> the underlying issue did involve conflict between state written by the original run of the build system versus state written by the secondary run executed for the data assets.  If the secondary run can be avoided, then that seems cleaner.

Yep, removed.


jason-simmons (commented):
LGTM

bkonyi (approved):

PR #181550, opened 2026-01-27
Updates GetDartSize to return the number of non-padding bytes

walley892:
Updates GetDartSize to return the number of non-padding bytes. We used to send uniform information to the FragmentShader code that included the logical size of the uniforms (no padding), after https://github.com/flutter/flutter/pull/181340, we started to send the physical size of the uniforms (including padding).

This caused us to mis-represent the size of the uniforms in the FragmentShader API, and possibly caused other issues. 

This PR updates GetDartSize to return the number of non-padding bytes (old behavior), and re-enables tests for metal that verify this behavior.

### DIFF
--- a/engine/src/flutter/impeller/core/runtime_types.cc
+++ b/engine/src/flutter/impeller/core/runtime_types.cc
@@ -9,10 +9,16 @@
 namespace impeller {
 
 size_t RuntimeUniformDescription::GetDartSize() const {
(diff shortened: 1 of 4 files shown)


gaaclarke (changes requested):
Can we incorporate the reland of https://github.com/flutter/flutter/pull/181340 with this pr since it is fixing it?

PR #178809, opened 2025-11-19
Allow SliverMainAxisGroup children to remain pinned

manu-sncf:
Fix #178808.

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [X] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[Data Driven Fixes]: https://github.com/flutter/flutter/blob/main/docs/contributing/Data-driven-Fixes.md

PR #175671, opened 2025-09-19
Fix SliverMainAxisGroup SliverEnsureSemantics support

manu-sncf:
To complete the work in PR #166889, this PR ensures that `SliverMainAxisGroup` correctly supports `SliverEnsureSemantics` and respects `cacheExtent` rules within its `visitChildrenForSemantics` implementation.

The fix is adapted from `RenderViewPort.visitChildrenForSemantics` code.

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [X] All existing and new tests are passing.

PR #174369, opened 2025-08-25
Fix SliverMainAxisGroup scrollOffsetCorrection

manu-sncf:
Fixes : [#174368](https://github.com/flutter/flutter/issues/174368)

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/main/docs/contributing/Chat.md
[...]

Issue #18345, opened 2018-06-10
SliverAppBar flexibleSpace with height depending on children


adriancmurray:
I have a SliverAppBar with a flexibleSpace that holds information that will change in height. I simply want my AppBar to determine it's height at runtime rather than the static expandedHeight option. Do I really need to calculate the height of this widget before I pass it to the flexibleSpace? Or is there a built in method to allow the SliverAppBar to determine that on it's own?

noordawod:
Have you managed to achieve this somehow, @adriancmurray?

adriancmurray:
@noordawod I did a complete redesign of my app since I wanted to implement this feature so I ended up not needing it.

noordawod:
Alright, cool, thanks.

vinicioslc:
I would like if there was an option to set a fixed height for the AppBar.

zoechi:
Did you try to get an answer on StackOverflow?

vinicioslc:
Yes, but when I'm using a SliverAppBar, I can not set the height of it to a fixed size, or to use SliverAppBar's children's (title) size.

And the PreferredSize used inside a Sliver does not force the child (SliverAppBar) to stick to the height I want.

javierforero:
@vinicioslc @noordawod did you guys figure out away to do this? I have a text widget that could be very long. I am setting it as the background so I need the `expandableHeight` to be whatever the height the text widget is.

noordawod:
No, unfortunately.

vinicioslc:
@javierforero @noordawod im using this  package for now https://pub.dev/packages/flutter_sticky_header and its highly customizable

vinicioslc:
@javierforero see my comment

singlakaran:
@vinicioslc   https://pub.dev/packages/flutter_sticky_header 
this library use Flutter SDK version >=1.5.9-pre.94 <2.0.0, but i don't want to downgrade version. so, what can i do?

noordawod:
@singlakaran 

Learn about dependency overrides: https://dart.dev/tools/pub/pubspec#dependencies

scopendo:
I've been able to get something that works for me by using SliverAppBar with a combination FlexibleSpaceBar with title padding and a bottom widget to set the overall height.

```
SliverAppBar(
      pinned: true,
      flexibleSpace: FlexibleSpaceBar(
        titlePadding: EdgeInsets.only(top: 70),
        title: Column(
          children: <Widget>[
[...]

PR #181569, opened 2026-01-27
POC NestedSlivers

zemanux:
Related to #178809 and #18345.

This PR explores a new approach by introducing the `NestedSlivers` widget (note: this name is provisional and open to suggestions).

The core characteristic of this widget is that it leverages `ViewportBase`'s `layoutChildSequence` algorithm via a new mixin, `SliverSequenceMixin` (introduced here as a basic version for this POC).

By design, this unifies the construction of multiple nested slivers, offering several key advantages:

* **Robustness:** We benefit from the `Viewport`'s proven and robust algorithm.
* **Consistency & Maintenance:** Any future modifications to the `Viewport`'s behavior regarding slivers will naturally prompt developers to consider applying those changes here as well. This is particularly relevant for issues such as #174369, #175671, and #175760. I also want to highlight `SliverPaintOrder`, which is not currently handled but could be aligned.
* **Simplification:** I was able to significantly simplify the implementation of `SliverMainAxisGroup`. I achieved this by wrapping `NestedSlivers` with a simple "modifier" (`SliverUnpinned`) to readjust the non-persistent header behavior. This adheres to the principle of [aggressive composability](https://docs.flutter.dev/resources/inside-flutter#aggressive-composability). The behavior remains strictly identical to the previous version (verified via `sliver_main_axis_group_test.dart`), but with improved overlap support (fixing a bug related to `SliverAppBar` overscroll).

**master `SliverMainAxisGroup`**

https://github.com/user-attachments/assets/47536308-169f-4a78-adf9-1c3347d101c8

**`SliverMainAxisGroup` with `NesteSlivers` and `SliverUnpinned`**

https://github.com/user-attachments/assets/75e5cdd8-13ac-4611-8165-d08ee4e72633

<details><summary>Code sample overscroll demo</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
  Widget build(BuildContext context) => MaterialApp(
    title: 'NestedSlivers Demo',
    theme: ThemeData(primarySwatch: Colors.blue, useMaterial3: true),
    home: const Scaffold(body: NumberList()),
  );
}

class NumberList extends StatelessWidget {
  const NumberList({super.key});

  @override
  Widget build(BuildContext context) => CustomScrollView(
    physics: const BouncingScrollPhysics(),
    slivers: <Widget>[
      SliverMainAxisGroup(
        slivers: [
          SliverAppBar(
            pinned: true,
            stretch: true,
            expandedHeight: 200,
            backgroundColor: Colors.yellow,
            flexibleSpace: FlexibleSpaceBar(title: Text('Coucou')),
          ),
          for (int i = 1; i <= 5; i++)
            SliverMainAxisGroup(
              slivers: <Widget>[
                PinnedHeaderSliver(
                  child: Container(
                    color: Theme.of(context).colorScheme.primaryContainer,
                    padding: const EdgeInsetsGeometry.all(16),
                    child: Text('Group $i', style: const TextStyle(fontSize: 18, fontWeight: FontWeight.bold)),
                  ),
                ),
                SliverList.builder(
                  itemCount: 10,
                  itemBuilder: (BuildContext context, int index) => ListTile(
                    title: Text('Item $index'),
                    leading: CircleAvatar(child: Text('$index')),
                    onTap: () {},
                  ),
                ),
              ],
            ),
        ],
      ),
    ],
  );
}
```

</details>

**Implementation Note:**
Regarding `SliverSequenceMixin`, I debated between extracting it or simply duplicating the `layoutChildSequence` code. Paradoxically, I found the logic more approachable through `NestedSlivers` than through `Viewport`, as the latter requires understanding a much larger set of concepts.

WDYT @Piinks?

Here is the code and a demo:

<details><summary>Code sample full demo</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
  Widget build(BuildContext context) => MaterialApp(
    title: 'NestedSlivers Demo',
    theme: ThemeData(primarySwatch: Colors.blue, useMaterial3: true),
    home: const Scaffold(body: NumberList()),
  );
}

class NumberList extends StatelessWidget {
  const NumberList({super.key});

  @override
  Widget build(BuildContext context) => CustomScrollView(
    physics: const BouncingScrollPhysics(),
    slivers: <Widget>[
      MyHeaderSliver(),
      for (int i = 1; i <= 5; i++)
        SliverMainAxisGroup(
          slivers: <Widget>[
            PinnedHeaderSliver(
              child: Container(
                color: Theme.of(context).colorScheme.primaryContainer,
                padding: const EdgeInsetsGeometry.all(16),
                child: Text('Group $i', style: const TextStyle(fontSize: 18, fontWeight: FontWeight.bold)),
              ),
            ),
            SliverList.builder(
              itemCount: 10,
              itemBuilder: (BuildContext context, int index) => ListTile(
                title: Text('Item $index'),
                leading: CircleAvatar(child: Text('$index')),
                onTap: () {},
              ),
            ),
          ],
        ),
      SliverToBoxAdapter(child: SizedBox(height: MediaQuery.paddingOf(context).bottom)),
    ],
  );
}

class MyHeaderSliver extends StatelessWidget {
  const MyHeaderSliver({super.key});

  @override
  Widget build(BuildContext context) => NestedSlivers(
    slivers: <Widget>[
      SliverAppBar(
        pinned: true,
        stretch: true,
        // floating: true,
        // snap: true,
        expandedHeight: 200,
        backgroundColor: Colors.yellow,
        flexibleSpace: FlexibleSpaceBar(title: Text('My App Bar')),
      ),
      SliverToBoxAdapter(
        child: Container(
          height: 150,
          color: Colors.grey[200],
          alignment: Alignment.center,
          child: const Column(
            children: <Widget>[
              Text('SliverToBoxAdapter', style: TextStyle(fontSize: 18, fontWeight: FontWeight.bold)),
              SizedBox(height: 8),
              Text('This content will disappear under the pinned headers'),
            ],
          ),
        ),
      ),
      PinnedHeaderSliver(
        child: Container(
          color: Theme.of(context).colorScheme.tertiaryContainer,
          padding: const EdgeInsetsGeometry.all(16),
          child: Text('My Pinned header', style: const TextStyle(fontSize: 18, fontWeight: FontWeight.bold)),
        ),
      ),
    ],
  );
}
```

</details>

https://github.com/user-attachments/assets/2ea59070-3acd-401e-85ca-ae63b1b8072a

### DIFF
--- a/packages/flutter/lib/src/rendering/nested_slivers.dart
+++ b/packages/flutter/lib/src/rendering/nested_slivers.dart
@@ -1,0 +1,251 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// @docImport 'package:flutter/widgets.dart';
+library;
+
+import 'dart:math' as math;
+
+import 'package:flutter/src/rendering/viewport.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+import 'object.dart';
+import 'sliver.dart';
+
+/// A sliver that places multiple sliver children in a linear array along the
+/// main axis.
+///
+/// The layout algorithm lays out slivers one by one. If the sliver is at the top
+/// of the viewport or above the top, then we pass in a nonzero [SliverConstraints.scrollOffset]
+/// to inform the sliver at what point along the main axis we should start layout.
+/// For the slivers that come after it, we compute the amount of space taken up so
+/// far to be used as the [SliverPhysicalParentData.paintOffset] and the
+/// [SliverConstraints.remainingPaintExtent] to be passed in as a constraint.
+///
+/// Finally, this sliver will also ensure that all child slivers are painted within
+/// the total scroll extent of the group by adjusting the child's
+/// [SliverPhysicalParentData.paintOffset] as necessary. This can happen for
+/// slivers such as [SliverPersistentHeader] which, when pinned, positions itself
+/// at the top of the [Viewport] regardless of the scroll offset.
+class RenderNestedSlivers extends RenderSliver
+    with
+        ContainerRenderObjectMixin<RenderSliver, SliverPhysicalContainerParentData>,
+        SliverSequenceMixin {
+  RenderNestedSlivers({SliverPaintOrder? paintOrder}) : _paintOrder = paintOrder;
+
+  @override
+  SliverPaintOrder get paintOrder {
+    return _paintOrder ??
+        switch (RenderAbstractViewport.maybeOf(this)) {
+          RenderViewportBase(:final SliverPaintOrder paintOrder) => paintOrder,
+          _ => SliverPaintOrder.firstIsTop,
+        };
+  }
+
+  SliverPaintOrder? _paintOrder;
+  set paintOrder(SliverPaintOrder? value) {
+    if (_paintOrder == value) {
+      return;
+    }
+    _paintOrder = value;
+    markNeedsLayout();
+  }
+
+  @override
+  void setupParentData(RenderObject child) {
+    if (child.parentData is! SliverPhysicalContainerParentData) {
+      child.parentData = SliverPhysicalContainerParentData();
+    }
+  }
+
+  @override
+  double? childScrollOffset(RenderSliver child) {
+    assert(child.parent == this);
+
+    return scrollOffsetOf(child, -_maxScrollObstructionExtentBefore(child));
+  }
+
+  double _maxScrollObstructionExtentBefore(RenderSliver child) {
+    final GrowthDirection growthDirection = child.constraints.growthDirection;
+    switch (growthDirection) {
+      case GrowthDirection.forward:
+        var pinnedExtent = 0.0;
+        RenderSliver? current = firstChild;
+        while (current != child) {
+          pinnedExtent += current!.geometry!.maxScrollObstructionExtent;
+          current = childAfter(current);
+        }
+        return pinnedExtent;
+      case GrowthDirection.reverse:
+        var pinnedExtent = 0.0;
+        RenderSliver? current = lastChild;
+        while (current != child) {
+          pinnedExtent += current!.geometry!.maxScrollObstructionExtent;
+          current = childBefore(current);
+        }
+        return pinnedExtent;
+    }
+  }
+
+  @override
+  double childMainAxisPosition(RenderSliver child) {
+    final Offset childPaintOffset = paintOffsetOf(child);
+    return switch (applyGrowthDirectionToAxisDirection(
+      child.constraints.axisDirection,
+      child.constraints.growthDirection,
+    )) {
+      AxisDirection.down => childPaintOffset.dy,
+      AxisDirection.right => childPaintOffset.dx,
+      AxisDirection.up => geometry!.paintExtent - child.geometry!.paintExtent - childPaintOffset.dy,
+      AxisDirection.left =>
+        geometry!.paintExtent - child.geometry!.paintExtent - childPaintOffset.dx,
+    };
+  }
+
+  @override
+  double childCrossAxisPosition(RenderSliver child) => 0.0;
+
+  @override
+  void performLayout() {
+    final RenderSliver? leadingChild;
+    final RenderSliver? Function(RenderSliver) advance;
+
+    switch (constraints.growthDirection) {
+      case GrowthDirection.forward:
+        leadingChild = firstChild;
+        advance = childAfter;
+      case GrowthDirection.reverse:
+        leadingChild = lastChild;
+        advance = childBefore;
+    }
+
+    if (leadingChild == null) {
+      geometry = SliverGeometry.zero;
+      return;
+    }
+
+    final double scrollOffsetCorrection = layoutChildSequence(
+      child: leadingChild,
+      axisDirection: constraints.axisDirection,
+      crossAxisDirection: constraints.crossAxisDirection,
+      scrollOffset: constraints.scrollOffset,
+      overlap: constraints.overlap,
+      layoutOffset: -math.min(constraints.overlap, 0.0),
+      remainingPaintExtent: constraints.remainingPaintExtent,
+      mainAxisExtent: constraints.remainingPaintExtent,
+      crossAxisExtent: constraints.crossAxisExtent,
+      userScrollDirection: constraints.userScrollDirection,
+      growthDirection: constraints.growthDirection,
+      advance: advance,
+      remainingCacheExtent: constraints.remainingCacheExtent,
+      cacheOrigin: constraints.cacheOrigin,
+      updateParentGeometry: (geometry) {
+        this.geometry = geometry;
+      },
+    );
+
+    if (scrollOffsetCorrection != 0.0) {
+      geometry = SliverGeometry(scrollOffsetCorrection: scrollOffsetCorrection);
+    }
+
+    final AxisDirection appliedGrowthDirectionToAxisDirection = applyGrowthDirectionToAxisDirection(
+      constraints.axisDirection,
+      constraints.growthDirection,
+    );
+
+    if (appliedGrowthDirectionToAxisDirection == AxisDirection.left ||
+        appliedGrowthDirectionToAxisDirection == AxisDirection.up) {
+      RenderSliver? child = leadingChild;
+      while (child != null) {
+        final childParentData = child.parentData! as SliverPhysicalParentData;
+        childParentData.paintOffset += switch (axisDirectionToAxis(
+          appliedGrowthDirectionToAxisDirection,
+        )) {
+          Axis.horizontal => Offset(geometry!.paintExtent, 0.0),
+          Axis.vertical => Offset(0.0, geometry!.paintExtent),
+        };
+        child = advance(child);
+      }
+    }
+  }
+
+  @override
+  void paint(PaintingContext context, Offset offset) => paintContents(context, offset);
+
+  @override
+  Offset paintOffsetOf(RenderSliver child) {
+    final childParentData = child.parentData! as SliverPhysicalParentData;
+    return childParentData.paintOffset;
+  }
+
+  @override
+  void applyPaintTransform(RenderSliver child, Matrix4 transform) {
+    final childParentData = child.parentData! as SliverPhysicalParentData;
+    childParentData.applyPaintTransform(transform);
+  }
+
+  @override
+  bool hitTestChildren(
+    SliverHitTestResult result, {
+    required double mainAxisPosition,
+    required double crossAxisPosition,
+  }) {
+    RenderSliver? child = firstChild;
+    while (child != null) {
+      final Offset paintOffset = (child.parentData! as SliverPhysicalParentData).paintOffset;
+      final bool isHit = result.addWithAxisOffset(
+        mainAxisPosition: mainAxisPosition,
+        crossAxisPosition: crossAxisPosition,
+        paintOffset: paintOffset,
+        mainAxisOffset: childMainAxisPosition(child),
+        crossAxisOffset: childCrossAxisPosition(child),
+        hitTest: child.hitTest,
+      );
+      if (isHit) {
+        return true;
+      }
+      child = childAfter(child);
+    }
+    return false;
+  }
+
+  @override
+  RenderSliver? get origin => switch (applyGrowthDirectionToAxisDirection(
+    constraints.axisDirection,
+    constraints.growthDirection,
+  )) {
+    AxisDirection.left || AxisDirection.up => lastChild,
+    AxisDirection.right || AxisDirection.down => firstChild,
+  };
+
+  @override
+  void updateChildLayoutOffset(
+    RenderSliver child,
+    double layoutOffset,
+    GrowthDirection growthDirection,
+  ) {
+    final childParentData = child.parentData! as SliverPhysicalParentData;
+    childParentData.paintOffset = switch (applyGrowthDirectionToAxisDirection(
+      constraints.axisDirection,
+      constraints.growthDirection,
+    )) {
+      AxisDirection.left => Offset(-layoutOffset - child.geometry!.paintExtent, 0.0),
+      AxisDirection.up => Offset(0.0, -layoutOffset - child.geometry!.paintExtent),
+      AxisDirection.right => Offset(layoutOffset, 0.0),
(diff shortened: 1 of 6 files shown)


packages/flutter/lib/src/widgets/sliver.dart
```diff
@@ -1922,3 +1973,43 @@ class _RenderSliverEnsureSemantics extends RenderProxySliver {
   @override
   bool get ensureSemantics => true;
 }
+
+/// A sliver that makes its sliver child unpinned.
+class SliverUnpinned extends SingleChildRenderObjectWidget {
```
Piinks, line 1978: The introduction of SliverUnpinned feels like a logic reversal that creates a conflict of interest between parent and child.
In Flutter, a sliver (like SliverAppBar or PinnedHeaderSliver) identifies itself as pinned via its geometry. SliverUnpinned attempts to force-unpin a child by manually calculating the pinned offset and then subtracting it from the paintOrigin. This is fragile because it assumes the parent can perfectly guess or counteract the child's internal pinning logic. If the child sliver's pinning logic changes (e.g., adding floating or snap behavior), SliverUnpinned will likely calculate the wrong correction, leading to visual fighting where the child tries to pin and the parent tries to pull it away.
The style guide dictates that a parent should not have special knowledge of how a child is implementing its internal logic, and SliverUnpinned is explicitly designed to undo a child's internal decision to pin. Instead of being agnostic, SliverUnpinned inspects the child's geometry.maxScrollObstructionExtent. If it sees the child is trying to act like a pinned sliver, it applies a manual mathematical correction to neutralize that behavior.
A parent should provide a layout context, but it shouldn't be reaching inside to counteract a child's reported geometry. If a child says "I am pinning," the parent should respect that. If we don't want it to pin, we should be changing the constraints we give the child, not fighting the geometry it returns.

That being said, there is some similar prior art about counteracting pinning but in a very specific, limited way. In RenderSliverMainAxisGroup.performLayout (lines 374-399 in sliver_group.dart), the framework performs a paint correction if the children's total paint extent exceeds the remaining extent of the group. The key difference is the framework's current logic in SliverMainAxisGroup is positional correction (moving the child). This PR's logic in SliverUnpinned is geometry mutation (changing what the child reported).

packages/flutter/lib/src/rendering/sliver.dart
```diff
@@ -2117,3 +2117,340 @@ class RenderSliverToBoxAdapter extends RenderSliverSingleBoxAdapter {
...
+  ///  * `crossAxisExtent` is the [SliverConstraints.crossAxisExtent] to pass to
+  ///    each child.
+  ///  * `growthDirection` is the [SliverConstraints.growthDirection] to pass to
+  ///    each child.
+  ///
+  /// Returns the first non-zero [SliverGeometry.scrollOffsetCorrection]
+  /// encountered, if any. Otherwise returns 0.0. Typical callers will call this
+  /// function repeatedly until it returns 0.0.
+  @protected
+  double layoutChildSequence({
```
Piinks, line 2171: By extracting layoutChildSequence into NestedSlivers, this is effectively creating a mini-viewport. The main Viewport at the top of the tree expects slivers to be slices of content. When a sliver contains a whole sequence of other slivers that have their own internal coordination logic (via this mixin), that internal state is hidden from the main Viewport. This might become problematic for features like Scrollable.ensureVisible, focus management, and accessibility. If the main Viewport asks for the childScrollOffset of a sliver deep inside NestedSlivers, the math has to traverse multiple layers of custom sequence logic, which is much more prone to error than the current linear delegation.

packages/flutter/lib/src/rendering/nested_slivers.dart
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  @override
+  void applyPaintTransform(RenderSliver child, Matrix4 transform) {
+    final childParentData = child.parentData! as SliverPhysicalParentData;
+    childParentData.applyPaintTransform(transform);
+  }
+
+  @override
+  bool hitTestChildren(
```
Piinks, line 189: This current implementation of RenderNestedSlivers hit-tests children in layout order (first to last). Standard Viewport logic hit-tests in reverse paint order (top-most first). Because this PR explicitly aims to improve support for overlapping slivers (like pinned headers), hit-testing in layout order is a regression. If a pinned header overlaps a list, the list might steal the touch events even though the header is visually on top.

packages/flutter/lib/src/rendering/sliver.dart
Piinks: Some additional thoughts here, I can see how the SliverSequenceMixin aims to unify Viewport and SliverMainAxisGroup, but it currently does so by introducing a magic method (layoutChildSequence) with a massive parameter list (15+ arguments). This doesn't actually simplify the codebase; it just moves the most complex part of the Viewport into a mixin where it's harder to test in isolation. Any change to the core sliver protocol now requires updating this massive method signature across multiple implementations. Usually, if we need this much shared logic, we could look at a dedicated SliverSequenceConstraints object rather than a mixin that tries to be a Swiss Army knife for layout. :)

packages/flutter/lib/src/widgets/sliver.dart
```diff
@@ -1850,22 +1850,73 @@ class SliverCrossAxisGroup extends MultiChildRenderObjectWidget {
...
+//     children
+//         .where((Element e) {
+//           final renderSliver = e.renderObject! as RenderSliver;
+//           return renderSliver.geometry!.visible;
+//         })
+//         .forEach(visitor);
+//   }
+// }
+
+class SliverMainAxisGroup extends StatelessWidget {
```
Piinks, line 1881: A widget's behavior should be predictable and not change fundamentally because it's wrapped in a specific modifier that isn't apparent to the developer. SliverMainAxisGroup is now a StatelessWidget that forces a composition of SliverUnpinned + NestedSlivers. This creates a hidden contract. If a user uses NestedSlivers directly, they get one behavior (everything pins). If they use SliverMainAxisGroup, they get another (nothing pins).
Widgets should be atomic. When we start creating complex modifier stacks like SliverUnpinned(NestedSlivers(...)) just to make a standard widget work, it suggests that the atomic widgets themselves (the RenderObjects) aren't truly agnostic or self-contained.

packages/flutter/lib/src/rendering/sliver.dart
```diff
@@ -2117,3 +2117,340 @@ class RenderSliverToBoxAdapter extends RenderSliverSingleBoxAdapter {
...
+    required double layoutOffset,
+    required double remainingPaintExtent,
+    required double mainAxisExtent,
+    required double crossAxisExtent,
+    required ScrollDirection userScrollDirection,
+    required GrowthDirection growthDirection,
+    required RenderSliver? Function(RenderSliver child) advance,
+    required double remainingCacheExtent,
+    required double cacheOrigin,
+    void Function(SliverGeometry geometry)? updateParentGeometry,
```
Piinks, line 2186: This creates a circular dependency within a single layout pass. The child is laid out based on parent constraints, but then the child (via the mixin) reaches back up to update the parent's geometry while other children are still being laid out. 
Data (constraints) should flow down, and sizes (geometry) should flow up at the end of the layout function. Mid-loop updates to the parent state violate the expectation that a parent is a stable container for its children during the duration of performLayout.

Piinks (commented):
Hi @zemanux, first of all, thank you so much for your patience while I got caught up on this and the related PRs. I also want to specifically thank you for the detailed demos and the effort you've put into this SliverSequenceMixin POC.
I’ve spent some time digging into the 'why' behind these proposals, and I want to validate that the problem you’re trying to solve is a very real pain point in the current sliver system. However, after looking at the NestedSlivers approach, I have some deep architectural concerns that make me hesitant to move forward with this specific direction.
Instead of building a parallel layout system to handle nesting and unpinning, I think the more 'Flutter-idiomatic' win would be solving the intrinsic height problem at the source. If we can find a way for pinned headers to report and respond to dynamic height changes (like text scaling) without needing fixed extents or complex 'unpinning' wrappers, it would solve your use case and benefit the entire ecosystem.
I've left some specific technical comments on the code below to highlight the hit-testing and geometry issues I found, let me know your thoughts. 💙

packages/flutter/lib/src/widgets/sliver.dart
zemanux: > That being said, there is some similar prior art about counteracting pinning but in a very specific, limited way. In RenderSliverMainAxisGroup.performLayout (lines 374-399 in sliver_group.dart), the framework performs a paint correction if the children's total paint extent exceeds the remaining extent of the group. The key difference is the framework's current logic in SliverMainAxisGroup is positional correction (moving the child). This PR's logic in SliverUnpinned is geometry mutation (changing what the child reported).

I agree with you. The approach I took with `SliverUnpinned` is indeed extreme, and I definitely pushed this exploration too far by fighting the child's geometry. 

Rather than trying to force the unpinning behavior from the outside, I envision adding variations directly within `SliverSequenceMixin`. This could potentially be handled using a delegate or a similar system.

The main goal of this exploration was really to demonstrate that we can:
* Factor out the layout sequence algorithm.
* Specify its behavior (such as whether it should be pinned or not) in a cleaner way, for instance via a delegate.

packages/flutter/lib/src/rendering/sliver.dart
zemanux: > By extracting layoutChildSequence into NestedSlivers, this is effectively creating a mini-viewport. The main Viewport at the top of the tree expects slivers to be slices of content. When a sliver contains a whole sequence of other slivers that have their own internal coordination logic (via this mixin), that internal state is hidden from the main Viewport. This might become problematic for features like Scrollable.ensureVisible, focus management, and accessibility. If the main Viewport asks for the childScrollOffset of a sliver deep inside NestedSlivers, the math has to traverse multiple layers of custom sequence logic, which is much more prone to error than the current linear delegation.

When I implemented this, I actually didn't think of it as creating a mini-viewport—in my mind, the sequence remains fully integrated within the main Viewport. 

My intention was strictly focused on isolating the layout logic for the sliver sequence. Because it only touches the layout phase, the underlying mechanics, such as hit testing, remain exactly the same as before.

zemanux: > Some additional thoughts here, I can see how the SliverSequenceMixin aims to unify Viewport and SliverMainAxisGroup, but it currently does so by introducing a magic method (layoutChildSequence) with a massive parameter list (15+ arguments). This doesn't actually simplify the codebase; it just moves the most complex part of the Viewport into a mixin where it's harder to test in isolation. Any change to the core sliver protocol now requires updating this massive method signature across multiple implementations. Usually, if we need this much shared logic, we could look at a dedicated SliverSequenceConstraints object rather than a mixin that tries to be a Swiss Army knife for layout. :)

I don't have a strong preference on this either way, and I completely agree that we shouldn't force code factorization if it ends up being counterproductive. 

Regarding the massive parameter list, I intentionally left it untouched initially just to highlight that this was purely a "code transfer" from the Viewport. My goal was simply to show the extraction without altering the underlying logic. 

That being said, I agree that if we were to take this exploration further, we would absolutely need to refactor this—perhaps by introducing a dedicated `SliverSequenceConstraints` object, as you suggested, to clean up the API.

packages/flutter/lib/src/rendering/nested_slivers.dart
zemanux: > This current implementation of RenderNestedSlivers hit-tests children in layout order (first to last). Standard Viewport logic hit-tests in reverse paint order (top-most first). Because this PR explicitly aims to improve support for overlapping slivers (like pinned headers), hit-testing in layout order is a regression. If a pinned header overlaps a list, the list might steal the touch events even though the header is visually on top.

Good catch. I will remove or fix this hit-testing logic if we decide to continue exploring this path.

packages/flutter/lib/src/widgets/sliver.dart
zemanux: As mentioned in my previous reply, `SliverUnpinned` is being removed entirely, and we can handle this behavior by adjusting the `SliverSequenceMixin` instead.

packages/flutter/lib/src/rendering/sliver.dart
zemanux: > This creates a circular dependency within a single layout pass. The child is laid out based on parent constraints, but then the child (via the mixin) reaches back up to update the parent's geometry while other children are still being laid out. Data (constraints) should flow down, and sizes (geometry) should flow up at the end of the layout function. Mid-loop updates to the parent state violate the expectation that a parent is a stable container for its children during the duration of performLayout.

This actually isn't a circular dependency, but rather a poor choice of naming on my part that makes it look exactly like one! 

In reality, `updateParentGeometry` is called by the parent itself. It's simply the callback invoked by `SliverSequenceMixin` once it has calculated its final `geometry`. 

If we decide to move forward with this exploration, I will absolutely rename this method to make the flow much more explicit and avoid this kind of confusion.

Piinks:
Hi @zemanux, I am so sorry for losing track of this PR! It’s been far too long, and I really appreciate your patience while I got back to this.

After spending more time reviewing the NestedSlivers approach and considering our long-term goals for the sliver system, I’ve decided that this isn't a direction we want to take within the core framework. While the POC is impressive, NestedSlivers introduces a level of architectural complexity that we’d like to avoid in the core.

We’ve been moving away from "uber-complex" slivers like SliverAppBar (which is moving to the material_ui package) in favor of more modular primitives like PinnedHeaderSliver, SliverResizingHeader, and SliverFloatingHeader. These were introduced specifically to allow for more flexibility through composition without needing a parallel nesting system.

I still believe the composition approach you’re advocating for is the right way to solve these problems, but I think the better path forward for the framework is to continue expanding that existing suite of behavior-focused slivers. If there are specific effects we still can't achieve, we should look at what new modular primitives might be missing from that suite.

My suggestion would be to publish your work on NestedSlivers as a standalone package. This would allow you to iterate on the concept more quickly and help the community experiment with these composition patterns. The insights gained from a package would be incredibly valuable if we decide to bring new primitives into the core later on.

Thank you again for the incredible effort and the detailed demos you put into this POC. It has been very helpful for our internal discussions!

zemanux:
Hi @Piinks, 

Thank you for taking the time to review this draft. I completely understand your feedback and perspective. 
I will go ahead and close this PR now. Thanks again for your guidance!

Issue #157664, opened 2024-10-26
Autocomplete options overlay should automatically moves up when soft keyboard is visible


dikatok:
### Steps to reproduce

When soft keyboard is visible, unless options widget is constrained or set to a specific height, part of the options overlay will be hidden behind the keyboard. This is a dealbreaker especially in mobile, since we would prefer not to set a specific height value.

### Expected results

https://github.com/user-attachments/assets/9d3bdf77-3474-48fc-9631-cea09470ed73

### Actual results

https://github.com/user-attachments/assets/ce09822d-c6a9-4681-a84a-2103cda1a31e

### Code sample

```dart
import "package:flutter/material.dart";
import "package:tokowijaya/features/widgets/custom_autocomplete.dart";

/// Flutter code sample for [RawAutocomplete].

class AutocompleteExampleApp extends StatelessWidget {
  const AutocompleteExampleApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        resizeToAvoidBottomInset: false,
        appBar: AppBar(
          title: const Text("RawAutocomplete Basic"),
        ),
        body: const Center(
          child: AutocompleteBasicExample(),
        ),
      ),
    );
  }
}

class AutocompleteBasicExample extends StatelessWidget {
  const AutocompleteBasicExample({super.key});

  static const List<String> _options = <String>[
    "aardvark",
    "bobcat",
    "chameleon",
    "chameleon",
    "chameleon",
    "chameleon",
    "chameleon",
    "chameleon",
    "chameleon",
  ];

  @override
  Widget build(BuildContext context) {
    return CustomAutocomplete<String>(
      optionsBuilder: (TextEditingValue textEditingValue) {
        return _options.where((String option) {
          return option.contains(textEditingValue.text.toLowerCase());
        });
      },
      fieldViewBuilder: (
        BuildContext context,
        TextEditingController textEditingController,
        FocusNode focusNode,
        VoidCallback onFieldSubmitted,
      ) {
        return TextFormField(
          controller: textEditingController,
          focusNode: focusNode,
          onFieldSubmitted: (String value) {
            onFieldSubmitted();
          },
        );
      },
      optionsViewBuilder: (
        BuildContext context,
        AutocompleteOnSelected<String> onSelected,
        Iterable<String> options,
      ) {
        return Material(
          elevation: 4,
          child: ListView.builder(
            padding: const EdgeInsets.all(8),
            itemCount: options.length,
[...]

Issue #142921, opened 2024-02-05
menuChildren by MenuAnchor gets hidden behind mobile keyboard


norbertje:
### Steps to reproduce

Using `MenuAnchor` (or `DropDownMenu`) within a scrollable `ListView` where anchor itself is a `TextField`. When scrolled and anchor is positioned at the bottom of the screen and typed within, `menuChildren` pop up and get hidden behind mobile keyboard.

1. Use the given code in a mobile environment
2. Scroll until TextField is positioned at the bottom of the screen
3. Click on TextField
4. Type any character
5. menuChildren appear behind the keyboard

If anchor would be a button, then `menuChildren` would appear above the anchor as expected. 

When the screen is touched (and not when actual tap happened) outside of the anchor and its children, `MenuAnchor` closes (aka hides the children). Due to this, if the user tries to adjust the scroll position to see the `menuChildren`, those will immediately disappear.

Applying `scrollPadding` to `TextField` works IF it is not positioned around at the end of the screen/ListView. Because then it's limited how much it can use `scrollPadding` (ex.: if `TextField` is 20 pixels away from the end of `ListView`, setting a bottom `scrollPadding` to 100 pixels, still only applies 20 pixels). In that scenario would need to detect when if `TextField` is focused and apply additional bottom padding to the `ListView`.

As a side note I mention, when anchor is at the end of ListView (which is my case),`menuChildren` appears behind the keyboard. When closing the keyboard, `menuChildren` is repositioned above the anchor. Sometimes, an extra gap appears between the children and the anchor in this case.

### Expected results

`menuChildren` would appear above the anchor

### Actual results

`menuChildren` appear below the anchor and behind the mobile keyboard.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(_MenuAnchor());
}

class _MenuAnchor extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        body: ListView(
          children: [
            const SizedBox(height: 800),
            MenuAnchor(
              menuChildren: const [
                Padding(
                  padding: EdgeInsets.all(32),
                  child: Text('Text'),
                ),
              ],
              builder: (context, controller, child) => TextField(
                onChanged: (value) => controller.open(),
                decoration: const InputDecoration(
                  border: OutlineInputBorder(
                    borderSide: BorderSide(),
                  ),
                ),
              ),
            ),
[...]

PR #181579, opened 2026-01-28
Propagates Overlay's MediaQueryData to OverlayPortal child

patrickBillingsley:
Widgets like Scaffold strip MediaQueryData causing widgets rendered in the Overlay to be obscured, e.g., by the software keyboard.

[Related Issue - 142921](https://github.com/flutter/flutter/issues/142921)
[Related Issue - 157664](https://github.com/flutter/flutter/issues/157664)

This does not directly fix these two issues but appears to be an underlying issue with both which merited this being a separate PR. I have a separate fix for the MenuAnchor that I will supply in a follow up PR.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -154,6 +154,61 @@ void main() {
     expect(directionSeenByOverlayChild, textDirection);
   });
 
+  testWidgets('OverlayPortal overlayChild receives MediaQuery properties from Overlay context', (WidgetTester tester) async {
+    final OverlayPortalController controller = OverlayPortalController();
+    final EdgeInsets expectedPadding = EdgeInsets.all(10);
+    final EdgeInsets expectedViewInsets = EdgeInsets.only(bottom: 300);
+    final EdgeInsets expectedViewPadding = EdgeInsets.only(top: 50, bottom: 20);
+    final Size expectedSize = Size(800, 600);
```
victorsanni:
```suggestion
    final controller = OverlayPortalController();
    const expectedPadding = EdgeInsets.all(10);
    const expectedViewInsets = EdgeInsets.only(bottom: 300);
    const expectedViewPadding = EdgeInsets.only(top: 50, bottom: 20);
    const expectedSize = Size(800, 600);
```
New omit obvious types lint.

packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -154,6 +154,61 @@ void main() {
...
+
+    MediaQueryData? overlayChildData;
+
+    await tester.pumpWidget(
+      MediaQuery(
+        data: MediaQueryData(
```
victorsanni:
```suggestion
        data: const MediaQueryData(
```

victorsanni (approved):

packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -154,6 +154,61 @@ void main() {
...
   });
 
+  testWidgets('OverlayPortal overlayChild receives MediaQuery properties from Overlay context', (WidgetTester tester) async {
+    final OverlayPortalController controller = OverlayPortalController();
+    final EdgeInsets expectedPadding = EdgeInsets.all(10);
+    final EdgeInsets expectedViewInsets = EdgeInsets.only(bottom: 300);
+    final EdgeInsets expectedViewPadding = EdgeInsets.only(top: 50, bottom: 20);
+    final Size expectedSize = Size(800, 600);
+
+    MediaQueryData? overlayChildData;
```
victorsanni, line 227:
```suggestion
    MediaQueryData? overlayChildData;
    late final OverlayEntry entry;
    addTearDown(() {
       entry.remove();
       entry.dispose();
    });
```

packages/flutter/test/widgets/overlay_portal_test.dart
```diff
@@ -154,6 +154,61 @@ void main() {
...
+          padding: expectedPadding,
+          viewInsets: expectedViewInsets,
+          viewPadding: expectedViewPadding,
+          size: expectedSize,
+        ),
+        child: Directionality(
+          textDirection: TextDirection.ltr,
+          child: Overlay(
+            initialEntries: <OverlayEntry>[
+              OverlayEntry(
```
victorsanni:
```suggestion
              entry = OverlayEntry(
```

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -2029,11 +2030,25 @@ class _OverlayPortalState extends State<OverlayPortal> {
         child: Semantics(traversalParentIdentifier: this, child: widget.child),
       );
     }
+
+    MediaQueryData data = MediaQuery.of(context);
+    final OverlayState? overlay = Overlay.maybeOf(context);
```
LongCatIsLooong: Hmm I don't think this is correct because `OverlayPortal` can target a different Overlay than it's closest ancestor Overlay.

LongCatIsLooong (commented):
The approach makes sense to me, except for the way `OverlayState` is acquired in the PR.

packages/flutter/lib/src/widgets/overlay.dart
LongCatIsLooong: (could you add a test for this too? I think you can which Overlay to target using https://main-api.flutter.dev/flutter/widgets/OverlayPortal/overlayLocation.html)

patrick-billingsley: Good catch on that one! I have updated the call to Overlay.maybeOf to account for the overlayLocation parameter on the widget. I also added an additional test for this option.

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -2029,11 +2030,28 @@ class _OverlayPortalState extends State<OverlayPortal> {
         child: Semantics(traversalParentIdentifier: this, child: widget.child),
       );
     }
+
+    MediaQueryData data = MediaQuery.of(context);
+    final OverlayState? overlay = Overlay.maybeOf(
```
LongCatIsLooong: You can get the right context for the Overlay from the `_getLocation` call instead of looking it up again using `Overlay.of`.

patrickBillingsley: I just pushed an update. If I understand right, the `Overlay.maybeOf` call was duplicating what is a potentially cached value in the class. The `_OverlayEntryLocation` provided by the `_getLocation` method holds an `_OverlayEntryWidgetState` which has the necessary context provided from the correct location. This also removes the need for the conditional logic and potential manipulation of the initially fetched `MediaQuery` from the `build` method's context.

This should now only have to traverse the tree once to get the correct context based on the `overlayLocation`.

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -2029,11 +2030,16 @@ class _OverlayPortalState extends State<OverlayPortal> {
...
+
+    final _OverlayEntryLocation overlayLocation = _getLocation(zOrderIndex, widget.overlayLocation);
     return _OverlayPortal(
-      overlayLocation: _getLocation(zOrderIndex, widget.overlayLocation),
+      overlayLocation: overlayLocation,
       overlayChild: _DeferredLayout(
         childIdentifier: this,
-        child: Builder(builder: widget.overlayChildBuilder),
+        child: MediaQuery(
+          data: MediaQuery.of(overlayLocation._childModel.context),
```
LongCatIsLooong: Hmm I think we just want to override the "safe area" related properties on MediaQueryData right? If the developer adds a MediaQuery between an OverlayPortal and its target Overlay (in terms of the widget tree), other overrides on the new MediaQuery like `boldText` should still be respected should the developer chooses to override them?

patrickBillingsley: You are absolutely right.

LongCatIsLooong (approved):

PR #181610, opened 2026-01-28
Add statesController to ExpansionTile

rickhohler:
Exposes WidgetStatesController in ExpansionTile and passes it to the underlying ListTile to allow programmatic control of the tile's state. Fixes #180161.


QuncCccccc (commented):
Hi @rickhohler! Thanks for your contribution! It looks like this is a duplicate of another open PR that is already under review: https://github.com/flutter/flutter/pull/181238

Issue #181573, opened 2026-01-27
Content Sizing Resize called on raster thread


mboetger:
### Steps to reproduce

Start the new content sizing [sample](https://github.com/flutter/samples/pull/2787)

### Expected results

FlutterView resizes itself.

### Actual results

The app crashes sometimes.

### Logs

<details open><summary>Logs</summary>

2026-01-27 15:27:52.280 18860-18882 flutter                 dev.flutter.example.androidView      E  [ERROR:flutter/fml/platform/android/jni_util.cc(206)] android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views. Expected: main Calling: 1.raster (Ask Gemini)
                                                                                                    	at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:11023)
                                                                                                    	at android.view.ViewRootImpl.requestLayout(ViewRootImpl.java:2473)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at androidx.constraintlayout.widget.ConstraintLayout.requestLayout(ConstraintLayout.java:3605)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at androidx.recyclerview.widget.RecyclerView.requestLayout(RecyclerView.java:4927)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.View.requestLayout(View.java:28020)
                                                                                                    	at android.view.ViewGroup.onSetLayoutParams(ViewGroup.java:8112)
                                                                                                    	at android.view.View.setLayoutParams(View.java:20528)
                                                                                                    	at io.flutter.embedding.android.FlutterView$3.resizeEngineView(FlutterView.java:211)
                                                                                                    	at io.flutter.embedding.engine.FlutterJNI.maybeResizeSurfaceView(FlutterJNI.java:1335)

</details>

PR #181633, opened 2026-01-28
Disable content sizing

mboetger:
Due to https://github.com/flutter/flutter/issues/181573, disable content sizing in this candidate. 

A threading issue may cause the app to crash if the users try to use content sizing. 

Justification for disabling a feature between beta and its promotion to stable: 
* This feature has not be advertised or promoted, yet. 
* This feature has no flag opt in or out it happens on android when the size of the android view is wrap_content. 
* There is a threading issue that crashes the app in some situations.
* Existing android apps may have wrap_content and not know it because the size of the android view was not something flutter considered before.  

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterImageView.java
+++ b/engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterImageView.java
@@ -94,12 +94,6 @@ public class FlutterImageView extends View implements RenderSurface {
     setAlpha(0.0f);
   }
 
-  /*
-  @Override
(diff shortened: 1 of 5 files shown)


reidbaker (approved):
Approved as the android TL. This still needs an approval from the release team.

jtmcdole (approved):
Original feature did not have flags, a safe default, and results in a hard crash.  Approved removing the entry points - is there a test plan for verifying this works when built?

mboetger:
FYI, I've actually got the solution with a flag: [181686](https://github.com/flutter/flutter/pull/181686). I think it would be better to cherry-pick that once it lands in master.

PR #181634, opened 2026-01-28
Fix cross imports for all Cupertino tests

navaronbracke:
This PR fixes all of the Cupertino cross imports, except for the one that uses `SelectableText`, which needs further discussion.

Some Material specific tests for Cupertino are now under `tests/material/cupertino`.

Part of https://github.com/flutter/flutter/issues/177415

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/cupertino/switch_test.dart
```diff
@@ -1498,15 +1492,12 @@ void main() {
...
-                  trailing: CupertinoSwitch(value: value, onChanged: onChanged),
-                ),
+            return MergeSemantics(
+              child: TestListTile(
+                title: CupertinoSwitch(value: value, onChanged: onChanged),
+                onTap: () {
+                  onChanged(!value);
+                },
               ),
             );
```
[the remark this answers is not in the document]
navaronbracke: This test passes, the child in question was not relevant to the test

dev/bots/check_tests_cross_imports.dart
```diff
@@ -265,20 +265,13 @@ class TestsCrossImportChecker {
...
-    'packages/flutter/test/cupertino/date_picker_test.dart',
-    'packages/flutter/test/cupertino/switch_test.dart',
-    'packages/flutter/test/cupertino/magnifier_test.dart',
-    'packages/flutter/test/cupertino/sheet_test.dart',
     'packages/flutter/test/cupertino/form_row_test.dart',
     'packages/flutter/test/cupertino/colors_test.dart',
     'packages/flutter/test/cupertino/text_form_field_row_restoration_test.dart',
-    'packages/flutter/test/cupertino/slider_test.dart',
+    // TODO(justinmc): Should we put this test in Material?
+    'packages/flutter/test/cupertino/material/cupertino_sheet_material_snackbar_test.dart',
```
navaronbracke: My rubber ducky was asleep, will move this test to Material, per the doc comment above.

packages/flutter/test/cupertino/material/tab_scaffold_test.dart
navaronbracke, line 1: These tests were in the wrong place

packages/flutter/test/cupertino/date_picker_test.dart
```diff
@@ -2140,12 +2139,10 @@ void main() {
     handle.dispose();
   });
 
-  // TODO(justinmc): Don't test Material interactions in Cupertino tests.
-  // https://github.com/flutter/flutter/issues/177028
-  testWidgets('DatePicker adapts to MaterialApp dark mode', (WidgetTester tester) async {
+  testWidgets('DatePicker adapts to CupertinoApp dark mode', (WidgetTester tester) async {
```
justinmc, line 2142: Would you mind copying the original test here into Material somewhere so that we have both? I think it's reasonable to test both, especially because the theming under the hood is unrelated between Material and Cupertino, so it really is testing different mechanisms, and in situations that many real apps probably use.

Same for the other "adapts to MaterialApp" tests below.

packages/flutter/test/cupertino/material/tab_scaffold_test.dart
justinmc, line 1: Can you remove this `material` directory altogether? It should be Material's responsibility to test Cupertino, and not the other way around. There is a README in there that you should delete, and you should update the README at test/cupertion/README.md to not mention it. Instead of:

"
The 'material' subdirectory contains tests for cross-interactions of Material
Cupertino widgets in hybridized apps.
"

I would say:

"
Testing interactions between Material and Cupertino is the responsibility of the Material library.
"

justinmc: Also, in the case of this file, do you think it's actually testing anything useful related to Material? It looks like it was deliberately written to do so, maybe we should consider keeping it and moving it to Material.

For example, look at the "Last tab gets focus" test here. There is already another test with the same name in cupertino/tab_scaffold_test.dart. Is it testing exactly the same thing you did when you ported this test to that file?

packages/flutter/test/cupertino/sheet_test.dart
```diff
@@ -1026,7 +1020,7 @@ void main() {
     expect(find.text('Page 2'), findsNothing);
     expect(find.text('ListItem 0'), findsOneWidget);
 
-    await tester.tap(find.byKey(popupMenuButtonKey));
+    await tester.tap(find.text('Button'));
```
justinmc: For this problem with CupertinoContextMenu, you have to press and hold to get it to open, maybe that's why this tap isn't working. Try playing around with it in the dartpad example in the docs: https://main-api.flutter.dev/flutter/cupertino/CupertinoContextMenu-class.html

packages/flutter/test/material/cupertino/README.md
justinmc, line 1: Would you mind adding this sentence to the README?

"
Testing interactions between Material and Cupertino is the responsibility of the Material library, not Cupertino.
"

Otherwise, I support the creation of this directory in Material! I wonder if there are any other tests lurking around Material that should go in here (a question for another PR).

justinmc (commented):
Splitting off SelectableText sounds good to me.

I left a comment with some ideas about CupertinoContextMenu.

packages/flutter/test/cupertino/date_picker_test.dart
navaronbracke: Will do!

packages/flutter/test/cupertino/sheet_test.dart
navaronbracke: Ah, it's a long press, wow. I'll take a look, thanks!

packages/flutter/test/cupertino/material/tab_scaffold_test.dart
navaronbracke: That material directory was deleted but the git diff here is a bit funny (probably because Github does not list deleted directories, only deleted files?)

Screenshot 2026-01-29 at 20 57 42

> Is it testing exactly the same thing you did when you ported this test to that file?

As for the test, it was entirely duplicate, so I'm just going to remove it? It only tested focus for the cupertino tabs. The only thing different between the two was how the focus nodes were set up in a loop (with the same end result anyway)

I will do the other things, though!

packages/flutter/test/cupertino/date_picker_test.dart
navaronbracke: Moved the three tests from picker_test.dart & date_picker_test.dart to a new file

packages/flutter/test/cupertino/material/tab_scaffold_test.dart
justinmc: Sounds good, thanks!

packages/flutter/test/cupertino/sheet_test.dart
```diff
@@ -1026,11 +1025,19 @@ void main() {
     expect(find.text('Page 2'), findsNothing);
     expect(find.text('ListItem 0'), findsOneWidget);
 
-    await tester.tap(find.byKey(popupMenuButtonKey));
-    await tester.pumpAndSettle();
+    final Offset contextMenuButton = tester.getCenter(find.text('Button'));
+    expect(find.text('Item 0'), findsNothing);
 
+    await tester.startGesture(contextMenuButton);
```
justinmc: Could you take the returned TestGesture and call `.up` on it at some point? I just worry that having that hidden gesture hanging around could cause a bug in the future.

justinmc (approved):
LGTM minus one comment about startGesture that I think we should clean up.

packages/flutter/test/cupertino/sheet_test.dart
navaronbracke: Will do. I found it weird as well, but forgot about it once the tests were green

navaronbracke: Done

dev/bots/check_tests_cross_imports.dart
```diff
@@ -249,19 +249,8 @@ class TestsCrossImportChecker {
   // TODO(justinmc): Fix all of these tests so there are no cross imports.
   // See https://github.com/flutter/flutter/issues/177028.
   static final Set<String> knownCupertinoCrossImports = <String>{
-    'packages/flutter/test/cupertino/material/tab_scaffold_test.dart',
-    'packages/flutter/test/cupertino/route_test.dart',
+    // TODO(justinmc): This one uses SelectableText
     'packages/flutter/test/cupertino/text_selection_test.dart',
```
navaronbracke, line 252: I'm going to move these tests under material/cupertino, to unblock this, see https://github.com/flutter/flutter/issues/181682#issuecomment-3862648483

packages/flutter/test/cupertino/magnifier_test.dart
```diff
@@ -103,16 +102,20 @@ void main() {
         final Key outerKey = UniqueKey();
 
         await tester.pumpWidget(
-          ColoredBox(
+          CupertinoApp(
             key: outerKey,
-            color: const Color.fromARGB(255, 0, 255, 179),
-            child: MaterialApp(
-              home: Center(
+            theme: const CupertinoThemeData(primaryColor: Color(0xFF6750A4)),
```
navaronbracke, line 107: The primary color is explicitly Material 3 purple, to match the baseline golden image.

packages/flutter/test/cupertino/form_row_test.dart
```diff
@@ -148,15 +147,13 @@ void main() {
     expect(errorTextStyle.style.color, CupertinoColors.destructiveRed);
   });
 
-  // TODO(justinmc): Don't test Material interactions in Cupertino tests.
-  // https://github.com/flutter/flutter/issues/177028
-  testWidgets('CupertinoFormRow adapts to MaterialApp dark mode', (WidgetTester tester) async {
+  testWidgets('CupertinoFormRow adapts to CupertinoApp dark mode', (WidgetTester tester) async {
```
justinmc, line 150: Could you also copy the original test to Material? I think that testing CupertinoFormRow's behavior inside of a MaterialApp is also valid.

packages/flutter/test/cupertino/magnifier_test.dart
justinmc: Yeah strange color choice now, but we should keep it the same rather than change it 👍

packages/flutter/test/cupertino/route_test.dart
```diff
@@ -1646,17 +1645,17 @@ void main() {
             // _CupertinoEdgeShadowDecoration draws the shadows with a series of
             // differently colored 1px rects. Skip all rects not drawn by a
             // _CupertinoEdgeShadowDecoration.
-            if (rect.width != 1.0) {
-              return true;
-            }
-            throw '''
-    Expected: no rects with a width of 1px.
+            if (rect.width == 1.0 && rect.left >= 0.0) {
```
justinmc: Why did you add the case for rect.left?

dev/bots/check_tests_cross_imports.dart
```diff
@@ -249,19 +249,8 @@ class TestsCrossImportChecker {
   // TODO(justinmc): Fix all of these tests so there are no cross imports.
   // See https://github.com/flutter/flutter/issues/177028.
   static final Set<String> knownCupertinoCrossImports = <String>{
-    'packages/flutter/test/cupertino/material/tab_scaffold_test.dart',
-    'packages/flutter/test/cupertino/route_test.dart',
+    // TODO(justinmc): This one uses SelectableText
```
justinmc: Could you link to https://github.com/flutter/flutter/issues/181682?

justinmc (approved):
LGTM with nits 👍 

Looks like the bot didn't get the goldens label, so I removed it.

packages/flutter/test/cupertino/form_row_test.dart
navaronbracke: Added it in a new test file under material/cupertino

packages/flutter/test/cupertino/route_test.dart
navaronbracke: Updated the matcher, so that the expectation is a bit more clear than just checking left.

Let me know if this needs further refinement.

navaronbracke:
Going to look into the leaking test, `packages/flutter/test/cupertino/tab_scaffold_test.dart`

packages/flutter/test/cupertino/tab_scaffold_test.dart
```diff
@@ -7,7 +7,6 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import '../image_data.dart';
```
navaronbracke, line 10: To fix the leak in the test (it was leaking ImageStreamCompleter), I did the following:

1. Since the test used a transparent memory image, which was irrelevant for the expectations, I used a plain widget instead, so we don't need to worry about dangling images.
2. This surfaced the fact that only _some_ tests that used the CupertinoTabScaffold called `imageCache.clear()`, which is probably why the test was leaking.
3. Building the individual tabs for the tests was a bit all over the place, so I cleaned that up by using the existing `tabGenerator()` utility function in this test.

justinmc: Thanks for tracking that down! Looks good to me.

justinmc (approved):
LGTM 👍

Issue #181287, opened 2026-01-21
[impeller] ASAN failures in interactive tests on OpenGLES


planetmarshall:
## Summary

Interactive tests using the Impeller Playground trigger ASAN failures at runtime

## Expected Behaviour

The interactive impeller tests should run without errors on all supported backends

## Actual Behaviour

OpenGLES tests crash at runtime. ASAN (Address Sanitizer) failures are reported on an appropriately instrumented build - for example:

```
./impeller_unittests --enable_playground --gtest_filter='Play/RendererTest.BabysFirstTriangle/OpenGLES'
[INFO:flutter/testing/test_timeout_listener.cc(75)] Test timeout of 300 seconds per test case will be enforced.
Note: Google Test filter = Play/RendererTest.BabysFirstTriangle/OpenGLES
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Play/RendererTest
[ RUN      ] Play/RendererTest.BabysFirstTriangle/OpenGLES
=================================================================
==332411==ERROR: AddressSanitizer: heap-use-after-free on address 0x7c90e8ee3e70 at pc 0x5589370aa026 bp 0x7fff60cfdae0 sp 0x7fff60cfdad8
```

## Steps to Reproduce

1. Configure a build with Address Sanitizer
    ```
    $ gn --unoptimized --asan
    $ ninja -C out/host_debug_unopt/ exe.unstripped/impeller_unittests
    ```
2. Run an interactive renderer test with OpenGLES
    ```
    $ exe.unstripped/impeller_unittests --enable_playground --gtest_filter='Play/RendererTest.BabysFirstTriangle/OpenGLES'
    ```

## System Info

Linux
flutter master (`9c66d267b4d8f82e0e7c9cc6e06f4c3c3f7fe7a9`)

Reproduced on

* OpenGL ES 3.2 NVIDIA 590.48.01
* OpenGL ES 3.2 Mesa 25.3.3-arch1.3

## Notes 

Looks a reference to the `DeviceBufferGLES`'s backing store may be getting lost, resulting in a use after free when the underlying pointer is accessed.

[Attached log](https://github.com/user-attachments/files/24779821/asan.log)

gaaclarke:
Jason has a fix for this issue by just fixing the lifetimes of the buffers.  Ideally we'd get ASAN to verify this but it will take some effort to get that working on CI again.

Issue #165594, opened 2025-03-20
ubuntu 24 update causing issues with `Step('test: test: Host_Tests_for_host_debug_unopt (3)') (retcode: 1)`


jtmcdole:
This is not an engine build; but it is an engine test. 

https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20Production%20Engine%20Drone/1101193/overview

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8719900237536082737/+/u/test:_test:_Host_Tests_for_host_debug_unopt/stdout

```
SUMMARY:
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.BuildTree passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.GetRole passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.SetActions passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.SetExtents passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.SetFlags passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleNodeTest.SetName passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleTextFieldTest.AddSelection passed 0 / 2 times.
  /b/s/w/ir/cache/builder/engine/src/out/ci/host_debug_unopt/exe.unstripped/flutter_linux_unittests FlAccessibleTextFieldTest.CopyCutPasteText passed 0 / 2 times.
[...]

PR #181635, opened 2026-01-28
[Impeller] Ensure that HostBuffers/DeviceBuffers allocated by RendererTest tests are valid for the lifetime of the RenderPass

jason-simmons:
RenderPass holds a list of BufferViews that may hold raw pointers to DeviceBuffers.  HostBuffers and the underlying DeviceBuffers must not be deleted until the RenderPass is no longer using them.

See https://github.com/flutter/flutter/issues/181287


gaaclarke (approved):
LGTM.  I'd really like us to get ASAN turned back on to make sure we don't break this in the future.  We both looked into it briefly and it is a bit involved.  It's not worth holding back this issue for that though.

see also:
- https://github.com/flutter/flutter/issues/165594
- https://github.com/flutter/flutter/issues/181639

PR #151224, opened 2024-07-03
Add `shouldTriggerDismiss` callback parameter to the `Dismissible` widget

ndelanou:
### Description
The default flinging behaviour implemented in the `Dismissible` widget often results in undesired triggering of the dismiss action when used in a scrollable content (e.g. dismissible tiles in a `ListView`)
This PR adds an optional callback parameter `shouldTriggerDismiss` to make it possible to override the default behavior.

The type of the `shouldTriggerDismiss` callback is: `bool? Function(TriggerDismissDetails details)`
When returning `null`, the default behavior is preserved and the widget will check the drag threshold and the flinging gesture.
Returning `true` will trigger the dismiss animation of the widget, even if the threshold is not reached.

The object `details` can be accessed to check for `velocity`, `progress` (relative distance dragged), `dragExtent` (distance dragged in logical pixels) or `reached` (bool value indicating whether the threshold has been reached)

### Linked issue
Closes https://github.com/flutter/flutter/issues/151223

PR #181644, opened 2026-01-29
Add allow fling property fix 151223

### LINKED ISSUE #151223: Ability to disable the default flinging behaviour on `Dismissible` widget
### Use case

The default flinging behaviour implemented in the `Dismissible` widget often results in undesired triggering of the dismiss action when used in a scrollable content (eg. dismissible tiles in a `ListView`)

### Proposal

Adding an optional `allowFlinging` parameter (defaulting to `true` for compatibility) that would disable the default flinging behaviour could fix this.

CadenFehr:
### Description:
This adds a boolean property `allowFling` to the `Dismissible` widget to control the fling behavior. This is desirable as it's easy to do a dismissal accidentally when scrolling. I've kept the implementation simple and left the original behavior by default.

### Linked Issue:
Closes #151223


Piinks (commented):
Hey @CadenFehr thanks for contributing! Did you see the past PR that tried to resolve this? There was some discussion about how we thought we might solve this in https://github.com/flutter/flutter/pull/151224#issuecomment-2316477348, curious to know WDYT?

CadenFehr:
> Hey @CadenFehr thanks for contributing! Did you see the past PR that tried to resolve this? There was some discussion about how we thought we might solve this in [#151224 (comment)](https://github.com/flutter/flutter/pull/151224#issuecomment-2316477348), curious to know WDYT?

I did skim over the previous PR. Looking into it further now, I notice this simpler solution had previously been proposed. I haven't checked out that branch yet to see it's current state.

I can see the additional value and flexibility that may be provided via a callback parameter. What do you think?

Do you know if removing the dismissThresholds map is still on the roadmap?

Piinks:
Thanks for taking a look!

> I can see the additional value and flexibility that may be provided via a callback parameter. What do you think?

I think it might be worth revisiting. In my experience, boolean flags can later become rather limiting, or end up deprecated and replaced later.

> Do you know if removing the dismissThresholds map is still on the roadmap?

I don't think anyone is currently working on this.

Piinks:
Hey @CadenFehr is this a change you'd still like to work on?

pubspec.lock
```diff
@@ -534,10 +534,10 @@ packages:
     dependency: transitive
     description:
       name: objective_c
-      sha256: "9922a1ad59ac5afb154cc948aa6ded01987a75003651d0a2866afc23f4da624e"
+      sha256: "983c7fa1501f6dcc0cb7af4e42072e9993cb28d73604d25ebf4dab08165d997e"
```
Piinks: It looks like this is an unintended change?

Piinks (commented):
Thanks for coming back! :) Looks like pulling in those changes worked out great.

> I see from the old PR (https://github.com/flutter/flutter/pull/151224) it was on the wish list to also deprecate dismissThresholds, and update related APIs. Is this still something we'd want completed in a separate PR?

Yeah, I can see how with this addition the API is a bit "muddy" now. shouldTriggerDismiss solves the "fling" problem, but it creates a secondary problem where the widget's drag state (the threshold) and its dismiss state (the callback) can disagree.

shouldTriggerDismiss is essentially an "imperative" version of the "declarative" dismissThresholds.


   * If we keep both we have two sources of truth for the same event. Developers have to manually sync their shouldTriggerDismiss logic with their dismissThresholds map to avoid the UI flickering or lying to the user.
   * If we deprecate `dismissThresholds` we would need to rethink onUpdate, as it also relies on that threshold to report reached.
   
   We should probably figure out the full picture before we land anything. What are your thoughts on this?

pubspec.lock
CadenFehr: unintended, I'll rollback the lock file.

Piinks (commented):
Thanks for sticking with this! You've identified the core challenge: we have two competing ways to define the "trigger" logic, which leads to a fragmented source of truth between the UI reporting (the threshold) and the actual dismissal (the callback).

To resolve this, I think a `DismissPolicy` object would be the best path forward. This would allow us to encapsulate both the threshold logic (declarative) and the trigger logic (including flings) into a single, extensible configuration.

The goal would be to have a single object that `Dismissible` can query for both its "reached" state and its final "trigger" decision. This ensures they can never disagree.

This is just an idea of what it could be like:

```dart
abstract class DismissPolicy {
  const DismissPolicy();

  /// The threshold (0.0 to 1.0) at which the item is considered "reached."
  /// This is used for [onUpdate] reporting and the default dismissal trigger.
  double getThreshold(DismissDirection direction) => 0.4;

  /// Whether a fling gesture should trigger a dismissal.
  bool allowFling(DismissDirection direction) => true;

  /// The final decision on whether to trigger a dismissal based on the drag details.
  /// The default implementation respects the threshold and fling settings.
  bool shouldDismiss(TriggerDismissDetails details) {
    return details.progress >= getThreshold(details.direction) 
        || (allowFling(details.direction) && details.isFling);
  }
}
```

This could create that single source of truth. `Dismissible` would use `policy.getThreshold()` for its internal `_dismissThreshold` calculation (affecting `onUpdate`) and `policy.shouldDismiss()` for the final trigger decision.

Instead of adding more boolean flags like `allowFling`, we (and developers) can subclass `DismissPolicy` or add new methods to it later without muddying the `Dismissible` constructor.

We can add a `policy` property to `Dismissible` and eventually deprecate `dismissThresholds`. For backward compatibility, the default policy can simply look up values from the old `dismissThresholds` map if it exists.

For this PR, we could start by defining `DismissPolicy` and a `StandardDismissPolicy` implementation, then update `_DismissibleState` to rely on the policy for all its decision-making.

What do you think about moving in this direction? It feels like a much more "Flutter-y" way to handle the growing complexity of this widget's behavior.

Piinks:
Hey @CadenFehr is this something you would still like to work on?

Issue #165904, opened 2025-03-25
Error when running a headless FlutterEngine in an iOS App Extension


AlexNik4:
Can a headless FlutterEngine work in iOS App Extensions? I'm trying to share Dart code between my main app and an iOS network extension via a Flutter module by loosely following [Adding iOS app extensions](https://docs.flutter.dev/platform-integration/ios/app-extensions) and [Integrate a Flutter module into iOS](https://docs.flutter.dev/add-to-app/ios/project-setup)

When initializing the FlutterEngine and trying to invoke a MethodChannel in my Extension Service:

```
let engine = FlutterEngine(name: "extension", project: nil, allowHeadlessExecution: true)
engine.run(withEntrypoint: "main")*
// Create MethodChannel
```

I get this error:

`Modifying properties of a view's layer off the main thread is not allowed: view <UIView: 0x104a57b80> with no associated or ancestor view controller;*`

Is this architecture possible? The error suggests Flutter is still trying to create UI components despite running headless.

#### Addition Details
I have a large amount of Dart code that needs to be reused in my Push Provider Extension. A lot of it is pure Dart, so I can potentially at least reuse that using ffi(I think). But a good chunk uses Flutter as well (local notifications, audio, local storage). It would be awesome to reuse all of it in the iOS extension. I was hoping that running the FlutterEngine without the UI would create a smaller footprint and still allow access to the Flutter SDK. I really can't find any direct documentation around this.

darshankawar:
@AlexNik4 
Check if [this](https://github.com/flutter/flutter/issues/119236) maybe is related / similar to your case or not.

AlexNik4:
@darshankawar Thanks for your response! Since the approach isn’t outright impossible, I tested further in a[ minimal sample project](https://github.com/AlexNik4/flutter-ios-extension) using a Packet Tunnel Network Extension (to avoid App Push Provider entitlement requirements). I ended up avoiding any additional plugins for now.
### Current Setup in the Extension Service

```
FlutterEngine Initialization (Swift):
self.flutterEngine = FlutterEngine(  
    name: "extension_engine",  
    project: nil,  
    allowHeadlessExecution: true  
)  
flutterEngine.run(  
    withEntrypoint: "mainHeadless",  
    libraryURI: "package:flutter_ios_extension/headless.dart"  
```
Confirmed the entrypoint is correct (misnamed withEntrypoint or libraryURI throw Dart_LookupLibrary errors).

### headless.dart Entrypoint:

[...]

PR #181645, opened 2026-01-29
Run Headless FlutterEngine in iOS app service extensions

Gibbo97:
This PR aims to allow the flutter engine to run from iOS service extensions. We have an "offline first" app which has extensive dart code to manage state in an SQLite database among other things. We want to reuse this code and run it from a Notification Service Extension before a notification is delivered, to ensure the app has the data the user is being notified about. Service extensions like this are a common feature.

When allowHeadlessExecution: true is set, the Flutter engine was still trying to use the merged platform/UI thread optimization. In notification service extensions, the platform's CFRunLoop isn't being pumped, causing the engine to hang.

Existing Issue: https://github.com/flutter/flutter/issues/165904

Example project https://github.com/Gibbo97/flutter_headless_engine_example
In the example when a notification is sent to the device, the method "dispatcher" in main.dart is called from NotificationService.swift, and the print statement runs successfully.
Without the engine changes this statement is never run.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -907,12 +911,20 @@ - (BOOL)createShell:(NSString*)entrypoint
...
   fml::RefPtr<fml::TaskRunner> ui_runner;
-  if (settings.enable_impeller &&
-      settings.merged_platform_ui_thread == flutter::Settings::MergedPlatformUIThread::kEnabled) {
+  // For headless execution (like notification extensions), we CANNOT use merged platform/UI thread
+  // because the platform's CFRunLoop is not being actively pumped.
+  // Force separate UI thread for headless mode.
+  bool useMergedThread =
+      settings.enable_impeller &&
+      settings.merged_platform_ui_thread == flutter::Settings::MergedPlatformUIThread::kEnabled &&
+      !self.allowHeadlessExecution;
```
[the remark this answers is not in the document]
Gibbo97: I have implemented this change

gaaclarke (commented):
This sounds a bit dicy dealing with threading.  This PR has no tests, we'll need that as a minimum.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -445,6 +445,19 @@ FLUTTER_DARWIN_EXPORT
...
+ *
+ * When set to YES, the engine will use a separate UI thread instead of merging with the platform
+ * thread. This is necessary in extension contexts where the platform's CFRunLoop is not being
+ * actively pumped.
+ *
+ * This property must be set before calling `-run` or `-runWithEntrypoint:`.
+ *
+ * Defaults to NO.
+ */
+@property(nonatomic, assign) BOOL isRunningInExtension;
```
gaaclarke, line 459: This should be a parameter in init, not a property since setting it after `run` is meaningless.

gaaclarke (commented):
I'll let the iOS team review this.  Thanks for the test.  There is always a tension when increasing the FlutterEngine API but I think you have a good case to be made here.

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
Gibbo97: Happy to shift it to a parameter if that is preferred.

The use is like this:
```
       let engine = FlutterEngine(
            name: "notification_service",
            project: flutterProject,
            allowHeadlessExecution: true
        )
        engine.isRunningInExtension = true

        self.flutterEngine = engine

        // Run the dispatcher entrypoint
        let success = engine.run(withEntrypoint: "dispatcher",
                                 libraryURI: nil,
                                 initialRoute: nil,
                                 entrypointArgs: ["notification_service_extension"])
```

In FlutterEngine.h there is a kind of similar existing example of a settable BOOL property:                                                                                                         
```
  isGpuDisabled (line 441-446):
  /**
   * Whether or not GPU calls are allowed.
   *
   * Typically this is set when the app is backgrounded and foregrounded.
   */
  @property(nonatomic, assign) BOOL isGpuDisabled;
```

engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterEngine.h
```diff
@@ -445,6 +445,19 @@ FLUTTER_DARWIN_EXPORT
  */
 @property(nonatomic, assign) BOOL isGpuDisabled;
 
+/**
+ * Whether this engine is running in an app extension context (e.g., Notification Service Extension).
+ *
+ * When set to YES, the engine will use a separate UI thread instead of merging with the platform
+ * thread. This is necessary in extension contexts where the platform's CFRunLoop is not being
+ * actively pumped.
```
hellohuanlin, line 453: what does "pump" mean. 

Could you also add details on why this is necessary?

Gibbo97: Sure, this is the first time I've worked on this so it's a lot for me to wrap my head around.

In a Service Extension the Platform Thread is not being looped through. In a normal iOS app UIApplicationMain loops indefinitely. 

In the engine there is a check (which I have modified) which sees if it should merge the UI Thread into the Platform thread. If they are merged the dart code is never run because it is not looped through.

We need some way of forcing the two threads to be seperate if we want it to ever be able to run dart code without the Platform Thread looping. Alternatively we could potentially start manually looping the Platform Thread.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -883,9 +884,19 @@ - (BOOL)createShell:(NSString*)entrypoint
 
   SetEntryPoint(&settings, entrypoint, libraryURI);
 
+  // Determine whether to use a merged platform/UI thread. We need a separate UI thread when:
+  // - Impeller is disabled (merged thread only works with Impeller)
+  // - merged_platform_ui_thread setting is not enabled
+  // - Running in an extension (e.g., notification service extension) where the platform's
+  //   CFRunLoop is not being actively pumped
```
hellohuanlin, line 891: is there any apple documentation on this? does this apply to all extensions (e.g. share extension) or is it just for notification service extension

PR #181664, opened 2026-01-29
Make the Windows windowing_test in .ci.yaml have bringup as false

mattkae:
## What's new?
Windows windowing tests should no longer be flaky, so let's set `bringup: false` in `.ci.yaml`

## Pre-launch Checklist

- [X] I signed the [CLA].


.ci.yaml
```diff
@@ -6949,7 +6949,7 @@ targets:
       task_name: windows_startup_test
 
   - name: Windows windowing_test
-    bringup: true # Flaky https://github.com/flutter/flutter/issues/177715
+    bringup: false
```
loic-sharma: Silly nitpick: the default value for `bringup` is `false`, so usually we just omit `bringup: false`.

```suggestion
```

loic-sharma (approved):
Thanks!

.ci.yaml
mattkae: Sure! Why not :)

PR #181703, opened 2026-01-30
Remove obsolete null checks from style guide

nate-thegrate:
A couple of the style guide examples were written prior to Dart's null safety era, so this PR aims to update those examples.

[Test-exempt](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests) because it only affects the `.md` file.

### DIFF
--- a/docs/contributing/Style-guide-for-Flutter-repo.md
+++ b/docs/contributing/Style-guide-for-Flutter-repo.md
@@ -711,7 +711,7 @@ When referencing a parameter, use backticks. However, when referencing a paramet
   /// The [bar] argument allows the baz to quux.
   ///
   /// The `baz` argument must be greater than zero.
-  Foo({this.bar, int baz}) : assert(baz > 0);
+  Foo({required this.bar, required int baz}) : assert(baz > 0);
 ```
 
 Avoid using terms like "above" or "below" to reference one dartdoc section from another. Dartdoc sections are often shown alone on a Web page, the full context of the class is not present.
--- a/docs/contributing/Style-guide-for-Flutter-repo.md
+++ b/docs/contributing/Style-guide-for-Flutter-repo.md
@@ -706,12 +706,12 @@ When referencing a parameter, use backticks. However, when referencing a paramet
 ```dart
 // GOOD
 
-  /// Creates a foobar, which allows a baz to quux the bar.
+  /// Creates a foobar.
   ///
-  /// The [bar] argument must not be null.
+  /// The [bar] argument allows the baz to quux.
   ///
   /// The `baz` argument must be greater than zero.
-  Foo({ this.bar, int baz }) : assert(bar != null), assert(baz > 0);
+  Foo({this.bar, int baz}) : assert(baz > 0);
 ```
 
 Avoid using terms like "above" or "below" to reference one dartdoc section from another. Dartdoc sections are often shown alone on a Web page, the full context of the class is not present.
@@ -924,7 +924,6 @@ the following pattern:
 TheType get theProperty => _theProperty;
 TheType _theProperty;
 void set theProperty(TheType value) {
-  assert(value != null);
   if (_theProperty == value) {
     return;
   }


docs/contributing/Style-guide-for-Flutter-repo.md
```diff
@@ -918,7 +918,6 @@ the following pattern:
 TheType get theProperty => _theProperty;
 TheType _theProperty;
```
Piinks, line 925: I could see how `TheType` could be a nullable property. Maybe we should make it explicit as `TheType?`? I think it's a good example to have the assertion in the setter body.

docs/contributing/Style-guide-for-Flutter-repo.md
```diff
@@ -706,12 +706,12 @@ When referencing a parameter, use backticks. However, when referencing a paramet
...
 // GOOD
 
-  /// Creates a foobar, which allows a baz to quux the bar.
+  /// Creates a foobar.
   ///
-  /// The [bar] argument must not be null.
+  /// The [bar] argument allows the baz to quux.
   ///
   /// The `baz` argument must be greater than zero.
-  Foo({ this.bar, int baz }) : assert(bar != null), assert(baz > 0);
```
Piinks, line 714: These could also be nullable, the sample code here does not provide enough information to determine if they are or not. IMO, I think this one is ok as is.

Piinks (commented):
Thanks for contributing some tweaks here.

docs/contributing/Style-guide-for-Flutter-repo.md
nate-thegrate: In this example `_theProperty` is non-nullable, but you're right that the property could just as easily have been created as a nullable one.

However, I don't think it's good to have the assertion in the setter body; based on how this example is currently written, there's no way for the non-nullable `value` to be `null`.

docs/contributing/Style-guide-for-Flutter-repo.md
nate-thegrate: I disagree here as well: the best practice is to avoid null-asserts in constructors.

```dart
class Foo {
  Foo({this.bar}) : assert(bar != null); // BAD

  Object? bar;
}
```

Let's say that `bar` is a nullable field but you still want the constructor argument to be non-nullable. You could do it like so:

```dart
class Foo {
  Foo({required Object this.bar}); // GOOD

  Object? bar;
}
```

It's much better for null-safety to be enforced by a type annotation than by using an `assert`, since the latter makes it much easier for mistakes to go unnoticed until runtime.

nate-thegrate (commented):
Sorry I ended up disagreeing on both points 😅

I really appreciate the review!

docs/contributing/Style-guide-for-Flutter-repo.md
Piinks: Right, I think we might be over-indexing on code that lacks enough context to actually suggest this.

docs/contributing/Style-guide-for-Flutter-repo.md
Piinks: Yes. What I was suggesting was that you update this to be an explicitly nullable type to remove the ambiguity and keep the assertion.

docs/contributing/Style-guide-for-Flutter-repo.md
nate-thegrate: Fair enough haha

But the point I was trying to make is: validating a constructor argument with an `assert(bar != null)` is never a good practice, so I'd like this example to be changed to reflect that.

docs/contributing/Style-guide-for-Flutter-repo.md
nate-thegrate: Okay, so then it would look like this:

```dart
TheType? get theProperty => _theProperty;
TheType? _theProperty;
void set theProperty(TheType? value) {
  assert(value != null);
  if (_theProperty == value) {
    return;
  }
  _theProperty = value;
  markNeedsWhatever(); // the method to mark the object dirty
}
```

This would work, though it has the downside of being a bit more verbose. (In my opinion, being concise & readable is even more important in the style guide than anywhere else in this repo.)

Is there a downside to how I have it currently?

docs/contributing/Style-guide-for-Flutter-repo.md
Piinks: I appreciate you want to make the code snippet as accurate as possible. Given that this code snippet is meant to facilitate guidance around documenting parameters, and not to illustrate best practices around unnecessary null-asserts, does this change improve on the intended guidance?

docs/contributing/Style-guide-for-Flutter-repo.md
Piinks: Similar to the other example, I would consider the intention of the guidance here. This is about providing guidelines for performing dirty checks in setters. Does this change improve on that guidance? IMO, I don't think verbosity is a bad thing in docs meant to educate for working on the framework.

nate-thegrate: > IMO, I don't think verbosity is a bad thing in docs meant to educate for working on the framework.

I suppose we could agree to disagree here. There's a good reason "TL;DR" is so prevalent throughout the internet: if something is more verbose, the opportunity cost of reading & understanding is higher, ergo people become less likely to do it. In the case of this example, removing irrelevant code is a straight improvement IMO.

I really appreciate the time you've spent helping me out here. Since this PR would require two approvals to merge, maybe now would be a good time for another reviewer to step in with a fresh perspective.

docs/contributing/Style-guide-for-Flutter-repo.md
nate-thegrate: Yes it does!

Originally, this example documentation said:

```dart
  /// The [bar] argument must not be null.
```

At the time it was written, there wasn't a good way to enforce that a non-null argument is passed into a class constructor, so the best practice was to include this advice where applicable.

But thanks to null safety, this advice is no longer warranted: setting the class member as non-nullable means that static analysis will enforce it.

docs/contributing/Style-guide-for-Flutter-repo.md
Piinks: > There's a good reason "TL;DR" is so prevalent throughout the internet: if something is more verbose, the opportunity cost of reading & understanding is higher, ergo people become less likely to do it. In the case of this example, removing irrelevant code is a straight improvement IMO.

I am not sure this is really relevant. This is a change specifically within the context of the Flutter org. Not the internet at large. :)

I will leave this to another reviewer to make a pass. Honestly, I don't know this change is worth this much debate.

justinmc: I agree that this example should change one way or another, because the assertion is no longer necessary and could confuse people. Also, the assertion is not important to this section of the style guide. My interpretation of the two options are:

Option 1:

```diff
TheType _theProperty;
void set theProperty(TheType value) {
-   assert(value != null);
  if (_theProperty == value) {
    return;
  }
  _theProperty = value;
  markNeedsWhatever(); // the method to mark the object dirty
}
```

Option 2:

```diff
-TheType _theProperty;
+TheType? _theProperty;
-void set theProperty(TheType value) {
+void set theProperty(TheType? value) {
  assert(value != null);
  if (_theProperty == value) {
    return;
  }
  _theProperty = value;
  markNeedsWhatever(); // the method to mark the object dirty
}
```

I like option 1. Option 2 is oddly specific and could distract from the point of this section. You might do something like it if _theProperty was supposed to be null during initialization or something and then never nonnull again... but then why assert instead of making the setter nonnullable? These are questions that the reader shouldn't be thinking about :)

docs/contributing/Style-guide-for-Flutter-repo.md
justinmc: I think that `assert(bar != null)` should go since it is not best practice. I don't want to confuse users or LLMs that might read it.

justinmc (approved):
LGTM 👍 . I say let's update these to modern Dart and move on, no need to debate these too much. Even if the examples aren't perfect they will now avoid being outdated.

AbdeMohlbi (approved):

PR #181705, opened 2026-01-30
Remove `DisposableBuildContext` type argument

### LINKED ISSUE #181935: `DisposableBuildContext` has an unused type argument
The [DisposableBuildContext](https://main-api.flutter.dev/flutter/widgets/DisposableBuildContext-class.html) class has an type argument that the API doesn't use for anything.

<br>

This results in a few false negatives for the [omit_obvious_local_variable_types](http://dart.dev/lints/omit_obvious_local_variable_types) linter rule:

https://github.com/flutter/flutter/blob/ba5ba0d9230ad03b5ac406a96c6314eb396307bc/packages/flutter/test/widgets/disposable_build_context_test.dart#L17

https://github.com/flutter/flutter/blob/f916dd6887bf0242132ab6e86bd0a77bd6ff2089/packages/flutter/test/widgets/scroll_aware_image_provider_test.dart#L45

nate-thegrate:
[**DisposableBuildContext**](https://main-api.flutter.dev/flutter/widgets/DisposableBuildContext-class.html) has an unused type argument, so this PR aims to clean it up a bit.

I believe this PR is [test-exempt](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests) because it only removes dead code.

resolves #181935

### DIFF
--- a/packages/flutter/lib/src/widgets/disposable_build_context.dart
+++ b/packages/flutter/lib/src/widgets/disposable_build_context.dart
@@ -22,14 +22,14 @@ import 'framework.dart';
 ///
 /// This object will not hold on to the [State] after disposal.
 @optionalTypeArgs
-class DisposableBuildContext<T extends State> {
+class DisposableBuildContext {
   /// Creates an object that provides access to a [BuildContext] without leaking
   /// a [State].
   ///
   /// Creators must call [dispose] when the [State] is disposed.
   ///
   /// [State.mounted] must be true.
-  DisposableBuildContext(T this._state)
+  DisposableBuildContext(State this._state)
     : assert(
         _state.mounted,
         'A DisposableBuildContext was given a BuildContext for an Element that is not mounted.',
@@ -37,7 +37,7 @@ class DisposableBuildContext<T extends State> {
     assert(debugMaybeDispatchCreated('widgets', 'DisposableBuildContext', this));
   }
 
-  T? _state;
+  State? _state;
 
   /// Provides safe access to the build context.
   ///
--- a/packages/flutter/lib/src/widgets/disposable_build_context.dart
+++ b/packages/flutter/lib/src/widgets/disposable_build_context.dart
@@ -21,7 +21,6 @@ import 'framework.dart';
 ///   2. They call [dispose] from [State.dispose].
 ///
 /// This object will not hold on to the [State] after disposal.
-@optionalTypeArgs
 class DisposableBuildContext {
   /// Creates an object that provides access to a [BuildContext] without leaking
   /// a [State].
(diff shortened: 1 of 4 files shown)


packages/flutter/lib/src/widgets/disposable_build_context.dart
```diff
@@ -21,23 +21,22 @@ import 'framework.dart';
...
+  DisposableBuildContext(State this._state)
     : assert(
         _state.mounted,
         'A DisposableBuildContext was given a BuildContext for an Element that is not mounted.',
       ) {
     assert(debugMaybeDispatchCreated('widgets', 'DisposableBuildContext', this));
   }
 
-  T? _state;
+  State? _state;
```
navaronbracke, line 39: Should we provide the type argument here? Otherwise it is `State<dynamic>` ?

```suggestion
  State<StatefulWidget>? _state;
```

nate-thegrate: That's a good question—luckily, when the type argument has an `extends` it won't default to `dynamic`.

```dart
class A<T> {}
class B<T extends Object?> {}

void foo() {
  final a = A(); // inferred as A<dynamic>
  final b = B(); // inferred as B<Object?>
}
```

Since the framework declares it as `class State<T extends StatefulWidget>`, we can omit the type argument and it'll default to `State<StatefulWidget>`.

packages/flutter/lib/src/widgets/disposable_build_context.dart
```diff
@@ -21,23 +21,22 @@ import 'framework.dart';
 ///   2. They call [dispose] from [State.dispose].
 ///
 /// This object will not hold on to the [State] after disposal.
-@optionalTypeArgs
-class DisposableBuildContext<T extends State> {
+class DisposableBuildContext {
```
Piinks, line 24: Since this is public API, is it possible it is not dead code and is being used by someone else out there? I am not sure of that means this is a breaking change.

Piinks (commented):
Is there an open issue this PR is looking to resolve?

packages/flutter/lib/src/widgets/disposable_build_context.dart
nate-thegrate: It's [dead code](https://en.wikipedia.org/wiki/Dead_code) in the sense that it never does anything useful, but you have a good point that someone out there might have written the type annotation somewhere in their codebase.

As far as our breaking change policy goes, I believe we're good as long as `customer_testing` passes.

nate-thegrate (commented):
> Is there an open issue this PR is looking to resolve?

Totally neglected to make one, my bad! Just opened #181935.

nate-thegrate:
The motivation here comes from the first sentence of the [style guide](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md):

> Optimize for readability.

<br>

For any hypothetical developer who uses a disposable build context with a type argument, IMO it should be easy enough for them to delete it. But it's totally up to you guys whether you think this change is worth it :)

PR #181706, opened 2026-01-30
Bugfix in the `absorbTicker` method

### LINKED ISSUE #181699: AnimationController `resync()` sometimes throws an assertion error
If an animation controller starts its ticker and then immediately calls [`resync()`](https://api.flutter.dev/flutter/animation/AnimationController/resync.html), the framework incorrectly assumes that the ticker was disposed of and throws an error.

I'll submit a PR with a test to demonstrate this.

nate-thegrate:
Currently, `absorbTicker` assumes that between `_future` and `_startTime`, if one of them is null and the other is not then the ticker must have been disposed of.

https://github.com/flutter/flutter/blob/9b6f15e67d181fcbc4e3ccc7387aa0da8a3d19e6/packages/flutter/lib/src/scheduler/ticker.dart#L329-L337

I discovered that starting an animation and then immediately calling `resync()` leads to a false positive, so this PR tweaks the `absorbTicker` method accordingly.

<br>

resolves #181699


packages/flutter/lib/src/scheduler/ticker.dart
```diff
@@ -332,7 +332,7 @@ class Ticker {
     assert(_startTime == null);
     assert(_animationId == null);
     assert(
-      (originalTicker._future == null) == (originalTicker._startTime == null),
+      (originalTicker._future == null) || (originalTicker._startTime == null),
```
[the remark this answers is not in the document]
nate-thegrate: Copy-paste error 😅

Thank you bot

packages/flutter/lib/src/scheduler/ticker.dart
```diff
@@ -332,7 +332,7 @@ class Ticker {
     assert(_startTime == null);
     assert(_animationId == null);
     assert(
-      (originalTicker._future == null) == (originalTicker._startTime == null),
+      (originalTicker._future != null) || (originalTicker._startTime == null),
```
victorsanni, line 335: This makes sense to me, _future can be null (so the ticker is disposed and so _startTime must be null). But _startTime can be null even when _future is non-null.

victorsanni (approved):

packages/flutter/lib/src/scheduler/ticker.dart
justinmc: The logical difference before/after is that after, this case does not trigger the assertion:

`originalTicker._future != null && originalTicker._startTime == null`

That seems to make sense because that's not the case after dispose, it's the case when you create the Ticker and then pass it to absorbTicker before starting it, which seems to be exactly what @nate-thegrate describes in the issue.

justinmc (approved):
LGTM 👍

Issue #181700, opened 2026-01-30
Proposal: prefer `ValueListenable` over `Animation` in API signatures


nate-thegrate:
[**ValueListenable**](https://main-api.flutter.dev/flutter/foundation/ValueListenable-class.html) is similar to [**Animation**](https://main-api.flutter.dev/flutter/animation/Animation-class.html), but it doesn't have a concept of "forward vs. reverse" so there's no need to implement an animation status.

Sometimes you have a transition that doesn't have an explicit "dismissed" or "completed" state, such as a draggable UI component that moves based on user input. For these cases, your code can be much simpler when you don't use any `Animation` objects:

```dart
class _MyWidgetState extends State<MyWidget> {
  final position = ValueNotifier(Offset.zero);

  @override
  Widget build(BuildContext context) {
    return SlideTransition(position: position, child: widget.child);
  }
}
```

In order to make this possible, [**SlideTransition**](https://main-api.flutter.dev/flutter/widgets/SlideTransition-class.html) needs to be tweaked so the `position` field is a `ValueListenable<Offset>` instead of an `Animation<Offset>`.

nate-thegrate:
@dkwingsmt you're absolutely right that the `Animation` class itself doesn't require a ticker, though I'm not aware of any other concrete subtype of `Animation` we currently have that can be used in a manner similar to `ValueNotifier` or `AnimationController`.

It's also true that a `ValueListenable` can't access properties like `isDismissed`. For widgets like [**DefaultTextStyleTransition**](https://main-api.flutter.dev/flutter/widgets/DefaultTextStyleTransition-class.html), in most cases I don't think it makes a whole lot of sense to be able to query whether or not it's "dismissed", so I feel like using a `ValueListenable` instead would be the right call.

I agree that often you don't need `drive` at all, especially since it's a method in the base `Animation<T>` class that typecasts itself as an `Animation<double>`. Maybe if the style guide is updated to allow extension methods, we could instead have it implemented that way for better type safety.

dkwingsmt:
So let's split this topic into two.

1. Do you still want to do anything with `AnimationController`? If we only focus on `Animation` vs `ValueListenable` then can you remove the mention of `ticker` in the issue body for the clarity of this project? If you still do shall we put it into another issue?
2. WRT `Animation`, I'm also not a fan of the `drive`. Feel free to explore its usage and/or if there's anything we can do about it that interests you.

dkwingsmt:
A 3rd topic:
[...]

PR #181707, opened 2026-01-30
Export `ValueListenable` in widgets library

nate-thegrate:
#181700 proposes adopting the [**ValueListenable**](https://main-api.flutter.dev/flutter/foundation/ValueListenable-class.html) API for a few transition widgets.

The goal of this PR is to allow future changes to be made more cleanly and thus easier to review.

### DIFF
--- a/examples/api/lib/material/selection_area/selection_area.1.dart
+++ b/examples/api/lib/material/selection_area/selection_area.1.dart
@@ -2,7 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
 /// Flutter code sample for [SelectionArea].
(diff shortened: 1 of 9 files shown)


nate-thegrate:
@chunhtai currently there isn't a design doc. I remember some advice that Matan [shared a while back](https://discord.com/channels/608014603317936148/608018585025118217/1319731216169177190) that seems relevant here:

> A design doc is supposed to be "I need cross-cutting feedback from multiple teams/areas to proceed"

<br>

If/when this PR lands, solving issue #181700 is going to be very simple & straightforward. Here's what that upcoming PR would eventually look like:

- https://github.com/nate-thegrate/flutter/pull/11

Feel free to take a look, and then if you'd still like to see a design doc just let me know. Thanks!

nate-thegrate:
Okay, sounds good! I guess it's understandable that this might be viewed as a conceptual shift for the framework, and you're right that `Animation.fromValueListenable` currently exists as a workaround.

I can get a design doc put together, with the caveat that I'm currently working full-time so it might not be ready for another day or two :)

dkwingsmt (approved):
LGTM

PR #181716, opened 2026-01-30
Fix textscaler clamp assertion error

### LINKED ISSUE #174828: textScalar.clamp not working
### Steps to reproduce

            final mediaQuery = MediaQuery.of(context);
            final textScalar = mediaQuery.textScaler;
            final mediaQueryData = mediaQuery.copyWith(
              textScaler: textScalar.clamp(
                minScaleFactor: 1,
                maxScaleFactor: 1.1,
              ),
            );

### Expected results

Failed assertion: line 118 pos 80: 'maxScale > minScale': is not true.

### Actual results

Should be working fine as maxScaleFactor > minScaleFactor 

### Code sample

```dart
            final mediaQuery = MediaQuery.of(context);
            final textScalar = mediaQuery.textScaler;
            final mediaQueryData = mediaQuery.copyWith(
              textScaler: textScalar.clamp(
                minScaleFactor: 1,
                maxScaleFactor: 1.1,
              ),
            );
```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
'package:flutter/src/painting/text_scaler.dart': Failed assertion: line 118 pos 80: 'maxScale > minScale': is not true.


```

</details>

### Flutter Doctor output

```console

<img width="1181" height="619" alt="Image" src="https://github.com/user-attachments/assets/8278a35c-6df0-4576-bada-05d9c5fd6e0d" />

```

</details>

Valansch:
Fix `TextScaler.clamp` assertion failure when clamping already-clamped scalers

Fixes #174828

## Description

When calling `clamp()` on an already-clamped `TextScaler`, the `_ClampedTextScaler.clamp()` method was checking if `minScaleFactor == maxScaleFactor` **before** calculating the actual new min/max values. This caused assertion failures (`'maxScale > minScale': is not true`) when the calculated `newMinScale` and `newMaxScale` became equal after applying `max()`/`min()` operations.

### Example scenario that triggers the bug:
```dart
final textScaler = MediaQuery.of(context).textScaler;
// First clamp - creates a _ClampedTextScaler
textScaler.clamp(minScaleFactor: 1.0, maxScaleFactor: 1.1);
// Internal framework code calls clamp() again (e.g., in BottomNavigationBar)
// This could result in newMinScale == newMaxScale, triggering the assertion
```

### The fix:
Calculate `newMinScale` and `newMaxScale` **first**, then check for equality. If they are equal, return a `_LinearTextScaler` instead of creating a `_ClampedTextScaler` that would fail its constructor assertion.

### Important
This breaks before undefined behaviour. We now through an assertion error when new bounds are out of bounds even when min and max are equal. For more detail view my comment on this [here](https://github.com/flutter/flutter/issues/174828#issuecomment-3822766018)


packages/flutter/lib/src/painting/text_scaler.dart
```diff
@@ -132,9 +132,11 @@ final class _ClampedTextScaler implements TextScaler {
 
   @override
   TextScaler clamp({double minScaleFactor = 0, double maxScaleFactor = double.infinity}) {
-    return minScaleFactor == maxScaleFactor
-        ? _LinearTextScaler(minScaleFactor)
-        : _ClampedTextScaler(scaler, max(minScaleFactor, minScale), min(maxScaleFactor, maxScale));
+    final double newMinScale = max(minScaleFactor, minScale);
+    final double newMaxScale = min(maxScaleFactor, maxScale);
+    return newMinScale == newMaxScale
```
victorsanni: Instead of merging the outer and inner bounds, we should use the provided bounds (`minScaleFactor` and `maxScaleFactor`) to clamp and create a new class instance.

Valansch: Change it to your suggestion. This is a breaking change now.

packages/flutter/test/painting/text_scaler_test.dart
```diff
@@ -57,5 +57,20 @@ void main() {
       expect(scaler1, scaler3);
       expect(scaler1, isNot(scaler2));
     });
+
+    testWidgets('clamping', (WidgetTester tester) async {
```
victorsanni: Rename the test to make it easier to debug in the future. Maybe to something like "Initial clamping bounds are ignored when re-clamped".

packages/flutter/test/painting/text_scaler_test.dart
```diff
@@ -57,5 +57,20 @@ void main() {
...
+      final TextScaler defaultScaler = MediaQueryData.fromView(tester.view).textScaler;
+      final TextScaler outerClampedScaler = defaultScaler.clamp(
```
victorsanni: Instead of "outer" and "inner" terminology, can we use more obvious/relevant terms? Like 'initial'/'first' for 'outer' or 'final'/'second'/'reclamped' for 'inner' etc? To make it obvious that this is clamping an already clamped textscaler.

victorsanni (commented):
Can you run `dart format` on all files in this PR to fix failing checks?

packages/flutter/test/painting/text_scaler_test.dart
Valansch: Right. Before there were outer and inner bounds. Now these don't make sense anymore. Will fix1

packages/flutter/test/painting/text_scaler_test.dart
Valansch: Changed

victorsanni (approved):

victorsanni (approved):
To fix, run `dart format packages/flutter/test/painting/text_scaler_test.dart`

Valansch:
@victorsanni Resolved

packages/flutter/lib/src/painting/text_scaler.dart
```diff
@@ -134,7 +132,7 @@ final class _ClampedTextScaler implements TextScaler {
   TextScaler clamp({double minScaleFactor = 0, double maxScaleFactor = double.infinity}) {
     return minScaleFactor == maxScaleFactor
         ? _LinearTextScaler(minScaleFactor)
-        : _ClampedTextScaler(scaler, max(minScaleFactor, minScale), min(maxScaleFactor, maxScale));
+        : _ClampedTextScaler(scaler, minScaleFactor, maxScaleFactor);
```
LongCatIsLooong: I don't think this is correct. When you compose 2 clampers the first one shouldn't be ignored.

Valansch: View this comment and make up you mind you two :)
https://github.com/flutter/flutter/issues/174828#issuecomment-3844151880

I implemented the fix first respecting the old bounds. Then i was told to ignore them, now not?

LongCatIsLooong: Ah I thought by "discard" @victorsanni meant the `this.minScale` and `this.maxScale` should be removed from the expression. But with this implementation I think the first clamp is just going to be ignored and that would create some surprising behaviors. I would expect text scalers compose like functions so `clamp2 . clamp1 === clamp2 . identityScaler . clamp1` (where identifyScaler is just a text scaler that returns the input text size as-is).

Valansch: But what do you do it the clamps are incompatible? Please view my comment on the issue i explained my thoughts there

LongCatIsLooong: If `clamp1` has a range of `[1, 2]` and `clamp2` `[3, 4]`, what should the output size be when the input font size `9` goes through `clamp1` and then `clamp2`? I'm expecting 3 instead of 4 (but I think this implementation would return 4).  

The developer doesn't know the runtime type of the `TextScaler` they're calling `clamp` on (at least in most cases). And with this implementation I believe when you call `clamp` the exact behavior will depend on the runtime type of the receiver, if the receiver is a `_ClampTextScaler` then the receiver's scaling logic will be ignored but if it's not then the clamping will be applied after the receiver's scaling logic is applied.

Valansch: @LongCatIsLooong I now implemented a version that is applying first clamp1 then clamp2. Please check it.

LongCatIsLooong: I guess what I meant was just that line 135`: _ClampedTextScaler(scaler, minScaleFactor, maxScaleFactor);` probably should be `: _ClampedTextScaler(this, minScaleFactor, maxScaleFactor);` instead, so `this` is applied. @victorsanni thoughts on whether `clamp1.clamp(newMin, newMax)` should completely ignore `clamp1`?

victorsanni: I think passing `this` into the new `_ClampedTextScaler` makes sense to me, that way we can guarantee correct min/max values while clamping and reclamping without ignoring the first bounds. Thanks @LongCatIsLooong for the clarification here.

packages/flutter/lib/src/painting/text_scaler.dart
```diff
@@ -132,9 +132,11 @@ final class _ClampedTextScaler implements TextScaler {
...
   @override
   TextScaler clamp({double minScaleFactor = 0, double maxScaleFactor = double.infinity}) {
-    return minScaleFactor == maxScaleFactor
-        ? _LinearTextScaler(minScaleFactor)
-        : _ClampedTextScaler(scaler, max(minScaleFactor, minScale), min(maxScaleFactor, maxScale));
+    final double newMinScale = max(minScale, minScaleFactor);
+    final double newMaxScale = min(maxScale, maxScaleFactor);
+    return newMinScale >= newMaxScale
+        ? TextScaler.linear(newMinScale)
+        : _ClampedTextScaler(scaler, newMinScale, newMaxScale);
```
victorsanni: Should we change this to 

```suggestion
        : _ClampedTextScaler(this, newMinScale, newMaxScale);
```

As per @LongCatIsLooong's suggestions?

packages/flutter/lib/src/painting/text_scaler.dart
Valansch: I'm not sure why you would want to do that. Passing "this" will nest the scalers, causing each scale call to call scale on the entire stack.
The current implementation has the same result, but requires less invocations. I'm not sure if the difference is neglicable. It probably is.

If you insist i can change it.

Valansch: Thinking about this again: My current approach, as the previous (bugged) version is commutative. But clamping is not. So the current approach is not working. I will add another test and consider your suggestion.

Valansch: I tried to implement your approach to just nest the clamedScalers like this:
```
  @override
  TextScaler clamp({double minScaleFactor = 0, double maxScaleFactor = double.infinity}) {
    return minScaleFactor == maxScaleFactor
        ? TextScaler.linear(minScaleFactor)
        : _ClampedTextScaler(this, minScaleFactor, maxScaleFactor);
  }
```

Unfortunately that breaks equality, because only the last clamps are compared. Now I can fix the equality by implementing a numeric solution to this problem, but then why not just implement the numeric solution directly in the clamp and don't nest? This is in addition to my points raised above.

LongCatIsLooong: > Unfortunately that breaks equality, because only the last clamps are compared

Could you elaborate? The "inner" scaler is also compared in the `==` implementation. Also it's correct to always return `false` for `TextScaler.==` implementations because it's just an optimization to prevent unnecessary rebuilds. Function equality is inherently tricky so the documentation only asks that subclasses should identify "equal" text scalers on a best effort basis.

Valansch: Nesting the clamed scalers can result in the same scalers not being equal. Example:
```
final scaler = TexScaler.linear(2)
scaler.clamp(2,3) != scaler.clamp(1,4).clamp(2,3)

scaler.clamp(1,5).clamp(2,3) != scaler.clamp(1,4).clamp(2,3)
```

I believe those should be equal.

LongCatIsLooong: `TextScaler.==` is only used by the framework to skip unnecessary widget rebuilds so you can always choose to return `false`, it's totally optional. But if you really want to prevent rebuilds as much as possible I think you can switch to the original implementation (union of the two clamp intervals) when the intervals overlap.

packages/flutter/lib/src/painting/text_scaler.dart
```diff
@@ -132,9 +132,14 @@ final class _ClampedTextScaler implements TextScaler {
...
   TextScaler clamp({double minScaleFactor = 0, double maxScaleFactor = double.infinity}) {
-    return minScaleFactor == maxScaleFactor
-        ? _LinearTextScaler(minScaleFactor)
-        : _ClampedTextScaler(scaler, max(minScaleFactor, minScale), min(maxScaleFactor, maxScale));
+    final double newMinScale = max(minScale, minScaleFactor);
+    final double newMaxScale = min(maxScale, maxScaleFactor);
+
+    if (newMaxScale <= newMinScale) {
+      // Ranges don't overlap or collapse to a single point.
+      return TextScaler.linear(clampDouble(minScale, minScaleFactor, maxScaleFactor));
```
LongCatIsLooong, line 145: When `newMaxScale < newMinScale` I don't think this matches the API docs: https://main-api.flutter.dev/flutter/painting/TextScaler/clamp.html

Valansch: Do you mean for the inputs `minScaleFactor` `maxScaleFactor`? You are right this implementation removes the assert when maxScaleFactor > minScaleFactor.

I will expend the test and fix it.

But the goal of this `newMaxScale <= newMinScale` is to check if the new and old bounds overlap. So this condition is required here.

Valansch: I fixed it in a new commit and added more tests. Now it's consistent with the other two `clamp` functions

LongCatIsLooong (approved):
LGTM. Thank you for fixing this!

victorsanni (approved):

Issue #181717, opened 2026-01-30
Display P3 wide gamut color conversion is incorrect


westito:
### Steps to reproduce

**Reproduce:** Compare a solid P3 PNG image with equivalent color painted with `Color.from(..., colorSpace: ColorSpace.displayP3)`

**Problem:** The P3-to-sRGB conversion applies a 3x4 affine matrix directly to gamma-encoded color values. This is mathematically wrong because the color space conversion matrix must operate in linear light. The result is colors that are barely distinguishable from sRGB instead of showing the full P3 gamut difference.

**Fix:** Replace the single matrix multiply with the correct 3-step pipeline: decode gamma to linear (sRGB EOTF), apply the 3x3 P3-to-sRGB matrix in linear space, encode back to gamma (sRGB OETF). Both P3 and sRGB use the same sRGB transfer function. Extended range (negative values from out-of-gamut colors) is handled by mirroring the transfer function.

**Performance impact of the fix:** negligible. The conversion runs once per paint setup when a Dart Color with P3 colorspace is read into the engine (ReadColor in paint.cc), not per frame or per pixel. It adds six pow() calls (three for linearize, three for gamma encode) per P3 color, which is on the order of nanoseconds compared to the milliseconds spent rendering a frame.

### Expected results

You see the same colors (screenshots attached). macOS has no P3 support in Flutter right now, but I also implement that to see differences more clearly. With camera it is barely visible. Can't take proper screenshot on device because it is converted to sRGB. 

### Actual results

Painted color has different shades 

### Code sample

<details open><summary>Code sample</summary>

```dart
Column(
  crossAxisAlignment: CrossAxisAlignment.stretch,
  children: [
    Expanded(child: Image.asset('assets/p3.png', fit: BoxFit.cover)),
    Expanded(
      child: ColoredBox(
        color: const Color.from(
          alpha: 1,
          red: 30 / 255.0,
          green: 202 / 255.0,
          blue: 211 / 255.0,
          colorSpace: ColorSpace.displayP3,
        ),
      ),
    ),
  ],
),
```

Here is the fix (`engine/src/flutter/display_list/dl_color.cc`)
```dart
// Copyright 2013 The Flutter Authors. All rights reserved.
// Use of this source code is governed by a BSD-style license that can be
// found in the LICENSE file.

#include "flutter/display_list/dl_color.h"

#include <algorithm>
#include <cmath>

namespace flutter {

namespace {

[...]

PR #181720, opened 2026-01-30
Fix P3-to-sRGB color conversion to operate in linear light

westito:
### Problem
The P3-to-sRGB color conversion applies an affine matrix directly to gamma-encoded values in both the C++ engine (`dl_color.cc`) and Dart (`painting.dart`). This is mathematically incorrect — color space conversion matrices must operate in linear light. The result is P3 colors that appear nearly identical to sRGB instead of showing the full gamut difference.
For example, P3 `#1ECAD3` (30/255, 202/255, 211/255) converts to:
- **Old (wrong):** extendedSRGB (-0.12, 0.86, 0.87)
- **New (correct):** extendedSRGB (-0.38, 0.81, 0.84)
The red channel error of 0.26 is clearly visible — colors appear washed out instead of vivid.

### Fix
Replace the single affine matrix multiply with the correct 3-step pipeline:
1. **Linearize** — decode gamma via sRGB EOTF
2. **Transform** — apply 3x3 P3-to-sRGB matrix in linear space
3. **Encode** — re-apply gamma via sRGB OETF
Extended range (negative values from out-of-gamut colors) is handled by mirroring the transfer function.
**C++ (`dl_color.cc`):** Replace affine matrix with `p3ToExtendedSrgb()` using `double` precision.
**Dart (`painting.dart`):** Replace `_MatrixColorTransform` with `_P3ToSrgbTransform` and `_SrgbToP3Transform` classes. Also fixes the sRGB-to-P3 direction.

### Performance
Negligible. The C++ conversion runs once per paint setup in `ReadColor`, not per frame or per pixel. The Dart conversion runs once per `Color.withValues()` call.

### Tests
- **C++:** Added `ColorSpaceP3ToExtendedSRGBLinearLight` test in `dl_color_unittests.cc`
- **Dart:** Added 3 tests in `colors_test.dart`: mid-range P3→extendedSRGB, P3 green→extendedSRGB, sRGB→P3 round-trip
All new tests fail with the old code and pass with the fix. Existing tests continue to pass.

## Issue:
https://github.com/flutter/flutter/issues/181717

## Pre-launch Checklist

- [x] I signed the [CLA].
- [ ] All existing and new tests are passing. **New tests pass, can't run golden tests!**

### DIFF
--- a/engine/src/flutter/display_list/dl_color.cc
+++ b/engine/src/flutter/display_list/dl_color.cc
@@ -11,8 +11,8 @@ namespace flutter {
 
 namespace {
 
-// sRGB standard constants for transfer functions.
-// See https://en.wikipedia.org/wiki/SRGB.
+/// sRGB standard constants for transfer functions.
+/// See https://en.wikipedia.org/wiki/SRGB.
 constexpr double kSrgbGamma = 2.4;
 constexpr double kSrgbLinearThreshold = 0.04045;
 constexpr double kSrgbLinearSlope = 12.92;
@@ -20,7 +20,7 @@ constexpr double kSrgbEncodedOffset = 0.055;
 constexpr double kSrgbEncodedDivisor = 1.055;
 constexpr double kSrgbLinearToEncodedThreshold = 0.0031308;
 
-// sRGB electro-optical transfer function (gamma decode, gamma ~2.2 to linear).
+/// sRGB electro-optical transfer function (gamma decode, gamma ~2.2 to linear).
 double srgbEOTF(double v) {
   if (v <= kSrgbLinearThreshold) {
     return v / kSrgbLinearSlope;
@@ -28,7 +28,7 @@ double srgbEOTF(double v) {
   return std::pow((v + kSrgbEncodedOffset) / kSrgbEncodedDivisor, kSrgbGamma);
 }
 
-// sRGB opto-electronic transfer function (linear to gamma encode).
+/// sRGB opto-electronic transfer function (linear to gamma encode).
 double srgbOETF(double v) {
   if (v <= kSrgbLinearToEncodedThreshold) {
     return v * kSrgbLinearSlope;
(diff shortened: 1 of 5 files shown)


engine/src/flutter/display_list/dl_color.cc
```diff
@@ -5,36 +5,87 @@
...
+    1.2249401, -0.2249402, 0.0,        -0.0420569, 1.0420571,
+    0.0,       -0.0196376, -0.0786507, 1.0982884,
+};
+
+// Converts a Display P3 color (gamma-encoded) to extended sRGB (gamma-encoded).
+// Steps: P3 gamma decode -> linear P3 -> linear sRGB (via 3x3 matrix) -> sRGB
+// gamma encode.
+DlColor p3ToExtendedSrgb(const DlColor& color) {
+  // Linearize P3 values (P3 uses same transfer function as sRGB).
+  double r_lin = srgbEOTFExtended(static_cast<double>(color.getRedF()));
```
gaaclarke, line 67: no action required: do we want to just keep everything in floats?

engine/src/flutter/display_list/dl_color.cc
```diff
@@ -5,36 +5,87 @@
...
+  double g_lin = srgbEOTFExtended(static_cast<double>(color.getGreenF()));
+  double b_lin = srgbEOTFExtended(static_cast<double>(color.getBlueF()));
+
+  // Apply 3x3 P3-to-sRGB matrix in linear space.
+  double r_srgb_lin = kP3ToSrgbLinear[0] * r_lin + kP3ToSrgbLinear[1] * g_lin +
```
gaaclarke, line 72: no action required: we have matrix multiplication available in our api

engine/src/flutter/display_list/dl_color.cc
```diff
@@ -5,36 +5,87 @@
...
-                     matrix[5] * color.getGreenF() +  //
-                     matrix[6] * color.getBlueF() +   //
-                     matrix[7],                       //
-                 matrix[8] * color.getRedF() +        //
-                     matrix[9] * color.getGreenF() +  //
-                     matrix[10] * color.getBlueF() +  //
-                     matrix[11],                      //
-                 color_space);
+
+// sRGB standard constants for transfer functions.
```
gaaclarke: no action required: comments should be in doxygen format

engine/src/flutter/display_list/dl_color_unittests.cc
```diff
@@ -299,6 +299,35 @@ TEST(DisplayListColor, ColorSpaceP3ToExtendedSRGB) {
...
+// which produces significantly different results than applying a matrix
+// directly to gamma-encoded values.
+TEST(DisplayListColor, ColorSpaceP3ToExtendedSRGBLinearLight) {
+  // P3 #1ECAD3 (30/255, 202/255, 211/255)
+  // Correct extended sRGB: (-0.3764, 0.8065, 0.8372)
+  // Wrong (gamma-encoded matrix): (-0.1195, 0.8609, 0.8675)
+  // Red channel diff: 0.2569 — well above isClose tolerance of 1/256.
+  DlColor p3_teal(1.0, 30.0f / 255.0f, 202.0f / 255.0f, 211.0f / 255.0f,
+                  DlColorSpace::kDisplayP3);
+  DlColor expected_teal(1.0, -0.3764f, 0.8065f, 0.8372f,
```
gaaclarke, line 314: no action required: apple's color tool gives (-0.3778, 0.8084, 0.8358)

engine/src/flutter/lib/ui/painting.dart
```diff
@@ -3904,58 +3904,124 @@ class _ClampTransform implements _ColorTransform {
...
+  0.9668058,
+  0.0,
+  0.0170806,
+  0.0723974,
+  0.9105220,
+];
+
+// Converts Display P3 (gamma-encoded) to extended sRGB (gamma-encoded).
+// Pipeline: EOTF(decode) -> 3x3 matrix -> OETF(encode).
+class _P3ToSrgbTransform implements _ColorTransform {
```
gaaclarke, line 3970: no action required: We should call into c++ for this instead of implementing it twice (now that it is a bit more involved)

engine/src/flutter/lib/web_ui/lib/painting.dart
```diff
@@ -503,52 +503,138 @@ class _ClampTransform implements _ColorTransform {
   }
 }
 
-class _MatrixColorTransform implements _ColorTransform {
-  const _MatrixColorTransform(this.values);
+// sRGB standard constants for transfer functions.
```
gaaclarke, line 506: no action required: let's find a way to share this with web.

gaaclarke (commented):
This is awesome, thanks @westito.  I double checked all the tests with Apple's color utility and they are good.  Apple gives slightly different values but that might be captured in the epsilon of the assertions.  The test coverage is great, no need to add tests.

In fact there is nothing else you need to do.  I've made a few notes about minor cleanup and deduplication which I'll look into.  I'll commit them here or implement in a followup.

gaaclarke (approved):
I looked briefly into deduplicating the logic and it is a bit involved.  If dart:ui calls into c++ then the code can't be shared with web.  There is also no existing good way to share logic between the web and standard dart:ui so I don't want to try to do it on top of this PR.  Everything else is minor nits and LGTM.  Thanks again.

jtmcdole (approved):
LGTW (looks good to wikipedia specification)

PR #181743, opened 2026-01-31
Add buffer around rerasterized input to fragment shaders to maintain coordinate space when clipped

### LINKED ISSUE #181660: `BackdropFilter`  shader UV sampling is incorrect on beta when using `ClipRect` and a composed blur
This issue is a spiritual successor to https://github.com/flutter/flutter/issues/178798

### Steps to reproduce

1. Run the code sample on the current beta channel

### Expected results

The blur filter does not affect the rendering of the shader filter. Result should always be what it looks like when setting sigma to 0.

### Actual results

**With sigma = 0:**

Image

**Sigma = 1:**

Image

### Code sample

**assets/shaders/uv.frag**
```glsl

#version 460 core
precision mediump float;

#include <flutter/runtime_effect.glsl>

uniform vec2 uSize;
uniform sampler2D uTexture;

out vec4 fragColor;

void main() {
    vec2 fragCoord = FlutterFragCoord().xy;

    vec2 screenUV = vec2(fragCoord.x / uSize.x, fragCoord.y / uSize.y);        
    vec4 texColor = texture(uTexture, screenUV);
        
    #ifdef IMPELLER_TARGET_OPENGLES
        screenUV.y = 1.0 - screenUV.y;
    #endif
    
    // Check if we're within 20px of any edge
    float borderWidth = 20.0;
    bool inBorder = fragCoord.x < borderWidth || 
                    fragCoord.x > (uSize.x - borderWidth) ||
                    fragCoord.y < borderWidth || 
                    fragCoord.y > (uSize.y - borderWidth);
    
    if (inBorder) {
        fragColor = vec4(0.5, 0.0, 0.5, 1.0); // Purple border
    } else {
        fragColor = vec4(screenUV, 0.0, 1.0);
        fragColor = mix(fragColor, texColor, .5);
    }
}
```

**pubspec.yaml**
```yaml
name: flutter_sandbox
description: "A new Flutter project."
publish_to: 'none'
version: 0.1.0

environment:
  sdk: ^3.9.2

dependencies:
  flutter:
    sdk: flutter
  flutter_shaders: ^0.1.3

dev_dependencies:
  flutter_test:
    sdk: flutter
  flutter_lints: ^5.0.0

flutter:
  uses-material-design: true

  shaders:
    - assets/shaders/uv.frag
```

**lib/main.dart**
```dart
import 'dart:ui';

import 'package:flutter_shaders/flutter_shaders.dart';

void main() {
  runApp(const MainApp());
}

class MainApp extends StatelessWidget {
  const MainApp({super.key});

  @override
    return MaterialApp(
      home: Scaffold(
        body: Stack(
          children: [
            Positioned.fill(child: GridPaper()),
            Positioned.fill(
              child: ShaderBuilder(
                assetKey: 'assets/shaders/uv.frag',
                (context, shader, child) => MyShaderWidget(shader: shader),
              ),
            ),
          ],
        ),
      ),
    );
  }
}

class MyShaderWidget extends SingleChildRenderObjectWidget {
  const MyShaderWidget({super.key, required this.shader, super.child});

  final FragmentShader shader;

  @override
  RenderObject createRenderObject(BuildContext context) {
    return RenderMyShader(shader);
  }
}

class RenderMyShader extends RenderProxyBox {
  RenderMyShader(this.shader);

  final FragmentShader shader;

  final LayerHandle<ClipRectLayer> _clipRectLayer =
      LayerHandle<ClipRectLayer>();

  final LayerHandle<BackdropFilterLayer> _backdropFilterLayer =
      LayerHandle<BackdropFilterLayer>();

  static const inset = EdgeInsets.all(66);

  @override
    final rect = offset & size;
    _clipRectLayer.layer = context.pushClipRect(
      true,
      offset,
      inset.deflateRect(rect),
      (context, offset) {
        final layer = _backdropFilterLayer.layer ??= BackdropFilterLayer();

        layer.filter = ImageFilter.shader(shader);
        context.pushLayer(layer, (context, offset) {}, offset);
      },
      oldLayer: _clipRectLayer.layer,
    );

    super.paint(context, offset);
  }

  @override
  void dispose() {
    _clipRectLayer.layer = null;
    _backdropFilterLayer.layer = null;
    super.dispose();
  }
}
```
</details>

### Screenshots or Video

_No response_

### Logs

_No response_

### Flutter Doctor output

```console
[✓] Flutter (Channel beta, 3.41.0-0.1.pre, on macOS 15.7.3 24G419 darwin-arm64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 26.0)
[✓] Network resources

• No issues found!
```

</details>

gaaclarke:
fixes https://github.com/flutter/flutter/issues/181660

This works by changing the rerasterization logic such that we will rerasterize the output from the blur filter to an image that matches the dimensions of the input.  The undoes the optimization of the clip rect but standardizes the input to the fragment shader so that it is behaving as if the input was the same as a backdrop filter without the blur before hand.

I think there may be opportunities in the future to find a way to make this more efficient if we could find some way to rerasterize only to the clipped region but tweak the coordinates in the fragment shader to make it act like it has a whole image.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -98,6 +98,18 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
           [maybe_input_coverage,
            entity_offset](const Entity& entity) -> std::optional<Rect> {
             Rect coverage = maybe_input_coverage.value();
+            // The LT values come from the offset of the clip rect, that creates
+            // the clipping effect on the content that will be rendered from
+            // the fragment shader.  The RB values define the region we'll be
+            // synthesizing and ultimately defines the width and the height of
+            // the rasterized image.  The LT values can be thought of shifting
+            // the window that will be rasterized. Since we are shifting from
+            // the top-left corner, that is effectively pushing the the bottom
+            // right corner lower, outside of the rendering space.  So, we can
+            // clamp those values to the coverage's RB values.  This doesn't
+            // cause the fragment shader's rendering to deform because the
+            // magic width/height values sent to the fragment shader don't take
+            // the rasterized image's size into account.
             return Rect::MakeLTRB(entity_offset.x, entity_offset.y,
                                   coverage.GetRight(), coverage.GetBottom());
           });
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -14,9 +14,6 @@
 #include "impeller/geometry/point.h"
 #include "impeller/geometry/size.h"
 
-extern float fudge_x;
-extern float fudge_y;
-
 namespace impeller {
 
 void RuntimeEffectFilterContents::SetRuntimeStage(
@@ -97,9 +94,9 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
           },
           [maybe_input_coverage,
            entity_offset](const Entity& entity) -> std::optional<Rect> {
-            FML_LOG(ERROR) << "Doof: " << maybe_input_coverage.value();
-            // return maybe_input_coverage;
-            return Rect::MakeXYWH(entity_offset.x, entity_offset.y, 2048, 1536);
+            Rect coverage = maybe_input_coverage.value();
+            return Rect::MakeLTRB(entity_offset.x, entity_offset.y,
+                                  coverage.GetRight(), coverage.GetBottom());
           });
 
       Entity entity;
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -87,6 +87,9 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
       texture_contents.SetStencilEnabled(false);
       texture_contents.SetSamplerDescriptor(input_snapshot->sampler_descriptor);
 
+      // Use an AnonymousContents to restore the padding around the input that
+      // may have been cut out with a clip rect to maintain the correct
+      // coordinates for the fragment shader to perform.
       auto anonymous_contents = AnonymousContents::Make(
           [&texture_contents](const ContentContext& renderer,
                               const Entity& entity, RenderPass& pass) -> bool {
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -98,7 +98,8 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
           [maybe_input_coverage,
            entity_offset](const Entity& entity) -> std::optional<Rect> {
             Rect coverage = maybe_input_coverage.value();
-            return coverage.Shift(entity_offset - coverage.GetOrigin());
+            return Rect::MakeLTRB(entity_offset.x, entity_offset.y,
+                                  coverage.GetRight(), coverage.GetBottom());
           });
 
       Entity entity;
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -71,6 +71,8 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
   // this branch for the unit test `ComposePaintRuntimeOuter`, but do it for
   // `ComposeBackdropRuntimeOuterBlurInner`.
   if (input_snapshot->ShouldRasterizeForRuntimeEffects()) {
+    Vector2 entity_offset =
+        Vector2(entity.GetTransform().m[12], entity.GetTransform().m[13]);
     Matrix inverse = input_snapshot->transform.Invert();
     Quad quad = inverse.Transform(Quad{
         coverage.GetLeftTop(),     //
@@ -93,12 +95,11 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
                               const Entity& entity, RenderPass& pass) -> bool {
             return texture_contents.Render(renderer, entity, pass);
           },
-          [maybe_input_coverage](const Entity& entity) -> std::optional<Rect> {
+          [maybe_input_coverage,
+           entity_offset](const Entity& entity) -> std::optional<Rect> {
             FML_LOG(ERROR) << "Doof: " << maybe_input_coverage.value();
-            auto clip_rect_origin = Vector2(10, 10);
             // return maybe_input_coverage;
-            return Rect::MakeXYWH(-clip_rect_origin.x, -clip_rect_origin.y,
-                                  2048, 1536);
+            return Rect::MakeXYWH(entity_offset.x, entity_offset.y, 2048, 1536);
           });
 
       Entity entity;
--- a/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
+++ b/engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
@@ -84,10 +84,19 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
       texture_contents.SetStencilEnabled(false);
       texture_contents.SetSamplerDescriptor(input_snapshot->sampler_descriptor);
 
+      auto anonymous_contents = AnonymousContents::Make(
+          [&texture_contents](const ContentContext& renderer,
+                              const Entity& entity, RenderPass& pass) -> bool {
+            return texture_contents.Render(renderer, entity, pass);
+          },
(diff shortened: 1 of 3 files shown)


engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
```diff
@@ -84,10 +87,25 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
...
+      auto anonymous_contents = AnonymousContents::Make(
+          [&texture_contents](const ContentContext& renderer,
+                              const Entity& entity, RenderPass& pass) -> bool {
+            return texture_contents.Render(renderer, entity, pass);
+          },
+          [maybe_input_coverage,
+           entity_offset](const Entity& entity) -> std::optional<Rect> {
+            Rect coverage = maybe_input_coverage.value();
+            return Rect::MakeLTRB(entity_offset.x, entity_offset.y,
+                                  coverage.GetRight(), coverage.GetBottom());
```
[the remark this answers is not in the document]
gaaclarke: I mean the clipping is the point.

gaaclarke:
The goldens demonstrate some different behavior when the transformed textcoords are outside the bounds of the image.  That's fine since I don't think we've spent any effort making sure that that case is accounted for yet.

engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc
```diff
@@ -84,10 +87,25 @@ std::optional<Entity> RuntimeEffectFilterContents::RenderFilter(
...
 
+      // Use an AnonymousContents to restore the padding around the input that
+      // may have been cut out with a clip rect to maintain the correct
+      // coordinates for the fragment shader to perform.
+      auto anonymous_contents = AnonymousContents::Make(
+          [&texture_contents](const ContentContext& renderer,
+                              const Entity& entity, RenderPass& pass) -> bool {
+            return texture_contents.Render(renderer, entity, pass);
+          },
+          [maybe_input_coverage,
```
walley892, line 98: nit: 

maybe_input_coverage is an optional rect, right? Since `value()` will throw an exception anyway if there's nothing, is it more efficient to just dereference and not take the runtime hit of the .value check?

```
[coverage, entity_offset](const Entity& entity) -> std::optional<Rect> {
    return Rect::MakeLRTB(entity_offset.x, entity_offset.y, coverage->GetRight(), coverage->GetBottom());
}

**Review by walley892 — COMMENTED** [review]
I think flar will have more interesting feedback. Just a minor question

File: engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc (hunk 2 above)
Comment (gaaclarke) [reply engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:98]: > Since value() will throw an exception anyway if there's nothing

Pedantic note: We (google) compile c++ with exceptions disabled.

> is it more efficient to just dereference and not take the runtime hit of the .value check

Dereferencing an std::optional works with an operator overload, it has the same cost as .value().  Capturing the value instead of the optional type in the lambda would have a slight memory improvement because sizeof(std::optional<X>) > sizeof(X).  But it's just like 1 byte difference, maybe 4 bytes for padding.  I believe that should all be on the stack though so it's not a big deal.

File: engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc (hunk 1 above)
Comment (flar) [inline engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:114]: Shouldn't that be MakeXYWH(offset.xy, coverage.wh)?

**Review by flar — CHANGES_REQUESTED** [review]
I'm not sure what anonymous context does, but it looks like it lets you lie about the coverage geometry of the input and it looks like this will do what you want with the caveat that I'm assuming a lot about how the anonymous contents does its job.

I do agree with gemini here, I'm not sure xy from one source and right/bottom from another mix in any meaningful way. I think you want xy from one source and width/height from the other, no?

File: engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc (hunk 1 above)
Comment (flar) [reply engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:114]: Or coverage.Shift(offset.xy)?

Comment (flar) [reply engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:114]: If the input coverage is always 0,0 based then Right == Width and Bottom == Height, and the 2 are the same, but it "reads wrong". Shouldn't you be transferring the dimensions and the fact that the bottom right coordinates match the dimensions is "information not readily transparent from this code"? Even if it doesn't fix any issues, I think that using WH vs RB is the better way to express this line.

Or just using the Shift method...

Comment (flar) [reply engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:114]: No, actually, that's not the case. Even if the rect is 0,0 based, by not offsetting the RB coordinates you are indeed clipping it. Imagine a clip that is 100x100 but is positioned with a translation of 500,500. You'd end up with a rect that was Left=500, Right=100 which is backwards and therefore empty. Same with the height.

Comment (gaaclarke) [reply engine/src/flutter/impeller/entity/contents/filters/runtime_effect_filter_contents.cc:114]: > No, actually, that's not the case. Even if the rect is 0,0 based, by not offsetting the RB coordinates you are indeed clipping it. Imagine a clip that is 100x100 but is positioned with a translation of 500,500. You'd end up with a rect that was Left=500, Right=100 which is backwards and therefore empty. Same with the height.

I spent a while looking into the width and height of the coverage.  That's why I have so many parameters to the added unit test.  I too was worried that we could accidentally cause stretching of the circle if we got the width and the height wrong.  I switched to the Shift then reverted it.  I think this is the correct math and the reason why is that the coordinates are in the global coordinate space.  So there is no translation that will make the rect go inside out.  If you think there is still an issue with the math I think it would be better if we could express it as a test case to make sure we don't break it in the future.

**Comment by gaaclarke** [comment]
@flar I increased the limits on the clip rect so we can see all possible combinations for it to alleviate any concerns that there is a case we are missing.

**Comment by gaaclarke** [comment]
@flar I also tested this PR against this test that uses the composition on a layer filter that isn't backdrop and has a transform and while I'm not sure if the result is 100% what we want, it behaves similar to HEAD except that it doesn't flicker.  I think this addresses you concern about having a translate affecting the coverage?  Maybe that can be a jumping off point for you to express your concern.

```c++
TEST_P(AiksTest, ClippedImageFilterWithShader) {
  struct FragUniforms {
    Vector2 uSize;
  } frag_uniforms = {.uSize = Vector2(400, 400)};
  auto uniform_data = std::make_shared<std::vector<uint8_t>>();
  uniform_data->resize(sizeof(FragUniforms));
  memcpy(uniform_data->data(), &frag_uniforms, sizeof(FragUniforms));

  auto runtime_stages_result =
      OpenAssetAsRuntimeStage("runtime_stage_border.frag.iplr");
  ABSL_ASSERT_OK(runtime_stages_result);
  std::shared_ptr<RuntimeStage> runtime_stage =
      runtime_stages_result
          .value()[PlaygroundBackendToRuntimeStageBackend(GetBackend())];
  ASSERT_TRUE(runtime_stage);
  ASSERT_TRUE(runtime_stage->IsDirty());

  std::vector<std::shared_ptr<DlColorSource>> sampler_inputs = {
      nullptr,
  };

  auto runtime_filter = DlImageFilter::MakeRuntimeEffect(
      DlRuntimeEffectImpeller::Make(runtime_stage), sampler_inputs,
      uniform_data);

  Scalar sigma = 20.0;
  Scalar translate_x = 0.0;
  Scalar translate_y = 0.0;

  auto callback = [&]() -> sk_sp<DisplayList> {
    if (AiksTest::ImGuiBegin("Controls", nullptr,
                             ImGuiWindowFlags_AlwaysAutoResize)) {
      ImGui::SliderFloat("sigma", &sigma, 0, 50);
      ImGui::SliderFloat("translate_x", &translate_x, 0, 2048);
      ImGui::SliderFloat("translate_y", &translate_y, 0, 1536);
      ImGui::End();
    }

    auto blur_filter =
        DlImageFilter::MakeBlur(sigma, sigma, DlTileMode::kClamp);
    auto composed_filter =
        DlImageFilter::MakeCompose(runtime_filter, blur_filter);

    DisplayListBuilder builder;

    // Draw a background so we can see the effect against white
    DlPaint background_paint;
    background_paint.setColor(DlColor::kWhite());
    builder.DrawPaint(background_paint);

    builder.Save();

    builder.Translate(translate_x, translate_y);
    // Replicate the clip rect (inset by 66)
    builder.ClipRect(DlRect::MakeXYWH(66, 66, 268, 268));

    DlPaint save_paint;
    save_paint.setImageFilter(composed_filter);
    builder.SaveLayer(std::nullopt, &save_paint, nullptr);

    // Draw some pattern to be filtered
    DlPaint pattern_paint;
    pattern_paint.setColor(DlColor::kRed());
    builder.DrawRect(DlRect::MakeXYWH(0, 0, 200, 200), pattern_paint);
    pattern_paint.setColor(DlColor::kBlue());
    builder.DrawRect(DlRect::MakeXYWH(200, 200, 200, 200), pattern_paint);

    builder.Restore();  // Restore SaveLayer
    builder.Restore();  // Restore Save (Clip)

    return builder.Build();
  };

  ASSERT_TRUE(OpenPlaygroundHere(callback));
}
```

flar (approved):
LGTM

Issue #164557, opened 2025-03-04
[macOS] macOS desktop project should enable wide gamut when running with Impeller


jonahwilliams:
From a cursory look at the code, all that needs to change is FlutterSurface.mm:

pixelFormat = kCVPixelFormatType_40ARGBLEWideGamut;
bytesPerElement = 8;

and the texture format change to MTLPixelFormatBGRA10_XR

On macOS I think detecting support is trickier, but we can use it as long as the device is apple series 3 or higher.

FYI @knopp

loic-sharma:
Fixed by: https://github.com/flutter/flutter/pull/186277

PR #181769, opened 2026-02-01
Implement macOS wide gamut (Display P3) support

westito:
Adds wide gamut color support to macOS (matching iOS), upgrades the surface pixel format from 10-bit BGRA10_XR to 16-bit float RGBA16Float on both iOS and macOS when enabled, and fixes Impeller's blur filter P3 clamping on macOS.

**macOS Wide Gamut Support**
- Added DoesHardwareSupportWideGamut() hardware capability check (MTLGPUFamilyApple2 or MTLGPUFamilyMac2)
- Wide gamut enabled when both hardware supports it and FLTEnableWideGamut plist flag is YES
- Dynamic wide gamut switching when windows move between P3 and sRGB displays
- Added flutter/screenshot method channel on macOS for integration testing

**RGBA16Float Surface Format (iOS + macOS)**
- macOS IOSurface: kCVPixelFormatType_64RGBAHalf + MTLPixelFormatRGBA16Float
- iOS CAMetalLayer: MTLPixelFormatRGBA16Float
- Image decoder: always uses kRGBA_F16_SkColorType for all wide gamut images (previously only transparent images used 16-bit)

**Fix Blur P3 Clamping on macOS**
macOS uses the compositor/embedder path, not GPUSurfaceMetalImpeller, so UpdateOffscreenLayerPixelFormat was never called. Added the call in embedder.cc after wrapping the Metal resolve texture.

**Why RGBA16Float over BGRA10_XR?**
BGRA10_XR has only 10 bits per channel — values outside sRGB gamut get clamped in intermediate render targets (e.g. blur filters). RGBA16Float has 16 bits per channel with full floating-point range, preventing P3 color clamping in multi-pass rendering.

**Tests**
- 9 new iOS FlutterView unit tests verifying RGBA16Float pixel format and extended sRGB color space
- Updated macOS FlutterSurfaceManagerTest for RGBA16Float, dynamic switching, color space, and pixel format verification
- Updated image decoder and Impeller display list tests for kR16G16B16A16Float
- 11 macOS/iOS integration tests: image, saveLayer, codecImage, none, blur, drawnImage, container, linearGradient, radialGradient, conicalGradient, sweepGradient

### Issues
https://github.com/flutter/flutter/issues/164557

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke (changes requested):
Can you please split this pr between MacOS support and switching iOS's pixel format to 16bit floats?

macOS will be easier to approve and accept.  Switching iOS's pixel format is going to be much more involved since it will require doing some research about support of legacy devices, including performance.  Honestly I'm not sure we have a the resources to do that sort of evaluation soon.

westito:
Okay, I understand! I reverted iOS, so it uses 10bit. 
For Android: It is reasonable to use 8bit only. Thanks for telling me, I will do it that way.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -113,55 +119,23 @@ List<double> _deepRed = <double>[1.0931, -0.2268, -0.1501];
...
+  epsilon ??= _defaultEpsilon;
   expect(result, isNotNull);
   expect(result.length, 4);
   final [int width, int height, String format, Uint8List bytes] = result;
   return switch (format) {
-    'MTLPixelFormatBGR10_XR' => _findBGR10Color(bytes, width, height, color, epsilon: epsilon),
     'MTLPixelFormatBGRA10_XR' => _findBGRA10Color(bytes, width, height, color, epsilon: epsilon),
     'MTLPixelFormatRGBA16Float' => _findRGBAF16Color(bytes, width, height, color, epsilon: epsilon),
-    _ => fail('Unsupported pixel format: $format'),
+    _ => (false, <double>[0, 0, 0]),
```
gaaclarke: Let's turn this to an exception so that if we hit it the developer will get a meaningful error message.  We've replace a concrete error message with a "unsupported pixel format?" message.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -186,100 +160,119 @@ class _HasColor extends Matcher {
...
-    final closest = matchState['closest'] as List<double>;
+    final closest = matchState['closest'] as List<double>?;
+    if (closest == null) {
+      return mismatchDescription.add('could not find any colors (unsupported pixel format?)');
+    }
     return mismatchDescription.add('closest color to $color was $closest');
   }
 }
 
+const MethodChannel _screenshotChannel = MethodChannel('flutter/screenshot');
```
gaaclarke: I prefer to not have this as a global so we can limit it's scope and have better timing on it's creation.

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
```diff
@@ -186,100 +160,119 @@ class _HasColor extends Matcher {
...
+  await tester.runAsync(() => precacheImage(MemoryImage(base64Decode(app.displayP3Logo)), context));
+}
+
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
   group('end-to-end test', () {
     testWidgets('look for display p3 deepest red', (WidgetTester tester) async {
       app.run(app.Setup.image);
-      await tester.pumpAndSettle(const Duration(seconds: 2));
```
gaaclarke, line 200: Why are we removing the time on pump and settle?

dev/integration_tests/wide_gamut_test/lib/main.dart
```diff
@@ -291,18 +297,22 @@ class _MyHomePageState extends State<MyHomePage> {
...
+        imageWidget = Image.memory(base64Decode(displayP3Logo));
+      case Setup.codecImage:
+        imageWidget = _image != null
+            ? RawImage(image: _image, width: 100, height: 100)
+            : const SizedBox(width: 100, height: 100);
       case Setup.drawnImage:
         imageWidget = CustomPaint(painter: _SaveLayerDrawer(_image));
       case Setup.canvasSaveLayer:
-        imageWidget = CustomPaint(painter: _SaveLayerDrawer(_image));
+        imageWidget = Image.memory(base64Decode(displayP3Logo));
```
gaaclarke: As part of you PR can you avoid editing the meaning of existing tests, instead just add the new one that you want.

engine/src/flutter/impeller/display_list/aiks_dl_atlas_unittests.cc
```diff
@@ -165,7 +165,7 @@ TEST_P(AiksTest, DrawAtlasWithColorAdvancedAndTransform) {
 TEST_P(AiksTest, DrawAtlasPlusWideGamut) {
   DisplayListBuilder builder;
   EXPECT_EQ(GetContext()->GetCapabilities()->GetDefaultColorFormat(),
-            PixelFormat::kB10G10R10A10XR);
+            PixelFormat::kR16G16B16A16Float);
```
gaaclarke: We don't want this anymore.  We'll have to come up with another way so that we are testing both pixel formats instead of just testing one of them.

We either need to make a way to select the pixel format in the test or add another magic suffix to the test name, like DrawAtlasPlusWideGamutF16.

engine/src/flutter/impeller/display_list/aiks_dl_basic_unittests.cc
```diff
@@ -1730,7 +1730,7 @@ TEST_P(AiksTest,
 // This makes sure the WideGamut named tests use 16bit float pixel format.
 TEST_P(AiksTest, FormatWideGamut) {
   EXPECT_EQ(GetContext()->GetCapabilities()->GetDefaultColorFormat(),
-            PixelFormat::kB10G10R10A10XR);
+            PixelFormat::kR16G16B16A16Float);
```
gaaclarke: same

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -65,14 +66,18 @@ static void UpdateContentSubLayers(CALayer* layer,
...
+                                   BOOL enableWideGamut) {
   // Adjust sublayer count to paintRegion count.
   while (layer.sublayers.count > paintRegion.size()) {
     [layer.sublayers.lastObject removeFromSuperlayer];
   }
 
   while (layer.sublayers.count < paintRegion.size()) {
     CALayer* newLayer = [CALayer layer];
+    if (enableWideGamut) {
+      newLayer.contentsFormat = kCAContentsFormatRGBA16Float;
```
gaaclarke: Let's pull out this constant as a constant of our own since we use it in a few places.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -343,6 +385,12 @@ - (void)returnSurfaces:(nonnull NSArray<FlutterSurface*>*)returnedSurfaces {
   [self performSelectorOnMainThread:@selector(reschedule) withObject:nil waitUntilDone:NO];
 }
 
+- (void)flush {
+  @synchronized(self) {
+    [_surfaces removeAllObjects];
+  }
+}
```
gaaclarke, line 374: This does the same thing as onIdle, maybe onIdle should use it?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h
```diff
@@ -71,6 +71,14 @@
  */
 - (void)didUpdateMouseCursor:(nonnull NSCursor*)cursor;
 
+/**
+ * Updates the wide gamut setting on the surface manager. Called when
+ * the window moves to a screen with different gamut support.
+ *
+ * Must be called on the platform thread.
+ */
+- (void)setEnableWideGamut:(BOOL)enableWideGamut;
```
gaaclarke: I don't think we want this as part of the public interface.  Our engine is the one that should be calling it.  Let's move this to a category on the class instead.  See FlutterEngine_Internal.h

TESTOWNERS
```diff
@@ -115,6 +115,8 @@
 /dev/devicelab/bin/tasks/very_long_picture_scrolling_perf__e2e_summary.dart @flar @flutter/engine
 /dev/devicelab/bin/tasks/web_size__compile_test.dart @yjbanov @flutter/web
 /dev/devicelab/bin/tasks/wide_gamut_ios.dart @gaaclarke @flutter/engine
+/dev/devicelab/bin/tasks/wide_gamut_macos.dart @gaaclarke @flutter/engine
```
gaaclarke, line 118: 🫡

engine/src/flutter/impeller/playground/backend/metal/playground_impl_mtl.mm
```diff
@@ -80,7 +80,7 @@
       switches.flags, ShaderLibraryMappingsForPlayground(),
       is_gpu_disabled_sync_switch_, "Playground Library",
       switches.enable_wide_gamut
-          ? std::optional<PixelFormat>(PixelFormat::kB10G10R10A10XR)
+          ? std::optional<PixelFormat>(PixelFormat::kR16G16B16A16Float)
```
gaaclarke: Mentioned above, we don't want to do this since kB10G10R10A10XR gets much more traffic.  Ideally we test both.

gaaclarke (changes requested):
Very impressive work!  The big feedback is we need to keep our tests in kB10G10R10A10XR, ideally we add f16 tests instead of replacing them.  The integration test changes will be easier to review if you refrain from refactoring them and instead just add the new test that you want.  Refactoring can happen in a follow up PR.  The changes to macOS look good to me and seem well tested.  We'll have to get a review from someone more familiar with the macOS embedder though (pr always need 2 reviews anyways).

gaaclarke:
Head's up there are some legitimate failures in the golden tests above: https://flutter-gold.skia.org/cl/github/181769

dev/devicelab/lib/tasks/integration_tests.dart
```diff
@@ -239,11 +239,11 @@ TaskFunction createWindowingDriverTest() {
   };
 }
 
-TaskFunction createWideGamutTest() {
+TaskFunction createWideGamutTest({required String platform}) {
   return IntegrationTest(
     '${flutterDirectory.path}/dev/integration_tests/wide_gamut_test',
     'integration_test/app_test.dart',
-    createPlatforms: <String>['ios'],
+    createPlatforms: <String>[platform],
```
loic-sharma: If you make this `['ios', 'macos']` can you get rid of the `platform` argument? If that works, I'd consider doing that. Otherwise, I'd consider adding something like `assert(platform == 'ios' || platform == 'macos');` to validate the `platform` argument.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterDartProject.mm
```diff
@@ -7,9 +7,24 @@
...
+
+static BOOL DoesHardwareSupportWideGamut() {
+  static BOOL result = NO;
+  static dispatch_once_t once_token = 0;
+  dispatch_once(&once_token, ^{
+    id<MTLDevice> device = MTLCreateSystemDefaultDevice();
+    // Apple2 covers Apple Silicon Macs (M1+).
+    // Mac2 covers Intel Macs with AMD GPUs that support wide gamut pixel formats.
+    // Mac1 does not support wide gamut pixel formats.
+    result = [device supportsFamily:MTLGPUFamilyApple2] || [device supportsFamily:MTLGPUFamilyMac2];
```
loic-sharma: Instead of a blocklist of devices that don't support wide gamut, would it be better to have an allowlist of devices that do support wide gamut? Imagine Apple adds `MTLGPUFamilyApple3` in the future. If we forget to update this branch, those new devices won't get wide gamut. cc @gaaclarke  for thoughts.

gaaclarke: Yes, that's a good suggestion, except I think you have the terms switched.  Right now this is an allowlist, we want a blocklist since the list of blocked devices shouldn't be growing in the future =)

dev/integration_tests/wide_gamut_test/integration_test/app_test.dart
westito: Because the test using image preloading now. Before, there was no preload and UI had to async wait for the test image to appear. Now load is awaited and the test is much faster this way.

dev/integration_tests/wide_gamut_test/lib/main.dart
westito: You right, debug leftover code.  I miss that.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h
westito: On iOS this is a static option - checked at app start - but on macOS wide gamut state is dynamically changing based on monitor capabilities, thats why this is in FlutterView header. There was no internal header so I created one.

dev/devicelab/lib/tasks/integration_tests.dart
westito: It works. But I made this way because the integration test runs on different devices. So when CI runs macOS test, it is pointless to generate iOS project and vice-versa. But it is harmless to generate both.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterDartProject.mm
westito: These are feature flags, not actual GPU types. If a new Family3 GPU arrives the `supportsFamily:MTLGPUFamilyApple2` will be true also. I can't change anyways, because `supportsFamily:MTLGPUFamilyApple1` also returns true :D

engine/src/flutter/impeller/display_list/aiks_dl_basic_unittests.cc
westito: I added both 10bit and 16bit tests

engine/src/flutter/impeller/display_list/aiks_dl_atlas_unittests.cc
westito: I added both 10bit and 16bit tests

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
westito: You're right! fixed

gaaclarke (changes requested):
Those failing goldens reminded me of the problem with using float16 for the frame buffer.  We shipped float16 at some point and we found we were getting wildly different results than when wide gamut was turned off.  It turned out it was because in all of the shaders we relying on the fact that the 8bit pixel formats have an implicit `clamp(x, 0, 1)` around all operations and they have all been written with that assumption.  If we were to support float16, we'd have to rewrite all the shaders to have explicit `clamp(x, 0, 1)` around all addition and multiplication operations.  That also has performance ramifications for old devices that only support 8bit pixel formats.  I believe this is why those goldens were failing.  I think the easiest thing is to use bgra10xr, even for macos.  Hopefully it's supported on the desktop gpus too, I haven't looked into it.

edit: it gets even trickier in extended srgb since you aren't clamping on 0,1

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -104,12 +112,18 @@ @implementation FlutterSurfaceManager
...
+               enableWideGamut:(BOOL)enableWideGamut {
   if (self = [super init]) {
     _device = device;
     _commandQueue = commandQueue;
     _containingLayer = containingLayer;
     _delegate = delegate;
+    _enableWideGamut = enableWideGamut;
+
+    if (enableWideGamut) {
+      _containingLayer.contentsFormat = kFlutterWideGamutContentsFormat;
```
knopp: Containing layer doesn't have any content, is the content format necessary? I would assume it is not.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -118,6 +132,32 @@ - (instancetype)initWithDevice:(id<MTLDevice>)device
...
+- (void)setEnableWideGamut:(BOOL)enableWideGamut {
+  FML_DCHECK([NSThread isMainThread]);
+  if (_enableWideGamut == enableWideGamut) {
+    return;
+  }
+  _enableWideGamut = enableWideGamut;
+
+  NSString* contentsFormat =
+      enableWideGamut ? kFlutterWideGamutContentsFormat : kCAContentsFormatRGBA8Uint;
+  _containingLayer.contentsFormat = contentsFormat;
```
knopp: Same here.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.h
```diff
@@ -52,7 +55,17 @@
 - (nullable instancetype)initWithDevice:(nonnull id<MTLDevice>)device
                            commandQueue:(nonnull id<MTLCommandQueue>)commandQueue
                                   layer:(nonnull CALayer*)containingLayer
-                               delegate:(nonnull id<FlutterSurfaceManagerDelegate>)delegate;
+                               delegate:(nonnull id<FlutterSurfaceManagerDelegate>)delegate
+                        enableWideGamut:(BOOL)enableWideGamut;
```
knopp: This might be a nit but I'd keep the argument named "wideGamut" instead, as `enableWideGamut` seems like an action taken (i.e. when moving window to wide gamut display) rather then simply specifying a surface creation flag.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -118,6 +132,32 @@ - (instancetype)initWithDevice:(id<MTLDevice>)device
...
+    for (CALayer* sublayer in layer.sublayers) {
+      sublayer.contentsFormat = contentsFormat;
+    }
+  }
+
+  // Flush cached surfaces since they have the wrong pixel format.
+  [_backBufferCache flush];
+
+  // Clear front surfaces — they will be replaced on the next present.
+  [_frontSurfaces removeAllObjects];
```
knopp, line 134: What is triggering the next present?

knopp (commented):
Overall looks good. Left some questions.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
knopp: I *think* setting the `contentsFormat` might be unnecessary?

engine/src/flutter/impeller/playground/backend/metal/playground_impl_mtl.mm
westito: Fixed

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
westito: You sure. I removed and works. At the beginning when tests failed many times I tried set formats multiple places. It seems not need.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
westito: Removed

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
westito: The `setEnableWideGamut` method and remove inside it  triggered by `windowDidChangeScreen` and next frame naturally drawn because window change also schedules a new frame in `updateWindowMetricsForViewController`. Without remove, it is possible sRGB screen gets buffered frame with wide gamut pixel format in it. I think it isn't cause problem - GPU surely handles it - but it is a more clearer approach I think.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -65,7 +66,8 @@ static void UpdateContentSubLayers(CALayer* layer,
                                    CGFloat scale,
                                    CGSize surfaceSize,
                                    NSColor* borderColor,
-                                   const std::vector<FlutterRect>& paintRegion) {
+                                   const std::vector<FlutterRect>& paintRegion,
+                                   BOOL enableWideGamut) {
```
gaaclarke: This parameter was added to this method, but it isn't used.

gaaclarke (commented):
This LGTM.  Thanks again for another great PR.  I'll give it an approval as soon as we can get it passing all the tests on CI.  Looks like `dev/integration_tests/wide_gamut_test/integration_test/app_test.dart` isn't formatted correctly right now.  Even though you might be thinking we didn't change anything for Linux, Linux is where we run the formatting tests.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
westito: Removed

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
knopp: UpdateWindowMetricsForViewController answers my question. I agree recreating the surfaces is the right thing to do.

knopp:
There's one thing that is bugging me - I think we should check the surfaces inside `[FlutterSurfaceManager commit:]` and discard the commit if surfaces don't match current wide gamut mode.

Is possible that these surfaces have been held by raster thread during wide gamut switch and we need them out of circulation.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -152,8 +168,16 @@ - (BOOL)enableSurfaceDebugInfo {
 - (void)commit:(NSArray<FlutterSurfacePresentInfo*>*)surfaces {
   FML_DCHECK([NSThread isMainThread]);
 
-  // Release all unused back buffer surfaces and replace them with front surfaces.
-  [_backBufferCache returnSurfaces:_frontSurfaces];
+  // Only return surfaces to the cache if they match the current wide gamut mode.
```
knopp: This is not what I had in mind. You don't need to touch `_frontSurfaces`, those are already cleaned `setEnableWideGamut`. It's the incoming surfaces (method argument) that I'm concerned with. You should check if they match current wide gamut, and if not, simply return early from the method, don't do anything with the surfaces. They will be released.

westito: I fixed

gaaclarke (approved):
LGTM!  Thanks.  I'll file a followup issue for making wide gamut the default.  We'll have to audit our testing to make sure we are comfortable with that change.  Having it be optional is good for now.

knopp (approved):
LGTM! Thanks!

PR #181795, opened 2026-02-02
[framework] Fix Text.semanticsIdentifier being absorbed by ancestor nodes

### LINKED ISSUE #180894: semanticsIdentifier of Text widget does not applied Semantic to it when used in Dialog widget with showDialog
### Steps to reproduce

Use below dartpad link where I have added this example to reproduce the issue. 
[Dart pad link](https://dartpad.dev/c4dfd4225ec70dc77fe55183eec947ec)

Run the application and open inspect console and paste 
`document.querySelector('[aria-label="Enable accessibility"]').click();` to enable element selection. 

Encounter if we use below only, Text element won't be identified.
```dart
Text( 'I should have the Semantics', semanticsIdentifier:'From Text Parameter',) 
```

However, if we use below, Text element will be identified.
```dart
Semantics(identifier: 'From Semantics',container: true, child: Text( 'I should have the Semantics', semanticsIdentifier:'From Text Parameter',) )
```

Whaat would be the reason for first part of code where Semantics are not identified using the `Text` paramenter?

### Expected results

It should identify the Text element same as what it can do with `Semantics` widget having container as true.

### Actual results

It's not identifying `Text` element with `semanticsIdentifier` parameter only.   

### Code sample

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        useMaterial3: true,
      ),
      home: const DialogTextSemantics(),
    );
  }
}

class DialogTextSemantics extends StatefulWidget {
  const DialogTextSemantics({super.key});

  @override
  State<DialogTextSemantics> createState() => _DialogTextSemanticsState();
}

class _DialogTextSemanticsState extends State<DialogTextSemantics> {
  @override
    return Scaffold(
      body: Center(
        child: ElevatedButton(
          onPressed: () {
            showDialog(
                context: context,
                builder: (bContext) {
                  return Material(
                    color: Colors.transparent,
                    child: Dialog(
                      shape: RoundedRectangleBorder(
                          borderRadius: BorderRadius.circular(4)),
                      child: SizedBox(
                        width: 800,
                        child: Padding(
                          child: Column(
                            children: [
                              Row(
                                children: [
                                  IconButton(
                                      onPressed: () {
                                        Navigator.of(bContext).pop();
                                      },
                                      icon:
                                          const Icon(Icons.arrow_back_rounded)),
                                  Expanded(
                                    // this below Semantics does help to identify the element using element selection
                                    child: Semantics(
                                      identifier: 'From Semantics',
                                      container: true,
                                      child: const Text(
                                        'I should have the Semantics',
                                        // this does not let the element identified using element selection
                                        // semanticsIdentifier:
                                        //     'From Text Parameter',
                                      ),
                                    ),
                                  ),
                                  IconButton(
                                      onPressed: () {
                                        Navigator.of(bContext).pop();
                                      },
                                      icon: const Icon(Icons.close_rounded)),
                                ],
                              ),
                            ],
                          ),
                        ),
                      ),
                    ),
                  );
                });
          },
          child: const Text('Click Here'),
        ),
      ),
    );
  }
}

```

</details>

### Screenshots or Video

<details open>

Image
Image
Image

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
purvikrana@Purviks-MacBook-Pro dialog_text_semantics % flutter doctor
[✓] Flutter (Channel stable, 3.35.7, on macOS 26.1 25B78 darwin-arm64, locale en-IN)
[!] Xcode - develop for iOS and macOS (Xcode 15.4)
    ! CocoaPods 1.14.3 out of date (1.16.2 is recommended).
[✓] Android Studio (version 2024.1)
[✓] VS Code (version 1.96.4)
    ! Error: Browsing on the local area network for Purvik Rana. Ensure the device is unlocked and attached with a cable or associated with the same local area
      network as this Mac.
    ! Error: Browsing on the local area network for Purvik’s Apple Watch. Ensure the device is unlocked and discoverable via Bluetooth. (code -27)
[✓] Network resources

purvikrana@Purviks-MacBook-Pro dialog_text_semantics % 
```

</details>

flutter-zl:
Problem: Text widget's semanticsIdentifier was being absorbed into ancestor semantic nodes (like Dialog) instead of creating its own semantic element.

Root Cause: The Semantics wrapper in Text didn't set container: true, allowing the identifier to be merged upward during semantics tree construction.

Fix: Added container: semanticsIdentifier != null to the Text widget's Semantics wrapper.

Fixes #180894.

Before
https://flutter-demo-11-before.web.app/

After
https://flutter-demo-11-after.web.app/


packages/flutter/lib/src/widgets/text.dart
```diff
@@ -804,6 +804,9 @@ class Text extends StatelessWidget {
         textDirection: textDirection,
         label: semanticsLabel,
         identifier: semanticsIdentifier,
+        // Fix for issue #180894: When semanticsIdentifier is provided, set container: true
+        // to prevent the identifier from being absorbed into ancestor semantic nodes.
+        container: semanticsIdentifier != null,
```
chunhtai: semanticsIdentifier IIRC is mostly for unittest and uiautomation test, letting it change tree shape seems weird. This seems a bit drastic to fix the root cause, i.e. dialog role ignored semantics identifier.

mdebbar: @chunhtai we could look at `semanticsIdentifier` as being a signal from the user wanting to have a reference to this semantics node, so we should not merge it up.

That being said, if we decide to go this route, the fix should be more general, not only for the Text widget.

flutter-zl: Make fix more general at SemanticsConfiguration level. Please review.

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -6693,6 +6693,9 @@ class SemanticsConfiguration {
     if (_maxValue != null && other._maxValue != null) {
       return false;
     }
+    if (other._identifier.isNotEmpty) {
```
chunhtai: We don't promise how this is called, for example other may be the parent, child, or sibling. but this code will depends on how the api is called to do the right thing.

I assume the intention is that we are ok if a child merge to a "parent that has identifier", but disallows a "child that has identifer" to merge to parent? the code here will fail if we somehow do child.isCompatableWith(parent).

We may need find a different way around it.

flutter-zl: Good catch. I've updated the logic. Now when identifier is set, isSemanticBoundary is set to true, which forces a dedicated SemanticsNode before merging is even considered. This avoids the isCompatibleWith directionality issue entirely.

packages/flutter/lib/src/rendering/object.dart
```diff
@@ -4786,7 +4786,7 @@ mixin SemanticsAnnotationsMixin on RenderObject {
   @override
   void describeSemanticsConfiguration(SemanticsConfiguration config) {
     super.describeSemanticsConfiguration(config);
-    config.isSemanticBoundary = container;
+    config.isSemanticBoundary = container || (_properties.identifier != null);
```
chunhtai, line 4789: can you also update the doc for _SemanticsBase.container and SemanticsProperty.identifier to mention this behavior?

chunhtai: also this may be a potential breaking change even if it is not breaking customer test, so should watchout after merge.

chunhtai (commented):

packages/flutter/lib/src/rendering/object.dart
flutter-zl: 1. I updated the docs.
2. Good point. I'll keep an eye on it after landing.

chunhtai (approved):
LGTM

chunhtai (approved):
LGTM

PR #181826, opened 2026-02-02
[ Tool ] Fix regression introduced in flutter/flutter#181421

bkonyi:
`dev/devicelab/bin/task/flutter_test_performance.dart` regressed due to the CFE not being initialized with the correct dill file for the target test. This change updates `TestCompiler` to pass the correct dill file when creating the `ResidentCompiler` and adds a test to verify that the `ResidentCompiler` is initialized with the correct dill.

See: flutter#181421.


gaaclarke (approved):
lgtm! thanks

PR #181828, opened 2026-02-02
Migrate abi build paths to use new abi filtering api #AGP9

reidbaker:
Related to #180137 

Not equivalent migrations, Before [1]* release/debug `BuildType`'s were modified to clear the set abi values and force set the 3 abi values flutter supports. The proposed code allows overriding of the abi config in defaultConfig without a flag but *requires* `-P disable-abi-filtering=true` to set custom values in either BuildTypes or Flavors. 
Flavors requiring a flag is consistent behavior with no change. This is because the "newDsl" only allows a union of default and other build types AGP bug tracking this can be found here http://issuetracker.google.com/issues/285353854. 

[1] if --split-abi flag was not passed and the app did not disable abi filter in their build props

`onVariants` is not preferred because [variant](https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-api/src/main/java/com/android/build/api/variant/Variant.kt) does not contain the values we wish to set. Specifically any abi/ndk values. The "buildType" on variant is a string which is not the same as `BaseExtension.buildTypes`. 
`variantBuilder` similarly did not have the values we want to modify. [VariantBuilder Source](https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-api/src/main/java/com/android/build/api/variant/VariantBuilder.kt).

Command to run a specific jni integration tests from `packages/flutter_tools/` `FLUTTER_ROOT=~/flutter-work/ ../../bin/cache/dart-sdk/bin/dart test test/integration.shard/gradle_jni_packaging_test.dart --plain-name 'abiFilters provided by the user take precedence over the default'` 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -550,6 +506,68 @@ class FlutterPlugin : Plugin<Project> {
...
+            // Developers can set this value by passing `-P disable-abi-filtering=true`
+            // to flutter build. Where "disable-abi-filtering" comes from
+            // com.flutter.gradle.FlutterPluginUtils.PROP_DISABLE_ABI_FILTERING.
+            if (!FlutterPluginUtils.shouldProjectDisableAbiFiltering(projectToAddTasksTo)) {
+                extension.defaultConfig.ndk {
+                    abiFilters.clear()
+                    abiFilters.addAll(PLATFORM_ABI_LIST)
+                }
+            }
+        }
```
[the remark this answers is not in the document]
reidbaker: Please review as is. I will try this in a different branch to see if I think it is better.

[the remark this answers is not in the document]
reidbaker: After evaluate I think violates the intention behind gradles optimization features. Possibly makes sense for project but not for frameworks. 
It breaks Configuration Cache which slows down builds and is incompatible with modern Gradle features.
and it relies on the order of execution, leading to hard-to-debug issues.

reidbaker (commented):
adding comments

reidbaker (commented):
.

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
jesswrd: To confirm the default behavior is really working (flutter abi filtering), can you verify libflutter.so is packaged in the APK for all flutter supported architectures?


```suggestion
    // Verify that libflutter.so is packaged in the APK for Flutter supported architectures.
    expect(_checkLibIsInApk(projectDir, 'lib/arm64-v8a/libflutter.so'), false);
    expect(_checkLibIsInApk(projectDir, 'lib/x86_64/libflutter.so'), true);
    expect(_checkLibIsInApk(projectDir, 'lib/armeabi-v7a/libflutter.so'), true);
```

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -550,6 +506,68 @@ class FlutterPlugin : Plugin<Project> {
...
+                //    in pom files.
+                // 2. For project dependencies (implementation(project(":flutter"))): The Flutter
+                //    Gradle Plugin is not applied to the main app subproject, so this apply()
+                //    method is never called.
+                //
+                // abiFilters cannot be added to templates because it would break builds when
+                // --splits-per-abi is used due to conflicting configuration. This approach
+                // adds them programmatically only when splits are not configured.
+                //
+                // If the user has specified abiFilters in their build.gradle file's . DefaultConfig,
```
gmackall:
```suggestion
                // If the user has specified abiFilters in their build.gradle file's DefaultConfig,
```

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
```diff
@@ -45,7 +47,7 @@ void main() {
     expect(_checkLibIsInApk(projectDir, 'lib/x86/libflutter.so'), false);
   });
 
-  testWithoutContext('abiFilters provided by the user take precedence over the default', () async {
+  testWithoutContext('abiFilters provided by the user in buildTypes can be overridden', () async {
```
jesswrd: nit: the test name confuses me. It reads as though the user-provided abilFilters is the thing being overriden when the thing being overriden is actually the default Flutter abi filters.

```suggestion
  testWithoutContext('abiFilters provided by the user in buildTypes can override Flutter defaults', () async {
```

jesswrd (approved):
Thanks for including the testing command! Just had a few suggestions but lgtm to me.

gmackall:
> The proposed code allows overriding of the abi config in defaultConfig without a flag but requires -P disable-abi-filtering=true to set custom values in either BuildTypes or Flavors.

Is this true? It looks to me like previously, because we set the abiFilters on all types, that flag would be required by a user to set their own abiFilters on a build type. But now that we are only mucking with the defaultConfig, wouldn't they _not_ need that flag now, when setting abiFilters on their own build type?

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
```diff
@@ -550,6 +506,68 @@ class FlutterPlugin : Plugin<Project> {
...
+        /**
+         * Clears existing abi configuration and sets ABI's supported by flutter.
+         */
+        private fun configureAbiWithoutSplits(
+            projectToAddTasksTo: Project,
+            extension: ApplicationExtension
+        ) {
+            // Developers can set this value by passing `-P disable-abi-filtering=true`
+            // to flutter build. Where "disable-abi-filtering" comes from
+            // com.flutter.gradle.FlutterPluginUtils.PROP_DISABLE_ABI_FILTERING.
```
gmackall: nit: I think we should delete this comment. I don't think there is value in explaining the internals of the helper function inline with its use. I think we are overly verbose in comments in this area of the code in general.

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
```diff
@@ -70,14 +72,60 @@ void main() {
         .childFile('android/app/build.gradle.kts')
         .writeAsStringSync(updatedBuildGradleContents);
 
-    processManager.runSync(<String>[flutterBin, 'build', 'apk'], workingDirectory: projectDir.path);
+    processManager.runSync(<String>[
+      flutterBin,
+      'build',
+      'apk',
+      '-P',
+      'disable-abi-filtering=true',
```
gmackall: Does this test fail without this line? Given that we are only modifying the default config now that would surprise me

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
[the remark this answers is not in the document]
reidbaker: After evaluate I think violates the intention behind gradles optimization features. Possibly makes sense for project but not for frameworks.
It breaks Configuration Cache which slows down builds and is incompatible with modern Gradle features.
and it relies on the order of execution, leading to hard-to-debug issues.

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
reidbaker: Fair feedback but I disagree. I, when working in this code and with an above average understanding of agp and gradle, and with a way above average understanding of this package structure had to spend time to track down how to set this value. I think if you are reading this code then the comment can be really helpful. I think that when searching for "ABI" in the flutter/flutter repo because you do not understand why abi filtering does not work like you expect you are likely to find this code and less likely to find the flag. I also think that the cost of comments is low. 

Given these are opinions I checked the style guide to see if that would give us some guidance at what to include or exclude and it does not cover this topic. 
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo/3ec25f128e35aaf0d4862e0f80a1ab38d573e5e0#comments

@jesswrd what do you think?

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
reidbaker: Yes it does. I found this test file because on an early iteration this test failed. I also ran this locally as part of my debugging. 

Finally I confirmed by removing the `-P disable-abi-filtering=true` and running the test locally. Without the flag this test fails, With the flag it passes. 

`FLUTTER_ROOT=~/flutter-work/ ../../bin/cache/dart-sdk/bin/dart test test/integration.shard/gradle_jni_packaging_test.dart` from `packages/flutter_tools`

reidbaker: One odd thing, the code passes with disable-abi-filtering=true or disable-abi-filtering=false

reidbaker: Fixed in https://github.com/flutter/flutter/pull/181828/changes/b42e71e87ba28ee43e9e7a7ef35da4914d6cd0c3

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
gmackall: I think it could be helpful, as long as the internals of the method don't change. But if they do it is now actively harmful, and it is very unlikely that the person changing the internals of the method will find a comment about its use. I think the internals of the method are the methods business, not comments elsewhere.

reidbaker: ok then how about I move this comment to the method. Then it is still one jump but it is adjacent to the implementation.

gmackall: Sure, I think a comment on the helper method explaining that it makes use of the flutter tool's piping of gradle properties via the `-android-project-arg/-P` option makes sense, my opposition was only to making a comment at a call site explain the internals of a helper method defined elsewhere

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
```diff
@@ -40,12 +42,15 @@ void main() {
           .childFile('libflutter.so'),
       exists,
     );
-
+    // Verify that libflutter.so is packaged in the APK for Flutter supported architectures.
+    expect(_checkLibIsInApk(projectDir, 'lib/arm64-v8a/libflutter.so'), false);
```
gmackall: should this be true? I noticed the tests are failing on this line, and this is a supported architecture right

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
jesswrd: That sounds good to me. While I do agree this area of FGP is verbose with comments, I appreciate the comments existing. It helps me make sense of what's happening in this area. I don't have strong opinons of where to move the comments—I just care that they exist somewhere.

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
gmackall: Ok it turns out that the default config gets unioned with, not overridden by, build type specific configuration (see http://issuetracker.google.com/issues/285353854). This explains why the test fails without that line, though it is very surprising behavior by AGP. I left a comment on the issue expressing that.

gmackall (approved):

packages/flutter_tools/test/integration.shard/gradle_jni_packaging_test.dart
reidbaker: typo I believe, updating locally.

packages/flutter_tools/gradle/src/main/kotlin/FlutterPlugin.kt
reidbaker: > Sure, I think a comment on the helper method explaining that it makes use of the flutter tool's piping of gradle properties via the `-android-project-arg/-P` option makes sense, my opposition was only to making a comment at a call site explain the internals of a helper method defined elsewhere

That is good feedback updated.

PR #181849, opened 2026-02-03
Mark all `IconData` fields as entrypoints to prevent tree shaking its fields

mkustermann:
Flutter's tree shaking of icon data in fonts examines the tree shaken Kernel AST, searching for `IconData` constants. It will then examine their fields in order to shake parts of fonts.

We should mark those fields as entrypoints to prevent the tree shaker from removing the fields (an optimization that may land in [0]) as the `const_finder` tool relies on them being there.

[0] https://dart-review.googlesource.com/c/sdk/+/472541


mraleph (approved):

packages/flutter/lib/src/widgets/icon_data.dart
```diff
@@ -154,3 +159,10 @@ class _StaticIconProvider {
 /// }
 /// ```
 const Object staticIconProvider = _StaticIconProvider();
+
+/// Retain the fields of [IconData] to ensure the icon tree shaker can access
+/// those fields after tree shaking the application.
+///
+/// NOTE: We may replace this with a `@RecordUse()` annotation once icon tree
+/// shaking uses the new link hooks.
```
justinmc: Nit:

"
This may be replaced with a `@RecordUse()` annotation once icon tree shaking uses the new line hooks.
"

https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-empty-prose

https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#use-the-passive-voice-recommend-do-not-require-never-say-things-are-simple

justinmc (approved):
LGTM 👍 . Sorry for the delay!

PR #181861, opened 2026-02-03
Windowing implementation of `showDialog` that uses a native desktop window to display the content 

mattkae:
image

image

## What's new?
- Added the `MaterialWindowingManager` which will render all of the sub-windows that the material app relies
- The `MaterialWindowingManager` provides the `MaterialWindowRegistry` which enables its descendants to create and remove windows from the windowing manager
- Implemented a windowing path for `showDialog` which uses that `Navigator` to push a true window to the navigation stack
- In `material_window_manager.dart`, the dialog will be rendered appropriately with inherited theming, text direction, and other media query data
- Added the `api_multiwindow` example application to demonstrate the new behavior for both fullscreen and non-fullscreen dialogs
- Added tests to the `api_multiwindow` example application
- Updated the documentation of `showDialog` to explain how each field works when windowing is enabled
- Made it so that we fallback to the regular `showDialog` implementation if the windowing implementation fails

## How to test
On Win32, macOS, and Linux:

1. `fluter config --enable-windowing`
2. In `examples/api_multiwindow`: `flutter run lib/material/dialog/dialog.0.dart`
3. Click the two buttons and see the true dialogs

Also, run `flutter test` and see that the tests pass

## Pre-launch Checklist

- [x] I signed the [CLA].


examples/api_multiwindow/test/material/dialog/dialog_0_test.dart
```diff
@@ -0,0 +1,70 @@
...
+// ignore: avoid_relative_lib_imports
+import '../../../lib/material/dialog/dialog.0.dart' as dialog_0;
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/src/foundation/_features.dart' show isWindowingEnabled;
+
+void main() {
+  if (!isWindowingEnabled) {
+    const String windowingDisabledErrorMessage = '''
+Skipping dialog_0.dart because Windowing APIs are not enabled.
```
robert-ancell: `dialog_0.dart` -> `dialog.0.dart`

examples/api_multiwindow/README.md
```diff
@@ -0,0 +1,17 @@
+# api_multiwindow
```
robert-ancell: Missing description of test app (other examples have this).

examples/api_multiwindow/analysis_options.yaml
```diff
@@ -0,0 +1,28 @@
+# This file configures the analyzer, which statically analyzes Dart code to
```
robert-ancell: Nit: This is the template, you could probably strip out the comments and make this smaller (other examples seem to have).

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1384,107 @@ Widget _buildMaterialDialogTransitions(
...
+  final _DialogWindowRoute<dynamic> route;
+
+  @override
+  void onWindowCloseRequested(DialogWindowController controller) {
+    route.navigator?.pop();
+  }
+
+  @override
+  void onWindowDestroyed() {
+    // Window has been destroyed
```
robert-ancell: Perhaps note why we don't need to handle this.

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1384,107 @@ Widget _buildMaterialDialogTransitions(
...
+  late final List<OverlayEntry> _overlayEntries;
+
+  @override
+  List<OverlayEntry> get overlayEntries => _overlayEntries;
+
+  @override
+  void install() {
+    super.install();
+
+    // Create a minimal transparent overlay entry to satisfy Navigator requirements
```
robert-ancell: Missing trailing '.'

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1384,107 @@ Widget _buildMaterialDialogTransitions(
...
+    // The actual dialog content is rendered through ViewAnchor, not through this overlay
```
robert-ancell: Missing trailing '.'

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1384,107 @@ Widget _buildMaterialDialogTransitions(
...
+        onPop: () => nav.pop(),
+      );
+      _registry?.register(_entry!);
+    }
+  }
+
+  @override
+  TickerFuture didPush() {
+    super.didPush();
+    // No animation is needed since the window appears instantly
```
robert-ancell: Missing trailing '.'

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1384,107 @@ Widget _buildMaterialDialogTransitions(
...
+
+  @override
+  bool didPop(T? result) {
+    _controller?.destroy();
+    return super.didPop(result);
+  }
+
+  @override
+  void dispose() {
+    // Unregister from the registry
```
robert-ancell: Missing trailing '.'

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+
+  final MaterialWindowRegistry _registry;
+
+  @override
+  bool updateShouldNotify(_MaterialWindowRegistryScope oldWidget) {
+    return _registry != oldWidget._registry;
+  }
+}
+
+// Wrapper that makes dialogs fill the entire window without insets or rounded corners
```
robert-ancell: Missing trailing '.'

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+          removeBottom: true,
+          context: context,
+          child: child,
+        ),
+      ),
+    );
+  }
+}
+
+// Provides a pop callback that dialog content can use
```
robert-ancell: Missing trailing '.'

robert-ancell: Other cases below too.

examples/api_multiwindow/README.md
justinmc: Example: https://github.com/flutter/flutter/blob/master/examples/platform_channel/README.md

examples/api_multiwindow/.gitignore
justinmc, line 1: What's the relation between this example (api_multiwindow) and the [multiple_windows example](https://github.com/flutter/flutter/tree/master/examples/multiple_windows)? I guess this is supposed to be like [examples/api](https://github.com/flutter/flutter/tree/master/examples/api) but work with multiwindow?

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
```diff
@@ -0,0 +1,98 @@
...
+                    onPressed: () {
+                      Navigator.pop(context);
+                    },
+                    child: const Text('Close'),
+                  ),
+                ],
+              ),
+            ),
+          ),
+          child: const Text('Show Fullscreen Dialog'),
```
justinmc: Nit: Consider pulling onPressed into its own method to make this easier to read. Or maybe pull the dialog into its own method or widget.

justinmc: Same with the "Show Dialog" button above actually.

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
```diff
@@ -0,0 +1,98 @@
...
+class DialogExample extends StatelessWidget {
+  const DialogExample({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: <Widget>[
+        TextButton(
+          onPressed: () => showDialog<String>(
```
justinmc: Can I still choose to show a non-windowed dialog when I'm in useWindowing mode? If not, I worry that an app-level useWindowing flag might not give people fine enough control.

packages/flutter/lib/src/material/dialog.dart
justinmc: I thought I was the only stickler about trailing periods, thank you @robert-ancell :laughing:

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1504,6 +1614,24 @@ Future<T?> showDialog<T>({
...
+          builder: builder,
+          parentController: WindowScope.maybeOf(context),
+          context: context,
+          settings: routeSettings,
+          preferredSize: fullscreenDialog ? parentSize : null,
+        ),
+      );
+    } catch (e) {
+      // Fallback to normal dialog route if windowing is not supported
+    }
```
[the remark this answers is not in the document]
justinmc: I was going to say the same thing!

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+// Wraps content to provide a Navigator-like interface for popping
+class _DialogPopScope extends StatelessWidget {
+  const _DialogPopScope({required this.child, this.onPop});
+
+  final Widget child;
+  final VoidCallback? onPop;
+
+  @override
+  Widget build(BuildContext context) {
+    // Wrap with WillPopScope to handle back button and provide popNavigator function
```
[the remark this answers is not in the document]
justinmc: +1

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+      onPopPage: (Route<dynamic> route, dynamic result) {
+        // When the page is popped, call our onPop callback
+        widget.onPop?.call();
+        // Return false to prevent the route from being removed from the Navigator
+        // (since we're handling the pop externally by closing the dialog window)
+        return false;
+      },
+    );
+  }
+}
```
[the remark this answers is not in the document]
justinmc: Agreed.

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+class _NavigatorShimState extends State<_NavigatorShim> {
+  @override
+  Widget build(BuildContext context) {
+    // Create a Navigator with a single page that contains the child
+    // This allows Navigator.pop(context) calls from within the dialog to work
+    return Navigator(
+      pages: <Page<void>>[_DialogContentPage(child: widget.child)],
+      onPopPage: (Route<dynamic> route, dynamic result) {
+        // When the page is popped, call our onPop callback
+        widget.onPop?.call();
```
justinmc: Is this mainly so that users can programmatically call Navigator.pop and have it close the dialog window? Does it actually integrate with the Navigator system beyond that? Like I wonder if a user programmatically messed with the navigation stack to remove the dialog or something. Or does the Navigator even have a route for the dialog?

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,306 @@
...
+            view: subViews.isNotEmpty ? ViewCollection(views: subViews) : null,
+            child: child!,
+          );
+        },
+        child: widget.child,
+      ),
+    );
+  }
+
+  Widget _buildDialog(MaterialWindowEntry entry, DialogWindowController controller) {
```
justinmc: The build method should be the last thing here, see https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

justinmc (commented):
My main points:

 * I worry about someone that wants to show a dialog window in one case and show a regular Overlay dialog in another case. What would that take with this API?
 * I want to make sure the integration with PopScope etc. works in all cases. Flutter's navigation has a lot of edge cases.

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1504,6 +1616,24 @@ Future<T?> showDialog<T>({
...
+        _DialogWindowRoute<T>(
+          builder: builder,
+          parentController: WindowScope.maybeOf(context),
+          context: context,
+          settings: routeSettings,
+          preferredSize: fullscreenDialog ? parentSize : null,
+        ),
+      );
+    } on UnsupportedError {
+      // Fallback to normal dialog route if windowing is not supported
```
justinmc: Would it be useful to log a FlutterError here? I guess this is likely to happen in a multiplatform app where one of the platforms doesn't support multiwindow. Maybe the user doesn't care since it will fall back to a Flutter dialog.

examples/api_multiwindow/.gitignore
loic-sharma: If we keep this, what do y'all think of these alternative names instead?

* `multiple_windows_api`
* `experimental_api`

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1504,6 +1616,24 @@ Future<T?> showDialog<T>({
...
+      final Size? parentSize = WindowScope.maybeContentSizeOf(context);
+      return Navigator.of(context, rootNavigator: useRootNavigator).push<T>(
+        _DialogWindowRoute<T>(
+          builder: builder,
+          parentController: WindowScope.maybeOf(context),
+          context: context,
+          settings: routeSettings,
+          preferredSize: fullscreenDialog ? parentSize : null,
+        ),
+      );
```
justinmc: 1. Would there ever be a reason where a user would want to show a windowed dialog but *not* via showDialog? Maybe we just keep an eye out for any GitHub issues related to this, but for now keep showDialog the only way.

2. In the looong run this could be a breaking change, because someone with a desktop app that uses showDialog will suddenly get their dialog in a real window when multiwindow is turned on by default. That's probably not a big deal, but we could add a parameter to showDialog to opt in if we were worried.

3. What if an app dev has multiwindow enabled, but they want to show a Flutter-rendered dialog for some reason? They couldn't use showDialog anymore, but they could manually push a Dialog route as a workaround. Adding an opt-in parameter to showDialog as mentioned in 2 above might allow them to use showDialog. (I'm not sure about the parameter idea, just making sure we explore these options.)

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
```diff
@@ -0,0 +1,107 @@
...
+
+// ignore_for_file: implementation_imports
+// ignore_for_file: invalid_use_of_internal_member
+
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+
+/// Flutter code sample for [Dialog].
```
loic-sharma: I imagine we'll want to move this sample to `examples/api/lib/material/dialog/` in the future? If so, this name will conflict with the existing `dialog.0.dart` sample. What do you think of renaming this file to `dialog_window.0.dart`?

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,208 @@
...
+
+Windowing APIs are currently experimental. Do not use windowing APIs in
+production applications or plugins published to pub.dev.
+
+To try experimental windowing APIs:
+1. Switch to Flutter's main release channel.
+2. Turn on the windowing feature flag.
+
+See: https://github.com/flutter/flutter/issues/30701.
+''';
```
loic-sharma: What do you think of renaming this file from `material_windowing_manager.dart` to `_window.dart` or `_windowing.dart`? The `_` prefix makes it clearer that this is experimental and that this stuff isn't exported. Also a more generic file name means we can shove more things in here as needed :)

packages/flutter/lib/src/material/material_windowing_manager.dart
```diff
@@ -0,0 +1,208 @@
...
+/// pub.dev. Flutter will make breaking changes to this API, even in patch
+/// versions.
+///
+/// This API throws an [UnsupportedError] error unless Flutter’s windowing
+/// feature is enabled by [isWindowingEnabled].
+///
+/// See: https://github.com/flutter/flutter/issues/30701.
+/// {@endtemplate}
+@internal
+class MaterialWindowRegistry extends ChangeNotifier {
```
justinmc: Now that I think about this, how much stuff is there in this PR that actually relates to Material? showDialog and I see a themeData parameter below. There is also a [cupertinoShowDialog](https://main-api.flutter.dev/flutter/cupertino/showCupertinoDialog.html), have you thought about that at all? I guess it's less likely to be used in a desktop app, maybe a macos app, but maybe it should have the same multiwindow behavior as showDialog.

I'm thinking maybe we should make this code design-language-agnostic, where possible. We should think about what the story is for dialog windows outside of the context of Material/Cupertino, and then tie into showDialog etc. from there.

justinmc (commented):
I like this approach a lot more than the MaterialApp parameter :+1: .

About my [previous concern](https://github.com/flutter/flutter/pull/181861#pullrequestreview-3748318445) about the integration with Flutter's Navigator: I see now you have _DialogWindowRoute and you're adding it to the route stack like anything else, which looks like it should work like any other route. Would you be able to write a test or example that uses the Navigator 2.0 [pages](https://main-api.flutter.dev/flutter/widgets/Navigator/pages.html) API to make sure this approach is compatible with that?

One QA thing I wanted to check, probably unrelated to this PR: When I tried running the example on my Linux machine, the dialog window was not draggable independently of the parent window. If I tried grabbing it and dragging, the whole parent window moved. Is that how it's supposed to work?

packages/flutter/lib/src/material/dialog.dart
mattkae: We should just not override it.

examples/api_multiwindow/.gitignore
mattkae: Yeah this is supposed to be `api but with with multiwindow demonstrations`. The reason it needs to be separate is that the runner is different (for now).

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
mattkae: That will be a flag that comes later. We didn't want to introduce a new flag into `showDialog` just yet. We can do it now if you think it is appropriate.

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
mattkae: In the future, I assume that the regular dialog example will be the same exact thing, but just with a different behavior on desktop (i.e. a "true" window).

packages/flutter/lib/src/material/material_windowing_manager.dart
mattkae: I removed it altogether :tada:

packages/flutter/lib/src/material/material_windowing_manager.dart
mattkae: Yes that's what it's mainly for.

The Navigator will have a route for this dialog under the `_DialogWindowRoute`

packages/flutter/lib/src/material/dialog.dart
mattkae: Our understanding is that the user shouldn't necessarily care, as they will always fall back to a sane default.

For example, I can imagine a make-pretend platform where dialogs only fail to be created _sometimes,_ like when they are on a small display or something. This would let us do the true windowing behavior when we have a big display, and otherwise do the in-window behavior.

This is hypothetical of course, but this design of a "silent fallback" makes such platforms a possibility, which is pretty neat

packages/flutter/lib/src/material/dialog.dart
mattkae: 1. Yes, but then they'd have to wrap it in the `DialogWindow` widget themselves and manage that manually. It would be more work, but it is certainly available.
2. Yes, I think we should introduce a flag to force the behavior to be non-multiwindow if they wan to opt-out. There's nothing wrong with that in my opinion.
3. I think a parameter might be a good idea. In fact, we were thinking of having the parameter on the `MaterialApp` itself, i.e. `useWindowing` so that they can toggle all windowing behavior globally.

packages/flutter/lib/src/material/material_windowing_manager.dart
mattkae: I used at least the registry out, since that seems the most generic. I wonder if anything else would be a good idea here, I'm trying to tease that out now

examples/api_multiwindow/README.md
```diff
@@ -0,0 +1,19 @@
...
+## Running
+
+Each file in the `lib` folder is a self-contained example program. You may run
+each program via:
+
+```bash
+flutter run <lib/path/to/file.dart>
+```
+
+For example, to run the Material dialog tests, run:
```
justinmc: By "tests" did you mean "example"? I assume the tests should be run with `flutter test`.

packages/flutter/lib/src/material/app.dart
```diff
@@ -18,10 +18,20 @@ import 'dart:ui' as ui;
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter/src/foundation/_features.dart' show isWindowingEnabled;
+import '../widgets/_window.dart'
```
justinmc: Instead of a relative import could you import something like this? "package:flutter/widgets/src/_window.dart"

packages/flutter/lib/src/material/app.dart
```diff
@@ -1269,3 +1281,192 @@ class _MaterialInspectorButton extends InspectorButton {
...
+  final VoidCallback? onPop;
+
+  @override
+  Widget build(BuildContext context) {
+    // Wrap with PopupScope to handle back button and provide popNavigator function.
+    return PopScope(
+      canPop: false,
+      onPopInvokedWithResult: (bool didPop, Object? result) {
+        if (!didPop) {
+          onPop?.call();
```
justinmc: Should the result be passed into onPop?

packages/flutter/lib/src/material/app.dart
```diff
@@ -1269,3 +1281,192 @@ class _MaterialInspectorButton extends InspectorButton {
...
+        ? Directionality.of(entry.parentContext!)
+        : null;
+    final ThemeData? themeData = entry.parentContext != null
+        ? Theme.of(entry.parentContext!)
+        : null;
+    final MediaQueryData? mediaQuery = entry.parentContext != null
+        ? MediaQuery.of(entry.parentContext!)
+        : null;
+    final Widget dialogContent = _DialogPopScope(
+      onPop: Navigator.of(entry.parentContext!).pop,
```
justinmc: Does this really need the parentContext? It looks to me like _DialogPopScope should get the same Navigator using its own BuildContext, but I could be missing a layer somewhere.

packages/flutter/lib/src/material/app.dart
```diff
@@ -1269,3 +1281,192 @@ class _MaterialInspectorButton extends InspectorButton {
...
+        },
+      ),
+    );
+
+    return DialogWindow(
+      controller: controller,
+      child: Directionality(
+        textDirection: textDirection ?? Directionality.of(context),
+        child: Theme(
+          data: themeData ?? Theme.of(context),
```
justinmc: Alright here is a use of Material.

packages/flutter/lib/src/material/app.dart
```diff
@@ -1269,3 +1281,192 @@ class _MaterialInspectorButton extends InspectorButton {
...
+  Widget build(BuildContext context) {
+    final DialogThemeData windowDialogTheme = DialogTheme.of(context).copyWith(
+      insetPadding: EdgeInsets.zero,
+      shape: const RoundedRectangleBorder(), // No rounded corners.
+      alignment: Alignment.topLeft, // Align to top-left so it fills from corner.
+      constraints:
+          const BoxConstraints.expand(), // Remove default constraints so dialog can expand to fill available space.
+    );
+
+    return DialogTheme(
```
justinmc: Here's another use of Material, but not useful to Widgets.

packages/flutter/lib/src/material/app.dart
```diff
@@ -1269,3 +1281,192 @@ class _MaterialInspectorButton extends InspectorButton {
...
+    final Widget dialogContent = _DialogPopScope(
+      onPop: Navigator.of(entry.parentContext!).pop,
+      child: Builder(
+        builder: (BuildContext innerContext) {
+          return _FullWindowDialogWrapper(child: entry.builder(innerContext));
+        },
+      ),
+    );
+
+    return DialogWindow(
```
justinmc: Do we need a Semantics with SemanticsRole.dialog somewhere?

packages/flutter/lib/src/material/app.dart
justinmc: This is the only real use of Material that I'm seeing that prevents this from going to Widgets.

packages/flutter/lib/src/material/app.dart
```diff
@@ -1160,7 +1170,9 @@ class _MaterialAppState extends State<MaterialApp> {
 
     return ScrollConfiguration(
       behavior: widget.scrollBehavior ?? const MaterialScrollBehavior(),
-      child: HeroControllerScope(controller: _heroController, child: result),
+      child: _WindowManager(
+        child: HeroControllerScope(controller: _heroController, child: result),
```
justinmc: It would be really nice if we could get _WindowManager in at the Widgets level instead of Material, so that this can be used in Widgets and Cupertino without duplicating all this code.

What if you pulled out your _buildDialog method below (or part of it), and turned it into a parameter on WidgetsApp? Then you could move almost all of the code in this file into widgets/app.dart, and in this file you'd just pass in a builder function to that parameter that makes use of Theme.

Let me know if you think that would work. I think the parameter could be useful for other types of windows as well. Did you imagine a similar mechanism for non-dialog windows?

packages/flutter/lib/src/material/dialog.dart
justinmc: Ok I like the idea of a parameter of some sort... You have a point that most users might just want to set the windowed/non-windowed parameter once globally and forget about it. Could we make it an InheritedWidget or put it in MediaQuery or something like that?

packages/flutter/lib/src/material/dialog.dart
justinmc: Have you considered logging the error in case the developer cares, but continuing execution without rethrowing?

```
} on UnsupportedError (error, stacktrace) {
  FlutterError.reportError(
    FlutterErrorDetails(
      exception: error,
      library: 'material library',
      stack: stacktrace,
    ),
  );
}
```

packages/flutter/lib/src/material/app.dart
mattkae: This is the result unfortunately:

```
Navigator operation requested with a context that does not include a Navigator.
The context used to push or pop routes from the Navigator must be that of a widget that is a
descendant of a Navigator widget.

The relevant error-causing widget was:
  MaterialApp
  MaterialApp:file:///home/matthew/Github/flutter/examples/api_multiwindow/lib/material/dialog/dialog.0.dart:38:12

When the exception was thrown, this was the stack:
#0      Navigator.of.<anonymous closure> (package:flutter/src/widgets/navigator.dart:2929:9)
#1      Navigator.of (package:flutter/src/widgets/navigator.dart:2936:6)
#2      _WindowManagerState._buildDialog (package:flutter/src/material/app.dart:1307:24)
#3      _WindowManagerState.build.<anonymous closure>.<anonymous closure>
(package:flutter/src/material/app.dart:1341:22)
```

packages/flutter/lib/src/material/app.dart
mattkae: Yes I imagine a similar mechanism for non-dialog windows. Let me try this out!

mattkae: Alright, I think I have something good here

packages/flutter/lib/src/material/app.dart
mattkae: That comes from the `Dialog` widget which users render inside of `showDialog` from what I can tell:

https://github.com/flutter/flutter/blob/bacb47284f209337fbae4011636ab915474629f3/packages/flutter/lib/src/material/dialog.dart#L84

packages/flutter/lib/src/material/app.dart
mattkae: Yes, I let the `showDialog` code take care of this wrapper

packages/flutter/lib/src/material/dialog.dart
mattkae: We could check if there is a `WindowRegistry` in the context, which would only be constructed if the `WindowManager` wraps the root of the MaterialApp. If we don't want windowing, we can just not construct one of those and therefore we'd have no way to add the dialog window.

packages/flutter/lib/src/material/app.dart
mattkae: Yeah, which is why I'm punting the wraps here to the users of the `WindowManager` instead. It's useful to have full control at that level. It reduces potential "surprises"

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -14,7 +14,16 @@ import 'dart:ui' show SemanticsHitTestBehavior, SemanticsRole, clampDouble, lerp
 
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
-
+import '../foundation/_features.dart' show isWindowingEnabled;
+
+import '../widgets/_window.dart'
```
justinmc: (Let me know if this has been discussed somewhere already) This won't work post-decoupling when Material is in its own package. I guess we won't be able to use feature flagged stuff at all at that point? Even if Material and Cupertino had their own feature flag setups, maybe there would be no good way to use flagged stuff in the main repo.

justinmc: This is another strong argument for putting as much of the multiwindow code as we can into the widgets library.

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
```diff
@@ -0,0 +1,115 @@
...
+
+void main() {
+  WidgetsFlutterBinding.ensureInitialized();
+  final RegularWindowController controller = RegularWindowController(
+    preferredSize: Size(800, 600),
+    delegate: _RegularWindowControllerDelegate(),
+  );
+  runWidget(
+    RegularWindow(controller: controller, child: const DialogExampleApp()),
+  );
```
justinmc: Cool, so I just set up multiwindow globally, and then showDialog opens in a new window, no parameter needed. If ever in the future users want the ability to do both in one app, then we can always revisit this and it won't be hard to add a parameter. No need to add maintenance burden for that now.

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1381,10 +1390,210 @@ Widget _buildMaterialDialogTransitions(
...
+
 /// Displays a Material dialog above the current contents of the app, with
 /// Material entrance and exit animations, modal barrier color, and modal
 /// barrier behavior (dialog is dismissible with a tap on the barrier).
 ///
+/// If windowing is enabled via `flutter config --enable-windowing`,
+/// then the dialog is displayed in its own window using the windowing system,
+/// rather than as a modal overlay within the current window. This will only
+/// function on platforms that support the dialog window type. True dialog
+/// windows will lack animations and barriers.
```
justinmc: It's not quite this easy because you still have to setup multiwindow by wrapping your app in RegularWindow right? Maybe add a link to your example to show how this is done?

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1421,12 +1635,18 @@ Widget _buildMaterialDialogTransitions(
...
+/// manage focus traversal independently.
 ///
 /// {@template flutter.material.dialog.requestFocus}
 /// The `requestFocus` argument is used to specify whether the dialog should
 /// request focus when shown.
 /// {@endtemplate}
 /// {@macro flutter.widgets.navigator.Route.requestFocus}
+/// If windowing is enabled via `flutter config --enable-windowing`, then this
+/// argument is ignored as dialogs are displayed in their own windows which are
+/// focused by the windowing system.
```
justinmc, line 1567: Seems like a reasonable thing to want to work even in multiwindow, though. Is it possible? (Maybe not for this PR)

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1504,6 +1724,50 @@ Future<T?> showDialog<T>({
     to: Navigator.of(context, rootNavigator: useRootNavigator).context,
   );
 
+  final WindowRegistry? windowingConfiguration = WindowRegistry.maybeOf(context);
+  if (windowingConfiguration != null && isWindowingEnabled) {
+    try {
+      final Size? parentSize = WindowScope.maybeContentSizeOf(context);
+      final NavigatorState nav = Navigator.of(context, rootNavigator: useRootNavigator);
+      return nav.push<T>(
```
justinmc: Nit: I would call this `navigator` per https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-abbreviations

justinmc (commented):
The one thing that is probably a blocker here is the cross-library import that I commented on below. The effort to decouple is moving along so it might not be something we can punt on.

packages/flutter/lib/src/material/dialog.dart
mattkae: We can import the implementation from the package instead of an absolute path, right?

That would be a good stop-gap to make sure that Material can be decoupled, and once the windowing API is no longer experimental, it should be easy enough to adjust the import in Material.

What do you think?

packages/flutter/lib/src/material/dialog.dart
mattkae: This actually isn't strictly necessary  because failing to wrap your app in `RegularWindow` will just cause the `WindowScope` to not find a parent, which will cause the dialog to be modeless.

Should I call out that behavior here instead? Or do we not want modeless dialogs in the Material API? I can imagine that they might be unexpected for some people.

packages/flutter/lib/src/material/dialog.dart
mattkae: How would that work exactly? Would the first focusable element of the dialog be focused immediately? if so, then we can make this happen

packages/flutter/lib/src/material/dialog.dart
justinmc: Oh I see, I guess just call out the behavior here. If a user finds this unexpected they will probably look at these docs for clarity.

packages/flutter/lib/src/material/app.dart
```diff
@@ -1158,9 +1160,14 @@ class _MaterialAppState extends State<MaterialApp> {
       return true;
     }());
 
+    Widget child = HeroControllerScope(controller: _heroController, child: result);
+    if (isWindowingEnabled) {
+      child = WindowManager(child: child);
+    }
```
justinmc: I think you could stick this in WidgetsApp.

packages/flutter/lib/src/material/dialog.dart
justinmc: Yes, here's how it's done for ModalRoute:

https://github.com/flutter/flutter/blob/544684634bd0518d50804d60d1b6db0446d7a518/packages/flutter/lib/src/widgets/routes.dart#L1690-L1693

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1516,6 +1736,50 @@ Future<T?> showDialog<T>({
     to: Navigator.of(context, rootNavigator: useRootNavigator).context,
   );
 
+  final WindowRegistry? windowingConfiguration = WindowRegistry.maybeOf(context);
```
justinmc: Tomorrow I'm going to check out your branch and see if I can abstract the windowing parts here into a showRawDialog in the widgets library, or something like that.

packages/flutter/lib/src/material/dialog.dart
mattkae: But does this work if the dialog is an actual window? A modal dialog will automatically get focus, and the first node in its tree will also automatically get focus as a result of the window opening. Unless I am missing something?

packages/flutter/lib/src/material/dialog.dart
mattkae: Ok let me know if you need help!

packages/flutter/lib/src/material/dialog.dart
justinmc: Oh sorry I was assuming it would not get focus by default. In that case then I agree with leaving this as-is. Users might want finer grained control over focus here but we can deal with that if/when it comes.

justinmc:
FYI running the example on Linux and clicking "Show Dialog" gives me an error:

```
══╡ EXCEPTION CAUGHT BY WIDGET LIBRARY ╞════════════════════════════════════════════════════════════
The following assertion was thrown:
A HeroController can not be shared by multiple Navigators. The Navigators that share the same
HeroController are:
- NavigatorState#3fc4e(tickers: tracking 1 ticker)
- NavigatorState#ddbb1(tickers: tracking 1 ticker)
Please create a HeroControllerScope for each Navigator or use a HeroControllerScope.none to prevent
subtree from receiving a HeroController.

When the exception was thrown, this was the stack:
#0      NavigatorState._updateHeroController.<anonymous closure>.<anonymous closure> (package:flutter/src/widgets/navigator.dart:3981:41)
#1      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1430:15)
#2      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1357:11)
#3      SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1198:5)
#4      _invoke (dart:ui/hooks.dart:423:13)
#5      PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:447:5)
#6      _drawFrame (dart:ui/hooks.dart:395:31)
════════════════════════════════════════════════════════════════════════════════════════════════════
```

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+        DialogWindowControllerDelegate,
+        WindowEntry,
+        WindowRegistry,
+        WindowScope;
+
+typedef RouteBuilder<T> = Route<T> Function(BuildContext context, WidgetBuilder builder);
+
+/// Displays a dialog over the contents of the app.
+///
+/// When windowing is available, the dialog is a separate window.
```
justinmc: By the time windowing is public, we would want these docs to explain how to make windowing available (or link to something or use a macro). How have we been explaining this elsewhere? Or is this actually the first public API that uses windowing under the hood?

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+///
+/// The parameter `routeBuilder` builds the [Route] that will be pushed to the
+/// [Navigator]. Defaults to building a [RawDialogRoute]. When windowing is
+/// available, this parameter will be silently ignored.
+///
+/// The parameter `builder` builds the content of the dialog.
```
justinmc: Any other docs that should be copied over (or macro'd) from showDialog?

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+        anchorPoint: anchorPoint,
+        traversalEdgeBehavior: traversalEdgeBehavior ?? TraversalEdgeBehavior.closedLoop,
+        requestFocus: requestFocus,
+        fullscreenDialog: fullscreenDialog,
+      );
+
+  return navigator.push<T>(route);
+}
+
+// TODO(justinmc): This was copied from Material, is it needed?
```
justinmc: Maybe we can get rid of the _debugIsActive in Material and only use this one?

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+bool _debugIsActive(BuildContext context) {
+  if (context is Element && !context.debugIsActive) {
+    throw FlutterError.fromParts(<DiagnosticsNode>[
+      ErrorSummary('This BuildContext is no longer valid.'),
+      ErrorDescription(
+        'The showDialog function context parameter is a BuildContext that is no longer valid.',
```
justinmc: If we do keep this one, then update showDialog to showRawDialog here and below.

packages/flutter/lib/src/material/dialog.dart
```diff
@@ -1515,27 +1651,63 @@ Future<T?> showDialog<T>({
...
+        themes: themes,
+        anchorPoint: anchorPoint,
+        traversalEdgeBehavior: traversalEdgeBehavior ?? TraversalEdgeBehavior.closedLoop,
+        requestFocus: requestFocus,
+        animationStyle: animationStyle,
+        fullscreenDialog: fullscreenDialog,
+      );
+    },
+    builder: (BuildContext context) {
+      // Wrap the build dialog with the theme, text direcction, and media
```
justinmc: direcction spelling

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+  bool fullscreenDialog = false,
+  bool? requestFocus,
+}) {
+  assert(_debugIsActive(context));
+  assert(debugCheckHasWidgetsLocalizations(context));
+
+  final NavigatorState navigator = Navigator.of(context, rootNavigator: useRootNavigator);
+
+  final WindowRegistry? windowingConfiguration = WindowRegistry.maybeOf(context);
+  if (windowingConfiguration != null && isWindowingEnabled) {
```
justinmc: Nit: I try to keep the names of the variables and their types the same (`final WindowRegistry? windowRegistry`), but not a big deal if there's a reason for it.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+    return context.dependOnInheritedWidgetOfExactType<WindowRegistryScope>()?._registry;
+  }
+}
+
+/// An inherited widget that provides access to the [WindowRegistry].
+////
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+class WindowRegistryScope extends InheritedWidget {
+  /// Creates a Material window registry scope.
```
justinmc: No longer "Material" right?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+  @override
+  bool updateShouldNotify(WindowRegistryScope oldWidget) {
+    return _registry != oldWidget._registry;
+  }
+}
+
+/// Represents an entry for a Material window such as a dialog.
+///
+/// This class holds the necessary information to build and manage
+/// a window within the Material application.
```
justinmc: Two more references to "Material" that should go.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [WindowRegistry], where window entries can be registered.
+@internal
+class WindowEntry {
+  /// Creates a Material window entry.
```
justinmc: No "Material"

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+///
+/// If windowing is not enabled, this widgets renders
+/// [child] and does not provide the [WindowRegistryScope].
+///
+/// {@tool snippet}
+/// An example usage might look like this, where the window manager wraps
+/// the root of the widget tree so that dialogs can be rendered at the same level
+/// as a [RegularWindow].
+///
+///
```
justinmc: Should just be one empty line.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+///   final WindowRegistry registry;
+///
+///   @override
+///   void onWindowDestroyed() {
+///     super.onWindowDestroyed();
+///     mainWindow.closeDialog(registry);
+///   }
+/// }
+///
+/// ```
```
justinmc: Have you thought about putting this example in the examples directory? It looks like it's a complete app already. If you put it there then we can guarantee it still runs in the future.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,325 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+/// An example usage might look like this, where the window manager wraps
+/// the root of the widget tree so that dialogs can be rendered at the same level
+/// as a [RegularWindow].
+///
+///
+/// ```dart
+/// // TODO(mattkae): remove invalid_use_of_internal_member ignore comment when this API is stable.
+/// // ignore_for_file: invalid_use_of_internal_member
+/// // ignore_for_file: implementation_imports
+/// import 'package:flutter/material.dart';
```
justinmc: Can you convert this example to use Widgets only? Maybe you just need to get rid of MaterialApp and Scaffold.

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+        _DialogWindowRoute<T>(
+          builder: builder,
+          parentController: WindowScope.maybeOf(context),
+          context: context,
+          settings: routeSettings,
+          preferredSize: fullscreenDialog ? parentSize : null,
+        ),
+      );
+    } on UnsupportedError catch (error, stacktrace) {
+      // Fallback to normal dialog route if windowing is not supported
```
justinmc: Period at the end here.

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
justinmc, line 1: Should we reference the existence of this example anywhere?

justinmc: Also, does this example really need to be in its own directory (api_multiwindow) or could it just be in api with most other examples?

justinmc (approved):
LGTM with nits, assuming tests will pass 👍 . Also probably @loic-sharma should double confirm my API here. Thanks for integrating my showRawDialog idea, I like how it turned out.

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,186 @@
...
+/// Displays a dialog over the contents of the app.
+///
+/// When windowing is available, the dialog is a separate window.
+///
+/// The parameter `routeBuilder` builds the [Route] that will be pushed to the
+/// [Navigator]. Defaults to building a [RawDialogRoute]. When windowing is
+/// available, this parameter will be silently ignored.
+///
+/// The parameter `builder` builds the content of the dialog.
+Future<T?> showRawDialog<T>({
```
victorsanni, line 69: Looking at the impl. of `showDialog`, it only pushes `DialogRoute`. Why are we abstracting `showDialog` instead of just using Navigator.push on the route?

packages/flutter/lib/src/widgets/dialog.dart
mattkae: This is the first public API that uses windowing under the hood. However, the wording in this comment should remain valid once windowing is available, as it points to whether windowing is available on your platform or not. Should I be more specific there?

packages/flutter/lib/src/widgets/dialog.dart
mattkae: We can macro a bunch

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
mattkae: Unfortunately, it needs to because it has a different runner (i.e. one that does NOT create an initial window)

mattkae: Or perhaps this doesn't matter? I will try it out

packages/flutter/lib/src/widgets/dialog.dart
justinmc: Looks good but now you should remove this TODO.

examples/api/test/widgets/windows/window_manager.0_test.dart
```diff
@@ -0,0 +1,15 @@
...
+import 'package:flutter_api_samples/widgets/windows/window_manager.0.dart'
+    as example;
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  testWidgets('Calling window_manager main returns normally', (
+    WidgetTester tester,
+  ) async {
+    expect(() => example.main(), returnsNormally);
+  });
```
justinmc, line 14: Would it be possible to pump the example and expect that "Unsupported" is shown?

examples/api_multiwindow/lib/material/dialog/dialog.0.dart
justinmc: Looks good after the move if it works!

packages/flutter/lib/src/widgets/dialog.dart
justinmc: Is it possible to just reference the `[WindowManager]` docs here? Those docs explain everything a user would need to get windowing set up. You could just say "See [WindowManager] for more on how to set up windowing."

justinmc (approved):
I'm currently discussing my API with @victorsanni but otherwise renewing my LGTM.

examples/api/test/widgets/windows/window_manager.0_test.dart
mattkae: I left it as "returns normally" so that when we "turn on" multi-window, it will _just work_

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,244 @@
...
+/// [Navigator.pop] when the dialog was closed.
+///
+/// See also:
+///
+///  * [WindowManager], for more information on how to set up windowing.
+Future<T?> showRawDialog<T>({
+  required BuildContext context,
+  required WidgetBuilder builder,
+  RouteBuilder<T>? routeBuilder,
+  Color? barrierColor,
```
victorsanni: Why do we have all these parameters when `routeBuilder` already exists. Since we already have a routeBuilder, why not call it in the non-windowing case, then fallback to a bare-bones RawDialogRoute? and let customers of showRawDialog do all their route customizations in routeBuilder?

packages/flutter/lib/src/widgets/dialog.dart
victorsanni: I see, we want to add windowing to showDialog itself, but keep the logic out of Material/.

packages/flutter/lib/src/widgets/dialog.dart
justinmc: I agree with this, we shouldn't add all this API bloat just because showDialog has it. We can still keep the parameters that are used by _DialogWindowRoute (routeSettings, fullscreenDialog, parentSize), but I think most others used by RawDialogRoute can be removed.

mattkae: Agreed! Done :+1:

packages/flutter/lib/src/widgets/dialog.dart
loic-sharma: Re-opening as Justin's comment is still relevant :)

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+/// The registry is particularly useful when deep descendents of the widget
+/// tree create windows which are logically nested under a particular widget
+/// while at the same time need to be rendered at the root of the widget tree.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [WindowManager], responsible for listening for new windows and rendering them.
+///  * [WindowRegistryScope], the scope that provides access to the registry
```
loic-sharma:
```suggestion
///  * [WindowRegistryScope], the scope that provides access to the registry.
```

packages/flutter/lib/src/widgets/app.dart
```diff
@@ -1728,6 +1730,10 @@ class _WidgetsAppState extends State<WidgetsApp> with WidgetsBindingObserver {
       result = routing!;
     }
 
+    if (isWindowingEnabled) {
+      result = WindowManager(child: result);
+    }
```
loic-sharma, line 1735: This adds a `WindowManager` if windowing is enabled, even if the platform doesn't support multi-window (like Web). Is that desirable? Or should we not add `WindowManager` if the platform does not support multi-window?

The docs on `WindowRegistry.of` seem to indicate that platforms that don't support multi-window will return `null`.

I don't have a strong opinion here, I just want to check we thought this case out :)

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+class WindowManager extends StatefulWidget {
+  /// Creates a window manager.
+  ///
+  /// The [child] is the content inside of the window manager.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  WindowManager({super.key, required this.child});
+
+  final WindowRegistry _registry = WindowRegistry();
```
loic-sharma: Should this be created in `_WindowManagerState.initState`?

Keep in mind that whenever `WindowManager`'s parent rebuilds, the parent will create a new instance of `WindowManager` and pass that to `_WindowManagerState`. In other words, this will create a new `WindowRegistry` whenever `WindowManager`'s parent rebuilds.

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,215 @@
...
+    final BuildContext? routeContext = nav?.context;
+    if (routeContext != null && nav != null) {
+      _entry = WindowEntry(controller: _controller!, builder: builder);
+      _registry?.register(_entry!);
+    }
+  }
+
+  @override
+  TickerFuture didPush() {
+    super.didPush();
```
loic-sharma: Instead of creating our own `TickerFuture`, should we return the `TickerFuture` created by `super.didPush()`? It also completes immediately: https://api.flutter.dev/flutter/widgets/Route/didPush.html

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    _windows.add(entry);
+    notifyListeners();
+  }
+
+  /// Unregisters a window.
+  ///
+  /// The [entry] parameter specifies the window to unregister.
```
loic-sharma, line 1836: Are there ordering requirements between when you should call `BaseWindowController.destroy` vs when you should call `WindowRegistry.unregister`?

Do you need to call `WindowRegistry.unregister` before you call `BaseWindowController.destroy` to ensure a window widget never uses a destroyed window? Or does that not matter?

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,215 @@
...
+    // No animation is needed since the window appears instantly.
+    return TickerFuture.complete();
+  }
+
+  @override
+  bool didPop(T? result) {
+    _controller?.destroy();
```
loic-sharma, line 183: I'm surprised that we destroy the window controller before we unregister the window entry. I would've expected that you need to unregister the entry before you destroy the controller to ensure the window widget never references a destroyed window controller. Is that not the case?

Also, could you explain why we unregister the window entry in `dispose` instead of `didPop`?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
     );
   }
 }
+
+/// A listenable registry for windows.
+///
+/// Users should prefer to use the [WindowManager] to access an instance of the
+/// registry instead of declaring one themselves.
```
loic-sharma: I suspect this comment might be stale. Currently you use `WindowRegistry.of(context)` to get the registry, not `WindowManager`.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+
+  /// The builder function that builds the content of the window.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  final WidgetBuilder builder;
+}
+
+/// The window manager provides a convient way to render windows
+/// at the root of an application. Descendents of the [WindowManager] may
```
loic-sharma: Nit: The first paragraph should be only a single sentence as that's what shows up in the IDE summary.

```suggestion
/// at the root of an application.
///
/// Descendents of the [WindowManager] may
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+  static WindowRegistry? maybeOf(BuildContext context) {
+    return context.dependOnInheritedWidgetOfExactType<WindowRegistryScope>()?._registry;
+  }
+}
+
+/// An inherited widget that provides access to the [WindowRegistry].
+////
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+class WindowRegistryScope extends InheritedWidget {
```
loic-sharma: Should we make `WindowRegistryScope` private? AFAICT it is an implementation detail that a user never needs to access:

1. You use `WindowManager` to create a `WindowRegistryScope`
2. You use `WindowRegistry` to look up the `WindowRegistryScope`.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
     );
   }
 }
+
+/// A listenable registry for windows.
+///
+/// Users should prefer to use the [WindowManager] to access an instance of the
+/// registry instead of declaring one themselves.
+///
+/// Applications may wrap the root of their application in a [WindowRegistryScope].
```
loic-sharma: Should this be `WindowManager`? It seems like if you add a `WindowRegistryScope` directly without a `WindowManager`, the app won't behave as expected.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,237 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+                child: entry.builder(context),
+              ),
+              final PopupWindowController popup => PopupWindow(
+                controller: popup,
+                child: entry.builder(context),
+              ),
+            };
+          }).toList();
+
+          return ViewAnchor(
```
loic-sharma, line 1999: I'm surprised this uses a `ViewAnchor` as `ViewAnchor` requires an ancestor `View`, but `WindowManager` is meant to host top-level windows.

I would've expected this to need to use `View.maybeOf` to check if there's a parent view, and then switch between `ViewAnchor` or `ViewCollection` depending on whether there's a parent view. Is that not the case? Or is `ViewAnchor` not properly asserting that it has a parent view? Let me know if I'm misunderstanding what's happening here!

See: https://github.com/flutter/flutter/blob/f36a0991d3f40f8a7f5edeb733831cfc257f15d4/packages/flutter/lib/src/widgets/view.dart#L679-L680

packages/flutter/lib/src/widgets/app.dart
wanjm: isWindowingEnabled can be false if the platform doesn't support.

mattkae: You can add `WindowManager` if `isWindowingEnabled` is false, as it will just be a passthrough widget in that case. It will not supply a `WindowRegistryScope` then.

On top of that, we can just make it false in the future if the platform doesn't support windows at all.

packages/flutter/lib/src/widgets/dialog.dart
mattkae: You are right. We were probably getting lucky. We should listen to `onWindowCloseRequested` to call pop, which will trigger a dispose, which remove us from the registry and destroy the window.

packages/flutter/lib/src/widgets/_window.dart
mattkae: You're right! I think we always just want to return a `ViewCollection` here. It will just hold the views at the top of the tree and nothing else. I don't even think an anchor is needed at all.

mattkae: Oh. The anchor is obviously needed

packages/flutter/lib/src/widgets/dialog.dart
mattkae: If I do not put the call to unregister + destroy in `didPop`, I never get it in `dispose`, it seems.

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: @mattkae Follow-up to this, does `ViewAnchor` assert that it has an ancestor `View`? We should add that assert if it's missing, but that's not a blocker for this PR though.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,239 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+  final BaseWindowController controller;
+
+  /// The builder function that builds the content of the window.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  final WidgetBuilder builder;
+}
+
+/// The window manager provides a convient way to render windows
```
loic-sharma:
```suggestion
/// The window manager provides a convenient way to render windows
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,239 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+  late final WindowRegistry _registry;
+
+  @override
+  void initState() {
+    super.initState();
+    _registry = WindowRegistry();
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
loic-sharma, line 1963: Do you think it'd be worth checking that `_WindowRegistryScope`s aren't nested? For example, you might get unexpected behavior if you use `MaterialApp` and then manually add your own `WindowManager` to the widget tree. I don't feel strongly about this.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,239 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+  WindowRegistry() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+  }
+
+  final List<WindowEntry> _windows = <WindowEntry>[];
+
+  /// The list of registered windows.
+  List<WindowEntry> get windows => List<WindowEntry>.unmodifiable(_windows);
```
loic-sharma, line 1813: Should this be internal with the experimental docs macro?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,239 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+///
+/// The registry is particularly useful when deep descendents of the widget
+/// tree create windows which are logically nested under a particular widget
+/// while at the same time need to be rendered at the root of the widget tree.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+///
+/// See also:
+///
+///  * [WindowManager], responsible for listening for new windows and rendering them.
```
loic-sharma, line 1794: I'd consider moving the 5th paragraph - which explains when `WindowRegistry` is useful - up to the 2nd paragraph. I think that's useful context when reading the rest of the docs.

Maybe something like:

```suggestion
/// A registry used to render top-level windows.
///
/// The registry is often used to render top-level windows
/// that are logically nested under a widget deep in the tree.
///
/// The [WindowManager] provides a [WindowRegistry] to its descendents.
///
/// Descendents of the manager can use [WindowRegistry.maybeOf] to access the
/// registry. With the registry, they can call [WindowRegistry.register] to
/// add a new window to the registry and [WindowRegistry.unregister] to remove
/// a window from the registry.
///
/// {@macro flutter.widgets.windowing.experimental}
///
/// See also:
///
///  * [WindowManager], responsible for listening for new windows and rendering them.
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1774,3 +1774,239 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
...
+}
+
+class _WindowManagerState extends State<WindowManager> {
+  late final WindowRegistry _registry;
+
+  @override
+  void initState() {
+    super.initState();
+    _registry = WindowRegistry();
+  }
```
loic-sharma: Oh sorry, my suggestion to create this in `initState` was unnecessary, you can just initialize the field directly:

```suggestion
  final WindowRegistry _registry = WindowRegistry();
```

Or if you're feeling fancy:

```suggestion
  final WindowRegistry _registry = .new();
```

packages/flutter/lib/src/widgets/dialog.dart
```diff
@@ -0,0 +1,212 @@
...
+              Animation<double> secondaryAnimation,
+            ) => builder(context),
+        settings: routeSettings,
+        fullscreenDialog: fullscreenDialog,
+      );
+
+  return navigator.push<T>(route);
+}
+
+// TODO(justinmc): This was copied from Material, is it needed?
```
loic-sharma: @mattkae Bubbling up this comment, it looks like it got missed since it's on an old thread: https://github.com/flutter/flutter/pull/181861/changes#r2879491013

packages/flutter/lib/src/widgets/dialog.dart
loic-sharma: Looks good now, thanks!

packages/flutter/lib/src/widgets/app.dart
loic-sharma: Sounds good to me 👍

loic-sharma (approved):
Great work!

packages/flutter/lib/src/widgets/_window.dart
mattkae: Yes I check `maybeOf`, first. If it doesn't, I just return a `ViewCollection`.

packages/flutter/lib/src/widgets/_window.dart
mattkae: This sounds like an unnecessary restriction. You may even want it in some places I imagine. Let's leave it like this for now, and if it proves to be a problem, we can go back and add the restriction :)

packages/flutter/lib/src/widgets/_window.dart
mattkae: I like it!

packages/flutter/lib/src/widgets/dialog.dart
mattkae: We certainly need this in Material. I think we can remove it at this layer though. It needs to happen before much of the build in `showDialog`

PR #181869, opened 2026-02-03
Avoid unnecessary ImageFilter::makeWithLocalMatrix in ImageFilterLayer::Diff

### LINKED ISSUE #159313: ImageFilterLayer::Diff uses ImageFilter::makeWithLocalMatrix unnecessarily
The following code in `ImageFilterLayer` snapshots the current transform into the ImageFilter using `makeWithLocalMatrix`:

https://github.com/flutter/engine/blob/03eb3dbc10f34719448690e87d96aa80ca5d8160/flow/layers/image_filter_layer.cc#L37

An easier implementation would be to pass the current matrix into the lambda function to replace the identity `DlMatrix()` arg in the call to `map_device_bounds`. This would avoid the overhead of creating a new filter and having the unnecessary `DlLocalMatrixImageFilter` have to concatenate the 2 matrices on each bounds update.

Something like:

```
  if (filter_) {
    context->PushFilterBoundsAdjustment([ctm, filter_](SkRect rect) {
      DlIRect filter_out_bounds;
      filter_->map_device_bounds(ToDlIRect(rect.roundOut()), ctm,
                                filter_out_bounds);
      return SkRect::Make(ToSkIRect(filter_out_bounds));
    });
  }
```

AbdeMohlbi:
FIxes #159313
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/flow/layers/image_filter_layer.cc
+++ b/engine/src/flutter/flow/layers/image_filter_layer.cc
@@ -34,7 +34,7 @@ void ImageFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
   }
 
   if (filter_) {
-    DlMatrix current_matrix = context->GetMatrix();
+    const DlMatrix& current_matrix = context->GetMatrix();
     // This transform will be applied to every child rect in the subtree
     context->PushFilterBoundsAdjustment(
         [filter = filter_, current_matrix](DlRect rect) {
--- a/engine/src/flutter/flow/layers/image_filter_layer.cc
+++ b/engine/src/flutter/flow/layers/image_filter_layer.cc
@@ -35,6 +35,7 @@ void ImageFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
 
   if (filter_) {
     DlMatrix current_matrix = context->GetMatrix();
+    // This transform will be applied to every child rect in the subtree
     context->PushFilterBoundsAdjustment(
         [filter = filter_, current_matrix](DlRect rect) {
           DlIRect filter_out_bounds;
--- a/engine/src/flutter/flow/layers/image_filter_layer.cc
+++ b/engine/src/flutter/flow/layers/image_filter_layer.cc
@@ -34,16 +34,14 @@ void ImageFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
   }
 
   if (filter_) {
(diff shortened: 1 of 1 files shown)


engine/src/flutter/flow/layers/image_filter_layer.cc
```diff
@@ -34,16 +34,15 @@ void ImageFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
...
-    if (filter) {
-      // This transform will be applied to every child rect in the subtree
-      context->PushFilterBoundsAdjustment([filter](DlRect rect) {
-        DlIRect filter_out_bounds;
-        filter->map_device_bounds(DlIRect::RoundOut(rect), DlMatrix(),
-                                  filter_out_bounds);
-        return DlRect::Make(filter_out_bounds);
-      });
-    }
+    DlMatrix current_matrix = context->GetMatrix();
```
flar: Make this `const DlMatrix&` to avoid copying the contents of the matrix.

flar (changes requested):

engine/src/flutter/flow/layers/image_filter_layer.cc
flar:
```suggestion
    const DlMatrix& current_matrix = context->GetMatrix();
```

flar:
The easiest way would be to look at the existing difftest for ImageFilterLayer and either add another case to it that does the "transformlayer" sandwhich or just make a new `TEST_F` instance that just tests the condition you are looking at.

https://github.com/flutter/flutter/blob/cf4291157bea13eb597d49b905e00e37a58ef92f/engine/src/flutter/flow/layers/image_filter_layer_unittests.cc#L630

The method constructs a number of layers, puts them into a few different tree formations and then runs a diff test on them and checks the result. Some of them diff 2 different layer trees against each other, some of them diff a layer tree against an empty tree.

The EXPECT_EQ (equal) lines double check that the results match your expectations and will cause the test to fail if the condition is not met. There are other variants such as EXPECT_LT (less than) etc. There are also macros of the form ASSERT_EQ, ASSERT_LT when you want the test to halt if the condition fails, but mostly we use the EXPECT versions as they allow the other tests in the same unit test to also run.

AbdeMohlbi:
> @AbdeMohlbi you don't have to ask it to write code, just ask it to run the tests and it should show you how it's done.

I will try giving it another shot soon thanks!

PR #181877, opened 2026-02-03
Update web ui fragment shader tests

walley892:
Adds a bunch of tests for uniform setting functionality for custom fragment shaders on the web.

Deletes redundant tests.

Fixes a discovered issue in the uniform offset calculation. We were previously using the `location`, which is the integer offset of the uniform, not the offset in floats.


engine/src/flutter/lib/web_ui/test/ui/fragment_shader_test.dart
```diff
@@ -325,12 +326,112 @@ const String kManyArraysSksl = r'''
 }
 ''';
 
+// Simple shader with a float uniform
+// Generated from engine/src/flutter/lib/ui/fixtures/shaders/uniforms/float_uniform.frag
+// For changes, update that file and regenerate using impellerc.
```
gaaclarke, line 331: worth giving the impellerc invocation since it is a bit involved

gaaclarke (approved):
nice! thanks lgtm

engine/src/flutter/lib/web_ui/test/ui/fragment_shader_test.dart
walley892: Aye, aye captain. Added to the impellerc readme.

PR #181884, opened 2026-02-04
[flutter_tools] Deprecate web hot reload flag

nshahan:
Warn when the flag is passed, and prompt to file an issue if disabling is required for a web development workflow.

The flag is enabled by default. It is still functional, but will be removed in an upcoming release.


mdebbar (approved):
LGTM

PR #181885, opened 2026-02-04
Improve FlWindowMonitor API

robert-ancell:
Use named parameters and allow callbacks to be omitted.


packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -434,22 +434,23 @@ class _FlView extends _GtkWidget {
...
-    _GtkWindow window,
-    void Function() onConfigure,
-    void Function() onStateChanged,
-    void Function() onIsActiveNotify,
-    void Function() onTitleNotify,
-    void Function() onClose,
-    void Function() onDestroy,
-  ) {
+    _GtkWindow window, {
+    void Function()? onConfigure,
```
mattkae: Nit: You can probably use `VoidCallback` for these from `dart:ui`, but I am ok either way

mattkae (approved):
Just a suggestion, but good by me!

PR #181889, opened 2026-02-04
[flutter_tools] Support flavors and transformers for shaders

### LINKED ISSUE #181888: Support transformers for flutter.shaders entries in pubspec.yaml
### Use case

`flutter.assets` entries can declare `transformers` for build-time processing. `flutter.shaders` historically only accepted a list of paths, which makes it hard to apply the same build-time transformation pipeline to fragment shaders.

When porting existing shaders (e.g. Shadertoy/GLSL Sandbox style sources) to Flutter runtime effects, developers often need to inject Flutter-specific boilerplate (headers/uniforms + an entrypoint bridge) while keeping the upstream shader logic intact.

Without build-time transformers for shaders, developers must either:
- manually edit shader sources (hard to keep in sync with upstream), or
- maintain duplicated “Flutterized” copies (noise + churn).

### Implemented (PR)

PR: https://github.com/flutter/flutter/pull/181889

This PR extends `flutter.shaders` to support the same map form used by assets entries, including `transformers`.

YAML sketch:
...
flutter:
  shaders:
    # Legacy string format (backwards compatible)
    - shaders/simple.frag

    # New map format with transformers
    - path: assets/shaders/shadertoy_port.frag
      transformers:
        - package: shadertoy_transpiler
          # args: [...] (optional, same as assets)
...

Semantics:
- If a shader entry declares `transformers`, the transformer pipeline runs first.
- The shader compiler (`impellerc`) is then invoked on the transformed output.

Backwards compatibility:
- The legacy string-list format remains supported.
- Only shader entries that explicitly declare `transformers` are transformed.

Non-goals:
- No change to `impellerc` / shader language compatibility.
- No Shadertoy-specific behavior in Flutter itself — this only enables a build-time hook.
- No change to built-in/material shaders.

### Tests

Added `flutter_tools` unit tests to verify:
- transformers are invoked for shader entries that declare them
- `impellerc` compiles the transformed output
- built-in/material shaders are not affected

Local verification command:
```
cd $FLUTTER_ROOT/packages/flutter_tools
FLUTTER_ROOT=$FLUTTER_ROOT ../../bin/cache/dart-sdk/bin/dart pub get
FLUTTER_ROOT=$FLUTTER_ROOT ../../bin/cache/dart-sdk/bin/dart test \
  test/general.shard/flutter_manifest_test.dart \
  test/general.shard/build_system/targets/assets_test.dart \
  -r expanded
```

### Optional example

I also put together a minimal example repo demonstrating how to declare shader transformers in `pubspec.yaml` and how the transformer output is used during shader compilation:

- Example: https://github.com/Laurie-Lin/shader_transformers_example

This is optional reference material; the implementation and tests live entirely in the Flutter repo.

Laurie-Lin:
This PR adds support for `transformers` on entries under `flutter: shaders:` in `pubspec.yaml`, bringing it closer to the existing assets entry model.

When a shader entry declares transformers, Flutter tools will run the transformer chain prior to shader compilation (Impeller/impellerc). This is wired through the build/bundle paths as well as DevFS shader recompilation during hot reload, so transformed sources are consistently used.

Fixes https://github.com/flutter/flutter/issues/181888

## Testing
```bash
cd $FLUTTER_ROOT/packages/flutter_tools
../../bin/flutter test test/general.shard/flutter_manifest_test.dart test/general.shard/build_system/targets/assets_test.dart -r expanded
```

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/lib/src/flutter_manifest.dart
```diff
@@ -567,17 +543,7 @@ void _validateFlutter(YamlMap? yaml, List<String> errors) {
...
-            'Expected "$yamlKey" to be a list, but got $yamlValue (${yamlValue.runtimeType}).',
-          );
-        } else if (yamlValue.isEmpty) {
-          break;
-        } else if (yamlValue[0] is! String) {
-          errors.add(
-            'Expected "$yamlKey" to be a list of strings, but the first element is $yamlValue (${yamlValue.runtimeType}).',
-          );
-        }
+        errors.addAll(_validateAssets(yamlValue));
```
chingjun, line 551: Correct me if I'm wrong. I don't believe we support `flavors` and `platforms` for shaders. Can we make sure that if `flavors` or `platforms` are passed, this part will fail?

Thanks!

packages/flutter_tools/lib/src/asset.dart
```diff
@@ -1116,9 +1117,9 @@ class ManifestAssetBundle implements AssetBundle {
         packageName: packageName,
         attributedPackage: attributedPackage,
         assetKind: AssetKind.shader,
-        flavors: <String>{},
-        platforms: <String>{},
-        transformers: <AssetTransformerEntry>[],
+        flavors: shaderEntry.flavors,
```
chingjun, line 1099: Same as the other comments, I think flavors and platforms are not supported yet, let's keep passing the empty set here.

chingjun (commented):
Mostly LGTM other than stricter input check needed.

Also, can we add tests for the changes in `bundle_builder.dart` and `devfs.dart`? Thanks!

packages/flutter_tools/lib/src/flutter_manifest.dart
Laurie-Lin: You are correct. However, I found that adding flavor support for shaders is not complicated, so I have implemented it and added the corresponding unit tests.

packages/flutter_tools/lib/src/asset.dart
Laurie-Lin: As mentioned in the other comment, I decided to support flavors here as well since it was straightforward.

bkonyi (approved):
Looks good to me! Just need signoff from @chingjun before we can submit this.

chingjun (approved):
LGTM, thanks for adding the support for flavors and platforms!

Issue #125975, opened 2023-05-03
Pressing and releasing a key before the framework has started throws


bleroux:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)
- [X] I have read the [guide to filing a bug](https://flutter.dev/docs/resources/bug-reports)

### Steps to reproduce

1. Launch the code sample (tried on Linux and on Android).
2. Do a hot restart and very quickly focused the Flutter app, press and release a key (this might need several tries, because the key has to be pressed and released in the short time where the engine is ready to listen to platform keyboard events and the framework is not fully initialized).

### Expected results

No error

### Actual results

Error thrown: `flutter: A KeyUpEvent is dispatched, but the state shows that the physical key is not pressed.`

### Code sample

<details><summary>Code sample</summary>

```dart

import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      home: Scaffold(
        body: Center(
          child: TextField(
          ),
        ),
      ),
    );
  }
}
```

</details>

### Screenshots or Video

<details>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details><summary>Logs</summary>

```console
flutter: ══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
flutter: The following assertion was thrown during a platform message callback:
flutter: A KeyUpEvent is dispatched, but the state shows that the physical key is not pressed. If this occurs
flutter: in real application, please report this bug to Flutter. If this occurs in unit tests, please ensure
flutter: that simulated events follow Flutter's event model as documented in `HardwareKeyboard`. This was the
flutter: event: KeyUpEvent#17c68(physicalKey: PhysicalKeyboardKey#a8a55(usbHidUsage: "0x00070014", debugName:
flutter: "Key Q"), logicalKey: LogicalKeyboardKey#12bb1(keyId: "0x00000061", keyLabel: "A", debugName: "Key
flutter: A"), character: null, timeStamp: 5:07:41.188000)
flutter: 'package:flutter/src/services/hardware_keyboard.dart':
flutter: Failed assertion: line 471 pos 16: '_pressedKeys.containsKey(event.physicalKey)'
flutter:
[...]

PR #181894, opened 2026-02-04
Disable hardware keyboard regularity warning by default

dkwingsmt:
Suppresses https://github.com/flutter/flutter/issues/125975

This PR disable hardware keyboard regularity warning unless `debugPrintKeyboardEvents` is set.

Keyboard state synchronization has proven extremely difficult, largely due to the asynchronous communication between the engine and framework. Unfortunately, I haven't had the bandwidth to fix it, causing developers to experience false errors during hot restarts. It's time to acknowledge this and prioritize development experience over idealism by disabling these useless warnings.

I apologize for the issues I've caused over the years. Perhaps someday we can make this work with a fully regular event stream.

Replaces https://github.com/flutter/flutter/pull/181719 .

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/flutter/lib/src/services/hardware_keyboard.dart
```diff
@@ -504,36 +504,62 @@ class HardwareKeyboard {
...
+            () =>
+                'Received ${event.runtimeType} for key with mismatched logical key:\n'
+                'Event: $event\n'
+                'Pressed logical key: ${_pressedKeys[event.physicalKey]}',
+          );
+        }
+        return true;
+      }());
+      return shouldProcess;
+    } else {
```
loic-sharma: This branch is still unexpected, no? Should we keep the assertion?

My understanding is if this is reachable, it indicates that we introduced a new `KeyEvent` but forgot to update this code path.

packages/flutter/lib/src/services/hardware_keyboard.dart
```diff
@@ -504,36 +504,62 @@ class HardwareKeyboard {
...
+                'Event: $event\n'
+                'Pressed logical key: ${_pressedKeys[event.physicalKey]}',
+          );
+        }
+        return true;
+      }());
+      return shouldProcess;
+    } else if (event is KeyRepeatEvent || event is KeyUpEvent) {
+      // Only checks physical key here and don't reject based on logical key,
+      // since otherwise the key will never be released.
```
loic-sharma: When I first saw this "don't reject based on logical key" comment, I incorrectly thought this meant "don't check if the map `_pressedKeys` has a key entry for the logical key". I didn't realize what this instead meant was "don't check if the map's _value_ matches the logical key".

I'd consider moving the second half of this comment down to line 544. Alternatively, consider adding more context to this command. Maybe something like:

```suggestion
      // Reject repeat or up events if the physical isn't known to be pressed.
      // However, do process these events even if the logical key is mismatched -
      // this ensures the key is released or updated to the latest logical key value.
```

packages/flutter/lib/src/services/hardware_keyboard.dart
```diff
@@ -504,36 +504,62 @@ class HardwareKeyboard {
...
+      final bool shouldProcess = _pressedKeys.containsKey(event.physicalKey);
```
loic-sharma: Should we process a `KeyRepeatEvent` if it's not in the pressed keys? I'd imagine that means we missed a `KeyDownEvent`?

packages/flutter/lib/src/services/hardware_keyboard.dart
```diff
@@ -504,36 +504,62 @@ class HardwareKeyboard {
...
+  bool _eventShouldProceed(KeyEvent event) {
+    if (event is KeyDownEvent) {
+      final bool shouldProcess = !_pressedKeys.containsKey(event.physicalKey);
+      assert(() {
+        if (!shouldProcess) {
+          _keyboardDebug(
+            () =>
+                'Received ${event.runtimeType} for key that is already pressed:\n'
+                'Event: $event\n'
+                'Pressed logical key: ${_pressedKeys[event.physicalKey]}',
```
loic-sharma: The previous error message had suggested reporting the bug to Flutter, or if it's a simulated event, making sure it follows Flutter's event model as documented by `HardwareKeyboard`. Should we keep that helpful message? It makes the debug log a bit more actionable even if app developers ignore it.

packages/flutter/lib/src/services/hardware_keyboard.dart
```diff
@@ -504,36 +504,62 @@ class HardwareKeyboard {
...
+  // This function will also print debug messages if the event is inconsistent
+  // with the current state.
+  bool _eventShouldProceed(KeyEvent event) {
+    if (event is KeyDownEvent) {
+      final bool shouldProcess = !_pressedKeys.containsKey(event.physicalKey);
+      assert(() {
+        if (!shouldProcess) {
+          _keyboardDebug(
+            () =>
+                'Received ${event.runtimeType} for key that is already pressed:\n'
```
loic-sharma: It might not be clear to the reader that this unexpected and indicates a bug. Consider:

```suggestion
                'Received unexpected ${event.runtimeType} for key that is already pressed:\n'
```

loic-sharma: Same comment for the other messages below :)

packages/flutter/lib/src/services/hardware_keyboard.dart
loic-sharma: If this is indeed correct, I'd consider adding a quick comment explaining why as that's not immediately obvious.

loic-sharma (approved):
LGTM, with one question on why `KeyRepeatEvent`s aren't processed if the key isn't known to be pressed.

packages/flutter/lib/src/services/hardware_keyboard.dart
dkwingsmt: Good idea. Fixed!

packages/flutter/lib/src/services/hardware_keyboard.dart
dkwingsmt: Good idea. Fixed!

packages/flutter/lib/src/services/hardware_keyboard.dart
dkwingsmt: I removed the messages because the messages are only displayed if explicitly enabled by developers that debug keyboard, who should know well enough what these mean. I've added some back for guidance.

packages/flutter/lib/src/services/hardware_keyboard.dart
dkwingsmt: That's an awesome insight. Thank you so much for spotting that. We should indeed process `KeyRepeatEvent` whose key is not pressed.

dkwingsmt:
I've changed the design, and now all key events are processed with their effect applied to the key pressing state. I find it hard to draw a line between irregular events to be processed and those not. Now this PR simply hides the warnings behind a flag.

Issue #181895, opened 2026-02-04
MenuAnchor - SubmenuButton submenu pops at wrong position


westito:
### Steps to reproduce

1. Create a MenuAnchor at bottom position with many items that does not fit downward. 
2. Or create a MenuAnchor with `topStart` alignment SubmenuButton that should pops up left of menu
3. Or create a MenuAnchor with `topEnd` alignment positioned at very end of screen (submenu will not fit and should pops up to the left of menu)

### Expected results

- Submenu first item in align with anchor item
- Submenu pops to the left of parent menu, not covers it

### Actual results

- Submenu first item not in align with anchor item (offsets bottom of subitem to the top of parent item)
- Submenu pops over to parent menu

### Code sample

<details open><summary>Code sample</summary>

```dart
// Minimal sample demonstrating MenuAnchor submenu positioning issues
//
// Issue 1: Submenu with topStart alignment overlaps the anchor instead of opening to the left
// Issue 2: Submenu that flips vertically (near bottom of screen) doesn't align bottom to anchor bottom
//
// To run: flutter run -t menu_anchor_issue_sample.dart

import 'package:flutter/material.dart';

void main() => runApp(const MyApp());

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'MenuAnchor Submenu Position Bug',
      theme: ThemeData.dark(),
      home: const MenuPositionDemo(),
    );
  }
}

class MenuPositionDemo extends StatelessWidget {
  const MenuPositionDemo({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(title: const Text('Submenu Position Issues')),
      body: const Column(
        children: [
          SizedBox(height: 20),
          // Issue 1: topStart alignment - submenu should open LEFT, but overlaps anchor
          Padding(
            padding: EdgeInsets.all(16),
            child: Row(
              children: [
                Text('Issue 1 (topStart): '),
                SizedBox(width: 20),
                _TopStartSubmenuDemo(),
              ],
            ),
          ),
          Spacer(),
          // Issue 2: Vertical flip - submenu bottom should align with anchor bottom
          Padding(
            padding: EdgeInsets.only(bottom: 100, left: 16, right: 16),
            child: Row(
              children: [
                Text('Issue 2 (flip): '),
                SizedBox(width: 20),
                _FlipSubmenuDemo(),
              ],
            ),
[...]

PR #181908, opened 2026-02-04
Fix SubmenuButton submenu positioning

westito:
## Description

Fixes `SubmenuButton` submenu positioning to account for parent menu's padding, flip and alignment.

### Problem

1. Submenus ignoring the parent menu's padding
2. Anchor point not flipped vertically with direction flip. 
3. Submenu anchor point not flipped horizontally, so submenu covered parent. 

### Solution

1. **Expand anchorRect to include parent padding** - For submenus, the `anchorRect` is now expanded to include the parent menu's horizontal padding, representing the visual edge of the parent menu rather than just the button bounds.
2. **Fix vertical flip alignment** - When a submenu flips upward, it now aligns its bottom edge to the anchor's bottom edge (accounting for padding), providing consistent visual alignment.
3. **Fix horizontal flip alignment** - When a submenu flips to the opposite side, it now anchors from its trailing edge to the parent's leading edge.
+1. ScrollView padding fixed, now scrollbar is at the edge of submenu even with added padding.

### Deprecation
<s>**Optional - I revert if deprecation not acceptable **
MenuStyle.alignment currently accepts four values (topStart, bottomStart, topEnd, bottomEnd), but the vertical component is effectively meaningless — a bottomEnd submenu that flips upward is no longer "bottom", making the name misleading. The horizontal component (start/end) is also questionable: opening a submenu toward the opposite side of its arrow indicator is counterintuitive, and the menu should simply open in the direction where space is available.
**Proposal:** Deprecate MenuStyle.alignment in favor of a simpler start/end parameter (or remove it entirely), letting the menu system automatically determine optimal placement based on available space and anchor position.</s>

## Related Issues
https://github.com/flutter/flutter/issues/181895

## Tests

Added comprehensive tests for:
- Submenu positioning with various alignments (topStart, topEnd)
- RTL support
- Vertical flip alignment
- Asymmetric parent/submenu padding combinations
- Deeply nested submenus
- Zero padding edge case
- alignmentOffset interaction with padding

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I added new tests to check the change I am making.

### DIFF
--- a/packages/flutter/lib/src/material/menu_anchor.dart
+++ b/packages/flutter/lib/src/material/menu_anchor.dart
@@ -3440,15 +3440,15 @@ class _MenuLayout extends SingleChildLayoutDelegate {
       y = desiredPosition.dy;
       // For submenus: shift left when RTL XOR *Start alignment.
       // RTL+End → left, RTL+Start → right, LTR+End → right, LTR+Start → left.
-      final bool isRtl = textDirection == TextDirection.rtl;
+      final isRtl = textDirection == TextDirection.rtl;
       final bool isStartAligned =
           parentOrientation == Axis.vertical &&
           switch (alignment) {
-            AlignmentDirectional a => a.start < 0,
-            Alignment a => a.x < 0,
+            final AlignmentDirectional a => a.start < 0,
+            final Alignment a => a.x < 0,
             _ => false,
           };
-      final bool shiftLeft = isRtl != isStartAligned;
+      final shiftLeft = isRtl != isStartAligned;
       if (shiftLeft) {
         x -= childSize.width;
       }
@@ -3849,7 +3849,7 @@ class _Submenu extends StatelessWidget {
     // outside the parent menu's visual bounds, not just outside the button.
     final Rect anchorRect;
     if (layerLink == null) {
-      Rect baseRect = Rect.fromLTRB(
+      var baseRect = Rect.fromLTRB(
         menuPosition.anchorRect.left + dx,
         menuPosition.anchorRect.top,
         menuPosition.anchorRect.right,
--- a/packages/flutter/lib/src/material/menu_anchor.dart
+++ b/packages/flutter/lib/src/material/menu_anchor.dart
@@ -3356,6 +3356,8 @@ class _MenuLayout extends SingleChildLayoutDelegate {
   });
 
   // Rectangle of underlying button, relative to the overlay's dimensions.
+  // For submenus, this is expanded to include the parent menu's padding,
+  // representing the visual edge of the parent menu rather than just the button.
   final Rect anchorRect;
 
   // Whether to prefer going to the left or to the right.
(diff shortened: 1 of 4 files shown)


packages/flutter/lib/src/material/menu_anchor.dart
```diff
@@ -3685,26 +3697,24 @@ class _MenuPanelState extends State<_MenuPanel> {
...
-              controller: scrollController,
-              scrollDirection: widget.orientation,
-              child: Flex(
-                crossAxisAlignment: CrossAxisAlignment.start,
-                textDirection: Directionality.of(context),
-                direction: widget.orientation,
-                mainAxisSize: MainAxisSize.min,
-                children: children,
-              ),
+    Widget menuPanel = ScrollConfiguration(
```
QuncCccccc, line 3700: Why are we removing the Padding here? I think this caused some failures in the google testing and seem not related to the issue reported.

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -4013,6 +4013,957 @@ void main() {
       );
     });
 
+    testWidgets('submenu with topStart alignment opens to the left of anchor', (
+      WidgetTester tester,
+    ) async {
+      await changeSurfaceSize(tester, const Size(800, 600));
+      await tester.pumpWidget(
+        MaterialApp(
+          theme: ThemeData(useMaterial3: false),
```
QuncCccccc, line 4022: Here and everywhere else.
```suggestion
```

MenuAnchor is a new widget in M3, so there are no difference when useMaterial3 is on or off.

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -4013,6 +4013,957 @@ void main() {
       );
     });
 
+    testWidgets('submenu with topStart alignment opens to the left of anchor', (
+      WidgetTester tester,
+    ) async {
+      await changeSurfaceSize(tester, const Size(800, 600));
```
QuncCccccc, line 4019: Why do we need to set a screen size for the tests? I think we can remove this line if it's not useful. Here and everywhere else.

packages/flutter/test/material/menu_anchor_test.dart
```diff
@@ -4013,6 +4013,957 @@ void main() {
...
+        reason: 'Flipped submenu should be positioned above the anchor',
+      );
+    });
+
+    testWidgets('submenu with topStart alignment opens to the right in RTL', (
+      WidgetTester tester,
+    ) async {
+      await changeSurfaceSize(tester, const Size(800, 600));
+      await tester.pumpWidget(
+        MaterialApp(
```
QuncCccccc, line 4240: It seems several tests have duplicated code with only alignment differences. Can we consider creating a method to deduplicate the test code?

QuncCccccc (commented):
Thanks for your contribution and providing a thorough test set. The code change in menu_anchor.dart look good to me with just one comment.

packages/flutter/lib/src/material/menu_anchor.dart
davidhicks980: Adding to @QuncCccccc's comment, the padding is added inside the SingleChildScrollView, which means it is cut off if the menu overflows.

This menu: 

Image

Old menu: 

Image

Note the padding at the bottom of the old menu.

dkwingsmt:

Issue #181684, opened 2026-01-29
Migrate the Flutter engine to Swift 6


jmagman:
Compile the engine with Swift 6, fix all concurrency and other issues.

Note https://github.com/flutter/flutter/issues/160969 is about making the public headers Swift 6 friendly so they can be called safely and idiomatically from Swift 6 code. This is about actually compiling the engine with Swift 6.

cbracken:
Self-assigning (for the bits I'm working on) so I don't have to bookmark this. I know others are working on this too though :)

PR #167530, opened 2025-04-22
Support Swift in the iOS embedder

cbracken:
Adds a `swift` tool to the gn mac toolchain and wires up support for Swift `source_set` targets. Ports `FlutterUIPressProxy` to Swift.

`gn` toolchain modifications include:
* Removes `//flutter/third_party/libcxx/include` and `//flutter/third_party/libcxxabi/include` from the default `include_dirs` on all `source_set` targets, and instead adds them via `cflags` and `cflags_cc` so they continue to be picked up by C, C++, Obj-C, Obj-C++ targets. Flutter's custom libc++ library headers should be picked up by C, C++, Objective-C, and Objective-C++ targets, but not Swift targets. Swift code does not use libc++ directly, but many Apple core libraries such as Foundation.framework are implemented in Objective-C/Objective-C++ and directly or indirectly rely on Apple's libc++ implementation in System.framework. Because of this, and because Swift compiles use Clang modules, we should ensure the correct (Apple) libc++ is picked up from Swift targets by ensuring our own libc++ is not.
* Adds a default empty list for `swiftflags` to `source_set` targets. This list is used by for Swift `source_set` targets by the `swift` tool in the gn toolchain.
* Adds a `//build/config/ios:compiler` config to `engine/src/build/config/ios/BUILD.gn`. This sets default `swiftflags` for iOS targets. This attribute is (only) used by the new `swift` tool in the toolchain.
* Adds `-g` to `swiftflags` for debug (unopt) builds in `engine/src/build/config/compiler/BUILD.gn`.
* Adds a `swift` tool to the gn toolchain that invokes `swiftc.py` to perform compilation and generate a bridging header. This tool is invoked automatically by gn for Swift `source_set` targets.
* Adds `//build/toolchain/apple/swiftc.py`, a wrapper script around `swiftc` in Xcode's Swift toolchain. This script ensures builds are invoked hermetically using a new build cache directory for the emitted swiftmodule, precompiled header, and bridging header, for each compile. The outputs are then copied into the gn output directory. Also adds supporting script `//build/toolchain/apple/get_tool_mtime.py` and data `//build/toolchain/apple/swift_const_gather_protocols.json`, both used by `swiftc.py`.
* Adds a `cxx_module` tool to the gn toolchain that supports clang module-based compiles. This is used for libc++ transitively via System.framework and others that rely on it.

`gn` variable modifications include:
[...]

PR #181921, opened 2026-02-04
Migrate json decoder in FlutterCodecs and the tests to swift

LongCatIsLooong:
- Move the implementation to swift to prevent a crash when the framework sends malformed UTF16 #179727
- Simplifies the "simple" jSON value handling since `JSONSerialization` can already handle that.
- Migrate the relevant tests to swift. ~This is mostly for strict concurrency migration (#181684) because we won't be able to test isolation contracts in objective-c (although it is unlikely that the codecs stuff would need extra isolation keywords since they should be thread safe but it might need to be marked as sendable).~

`JSONSerialization.jsonObject` sanitizes the input string data in swift but not in objective-c because this initializer does string sanitization in swift but not in objective-c: developer.apple.com/documentation/foundation/nsstring/init(bytes:length:encoding:) 

This might have to do with the `NSString` and `String` bridging. This may imply that the swift implementation can be slower because it has to sanitize the string before sending it to the JSON parser.

Closes https://github.com/flutter/flutter/pull/179978

## Pre-launch Checklist

- [ ] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
```diff
@@ -0,0 +1,109 @@
...
+    )
+    let decoded = codec.decode(rawJSON.data(using: .utf16)) as? String
+    #expect("\u{FFFD}" == decoded)
+  }
+
+  @Test func decodeZeroLength() {
+    #expect(codec.decode(Data()) == nil)
+  }
+
+  @available(macOS 13.0, *)
```
LongCatIsLooong: This (and same for the test below this one) is unfortunate because https://developer.apple.com/documentation/swift/collection/contains(_:) is only available  on macOS 13.0+.

LongCatIsLooong: I'm using the `@available` (instead of re-implementing) under the assumption that we have macOS 13+ bots. Reimplementing `contains` would be annoying to maintain and the implementation has to be copied to the test below because they are both exit tests and [exit tests can't capture anything from the context](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0008-exit-tests.md).

engine/src/flutter/testing/symbols/verify_exported.dart
```diff
@@ -13,9 +13,14 @@ import 'package:path/path.dart' as p;
 // Android binaries (libflutter.so) should only export one symbol "JNI_OnLoad"
 // of type "T".
 //
-// iOS binaries (Flutter.framework/Flutter) should only export Objective-C
-// Symbols from the Flutter namespace. These are either of type
-// "(__DATA,__common)" or "(__DATA,__objc_data)".
+// Ideally, iOS binaries (Flutter.framework/Flutter) should only export
+// Objective-C Symbols from the Flutter namespace, of type "(__DATA,__common)" or
+// "(__DATA,__objc_data)". However, to allow Swift symbols to be exported in
```
LongCatIsLooong: copied from the PR description of https://github.com/flutter/flutter/pull/167530

LongCatIsLooong:
After looking at the internal golden changes, I don't think any of the google testing failures are relevant.

hellohuanlin:
Just wanna double check if this is ready for review? The checkboxes in PR description aren't filled up, and it's unclear to me in the description whether this fixes https://github.com/flutter/flutter/issues/179727 - If it does, you can put "fixes https://github.com/flutter/flutter/issues/179727" so that it auto-closes the issue.

engine/src/flutter/shell/platform/darwin/common/framework/Source/flutter_codecs_unittest.mm
```diff
@@ -8,41 +8,8 @@
...
-
-TEST(FlutterStringCodec, CanEncodeAndDecodeNonBMPString) {
-  NSString* value = @"hello \U0001F602 world";
-  FlutterStringCodec* codec = [FlutterStringCodec sharedInstance];
-  NSData* encoded = [codec encode:value];
-  NSString* decoded = [codec decode:encoded];
-  ASSERT_TRUE([value isEqualTo:decoded]);
-}
+// Consider adding new tests to FlutterCodecsTests.swift instead.
+// These legacy tests are kept until Swift 6.2 becomes available on CI.
```
hellohuanlin, line 12: why are some tests in this files moved to swift but not others?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecs.mm
```diff
@@ -87,29 +88,10 @@ - (id)decode:(NSData*)message {
...
-  id decoded = nil;
   NSError* error;
-  if (0 < message.length) {
-    UInt8 first;
-    [message getBytes:&first length:1];
-    isSimpleValue = first != '{' && first != '[';
-    if (isSimpleValue) {
-      // NSJSONSerialization does not support top-level simple values.
-      // We expand encoding to singleton array, then decode that and extract
-      // the single entry.
```
hellohuanlin, line 100: I assume this is done by `fragmentsAllowed`? why we didn't use this option before?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
```diff
@@ -0,0 +1,115 @@
...
+
+  @Test func decodeZeroLength() {
+    #expect(codec.decode(Data()) == nil)
+  }
+
+  // Exit tests are only available on Swift 6.2 and later.
+  // These two tests currently do not run on CI.
+  #if compiler(>=6.2)
+    @available(macOS 13.0, *)
+    @Test
```
hellohuanlin: nit: remove line break to be consistent with your other tests

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
```diff
@@ -0,0 +1,115 @@
...
+    func encodingAssertsOnInvalidInput() async {
+      let result = await #expect(
+        processExitsWith: .signal(SIGABRT),
+        observing: [\.standardErrorContent]
+      ) {
+        let malformedString = NSString(
+          data: Data([0xDF, 0xFF]),
+          encoding: NSUTF16StringEncoding
+        )!
+        FlutterJSONMessageCodec.sharedInstance().encode(malformedString)
```
hellohuanlin, line 43: does the crash happen on this line?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
```diff
@@ -0,0 +1,115 @@
...
+    let decoded = codec.decode(rawJSON.data(using: .utf16)) as? String
+    #expect("\u{FFFD}" == decoded)
+  }
+
+  @Test func decodeZeroLength() {
+    #expect(codec.decode(Data()) == nil)
+  }
+
+  // Exit tests are only available on Swift 6.2 and later.
+  // These two tests currently do not run on CI.
```
hellohuanlin, line 31: is there a timeline on supporting Swift 6.2 on CI? @vashworth

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
```diff
@@ -0,0 +1,19 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
hellohuanlin, line 1: file naming convention: `ClassName+ExtensionName.swift`

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
```diff
@@ -0,0 +1,19 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+
+// The extension is implemented in Swift because `JSONSerialization.jsonObject`
+// converts `data` to a `String` using [NSString initWithBytes:length:encoding:],
+// before passing it to the JSON parser. In Swift the initializer replaces
+// unpaired surrogates in the String with `U+FFFD`, but this doesn't happen in
+// Objective-C.
```
hellohuanlin, line 11: Can you link to the issue here? Also this behavioral difference sounds like a bug to me. May worth filing a radar?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONDecoding.swift
```diff
@@ -0,0 +1,19 @@
...
+// found in the LICENSE file.
+
+import Foundation
+
+// The extension is implemented in Swift because `JSONSerialization.jsonObject`
+// converts `data` to a `String` using [NSString initWithBytes:length:encoding:],
+// before passing it to the JSON parser. In Swift the initializer replaces
+// unpaired surrogates in the String with `U+FFFD`, but this doesn't happen in
+// Objective-C.
+@objc extension JSONSerialization {
```
hellohuanlin: do you still need objc here? (the func is also marked as objc)

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONDecoding.swift
```diff
@@ -0,0 +1,19 @@
...
+  @objc public class func decodeJSON(_ data: Data) throws -> Any {
```
hellohuanlin: nit: encode/decode typically refers to `Codable` API in modern swift (post-JSONSerialization-era). Try "serialization/deserialization" instead.

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
```diff
@@ -0,0 +1,19 @@
...
+@objc extension JSONSerialization {
+  @objc public class func decodeJSON(_ data: Data) throws -> Any {
+    return try JSONSerialization.jsonObject(
+      with: data,
+      options: [.fragmentsAllowed]
```
hellohuanlin, line 18: This is hidden logic that could very well shoot us in the foot in the future. Consider incorporating this info in your function name (e.g. func jsonObjectAllowingFragments).

engine/src/flutter/testing/symbols/verify_exported.dart
```diff
@@ -109,17 +114,7 @@ int _checkIos(String outPath, String nmPath, Iterable<String> builds) {
           (entry.type == '(__DATA,__objc_data)' || entry.type == '(__DATA,__data)') &&
           (entry.name.startsWith(r'_OBJC_METACLASS_$_Flutter') ||
               entry.name.startsWith(r'_OBJC_CLASS_$_Flutter'));
-      // Swift's name mangling uses s followed by symbol length followed by symbol.
-      final swiftInternalRegExp = RegExp(r'^_\$s\d+InternalFlutterSwift');
```
hellohuanlin, line 120: I assume this regex doesn't support extension, and that's why you manually wrote all the matching logic? I wonder if you have tried writing another regex for extension case?

engine/src/flutter/testing/symbols/verify_exported.dart
```diff
@@ -228,6 +223,74 @@ class NmEntry {
...
+            (null, _) => false,
+            (final String moduleName, final String tail) =>
+              allowedModules.contains(moduleName) && tail.startsWith('E'),
+          };
+    }
+    // Simple case: the mangled name starts with the correct module name.
+    // exmaple: _$s26InternalFlutterSwiftCommon8LogLevelOSYAAMc
+    final (String? moduleName, _) = _parseLeadingSwiftMangledIdentifier(rest);
+    return moduleName != null && allowedModules.contains(moduleName);
+  }
```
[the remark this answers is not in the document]
hellohuanlin: is this resolved?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
hellohuanlin: With this extension, we end up with 2 very similarly named APIs, one with [.fragmentsAllowed] option, and the other with [] option:

```
JSONSerialization.jsonObject(with: data)
JSONSerialization.decodeJSON(data)
```

So this can be quite confusing and open to misuse

engine/src/flutter/testing/symbols/verify_exported.dart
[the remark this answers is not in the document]
LongCatIsLooong: Yeah there was a typo, I missed a `!` and gemini caught it.

engine/src/flutter/shell/platform/darwin/common/framework/Source/flutter_codecs_unittest.mm
LongCatIsLooong: I added re-wrote these two tests in swift but found out exit tests are only supported in Swift 6.2 (so the swift versions currently do not run on CI yet).

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecs.mm
LongCatIsLooong: Yeah it appears so and the tests are passing. I assume it was an oversight since the reason for not using `.fragmentsAllowed` wasn't documented.

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
LongCatIsLooong: I think @okorohelijah is working on this and if I recall it's one PR away?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterCodecsTest.swift
LongCatIsLooong: Yeah, 

```
let malformedString = NSString(
  data: Data([0xDF, 0xFF]),
  encoding: NSUTF16StringEncoding
)!
```
evaluates fine in a swift playground. The test does check stderr for error message and verifies if it matches the message we set in the implementation ("failed to convert to UTF8").

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
LongCatIsLooong: Updated the comment.

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONDecoding.swift
LongCatIsLooong: changed the extension from `JSONSerialization` to `FlutterJSONCodec`

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
LongCatIsLooong: changed the extension from `JSONSerialization` to `FlutterJSONCodec`, so from the context it should be clear that this is specialized for FlutterJSONCodec.

engine/src/flutter/testing/symbols/verify_exported.dart
LongCatIsLooong: Yeah regular expressions would work, but there can be false negatives for extensions, e.g., `_$sSo19NSJSONSerializationC26InternalFlutterSwiftCommonE10decodeJSONyyp10Foundation4DataVKFZ` we don't know the external class name and the class name may have digits in it.

LongCatIsLooong:
@hellohuanlin could you take a look when you get a chance

engine/src/flutter/shell/platform/darwin/common/framework/Source/flutter_codecs_unittest.mm
hellohuanlin: > exit tests are only supported in Swift 6.2

Can you update the comment to include this info?

engine/src/flutter/shell/platform/darwin/common/framework/Source/FlutterJSONMessageCodec+SafeDecode.swift
```diff
@@ -0,0 +1,21 @@
...
+
+import Foundation
+
+// The extension is implemented in Swift because `JSONSerialization.jsonObject`
+// converts `data` to a `String` using [NSString initWithBytes:length:encoding:],
+// before passing it to the JSON parser. In Swift the initializer replaces
+// unpaired surrogates in the String with `U+FFFD`, but this doesn't happen in
+// Objective-C.
+//
+// Seealso: https://github.com/flutter/flutter/issues/179727.
```
hellohuanlin, line 13: space

engine/src/flutter/testing/symbols/verify_exported.dart
```diff
@@ -13,9 +13,14 @@ import 'package:path/path.dart' as p;
 // Android binaries (libflutter.so) should only export one symbol "JNI_OnLoad"
 // of type "T".
 //
-// iOS binaries (Flutter.framework/Flutter) should only export Objective-C
-// Symbols from the Flutter namespace. These are either of type
-// "(__DATA,__common)" or "(__DATA,__objc_data)".
```
hellohuanlin, line 18: I believe @cbracken's recent change supported swift extensions so this can be reverted?

LongCatIsLooong:
Moved the `verify_export` part to [a different PR](https://github.com/flutter/flutter/pull/186835). Since our bots are all updated I'm going to remove the compiler version check in the tests.

Piinks:

LongCatIsLooong:
Potentially superseded by https://github.com/flutter/flutter/pull/190329.

PR #181933, opened 2026-02-05
Adds opengles to engine dart tests

### LINKED ISSUE #180601: Run flutter_test tests against Metal and OpenGLES backends
Currently, tests like fragment_shader_test.dart will [only execute against the vulkan backend](https://github.com/flutter/flutter/blob/1e87bd8e8f0c625a3f02953923d3d0b614061766/engine/src/flutter/shell/testing/tester_main.cc#L200).  `flutter_tester` lacks the ability to run against any other backend which is important when the test incorporates things like impellerc which behaves differently for each backend.

notes:
- while flutter_tester technically runs with the vulkan backend, it doesn't actually run the tests in question  https://github.com/flutter/flutter/blob/1e87bd8e8f0c625a3f02953923d3d0b614061766/engine/src/flutter/testing/dart/fragment_shader_test.dart#L46  https://github.com/flutter/flutter/issues/180595 will fix that

b-luk:
Forked from https://github.com/flutter/flutter/pull/180704. See comments there for more context.

Enables most of the dart tests for opengles. There are some which aren't passing, which are filtered out in run_tests.py.

Fixes an issue in snapshot_controller_impeller.cc where a returned GLContextResult is ignored. The returned GLContextResult is potentially an RAII object, so it must stay in scope for subsequent GL calls.

Fixes https://github.com/flutter/flutter/issues/180601

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/testing/run_tests.py
```diff
@@ -926,43 +931,53 @@ def gather_dart_tests(
...
   )
 
-  dart_vm_service_tests = glob.glob(f'{dart_tests_dir}/vm_service/*_test.dart')
+  dart_vm_service_tests = [] if 'release' in build_dir else glob.glob(
+      f'{dart_tests_dir}/vm_service/*_test.dart'
+  )
   dart_tests = glob.glob(f'{dart_tests_dir}/*_test.dart')
 
-  impeller_backends = ['', 'vulkan']
+  opengles_skipped_tests = [
```
b-luk, line 939: From @gaaclarke in https://github.com/flutter/flutter/pull/180704:

> Are these every test that actually uses opengles though? I'm seeing other tests that pass are things like platform channels tests, which never touch graphics. What are some of the failures we have here?

I did some digging into what was causing these tests to fail.

I think I tracked down the majority of the failures to one bug. In snapshot_controller_impeller.cc's DoMakeRasterSnapshot() function, we call MakeRenderContextCurrent(), which returns a GLContextResult. The returned GLContextResult is a RAII object which obtains the GL context when it is created and releases it when it goes out of scope. In DoMakeRasterSnapshot(), MakeRenderContextCurrent() is called but the return value is unused. So the GL context is released immediately, and the subsequent GL commands would fail. This is why I found many of these test failures to only happen in non-optimized builds. The nonopt builds has looser memory management, so the GL context would stay current for the subsequent GL commands.

I fixed this issue in snapshot_controller_impeller.cc by keeping the return value in scope for the duration of the rest of the function. This makes the following tests now pass on opengles:
```
assets_test.dart
canvas_test.dart
color_filter_test.dart
compositing_test.dart
image_events_test.dart
image_filter_test.dart
image_shader_test.dart
image_test.dart
mask_filter_test.dart
painting_test.dart
text_test.dart
tracing_test.dart
```

There is a handful of tests which still fail on opengles, which remain filtered out here.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
```diff
@@ -0,0 +1,304 @@
...
+ private:
+  std::unique_ptr<TesterGLESDelegate> delegate_;
+  std::shared_ptr<TesterGLESWorker> worker_;
+  std::shared_ptr<impeller::ContextGLES> context_;
+};
+
+}  // namespace
+
+std::unique_ptr<TesterContext> TesterContextGLESFactory::Create() {
+  flutter::testing::SetupSwiftshaderOnce(true);
```
b-luk, line 300: From @gaaclarke in https://github.com/flutter/flutter/pull/180704:

> This function exists already in impeller/playground/backend/vulkan/swiftshader_utilities.cc, can we share it?

Done. I moved swiftshader_utilities to a sharable location in flutter/testing. I renamed it test_swiftshader_utils to be more consistent with other files in that directory (e.g. test_gl_utils).

Additionally, the CreateSwangleDisplay/HasExtension functions in this file already exist as functions in engine/src/flutter/testing/test_gl_context.cc. Similar to test_swiftshader_utils, I moved these functions to a shared location (flutter/testing/test_swangle_utils) so they can be reused here.

engine/src/flutter/shell/testing/tester_context_mtl_factory.h
```diff
@@ -10,7 +10,7 @@
 
 // Impeller should only be enabled if the Metal backend is enabled.
 #define TESTER_ENABLE_METAL \
-  (IMPELLER_SUPPORTS_RENDERING && IMPELLER_ENABLE_METAL)
+  (IMPELLER_SUPPORTS_RENDERING && IMPELLER_ENABLE_METAL && SHELL_ENABLE_METAL)
```
b-luk, line 13: From @gaaclarke in https://github.com/flutter/flutter/pull/180704:

> Aren't IMPELLER_ENABLE_METAL and SHELL_ENABLE_METAL synonymous? I'm not sure when you'd have one and not the other.

I don't know the full context of these IMPELLER_ENABLE_ vs SHELL_ENABLE_  defines. But they are defined in different situations. IMPELLER_ENABLE_ is solely based on the platform: e.g. https://github.com/flutter/flutter/blob/70d9d3188044627f4f32cc016f2f49a4c3f9b754/engine/src/flutter/impeller/tools/args.gni#L13 They are the same regardless of whether we're building the flutter_tester executable or the flutter_tester_opengles executable. We want the two executables to behave differently, so using the IMPELLER_ENABLE_ defines won't work for this.

SHELL_ENABLE_METAL is based on the shell_gpu_configuration config: https://github.com/flutter/flutter/blob/70d9d3188044627f4f32cc016f2f49a4c3f9b754/engine/src/flutter/shell/gpu/gpu.gni

We want vulkan and metal and not opengles for flutter_tester. And we want only opengles for flutter_tester_opengles. So engine/src/flutter/shell/testing/BUILD.gn uses two different shell_gpu_configurations for the two executables to specifically enable/disable the desired backends.

engine/src/flutter/shell/testing/BUILD.gn
```diff
@@ -9,65 +9,120 @@ if (is_fuchsia) {
...
+  public_deps = [ ":flutter_tester" ]
+  if (current_toolchain == host_toolchain) {
+    public_deps += [ ":flutter_tester_opengles" ]
+  }
+}
+
 shell_gpu_configuration("tester_gpu_configuration") {
   enable_software = true
-  enable_gl = true
+  enable_gl = false
```
gaaclarke, line 21: This won't affect any current usage of `flutter_tester` right?  I think CI will tell us that.

engine/src/flutter/shell/testing/BUILD.gn
```diff
@@ -9,65 +9,120 @@ if (is_fuchsia) {
...
+      "//flutter/third_party/swiftshader/src/Vulkan:swiftshader_libvulkan_static",
+    ]
   }
 
   metadata = {
     entitlement_file_path = [ "flutter_tester" ]
   }
 }
+
+if (current_toolchain == host_toolchain) {
```
gaaclarke, line 114: What's this for?  I'm not familiar with it.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
```diff
@@ -0,0 +1,304 @@
...
+
+ private:
+  EGLDisplay display_;
+  EGLSurface surface_;
+  EGLContext context_;
+};
+
+class TesterGLESDelegate : public GPUSurfaceGLDelegate {
+ public:
+  TesterGLESDelegate() {
```
gaaclarke: This should be transitioned to a factory function that returns a `absl::StatusOr<TesterGLESDelegate>`

https://google.github.io/styleguide/cppguide.html#Doing_Work_in_Constructors

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
```diff
@@ -0,0 +1,304 @@
...
+        ::eglDestroySurface(display_, surface_);
+      }
+      if (context_ != EGL_NO_CONTEXT) {
+        ::eglDestroyContext(display_, context_);
+      }
+      ::eglTerminate(display_);
+    }
+  }
+
+  bool IsValid() const {
```
gaaclarke: Once you create the factory function, you won't need this anymore.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
```diff
@@ -0,0 +1,304 @@
...
+  bool GLContextClearCurrent() override {
+    return ::eglMakeCurrent(display_, EGL_NO_SURFACE, EGL_NO_SURFACE,
+                            EGL_NO_CONTEXT) == EGL_TRUE;
+  }
+
+  // |GPUSurfaceGLDelegate|
+  bool GLContextPresent(const GLPresentInfo& present_info) override {
+    return true;  // PBuffer doesn't present? Or maybe eglSwapBuffers?
+    // For pbuffer, eglSwapBuffers is usually valid but doesn't do much on
+    // screen. return ::eglSwapBuffers(display_, surface_) == EGL_TRUE;
```
gaaclarke: I think this is some llm detritus?

engine/src/flutter/shell/testing/tester_context_mtl_factory.h
gaaclarke: I think I see.  I guess it's not really a big deal if flutter_tester_opengles has metal if we don't run metal tests with it, right?  Not a big deal, just making sure I understand.

engine/src/flutter/testing/run_tests.py
gaaclarke: Ahh nice work.  It sounds like we should add the `[[nodiscard]]` annotation to MakeRenderContextCurrent().

We might have to change the test's to artificially keep that context in scope and active to mimic the runtime behavior.

gaaclarke: Let's add an issue to get these turned back on.

gaaclarke (commented):
Awesome, this is looking good.  There is a lot of fiddly code for setting up swangle but ultimately if it run on CI that's the ultimate test and that's what its use is limited to.  I just have a few cleanup suggestions.

engine/src/flutter/shell/testing/BUILD.gn
b-luk: I assume not, based on CI. But I don't know the details of exactly what behavior might change.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
```diff
@@ -0,0 +1,304 @@
...
+class TesterContextGLES : public TesterContext {
+ public:
+  TesterContextGLES() = default;
+
+  ~TesterContextGLES() override {
+    if (context_) {
+      std::shared_ptr<impeller::Context> raw_context = context_;
+      raw_context->Shutdown();
+    }
+  }
```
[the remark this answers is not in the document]
b-luk: context_ is a ContextGLES, and Shutdown() is made a private method in ContextGLES. So I can't do `context_->Shutdown()` here and have to call it using the base impeller::Context class.

All the overridden methods of impeller::Context are made private in ContextGLES. Is that intentional? Or should I make them public instead, so I can do `context_->Shutdown()` here?

engine/src/flutter/shell/testing/BUILD.gn
b-luk: This is to make this build target only available when building for the host platform: https://github.com/flutter/flutter/blob/3e015d2dab750fa19e9dbfa858927c7d02f38d2a/engine/src/build/config/BUILDCONFIG.gn#L34-L35

Having libEGL_static and libGLESv2_static as dependencies for any build targets breaks building anything for ios/android/fuchsia platform targets. It results in an error like:
```
❯ et build -c ios_debug_unopt
[2026-02-06 16:46:06.415699][macos/ios_debug_unopt: GN]: FAILED
COMMAND:
/Users/bensonluk/development/flutter/engine/src/flutter/tools/gn --ios --runtime-mode debug --unoptimized --no-stripped --no-lto --xcode-symlinks --rbe --no-goma
STDOUT:
ERROR at //flutter/third_party/angle/gni/angle.gni:460:5: Undefined identifier.
    configs += invoker.configs
    ^------
See //flutter/third_party/angle/BUILD.gn:1407:3: whence it was called.
  angle_shared_library(target_name) {
  ^----------------------------------
See //flutter/third_party/angle/BUILD.gn:1436:1: whence it was called.
angle_libGLESv2("libGLESv2") {
^-----------------------------
See //flutter/testing/BUILD.gn:248:7: which caused the file to be included.
      "//flutter/third_party/angle:libGLESv2_static",
      ^---------------------------------------------
```

So I add this check around anything that uses those angle libraries (`test_swangle_utils` and this `flutter_tester_opengles` target) to prevent breaking builds for the ios/android/fuchsia platform targets. As far as I can tell, the `flutter_tester` and `flutter_tester_opengles` executables are only built for host targets, so I think this should be fine.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
b-luk: This (along with the other odd comment in CreateRenderingSurface() below) was part of your original PR.

As someone who is not very familiar with EGL, I don't fully understand what the comments are about, or if whether they'd be helpful for someone that does know more about EGL than me.

I deleted the comments, but if you think it'd be helpful to have some comments here that make more sense, I'm be open to suggestions.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
b-luk: Done

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
b-luk: Removed

engine/src/flutter/shell/testing/tester_context_mtl_factory.h
b-luk: Right, I don't think it's a big deal to make flutter_tester_opengles not have metal.

There's nothing technically stopping us from making it include metal. But I think that just makes things less confusing. As currently implemented flutter_tester_opengles is specifically only for running with opengles and flutter_tester is for all other backends (metal, vulkan, skia), with no overlap between the two. I think making some backends work with both versions would just make things more confusing.

engine/src/flutter/testing/run_tests.py
b-luk: Acknowledged. I'll create an issue for this.

engine/src/flutter/shell/testing/tester_context_gles_factory.cc
gaaclarke: Sometimes when you ask gemini to make changes it inserts comments that are like it's thinking phase.  Something like "// No, wait but if I do this, how will I do X?".  I think that's what this was here.  I guess maybe if the LLM got confused it's an indicator that comments would be helpful?  Or it's just LLMs being weird and it doesn't mean anything.  I'm fine with none.

gaaclarke (approved):
LGTM!  There is some bad changes in the golden tests.  I'm not sure what those are about.  Maybe try rebasing.

gaaclarke:
@b-luk ahh, okay I missed the backend dimension.  Okay, I'd approve those new goldens and file a bug.  I think there may be a way to discard the old approvals after we fix the bug.

b-luk:
> Sounds good. Assuming the goldens after the merge come back the same way, I'll approve them and get this PR submitted, then file an issue about the one weird test.

The goldens failed in the same way. I think they're all fine except for the weird drawImageNine issue I mentioned.

I approved the changed goldens and filed https://github.com/flutter/flutter/issues/182101

engine/src/flutter/testing/run_tests.py
b-luk: Filed https://github.com/flutter/flutter/issues/182102 and its sub-issues

PR #181956, opened 2026-02-05
Temporarily remove the Pixel 9/API 36 device from the Firebase Test Lab tests.

jason-simmons:
There appears to be a Firebase outage that currently affects this device type.

See https://github.com/flutter/flutter/issues/181954

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -712,6 +712,8 @@ targets:
       tags: >
         ["firebaselab"]
       task_name: release_smoke_test
+      # TODO(flutter/flutter#181954) Restore the Pixel 9/API 36 device when
+      # it is available in Firebase Test Lab.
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -714,7 +714,8 @@ targets:
       task_name: release_smoke_test
       physical_devices: >-
         [
-          "--device", "model=tokay,version=36",
+          # Temporarily disabled due to Firebase Test Lab issues.
```
jtmcdole: I'm pretty sure this is a json string. The comment should be above phsyical_devices.

jtmcdole (approved):

.ci.yaml
jason-simmons: done

PR #181966, opened 2026-02-05
[Impeller] Fix VkImageView leak in WrappedTextureSourceVK

### LINKED ISSUE #181967: [Impeller] VkImageView leak in WrappedTextureSourceVK (embedder delegate path)
## Description

`WrappedTextureSourceVK` in `shell/gpu/gpu_surface_vulkan_impeller.cc` leaks a `VkImageView` every frame when using the embedder delegate code path.

## Steps to reproduce

1. Create a custom embedder using `FlutterVulkanRendererConfig` (the Vulkan delegate path)
2. Run any Flutter app
3. Each frame, `AcquireFrame` calls `createImageView()` and stores the result as a raw `vk::ImageView`
4. The destructor `~WrappedTextureSourceVK() {}` is empty — the `VkImageView` is never destroyed

This leaks ~60 `VkImageView` handles per second. On a long-running app this will eventually exhaust driver resources.

## Analysis

The fix is straightforward: use `vk::UniqueImageView` instead of raw `vk::ImageView` for RAII ownership. The resource lifetime is safe because Impeller's `TrackedObjectsVK` holds a `shared_ptr<TextureSourceVK>` until the GPU fence signals completion, preventing premature destruction.

This only affects the embedder delegate path (`delegate_ != nullptr`). Android passes `nullptr` as the delegate and uses `AcquireNextSurface()` instead, so Android is not affected.

## Environment

- Found on Linux/Wayland with a custom Vulkan embedder
- NVIDIA GeForce RTX 5070 Ti, driver 570.x
- Flutter engine at commit f916dd6887b

## Labels

- `e: impeller`
- `platform-linux`
- `P2`

RickyvdBerg:
## Description

Fixes https://github.com/flutter/flutter/issues/181967

`WrappedTextureSourceVK` in the embedder delegate path creates a `VkImageView` every frame via `createImageView()` but never destroys it. The handle is stored as a raw non-owning `vk::ImageView` and the destructor is empty, leaking one `VkImageView` per frame (~60/sec on a typical app).

This switches to `vk::UniqueImageView` for RAII ownership so the image view is destroyed when the `WrappedTextureSourceVK` is released.

## Why this is safe

The resource lifetime is correct because Impeller's `TrackedObjectsVK` holds a `shared_ptr` to the `TextureSourceVK` until the GPU fence signals completion:

1. `RenderPassVK` calls `Track(attachment.texture)` on all render target textures
2. `CommandBufferVK::Track()` stores `shared_ptr<TextureSourceVK>` in `TrackedObjectsVK`
3. `TrackedObjectsVK` is cleared only in the fence completion callback, after the GPU finishes

The `VkImageView` cannot be destroyed while the GPU is still using it.

## Scope

This only affects the **embedder delegate code path** (when `delegate_ != nullptr` in `GPUSurfaceVulkanImpeller::AcquireFrame`). Android passes `nullptr` as the delegate and uses `AcquireNextSurface()` instead, so this change has **no effect on Android, iOS, or standard swapchain rendering**.

## Pre-launch Checklist

- [ ] I added a test for the change if applicable. *(See note below)*

**Note on testing:** This is a resource leak fix in a Vulkan code path that requires a running Vulkan device and embedder delegate. I was unable to find existing test infrastructure for the embedder delegate path in `GPUSurfaceVulkanImpeller`. Happy to add a test if reviewers can point me to the right test harness.

Found while working on Impeller Vulkan support for Linux/Wayland with a custom embedder.

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md

### DIFF
--- a/engine/src/flutter/shell/gpu/gpu_surface_vulkan_impeller.cc
+++ b/engine/src/flutter/shell/gpu/gpu_surface_vulkan_impeller.cc
@@ -25,25 +25,29 @@ namespace flutter {
 class WrappedTextureSourceVK : public impeller::TextureSourceVK {
  public:
   explicit WrappedTextureSourceVK(impeller::vk::Image image,
-                                  impeller::vk::ImageView image_view,
+                                  impeller::vk::UniqueImageView image_view,
                                   impeller::TextureDescriptor desc)
-      : TextureSourceVK(desc), image_(image), image_view_(image_view) {}
+      : TextureSourceVK(desc),
+        image_(image),
+        image_view_(std::move(image_view)) {}
 
-  ~WrappedTextureSourceVK() {}
+  ~WrappedTextureSourceVK() override = default;
 
  private:
   impeller::vk::Image GetImage() const override { return image_; }
 
-  impeller::vk::ImageView GetImageView() const override { return image_view_; }
+  impeller::vk::ImageView GetImageView() const override {
+    return image_view_.get();
+  }
 
   impeller::vk::ImageView GetRenderTargetView() const override {
-    return image_view_;
+    return image_view_.get();
   }
 
   bool IsSwapchainImage() const override { return true; }
 
   impeller::vk::Image image_;
-  impeller::vk::ImageView image_view_;
+  impeller::vk::UniqueImageView image_view_;
 };
 
 GPUSurfaceVulkanImpeller::GPUSurfaceVulkanImpeller(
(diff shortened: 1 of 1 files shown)


gaaclarke:
@jason-simmons this seems reasonable to me, can you do a pass on it please?

jason-simmons (approved):
Thank you for finding and fixing this!

Ran the https://github.com/flutter/flutter/tree/master/engine/src/flutter/examples/vulkan_glfw sample app with Vulkan validation layers enabled.

Confirmed that the object tracking layer reports leaks of `VkImageView` objects on the current head.  With this PR, the validation layers do not show any errors.

gaaclarke (approved):
thanks!

PR #181978, opened 2026-02-06
Add GitHub workflows to assist with release tasks

eyebrowsoffire:
This is a PR which adds some new GitHub Action workflows which assist in release tasks for the engineer on the release rotation.

## Release Tracker

This workflow creates a GitHub issue which tracks the release process for a particular release. When running the workflow, the engineer provides a version number, the release channel (beta or stable) and whether the release is a hotfix or not. The issue contains a checklist with all the steps needed for a particular release type and channel.

I have been testing this workflow on a separate repository here: https://github.com/eyebrowsoffire/flutter_release_workflow_test

You can see examples of the issues that were generated from the workflow here:

Stable initial release: https://github.com/eyebrowsoffire/flutter_release_workflow_test/issues/9
Stable hotfix release: https://github.com/eyebrowsoffire/flutter_release_workflow_test/issues/10
Beta initial release: https://github.com/eyebrowsoffire/flutter_release_workflow_test/issues/11
Beat hotfix release: https://github.com/eyebrowsoffire/flutter_release_workflow_test/issues/12

## Cut Release Branch

This workflow creates a new `flutter-X.Y-candidate.0` branch with a commit which writes out the `bin/internal/release-candidate-branch.version` file with the proper branch name inside. The user supplies the version number and the git hash that the branch should be forked from.

## Roll Dart Dependencies

This workflow creates a PR which updates the DEPS file with a new dart hash and updates the transitive dependencies with the `create_updated_flutter_deps.py` script. The user runs the workflow on a specific candidate branch and passes the dart hash to update to. Unlike the dart autoroller, this can be manually invoked on any branch. If the dart hash and deps are already up to date, the workflow succeeds and doesn't create a PR.

## Sync Engine Version

This workflow creates a PR that updates the engine.version file. The user runs the workflow on a specific candidate branch, and the correct dart hash is automatically calculated by the workflow. If the engine hash is already up to date, the workflow succeeds and doesn't create a PR.


eyebrowsoffire:
> Is this Dart code? Is there any way for the code to be in a...like...regular file?

No. This uses the standard Github script action which is Node JS by default. We might not be able to actually use that action though, so I might have to rewrite this in some other way. This is a proof of concept.

eyebrowsoffire:
Discussed with @jtmcdole, and apparently the GitHub official actions are allowlisted as part of our actions that we can use. We explicitly added it as one of the actions allowed for this project.

With that being the case, I personally think it's better to actually leave this as inline NodeJS code, for a few reasons:
* Github's standard action runs really fast and doesn't require any extra setup (like downloading the Dart SDK onto whatever host is running the action).
* All of Github's API for reading and manipulating issues and pull requests are simple and available by default.
* The logic here is fairly straightforward, and keeping it encapsulated in a simple single file seems to be the path of least maintenance.

engine/src/tools/dart/create_updated_flutter_deps.py
```diff
@@ -191,8 +195,13 @@ def Main(argv):
   lines = file.readlines()
   i = 0
   while i < len(lines):
-    updatedfile.write(lines[i])
     if lines[i].startswith("  'dart_revision':"):
+      if args.dart_revision is None:
+        # No dart revision supplied. Leave as-is.
+        updatedfile.write(lines[i])
+      else:
+        updatedfile.write("  'dart_revision': '%s',\n" % args.dart_revision)
```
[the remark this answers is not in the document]
eyebrowsoffire: I added a step that validates the format of the argument.

.github/workflows/cut-release-branch.yml
```diff
@@ -0,0 +1,65 @@
+name: Cut Release Branch
+
+on:
+  workflow_dispatch:
```
jtmcdole, line 4: Anyone with WRITE access can trigger workflow dispatch. I suspect this will fail though since we have branch protections set to `flutter-release-eng` for creating branches.

That said, do we want something like:

```
steps:
      - name: Check team membership using API
        uses: actions/github-script@v6
        with:
          script: |
            try {
              const username = context.actor;
              const org = 'flutter';
              const team_slug = 'flutter-release-eng';

              await github.rest.teams.getMembershipForUserInOrg({
                org: org,
                team_slug: team_slug,
                username: username
              });
              console.log(`${username} is a member of the team.`);
              // Set an output or continue the job
            } catch (error) {
              if (error.status === 404) {
                console.log(`${username} is not a member of the flutter-release-eng. Stopping workflow.`);
                process.exit(1); // Exit with failure
              }
              throw error; // Re-throw other errors
            }
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

jtmcdole: now that I think about it - I bet workflows don't inherit from the user, so we might want something like the above.

.github/workflows/cut-release-branch.yml
```diff
@@ -0,0 +1,65 @@
...
+          git config user.email "<>"
+
+      - name: Create Branch and Version File
+        run: |
+          # Create and switch to the new branch locally
+          git checkout -b ${{ env.BRANCH_NAME }}
+
+          # Create the version file
+          mkdir -p bin/internal/
+          echo "${{ env.BRANCH_NAME }}" > bin/internal/release-candidate-branch.version
```
jtmcdole: I don't think this file has a new line in it? Not sure it matters, but `echo -n`

.github/workflows/release-tracker.yml
```diff
@@ -0,0 +1,124 @@
+name: 🚀 Create Release Tracker
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version'
```
jtmcdole: match other workflow by adding " (format X.Y)"

.github/workflows/roll-dart-dependencies.yml
```diff
@@ -0,0 +1,97 @@
+name: Roll Dart Dependencies
+
+on:
+  workflow_dispatch:
```
jtmcdole, line 4: Are we only ever checking out master or would you want a branch here?

jtmcdole (changes requested):

.github/workflows/roll-dart-dependencies.yml
eyebrowsoffire: `workflow_dispatch` automatically takes a branch without specifying it as an input parameter, since it needs to know which version of the workflow to use.

jtmcdole (approved):

PR #182009, opened 2026-02-06
Add alwaysSizeToContent argument to Overlay.

### LINKED ISSUE #182008: Overlay can't size to child when the constraints are loose
Right now when the `Overlay` widget gets passed loose (but not infinite) constraints, it picks `constraints.biggest` as its size. In order for `Overlay` to work for views sized to content, a behavior was added where the `Overlay` can a child to size itself. This however only works, when constraints are infinite.

There is a missing gap here where we need `Overlay` to size itself to content, but in constrained but loose environment. The usecase for this is `Popover` window, where the size of the popover is determined by content, but there are incoming constraints from the desktop (popover can not be larger than usable desktop area)/

knopp:
Fixes https://github.com/flutter/flutter/issues/182008.

This PR introduces `alwaysSizeToContent` argument to `Overlay`. When set to true, this ensures the Overlay
is always sized to content, even if the constraints are finite.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/overlay.dart
+++ b/packages/flutter/lib/src/widgets/overlay.dart
@@ -489,6 +489,7 @@ class Overlay extends StatefulWidget {
     super.key,
     this.initialEntries = const <OverlayEntry>[],
     this.clipBehavior = Clip.hardEdge,
+    this.alwaysSizeToContent = false,
   });
 
   /// Wrap the provided `child` in an [Overlay] to allow other visual elements
@@ -497,8 +498,18 @@ class Overlay extends StatefulWidget {
   /// This is a convenience method over the regular [Overlay] constructor: It
   /// creates an [Overlay] and puts the provided `child` in an [OverlayEntry]
   /// at the bottom of that newly created Overlay.
-  static Widget wrap({Key? key, Clip clipBehavior = Clip.hardEdge, required Widget child}) {
-    return _WrappingOverlay(key: key, clipBehavior: clipBehavior, child: child);
+  static Widget wrap({
+    Key? key,
+    Clip clipBehavior = Clip.hardEdge,
+    bool alwaysSizeToContent = false,
+    required Widget child,
+  }) {
+    return _WrappingOverlay(
+      key: key,
+      clipBehavior: clipBehavior,
+      alwaysSizeToContent: alwaysSizeToContent,
+      child: child,
+    );
   }
 
   /// The entries to include in the overlay initially.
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -521,6 +532,17 @@ class Overlay extends StatefulWidget {
   /// Defaults to [Clip.hardEdge].
   final Clip clipBehavior;
 
+  /// Determines that the overlay should always size itself to content.
+  ///
+  /// Normally overlay will only size itself to content if the the incoming
+  /// constraints are infinite and there is a [OverlayEntry] that can size
+  /// the overlay. Setting this to `true` will force this behavior even for
```
chunhtai, line 539: also mention [OverlayEntry.canSizeOverlay] set to true

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -521,6 +532,17 @@ class Overlay extends StatefulWidget {
...
+  /// finite (but possibly loose) constraints.
+  ///
+  /// Setting this to true requires an [OverlayEntry] that can size the overlay
+  /// based on itself.
+  final bool alwaysSizeToContent;
```
chunhtai, line 545: does it worth to make this a tri state? either enum or `bool?`, where always size by content, always not size by content, and size depends on infinite constraint

packages/flutter/lib/src/widgets/overlay.dart
```diff
@@ -521,6 +532,17 @@ class Overlay extends StatefulWidget {
...
   final Clip clipBehavior;
 
+  /// Determines that the overlay should always size itself to content.
+  ///
+  /// Normally overlay will only size itself to content if the the incoming
+  /// constraints are infinite and there is a [OverlayEntry] that can size
+  /// the overlay. Setting this to `true` will force this behavior even for
+  /// finite (but possibly loose) constraints.
+  ///
+  /// Setting this to true requires an [OverlayEntry] that can size the overlay
```
chunhtai, line 542: should also mention it throws when there isn't a overlay entry that can size the overlay

packages/flutter/lib/src/widgets/overlay.dart
knopp: Not sure, if there is consensus tri-state would be better I'll can rework this.

chunhtai: thinking a bit more, I think this is fine for now.

chunhtai (commented):
besides the doc related comments, otherwise LGTM

mattkae (approved):

Issue #46235, opened 2019-12-06
Focus can be stolen from the current ModalRoute


ds84182:
This involves dialogs or any modal routes that maintain their state. If an event happens and they spawn an auto-focusing FocusNode, the route will steal focus from the rest of the application, sometimes even invisibly.

Example app:
```dart
import 'dart:async';

import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';

void main() {
  // If testing on desktop:
  debugDefaultTargetPlatformOverride = TargetPlatform.fuchsia;
  runApp(MaterialApp(
    home: HomePage(),
  ));
}

class HomePage extends StatefulWidget {
  @override
  _HomePageState createState() => _HomePageState();
}

class _HomePageState extends State<HomePage> {
  Timer _timer;

  void _stealSoon() {
    setState(() {
      _timer = Timer(Duration(seconds: 5), () {
        setState(() {});
      });
    });
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Padding(
        padding: const EdgeInsets.all(24.0),
        child: SizedBox(
          width: double.infinity,
          child: Column(
            children: <Widget>[

              RaisedButton(
                child: Text('Open dialog'),
                onPressed: () {
                  showDialog(
                    context: context,
                    child: AlertDialog(
                      title: Text('Hi!'),
                      actions: <Widget>[
                        FlatButton(
                          autofocus: true,
                          child: Text('OK'),
                          onPressed: () {
                            Navigator.of(context).pop();
                          },
                        ),
                      ],
                    ),
                  );
                },
              ),
              RaisedButton(
                child: Text('Steal focus soon'),
                onPressed: _timer == null ? _stealSoon : null,
              ),
              if (_timer != null && !_timer.isActive)
                TextField(
                  autofocus: true,
                ),
            ],
          ),
        ),
      ),
    );
  }
}
```

Clicking "Steal focus soon" and then "Open dialog" will show this issue. The topmost modal route should be the only thing that can request and maintain focus.

This can happen in an actual application. e.g. a dialog shows up, and the route below goes to a username+password login page because credentials expired. If the username field has autofocus, this will steal focus from the dialog. Worse, this can happen with an opaque route if another opaque route is on top of it, assuming the previous route has maintainState set.

cc @gspencergoog

maheshmnj:
Hi @ds84182, Thanks for filing the issue. I am not sure if this statement is correct   
[...]

PR #182024, opened 2026-02-06
[web] Fix autofill in iOS 26 Safari

### LINKED ISSUE #177248: [web] Autofill does not work on browsers running on iOS 26 and iPadOS 26 devices
### Steps to reproduce

Create a fresh project, and launch the app 

```
flutter create test_app --platform web
cd test_app
flutter run -d web-server --web-hostname 0.0.0.0 --web-port 8080
```

Notes: 
- This issue can be replicated regardless of whether or not `--release` or `--debug` is used
- This issue can only be replicated on iOS 26 devices and iPadOS 26 devices, on other devices such as iOS 18.x.x this seems to work properly
- This issue can also be replicated on iOS 26 Simulator
- This has been replicated using the ff password apps Apple Passwords, Lastpass, and Bitwarden
- This issue can be replicated on Flutter Versions 3.32.8, and 3.35.6 unfortunately
- This issue can be replicated on the ff browsers on iOS 26 Safari, Chrome, Firefox

### Expected results

Autofill should work properly 

### Actual results

Autofill is not working

### Code sample

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Login Demo',
      theme: ThemeData(
        useMaterial3: true,
      ),
      home: const LoginPage(),
    );
  }
}

class LoginPage extends StatefulWidget {
  const LoginPage({super.key});

  @override
  State<LoginPage> createState() => _LoginPageState();
}

class _LoginPageState extends State<LoginPage> {
  final _formKey = GlobalKey<FormState>();
  final _emailController = TextEditingController();
  final _passwordController = TextEditingController();
  bool _isPasswordVisible = false;
  bool _isLoading = false;

  @override
  void dispose() {
    _emailController.dispose();
    _passwordController.dispose();
    super.dispose();
  }

  String? _validateEmail(String? value) {
    if (value == null || value.isEmpty) {
      return 'Please enter your email';
    }
    final emailRegex = RegExp(r'^[^@]+@[^@]+\.[^@]+');
    if (!emailRegex.hasMatch(value)) {
      return 'Please enter a valid email address';
    }
    return null;
  }

  String? _validatePassword(String? value) {
    if (value == null || value.isEmpty) {
      return 'Please enter your password';
    }
    if (value.length < 6) {
      return 'Password must be at least 6 characters long';
    }
    return null;
  }

  Future<void> _handleLogin() async {
    if (!_formKey.currentState!.validate()) {
      return;
    }

    setState(() {
      _isLoading = true;
    });

    // Simulate login delay
    await Future.delayed(const Duration(seconds: 2));

    setState(() {
      _isLoading = false;
    });

    if (mounted) {
      ScaffoldMessenger.of(context).showSnackBar(
        SnackBar(
          content: Text('Login attempt for ${_emailController.text}'),
          backgroundColor: Colors.green,
        ),
      );
    }
  }

  Widget _buildLoginForm() {
    return AutofillGroup(
      child: Form(
        key: _formKey,
        child: Column(
          children: [
            // Logo or Title
            Icon(
              Icons.lock_outline,
              size: 80,
              color: Theme.of(context).colorScheme.primary,
            ),
            const SizedBox(height: 16),
            Text(
              'Welcome Back',
              style: Theme.of(
                context,
              ).textTheme.headlineMedium?.copyWith(fontWeight: FontWeight.bold),
              textAlign: TextAlign.center,
            ),
            const SizedBox(height: 8),
            Text(
              'Sign in to your account',
              style: Theme.of(context).textTheme.bodyLarge?.copyWith(
                color: Theme.of(context).colorScheme.onSurfaceVariant,
              ),
              textAlign: TextAlign.center,
            ),
            const SizedBox(height: 32),

            // Email Field
            TextFormField(
              controller: _emailController,
              keyboardType: TextInputType.emailAddress,
              autofillHints: const [AutofillHints.email],
              decoration: const InputDecoration(
                labelText: 'Email',
                hintText: 'Enter your email address',
                prefixIcon: Icon(Icons.email_outlined),
                border: OutlineInputBorder(),
              ),
              validator: _validateEmail,
              textInputAction: TextInputAction.next,
            ),
            const SizedBox(height: 16),

            // Password Field
            TextFormField(
              controller: _passwordController,
              obscureText: !_isPasswordVisible,
              autofillHints: const [AutofillHints.password],
              decoration: InputDecoration(
                labelText: 'Password',
                hintText: 'Enter your password',
                prefixIcon: const Icon(Icons.lock_outlined),
                suffixIcon: IconButton(
                  icon: Icon(
                    _isPasswordVisible
                        ? Icons.visibility_off_outlined
                        : Icons.visibility_outlined,
                  ),
                  onPressed: () {
                    setState(() {
                      _isPasswordVisible = !_isPasswordVisible;
                    });
                  },
                ),
                border: const OutlineInputBorder(),
              ),
              validator: _validatePassword,
              textInputAction: TextInputAction.done,
              onFieldSubmitted: (_) => _handleLogin(),
            ),
            const SizedBox(height: 24),

            // Login Button
            FilledButton(
              onPressed: _isLoading ? null : _handleLogin,
              style: FilledButton.styleFrom(
                padding: const EdgeInsets.symmetric(vertical: 16),
              ),
              child: _isLoading
                  ? const SizedBox(
                      height: 20,
                      width: 20,
                      child: CircularProgressIndicator(
                        strokeWidth: 2,
                        valueColor: AlwaysStoppedAnimation<Color>(Colors.white),
                      ),
                    )
                  : const Text('Sign In'),
            ),
            const SizedBox(height: 16),

            // Forgot Password Link
            TextButton(
              onPressed: () {
                ScaffoldMessenger.of(context).showSnackBar(
                  const SnackBar(
                    content: Text(
                      'Forgot password functionality not implemented',
                    ),
                  ),
                );
              },
              child: const Text('Forgot Password?'),
            ),
          ],
        ),
      ),
    );
  }

  @override
    final isWeb = kIsWeb;
    final screenWidth = MediaQuery.of(context).size.width;
    final isWideScreen = screenWidth > 600;

    return Scaffold(
      body: SafeArea(
        child: isWeb && isWideScreen
            ? Center(
                child: SingleChildScrollView(
                  padding: const EdgeInsets.all(24),
                  child: ConstrainedBox(
                    constraints: const BoxConstraints(maxWidth: 400),
                    child: Card(
                      elevation: 8,
                      child: Padding(
                        padding: const EdgeInsets.all(32),
                        child: _buildLoginForm(),
                      ),
                    ),
                  ),
                ),
              )
            : SingleChildScrollView(
                padding: const EdgeInsets.all(24),
                child: Column(
                  children: [
                    SizedBox(height: MediaQuery.of(context).size.height * 0.1),
                    _buildLoginForm(),
                  ],
                ),
              ),
      ),
    );
  }
}```

</details>


### Screenshots or Video

<details open>

[Upload media here]

</details>


### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console

[✓] Flutter (Channel stable, 3.35.6, on macOS 15.7.1 24G231 darwin-arm64, locale en-PH) [339ms]
    • Flutter version 3.35.6 on channel stable at /Users/..../fvm/versions/3.35.6
    • Framework revision 9f455d2486 (12 days ago), 2025-10-08 14:55:31 -0500
    • Engine revision d2913632a4
    • Dart version 3.9.2
    • DevTools version 2.48.0

[!] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [1,085ms]
    • Android SDK at /Users/..../Library/Android/sdk
    • ANDROID_HOME = /Users/..../Library/Android/sdk
    • ANDROID_SDK_ROOT = /Users/..../Library/Android/sdk

[✓] Xcode - develop for iOS and macOS (Xcode 26.0.1) [769ms]
    • Build 17A400
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [6ms]

[✓] Android Studio (version 2025.1) [6ms]
    • Java version openjdk version "21.0.6" 2025-01-21

[✓] VS Code (version 1.105.1) [5ms]
    • Flutter extension version 3.120.0

[✓] Connected device (6 available) [5.9s]
    • ...

[✓] Network resources [447ms]
```

Notes: 
- Output has been edited to remove some information and been replaced with `...` for privacy
- I am using FVM but this has been replicated as well on a separate machine running a fresh flutter install (manually and not via FVM)

</details>

mdebbar:
This fix has 2 parts basically:

1. Reuse autofill forms. Prior to this PR, the autofill forms were being recreated every time a text input connection is established. This behavior prevents Safari from autofilling the entire form. This PR reuses the existing form and fields instead of recreating them.

2. Re-establish the text input connection with the framework when the text field receives focus from the browser. This is necessary for iOS26's new focus behavior where it blurs the text field then focuses it before autofilling it. That blur-then-focus is a new behavior that was not accounted for by the web engine.

Fixes https://github.com/flutter/flutter/issues/177248


packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1369,6 +1369,9 @@ mixin TextInputClient {
   /// This method will only be called on iOS.
   void showAutocorrectionPromptRect(int start, int end);
 
+  /// Requests that this client refocus the text input control.
+  void refocus();
```
justinmc: This will be a breaking change for anyone that mixes in TextInputClient. This at least includes super_editor in customer_testing. You probably want to include an empty implementation here to avoid this (like showToolbar does below), unless maybe if the refocus method is critical enough that it should be a breaking change to not implement it.

justinmc (commented):
One comment about a breaking change, which might be causing some of the failures. Otherwise looks good to me overall.

packages/flutter/lib/src/services/text_input.dart
Renzo-Olivares: +1 about this being a breaking change. 

My thoughts are that we probably do want to make this a breaking change since without it `Autofill` is broken on iOS 26 and this way implementers are made aware that implementing the method is required to fix that feature. This `refocus` approach LGTM!

mdebbar: Alright, now that the approach sounds good to both of you, I'll go ahead and add tests and fix the failures.

Regarding the breaking change, I think the most reasonable path is to include an empty implementation of `refocus()` to avoid the breaking change. Reasoning:

Advanced text field implementations (like `super_editor`) do not need or use autofill functionality. They are addressing a different use case (rich text editing and document rendering). And I doubt autofill ever worked with `super_editor`given that they don't use AutofillGroup, etc.

Does that sound reasonable?

Renzo-Olivares: My initial thought was that because Autofill is a common feature of a `TextField` that we should get this in as a breaking change to make it clear to implementers that this is necessary for Autofill to work. However, giving this more thought I think for the reasons you stated @mdebbar it is reasonable to keep this change non-breaking since it is not required for basic text input, i.e if a user does not want autofill on Safari they do not need to implement `refocus`.

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -342,43 +405,19 @@ class EngineAutofillForm {
...
+      // Maybe only apply text but not selection?
+      autofill.editingState.applyToDomElement(element);
     }
-    _insertEditingElementInView(formElement, viewId);
-  }
-
-  void storeForm() {
-    formsOnTheDom[formIdentifier] = formElement;
-    _styleAutofillElements(formElement, isOffScreen: true);
   }
```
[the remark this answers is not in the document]
mdebbar: Good bot!

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -2485,6 +2534,24 @@ class HybridTextEditing {
       channel.onConnectionClosed(_clientId);
     }
   }
+
+  void _addRefocusListenerToView(int viewId) {
+    final EngineFlutterView? view = EnginePlatformDispatcher.instance.viewManager[viewId];
+    view!.dom.textEditingHost.addEventListener('focusin', createDomEventListener(_handleRefocus));
```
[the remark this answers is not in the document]
mdebbar: The `_addRefocusListenerToView` method is always given an ID of a live view. If `view` turns out to be null, then there's a serious bug somewhere.

packages/flutter/test/services/text_input_test.dart
```diff
@@ -726,6 +726,57 @@ void main() {
...
+
+      expect(connection.attached, isTrue);
+      expect(client.latestMethodCall, isEmpty);
+
+      connection.connectionClosedReceived();
+
+      expect(connection.attached, isFalse);
+      expect(client.latestMethodCall, isEmpty);
+
+      // Send refocu message to re-establish the connection.
```
loic-sharma, line 742:
```suggestion
      // Send refocus message to re-establish the connection.
```

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2150,6 +2158,15 @@ class TextInput {
       case 'TextInputClient.scribbleInteractionFinished':
         _scribbleInProgress = false;
         return;
+      case 'TextInputClient.refocus':
+        final args = methodCall.arguments as List<dynamic>;
+        final clientId = args[0] as int;
+        if (_currentConnection == null &&
```
loic-sharma: This behavior - that `TextInputClient.refocus` is ignored unless the connection was previously closed - might be confusing to other embedders that want to send this message. Should we maybe make this message's name a bit more descriptive? For example, `TextInputClient.refocusIfConnectionClosed` or `TextInputClient.refocusLastConnection`?

packages/flutter/test/services/text_input_test.dart
```diff
@@ -726,6 +726,57 @@ void main() {
...
+
+    test('TextInputClient refocus method is not called if already connected', () async {
+      final client = FakeTextInputClient(const TextEditingValue(text: 'test3'));
+      const configuration = TextInputConfiguration();
+      final TextInputConnection connection = TextInput.attach(client, configuration);
+
+      expect(connection.attached, isTrue);
+      expect(client.latestMethodCall, isEmpty);
+
+      // Send refocu message to re-establish the connection.
```
loic-sharma:
```suggestion
      // Send refocus message to re-establish the connection.
```

packages/flutter/lib/src/services/text_input.dart
mdebbar: Rationale for current implementation: if a connection is already open, then the text field is already focused, so refocusing is a no-op. Is that a correct assumption?

loic-sharma: > if a connection is already open, then the text field is already focused, so refocusing is a no-op.

While this _might_ be true for `EditableText`, I don't think we can make this assumption universally. You can make a custom text input client with weird focus behaviors.

Also @LongCatIsLooong bubbled up an interesting concern during Jan 8's text input sync ([Google internal link](https://docs.google.com/document/d/1vimZnqmmv5ZjTc9mDZ73cX3ztM6hlAXTSiznd5gB80c/edit?resourcekey=0-9Zl8Zp8pdQwYsSRm8sLcCA&tab=t.0#heading=h.eimx21agxelp)): on iOS, the input method can do actions even after you unfocused. For example, if you start an IME composition but don't complete it and then move focus, IME composition completes after focus has moved. The ideal solution would be keeping connections open for all focusable text fields, regardless of whether they are actually focused. This might be something we do in the future.

loic-sharma: (BTW I'm happy to stick with the current implementation as-is, my suggestion is only for the name of the platform message)

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -238,105 +240,166 @@ class EngineAutofillForm {
...
+    assert(elements.isEmpty);
+
+    final DomHTMLFormElement formElement = createDomHTMLFormElement();
+    // Validation is in the framework side.
+    formElement.noValidate = true;
+    formElement.method = 'post';
+    formElement.action = '#';
+    formElement.addEventListener('submit', preventDefaultListener);
+
+    // We need to explicitly disable pointer events on the form in Safari Desktop,
```
loic-sharma: Is this comment up-to-date? It looks like this disables pointer events on iOS too.

packages/flutter/lib/src/services/text_input.dart
mdebbar: I gave this a little more thought.

Based on what you said (which makes sense), I think we should be less opinionated on how the input client should behave and instead treat this as an event rather than a request for a specific action. I.e. I'm proposing we name it "onFocusReceived" or something similar, and each client should react to it as it sees fit (default would be to refocus, which will re-establish the connection). This is similar to the existing "onConnectionClosed".

I'm also going to remove the condition `if (!alreadyFocused) { refocus(); }`, and make it so it always calls `requestFocus` (which is mostly a no-op if the node already has focus: [see code](https://github.com/flutter/flutter/blob/167f30fc7d2ce4f83e2a5439eb5e074c431c27b3/packages/flutter/lib/src/widgets/focus_manager.dart#L1184-L1187)).

loic-sharma (dismissed):
This looks good to me modulo some minor nitpicks 👍

packages/flutter/lib/src/services/text_input.dart
loic-sharma: > I'm proposing we name it "onFocusReceived" 

Love it! Thanks for the thoughtful response :)

engine/src/flutter/lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -3385,8 +3401,9 @@ Future<void> testMain() async {
           createAutofillInfo('email', 'field1'),
           fields,
         )!;
+        // TODO(mdebbar): Things are different now. DOM elements aren't created immediately like they used to.
```
justinmc: Did you plan to come back to this TODO in this PR?

engine/src/flutter/lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -3413,8 +3430,9 @@ Future<void> testMain() async {
         createAutofillInfo('email', 'field1'),
         fields,
       )!;
+      // TODO(mdebbar): Things are different now. DOM elements aren't created immediately like they used to.
```
justinmc: Calling out another TODO

justinmc (dismissed):
LGTM assuming the tests pass 👍

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -1369,6 +1369,12 @@ mixin TextInputClient {
   /// This method will only be called on iOS.
   void showAutocorrectionPromptRect(int start, int end);
 
+  /// Requests that this client refocus the text input control.
```
loic-sharma: My understanding is the "text input control" is the platform side of the text input connection, whereas this is referring to the client side of the text input connection. Maybe instead say:

```suggestion
  /// Requests that this client refocus itself.
```

engine/src/flutter/lib/web_ui/test/engine/text_editing_test.dart
mdebbar: Yep, done.

packages/flutter/lib/src/services/text_input.dart
mdebbar: Actually, this comment needs to be updated to reflect the rename. I'm thinking something like:
```
/// Notifies the client that the platform moved focus back to this input.
```
WDYT?

loic-sharma: Yup sounds good 👍

packages/flutter/test/services/text_input_test.dart
```diff
@@ -726,6 +726,56 @@ void main() {
...
+
+      expect(connection.attached, isTrue);
+      expect(client.latestMethodCall, isEmpty);
+
+      connection.connectionClosedReceived();
+
+      expect(connection.attached, isFalse);
+      expect(client.latestMethodCall, isEmpty);
+
+      // Send refocu message to re-establish the connection.
```
loic-sharma, line 742: FYI this comment got missed: https://github.com/flutter/flutter/pull/182024/changes#r2874745020

```suggestion
      // Send refocus message to re-establish the connection.
```

packages/flutter/test/services/text_input_test.dart
loic-sharma: Resolving this in favor of a new comment so that it is bubbled up: https://github.com/flutter/flutter/pull/182024/changes#r2942820422

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -1736,6 +1736,250 @@ void main() {
...
+    await tester.pump();
+
+    expect(editableText.widget.focusNode.hasFocus, isFalse);
+
+    // Scroll far away from the text field.
+    final ScrollPosition position = scrollController.position;
+    scrollController.jumpTo(position.maxScrollExtent);
+    await tester.pumpAndSettle();
+
+    // The text field is unmounted, so it should not regain focus.
```
loic-sharma, line 1985: I'm surprised this works since editable text mixes `AutomaticKeepAliveClientMixin`. My understanding is that it will stay mounted even if it is in a lazy list and scrolled out of view. I suspect this comment is incorrect and something else is blocking the refocus.

<details><summary>Sample app that shows automatic keep alive stays mounted in a lazy list...</summary>

Generated by Gemini, run the app and scroll down:

```dart
import 'dart:async';

void main() => runApp(const MaterialApp(home: KeepAliveDemo()));

class KeepAliveDemo extends StatelessWidget {
  const KeepAliveDemo({super.key});

  @override
    return Scaffold(
      appBar: AppBar(title: const Text('KeepAlive Mounted Proof')),
      body: ListView.builder(
        itemCount: 50,
        itemBuilder: (context, index) => KeepAliveItem(index: index),
      ),
    );
  }
}

class KeepAliveItem extends StatefulWidget {
  final int index;
  const KeepAliveItem({super.key, required this.index});

  @override
  State<KeepAliveItem> createState() => _KeepAliveItemState();
}

// 1. Add the Mixin
class _KeepAliveItemState extends State<KeepAliveItem> 
    with AutomaticKeepAliveClientMixin {
  
  Timer? _timer;

  @override
  void initState() {
    super.initState();
    // 2. Start a timer only for the first item to track its status
    if (widget.index == 0) {
      _timer = Timer.periodic(const Duration(seconds: 1), (timer) {
        // This will print even when off-screen
        debugPrint("Item ${widget.index} - mounted: $mounted, time: ${DateTime.now()}");
      });
    }
  }

  @override
  void dispose() {
    _timer?.cancel();
    debugPrint("Item ${widget.index} DISPOSED"); // This won't fire for index 0
    super.dispose();
  }

  // 3. Return true to keep the state alive
  @override
  bool get wantKeepAlive => widget.index == 0; // Only keep the first item alive

  @override
    // 4. Required super.build call
    super.build(context);

    return Container(
      height: 200,
      margin: const EdgeInsets.all(8),
      color: Colors.blueGrey[100 * (widget.index % 9 + 1)],
      child: Center(
        child: Text(
          "Item ${widget.index}\n${widget.index == 0 ? 'Watch Console' : ''}",
          textAlign: TextAlign.center,
          style: const TextStyle(fontSize: 20),
        ),
      ),
    );
  }
}
```

</summary>

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -1736,6 +1736,250 @@ void main() {
...
+
+    editableText.onFocusReceived();
+    await tester.pump();
+
+    expect(focusNode.hasFocus, isFalse);
+  });
+
+  testWidgets('does not refocus when it is hidden by a new route', (WidgetTester tester) async {
+    final navigatorKey = GlobalKey<NavigatorState>();
+    final focusNode = FocusNode(debugLabel: 'EditableText Node');
```
loic-sharma, line 1865: I suspect this needs a `addTearDown(focusNode.dispose);`. Same comment for tests below

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -1736,6 +1736,250 @@ void main() {
...
+
+    // The text field is visible again, so it should regain focus this time.
+    editableText.onFocusReceived();
+    await tester.pump();
+    expect(editableText.widget.focusNode.hasFocus, isTrue);
+  }, skip: true);
+
+  testWidgets('does not refocus when scrolled away in a ListView', (WidgetTester tester) async {
+    final focusNode = FocusNode(debugLabel: 'EditableText Node');
+    final scrollController = ScrollController();
```
loic-sharma, line 1938: I suspect this needs a `addTearDown(scrollController.dispose)`

packages/flutter/test/widgets/editable_text_test.dart
mdebbar: Keep-alive is only active while the text field has focus. Once the text field loses focus, keep-alive is no more:
https://github.com/flutter/flutter/blob/7290bb4e281ecd5abc386f9ff19399e3f3a1a941/packages/flutter/lib/src/widgets/editable_text.dart#L2622

What the test is doing:
- Pump the text field widget inside a ListView.
- The field has autofocus so it receives focus right away.
- Make the field lose focus by calling `connectionClosed()`.
- Scroll down until the field is hidden and unmounted.
- Try to refocus.
- Check that it didn't refocus because it's not alive anymore.

I thought this was the use case you wanted to test. If I have misunderstood, please let me know.

loic-sharma: Ahhh I didn't realize `EditableText` only has keep alive behavior if it is focused. `EditableText` is indeed unmounted here. Ignore me!

EDIT: I didn't see your comment until I reloaded the page :)

packages/flutter/test/widgets/editable_text_test.dart
```diff
@@ -1736,6 +1736,250 @@ void main() {
...
+
+    // Pop the route.
+    navigatorKey.currentState!.pop();
+    await tester.pumpAndSettle();
+
+    // The text field is visible again, so it should regain focus this time.
+    editableText.onFocusReceived();
+    await tester.pump();
+    expect(editableText.widget.focusNode.hasFocus, isTrue);
+  }, skip: true);
```
loic-sharma: If we land this without fixing this issue, could you add TODO + GitHub issue URL for this?  See: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#comment-all-test-skips

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -4063,6 +4063,13 @@ class EditableTextState extends State<EditableText>
     }
   }
 
+  @override
+  void onFocusReceived() {
```
loic-sharma: I believe this is the logic that'll fix the push route case:

```dart
final bool isCurrent = ModalRoute.of(context)?.isCurrent ?? true;
if (!isCurrent) {
  return;
}
```

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2150,6 +2158,13 @@ class TextInput {
       case 'TextInputClient.scribbleInteractionFinished':
         _scribbleInProgress = false;
         return;
+      case 'TextInputClient.onFocusReceived':
+        final args = methodCall.arguments as List<dynamic>;
+        final clientId = args[0] as int;
+        if (_lastConnection != null && _lastConnection!._id == clientId) {
+          _lastConnection!._client.onFocusReceived();
+        }
+        return;
```
loic-sharma: Should we return a boolean on whether the text input client was focused?

Imagine if Flutter Web sends the `TextInputClient.onFocusReceived` message but the framework decides not to focus the text input client. Will Flutter Web and the framework be out-of-sync? Could that cause issues?

loic-sharma (dismissed):
LGTM module pending comments!

packages/flutter/lib/src/widgets/editable_text.dart
mdebbar: Hmm I have mixed feelings about this. I feel like that should be handled in `focusNode.canRequestFocus`, not in `EditableText`.

packages/flutter/lib/src/services/text_input.dart
mdebbar: Good idea! I'll add a return value.

packages/flutter/lib/src/widgets/editable_text.dart
loic-sharma: I checked with the team, this does indeed appear to be a bug with `focusNode.canRequestFocus` that we should fix. That's tracked by: https://github.com/flutter/flutter/issues/46235.

Let's keep the `onFocusReceived` logic as-is, but let's update the `skip: true` on the route test to link this issue: https://github.com/flutter/flutter/issues/46235

packages/flutter/test/widgets/editable_text_test.dart
loic-sharma: Let's link to the bug for `FocusNode.canRequestFocus` is `true` even if the route isn't current:

```suggestion
  }, skip: true); // https://github.com/flutter/flutter/issues/46235
```

That'll make it more clear what we need to fix before turning on this test :)

loic-sharma: Closing in favor of: https://github.com/flutter/flutter/pull/182024/changes#r2977571051

packages/flutter/lib/src/widgets/editable_text.dart
loic-sharma: I'll go ahead and mark this as resolved in favor of this comment: https://github.com/flutter/flutter/pull/182024#discussion_r2977571051

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -2432,6 +2483,17 @@ class TextEditingChannel {
...
   }
+
+  /// Sends the 'TextInputClient.onFocusReceived' message to the framework.
+  void onFocusReceived(int? clientId) {
+    EnginePlatformDispatcher.instance.invokeOnPlatformMessage(
+      'flutter/textinput',
+      const JSONMethodCodec().encodeMethodCall(
+        MethodCall('TextInputClient.onFocusReceived', <dynamic>[clientId]),
+      ),
+      _emptyCallback,
```
loic-sharma: Should we do something like log an error in debug mode if the framework returns false? That would indicate a logical error in our code where the embedder expected to be able to request focus, but the framework could not.

I don't feel super strongly about this, feel free to punt.

loic-sharma (dismissed):
Re-LGTM

engine/src/flutter/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
mdebbar: There may be legit situations where this could happen e.g: if the app (for whatever reason) unmounts the text  field as soon as it loses focus.

I think logging a warning is fair though.

loic-sharma (approved):
Re-re-re-LGTM 😂

PR #181976, opened 2026-02-05
Update DEPS for 3.41.0-0.3 beta hotfix release

justinmc:
There were 2 cherry picks include:

 * https://github.com/flutter/flutter/pull/181732
 * https://github.com/flutter/flutter/pull/181670 (containing latest engine change 0aafd14efaa95e7adb8bd23bf442bf19b2a669b4)

PR #182035, opened 2026-02-06
Update engine.version for the beta hotfix

justinmc:
Following up from updating DEPS in https://github.com/flutter/flutter/pull/181976.


camsim99 (commented):
I'm pretty sure should be the DEPS update commit https://github.com/flutter/flutter/commit/492bf6df86d056e47d05a3d5404f3d78939b3dae

What did you use to update engine.verison? We should use `bin/internal/last_engine_commit.sh > bin/internal/engine.version`.

justinmc:
Whoops, thanks for catching that. I think I did that but maybe before the DEPS PR had merged. I'll update it.

camsim99 (approved):

PR #137936, opened 2023-11-06
Implement CupertinoMenu and CupertinoNestedMenu

davidhicks980:
Resolves https://github.com/flutter/flutter/issues/60298, https://github.com/notDmDrl/pull_down_button/issues/26

This PR adds a CupertinoMenu, CupertinoNestedMenu, and associated widgets that increase feature parity with iOS. 

I'm submitting a draft PR to collect feedback on the overall API and layout design before writing more extensive layout and API tests. I have added a few tests based on the PopupMenu widget, and I've done some manual accessibility testing across iOS, Android, and MacOS. Comments and an example have been added, but could use review. I also added TODO annotations for any particularly pressing questions I have. 

I'm still fairly new to Flutter, so I apologize for any unconventional code. Let me know if there are any pieces we need to tear out or rethink. I'm interested to hear different ideas regarding how to solve this menu, so I appreciate you all taking a look!

Also, I used @notDmDrl's repo as a base for my contributions: https://github.com/notDmDrl/pull_down_button, so I'm not sure whether I should contact him with regards to signing a CLA. 

Basic example: 
```dart
import 'package:flutter/cupertino.dart';

/// Flutter code sample for [CupertinoMenuButton]
void main() => runApp(const CupertinoMenuApp());
class CupertinoMenuApp extends StatelessWidget {
  const CupertinoMenuApp({super.key});
  @override
  Widget build(BuildContext context) {
    return const CupertinoApp(
      theme: CupertinoThemeData(brightness: Brightness.light),
      home: CupertinoMenuExample(),
    );
  }
}

class CupertinoMenuExample extends StatefulWidget {
  const CupertinoMenuExample({super.key});

  @override
  State<CupertinoMenuExample> createState() => _CupertinoMenuExampleState();
}

class _CupertinoMenuExampleState extends State<CupertinoMenuExample> {
  String _checkedValue = 'Cat';
  @override
  Widget build(BuildContext context) {
    return CupertinoPageScaffold(
      navigationBar: const CupertinoNavigationBar(
        middle: Text('CupertinoMenu Sample'),
      ),
      child: Align(
        alignment: Alignment(0, -0.5),
        child: SizedBox(
          width: 100,
          height: 100,
          child:  CupertinoMenuButton<String>(
            itemBuilder: (BuildContext context) {
              return <CupertinoMenuEntry<String>>[
                CupertinoStickyMenuHeader(
                  leading: Container(
                    width: 42,
                    height: 42,
                    decoration:  ShapeDecoration(
[...]

PR #182036, opened 2026-02-06
[cupertino.dart] Implement CupertinoMenuAnchor and CupertinoMenuItem using RawMenuAnchor

### LINKED ISSUE #60298: Pull-Down Menus for iOS 14
On iOS 14, Pull-Down Menus have been added.

> Menus offer several advantages over action sheets, context menus, and popovers. For example:
>
> A menu opens very near the button that reveals it, so people can instantly understand the relationship between the menu's items and the action they're performing.
>
> In addition to listing actions, a menu can offer selections people can use to affect the primary action.
>
> Menus animate quickly into view and don't dim the screen when they appear, giving both the transition and the overall experience a lightweight feel.

Image

Image

https://developer.apple.com/design/human-interface-guidelines/pull-down-buttons

davidhicks980:
This PR implements `CupertinoMenuAnchor`, `CupertinoMenuItem`, and `CupertinoMenuDivider` using `RawMenuAnchor`. 

Resolves https://github.com/flutter/flutter/issues/60298, https://github.com/notDmDrl/pull_down_button/issues/26, https://github.com/flutter/flutter/pull/137936.

https://github.com/user-attachments/assets/cdaee8da-888b-4f64-8bf3-49c873f6d6e1

Dartpad: https://dartpad.dev/8c6bba779b6a00e95582b61b132292bc

@dkwingsmt 

----

Edit: Reland of https://github.com/flutter/flutter/pull/174695 after revert https://github.com/flutter/flutter/pull/182010 due to breakage when tests were randomized. 

Per @jason-simmons -
> The error seen on CI can be reproduced by running:
> flutter test --test-randomize-ordering-seed=20260206 test/cupertino/menu_anchor_test.dart 

Thanks to those who caught the error -- sorry I didn't notice.

From what I can deduce, two errors were present.

**Test 1: Menus do not close on root menu internal scroll (and others)**
The first error occurred due to TestPointers being created with an existing pointer number. This was fixed by using tester.nextPointer instead of a fixed number.

Code: https://github.com/davidhicks980/flutter/blob/054f3b03f2ab04822760c600e4259fbd828ebd4d/packages/flutter/test/cupertino/menu_anchor_test.dart#L968

```dart
    final pointer = TestPointer(
       1, // Fixed with tester.nextPointer
       ui.PointerDeviceKind.mouse
    );
    await tester.sendEventToBinding(pointer.hover(tester.getCenter(find.text(Tag.a.text))));
    await tester.pump();
```

**Test 2: `Menu scale rebounds to full size when swipe gesture ends`**

The second error occurred due to the _SwipeRegion's GestureRecognizer outliving its state and calling didSwipeLeave on a disposed CupertinoMenuItem. <s>This shouldn't be a problem -- swiping is supposed to continue so long as a user has their pointer down -- but the CupertinoMenuItem should have checked whether it was mounted before entering code that could schedule a rebuild (e.g. _handleActivation, isSwiped)</s> Nevermind -- I realize now there isn't much of a reason for letting the swipe outlive the region, so I just disposed of the swipe when the SwipeRegion is unmounted. I still kept the mounted check.

Code:
https://github.com/davidhicks980/flutter/blob/054f3b03f2ab04822760c600e4259fbd828ebd4d/packages/flutter/test/cupertino/menu_anchor_test.dart#L1285

```dart
@override
void didSwipeLeave({bool pointerUp = false}) {
    // This function could be called after disposal, but _handleActivation and isSwiped both could trigger rebuilds. 
    // Fix:
    // if (!mounted) {
    //   return
    // }

    if (isEnabled && pointerUp) {
      _handleActivation();
    }

    isSwiped = false;
  }
```

I ran an additional 500 randomized menu_anchor_test.dart runs and did not experience another exception. I believe everything should be good-to-go, but I'll pay closer attention to cocoon.

---

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/api/lib/cupertino/menu_anchor/menu_anchor.0.dart
+++ b/examples/api/lib/cupertino/menu_anchor/menu_anchor.0.dart
@@ -1,0 +1,86 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/cupertino.dart';
+
+/// Flutter code sample for a basic [CupertinoMenuAnchor].
+void main() => runApp(const CupertinoMenuAnchorApp());
+
+class CupertinoMenuAnchorApp extends StatelessWidget {
+  const CupertinoMenuAnchorApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return const CupertinoApp(
+      home: CupertinoPageScaffold(child: CupertinoMenuAnchorExample()),
+    );
+  }
+}
+
+class CupertinoMenuAnchorExample extends StatefulWidget {
+  const CupertinoMenuAnchorExample({super.key});
+
+  @override
+  State<CupertinoMenuAnchorExample> createState() =>
+      _CupertinoMenuAnchorExampleState();
+}
+
+class _CupertinoMenuAnchorExampleState
+    extends State<CupertinoMenuAnchorExample> {
+  // Optional: Create a focus node to allow focus traversal between the menu
+  // button and the menu overlay.
+  final FocusNode _buttonFocusNode = FocusNode(debugLabel: 'Menu Button');
+  AnimationStatus _animationStatus = AnimationStatus.dismissed;
+
+  @override
+  void dispose() {
+    _buttonFocusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      children: <Widget>[
+        Center(
+          child: CupertinoMenuAnchor(
+            onAnimationStatusChanged: (AnimationStatus status) {
+              // Since we are only checking the animation status when the button
+              // is pressed, we don't need to call setState here.
+              _animationStatus = status;
+            },
+            childFocusNode: _buttonFocusNode,
+            menuChildren: <Widget>[
+              CupertinoMenuItem(
+                onPressed: () {},
+                subtitle: const Text('Subtitle'),
+                trailing: const Icon(CupertinoIcons.star),
+                child: const Text('Menu Item'),
+              ),
+            ],
+            builder:
+                (
+                  BuildContext context,
+                  MenuController controller,
+                  Widget? child,
+                ) {
+                  return CupertinoButton(
+                    sizeStyle: CupertinoButtonSize.small,
+                    focusNode: _buttonFocusNode,
+                    onPressed: () {
+                      if (_animationStatus.isForwardOrCompleted) {
+                        controller.close();
+                      } else {
+                        controller.open();
+                      }
+                    },
+                    child: const Icon(CupertinoIcons.ellipsis_vertical_circle),
+                  );
+                },
+          ),
+        ),
+      ],
+    );
+  }
+}
--- a/examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
+++ b/examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
@@ -1,0 +1,133 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/cupertino.dart';
+
+/// Flutter code sample for a [CupertinoMenuAnchor] that shows a menu with three
+/// [CupertinoMenuItem]s and one [CupertinoMenuDivider].
+void main() => runApp(const CupertinoMenuAnchorApp());
+
+class CupertinoMenuAnchorApp extends StatelessWidget {
+  const CupertinoMenuAnchorApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return const CupertinoApp(
+      home: CupertinoPageScaffold(
+        navigationBar: CupertinoNavigationBar(
+          middle: Text('CupertinoMenuAnchor Example'),
+        ),
+        child: CupertinoMenuAnchorExample(),
+      ),
+    );
+  }
+}
+
+class CupertinoMenuAnchorExample extends StatefulWidget {
+  const CupertinoMenuAnchorExample({super.key});
+
+  @override
+  State<CupertinoMenuAnchorExample> createState() =>
+      _CupertinoMenuAnchorExampleState();
+}
+
+class _CupertinoMenuAnchorExampleState
+    extends State<CupertinoMenuAnchorExample> {
+  // Optional: Create a focus node to allow focus traversal between the menu
+  // button and the menu overlay.
+  final FocusNode _buttonFocusNode = FocusNode(debugLabel: 'Menu Button');
+  String _pressedItem = '';
+  AnimationStatus _status = AnimationStatus.dismissed;
+
+  @override
+  void dispose() {
+    _buttonFocusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Column(
+        spacing: 20,
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: <Widget>[
+          CupertinoMenuAnchor(
+            onAnimationStatusChanged: (AnimationStatus status) {
+              _status = status;
+            },
+            childFocusNode: _buttonFocusNode,
+            menuChildren: <Widget>[
+              CupertinoMenuItem(
+                onPressed: () {
+                  setState(() {
+                    _pressedItem = 'Regular Item';
+                  });
+                },
+                subtitle: const Text('Subtitle'),
+                child: const Text('Regular Item'),
+              ),
+              CupertinoMenuItem(
+                onPressed: () {
+                  setState(() {
+                    _pressedItem = 'Colorful Item';
+                  });
+                },
+                decoration: const WidgetStateProperty<BoxDecoration>.fromMap(<
+                  WidgetStatesConstraint,
+                  BoxDecoration
+                >{
+                  WidgetState.dragged: BoxDecoration(color: Color(0xAEE48500)),
+                  WidgetState.pressed: BoxDecoration(color: Color(0xA6E3002A)),
+                  WidgetState.hovered: BoxDecoration(color: Color(0xA90069DA)),
+                  WidgetState.focused: BoxDecoration(color: Color(0x9B00C8BE)),
+                  WidgetState.any: BoxDecoration(color: Color(0x00000000)),
+                }),
+                child: const Text('Colorful Item'),
+              ),
+              const CupertinoMenuDivider(),
+              CupertinoMenuItem(
+                trailing: const Icon(CupertinoIcons.delete),
+                isDestructiveAction: true,
+                child: const Text('Destructive Item'),
+                onPressed: () {
+                  setState(() {
+                    _pressedItem = 'Destructive Item';
+                  });
+                },
+              ),
+            ],
+            builder:
+                (
+                  BuildContext context,
+                  MenuController controller,
+                  Widget? child,
+                ) {
+                  return CupertinoButton(
+                    sizeStyle: CupertinoButtonSize.medium,
+                    focusNode: _buttonFocusNode,
+                    onPressed: () {
+                      if (_status.isForwardOrCompleted) {
+                        controller.close();
+                      } else {
+                        controller.open();
+                      }
+                    },
+                    child: Text(
+                      _status.isForwardOrCompleted ? 'Close Menu' : 'Open Menu',
+                    ),
+                  );
+                },
+          ),
+          Text(
+            _pressedItem.isEmpty
+                ? 'No items pressed'
+                : 'You Pressed: $_pressedItem',
+            style: CupertinoTheme.of(context).textTheme.textStyle,
+          ),
+        ],
+      ),
+    );
+  }
+}
--- a/examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
+++ b/examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
@@ -55,7 +55,9 @@ class _CupertinoMenuAnchorExampleState
         children: <Widget>[
           CupertinoMenuAnchor(
             onAnimationStatusChanged: (AnimationStatus status) {
-              _status = status;
+              setState(() {
+                _status = status;
+              });
             },
             childFocusNode: _buttonFocusNode,
             menuChildren: <Widget>[
--- a/packages/flutter/lib/src/cupertino/menu_anchor.dart
+++ b/packages/flutter/lib/src/cupertino/menu_anchor.dart
@@ -42,6 +42,9 @@ bool get _isCupertino {
   }
 }
 
+// TODO(davidhicks980): Move typography constants to a more appropriate
+// location. https://github.com/flutter/flutter/issues/182933
+
 /// The font family for menu items at smaller text scales.
 const String _kBodyFont = 'CupertinoSystemText';
 
--- a/packages/flutter/lib/src/cupertino/menu_anchor.dart
+++ b/packages/flutter/lib/src/cupertino/menu_anchor.dart
@@ -443,6 +443,8 @@ class CupertinoMenuAnchor extends StatefulWidget {
   ///
   /// Typically, this is a button that calls [MenuController.open] when pressed.
   ///
+  /// The [builder] will rebuild when the menu's [AnimationStatus] changes.
+  ///
   /// If null, the [CupertinoMenuAnchor] will be the size that its parent
   /// allocates for it.
   final RawMenuAnchorChildBuilder? builder;
@@ -539,7 +541,7 @@ class _CupertinoMenuAnchorState extends State<CupertinoMenuAnchor> with TickerPr
 
   MenuController get _menuController => widget.controller ?? _internalMenuController!;
   MenuController? _internalMenuController;
-  bool get isOpening => _animationStatus.isForwardOrCompleted;
+  bool get isOpenOrOpening => _animationStatus.isForwardOrCompleted;
   bool get enableSwipe =>
       widget.enableSwipe &&
       switch (_animationStatus) {
@@ -622,7 +624,7 @@ class _CupertinoMenuAnchorState extends State<CupertinoMenuAnchor> with TickerPr
   }
 
   void _handleAnchorSwipeStart() {
-    if (isOpening || !widget.enableLongPressToOpen) {
+    if (isOpenOrOpening || !widget.enableLongPressToOpen) {
       return;
     }
     _menuController.open();
@@ -666,11 +668,11 @@ class _CupertinoMenuAnchorState extends State<CupertinoMenuAnchor> with TickerPr
 
   Widget _buildMenuOverlay(BuildContext childContext, RawMenuOverlayInfo info) {
     return ExcludeSemantics(
-      excluding: !isOpening,
+      excluding: !isOpenOrOpening,
       child: IgnorePointer(
-        ignoring: !isOpening,
+        ignoring: !isOpenOrOpening,
         child: ExcludeFocus(
-          excluding: !isOpening,
+          excluding: !isOpenOrOpening,
           child: _MenuOverlay(
             constrainCrossAxis: widget.constrainCrossAxis,
             visibilityAnimation: _animationController.view,
--- a/packages/flutter/lib/src/cupertino/menu_anchor.dart
+++ b/packages/flutter/lib/src/cupertino/menu_anchor.dart
@@ -18,7 +18,6 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
 import 'colors.dart';
-import 'dialog.dart';
 import 'theme.dart';
 
 // Dismiss is handled by RawMenuAnchor
@@ -41,6 +40,11 @@ bool get _isCupertino {
       return false;
   }
 }
+
+const Color _kMenuSurfaceColor = CupertinoDynamicColor.withBrightness(
+  color: Color(0xCCF2F2F2),
+  darkColor: Color(0xCC2D2D2D),
+);
 
 /// The font family for menu items at smaller text scales.
 const String _kBodyFont = 'CupertinoSystemText';
@@ -1061,6 +1065,7 @@ class _MenuOverlayState extends State<_MenuOverlay>
       );
       constraints = BoxConstraints.tightFor(width: menuWidth.points);
     }
+
     Widget child = _SwipeSurface(
       child: TapRegion(
         groupId: widget.tapRegionGroupId,
@@ -1091,18 +1096,28 @@ class _MenuOverlayState extends State<_MenuOverlay>
                 child: FadeTransition(
                   opacity: _fadeAnimation,
                   alwaysIncludeSemantics: true,
-                  child: CupertinoPopupSurface(
-                    child: AnimatedBuilder(
-                      animation: _sizeAnimation,
-                      builder: _buildAlign,
-                      child: Semantics(
-                        explicitChildNodes: true,
-                        scopesRoute: true,
-                        child: ConstrainedBox(
-                          constraints: constraints,
-                          child: SingleChildScrollView(
-                            clipBehavior: Clip.none,
-                            child: Column(mainAxisSize: MainAxisSize.min, children: _children),
+                  // TODO(davidhicks980): Use CupertinoPopupSurface when
+                  // appearance is fixed on Impeller.
+                  // https://github.com/flutter/flutter/issues/182066
+                  child: ClipRSuperellipse(
+                    borderRadius: const BorderRadius.all(Radius.circular(13)),
+                    child: BackdropFilter(
+                      filter: ui.ImageFilter.blur(sigmaX: 30, sigmaY: 30),
+                      child: ColoredBox(
+                        color: CupertinoDynamicColor.resolve(_kMenuSurfaceColor, context),
+                        child: AnimatedBuilder(
+                          animation: _sizeAnimation,
+                          builder: _buildAlign,
+                          child: Semantics(
+                            explicitChildNodes: true,
+                            scopesRoute: true,
+                            child: ConstrainedBox(
+                              constraints: constraints,
+                              child: SingleChildScrollView(
+                                clipBehavior: Clip.none,
+                                child: Column(mainAxisSize: MainAxisSize.min, children: _children),
+                              ),
+                            ),
                           ),
                         ),
                       ),
@@ -1197,6 +1212,8 @@ class _ShadowPainter extends CustomPainter {
       ..addRect(rect.inflate(200))
       ..addRRect(RRect.fromRectAndRadius(rect, radius));
 
+    // Clip the shadow underneath the menu shape to make the shadow appear more
+    // vibrant.
     canvas
       ..save()
       ..clipPath(maskPath)
@@ -2780,14 +2797,6 @@ class _SwipeRegionState extends State<_SwipeRegion> {
     _recognizer!.addPointer(event);
   }
 
-  void _handleSwipeEnd(DragEndDetails position) {
-    _completeSwipe();
-  }
-
-  void _handleSwipeCancel() {
-    _completeSwipe();
-  }
-
   void _handleSwipeUpdate(DragUpdateDetails updateDetails) {
     _position = _position! + updateDetails.delta;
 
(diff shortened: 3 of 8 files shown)


examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
```diff
@@ -0,0 +1,133 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Column(
+        spacing: 20,
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: <Widget>[
+          CupertinoMenuAnchor(
+            onAnimationStatusChanged: (AnimationStatus status) {
+              _status = status;
```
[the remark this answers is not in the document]
davidhicks980: This is actually not needed because the builder already is called on status changes, but it's probably better to include since it's a good practice.

dkwingsmt (dismissed):
RSLGTM

MaherSafadii:
> Oh I see it on dark mode. I'll update

I was actually referring to light mode, the example you replied with looks good, but the original video recording maybe had some darker shadow, I did't try dark mode.

davidhicks980:
@MaherSafadii You're correct. I'll update the video when I get a chance. Impeller is causing some problems that I need to sort out...

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3063 @@
...
+      initialPosition: position,
+      onSwipeUpdate: _handleSwipeUpdate,
+      onSwipeEnd: _handleSwipeEnd,
+      onSwipeCanceled: _handleSwipeCancel,
+    );
+  }
+
+  void _completeSwipe() {
+    _position = null;
+    widget.onDistanceChanged(0);
```
chunhtai: This should check for is mount in case the cancel is receive after dispose, it is a foot gun to call callback after the widget state is disposed.

chunhtai: if it is not logically possible, at least put an assert is mounted

davidhicks980: Good point. There shouldn't be any logical reasons not to. Here's the final change:
```dart
  void _completeSwipe() {
    _recognizer?.dispose();
    _recognizer = null;
    _position = null;
    if (mounted) {
      widget.onDistanceChanged(0);
    }
  }
```

dkwingsmt:
Thanks for the diff gist.

> * Replaced CupertinoPopupSurface with blur/clip path/colored box while waiting for https://github.com/flutter/flutter/issues/182066 to land

This seems like a bug of bounded blur. I'll look into it. Meanwhile feel free to change `bounded` to false while linking that issue as TODO in order to unblock this PR.

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3070 @@
...
+///
+/// Empirically, the jump from one policy to the other occurs at the following text
+/// scale factors:
+/// * Max "regular" scale factor ≈ 23/17 ≈ 1.352... (normalized text scale: 6)
+/// * Min "accessible" scale factor  ≈ 28/17 ≈ 1.647... (normalized text scale: 11)
+///
+/// The following constant represents a division in text scale factor beyond which
+/// we want to change how the menu is laid out.
+///
+/// This explanation was ported from cupertino/dialog.dart.
```
chunhtai: should also these constant be defined elsewhere to be share across similar to _DynamicTypeStyle? maybe add todo and group all these together some how?

examples/api/lib/cupertino/menu_anchor/menu_anchor.0.dart
```diff
@@ -0,0 +1,86 @@
...
+                (
+                  BuildContext context,
+                  MenuController controller,
+                  Widget? child,
+                ) {
+                  return CupertinoButton(
+                    sizeStyle: CupertinoButtonSize.small,
+                    focusNode: _buttonFocusNode,
+                    onPressed: () {
+                      if (_animationStatus.isForwardOrCompleted) {
```
chunhtai, line 72: why uses `_animationStatus` instead of controller.isopen

examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
```diff
@@ -0,0 +1,135 @@
...
+                (
+                  BuildContext context,
+                  MenuController controller,
+                  Widget? child,
+                ) {
+                  return CupertinoButton(
+                    sizeStyle: CupertinoButtonSize.medium,
+                    focusNode: _buttonFocusNode,
+                    onPressed: () {
+                      if (_status.isForwardOrCompleted) {
```
chunhtai, line 113: same

examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
```diff
@@ -0,0 +1,135 @@
...
+                        controller.close();
+                      } else {
+                        controller.open();
+                      }
+                    },
+                    child: Text(
+                      _status.isForwardOrCompleted ? 'Close Menu' : 'Open Menu',
```
chunhtai, line 120: are we guaranteed this build method will be called when status change?

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3070 @@
...
+  /// A list of menu items to display in the menu.
+  final List<Widget> menuChildren;
+
+  /// The widget that this [CupertinoMenuAnchor] surrounds.
+  ///
+  /// Typically, this is a button that calls [MenuController.open] when pressed.
+  ///
+  /// If null, the [CupertinoMenuAnchor] will be the size that its parent
+  /// allocates for it.
+  final RawMenuAnchorChildBuilder? builder;
```
chunhtai, line 452: should also mention this is called when animation status changes

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3070 @@
...
+  /// menu changes.
+  ///
+  /// This callback provides a way to determine when the menu is opening or
+  /// closing. This is necessary because the [MenuController.isOpen] property
+  /// remains true throughout the opening, opened, and closing phases, and
+  /// therefore cannot be used on its own to determine the current animation
+  /// direction.
+  ///
+  /// Defaults to null.
+  final CupertinoMenuAnimationStatusChangedCallback? onAnimationStatusChanged;
```
chunhtai, line 368: I personally feel this is a weird api, but looks like this is mirroring the material version's. So I guess it is fine.

Personally I think if we want to expose this, it should pass into the builder.

examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
chunhtai: looks like it will based on the code, should document on the builder method

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3070 @@
...
+  );
+
+  late final AnimationController _animationController;
+  final FocusScopeNode _menuScopeNode = FocusScopeNode(debugLabel: 'Menu Scope');
+  final ValueNotifier<double> _swipeDistanceNotifier = ValueNotifier<double>(0);
+  bool? _hasLeadingWidget;
+
+  MenuController get _menuController => widget.controller ?? _internalMenuController!;
+  MenuController? _internalMenuController;
+  bool get isOpening => _animationStatus.isForwardOrCompleted;
```
chunhtai: this name is a bit confusing, maybe just isOpen.

IsOpening implies the animation is still ongoing

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3070 @@
...
+      ..dispose();
+    _internalMenuController = null;
+    _swipeDistanceNotifier.dispose();
+    super.dispose();
+  }
+
+  bool _resolveHasLeading() {
+    return widget.menuChildren.any((Widget element) {
+      return switch (element) {
+        final CupertinoMenuEntry entry => entry.hasLeading(context),
```
chunhtai, line 603: what is the use case for passing context to this method? I think this should be the root of the menu? can there be Anchor scope above?

lukemmtt:
Just wanted to say, super grateful for your work on this @davidhicks980! I've been a user of https://pub.dev/packages/pull_down_button for a while, and your work here is a worthy successor to that and a great improvement in several dimensions, so I'm excited to see this merged soon. I appreciate you! 🥳

Thank you also @chunhtai for your thoughtful reviewing here! 🙏🏻

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: It's for submenus. IIRC, submenus on iPadOS display their arrow aligned to the end of a nested menu item, because menus open to the end of their parent. I'm using a media query to test for this in my project, but passing context is the safest way to allow for customization imo.

```dart
  @override
  bool hasLeading(BuildContext context) {
    return switch (_computeMenuWidth(context)) {
      CupertinoMenuWidth.iOS || CupertinoMenuWidth.iOSAccessible => true,
      CupertinoMenuWidth.iPadOS || CupertinoMenuWidth.iPadOSAccessible => false,
    };
  }
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: I agree. It would have been a breaking change on the material builder, so I opted for a callback. The other benefit is that users can call setState from the callback, whereas passing the animation status directly to a builder would have prevented that.

examples/api/lib/cupertino/menu_anchor/menu_anchor.0.dart
davidhicks980: controller.isOpen corresponds to whether the overlay is open, meaning it is true when the animation is AnimationStatus.forward, AnimationStatus.completed, and AnimationStatus.reverse. If we just check controller.isOpen, we have to wait until the menu is AnimationStatus.dismissed before it can be opened again. isForwardOrCompleted allows the menu to toggle open while it is closing.

davidhicks980:
@MaherSafadii fixed the spacing btw!

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: What about isOpenOrOpening?

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: I added 

```dart
  /// The [builder] will rebuild when the menu's [AnimationStatus] changes.
```

Let me know if more detail is needed.

examples/api/lib/cupertino/menu_anchor/menu_anchor.1.dart
davidhicks980: We're calling setState within onAnimationStatusChanged. However, as you mentioned, this isn't actually necessary. I'm only doing it so that devs aren't confused.

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: Oh, and I try to avoid builders with more than 3 parameters. It's not very common across the framework (at least from what I've read).

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: Forgot to comment -- this is done

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3075 @@
...
+        break;
+    }
+
+    isSwiped = true;
+    return true;
+  }
+
+  @override
+  void didSwipeLeave() {
+    if (mounted) {
```
dkwingsmt: Maybe you can always set `isSwiped` to false regardless of `mounted`. Just to be safe, and it makes sense.

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3075 @@
...
+  set isPressed(bool value) {
+    if (_isPressed != value) {
+      _isPressed = value;
+      _statesController.update(WidgetState.pressed, value);
+    }
+  }
+
+  bool get isSwiped => _isSwiped;
+  bool _isSwiped = false;
+  set isSwiped(bool value) {
```
dkwingsmt, line 2397: I find this name a bit confusing. I understand that it might have come from `isHovered`. What do you think of something like `isOnSwipe`?

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3075 @@
...
+  ///  * The pointer is no longer in contact with the screen.
+  ///  * The pointer is canceled.
+  ///  * The gesture loses the arena.
+  ///  * The gesture ends. In this case, this method is called immediately
+  ///    before [didSwipeActivate].
+  void didSwipeLeave();
+
+  /// The drag gesture is completed in this region.
+  ///
+  /// This method is called immediately after a [didSwipeLeave].
```
dkwingsmt: nit: I'm leaning towards making it a flag of `didSwipeLeave`, which is basically your previous design (although I prefer a name that describes the function, such as `fromActivation`). What made you change your design?

If you'd like to keep the current design, it might make a bit more sense if `didSwipeActivate` is called before `didSwipeLeave`, since then `didSwipeLeave` can feel better to clean up without worrying about a resource that `didSwipeActivate` might want to use.

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3075 @@
...
+      }
+
+      minimumSquaredDistance = math.min(squaredDistance, minimumSquaredDistance);
+    }
+
+    final double distance = minimumSquaredDistance == 0 ? 0 : math.sqrt(minimumSquaredDistance);
+    widget.onDistanceChanged(distance);
+  }
+
+  void _handleSwipeEnd(DragEndDetails position) {
```
dkwingsmt:
```suggestion
  void _handleSwipeEnd(DragEndDetails updateDetails) {
```

dkwingsmt (commented):
I read the diff of the first 70% of `cupertino/menu_anchor.dart`. Will finish the remaining tomorrow.

dkwingsmt (dismissed):
LGTM. Thank you!

packages/flutter/lib/src/cupertino/menu_anchor.dart
chunhtai: I don't think we have guideline on the limit of builder parameters. In this case, If choosing between one more parameter in builder versus a new parameter in constructor. I think it is better to be in builder  instead.

anyway this doesn't block the merge

chunhtai (dismissed):
LGTM

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: Setting isSwiped can trigger a state update via the WidgetStatesController, so it'd cause an error if the widget wasn't mounted.

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: Hmm I also find isOnSwipe a bit confusing. I was originally using isPanned instead of isSwiped, but I'm not sure what the best nomenclature is here. I'll let you make the final call.

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: I changed the old design because I preferred your implementation on CupertinoDialog 😆. Either way works for me, although one benefit of the new design is that it decouples activation from the pointer leaving. 

The only problem I have with calling didSwipeActivate after didSwipeLeave is that didSwipeActivate logically follows didSwipeLeave, since the swipe leaving is what triggers the activation. 

Here's the code I was using:
(https://github.com/flutter/flutter/blob/bd7d9437a93b9bb5f04f2c8db948ea1e5b2ba9b1/packages/flutter/lib/src/cupertino/dialog.dart#L828-L880)

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: I agree that it feels less intrusive to add a builder parameter, but I do think the fact that the builder can't call setState in response to animation status changes could prove to be frustrating. For example, I don't think there would be a way for the user to change a menu item's appearance in response to the AnimationStatus changing:

```dart
CupertinoMenuAnchor(
  onAnimationStatusChanged: (AnimationStatus status) {
    setState(() {
      _animationStatus = status;
    });
  },
  childFocusNode: _buttonFocusNode,
  menuChildren: <Widget>[
    CupertinoMenuItem(
      onPressed: () {},
     // This wouldn't be possible if the animation status was set from the builder.
      child: _animationStatus.isForwardOrCompleted
          ? const Text('Opening Item')
          : const Text('Closing Item'),
    ),
  ],
```

As a side note, you're right that there aren't any guidelines against adding more than 3 params. It's a personal preference of mine because positional parameters don't retain documentation or their names when hovered. This can make it frustrating to track down how to use each parameter in a callback with a large definition. The example I always think of is PageRouteBuilder.transitionsBuilder:

image

dkwingsmt (dismissed):
LGTM. Thank you very much for the amazing work! 

Let's see if @chunhtai has more thoughts.

packages/flutter/lib/src/cupertino/menu_anchor.dart
chunhtai: thanks for the reply, yup this sounds good to me

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+    );
+    properties.add(
+      FlagProperty('useRootOverlay', value: useRootOverlay, ifTrue: 'uses root overlay'),
+    );
+    properties.add(DiagnosticsProperty<EdgeInsetsGeometry>('overlayPadding', overlayPadding));
+  }
+}
+
+class _CupertinoMenuAnchorState extends State<CupertinoMenuAnchor> with TickerProviderStateMixin {
+  static const Duration longPressToOpenDuration = Duration(milliseconds: 400);
```
chunhtai:
```suggestion
  static const Duration kLongPressToOpenDuration = Duration(milliseconds: 400);
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+  static const Tolerance springTolerance = Tolerance(velocity: 0.1);
```
chunhtai:
```suggestion
  static const Tolerance kSpringTolerance = Tolerance(velocity: 0.1);
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+  /// against the menu background.
+  // The following colors were measured from the iOS 17.2 simulator, and opacity was
+  // extrapolated:
+  // Dark mode on black       Color.fromRGBO(97, 97, 97)
+  // Dark mode on white       Color.fromRGBO(132, 132, 132)
+  // Light mode on black      Color.fromRGBO(147, 147, 147)
+  // Light mode on white      Color.fromRGBO(187, 187, 187)
+  //
+  // Colors were also compared atop a red, green, and blue backgrounds.
+  static const CupertinoDynamicColor overlayColor = CupertinoDynamicColor.withBrightness(
```
chunhtai:
```suggestion
  static const CupertinoDynamicColor kOverlayColor = CupertinoDynamicColor.withBrightness(
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+    color: Color.fromRGBO(140, 140, 140, 0.3),
+    darkColor: Color.fromRGBO(255, 255, 255, 0.25),
+  );
+
+  /// The default color applied to the [_CupertinoMenuImplicitDivider], atop the
+  /// [overlayColor], with [BlendMode.srcOver].
+  ///
+  /// This color is used to make the divider more opaque.
+  static const CupertinoDynamicColor color = CupertinoDynamicColor.withBrightness(
```
chunhtai:
```suggestion
  static const CupertinoDynamicColor kDividerColor = CupertinoDynamicColor.withBrightness(
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+  //
+  // Dark mode on white background     rgb(111, 111, 111)
+  // Dark mode on black                rgb(61, 61, 61)
+  // Light mode on black               rgb(177, 177, 177)
+  // Light mode on white               rgb(225, 225, 225)
+  //
+  // Blend mode is used to mimic the visual effect of the iOS
+  // menu item. As a result, the default pressed color does not match the
+  // reported colors on the iOS 18.5 simulator.
+  static const WidgetStateProperty<BoxDecoration> defaultDecoration =
```
chunhtai:
```suggestion
  static const WidgetStateProperty<BoxDecoration> kDefaultDecoration =
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+
+  static final WidgetStateProperty<MouseCursor> _defaultCursor =
+      WidgetStateProperty.resolveWith<MouseCursor>((Set<WidgetState> states) {
+        return !states.contains(WidgetState.disabled) && kIsWeb
+            ? SystemMouseCursors.click
+            : MouseCursor.defer;
+      });
+
+  // Measured from the iOS 18.5 simulator debug view.
+  static const Color _defaultTextColor = CupertinoDynamicColor.withBrightness(
```
chunhtai:
```suggestion
  static const Color _kDefaultTextColor = CupertinoDynamicColor.withBrightness(
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+    darkColor: Color.from(alpha: 0.96, red: 1, green: 1, blue: 1),
+  );
+
+  /// The default [Color] applied to a [CupertinoMenuItem]'s [subtitle]
+  /// widget, if a subtitle is provided.
+  ///
+  /// A custom blend mode is applied to the subtitle to mimic the visual effect
+  /// of the iOS menu subtitle. As a result, the defaultSubtitleStyle color does
+  /// not match the reported color on the iOS 18.5 simulator.
+  static const Color _defaultSubtitleTextColor = CupertinoDynamicColor.withBrightness(
```
chunhtai:
```suggestion
  static const Color _kDefaultSubtitleTextColor = CupertinoDynamicColor.withBrightness(
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+    color: Color.from(alpha: 0.55, red: 0, green: 0, blue: 0),
+    darkColor: Color.from(alpha: 0.4, red: 1, green: 1, blue: 1),
+  );
+
+  /// The maximum number of lines for the [child] widget when
+  /// [MediaQuery.textScalerOf] returns a [TextScaler] that is less than or
+  /// equal to 1.25.
+  ///
+  /// Measured from the iOS 18.5 simulator debug view.
+  static const int _defaultMaxLines = 2;
```
chunhtai:
```suggestion
  static const int _kDefaultMaxLines = 2;
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+
+  /// The maximum number of lines for the [child] widget when
+  /// [MediaQuery.textScalerOf] returns a [TextScaler] that is greater than
+  /// 1.25.
+  static const int _defaultAccessibilityModeMaxLines = 100;
```
chunhtai:
```suggestion
  static const int _kDefaultAccessibilityModeMaxLines = 100;
```

packages/flutter/lib/src/cupertino/menu_anchor.dart
```diff
@@ -0,0 +1,3043 @@
...
+
+/// The minimum normalized text scale factor supported on iOS.
+const double _kMinimumTextScaleFactor = 1 - 3 / _kCupertinoMobileBaseFontSize;
+
+/// The minimum normalized text scale factor supported on iOS.
+const double _kMaximumTextScaleFactor = 1 + 36 / _kCupertinoMobileBaseFontSize;
+
+// Accessibility mode on iOS is determined by the text scale factor that the
+// user has selected.
+bool _isAccessibilityModeEnabled(BuildContext context) {
```
chunhtai: This needs a better name, 
```suggestion
bool _largeTextModeEnabled(BuildContext context) {
```

chunhtai (dismissed):
LGTM, left some nitpicking

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: I agree and I'll switch it over, but we may need to change the naming from CupertinoDialog: https://github.com/flutter/flutter/blob/7f3269bf79166938527507f414cddcc5ffa91b96/packages/flutter/lib/src/cupertino/dialog.dart#L188-L194

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: The style guide says to only use the k prefix on global constants: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#begin-global-constant-names-with-prefix-k. 

I also prefer adding it to local constants, but wanted to give you a heads up prior to the change

chunhtai: sgtm, thanks!

davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

packages/flutter/lib/src/cupertino/menu_anchor.dart
davidhicks980: done

chunhtai (approved):
LGTM

dkwingsmt (approved):

dkwingsmt (approved):

PR #182042, opened 2026-02-07
Factor out unnecessary `assert`s

nate-thegrate:
The conversation in #181703 got me thinking about whether there were any `assert(value != null)` statements in our codebase that were written prior to Dart's null safety era and could be factored out.

It turns out that there are!

This PR updates a few setters so that incorrect `null` assignments are caught by static analysis instead of potentially going unnoticed until runtime.


dkwingsmt (approved):

reidbaker (approved):

loic-sharma (approved):
Nice clean up, thanks!

PR #182065, opened 2026-02-08
Hardening SPM: Model-Based Regeneration

rickhohler:
Refactors Flutter's Swift Package Manager integration to use deterministic regeneration instead of regex patching. Includes design document in packages/flutter_tools/doc/spm_hardening_design_doc.md for review. cc @vashworth


reidbaker (commented):
Android readme changes approved.

vashworth:
Thank you for your contribution. At this time this is not really the direction in which we would like to take this. The Package.swift that is being altered is generated by Flutter, so we are confident in the replacement. There does not seem to be any indication that this causing a problem for people. If you're having a problem with it, please file an issue.

PR #182093, opened 2026-02-09
Make positionInlineChildren assert much clearer

### LINKED ISSUE #182090: Misleading error message in positionInlineChildren assert
When incorrectly implementing a custom InlineSpan, I encountered an assert inside the `positionInlineChildren` function with the following message:

> The length of boxes (5) should be greater than childCount (1)

Looking only at this message, I would think my implementation is correct (5 is indeed greater than 1). However, the logic in the `positionInlineChildren` function expects the opposite: the number of boxes must NOT exceed the number of children (childCount).
```dart
  @protected
  void positionInlineChildren(List<ui.TextBox> boxes) {
    RenderBox? child = firstChild;
    for (final box in boxes) {
      if (child == null) {
        assert(
          false,
          'The length of boxes (${boxes.length}) should be greater than childCount ($childCount)',
        );
        return;
      }
      final textParentData = child.parentData! as TextParentData;
      textParentData._offset = Offset(box.left, box.top);
      child = childAfter(child);
    }
    while (child != null) {
      final textParentData = child.parentData! as TextParentData;
      textParentData._offset = null;
      child = childAfter(child);
    }
  }
```
The current message is logically incorrect and misleading for debugging.

BrainLUX:
fixes [flutter/flutter#182090](https://github.com/flutter/flutter/issues/182090)

The assertion message should clearly state the actual requirement of the code to help developers understand and fix the error.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -219,8 +219,8 @@ mixin RenderInlineChildrenContainerDefaults
     for (final box in boxes) {
       if (child == null) {
         assert(
-          false,
-          'The length of boxes (${boxes.length}) should be greater than childCount ($childCount)',
+          boxes.length <= childCount,
```
victorsanni: I'm assuming the fact that we are in this if-block means there is an additional box not accounted for.

victorsanni (approved):

packages/flutter/lib/src/rendering/paragraph.dart
BrainLUX: In fact, yes, that's true.
But explicitly specifying the condition will make debugging easier for developers.

LongCatIsLooong (approved):
LGTM but I think this may need a test, since you already have an implementation that would trigger the assert.

justinmc:
@BrainLUX Would you be able to add a test here that catches the logical change of when the error occurs?

packages/flutter/lib/src/rendering/paragraph.dart
LongCatIsLooong: I think I initially wrote the assert under the assumption that it wasn't going to be an app developer facing error message (so the error message was only useful for framework developers), but it looks like this is not the case. Was it because you didn't push placeholders in the custom InlineSpan in `InlineSpan.build`? In any case it might be helpful to expand the error message a little more to include the most probable cause (e.g. did you forget to push placeholders if you are using a custom InlineSpan subclass?)

LongCatIsLooong:
still LGTM, but maybe this should be expanded to a more detailed message, like the one over here: https://github.com/flutter/flutter/blob/1e97fd38cf89f94f5159f4e9d0b598a0871103ab/packages/flutter/lib/src/rendering/object.dart#L2224-L2241

so people ran into the same problem don't have to figure out what could have caused the assert to fire.

packages/flutter/lib/src/rendering/paragraph.dart
```diff
@@ -218,10 +218,28 @@ mixin RenderInlineChildrenContainerDefaults
     RenderBox? child = firstChild;
     for (final box in boxes) {
       if (child == null) {
-        assert(
-          false,
-          'The length of boxes (${boxes.length}) should be greater than childCount ($childCount)',
-        );
+        assert(() {
+          if (boxes.length > childCount) {
```
LongCatIsLooong: nit: I feel this is no longer needed since there is now a detailed message.

LongCatIsLooong (approved):
Still LGTM, thank you for adding the detailed error message!

packages/flutter/lib/src/rendering/paragraph.dart
BrainLUX: Agreed.

victorsanni (approved):

PR #182094, opened 2026-02-09
Update Skia logging ports: Replace SkDebug with SkLog

jbeta51:
Update Skia BUILD file names to match new SkLog name.

Skia is updating its logging tools. This includes changes that rename the SkDebug ports, so all GN's referencing them need to be updated


gaaclarke:
(fyi you're going to have to rebase)

Eliezermga (commented):
okay...

PR #182145, opened 2026-02-10
Add information to issue triage page

stuartmorgan-g:
- Adds an overview at the top about how triage works from a very high level. Right now the page discusses primary and secondary/team triage without ever actually introducing those concepts. Also sets some expectations around timelines, as we do for reviews in other docs.
- Clarifies that some sub-sections of the team assignment flow chart are additional steps rather than alternate path (especially since the intro says to stop once a team assignment matches, so it was unclear that the sub-bullets applied as well).
- Adds more discussion of common labels.


Piinks (commented):
Ah thank you, realizing there was a lot of things there were just known. Thanks for writing it down!
There might be some overlap with https://github.com/flutter/flutter/blob/master/docs/contributing/issue_hygiene/README.md#labels

Skimming through, should this page also mention:

- `found in release..`?
- `r:..` resolution labels? (e.g. fixed, invalid, duplicate)
- `browser-` for web issues that are happening on specific browsers?
- I noted `waiting for customer response`, but it looks like that is already mentioned. 👍

Piinks (approved):
LGTM to me then! Thanks!

Issue #182155, opened 2026-02-10
Automated Flutter "Prompt-to-Reload" Workflow


jwren:
This feature aims to introduce a frictionless development loop by bridging the gap between an AI agent's code modifications and the live execution environment. By leveraging the **Dart Model Context Protocol (MCP)** and the latest Flutter discovery tools, the agent moves beyond simple code generation to become a runtime-aware development partner.

---

### The Heart of the Workflow: Hot Reload
In Flutter development, **Hot Reload** is the indispensable "pulse" of the iteration cycle. It allows developers to inject updated source code files into the running Dart Virtual Machine (VM) without losing the application state or requiring a full restart. 

By automating this step, the agent eliminates the need for the developer to manually switch focus to a terminal or IDE to trigger a reload. The "prompt-to-screen" latency is reduced to near-zero, making the AI-driven development feel native to the Flutter experience.

---

### The Discovery Engine: `flutter running-apps`
The technical backbone of this automation is the newly landed `flutter running-apps` command ([PR #180098](https://github.com/flutter/flutter/pull/180098)). This provides the specific mechanism for the agent to connect to an existing session without user configuration:

* **mDNS Discovery**: The command utilizes Multicast DNS (mDNS) to scan the local network and connected devices for active Flutter applications.
* **Automatic URI Resolution**: It identifies and retrieves the **DTD URI** (Dart Tooling Daemon) or VM Service URI.
* **Zero-Config Connection**: Because the agent can programmatically call `flutter running-apps`, the developer no longer needs to find, copy, and paste debugger URIs. The agent autonomously establishes the communication link.

---

### The End-to-End Automated Loop

1. **Natural Language Prompt**: The developer provides a request (e.g., *"Refactor the search bar to use the new branding colors and center the icon"*).
2. **Agent Modification**: The agent analyzes the codebase and applies the necessary logic or UI changes.
3. **Atomic Save**: The agent automatically saves the modified files to the disk.
4. **Process Discovery**: The agent executes `flutter running-apps` to find the active Flutter process and its corresponding DTD URI. This may best be done through MCP modifications, an Agent Skill recommendation, a combination, or some other mechanism.
5. **Automated Hot Reload**: Once the connection is established, the agent immediately triggers a **Hot Reload**.
6. **Instant Verification**: The changes are reflected on the device or emulator instantly. If any runtime errors occur, the agent can use MCP tools to inspect the stack trace and propose a fix.

---

[...]

PR #182172, opened 2026-02-10
running-apps: update running-apps to use Duration.ago()

jwren:
Cleanup from https://github.com/flutter/flutter/pull/180098

https://github.com/flutter/flutter/issues/182155

### DIFF
--- a/dev/integration_tests/widget_preview_scaffold/pubspec.yaml
+++ b/dev/integration_tests/widget_preview_scaffold/pubspec.yaml
@@ -45,7 +45,7 @@ dependencies:
   term_glyph: 1.2.2
   test_api: 0.7.9
   typed_data: 1.4.0
-  unified_analytics: 8.0.10
+  unified_analytics: 8.0.11
   url_launcher_android: 6.3.28
   url_launcher_ios: 6.4.1
   url_launcher_linux: 3.2.2
--- a/dev/integration_tests/widget_preview_scaffold/pubspec.yaml
+++ b/dev/integration_tests/widget_preview_scaffold/pubspec.yaml
@@ -45,7 +45,7 @@ dependencies:
   term_glyph: 1.2.2
   test_api: 0.7.9
   typed_data: 1.4.0
-  unified_analytics: 8.0.11
+  unified_analytics: 8.0.10
   url_launcher_android: 6.3.28
   url_launcher_ios: 6.4.1
   url_launcher_linux: 3.2.2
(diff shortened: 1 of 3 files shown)


packages/flutter_tools/lib/src/commands/running_apps.dart
```diff
@@ -108,7 +108,7 @@ class RunningAppsCommand extends FlutterCommand {
       final String deviceId = app.deviceId;
       final String platform = app.targetPlatform;
       final String vmServiceUri = app.wsUri;
-      final String age = processAge(app.epoch, _systemClock);
+      final String age = _systemClock.now().difference(DateTime.fromMillisecondsSinceEpoch(app.epoch)).ago();
```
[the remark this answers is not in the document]
jwren: Ack, second precision is not important here.

dev/integration_tests/widget_preview_scaffold/pubspec.yaml
```diff
@@ -45,7 +45,7 @@ dependencies:
   term_glyph: 1.2.2
   test_api: 0.7.9
   typed_data: 1.4.0
-  unified_analytics: 8.0.10
+  unified_analytics: 8.0.11
```
bkonyi: Is there any particular dependency that needed to be updated for this change? If not, I think we can just revert the pubspec related updates.

jwren: @bkonyi I reverted the change to just that `pubspec.yaml`, but now have failures with the bots.  Leaving them red so you can see.  From the logs: ease run "flutter update-packages --force-upgrade --update-hashes" to update them correctly. The hash (fv45lk) does not match the expectation (nepob0).

bkonyi: I'm just wondering how the pubspecs got updated in the first place. You shouldn't need to run `flutter update-packages` at all for the changes you made since they don't touch any dependencies.

In general, I only run `flutter update-packages` if I'm trying to pick up a new dependency for the PR I'm working on or if I'm trying to manually unblock a packages roll. It's not something that I need to do often :-)

I'd try rolling back all the pubspec and pubspec.lock changes and uploading that. As long as the parent commit was in a good state (it should have been), that should clear things up. If you rebased, it's possible something got messed up during conflict resolution and it might be worth reapplying your changes on top of master.

bkonyi (approved):

PR #182196, opened 2026-02-11
refactor: Centralize table formatting logic into a new `formatTable` utility function.

### LINKED ISSUE #180949: Consolidate table output from `flutter devices` and `flutter running-apps`
https://github.com/flutter/flutter/pull/180098 introduces another table displayed through the flutter CLI.  The output from `flutter devices` as well as this new action, `flutter running-apps` should be consolidated to have one source of true for output of this form

jwren:
This resolves https://github.com/flutter/flutter/issues/180949

This is follow-up on https://github.com/flutter/flutter/pull/180098

### DIFF
--- a/packages/flutter_tools/lib/src/base/utils.dart
+++ b/packages/flutter_tools/lib/src/base/utils.dart
@@ -611,3 +611,30 @@ extension StackTraceTransform<T> on Stream<T> {
 final utf8LineDecoder = StreamTransformer<List<int>, String>.fromBind(
   (stream) => stream.transformWithCallSite(utf8.decoder).transform(const LineSplitter()),
 );
+
+/// Formats a list of rows into a table with aligned columns.
+///
+/// [table] is a list of rows, where each row is a list of strings.
+/// [separator] is the string used to separate columns (default is ' • ').
+///
+/// Returns a list of strings, where each string is a formatted row.
+List<String> formatTable(List<List<String>> table, {String separator = ' • ', int indent = 0}) {
+  if (table.isEmpty) {
+    return <String>[];
+  }
+
+  // Calculate column widths
+  final indices = List<int>.generate(table[0].length - 1, (int i) => i);
+  List<int> widths = indices.map<int>((int i) => 0).toList();
+  for (final row in table) {
+    widths = indices.map<int>((int i) => math.max(widths[i], row[i].length)).toList();
+  }
+
+  final String indentString = ' ' * indent;
+
+  // Join columns into lines of text
+  return <String>[
+    for (final List<String> row in table)
+      '$indentString${indices.map<String>((int i) => row[i].padRight(widths[i])).followedBy(<String>[row.last]).join(separator)}',
+  ];
+}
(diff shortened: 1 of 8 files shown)


packages/flutter_tools/lib/src/base/utils.dart
```diff
@@ -611,3 +611,30 @@ extension StackTraceTransform<T> on Stream<T> {
...
+  for (final row in table) {
+    widths = indices.map<int>((int i) => math.max(widths[i], row[i].length)).toList();
+  }
+
+  final String indentString = ' ' * indent;
+
+  // Join columns into lines of text
+  return <String>[
+    for (final List<String> row in table)
+      '$indentString${indices.map<String>((int i) => row[i].padRight(widths[i])).followedBy(<String>[row.last]).join(separator)}',
```
bkonyi: This is a bit hard to read. How about this?

```dart
  return table.map<String>((row) {
    final formatted = indices
        .map<String>((int i) => row[i].padRight(widths[i]))
        .followedBy(<String>[row.last])
        .join();
    return '$indentString$formatted';
  }).toList();
```

packages/flutter_tools/lib/src/base/utils.dart
```diff
@@ -611,3 +611,30 @@ extension StackTraceTransform<T> on Stream<T> {
...
+  if (table.isEmpty) {
+    return <String>[];
+  }
+
+  // Calculate column widths
+  final indices = List<int>.generate(table[0].length - 1, (int i) => i);
+  List<int> widths = indices.map<int>((int i) => 0).toList();
+  for (final row in table) {
+    widths = indices.map<int>((int i) => math.max(widths[i], row[i].length)).toList();
+  }
```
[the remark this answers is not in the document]
bkonyi: We should probably throw if the first table entry is empty since it's the header. Otherwise, this comment seems reasonable.

packages/flutter_tools/test/general.shard/utils_test.dart
```diff
@@ -568,4 +568,42 @@ needs to be wrapped.
...
+
+  testWithoutContext('formatTable', () {
+    expect(formatTable(<List<String>>[]), isEmpty);
+
+    expect(
+      formatTable(<List<String>>[
+        <String>['Col1', 'Col2'],
+        <String>['Value1', 'Value2'],
+      ]),
+      <String>['Col1   • Col2', 'Value1 • Value2'],
```
bkonyi: Is it maybe worth joining the result and verifying against the resulting formatted string? Visually, it'll be easier to tell that the formatting is correct:

```dart
const expected = '''
Col1   • Col2
Value1 • Value2
''';

expect(
  formatTable(<List<String>>[
    <String>['Col1', 'Col2'],
    <String>['Value1', 'Value2'],
  ]).join(Platform.lineTerminator),
  expected,
);
```

packages/flutter_tools/lib/src/base/utils.dart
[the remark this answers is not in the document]
jwren: done

packages/flutter_tools/lib/src/base/utils.dart
jwren: done

packages/flutter_tools/test/general.shard/utils_test.dart
jwren: done

bkonyi (approved):
LGTM once analysis check is passing. Thanks for coming back to this!

PR #182200, opened 2026-02-11
Preparation to add contentTextStyle flag to SimpleDialog.

chingjun:
This commit adds a `contentTextStyle` flag to `SimpleDialog` to make it easier to land the internal fixes needed to PR #178824.

The flag currently does nothing. Actual implementation will be landed in PR #178824.


QuncCccccc (approved):
LGTM! Thanks!

PR #178383, opened 2025-11-12
[Android] Add integration test for setting shell arguments in release mode

camsim99:
> [!NOTE]  
> 🚧 WIP. Intended to increase test coverage of https://github.com/flutter/flutter/pull/177127. Note that this PR currently contains changes from that PR.

Adds integration test (see `dev/devicelab/lib/tasks/android_flutter_shell_args_test.dart`) for setting Android Flutter shell arguments for apps running in release mode.

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #182241, opened 2026-02-11
[Android] Add integration test for setting engine flags via the manifest

camsim99:
Adds integration tests for setting engine flags via the manifest for Flutter Android apps: https://github.com/flutter/flutter/pull/182522. Technically, a continuation of https://github.com/flutter/flutter/pull/178383.

Additionally:

1. Wires up `--test-flag` added in https://github.com/flutter/flutter/pull/182522 so I can test command line flag specification taking precedence over manifest flag specification in debug mode only.
2. Fixes documentation/test bug of my previous claim that developers do not need to specify a manifest metadata value for keys/flags that do not take arguments in the format of `--flag=value` or `--flag=value1,value2,...`. This is false; metadata requires a key-value or key-resource pair (see [documentation proof](https://developer.android.com/guide/topics/manifest/meta-data-element#:~:text=receiver%3E%0A%3Cservice%3E-,description%3A,-A%20name%2Dvalue)).
3. Refactors devicelab util method for adding metadata to the app manifest to a central location.

## Pre-launch Checklist

- [x] I signed the [CLA].


dev/devicelab/lib/tasks/android_engine_flags_test.dart
```diff
@@ -0,0 +1,259 @@
...
+    } on TaskResult catch (taskResult) {
+      return taskResult;
+    } catch (e, stackTrace) {
+      print('Task exception stack trace:\n$stackTrace');
+      return TaskResult.failure(e.toString());
+    } finally {
+      rmTree(tempDir);
+    }
+  };
+}
```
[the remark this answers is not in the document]
camsim99: I personally think this makes it a bit harder to parse what is happening in each test, but I don't have a strong preference if a reviewer prefers a refactor like this.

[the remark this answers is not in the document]
reidbaker: trust your gut over a bot.  I would think of it this way. If you had to add a test because of a regression you missed would you need to modify every test. If you needed to modify every test and that new code didnt make sense when reviewing a single test then it is a good sign to pull out shared logic.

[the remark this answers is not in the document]
camsim99: Great general advice! I think this would be a refactor only for readability. I went ahead and pulled shared logic into a helper method, but left the rest the same because I don't love the readability of this suggested refactor.

packages/flutter_tools/lib/src/runner/flutter_command.dart
```diff
@@ -1330,6 +1330,14 @@ abstract class FlutterCommand extends Command<void> {
     );
   }
 
+  void addTestFlag({required bool verboseHelp}) {
+    argParser.addFlag(
+      'test-flag',
+      hide: !verboseHelp,
+      help: 'No-op flag for testing purposes; use for testing only.',
```
reidbaker: nit:
```suggestion
      help: 'No-op flag for testing purposes; use for testing flag priority only.',
```

engine/src/flutter/shell/platform/android/test/io/flutter/embedding/engine/loader/FlutterLoaderTest.java
```diff
@@ -784,7 +784,7 @@ public void getSofwareRenderingEnabledViaManifest_returnsExpectedValueWhenSetVia
   public void itSetsSkiaDeterministicRenderingFromMetadata() {
     testFlagFromMetadataPresent(
         "io.flutter.embedding.android.SkiaDeterministicRendering",
-        null,
+        true,
```
reidbaker: instead of setting the default here is there a place where this test could reference some flagObject.defaultValue?

reidbaker (dismissed):

dev/devicelab/lib/framework/android_manifest_utils.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:path/path.dart' as path;
+import 'package:xml/xml.dart';
+
```
reidbaker, line 9: Please give this file a dart doc that includes the scope of "utils" that should be included. Also consider giving this a more specific name to match the scope of what utils belong here.

dev/devicelab/lib/framework/android_utils.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:path/path.dart' as path;
+import 'package:xml/xml.dart';
+
+void addMetadataToManifest(String testDirectory, List<(String, String)> keyPairs) {
```
reidbaker: Give a java doc for assumptions. Also consider making `testDirectory` an object that is a directory instead of a string.

dev/devicelab/lib/tasks/android_engine_flags_test.dart
```diff
@@ -0,0 +1,250 @@
...
+            }
+          });
+
+      section('Check that warning log for invalid AOT shared library name is in STDOUT');
+      final Object result = await Future.any(<Future<Object>>[
+        foundInvalidAotLibraryLog.future,
+        run.exitCode,
+      ]);
+
+      if (result is int) {
```
reidbaker, line 102: Why does this not fail on 0. I would think that 0 is an int and the result code 0 means everything is ok.

reidbaker (dismissed):

dev/devicelab/lib/tasks/android_engine_flags_test.dart
camsim99: If the run completes before the test finishes (i.e. before the logs we expect to be found or found), we should consider that a test failure.

dev/devicelab/lib/framework/android_utils.dart
camsim99: Added a doc but kept the String directory because that would require a larger refactor beyond the scope of this PR that I don't think is worth it, but if you feel strongly, I'll give it a go.

dev/devicelab/lib/framework/android_manifest_utils.dart
camsim99: I renamed the file to be clearer and added a doc for the method instead since I don't have a class here.

packages/flutter_tools/lib/src/device.dart
```diff
@@ -1027,7 +1026,6 @@ class DebuggingOptions {
     this.usingCISystem = false,
     this.debugLogsDirectoryPath,
     this.webDevServerConfig,
```
reidbaker, line 1027: Was this an intentional change?

packages/flutter_tools/lib/src/commands/run.dart
reidbaker, line 257: Same here?

reidbaker (dismissed):

mboetger (dismissed):

packages/flutter_tools/lib/src/commands/run.dart
camsim99: Whatever you're seeing must be outdated!

mboetger (dismissed):

mboetger (approved):

PR #182246, opened 2026-02-11
Prompt plugin authors to add Swift Package Manager compatibility to their plugin

### LINKED ISSUE #148222: Prompt plugin authors to add Swift Package Manager compatibility to their plugin
When detected that a flutter project is a plugin example app that has a podspec but doesn't have a Package.swift, prompt the user to add Swift Package Manager compatibility and point to documentation. 

https://github.com/flutter/flutter/blob/708fe97f9e3a0fd0c176305c5f63a506d6c78777/packages/flutter_tools/lib/src/plugins.dart#L392-L418
https://github.com/flutter/flutter/blob/708fe97f9e3a0fd0c176305c5f63a506d6c78777/packages/flutter_tools/lib/src/project.dart#L269
https://github.com/flutter/flutter/blob/708fe97f9e3a0fd0c176305c5f63a506d6c78777/packages/flutter_tools/lib/src/project.dart#L253-L257
https://github.com/flutter/flutter/blob/708fe97f9e3a0fd0c176305c5f63a506d6c78777/packages/flutter_tools/lib/src/project.dart#L263

okorohelijah:
Prompt plugin authors to add Swift Package Manager compatibility to their plugin

Fixes #148222

## Pre-launch Checklist

- [x] I signed the [CLA].


vashworth (commented):
Can you move all logic into `DarwinDependencyManagement`?

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
vashworth, line 1: I think this could be simplified by adding the warning to `_evaluatePluginsAndPrintWarnings`, which already checks if the plugin is SwiftPM/CocoaPods compatible.

https://github.com/flutter/flutter/blob/64866862f623ceeb45fd8be4782e8db8b58910c0/packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart#L143-L172

I'd do something like

```dart
Future<({int totalCount, int swiftPackageCount, int podCount})> _evaluatePluginsAndPrintWarnings({
   ...
  }) async {
    String? pluginFromExampleApp = _loadPluginFromExampleProject(...);
    for (final Plugin plugin in _plugins) {
      ...
       if (!swiftPackageManagerCompatible && !cocoaPodsCompatible) { 
         continue; 
       } 
      if (plugin.name == pluginFromExampleApp) {
          if (swiftPackageManagerCompatible && !_hasFlutterFrameworkDependency()) {
            _logger.printWarning(...)
          } else if (!swiftPackageManagerCompatible) {
            _logger.printWarning(...)
          }
      }
    }
}
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+        fileSystem.file(packageSwiftPath),
+      );
+      if (!hasFlutterFrameworkDependency) {
+        messages.add(
+          'Plugin ${plugin.name} has a Package.swift for $platform but is missing a dependency '
+          'on FlutterFramework. Add the following to your Package.swift dependencies:\n'
+          '    .package(name: "FlutterFramework", path: "../FlutterFramework")\n'
+          'And add FlutterFramework as a target dependency:\n'
+          '    .product(name: "FlutterFramework", package: "FlutterFramework")\n'
+          'See $kSwiftPackageManagerDocsUrl for more information.',
```
vashworth: Let's wait for https://github.com/flutter/website/pull/12979 to land, so we can point to the docs

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
okorohelijah: `_evaluatePluginsAndPrintWarnings` operates on the app's dependency plugins during every build and targets app developers while the validation only fires for plugin example apps, targets the parent plugin being developed, and does a deeper content analysis of the Package.swift (checking for FlutterFramework dependency declarations while filtering out comments). I thought merging them could be be a mix of 2 different things so I kept them separate for clarity wdyt?

hellohuanlin: If you prefer to separate the call path, maybe worth creating a common helper function to be called by both path?

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -129,6 +129,9 @@ class DarwinDependencyManagement {
     );
 
     _analytics.send(event);
+
+    // Validate Swift Package Manager support for plugin example apps and print warning if it doesn't support Swift Package Manager.
+    _validatePluginSwiftPackageManagerSupport(platform: platform);
```
hellohuanlin: nit: it's unclear from the function name that this is for plugin's example app, rather than a regular app. How about something like `_validatePluginSwiftPMSupportViaPluginExampleApp`

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
 
     return '';
   }
+
+  /// Tracks which plugin/platform combinations have already been warned about
+  /// during this session to avoid duplicate warnings.
+  static final Set<String> _spmValidationWarningsShown = <String>{};
```
hellohuanlin: 1. Worth commenting how duplicate can happen. (i'm also curious)
2. why static? 
3. let's be consistent and pick one among SPM vs SwiftPM vs SwiftPackageManager naming

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+
+    final Plugin? parentPlugin = _plugins
+        .where((Plugin p) => p.name == parentProject.manifest.appName)
+        .firstOrNull;
+
+    if (parentPlugin == null) {
+      return;
+    }
+
+    final cacheKey = '${parentPlugin.name}:${platform.name}';
```
hellohuanlin: uber nit: `cacheKey` sounds like you are fetching something from the cache. How about something like `dedupKey`

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+      hasFlutterFrameworkDependency: hasFlutterFrameworkDependency,
+      validationMessages: messages,
+    );
+  }
+
+  /// Checks if a Package.swift file contains a dependency on FlutterFramework.
+  ///
+  /// This looks for common patterns used to declare a FlutterFramework dependency:
+  /// - `.package(name: "FlutterFramework", path: "../FlutterFramework")` - package dependency
+  /// - `.product(name: "FlutterFramework", package: "FlutterFramework")` - target dependency
```
hellohuanlin: I'm not familiar with SwiftPM, but can target dependency be just a string (if no naming conflict)? Something like: 

```
  dependencies: [
    "FlutterFramework",
  ]
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+
+/// The URL for documentation on adding Swift Package Manager support to a plugin.
+const String kSwiftPackageManagerDocsUrl =
+    'https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors';
+
+/// The result of validating a plugin's Swift Package Manager compatibility.
+class SwiftPackageManagerPluginValidationResult {
+  SwiftPackageManagerPluginValidationResult({
+    required this.hasPodspec,
+    required this.hasPackageSwift,
```
hellohuanlin: Nit: I assume this refers to "Package.swift" file? How about `hasSwiftPMManifest`?

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+    required this.hasFlutterFrameworkDependency,
+    this.validationMessages = const <String>[],
+  });
+
+  final bool hasPodspec;
+  final bool hasPackageSwift;
+  final bool hasFlutterFrameworkDependency;
+  final List<String> validationMessages;
+
+  bool get isFullyCompatible => hasPackageSwift && hasFlutterFrameworkDependency;
```
hellohuanlin: what does "fully compatible" mean? (and what is "partially" compatible? the ones without 
"FlutterFramework" dependency?

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
vashworth: I agree with @hellohuanlin. If we're going to keep them separate, let's move the common logic into a helper function

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
vashworth: Or `_validateExampleAppPluginSupportsSwiftPackageManager`

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
vashworth: I don't think this is needed

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+  /// Tracks which plugin/platform combinations have already been warned about
+  /// during this session to avoid duplicate warnings.
+  static final Set<String> _spmValidationWarningsShown = <String>{};
+
+  /// Validates Swift Package Manager support for plugins in the current project.
+  ///
+  /// Checks if the current project is a plugin example app and validates
+  /// the parent plugin's Swift Package Manager compatibility:
+  /// 1. If the plugin has a podspec but no Package.swift, prompts the user to
+  ///    add SPM support.
```
vashworth:
```suggestion
  ///    add SwiftPM support.
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
 
     return '';
   }
+
+  /// Tracks which plugin/platform combinations have already been warned about
+  /// during this session to avoid duplicate warnings.
+  static final Set<String> _spmValidationWarningsShown = <String>{};
+
+  /// Validates Swift Package Manager support for plugins in the current project.
```
vashworth:
```suggestion
  /// Validates Swift Package Manager support for the plugin when building it's example app.
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+  ///
+  /// Checks if the current project is a plugin example app and validates
+  /// the parent plugin's Swift Package Manager compatibility:
+  /// 1. If the plugin has a podspec but no Package.swift, prompts the user to
+  ///    add SPM support.
+  /// 2. If the plugin has a Package.swift, validates that it has a dependency
+  ///    on FlutterFramework.
+  ///
+  /// Warnings are printed to inform plugin authors about SPM compatibility issues.
```
vashworth:
```suggestion
  /// Warnings are printed to inform plugin authors about SwiftPM compatibility issues.
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+      return;
+    }
+
+    final Plugin? parentPlugin = _plugins
+        .where((Plugin p) => p.name == parentProject.manifest.appName)
+        .firstOrNull;
+
+    if (parentPlugin == null) {
+      return;
+    }
```
vashworth, line 347: I'd move this logic into it's own method so it's clear what it's doing. For example, something like 
```dart
Plugin? pluginFromExampleApp = _loadPluginFromExampleProject(...);
```

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+        validatePluginSwiftPackageManagerSupport(
+          parentPlugin,
+          fileSystem: _fileSystem,
+          platform: platform.name,
+        );
+
+    if (result.validationMessages.isNotEmpty) {
+      _spmValidationWarningsShown.add(cacheKey);
+      result.validationMessages.forEach(_logger.printWarning);
+    }
```
vashworth: `validationMessages` does not need to be a List, it can just be a string since there is only ever one

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+    if (parentPlugin == null) {
+      return;
+    }
+
+    final cacheKey = '${parentPlugin.name}:${platform.name}';
+    if (_spmValidationWarningsShown.contains(cacheKey)) {
+      return;
+    }
+
+    final SwiftPackageManagerPluginValidationResult result =
```
vashworth: Nothing out of this result is used except the `validationMessages`. Instead of returning a class, just return a String

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
```diff
@@ -259,4 +262,177 @@ class DarwinDependencyManagement {
...
+
+      // Remove both single-line (//) and block (/* ... */) comments.
+      final String uncommentedContents = contents.replaceAll(RegExp(r'//.*|/\*[\s\S]*?\*/'), '');
+
+      final bool hasPackageDependency = uncommentedContents.contains(
+        RegExp(r'\.package\s*\(\s*name\s*:\s*"FlutterFramework"'),
+      );
+      final bool hasTargetDependency = uncommentedContents.contains(
+        RegExp(r'\.product\s*\(\s*name\s*:\s*"FlutterFramework"'),
+      );
```
vashworth: I'd rather make this check overly generous rather than have potential false positives. Let's just check if the Package.swift contains the string `FlutterFramework`.

vashworth (approved):
LGTM

packages/flutter_tools/lib/src/macos/darwin_dependency_management.dart
hellohuanlin: This seems to be outstanding? @vashworth is this a possible case that we should watch out?

Issue #168288, opened 2025-05-04
Support MediaQuery's highContrast property on all platforms


12people:
### Use case

Right now, based on the [official Flutter docs](https://api.flutter.dev/flutter/widgets/MediaQueryData/highContrast.html), the `MediaQuery.highContrastOf(context)` property is available only for iOS.

Having tested it on macOS, this property seems to always return false there, even though the two platforms are quite similar. Based on the documentation, I assume all platforms outside of iOS also always return false.

### Proposal

It'd be great if Flutter recognized the system's high contrast settings on all supported platforms, including Linux, macOS, and Windows.

justinmc:
I agree that Flutter should support this on all platforms where it's possible. Do we know for sure which platforms this setting is supported on? Looks like @12people you've checked macOS?

12people:
Yup, only tried macOS so far.

PR #182263, opened 2026-02-11
Add support for high contrast and color inversion on Android

xxxOVALxxx:
This PR implements Android support for two previously unsupported accessibility features in `AccessibilityBridge`

Part of #168288

On Android, unlike iOS, color contrast can range from -1.0 to 1.0. The current implementation of `AccessibilityBridge` only allows communication using bitmasks. In the future, this implementation will need to be changed to allow other types of data to be sent (And probably modifications to the corresponding ColorScheme).

In the video, you can also see that color inversion is enabled with a delay (about 10 seconds). I tested it on all my devices and created an empty project on Jetpack Compose for testing, and it turned on with a delay everywhere. The only thing I found was that older versions of the Android API do not have this problem. I have not determined exactly which version this problem starts with, but API 28 and 24 do not have this problem.

[output.webm](https://github.com/user-attachments/assets/f2feba77-f04e-465c-8527-8c05e0a7a19f)

<details><summary>Code Example</summary>
<p>

```

void main() {
  runApp(const AccessibilityTestApp());
}

class AccessibilityTestApp extends StatelessWidget {
  const AccessibilityTestApp({super.key});

  @override
    return MaterialApp(
      highContrastTheme: ThemeData(
        colorScheme: const ColorScheme.highContrastLight(
          primary: Colors.black,
          secondary: Colors.red,
        ),
      ),
      home: const TestScreen(),
    );
  }
}

class TestScreen extends StatelessWidget {
  const TestScreen({super.key});

  @override
    final mediaQuery = MediaQuery.of(context);
    final bool isHighContrast = mediaQuery.highContrast;
    final bool isInverted = mediaQuery.invertColors;

    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.primaryContainer,
      ),
      body: SingleChildScrollView(
        child: Padding(
          padding: const EdgeInsets.all(20.0),
          child: Column(
            children: [
              _StatusCard(
                title: 'High Contrast',
                isActive: isHighContrast,
                activeColor: Colors.green,
                icon: Icons.contrast,
              ),

              const SizedBox(height: 20),

              _StatusCard(
                title: 'Invert Colors',
                isActive: isInverted,
                activeColor: Colors.green,
                icon: Icons.invert_colors,
              ),

              const SizedBox(height: 20),

              Image.network('https://picsum.photos/400/200', fit: BoxFit.cover),

              const SizedBox(height: 20),

              ElevatedButton(onPressed: () {}, child: const Text('Button')),
            ],
          ),
        ),
      ),
    );
  }
}

class _StatusCard extends StatelessWidget {
  final String title;
  final bool isActive;
  final Color activeColor;
  final IconData icon;

  const _StatusCard({
    required this.title,
    required this.isActive,
    required this.activeColor,
    required this.icon,
  });

  @override
    return Card(
      child: ListPadding(
        child: ListTile(
          leading: Icon(
            icon,
            color: isActive ? activeColor : Colors.grey,
            size: 32,
          ),
          title: Text(
            title,
            style: const TextStyle(fontWeight: FontWeight.bold),
          ),
          trailing: Icon(
            isActive ? Icons.check_circle : Icons.remove_circle_outline,
            color: isActive ? activeColor : Colors.grey,
          ),
        ),
      ),
    );
  }
}

class ListPadding extends StatelessWidget {
  final Widget child;
  const ListPadding({super.key, required this.child});
  @override
  Widget build(BuildContext context) =>
      Padding(padding: const EdgeInsets.symmetric(vertical: 8), child: child);
}
```

</p>
</details> 

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -1446,6 +1449,168 @@ public void itSetsBoldTextFlagCorrectly() {
...
+
+    Settings.Secure.putInt(
+        spyContentResolver, Settings.Secure.ACCESSIBILITY_DISPLAY_INVERSION_ENABLED, 1);
+    observer.onChange(false);
+
+    ArgumentCaptor<Integer> featuresCaptor = ArgumentCaptor.forClass(Integer.class);
+    verify(mockChannel, atLeastOnce()).setAccessibilityFeatures(featuresCaptor.capture());
+    assertEquals(
+        ACCESSIBILITY_FEATURE_INVERT_COLORS,
+        (featuresCaptor.getValue() & ACCESSIBILITY_FEATURE_INVERT_COLORS));
```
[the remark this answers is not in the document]
xxxOVALxxx: Fixed it

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -1446,6 +1449,168 @@ public void itSetsBoldTextFlagCorrectly() {
...
+    reset(mockChannel);
+
+    Settings.Secure.putInt(
+        spyContentResolver, Settings.Secure.ACCESSIBILITY_DISPLAY_INVERSION_ENABLED, 0);
+    observer.onChange(false);
+
+    verify(mockChannel, atLeastOnce()).setAccessibilityFeatures(featuresCaptor.capture());
+    assertNotEquals(
+        ACCESSIBILITY_FEATURE_INVERT_COLORS,
+        (featuresCaptor.getValue() & ACCESSIBILITY_FEATURE_INVERT_COLORS));
```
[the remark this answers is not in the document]
xxxOVALxxx: Fixed it

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -1446,6 +1449,168 @@ public void itSetsBoldTextFlagCorrectly() {
...
+    reset(mockChannel);
+
+    when(mockUiModeManager.getContrast()).thenReturn(0.5f);
+    listener.onContrastChanged(0.5f);
+
+    ArgumentCaptor<Integer> featuresCaptor = ArgumentCaptor.forClass(Integer.class);
+    verify(mockChannel, atLeastOnce()).setAccessibilityFeatures(featuresCaptor.capture());
+    assertTrue(
+        (featuresCaptor.getValue() & ACCESSIBILITY_FEATURE_HIGH_CONTRAST)
+            == ACCESSIBILITY_FEATURE_HIGH_CONTRAST);
```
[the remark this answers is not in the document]
xxxOVALxxx: Fixed it

engine/src/flutter/shell/platform/android/test/io/flutter/view/AccessibilityBridgeTest.java
```diff
@@ -1446,6 +1449,168 @@ public void itSetsBoldTextFlagCorrectly() {
...
+
+    reset(mockChannel);
+
+    when(mockUiModeManager.getContrast()).thenReturn(0.0f);
+    listener.onContrastChanged(0.0f);
+
+    verify(mockChannel, atLeastOnce()).setAccessibilityFeatures(featuresCaptor.capture());
+    assertNotEquals(
+        ACCESSIBILITY_FEATURE_HIGH_CONTRAST,
+        (featuresCaptor.getValue() & ACCESSIBILITY_FEATURE_HIGH_CONTRAST));
```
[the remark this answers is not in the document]
xxxOVALxxx: Fixed it

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -664,10 +708,50 @@ private void setBoldTextFlag() {
...
+    Context context = rootAccessibilityView.getContext();
+    UiModeManager uiModeManager = (UiModeManager) context.getSystemService(Context.UI_MODE_SERVICE);
+
+    if (uiModeManager == null) {
+      updateAccessibilityFeature(AccessibilityFeature.HIGH_CONTRAST, false);
+      return;
+    }
+
+    float uiContrast = uiModeManager.getContrast();
+    // 0.0 (standard), 0.5 (medium), 1.0 (high)
```
chunhtai, line 750: is the value a slider in android setting?

looks like we need a new API in framework to accomodate this, something like PlatformDispatcher.colorContrastScale.

I assume the 1.0 is the same as highcontrast in iOS?

chunhtai (commented):
reusing AccessibilityFeature is fine for now, but I think we do need a property for this setting sooner or later

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
xxxOVALxxx: I agree that in the future we should move the contrast value into a separate API. I’ll leave a TODO.

The contrast value can range from -1.0 to 1.0 [link](https://developer.android.com/reference/android/app/UiModeManager#getContrast()) I don’t fully understand how to obtain a value below zero. Possibly this will be added in future versions of Android.

In stock Android, the contrast selection is represented by three values. However, in different Android skins, the way the value is selected may be different

Screenshot 2026-02-18 at 15 40 37


On iOS, contrast is represented as a boolean value

xxxOVALxxx: And we already have `highContrastTheme`. In the case of a floating contrast value, we will need to modify this slightly.

Maybe we could do this:
`highContrastTheme: (double contrast) => ThemeData()`

Or merge `theme` and `highContrastTheme` into a single parameter:
`theme: (double contrast) => ThemeData()`

Or come up with a more elegant solution that doesn't break the existing API and takes negative contrast values ​​into account

mboetger: You could probably set the contrast to -1 via adb, but I think that would correlate to 'low contrast' - which android doesn't seem to do right now.  OEMs could probably set it differently.  

For now, since we only have the on/off switch for high contrast, this seems like the way to go.  Please create an issue to track the todo (if you haven't already and put that in the comment instead of your username).

xxxOVALxxx: Thanks for the advice. I created an issue and updated the TODO: https://github.com/flutter/flutter/issues/182863

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
```diff
@@ -422,36 +415,17 @@ public void onAccessibilityStateChanged(boolean accessibilityEnabled) {
   private final AccessibilityManager.TouchExplorationStateChangeListener
       touchExplorationStateChangeListener;
 
+  // Listener that is notified when the high contrast mode is turned on/off.
+  // This is a UiModeManager.ContrastChangeListener on API 34+, null otherwise.
+  // Type is Object to avoid requiring API 34 for the field declaration.
+  private final Object highContrastObserver;
```
mboetger: Not a fan of setting this to Object.  I'd rather see something like this: 


```suggestion
  private interface ContrastChangedListener {
    void onContrastChanged(float contrast);
  }

  private final ContrastChangedListener highContrastObserver =
          (Build.VERSION.SDK_INT >= API_LEVELS.API_34) ?
                  (float contrast) -> setHighContrastFlag() :
                  (contrast) -> { /* no-op */};
```

mboetger (changes requested):

engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java
xxxOVALxxx: Yes, I think so too. Thank you for suggestion

mboetger (dismissed):
LGTM

chunhtai (dismissed):
LGTM

mboetger (approved):

camsim99 (approved):

chunhtai (approved):
LGTM

justinmc:
Looks like there are Google test failures here.

reidbaker:
@justinmc fyi @camsim99 went to make a g3 fix for this and it quickly grew beyond the time she had allocated. We might need to change this pr behavior in order to actually get it landed or maybe there are superpowers someplace to approve a lot of scuba diffs.

camsim99:
I assume go/lssc is the path forward here, I just don't have the bandwidth currently.

chunhtai:
There should be a fusion view that let you triage all scubas diff at once. also scuba doesn't need g3 fix, they will auto generate upon g3 roll

chunhtai:
also looking at the code change, I won't expect scuba change, do you have link the the failures? the frob entry seems to be dead https://frob.corp.google.com/#/pr/182263

PR #182265, opened 2026-02-11
[Impeller] Dispose thread local caches on each frame when using GPUSurfaceVulkanImpeller with a delegate

jason-simmons:
The Impeller Vulkan back end caches command pools in thread-local storage.  These caches can grow unbounded unless the embedder calls ContextVK::DisposeThreadLocalCachedResources()

If the GPUSurfaceVulkanImpeller does not have a delegate, then AcquireFrame will invoke DisposeThreadLocalCachedResources through its call to SurfaceContextVK::AcquireNextSurface.

With a delegate, AcquireFrame follows a different code path that must explicitly call DisposeThreadLocalCachedResources.

### DIFF
--- a/engine/src/flutter/shell/gpu/BUILD.gn
+++ b/engine/src/flutter/shell/gpu/BUILD.gn
@@ -74,7 +74,7 @@ source_set("gpu_surface_vulkan") {
   }
 }
 
-if (impeller_enable_vulkan) {
+if (test_enable_vulkan) {
   impeller_component("gpu_surface_vulkan_unittests") {
     testonly = true
 
--- a/engine/src/flutter/shell/gpu/BUILD.gn
+++ b/engine/src/flutter/shell/gpu/BUILD.gn
@@ -78,7 +78,6 @@ if (test_enable_vulkan) {
   impeller_component("gpu_surface_vulkan_unittests") {
     testonly = true
 
-    target_type = "executable"
     sources = [ "gpu_surface_vulkan_impeller_unittests.cc" ]
 
     deps = [
@@ -127,7 +126,6 @@ if (is_mac) {
     cflags_objc = flutter_cflags_objc
     cflags_objcc = flutter_cflags_objcc
 
-    target_type = "executable"
     sources = [
       "gpu_surface_metal_impeller_unittests.mm",
       "gpu_surface_noop_unittests.mm",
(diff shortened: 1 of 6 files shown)


engine/src/flutter/shell/gpu/BUILD.gn
```diff
@@ -74,6 +74,26 @@ source_set("gpu_surface_vulkan") {
   }
 }
 
+if (impeller_enable_vulkan) {
```
gaaclarke: Please add these tests as part of impeller_unittests instead of making a new executable.

jason-simmons:
Noticed this memory leak while looking at https://github.com/flutter/flutter/pull/181966 / https://github.com/flutter/flutter/issues/181967

cc @RickyvdBerg - if you are using the Impeller/Vulkan back end with the embedder and `GPUSurfaceVulkanImpeller` delegates, then you will also want this fix.

gaaclarke (changes requested):
Code looks good and test looks good.  I'd just like to see this incorporated as part of impeller_unittests instead of adding yet another executable.

engine/src/flutter/shell/gpu/BUILD.gn
jason-simmons: Impeller is mostly independent of the rest of Flutter.

I'd prefer to avoid having the `impeller_unittests` binary incorporate code from the `flutter/shell/gpu` tree.

This PR adds a `shell/gpu` test binary for Vulkan similar to the existing one for Metal.

gaaclarke: Okay, not adding it to impeller_unittests sounds good.  Can we add this test to the existing metal one then?

jason-simmons: `gpu_surface_metal_unittests` in https://github.com/flutter/flutter/blob/master/engine/src/flutter/shell/gpu/BUILD.gn is specific to Apple platforms.

The Vulkan test should be buildable on the platforms that support Vulkan/SwiftShader.

gaaclarke: Right, I'm thinking we do like what we do in impeller_unittests where we only link in tests that are applicable to the testing platform.  So the executable will be `gpu_surface_unittests`, if you are building for macos you will build and link in the metal tests, on linux you won't.  We can have TEST_P tests if we want the same test across multiple backens.

jason-simmons: Merged the `shell/gpu` test binaries - PTAL

gaaclarke (approved):
Thanks jason!  Our testing setup for the engine is pretty complicated, I'm operating under the theory that the less testing executables there are the easier it is for people to find and execute tests.  When there is separate binaries too it's easier for them to drift too.  Hopefully his helps.

PR #182268, opened 2026-02-12
[ Tool ] Allow creation of projects inside the engine's examples directory

jason-simmons:
engine/src/flutter/examples contains sample code for the engine embedder library.  The examples include scripts that create and build Flutter projects that are run through the embedder.


bkonyi (approved):
LGTM overall.

Issue #174785, opened 2025-09-01
Visibility widget still processes and renders child when visible=false


PeacefulBlack:
### Steps to reproduce

I've noticed that when using the `Visibility` widget to control the visibility of a child widget that contains dynamic content/animations, the child widget continues to be processed and rendered even when `visible` is set to `false`.

This was observed by devtools  `performance`  tab, in where ui/raster thread continue to execute despite the widget being invisible.

<img width="2928" height="1590" alt="Image" src="https://github.com/user-attachments/assets/3f8bdb80-f933-4e25-8eec-4b4e8a2ef457" />

It can be reproduced using `code sample`

Is this behavior intentional? If so, what are the design considerations behind keeping invisible widgets in the rendering pipeline?

If it's intentional, would it be possible to add a parameter to Visibility that allows complete exclusion from rendering when invisible?

### Expected results

When `visible` is `false`, I would expect the child widget to be completely excluded from the rendering pipeline to save resources, especially for animated or frequently updated widgets.

### Actual results

The child widget continues to be paint despite being invisible.

### Code sample

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Animated Visibility Demo',
      theme: ThemeData(primarySwatch: Colors.blue),
      home: const AnimatedVisibilityDemo(),
    );
  }
}

class AnimatedVisibilityDemo extends StatefulWidget {
  const AnimatedVisibilityDemo({super.key});

  @override
  State<AnimatedVisibilityDemo> createState() => _AnimatedVisibilityDemoState();
}

class _AnimatedVisibilityDemoState extends State<AnimatedVisibilityDemo> with SingleTickerProviderStateMixin {
  bool _isVisible = true;
  late AnimationController _animationController;
  late Animation<double> _sizeAnimation;

  @override
  void initState() {
    super.initState();
    _animationController = AnimationController(
      vsync: this,
      duration: const Duration(seconds: 1),
    )..repeat(reverse: true);

    _sizeAnimation = CurvedAnimation(
      parent: _animationController,
      curve: Curves.easeInOut,
    ).drive(Tween<double>(begin: 100, end: 200));
  }

  @override
  void dispose() {
    _animationController.dispose();
    super.dispose();
  }

  void _toggleVisibility() {
    setState(() {
[...]

PR #182270, opened 2026-02-12
Add pauseWhenNoListeners optimization to AnimationController

zionjo89757:
Add an optional optimization to AnimationController that pauses frame scheduling when there are no listeners.

Towards https://github.com/flutter/flutter/issues/174785
Add an optional optimization to AnimationController that pauses frame scheduling when there are no listeners.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/animation/animation_controller.dart
+++ b/packages/flutter/lib/src/animation/animation_controller.dart
@@ -242,6 +242,11 @@ class AnimationController extends Animation<double>
   /// * `vsync` is the required [TickerProvider] for the current context. It can
   ///   be changed by calling [resync]. See [TickerProvider] for advice on
   ///   obtaining a ticker provider.
+  ///
+  /// * [pauseWhenNoListeners] controls whether the animation controller should
+  ///   pause frame scheduling when there are no listeners. When set to true,
+  ///   the controller will not request frames (saving CPU/GPU resources) until
+  ///   a listener is added. Defaults to false for backward compatibility.
   AnimationController({
     double? value,
     this.duration,
@@ -250,6 +255,7 @@ class AnimationController extends Animation<double>
     this.lowerBound = 0.0,
     this.upperBound = 1.0,
     this.animationBehavior = AnimationBehavior.normal,
+    this.pauseWhenNoListeners = false,
     required TickerProvider vsync,
   }) : assert(upperBound >= lowerBound),
        _direction = _AnimationDirection.forward {
@@ -272,6 +278,9 @@ class AnimationController extends Animation<double>
   ///   be changed by calling [resync]. See [TickerProvider] for advice on
   ///   obtaining a ticker provider.
   ///
+  /// * [pauseWhenNoListeners] controls whether the animation controller should
+  ///   pause frame scheduling when there are no listeners. Defaults to false.
+  ///
   /// This constructor is most useful for animations that will be driven using a
   /// physics simulation, especially when the physics simulation has no
   /// pre-determined bounds.
@@ -282,6 +291,7 @@ class AnimationController extends Animation<double>
     this.debugLabel,
     required TickerProvider vsync,
     this.animationBehavior = AnimationBehavior.preserve,
+    this.pauseWhenNoListeners = false,
   }) : lowerBound = double.negativeInfinity,
        upperBound = double.infinity,
        _direction = _AnimationDirection.forward {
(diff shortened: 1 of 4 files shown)


packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -864,6 +930,18 @@ class AnimationController extends Animation<double>
     _lastElapsedDuration = Duration.zero;
     _value = clampDouble(simulation.x(0.0), lowerBound, upperBound);
     final TickerFuture result = _ticker!.start();
+
+    if (pauseWhenNoListeners) {
+      if (_listenerCount == 0) {
+        if (!_ticker!.isPaused) {
+          _ticker!.pause();
+        }
+      } else {
```
warioddly, line 939: I think you can do this here `else if (_ticker!.isPaused)`

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -308,6 +318,24 @@ class AnimationController extends Animation<double>
...
+  /// Defaults to false for backward compatibility. Set to true to enable this
+  /// optimization.
+  ///
+  /// Example usage:
+  /// ```dart
+  /// final controller = AnimationController(
+  ///   vsync: this,
+  ///   duration: const Duration(seconds: 1),
+  ///   pauseWhenNoListeners: true, // Enable optimization
+  /// );
```
justinmc, line 335: IMO this code block is not needed, all it shows is setting the parameter to true right?

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -327,6 +355,44 @@ class AnimationController extends Animation<double>
...
+  ///
+  /// When [pauseWhenNoListeners] is true, this resumes the ticker when the
+  /// first listener is added while the animation is running.
+  @override
+  void didRegisterListener() {
+    if (!pauseWhenNoListeners) {
+      return;
+    }
+    _listenerCount += 1;
+    if (_listenerCount > 0 && isAnimating) {
```
justinmc, line 374: _listenerCount can never be less than 0 here, right? Maybe assert that.

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -327,6 +355,44 @@ class AnimationController extends Animation<double>
...
+  /// Called when a listener is unregistered with [removeListener] or [removeStatusListener].
+  ///
+  /// When [pauseWhenNoListeners] is true, this pauses the ticker when the
+  /// last listener is removed while the animation is running.
+  @override
+  void didUnregisterListener() {
+    if (!pauseWhenNoListeners) {
+      return;
+    }
+    _listenerCount -= 1;
```
justinmc, line 389: Maybe also assert that _listenerCount is not less than 0.

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -864,6 +930,18 @@ class AnimationController extends Animation<double>
...
+      if (_listenerCount == 0) {
+        if (!_ticker!.isPaused) {
+          _ticker!.pause();
+        }
+      } else {
+        if (_ticker!.isPaused) {
+          _ticker!.resume();
+        }
+      }
+    }
```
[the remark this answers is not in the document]
justinmc: I think I agree with this.

packages/flutter/lib/src/animation/animation_controller.dart
```diff
@@ -864,6 +930,18 @@ class AnimationController extends Animation<double>
     _lastElapsedDuration = Duration.zero;
     _value = clampDouble(simulation.x(0.0), lowerBound, upperBound);
     final TickerFuture result = _ticker!.start();
+
+    if (pauseWhenNoListeners) {
+      if (_listenerCount == 0) {
+        if (!_ticker!.isPaused) {
```
justinmc, line 936: This could be a single `if` with `&&`.

justinmc (commented):
I think you're probably right that something like this is needed to make this optimization, and this is probably an optimization worth making. I just want to make sure we get the API right.

The API for Ticker is getting big. We now have: 

 * pause
 * resume
 * stop
 * muted
 * isActive
 * isPaused

Why does `muted` not work for this case? If we do need to add these new methods, we need to be very clear in the docs how all of them work and how they relate to each other. You should probably modify some of the existing docs to point to the new members, e.g. add a "See also" reference to `pause` in the docs for `muted`.

justinmc:
@zionjo89757 I'm going to close this as stale, but let me know if you still have plans to finish it and I'm happy to reopen.

PR #182304, opened 2026-02-12
Add plugin dependencies to Add to App FlutterPluginRegistrant

### LINKED ISSUE #181207: [SwiftPM Add2App] Copy plugins and add FlutterFramework dependency
## Design

When building for SwiftPM add2app, the plugins that support SwiftPM should be copied from pubcache to the `--output` directory, under a `Plugins` subdirectory.
```
├── FlutterPluginRegistrant
├── Plugins
│   ├── plugin_a
│   ├── plugin_b
...
```

If a plugin does not have a dependency on the FlutterFramework, it should add it. For example, if the plugin's Package.swift looks like this:

```swift
let package = Package(
    name: "plugin_a",
    products: [
        .library(name: "plugin-a", targets: ["plugin_a"]),
    ],
    dependencies: [],
    targets: [
        .target(
            name: "plugin_a",
            dependencies: [
                .target(name: "plugin_a_helper"),
            ]
        ),
        .target(
            name: "plugin_a_helper",
            dependencies: []
        ),
    ]
)
```

You can run the following commands to update the Package.swift
* `swift package add-dependency ../FlutterFramework --type path`
* `swift package add-target-dependency FlutterFramework plugin_a --package FlutterFramework`
* `swift package add-target-dependency FlutterFramework plugin_a_helper --package FlutterFramework`

The resulting Package.swift would looked like:

```swift
let package = Package(
    name: "plugin_a",
    products: [
        .library(name: "plugin-a", targets: ["plugin_a"]),
    ],
    dependencies: [
        .package(name: "FlutterFramework", path: "../FlutterFramework")
    ],
    targets: [
        .target(
            name: "plugin_a",
            dependencies: [
                .target(name: "plugin_a_helper"),
                .product(name: "FlutterFramework", package: "FlutterFramework")
            ]
        ),
        .target(
            name: "plugin_a_helper",
            dependencies: [
                .product(name: "FlutterFramework", package: "FlutterFramework")
            ]
        ),
    ]
)
```

vashworth:
This PR copies Flutter plugins that support SwiftPM into the FlutterPluginRegistrant directory and adds them as dependencies in the FlutterPluginRegistrant Package.swift. It also parses the [SupportedPlatform](https://developer.apple.com/documentation/packagedescription/supportedplatform) out of the Package.swift for each plugin and finds the highest one. It then makes that value the SupportedPlatform for the FlutterPluginRegistrant.

Fixes https://github.com/flutter/flutter/issues/181207.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/lib/src/commands/build_swift_package.dart
+++ b/packages/flutter_tools/lib/src/commands/build_swift_package.dart
@@ -490,10 +490,7 @@ class FlutterPluginSwiftDependencies {
       // Example: https://github.com/firebase/flutterfire/blob/198aef8db6c96a08f57d750f1fa756da5e4a68a5/packages/firebase_core/firebase_core/ios/firebase_core/Package.swift#L21-L26
       final Directory pluginDestination = pluginsDirectory.childDirectory(plugin.name)
         ..createSync(recursive: true);
-      copyDirectory(
-        _utils.fileSystem.directory(plugin.path),
-        pluginDestination,
-      );
+      copyDirectory(_utils.fileSystem.directory(plugin.path), pluginDestination);
 
       final String? swiftPackagePath = plugin.pluginSwiftPackagePath(
         _utils.fileSystem,
@@ -571,20 +568,18 @@ class FlutterPluginSwiftDependencies {
     if (!manifestContents.contains('platforms')) {
       return null;
     }
-    // First attempt to parse with regex, which is fast
-    // e.g. \.iOS\(([\.v\d"\s]*)\) matches .iOS("13.0") or .iOS(.v13) or .iOS(.v10_15)
+    // First, attempt to parse with regex, which is fast
+    // e.g. \.iOS\([\s"]*([\._v\d]*)[\s"]*\) matches .iOS("13.0") or .iOS(.v13) or .iOS(.v10_15)
     final pattern = RegExp(
-      '\\${_targetPlatform.swiftPackagePlatform.displayName}\\(([_v\\.\\d\\s"]*)\\)',
+      r'\'
+      '${_targetPlatform.swiftPackagePlatform.displayName}'
+      r'\([\s"]*([\._v\d]*)[\s"]*\)',
     );
     final Iterable<RegExpMatch> matches = pattern.allMatches(manifestContents);
     if (matches.length == 1) {
       final String? match = matches.first.group(1);
       if (match != null) {
-        final String normalizedVersionString = match
-            .replaceAll('"', '')
-            .replaceAll(' ', '')
-            .replaceAll('.v', '')
-            .replaceAll('_', '.');
+        final String normalizedVersionString = match.replaceAll('.v', '').replaceAll('_', '.');
         final Version? parsedVersion = Version.parse(normalizedVersionString);
         if (parsedVersion != null) {
           return parsedVersion;
@@ -608,8 +603,7 @@ class FlutterPluginSwiftDependencies {
     }
     throwToolExit(
       'Unable to parse ${_targetPlatform.name} supported platform version from '
-      '${swiftPackageManifest.path}. $_kFileAnIssue and include the contents of the aforementioned '
-      'file.',
+      '${swiftPackageManifest.path}. $_kFileAnIssue and include the contents of this file.',
     );
   }
 
@@ -624,8 +618,8 @@ class FlutterPluginSwiftDependencies {
       return json.decode(parsedManifest.stdout.toString()) as Map<String, Object?>;
     } on Exception catch (e, stackTrace) {
       throwToolExit(
-        'Failed to decode ${swiftPackageManifest.path}. $_kFileAnIssue and include aforementioned '
-        'file and the following stack trace: \n$e\n$stackTrace',
+        'Failed to decode ${swiftPackageManifest.path}. $_kFileAnIssue and include this file '
+        'and the following stack trace: \n$e\n$stackTrace',
       );
     }
   }
--- a/packages/flutter_tools/lib/src/commands/build_swift_package.dart
+++ b/packages/flutter_tools/lib/src/commands/build_swift_package.dart
@@ -428,14 +428,16 @@ class FlutterPluginSwiftDependencies {
 
   /// The highest [SwiftPackageSupportedPlatform] among all of the Flutter SwiftPM plugins.
   /// Defaults to the Flutter framework's [SwiftPackageSupportedPlatform].
-  late SwiftPackageSupportedPlatform highestSupportedVersion =
+  SwiftPackageSupportedPlatform get highestSupportedVersion => _highestSupportedVersion;
+  late SwiftPackageSupportedPlatform _highestSupportedVersion =
       _targetPlatform.supportedPackagePlatform;
 
   @visibleForTesting
   /// A list of [Plugin]s copied and path to the copied Swift package.
   final List<(Plugin, String)> copiedPlugins = [];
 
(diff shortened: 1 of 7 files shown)


packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,261 @@ class FlutterPluginRegistrantSwiftPackage {
...
+    }
+    for (final plugin in plugins) {
+      // If plugin does not support the platform, skip it.
+      if (!plugin.supportSwiftPackageManagerForPlatform(_utils.fileSystem, _targetPlatform.name)) {
+        continue;
+      }
+
+      // The entire plugin is copied (excluding the example app) instead of just the Swift package
+      // to maintain any relative links within the plugin.
+      // Example: https://github.com/firebase/flutterfire/blob/198aef8db6c96a08f57d750f1fa756da5e4a68a5/packages/firebase_core/firebase_core/ios/firebase_core/Package.swift#L21-L26
```
stuartmorgan-g, line 490: In theory couldn't someone do this with a file in `example/`?

packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,261 @@ class FlutterPluginRegistrantSwiftPackage {
...
+  }
+
+  /// Determines the highest [SwiftPackageSupportedPlatform] version among the plugins and saves
+  /// the value to [highestSupportedVersion] to later be used when creating the
+  /// FlutterPluginRegistrant.
+  ///
+  /// Saves the value to a file in [cacheDirectory] for quicker lookup when the list of Swift
+  /// package manifests has not changed.
+  @visibleForTesting
+  Future<bool> determineHighestSupportedVersion({
```
stuartmorgan-g: The return value is very non-obvious, so should be documented clearly.

Optionally, consider using a record here to return a bool and the highest version, and then using that at the call site, instead of having the thing that the method sounds like it should be returning only be captured via side-effect.

packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,261 @@ class FlutterPluginRegistrantSwiftPackage {
...
+  /// Parses the [SwiftPackageSupportedPlatform] from the Package.swift using either regex or
+  /// `swift` command line tool.
+  Future<Version?> _parseSwiftPackageSupportedPlatform(File swiftPackageManifest) async {
+    final String manifestContents = swiftPackageManifest.readAsStringSync();
+    if (!manifestContents.contains('platforms')) {
+      return null;
+    }
+    // First attempt to parse with regex, which is fast
+    // e.g. \.iOS\((.*)\) matches .iOS("13.0") or .iOS(.v13) or .macOS(.v10_15)
+    final pattern = RegExp('\\${_targetPlatform.swiftPackagePlatform.displayName}\\((.*)\\)');
```
stuartmorgan-g: This should probably be a tighter match to the expected format; I assume someone could do `.iOS("\(some complex expression)")`, in which case we need the other branch.

vashworth: Done. I made it so it only accepts values of `_`, `v`, `.`, `\d` (any digit), `\s` (whitespace), and `"`

packages/flutter_tools/lib/src/commands/build_swift_package.dart
vashworth: Refactored to return the version and then set the value

packages/flutter_tools/lib/src/commands/build_swift_package.dart
vashworth: I suppose so, although, seems unlikely. It doesn't hurt much to copy the example app. I'll just include it for now.

stuartmorgan-g: > I suppose so, although, seems unlikely.

Agreed, but I would have said reaching multiple levels outside the package to dynamically construct a version from external information was unlikely with the link to someone doing that 😉

packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,262 @@ class FlutterPluginRegistrantSwiftPackage {
...
+  /// `swift` command line tool.
+  Future<Version?> _parseSwiftPackageSupportedPlatform(File swiftPackageManifest) async {
+    final String manifestContents = swiftPackageManifest.readAsStringSync();
+    if (!manifestContents.contains('platforms')) {
+      return null;
+    }
+    // First attempt to parse with regex, which is fast
+    // e.g. \.iOS\(([\.v\d"\s]*)\) matches .iOS("13.0") or .iOS(.v13) or .iOS(.v10_15)
+    final pattern = RegExp(
+      '\\${_targetPlatform.swiftPackagePlatform.displayName}\\(([_v\\.\\d\\s"]*)\\)',
```
stuartmorgan-g: IIRC you can autoconcatenate raw and non-raw strings, so you may be able to make this less `\`-noisy by doing `r'\' '${...}' r'\(...'`.

stuartmorgan-g: Should the `\s` really be in the `[]`, or should it be `\s*[...]*\s*`?

packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,262 @@ class FlutterPluginRegistrantSwiftPackage {
...
+    );
+    final Iterable<RegExpMatch> matches = pattern.allMatches(manifestContents);
+    if (matches.length == 1) {
+      final String? match = matches.first.group(1);
+      if (match != null) {
+        final String normalizedVersionString = match
+            .replaceAll('"', '')
+            .replaceAll(' ', '')
```
stuartmorgan-g: `RegExp(r'\s')`? In theory, some heathen could put a tab inside the `()`, and it would match above but then not be stripped out here.

Alternately, if you move the `\s` out above, you could `trim()` instead of this.

packages/flutter_tools/lib/src/commands/build_swift_package.dart
```diff
@@ -381,6 +413,262 @@ class FlutterPluginRegistrantSwiftPackage {
...
+        if (parsedPlatform != null &&
+            parsedPlatform.platform == _targetPlatform.swiftPackagePlatform) {
+          return parsedPlatform.version;
+        }
+      }
+      return null;
+    }
+    throwToolExit(
+      'Unable to parse ${_targetPlatform.name} supported platform version from '
+      '${swiftPackageManifest.path}. $_kFileAnIssue and include the contents of the aforementioned '
```
stuartmorgan-g: Maybe just "of this file"? (Mostly thinking that a simpler phrasing might be easier for non-native speakers.)

stuartmorgan-g (approved):
All comments optional; LGTM

packages/flutter_tools/lib/src/commands/build_swift_package.dart
vashworth: Done.

packages/flutter_tools/lib/src/commands/build_swift_package.dart
vashworth: Done.

vashworth: Oh good idea! This is great, I don't have to strip the whitespace out when I normalize it since it's not allowed in the capture group. I can do the same for the `"` too

Issue #179673, opened 2025-12-10
Linux_android_emu android_engine_opengles_tests is 2.04% flaky


fluttergithubbot:
The post-submit test builder `Linux_android_emu android_engine_opengles_tests` had a flaky ratio 2.04% for the past (up to) 100 commits, which is above our 2.00% threshold.

One recent flaky example for a same commit: https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android_emu%20android_engine_opengles_tests/4609
Commit: https://github.com/flutter/flutter/commit/4000ce2b4e576e356ee0f8a3841bbb5e280cac2c

Flaky builds:
https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android_emu%20android_engine_opengles_tests/4609
https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android_emu%20android_engine_opengles_tests/4585

Recent test runs:
https://flutter-dashboard.appspot.com/#/build?taskFilter=Linux_android_emu%20android_engine_opengles_tests

Please follow https://github.com/flutter/flutter/blob/master/docs/infra/Reducing-Test-Flakiness.md#fixing-flaky-tests to fix the flakiness and enable the test back after validating the fix (internal dashboard to validate: go/flutter_test_flakiness).

Piinks:
This did not get a team issue, added team-android to route to triage.

fluttergithubbot:
[prod pool] flaky ratio for the past (up to) 100 commits between 2025-12-17 and 2025-12-23 is 1.01%. Flaky number: 1; total number: 99.
One recent flaky example for a same commit: https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android_emu%20android_engine_opengles_tests/4778
Commit: https://github.com/flutter/flutter/commit/a96687c0945f6dee6a9fa083b61aecee716e1312
Flaky builds:
https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android_emu%20android_engine_opengles_tests/4778

Recent test runs:
https://flutter-dashboard.appspot.com/#/build?taskFilter=Linux_android_emu%20android_engine_opengles_tests

fluttergithubbot:
[prod pool] flaky ratio for the past (up to) 100 commits between 2026-01-13 and 2026-01-20 is 0.00%. Flaky number: 0; total number: 99.

fluttergithubbot:
[...]

PR #182326, opened 2026-02-12
Fix Content Sizing Race Condition

mboetger:
Only call MaybeResizeSurfaceView if the expected frame_size is not the same size as what is returned from the Dart layout. Also, remove the gating function that prevents viewport metrics from being sent on a resize.

The AtomicBoolean that is gating calls back into the engine has to be removed.  This is necessary because the buffer for the SurfaceView is only updated once the surface finishes drawing to the screen.  Since the frame can be submitted before the buffer is updated (but after the resize call happens), the UI rejects the frame because the buffer dimensions do not match.  Therefore, the callback with the updated viewport metrics is necessary for the frame to be re-tried. 

I believe this AtomicBoolean is ultimately the cause of the initial flake. Since rotations are actually changing the size, the AtomicBoolean would have been set to not send the viewport metrics back when necessary.

The call from the engine should remain non-blocking to prevent deadlocks.

Fixes: #182242

## Pre-launch Checklist

- [X] I signed the [CLA].


mboetger:
Local testing shows this also fixes the [flake](https://github.com/flutter/flutter/issues/179673) (if the listener is always added regardless of if content sizing is enabled - though this PR does NOT remove that condition check).

mboetger:
> Sorry. Not really. Probably I misinterpreted the fix and got the results I needed by luck or another commit. =/ I was facing the problem using `stable` (3.41.4) and it was fixed by using your branch.

@rwrz - even more interesting.  How did you decide to use my branch?  Were you seeing this line in the logs: "Resize was in response to the engine resizing the view. Not sending viewport metrics."?

mboetger:
@rwrz I guess it's possible another PR fixed the issue somewhere in between stable and HEAD for my PR.  I would be extra curious if you cherrypicked just my fix into the stable branch and saw the fix.

gmackall (approved):

Issue #182105, opened 2026-02-09
Fix broken opengles dart engine tests that crash with GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT


b-luk:
Some opengles dart engine tests are currently skipped because they crash with the following error:
```
[ERROR:flutter/impeller/renderer/backend/gles/blit_command_gles.cc(79)] Break on 'ImpellerValidationBreak' to inspect point of failure: Could not create a complete framebuffer: GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT
[FATAL:flutter/impeller/renderer/backend/gles/blit_pass_gles.cc(88)] Check failed: result. Must be able to encode GL commands without error.
```

These tests seem to pass when running with an unopt build (e.g. `host_debug_unopt_arm64`), and crash when running on a non-unopt build (e.g. `ci/host_debug_arm64_tests`).

The following tests have this behavior:
```
codec_test.dart
decode_image_from_pixels_sync_test.dart
encoding_test.dart
image_dispose_test.dart
image_resize_test.dart
```

b-luk:
This failure behavior is very similar a bunch of tests which originally failed but were fixed as part of https://github.com/flutter/flutter/pull/181933.

The fix for those tests was to properly handle the RAII return value of `MakeRenderContextCurrent()` in a place which wasn't properly handling it before.

See https://github.com/flutter/flutter/pull/181933/changes#r2775519231

It's possible that the still-failing tests are due to other places where we don't properly handle a `MakeRenderContextCurrent()` return value.

PR #182329, opened 2026-02-12
Remove `FML_DCHECK` around `CanReactOnCurrentThread()` in `ReactorGLES::CreateUntrackedHandle()

b-luk:
`ReactorGLES::CreateUntrackedHandle()` requires the current thread to have the current opengles context. In environments that use `SwitchableGLContext`, calling `CanReactOnCurrentThread()` has the side effect of setting the current opengles context. So it must be called. Having it inside `FML_DCHECK` made the call only happen in unopt builds, resulting in https://github.com/flutter/flutter/issues/182105.

Makes progress on https://github.com/flutter/flutter/issues/182105. The `GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT ` issues no longer happen. Newly passing tests:
```
decode_image_from_pixels_sync_test.dart
encoding_test.dart
image_dispose_test.dart
image_resize_test.dart
```

`codec_test.dart` no longer has the `GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT ` error, but still fails. It passes on mac, but on linux it fails with:
```
[ERROR:flutter/impeller/renderer/backend/gles/blit_command_gles.cc(343)] Break on 'ImpellerValidationBreak' to inspect point of failure: Only textures with pixel format RGBA are supported yet.
[FATAL:flutter/impeller/renderer/backend/gles/blit_pass_gles.cc(88)] Check failed: result. Must be able to encode GL commands without error.
```

I have a pending fix for this which I will send out afterwards.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
+++ b/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
@@ -201,7 +201,7 @@ bool ReactorGLES::RegisterCleanupCallback(const HandleGLES& handle,
 }
 
 HandleGLES ReactorGLES::CreateUntrackedHandle(HandleType type) const {
-  bool can_react = CanReactOnCurrentThread();
+  [[maybe_unused]] bool can_react = CanReactOnCurrentThread();
   FML_DCHECK(can_react);
   auto new_handle = HandleGLES::Create(type);
   std::optional<ReactorGLES::GLStorage> gl_handle =
--- a/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
+++ b/engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
@@ -201,7 +201,8 @@ bool ReactorGLES::RegisterCleanupCallback(const HandleGLES& handle,
 }
 
 HandleGLES ReactorGLES::CreateUntrackedHandle(HandleType type) const {
-  FML_DCHECK(CanReactOnCurrentThread());
+  bool can_react = CanReactOnCurrentThread();
+  FML_DCHECK(can_react);
   auto new_handle = HandleGLES::Create(type);
   std::optional<ReactorGLES::GLStorage> gl_handle =
       CreateGLHandle(GetProcTable(), type);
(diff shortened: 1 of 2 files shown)


engine/src/flutter/impeller/renderer/backend/gles/reactor_gles.cc
```diff
@@ -201,7 +201,8 @@ bool ReactorGLES::RegisterCleanupCallback(const HandleGLES& handle,
 }
 
 HandleGLES ReactorGLES::CreateUntrackedHandle(HandleType type) const {
-  FML_DCHECK(CanReactOnCurrentThread());
+  [[maybe_unused]] bool can_react = CanReactOnCurrentThread();
+  FML_DCHECK(can_react);
```
gaaclarke, line 205: Asked question about this in https://github.com/flutter/flutter/pull/182397#discussion_r2806429293

b-luk: > I'm not keen to accept this extra overhead. Do you not need this after my PR that eliminates multithreaded tests?

The issue happens even when multithreading is not enabled.

gaaclarke: Can we just call directly the thing you need from CanReactOnCurrentThread() directly?  The setting of the context.

Can we do that setting of the context once when things are initialized instead of each time we make an untracked handle?

b-luk: I don't think setting the context here directly would be right because we'd only want to do this when a SwitchableGLContext is used (currently only in tests like with TesterContextGLES). We don't want to do this for setups that don't use a SwitchableGLContext (all cases in prod right now).

We can't set the the context once when things are initialized because different things are initialized at different times, potentially in different threads. The current case that is making the tests fail is because the context is not current when creating a texture for an image, which can happen in a different thread than the main frame loop code which also needs to set the current context to do stuff.

TesterContextGLES had to implement this SwitchableGLContext logic in order to work properly with multithreading. Since we no longer are running opengles tests with multithreading enabled, we should be able to simplify TesterContextGLES and remove all of its SwitchableGLContext logic. If we do this, these test should work without calling CanReactOnCurrentThread() here, so this PR should no longer be needed. I'm working on this change to TesterContextGLES now. I'm running into a different problem with that change, but once it's ready I'll send it out and it should make this PR obsolete.

PR #182331, opened 2026-02-12
Eliminate rebuilds for Scaffold FAB animation

### LINKED ISSUE #182330: Proposal: take advantage of `relayout` parameter in `_ScaffoldLayout` class constructor
The Scaffold widget's build method creates an `AnimatedBuilder` so that the `CustomMultiChildLayout` widget is rebuilt every frame the FAB animation sends a notification.

https://github.com/flutter/flutter/blob/7bafe12c4f69c41b8bee4466296e6a30ba3564ea/packages/flutter/lib/src/material/scaffold.dart#L3207-L3233

<br>

I believe we can factor out the `AnimatedBuilder` by utilizing the [MultiChildLayoutDelegate constructor](https://main-api.flutter.dev/flutter/rendering/MultiChildLayoutDelegate/MultiChildLayoutDelegate.html)'s `relayout` parameter. This would allow the FAB animation to update the layout directly, rather than triggering a build and a subsequent [shouldRelayout](https://main-api.flutter.dev/flutter/rendering/MultiChildLayoutDelegate/shouldRelayout.html) call.

nate-thegrate:
This PR contains a tweak to the Scaffold widget's internal logic, reducing the amount of work that needs to be performed each time the FAB animation updates.

resolves #182330

### DIFF
--- a/packages/flutter/test/material/scaffold_test.dart
+++ b/packages/flutter/test/material/scaffold_test.dart
@@ -3684,6 +3684,55 @@ void main() {
     expect(find.byType(FloatingActionButton), findsNothing);
   });
 
+  testWidgets('Scaffold FAB animates without rebuilding', (WidgetTester tester) async {
+    bool? dirty;
+    bool? checkDirty() {
+      final Finder finder = find.descendant(
+        of: find.byType(ScrollNotificationObserver),
+        matching: find.ancestor(
+          of: find.byType(CustomMultiChildLayout),
+          matching: find.byWidgetPredicate((widget) {
+            return widget is AnimatedBuilder || widget is Builder;
+          }),
+        ),
+      );
+      return finder.tryEvaluate() ? tester.element(finder).dirty : null;
+    }
+
+    final fabLocation = ValueNotifier<FloatingActionButtonLocation>(.centerDocked);
+    const builderKey = Key('Builder');
+    await tester.pumpWidget(
+      MaterialApp(
+        home: ValueListenableBuilder(
+          key: builderKey,
+          valueListenable: fabLocation,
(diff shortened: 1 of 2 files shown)


packages/flutter/test/material/scaffold_test.dart
```diff
@@ -3597,6 +3597,55 @@ void main() {
     expect(find.byType(FloatingActionButton), findsNothing);
   });
 
+  testWidgets('Scaffold FAB animates without rebuilding', (WidgetTester tester) async {
```
dkwingsmt, line 3687: Do we have a test to verify that the animation is indeed applied (i.e. `relayout: ` is correctly assigned)?

dkwingsmt (approved):
LGTM. Thank you!

packages/flutter/test/material/scaffold_test.dart
nate-thegrate: Yes we do! I was able to find two tests that cover the FAB animation:

https://github.com/flutter/flutter/blob/d9556bd0db5dd92b4b257316816a1ef5264cbcf7/packages/flutter/test/material/scaffold_test.dart#L318-L366

https://github.com/flutter/flutter/blob/d9556bd0db5dd92b4b257316816a1ef5264cbcf7/packages/flutter/test/material/scaffold_test.dart#L3553-L3606

justinmc (approved):
LGTM :+1: Good catch!

nate-thegrate:
Dang it, looks like [Google Testing failed](https://github.com/flutter/flutter/pull/182331/checks?check_run_id=65026630522). I'll re-trigger the checks in hopes that it was a flake…

PR #182333, opened 2026-02-13
`AnimationStyle` methods

### LINKED ISSUE #160563: `AnimationStyle` API improvements
- It'd be nice to have `AnimationStyle.merge` (similar to [**TextStyle.merge**](https://main-api.flutter.dev/flutter/painting/TextStyle/merge.html))
- It'd also be nice if the `.lerp()` method was more than just a binary switch, since maybe someone would want to animate the animation style!
- Since it's an immutable configuration class, a `const` constructor would be nice to have as well.

nate-thegrate:
This pull request adds an `AnimationStyle.merge()` method (for parity with similar APIs) and updates `AnimationStyle.lerp()` so that it performs an accurate interpolation.

<br>

closes #160563

### DIFF
--- a/packages/flutter/lib/src/animation/animation_style.dart
+++ b/packages/flutter/lib/src/animation/animation_style.dart
@@ -71,11 +71,39 @@ class AnimationStyle with Diagnosticable {
       return a;
     }
     return AnimationStyle(
-      curve: t < 0.5 ? a?.curve : b?.curve,
-      duration: t < 0.5 ? a?.duration : b?.duration,
-      reverseCurve: t < 0.5 ? a?.reverseCurve : b?.reverseCurve,
-      reverseDuration: t < 0.5 ? a?.reverseDuration : b?.reverseDuration,
+      curve: _lerpCurve(a?.curve, b?.curve, t),
+      duration: _lerpDuration(a?.duration, b?.duration, t),
+      reverseCurve: _lerpCurve(a?.reverseCurve, b?.reverseCurve, t),
+      reverseDuration: _lerpDuration(a?.reverseDuration, b?.reverseDuration, t),
     );
+  }
+  
+  static Duration? _lerpDuration(Duration? a, Duration? b, double t) {
+    if (t == 0.0) {
+      return a;
+    }
+    if (t == 1.0) {
+      return b;
+    }
+    if (a == null || b == null) {
+      return a ?? b;
+    }
+    return Duration(
+      microseconds: (a.inMicroseconds * (1.0 - t) + b.inMicroseconds * t).round(),
+    );
+  }
+
+  static Curve? _lerpCurve(Curve? a, Curve? b, double t) {
+    if (t == 0.0) {
+      return a;
+    }
+    if (t == 1.0) {
+      return b;
+    }
+    if (a == null || b == null) {
+      return a ?? b;
+    }
+    return _LerpedCurve(a, b, t);
   }
 
   @override
@@ -106,4 +134,31 @@ class AnimationStyle with Diagnosticable {
       DiagnosticsProperty<Duration>('reverseDuration', reverseDuration, defaultValue: null),
     );
   }
+}
+
+class _LerpedCurve extends Curve {
+  const _LerpedCurve(this.first, this.second, this._t);
+
+  final Curve first;
+  final Curve second;
+  final double _t;
+
+  @override
+  double transform(double t) {
+    final double a = first.transform(t);
+    final double b = second.transform(t);
+
+    return a * (1.0 - _t) + b * _t;
+  }
+
+  @override
+  bool operator==(Object other) {
+    return other is _LerpedCurve
+        && other.first == first
+        && other.second == second
+        && other._t == _t;
+  }
+
+  @override
+  int get hashCode => Object.hash(first, second, _t);
 }
--- a/packages/flutter/lib/src/animation/animation_style.dart
+++ b/packages/flutter/lib/src/animation/animation_style.dart
@@ -65,6 +65,22 @@ class AnimationStyle with Diagnosticable {
     );
   }
 
+  /// Returns a modified version of the `other` style, where its `null` properties
+  /// are filled in with the non-null properties of this style, where applicable.
+  ///
+  /// If a `null` argument is passed, returns this animation style.
+  AnimationStyle merge(AnimationStyle? other) {
+    if (other == null) {
+      return this;
+    }
+    return copyWith(
+      curve: other.curve,
+      duration: other.duration,
+      reverseCurve: other.reverseCurve,
+      reverseDuration: other.reverseDuration,
+    );
+  }
+
   /// Linearly interpolate between two animation styles.
   static AnimationStyle? lerp(AnimationStyle? a, AnimationStyle? b, double t) {
     if (identical(a, b)) {
--- a/packages/flutter/lib/src/animation/animation_style.dart
+++ b/packages/flutter/lib/src/animation/animation_style.dart
@@ -87,29 +87,18 @@ class AnimationStyle with Diagnosticable {
       return a;
     }
     return AnimationStyle(
-      curve: _lerpCurve(a?.curve, b?.curve, t),
-      duration: _lerpDuration(a?.duration, b?.duration, t),
-      reverseCurve: _lerpCurve(a?.reverseCurve, b?.reverseCurve, t),
-      reverseDuration: _lerpDuration(a?.reverseDuration, b?.reverseDuration, t),
-    );
-  }
-  
-  static Duration? _lerpDuration(Duration? a, Duration? b, double t) {
-    if (t == 0.0) {
-      return a;
-    }
-    if (t == 1.0) {
-      return b;
-    }
-    if (a == null || b == null) {
-      return a ?? b;
-    }
-    return Duration(
-      microseconds: (a.inMicroseconds * (1.0 - t) + b.inMicroseconds * t).round(),
+      curve: _lerp(a?.curve, b?.curve, t, _LerpedCurve.new),
+      duration: _lerp(a?.duration, b?.duration, t, _lerpDuration),
+      reverseCurve: _lerp(a?.reverseCurve, b?.reverseCurve, t, _LerpedCurve.new),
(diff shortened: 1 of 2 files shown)


packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +65,55 @@ class AnimationStyle with Diagnosticable {
     );
   }
 
+  /// Returns a modified version of the `other` style, where its `null` properties
+  /// are filled in with the non-null properties of this style, where applicable.
+  ///
+  /// If a `null` argument is passed, returns this animation style.
```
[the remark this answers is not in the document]
nate-thegrate: Sure, why not?

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +65,55 @@ class AnimationStyle with Diagnosticable {
...
+      return a;
+    }
+    if (t == 0.0) {
+      return a;
+    }
+    if (t == 1.0) {
+      return b;
+    }
+    if (a == null || b == null) {
+      return a ?? b;
```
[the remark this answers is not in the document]
nate-thegrate: > The line `if (a == null || b == null) { return a ?? b; }` causes a sudden jump instead of a smooth transition.

I'll point out that even with the proposed changes, there could still be a "sudden jump", e.g. if the fallback `duration` is greater than zero.

That being said, I'll go ahead and update the logic 👍

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
...
 
+  /// Creates a new animation style that is a combination of this animation style
+  /// and the given `other` animation style.
+  ///
+  /// If `other` is non-null, its non-null properties are used to override the
+  /// corresponding properties of this style.
+  ///
+  /// If a `null` argument is passed, returns this animation style.
+  AnimationStyle merge(AnimationStyle? other) {
+    if (other == null) {
```
navaronbracke, line 75: Calling `merge(this)` should probably not call copyWith()?
```suggestion
    if (other == null || identical(other, this)) {
```

navaronbracke (changes requested):

packages/flutter/lib/src/animation/animation_style.dart
nate-thegrate: Looking at the implementation in other classes:

- [ButtonStyle.merge](https://main-api.flutter.dev/flutter/material/ButtonStyle/merge.html)
- [IconThemeData.merge](https://main-api.flutter.dev/flutter/widgets/IconThemeData/merge.html)
- [InputDecorationThemeData.merge](https://main-api.flutter.dev/flutter/material/InputDecorationThemeData/merge.html)
- [MenuStyle.merge](https://main-api.flutter.dev/flutter/material/MenuStyle/merge.html)
- [StepStyle.merge](https://main-api.flutter.dev/flutter/material/StepStyle/merge.html)
- [StrutStyle.merge](https://main-api.flutter.dev/flutter/painting/StrutStyle/merge.html)
- [TextStyle.merge](https://main-api.flutter.dev/flutter/painting/TextStyle/merge.html)

It looks like none of them make the `identical` check. Maybe calling `merge(this)` doesn't happen often enough for the overhead of that function to be worth it?

I think it makes sense for this method's implementation to match others in the framework, but let me know what you think.

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -107,3 +145,32 @@ class AnimationStyle with Diagnosticable {
     );
   }
 }
+
+class _LerpedCurve extends Curve {
```
victorsanni, line 146: Could a `toString` be useful for debugging this class?

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
...
+      curve: _lerp(a?.curve, b?.curve, t, _LerpedCurve.new),
+      duration: _lerp(a?.duration, b?.duration, t, _lerpDuration),
+      reverseCurve: _lerp(a?.reverseCurve, b?.reverseCurve, t, _LerpedCurve.new),
+      reverseDuration: _lerp(a?.reverseDuration, b?.reverseDuration, t, _lerpDuration),
+    );
+  }
+
+  @optionalTypeArgs
+  static T? _lerp<T extends Object>(T? a, T? b, double t, T Function(T? a, T? b, double t) lerp) {
+    if (identical(a, b)) {
```
victorsanni: Can `identical(a,b) != (a == b)`?

nate-thegrate: Good question—`identical` and `==` do the same thing unless you override the `==` operator.

We could switch this line to `if (a == b)` in order to catch a few non-identical objects that still should be considered "equal", but the tradeoff would be a higher computational cost anytime we actually do the interpolation.

packages/flutter/lib/src/animation/animation_style.dart
nate-thegrate: That's a wonderful idea 🙂

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
     );
   }
 
+  /// Creates a new animation style that is a combination of this animation style
```
navaronbracke:
```suggestion
  /// Creates a new [AnimationStyle] that is a combination of this animation style
```

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
...
+  /// and the given `other` animation style.
+  ///
+  /// If `other` is non-null, its non-null properties are used to override the
+  /// corresponding properties of this style.
+  ///
+  /// If a `null` argument is passed, returns this animation style.
```
navaronbracke:
```suggestion
  /// Returns this animation style if `other` is null.
```

packages/flutter/lib/src/animation/animation_style.dart
navaronbracke: Makes sense to follow that

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -107,3 +145,35 @@ class AnimationStyle with Diagnosticable {
...
+        other.first == first &&
+        other.second == second &&
+        other._t == _t;
+  }
+
+  @override
+  int get hashCode => Object.hash(first, second, _t);
+
+  @override
+  String toString() => 'interpolated curve ($first, $second, t: $_t)';
```
navaronbracke: I don't like how it is lowercase words here

```suggestion
  String toString() => '_LerpedCurve($first, $second, t: $_t)';
```

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -107,3 +145,35 @@ class AnimationStyle with Diagnosticable {
     );
   }
 }
+
+class _LerpedCurve extends Curve {
```
navaronbracke, line 146: I wonder if this curve class should be public and in flutter/foundation or something?

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
...
+      reverseDuration: _lerp(a?.reverseDuration, b?.reverseDuration, t, _lerpDuration),
+    );
+  }
+
+  @optionalTypeArgs
+  static T? _lerp<T extends Object>(T? a, T? b, double t, T Function(T? a, T? b, double t) lerp) {
+    if (identical(a, b)) {
+      return a;
+    }
+    if (t == 0.0) {
```
navaronbracke: Should lerp() not handle t == 0 and t == 1 for you?

packages/flutter/lib/src/animation/animation_style.dart
```diff
@@ -65,16 +64,55 @@ class AnimationStyle with Diagnosticable {
...
+      return a;
+    }
+    if (t == 1.0) {
+      return b;
+    }
+    return lerp(a, b, t);
+  }
+
+  static Duration _lerpDuration(Duration? a, Duration? b, double t) {
```
navaronbracke, line 110: Unfortunate that this can't be added to Duration, but it is what it is.

navaronbracke (changes requested):

packages/flutter/lib/src/animation/animation_style.dart
nate-thegrate: I know that for methods like [**Color.lerp**](https://main-api.flutter.dev/flutter/dart-ui/Color/lerp.html), handling t == 0 and t == 1 this way can be problematic; oftentimes the method converts both colors to the extended RGB color space in order to perform the interpolation, so returning one of the inputs might cause a jump.

Since this is a private method that we know is only interpolating Curve and Duration objects, I believe we're fine to keep it as-is.

packages/flutter/lib/src/animation/animation_style.dart
nate-thegrate: Yeah, I could see how that might be useful. Might be outside the scope of this PR, but maybe after this lands we could open a new one to make it public?

navaronbracke (dismissed):

packages/flutter/lib/src/animation/animation_style.dart
victorsanni: I am just wondering if missing some non-identical objects can lead to future issues. Maybe it is safest to just do a == b?

nate-thegrate: Works for me!

victorsanni (dismissed):

navaronbracke:
@nate-thegrate it looks like some of the feedback that was addressed, was reverted in code?

navaronbracke (approved):

victorsanni (approved):

justinmc (approved):
LGTM, let's see if the Google tests pass.

justinmc (approved):
LGTM again after the merge.

PR #182348, opened 2026-02-13
Implement tooltip windows on Linux

robert-ancell:
Add support for tooltips on Linux. These are implemented using a popup window. Due to a limitation of the Wayland xdg_popup implementation they do not reposition after being shown, which means the example application does not behave the same on Linux.


engine/src/flutter/shell/platform/linux/fl_view_monitor.cc
```diff
@@ -0,0 +1,60 @@
...
+
+  // Callbacks.
+  void (*on_first_frame)(void);
+};
+
+G_DEFINE_TYPE(FlViewMonitor, fl_view_monitor, G_TYPE_OBJECT)
+
+static void first_frame_cb(FlViewMonitor* self) {
+  flutter::IsolateScope scope(self->isolate);
+  self->on_first_frame();
```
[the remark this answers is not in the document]
robert-ancell: Fixing in other PR

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -495,6 +535,180 @@ class DialogWindowControllerLinux extends DialogWindowController {
...
+  }) : _owner = owner,
+       _delegate = delegate,
+       _parent = parent,
+       _window = _GtkWindow(_GtkWindowType.popup),
+       super.empty() {
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+
+    //_window.setTypeHint(_GDK_WINDOW_TYPE_HINT_TOOLTIP);
```
mattkae: Unneeded?

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -636,11 +877,36 @@ class _GtkWidget extends _GObject {
     return _GdkWindow(_gtkWidgetGetWindow(instance));
   }
 
+  /// Get the scale factor that maps window coordinates to device pixels.
+  int getScaleFactor() {
+    return _gtkWidgetGetScaleFactor(instance);
+  }
+
+  (int, int)? translateCoordinates(_GtkWidget destWidget, (int, int) src) {
```
mattkae, line 888: Nit: Can this just return `src` if `translated` is false?

packages/flutter/lib/src/widgets/_window_linux.dart
```diff
@@ -495,6 +535,180 @@ class DialogWindowControllerLinux extends DialogWindowController {
...
+    rootView = WidgetsBinding.instance.platformDispatcher.views.firstWhere(
+      (FlutterView view) => view.viewId == viewId,
+    );
+    _view.show();
+    _window.add(_view);
+
+    final _GtkWindow? parentWindow = _owner._windows[_parent.rootView.viewId];
+    if (parentWindow != null) {
+      _window.setTransientFor(parentWindow);
+    }
```
mattkae, line 603: If a tooltip window lacks a parent, that should be a real problem, right? We shouldn't silently do something else in my opinion.

mattkae (commented):
Some minor questions and things!

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Ah, I was having problems with this, but it seems to work now.

packages/flutter/lib/src/widgets/_window_linux.dart
robert-ancell: Since this is wrapping GTK API I want it to match that as closely as possible.

mattkae (approved):
I personally have no issues with this 🎉 

Great stuff!

PR #182357, opened 2026-02-13
[widgets/raw_menu_anchor.dart] Always call onClose and onCloseRequested on descendants before parent.

### LINKED ISSUE #182355: [widgets/raw_menu_anchor.dart] onClose should be called by descendant menus before parent
### Steps to reproduce

1. Run example
2. Open until you reach "Close Root Menu"
3. Tap "Close Root Menu"
4. Read console

### Expected results

Should print 
```sh
> onClose Submenu
> onClose Root
```

### Actual results

Prints
```sh
> onClose Root
> onClose Submenu
```

### Code sample

```dart

void main() {
  runApp(const App());
}

class App extends StatefulWidget {
  const App({super.key});

  @override
  State<App> createState() => _AppState();
}

class _AppState extends State<App> {
  @override
    return const CupertinoApp(home: CupertinoPageScaffold(child: Bug()));
  }
}

class Bug extends StatelessWidget {
  const Bug({super.key});

  @override
    return Center(
      child: RawMenuAnchor(
        controller: MenuController(),
        builder: (context, controller, child) {
          return CupertinoButton.filled(onPressed: controller.open, child: const Text('Open'));
        },
        onClose: () {
          print('onClose Root');
        },
        overlayBuilder: (context, info) {
          return Positioned(
            left: info.anchorRect.left,
            top: info.anchorRect.bottom + 4,
            child: TapRegion(
              groupId: info.tapRegionGroupId,
              child: RawMenuAnchor(
                controller: MenuController(),
                onClose: () {
                  print('onClose Submenu');
                },
                builder: (context, controller, child) {
                  return CupertinoButton.filled(
                    onPressed: controller.open,
                    child: const Text('Open'),
                  );
                },
                overlayBuilder: (context, info) {
                  return Positioned(
                    left: info.anchorRect.left,
                    top: info.anchorRect.bottom + 4,
                    child: TapRegion(
                      groupId: info.tapRegionGroupId,
                      child: CupertinoButton.filled(
                        onPressed: () {
                          Actions.invoke(context, const DismissIntent());
                        },
                        child: const Text('Close Root Menu'),
                      ),
                    ),
                  );
                },
              ),
            ),
          );
        },
      ),
    );
  }
}

```

</details>

### Screenshots or Video

<details open>

N/a

</details>

### Logs

No relevant logs

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel master, 3.42.0-1.0.pre-18, on macOS 15.5 24F74 darwin-arm64, locale en-US) [1,297ms]
    • Flutter version 3.42.0-1.0.pre-18 on channel master at /Users/davidhicks/fvm/versions/master
    • Framework revision 73a7978d56 (4 days ago), 2026-02-09 00:59:49 -0500
    • Engine revision 73a7978d56
    • Dart version 3.12.0 (build 3.12.0-127.0.dev)
    • DevTools version 2.54.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios,
      cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration, enable-riscv64

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [2.6s]
    • Android SDK at /Users/davidhicks/Library/Android/sdk
    • Emulator version 36.3.10.0 (build_id 14472402) (CL:N/A)
    • Platform android-36.1, build-tools 36.1.0
    • ANDROID_HOME = /Users/davidhicks/Library/Android/sdk
    • Java binary at: /Users/davidhicks/Library/Java/JavaVirtualMachines/jbr-17.0.14/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment JBR-17.0.14+1-1367.22-nomod (build 17.0.14+1-b1367.22)

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [1,276ms]
    • Build 16F6
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [6ms]

[✓] Connected device (2 available) [6.1s]
    • macOS (desktop) • macos  • darwin-arm64   • macOS 15.5 24F74 darwin-arm64
    • Chrome (web)    • chrome • web-javascript • Google Chrome 145.0.7632.68
    ! Error: Browsing on the local area network for David’s iPhone. Ensure the device is unlocked and attached with a cable or associated
      with the same local area network as this Mac.

[✓] Network resources [261ms]

• No issues found!
```

</details>

davidhicks980:
This PR corrects the order in which onClose() and onCloseRequested() is called by a RawMenuAnchor tree. 

Per @dkwingsmt:

> A close is immediate and should close children immediately before closing itself, while a closeRequest [can be] async and should closeRequest children.

Before, calling close() on a parent would call handleCloseRequest() on its children, leading to children (potentially) finishing their closure after their parents.

Additionally, if a RawMenuAnchor is already closed, it will not trigger onCloseRequested.

**Details:**
 
The handleCloseRequest portion of the _RawMenuAnchorBaseMixin.closeChildren function has been moved into a requestChildrenClose function to be more explicit.

Additionally, this PR makes handleCloseRequest call requestChildrenClose() after calling widget.onCloseRequested on _RawMenuAnchorState. This makes submenus begin closing after their parent menu begins closing. I also modified the menu_anchor.dart and raw_menu_anchor.3.dart code to remove a closeChildren call (now redundant).

Last, this PR makes handleCloseRequest short-circuit if a menu isn't open. This prevents unncessary widget.onCloseRequest calls when a submenu is opened, since all sibling menus call handleCloseRequest before the submenu begins opening.  

This is a breaking change. I had to modify one of the DismissMenuAnchor tests since onCloseRequested is now called on all descendants when a dismissal occurs. I think it's best to introduce this change because it can lead to hard-to-track behavior when creating menus (that's how I discovered it). 

Migration guide: https://github.com/flutter/website/pull/13145

Additional tests were added to check the opening and closing behavior. 

Fixes https://github.com/flutter/flutter/issues/182355

## Pre-launch Checklist

- [x] I signed the [CLA].
- [] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.


packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -600,22 +600,31 @@ mixin _RawMenuAnchorBaseMixin<T extends StatefulWidget> on State<T> {
   @protected
   void handleCloseRequest();
 
-  /// Request that the submenus of this menu be closed.
+  /// Close the open submenus of this menu..
```
dkwingsmt:
```suggestion
  /// Close the open submenus of this menu.
```

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -600,22 +600,31 @@ mixin _RawMenuAnchorBaseMixin<T extends StatefulWidget> on State<T> {
...
+      child.close(inDispose: inDispose);
+    }
+  }
+
+  /// Request that the open submenus of this menu be closed.
+  ///
+  /// This method will call [handleCloseRequest] on each child of this
+  /// menu, which will trigger the closing sequence of each child.
+  @protected
+  void closeChildrenWithRequest() {
```
dkwingsmt: nit: Will a name like requestChildrenClose be clearer and more recognizable from closeChildren? I'm fine with the current name as well, just providing another option.

dkwingsmt (dismissed):
Overall LGTM.

Honestly I think the PR description might be over complicated and I don't really understand what it has to do with "if onCloseRequested was defined by a submenu and that submenu didn't immediately call hideMenu()". But I think the change is pretty easy to understand. Basically a `close` is immediate and should close children immediately before closing itself, while a `closeRequest` is async and should `closeRequest` children, which makes perfect sense.

packages/flutter/test/widgets/raw_menu_anchor_test.dart
```diff
@@ -470,6 +470,92 @@ void main() {
     expect(find.text(Tag.b.a.text), findsNothing);
   });
```
dkwingsmt, line 472: Just to make sure, do we have a test that verifies that the root's `onClose` is called after the entire subtree is closed in an _async_ way?

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: I went with closeChildrenWithRequest to indicate to the future editors that they should pay attention to the differentiator between the two functions (i.e. they both closeChildren, so users should pay attention to what makes one a request). That said, I was between those two. Your suggestion is more accurate since calling requestChildrenClose doesn't guarantee the children will close. Let me know what you prefer and I'll add it.

packages/flutter/test/widgets/raw_menu_anchor_test.dart
davidhicks980: Like onCloseRequest() -> wait 1 second -> hideMenu()? I'll add it in if we don't have one. The dismiss test I refactored did test for async but was so convoluted that I decided to simplify it down to checking that the dismiss occurred.

dkwingsmt: It was just that, while reading the code, I wondered how `onClose` was invoked, which slightly concerned me if not tested since the flow is full of async (and I didn't want to read everything to verify that  :)   )

davidhicks980: I added an additional test to validate that async closures occur as expected. There's actually no async code (where async is code involving futures) in RawMenuAnchor, but it can be used in an async way.

That said, done!

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: Done!

davidhicks980:
@chunhtai I'm requesting a review from you because I'm curious whether we should be calling closeChildrenWithRequest() on children that are not opened when MenuController.open() is called. This can lead to some weird behavior (e.g. a user forgets to cancel their timers, so every open() request creates a new timer for each submenu).  

Edit: @dkwingsmt this is also a question for you.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
chunhtai: +1 to @dkwingsmt , I think we should prioritize readaibiity here

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -708,7 +723,7 @@ class _RawMenuAnchorState extends State<RawMenuAnchor> with _RawMenuAnchorBaseMi
     assert(_debugMenuInfo('Opening $this at ${position ?? Offset.zero}'));
 
     // Close all siblings.
-    _parent?.closeChildren();
+    _parent?.closeChildrenWithRequest();
```
chunhtai: I know this is not introduced in this pr, why would we want to do async close here? it feels weird if the new menu opens before old one close completely

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -600,22 +606,31 @@ mixin _RawMenuAnchorBaseMixin<T extends StatefulWidget> on State<T> {
...
-  /// Request that the submenus of this menu be closed.
+  /// Close the open submenus of this menu.
   ///
-  /// By default, this method will call [handleCloseRequest] on each child of this
-  /// menu, which will trigger the closing sequence of each child.
+  /// This method will call [close] on each child of this menu, which will
+  /// immediately close the child.
   ///
   /// If `inDispose` is true, this method was triggered by the widget being
   /// unmounted.
```
chunhtai, line 622: add a see also to reference closeChildrenWithRequest and vice versa

chunhtai (commented):
I am onboard with this change, just left some suggestion.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
dkwingsmt: Now that you mentioned, I tested on macOS's system menu bar, and noticed that, while top menus have a tiny fade out animation, there is no fade out when switching between menus. I'm not sure if this applies to all other menus in other design languages (i.e. whether we should leave this as a flag or callback for design languages to implement), but the async closure here should indeed be wrong.

https://github.com/user-attachments/assets/b2407708-41b1-4dd4-ac99-794bc8fdbd07

dkwingsmt: Nvm, I just noticed that this place is for repositioning menus, and I think should probably have no animations at all.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -764,6 +779,7 @@ class _RawMenuAnchorState extends State<RawMenuAnchor> with _RawMenuAnchorBaseMi
 
   @override
   void handleCloseRequest() {
+    closeChildrenWithRequest();
```
dkwingsmt: I think we should call this after `widget.onCloseRequested` (i.e. the if clause below). The `onCloseRequested` callback is where themed menus change the animation status, and calling `closeChildrenWithRequest` later allows submenus to read the animation status and know what triggers the closure, and hence decide whether to display closing animation, which helps implementing the behavior here https://github.com/flutter/flutter/pull/182357/changes#r2824951515 .

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: Some menus do an async close. For example, Google Drive:

https://github.com/user-attachments/assets/d341cc80-e845-4382-b8b6-224469bbd532

dkwingsmt: I see. I guess we should somehow allow themed widget to customize this.

Edit: Actually, I think this is possible with my suggestion in https://github.com/flutter/flutter/pull/182357#discussion_r2830576068, which will allow submenus to know the cause of the closure, and decide the animation for themselves.

(Although I wonder a clearer way to explicitly tell this is possible, or needed at all)

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: I think this is reasonable. My only concern is that closeChildren begins at the topmost menu and travels to underlying menus. This would begin closing from the bottom menu and travel to the topmost menu. I'm not sure what implications this may have.

dkwingsmt: The full closing sequence is:
* Request propagates in (parent.requestClose, child.requestClose, grandchild.requestClose, ...)
* Close propagates out (..., grandchild.close, child.close, parent.close).

It makes perfect sense to me. What do you think?

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: Done

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: Done

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: 😿  It does make things a bit more complicated, but it is probably for the best.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: I think it makes sense... it doesn't seem to break anything so that's a good sign. I pushed the update implementing this.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
dkwingsmt: Also to be clear I mean that we should use `closeChildrenWithRequest` here since this allows children to use animation. They can choose to make the animation none.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -315,6 +315,12 @@ class RawMenuAnchor extends StatefulWidget {
   /// callback can be used to add a delay or a closing animation before the menu
   /// is hidden.
   ///
+  /// This callback is also triggered when a parent [RawMenuAnchor] is opened,
+  /// since that triggers [MenuController.close] on all descendant menu
```
dkwingsmt: Can you explain why `onCloseRequested` is also called on open? To my understanding, this is only a safeguard to ensure that menus are closed. If that's the case we should call `close` instead `requestClose`. Or do you think you actually want it to happen that:
* A menu is open.
* `MenuController.open` is called
* The menu closes with an animation, and opens with an animation (either at the same time or one after another)

davidhicks980: That's how we achieve this: https://github.com/flutter/flutter/pull/182357#discussion_r2857115524

To clarify, my explanation is incorrect and I pushed a fix. It now reads:  
```dart
  /// This callback is also triggered when a sibling [RawMenuAnchor] is opened
  /// while this menu is open. In this case, the callback can be used to add a
  /// delay or a closing animation while the sibling menu opens. When
  /// implementing this behavior, the closing menu should immediately stop being
  /// interactive, so that it doesn't interfere with the opening sibling menu.
  /// Semantics, focus, and hit testing for the closing menu should be disabled
  /// for the duration of the closing animation.
```

dkwingsmt: That makes much more sense.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -315,6 +315,11 @@ class RawMenuAnchor extends StatefulWidget {
   /// callback can be used to add a delay or a closing animation before the menu
   /// is hidden.
   ///
+  /// This callback is also triggered when a sibling [RawMenuAnchor] is opened.
+  /// Pending timers or animations previously started in a previous call to
+  /// [onCloseRequested] should be canceled when this callback is triggered to
+  /// prevent them from closing the menu at an unintended time.
+  ///
   /// If the menu is not closed, this callback will also be called when the root
   /// menu anchor is scrolled and when the screen is resized.
```
dkwingsmt: Given the current content, I think we should group the trigger timing of the callback together. Something like this: (you might want to rephrase it to reduce repetitiveness)
```dart
  /// This callback can be used to add a delay or a closing animation before the menu
  /// is hidden.
  ///
  /// This callback is triggered every time [MenuController.close] is called,
  /// regardless of whether the overlay is already hidden. 
  /// This callback is also triggered a sibling [RawMenuAnchor] is opened while this menu is open
  /// This callback will also be called when the root
  /// menu anchor is scrolled and when the screen is resized while this menu anchor is open.
```

And I suggest the following sentence be moved after the next paragraph, since the next paragraph mentions asynchronous operation (i.e. timer).

```
  /// Pending timers or animations previously started in a previous call to
  /// [onCloseRequested] should be canceled when this callback is triggered to
  /// prevent them from closing the menu at an unintended time.
```

davidhicks980: /// Called when a request is made to close the menu.
  ///
  /// This callback can be used to add a delay or a closing animation before the
  /// menu is hidden.
  ///
  /// This callback is triggered every time [MenuController.close] is called
  /// while this menu is open.
  ///
  /// This callback is also triggered when a sibling [RawMenuAnchor] is opened
  /// while this menu is open. In this case, the callback can be used to add a
  /// delay or a closing animation while the sibling menu opens. When
  /// implementing this behavior, the closing menu should immediately stop being
  /// interactive, so that it doesn't interfere with the opening sibling menu.
  /// Semantics, focus, and hit testing for the closing menu should be disabled
  /// for the duration of the closing animation.
  ///
  /// Pending timers or animations started in a previous call to
  /// [onCloseRequested] should be canceled when this callback is triggered to
  /// prevent them from closing the menu at an unintended time.
  ///
  /// If the menu is not closed, this callback will also be called when the root
  /// menu anchor is scrolled and when the screen is resized.
  ///
  /// After a close request is intercepted and closing behaviors have completed,
  /// the `hideOverlay` callback should be called. This callback sets
  /// [MenuController.isOpen] to false and hides the menu overlay widget. If the
  /// [RawMenuAnchor] is used in a themed menu that plays a closing animation,
  /// `hideOverlay` should be called after the closing animation has ended,
  /// since the animation plays on the overlay itself. This means that
  /// [MenuController.isOpen] will stay true while closing animations are
  /// running.
  ///
  /// Calling `hideOverlay` after disposal is a no-op, meaning it will not
  /// trigger [onClose] or hide the menu overlay.
  ///
  /// Typically, [onCloseRequested] consists of the following steps:
  ///
  ///  1. Optionally start the closing animation and wait for it to complete.
  ///  2. Call `hideOverlay` (whose call chain eventually invokes [onClose]).
  ///
  /// Throughout the closing sequence, menus should typically not be focusable
  /// or interactive.
  ///
  /// Defaults to a callback that immediately hides the menu.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -310,10 +310,23 @@ class RawMenuAnchor extends StatefulWidget {
...
+  /// This callback can be used to add a delay or a closing animation before the
+  /// menu is hidden.
+  ///
+  /// This callback is triggered every time [MenuController.close] is called
+  /// while this menu is open.
+  ///
+  /// This callback is also triggered when a sibling [RawMenuAnchor] is opened
+  /// while this menu is open. In this case, the callback can be used to add a
+  /// delay or a closing animation while the sibling menu opens. When
+  /// implementing this behavior, the closing menu should immediately stop being
```
dkwingsmt:
```suggestion
  /// implementing this behavior, consider ensuring the closing menu immediately stops being
```

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
```diff
@@ -310,10 +310,23 @@ class RawMenuAnchor extends StatefulWidget {
...
+  /// interactive so that it doesn't interfere with the opening sibling menu.
+  /// Semantics, focus, and hit testing for the closing menu should be disabled
+  /// for the duration of the closing animation.
```
dkwingsmt:
```suggestion
  /// Consider disabling semantics, focus, and hit testing for the closing menu
  /// for the duration of the closing animation.
```

dkwingsmt (dismissed):
LGTM. Thank you very much for the work!

dkwingsmt:
I wonder if @chunhtai would like to take a final look.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: I slightly modified it to  
```
/// When implementing this behavior, consider disabling interactions so that the
/// closing menu does not interfere with the opening sibling menu
```

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
davidhicks980: Done but I added a "Also," to the beginning.

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
chunhtai, line 771: I am slightly confused at the logic here. what happen if `SchedulerBinding.instance.schedulerPhase == SchedulerPhase.persistentCallbacks` and `inDispose == true`?

is there another code path the remove the overlay entry?

chunhtai: also curious why we need inDispose in the first place? I thought each menu and submenu's dispose will be called, so  if the dispose method only clean up the overlay entry this widget introduce should be enough?

chunhtai (commented):
just some questions left, other part lgtm

packages/flutter/lib/src/widgets/raw_menu_anchor.dart
chunhtai: > also curious why we need inDispose in the first place? I thought each menu and submenu's dispose will be called, so if the dispose method only clean up the overlay entry this widget introduce should be enough?

is this because 1 frame delay before the child overlay is removed? so if there are three opened submenu, it took 3 frame to completely remove them all?

davidhicks980: I'm fairly confident inDispose is actually unnecessary and exists only because it wasn't removed when the original MenuAnchor was converted from using OverlayEntry to OverlayPortal. It's on my list of things to remove in the future (it's removed on RawMenuNode, for example). 

When OverlayEntry was used, it wasn't automatically removed on dispose, so the original code had to go through each submenu and manually remove entries. I'm not sure why a post-frame callback wasn't used. When I first migrated the code from MenuAnchor to RawMenuAnchor, I tried to keep the logic the same so that no unintended consequences occurred. Now I'm more confident that it's okay to remove, but I was planning on doing it in a separate PR so that it's easy to revert.

chunhtai (approved):
LGTM

PR #182371, opened 2026-02-13
Implement popup windows for macOS

knopp:
*Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots.*

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
@@ -221,6 +221,13 @@ static char markerKey;
   __weak FlutterViewController* _currentViewController;
 
   /**
+   * The originally requested view controller. This may be attached to a window
+   * that is a descendant of _currentViewController window in case the window
+   * can not become a key window (i.e. popup window).
+   */
+  __weak FlutterViewController* _originalViewController;
+
+  /**
    * Used to obtain view controller on client creation
    */
   __weak id<FlutterTextInputPluginDelegate> _delegate;
@@ -432,6 +439,19 @@ static char markerKey;
         viewId = [(NSNumber*)requestViewId longLongValue];
       }
       _currentViewController = [_delegate viewControllerForIdentifier:viewId];
+      _originalViewController = _currentViewController;
+      while (!_currentViewController.view.window.canBecomeKeyWindow) {
+        NSWindow* parentWindow = [_currentViewController.view.window parentWindow];
+        if (parentWindow == nil) {
+          break;
+        }
+        NSViewController* controller = parentWindow.contentViewController;
+        if ([controller isKindOfClass:[FlutterViewController class]]) {
+          _currentViewController = (FlutterViewController*)controller;
+        } else {
+          break;
+        }
+      }
       FML_DCHECK(_currentViewController != nil);
     }
   } else if ([method isEqualToString:kShowMethod]) {
@@ -463,6 +483,7 @@ static char markerKey;
     _inputType = nil;
     _activeModel = nullptr;
     _currentViewController = nil;
+    _originalViewController = nil;
   } else if ([method isEqualToString:kSetEditingStateMethod]) {
     FML_DCHECK(_currentViewController != nil);
     NSDictionary* state = call.arguments;
@@ -993,7 +1014,7 @@ static char markerKey;
     farthest.y = MAX(farthest.y, y);
   }
 
-  const NSView* fromView = _currentViewController.flutterView;
+  const NSView* fromView = _originalViewController.flutterView;
   const CGRect rectInWindow = [fromView
       convertRect:CGRectMake(origin.x, origin.y, farthest.x - origin.x, farthest.y - origin.y)
            toView:nil];
@@ -1004,7 +1025,7 @@ static char markerKey;
 - (NSRect)firstRectForCharacterRange:(NSRange)range actualRange:(NSRangePointer)actualRange {
   // This only determines position of caret instead of any arbitrary range, but it's enough
   // to properly position accent selection popup
-  return !_currentViewController.viewLoaded || CGRectEqualToRect(_caretRect, CGRectNull)
+  return !_originalViewController.viewLoaded || CGRectEqualToRect(_caretRect, CGRectNull)
              ? CGRectZero
              : [self screenRectFromFrameworkTransform:_caretRect];
 }
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
@@ -658,4 +658,24 @@ void InternalFlutter_Window_UpdatePosition(void* window) {
   [owner updatePosition];
 }
 
+FlutterWindowOffset InternalFlutter_Window_GetOffsetInParent(void* window) {
+  NSWindow* w = (__bridge NSWindow*)window;
+  NSWindow* parent = w.parentWindow;
+  if (!parent) {
+    return {0, 0};
+  }
+  NSRect globalScreenFrame = ComputeGlobalScreenFrame();
+
+  NSRect parentRect = [parent contentRectForFrameRect:parent.frame];
+  FlipRect(parentRect, globalScreenFrame);
+
+  NSRect childRect = w.frame;
+  FlipRect(childRect, globalScreenFrame);
+
+  return {
+      .x = childRect.origin.x - parentRect.origin.x,
+      .y = childRect.origin.y - parentRect.origin.y,
+  };
+}
+
 // NOLINTEND(google-objc-function-naming)
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
@@ -427,7 +427,7 @@ static void FlipRect(NSRect& rect, const NSRect& globalScreenFrame) {
 
   window.collectionBehavior = NSWindowCollectionBehaviorAuxiliary;
   [parent addChildWindow:window ordered:NSWindowAbove];
-  // window.alphaValue = 0.0;
+  window.alphaValue = 0.0;
   return controller.viewIdentifier;
 }
 
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
@@ -244,6 +244,25 @@ static void FlipRect(NSRect& rect, const NSRect& globalScreenFrame) {
 
 @end
 
+@interface FlutterPopupWindow : NSPanel
+@end
+
+@implementation FlutterPopupWindow
+
+- (BOOL)canBecomeKeyWindow {
+  return NO;
+}
+
+- (BOOL)acceptsFirstResponder {
+  return NO;
+}
+
+- (BOOL)canBecomeMainWindow {
+  return NO;
+}
+
+@end
+
 @implementation FlutterWindowController
 
 - (instancetype)init {
@@ -361,6 +380,57 @@ static void FlipRect(NSRect& rect, const NSRect& globalScreenFrame) {
   return controller.viewIdentifier;
 }
 
+- (FlutterViewIdentifier)createPopupWindow:(const FlutterWindowCreationRequest*)request {
+  FlutterViewController* controller = [[FlutterViewController alloc] initWithEngine:_engine
+                                                                            nibName:nil
+                                                                             bundle:nil];
+  // By default this is kFlutterMouseTrackingModeInKeyWindow but popup window is never
+  // key window.
+  controller.mouseTrackingMode = kFlutterMouseTrackingModeInActiveApp;
+
+  NSWindow* window = [[FlutterPopupWindow alloc] init];
+  // If this is not set there will be double free on window close when
+  // using ARC.
+  [window setReleasedWhenClosed:NO];
+
+  window.contentViewController = controller;
+  window.styleMask = NSWindowStyleMaskBorderless;
+  window.hasShadow = NO;
+  window.opaque = NO;
+  window.backgroundColor = [NSColor clearColor];
+
+  FlutterWindowOwner* w = [[FlutterWindowOwner alloc] initWithWindow:window
+                                               flutterViewController:controller
+                                                     creationRequest:*request];
+
+  controller.flutterView.sizingDelegate = w;
+  [controller.flutterView setBackgroundColor:[NSColor clearColor]];
+  // Resend configure event after setting the sizing delegate.
+  [controller.flutterView constraintsDidChange];
+  w.closeWhenParentResignsKey = NO;
+
+  window.delegate = w;
+  [_windows addObject:w];
+
+  NSWindow* parent = nil;
+
+  if (request->parent_view_id != 0) {
+    for (FlutterWindowOwner* owner in _windows) {
+      if (owner.flutterViewController.viewIdentifier == request->parent_view_id) {
+        parent = owner.window;
+        break;
+      }
+    }
+  }
+
+  NSAssert(parent != nil, @"Popup window must have a parent window.");
+
+  window.collectionBehavior = NSWindowCollectionBehaviorAuxiliary;
+  [parent addChildWindow:window ordered:NSWindowAbove];
+  // window.alphaValue = 0.0;
+  return controller.viewIdentifier;
+}
+
 - (FlutterViewIdentifier)createRegularWindow:(const FlutterWindowCreationRequest*)request {
   FlutterViewController* controller = [[FlutterViewController alloc] initWithEngine:_engine
                                                                             nibName:nil
@@ -473,6 +543,14 @@ int64_t InternalFlutter_WindowController_CreateTooltipWindow(
   FlutterEngine* engine = [FlutterEngine engineForIdentifier:engine_id];
   [engine enableMultiView];
   return [engine.windowController createTooltipWindow:request];
+}
+
+int64_t InternalFlutter_WindowController_CreatePopupWindow(
+    int64_t engine_id,
+    const FlutterWindowCreationRequest* request) {
+  FlutterEngine* engine = [FlutterEngine engineForIdentifier:engine_id];
+  [engine enableMultiView];
+  return [engine.windowController createPopupWindow:request];
 }
 
 void InternalFlutter_Window_Destroy(int64_t engine_id, void* window) {
--- a/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
+++ b/engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
@@ -362,12 +362,11 @@ static void FlipRect(NSRect& rect, const NSRect& globalScreenFrame) {
 
   NSWindow* parent = nil;
 
-  if (request->parent_view_id != 0) {
-    for (FlutterWindowOwner* owner in _windows) {
-      if (owner.flutterViewController.viewIdentifier == request->parent_view_id) {
-        parent = owner.window;
-        break;
-      }
+  FML_DCHECK(request->parent_view_id != 0);
+  for (FlutterWindowOwner* owner in _windows) {
+    if (owner.flutterViewController.viewIdentifier == request->parent_view_id) {
+      parent = owner.window;
+      break;
     }
   }
 
@@ -414,12 +413,11 @@ static void FlipRect(NSRect& rect, const NSRect& globalScreenFrame) {
 
   NSWindow* parent = nil;
 
-  if (request->parent_view_id != 0) {
-    for (FlutterWindowOwner* owner in _windows) {
-      if (owner.flutterViewController.viewIdentifier == request->parent_view_id) {
-        parent = owner.window;
-        break;
-      }
+  FML_DCHECK(request->parent_view_id != 0);
+  for (FlutterWindowOwner* owner in _windows) {
+    if (owner.flutterViewController.viewIdentifier == request->parent_view_id) {
+      parent = owner.window;
+      break;
     }
   }
 
--- a/examples/multiple_windows/lib/app/element_position_tracker.dart
+++ b/examples/multiple_windows/lib/app/element_position_tracker.dart
@@ -14,9 +14,6 @@ class ElementPositionTracker {
     _ElementPositionTrackerManager.instance.remove(this);
   }
 
-  final BuildContext element;
-  Rect? _lastReportedRect;
-
   /// Returns current global rect for tracked element or `null` if not available.
   Rect? getGlobalRect() {
     final rect = _getGlobalRect();
@@ -27,6 +24,9 @@ class ElementPositionTracker {
   /// Callback invoked every time the global position of the tracked element changes
   /// compared to last result of [getGlobalRect].
   void Function(Rect rect)? onGlobalRectChange;
+
+  final BuildContext element;
+  Rect? _lastReportedRect;
 
   Rect? _getGlobalRect() {
     if (!element.mounted) {
--- a/examples/multiple_windows/lib/app/element_position_tracker.dart
+++ b/examples/multiple_windows/lib/app/element_position_tracker.dart
@@ -14,7 +14,7 @@ class ElementPositionTracker {
     _ElementPositionTrackerManager.instance.remove(this);
   }
 
-  /// Returns current global rect for tracked element or `null` if not available.
+  /// Returns current global rect for the tracked element, or `null` if not available.
   Rect? getGlobalRect() {
     final rect = _getGlobalRect();
     _lastReportedRect = rect;
@@ -51,8 +51,8 @@ class ElementPositionTracker {
     }
     if (_lastReportedRect != rect) {
       _lastReportedRect = rect;
+      onGlobalRectChange?.call(rect);
     }
-    onGlobalRectChange?.call(rect);
   }
 }
 
--- a/examples/multiple_windows/lib/app/element_position_tracker.dart
+++ b/examples/multiple_windows/lib/app/element_position_tracker.dart
@@ -1,0 +1,88 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
+
+/// Tracks global rect of an [Element].
+class ElementPositionTracker {
+  ElementPositionTracker({required this.element}) {
+    _ElementPositionTrackerManager.instance.add(this);
+  }
+
+  void dispose() {
+    _ElementPositionTrackerManager.instance.remove(this);
+  }
+
+  final BuildContext element;
+  Rect? _lastReportedRect;
+
+  /// Returns current global rect for tracked element or `null` if not available.
+  Rect? getGlobalRect() {
+    final rect = _getGlobalRect();
+    _lastReportedRect = rect;
+    return rect;
+  }
+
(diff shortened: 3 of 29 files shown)


engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
```diff
@@ -361,6 +380,57 @@ - (FlutterViewIdentifier)createTooltipWindow:(const FlutterWindowCreationRequest
...
+  // Resend configure event after setting the sizing delegate.
+  [controller.flutterView constraintsDidChange];
+  w.closeWhenParentResignsKey = NO;
+
+  window.delegate = w;
+  [_windows addObject:w];
+
+  NSWindow* parent = nil;
+
+  if (request->parent_view_id != 0) {
```
mattkae: A popup window must always have a parent, right? So this check may be unnecessary

mattkae (dismissed):
I have a single nit, but looks reasonable to me. Perhaps @justinmc or @loic-sharma may have more to say than me about the API changes

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -137,11 +137,21 @@ class WindowingOwnerMacOS extends WindowingOwner {
   PopupWindowController createPopupWindowController({
     required PopupWindowControllerDelegate delegate,
     required BoxConstraints preferredConstraints,
+    required bool isSizedToContent,
     required Rect anchorRect,
     required WindowPositioner positioner,
     required BaseWindowController parent,
   }) {
-    throw UnimplementedError('Popup windows are not yet implemented on MacOS.');
+    final res = PopupWindowControllerMacOS(
```
justinmc: Nit: See https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-abbreviations

What does `res` mean? Maybe just make it `controller`.

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -346,6 +356,108 @@ class TooltipWindowControllerMacOS extends TooltipWindowController with _WindowC
   Rect _anchorRect;
 }
 
+/// MacOS specific implementation of [PopupWindowController].
+///
+/// /// {@macro flutter.widgets.windowing.experimental}
```
justinmc: No need for the duplicate `///`.

examples/multiple_windows/lib/app/popup_button.dart
```diff
@@ -0,0 +1,199 @@
...
+    );
+  }
+}
+
+class _PopupWindowControllerDelegate extends PopupWindowControllerDelegate {
+  _PopupWindowControllerDelegate({required this.onDestroyed});
+
+  @override
+  void onWindowDestroyed() {
+    onDestroyed();
```
justinmc, line 111: Nit: It might be too late to change this, but see here about naming this kind of stuff: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

I think that means onWindowDestroyed should have been handleWindowDestroyed.

examples/multiple_windows/lib/app/main_window.dart
justinmc, line 1: Can you test the new changes to this example in examples/multiple_windows/test?

examples/multiple_windows/lib/app/popup_button.dart
```diff
@@ -0,0 +1,199 @@
...
+import 'models.dart';
+
+/// Manages all [_ElementPositionTracker] instances.
+///
+/// This class is a singleton because it needs to hook into the global frame
+/// callbacks to update all registered trackers. This callback may only be
+/// registered once, so we use a singleton to ensure that.
+class _ElementPositionTrackerManager {
+  _ElementPositionTrackerManager._() {
+    WidgetsBinding.instance.addPersistentFrameCallback((_) {
```
justinmc: So every frame, you look up the location of the popup window and report it to macos? Is that in order to keep the popup menu positioned on the correct place on the screen relative to the window that it came from?

justinmc: I think this would be easier to understand if I could play around with the example, but I'm getting an error (see main review text).

examples/multiple_windows/lib/app/popup_button.dart
```diff
@@ -0,0 +1,199 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+
+import 'window_content.dart';
+import 'models.dart';
+
+/// Manages all [_ElementPositionTracker] instances.
+///
+/// This class is a singleton because it needs to hook into the global frame
```
justinmc: I'm always wary of global singletons, are you sure this one is necessary or is there another way to do this?

justinmc: Sometimes a way around using a singleton is to use an InheritedWidget at the root of the tree instead. I'm not sure if that would be practical here or not.

examples/multiple_windows/lib/app/popup_button.dart
justinmc, line 1: My initial impression is that this is quite a lot of boilerplate for a user to set up in order to use a popup window. Do we have any plans in the longer term to put some of this code into Flutter for users to use?

justinmc (commented):
I tried running the example on my mac but I got an error when I clicked on "Show Popup". Maybe it's just because I haven't compiled the engine on this branch, I'll try that.

<details>

<summary>Error message</summary>

```

══╡ EXCEPTION CAUGHT BY GESTURE ╞═══════════════════════════════════════════════════════════════════
The following ArgumentError was thrown while handling a gesture:
Invalid argument(s): Couldn't resolve native function
'InternalFlutter_WindowController_CreatePopupWindow' in
'package:flutter/src/widgets/_window_macos.dart' : No asset with id
'package:flutter/src/widgets/_window_macos.dart' found. No available native assets. Attempted to
fallback to process lookup. dlsym(RTLD_DEFAULT, InternalFlutter_WindowController_CreatePopupWindow):
symbol not found.
```

</details>

examples/multiple_windows/lib/app/popup_button.dart
knopp: The crash - You need to run it with local engine. 

The tracker manager - I think this is actually just copied from the tooltip_button.dart, now that I'm looking into that. This should be shared.

The idea is that the Controller API itself only gets an anchor rect, it shouldn't concern itself with position tracking. That's a completely orthogonal problem. The tracker in example is something I built quickly to get the example running and it stuck there. It's by no means the best way to solve the problem.

examples/multiple_windows/lib/app/popup_button.dart
knopp: I moved the `ElementPositionTracker` to separate file so that it's not copied for each button and simplified the API. The singleton (`_ElementPositionTrackerManager`) is private and the reason for existence is to only register one persistent frame callback for all trackers, because you can't unregister persistent callbacks.

I don't think the tracker itself is too verbose (89 lines). At this point I'm not sure that's the best way is to track the anchor position, that's why the tracker is part of the example instead of framework.

examples/multiple_windows/lib/app/popup_button.dart
knopp: The `_ElementPositionTrackerManager` is  now just an implementation detail of `ElementPositionTracker`. It is a singleton because I don't want to register more than one persistent frame callback and there's no way to unregister them.

I don't think I'd want to pass this down widget hierarchy - the user shouldn't be concerned with it at all.

examples/multiple_windows/lib/app/popup_button.dart
knopp: > So every frame, you look up the location of the popup window and report it to macos? Is that in order to keep the popup menu positioned on the correct place on the screen relative to the window that it came from?

Yes, on every frame the position gets updated. This is to make sure the popup tracks the anchor even if it moves (i.e. during window resizing).

examples/multiple_windows/lib/app/popup_button.dart
knopp: I don't think we care much about not breaking things right now. I think we can rename it.  @mattkae ?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -220,6 +220,13 @@ @interface FlutterTextInputPlugin () {
    */
   __weak FlutterViewController* _currentViewController;
 
+  /**
+   * The originally requested view controller. This may be attached to a window
+   * that is a descendant of _currentViewController window in case the window
+   * can not become a key window (i.e. popup window).
```
loic-sharma, line 226: Could we update `_currentViewController`'s docs to call out this popup window case?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -432,6 +439,19 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+        if (parentWindow == nil) {
+          break;
+        }
+        NSViewController* controller = parentWindow.contentViewController;
+        if ([controller isKindOfClass:[FlutterViewController class]]) {
+          _currentViewController = (FlutterViewController*)controller;
+        } else {
+          break;
+        }
+      }
```
loic-sharma, line 454: Just to double check: is it possible for a developer to write an app that creates a window that has no parent but also cannot become a key window? 

I'm a bit nervous with the `FML_DCHECK` immediately below this line. I'm wondering if we should be a bit more defensive and return a `FlutterError` result instead? This would ensure we don't crash the app in this scenario, while still reporting an error to the developer.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
loic-sharma: FYI this comment is still relevant but likely got lost by GitHub's collapsed timeline: https://github.com/flutter/flutter/pull/182371/changes#r2865924154

examples/multiple_windows/lib/app/element_position_tracker.dart
```diff
@@ -0,0 +1,88 @@
...
+class ElementPositionTracker {
+  ElementPositionTracker({required this.element}) {
+    _ElementPositionTrackerManager.instance.add(this);
+  }
+
+  void dispose() {
+    _ElementPositionTrackerManager.instance.remove(this);
+  }
+
+  /// Returns current global rect for tracked element or `null` if not available.
```
loic-sharma:
```suggestion
  /// Returns current global rect for the tracked element, or `null` if not available.
```

examples/multiple_windows/lib/app/element_position_tracker.dart
```diff
@@ -0,0 +1,88 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
+
+/// Tracks global rect of an [Element].
```
loic-sharma:
```suggestion
/// Tracks the global rect of an [Element].
```

examples/multiple_windows/lib/app/element_position_tracker.dart
```diff
@@ -0,0 +1,88 @@
...
+  void _updateSelf() {
+    final rect = _getGlobalRect();
+    if (rect == null) {
+      _ElementPositionTrackerManager.instance.remove(this);
+      return;
+    }
+    if (_lastReportedRect != rect) {
+      _lastReportedRect = rect;
+    }
+    onGlobalRectChange?.call(rect);
```
loic-sharma: Should this line be in the `if` block above? It looks like we're always calling `onGlobalRectChange` currently, even if `_lastReportedRect` was correct.

If this is intentional - and I'm not misunderstanding something silly - please add a comment.

examples/multiple_windows/lib/app/popup_button.dart
loic-sharma: Ah TIL about that naming convention! These APIs are still experimental, we can break things willy nilly. Feel free to do this in a follow-up pull request though.

examples/multiple_windows/lib/app/popup_window_content.dart
```diff
@@ -0,0 +1,123 @@
...
+                width: 200,
+                decoration: BoxDecoration(
+                  color: const Color(0x33FFFFFF),
+                  borderRadius: BorderRadius.circular(8),
+                ),
+                padding: const EdgeInsets.symmetric(
+                  horizontal: 12,
+                  vertical: 8,
+                ),
+                child: EditableText(
```
loic-sharma: Can we use `TextField` here instead of `EditableText`?

I believe text selection, magnifier, and gestures won't work properly here. You'd get all that for free with `TextField`.

examples/multiple_windows/lib/app/regular_window_content.dart
```diff
@@ -94,6 +96,8 @@ class RegularWindowContent extends StatelessWidget {
                 const SizedBox(height: 20),
                 TooltipButton(parentController: window),
                 const SizedBox(height: 20),
+                PopupButton(parentController: window),
+                const SizedBox(height: 20),
```
loic-sharma, line 102: Can we use `Column.spacing` here? I don't feel strongly, feel free to keep as-is.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -859,35 +861,33 @@ abstract class PopupWindowController extends BaseWindowController {
...
 
-  /// Request change to the constraints of the window.
-  ///
-  /// The [constraints] describes the new constraints that the window should
-  /// satisfy. If the constraints disagree with the current size of the window,
-  /// the platform might resize the window to satisfy the new constraints.
+  /// Returns the offset of popup's top left corner in parent window client area.
   ///
-  /// The platform is free to ignore this request.
+  /// The offset is in logical coordinates with top left origin.
```
loic-sharma: Is this comment strictly necessary? I think that's the assumption for all Flutter APIs. I'd consider removing this, but I don't feel strongly.

If you keep this, consider:

```suggestion
  /// The offset is in logical coordinates with top-left origin.
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -859,35 +861,33 @@ abstract class PopupWindowController extends BaseWindowController {
...
   @internal
-  bool get isActivated;
+  void setConstraints(BoxConstraints constraints);
 
-  /// Requests that the window receive focus.
+  /// Updates the position of the popup.
   ///
-  /// The platform may also give the window input focus and bring it to the
-  /// top of the window stack. However, this behavior is platform-dependent.
+  /// This requests that the popup be repositioned according to the new [anchorRect] and/or [positioner].
```
loic-sharma, line 899: Does this need the "platform is free to ignore" disclaimer?

```suggestion
  /// This requests that the popup be repositioned according to the new [anchorRect] and/or [positioner].
  ///
  /// The platform is free to ignore this request.
```

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -639,6 +751,23 @@ final class _Size extends Struct {
...
+  external double y;
+
+  @override
+  String toString() {
+    return 'Offset(x: $x, y: $y)';
+  }
+
+  Offset toOffset() {
+    return Offset(x, y);
+  }
```
loic-sharma: Consider switching to `=>` syntax. I don't feel strongly, feel free to skip:

```suggestion
  @override
  String toString() => 'Offset(x: $x, y: $y)';

  Offset toOffset() => Offset(x, y);
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -827,13 +827,15 @@ abstract class PopupWindowController extends BaseWindowController {
     required Rect anchorRect,
     required WindowPositioner positioner,
     BoxConstraints? preferredConstraints,
+    bool isSizedToContent = true,
```
loic-sharma: Is this used? It looks like we should remove this.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -859,35 +861,33 @@ abstract class PopupWindowController extends BaseWindowController {
...
-  /// Requests that the window receive focus.
+  /// Updates the position of the popup.
   ///
-  /// The platform may also give the window input focus and bring it to the
-  /// top of the window stack. However, this behavior is platform-dependent.
+  /// This requests that the popup be repositioned according to the new [anchorRect] and/or [positioner].
   ///
   /// {@macro flutter.widgets.windowing.experimental}
   @internal
-  void activate();
```
loic-sharma, line 909: Could you explain why we are removing the `activate()` and `isActivated` APIs? That seems useful for popup windows since they can receive input focus.

packages/flutter/test/widgets/windowing_test.dart
```diff
@@ -848,7 +848,7 @@ void main() {
           ),
         );
 
-        expect(isActivated, equals(true));
+        expect(isActivated, equals(false));
```
loic-sharma: If we re-add activated APIs (see [this comment](https://github.com/flutter/flutter/pull/182371/changes#r2943546836)), don't forget to undo this.

packages/flutter_test/lib/src/binding.dart
```diff
@@ -746,8 +754,7 @@ class _TestWindowingOwner extends WindowingOwner {
         dialogParent.activate();
       case final TooltipWindowController _:
         fail('TooltipWindowController cannot be a parent of DialogWindowController.');
-      case final PopupWindowController popupParent:
-        popupParent.activate();
+      case final PopupWindowController _:
```
loic-sharma: If we re-add activate APIs, don't forget to undo this. If we remove activate APIs, should we add an error message here?

Also, is the `fail` message for the `TooltipWindowController` case above correct? It seems to incorrectly assume that the child is `DialogWindowController`.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
knopp: Why would the `FML_DCHECK` fail in this case? Unless I'm reading the code wrong it would only fail if `[_delegate viewControllerForIdentifier:viewId]` returns `nil`, which is definitely not expected.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterWindowController.mm
knopp: Thanks! Missed this one. Will turn it into a DCHECK for both tooltip and popup.

examples/multiple_windows/lib/app/element_position_tracker.dart
knopp: This really seems like it should be in the if block.

packages/flutter/lib/src/widgets/_window.dart
knopp: cc @mattkae

packages/flutter/lib/src/widgets/_window.dart
knopp: The popup window is never activated. It is always the parent window that is active. Active popup window basically means a focused `FocusNode` inside popup window view hierarchy. We have no insight into that in the controller. 

Because popup window is conceptually an extension of parent window, I think maybe we could consider forwarding `activate()` and `isActive()` into parent window controller. @mattkae @Saviq?

Saviq: Yeah, I think that makes sense, if we want to bring consistency here.

packages/flutter/lib/src/widgets/_window.dart
knopp: @Saviq, @mattkae

packages/flutter/lib/src/widgets/_window.dart
knopp: Not sure. Window positioning coordinates are tricky in general because on windows it is in physical pixels with no good way of converting to logical pixels. It can be done in this case because the windows are likely on same screen, but if we ever bring any window positioning API we'll need to deal with coordinate systems. 

The top left origin might be necessary, since we can flip macOS coordinate system and hide that insanity from users.

packages/flutter/lib/src/widgets/_window.dart
mattkae: It should be used since popups can be sized to content, no?

examples/multiple_windows/lib/app/regular_window_content.dart
knopp: We probably should, but maybe in another PR? @mattkae

packages/flutter/lib/src/widgets/_window.dart
knopp: I am a bit confused here. I didn't notice this after rebase. Right now at least on macOS popups and tooltips are always sized to content. If you want to specify size manually you can use tight constraints. What exactly is the purpose of this?

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
loic-sharma: Ah I read this wrong. This looks good to me!

(I was under the incorrect impression that `_currentViewController` would be `nil` if we didn't find a window that could become key)

examples/multiple_windows/lib/app/regular_window_content.dart
loic-sharma: I'm fine with punting 👍

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: Thanks for the details, that makes sense!

packages/flutter/lib/src/widgets/_window.dart
knopp: I will be removing `isSizedToContent`.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -859,35 +861,77 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  void updatePosition({Rect? anchorRect, WindowPositioner? positioner});
 
-  /// Request change to the constraints of the window.
+  /// Returns the offset of popup's top-left corner in parent window client area.
```
loic-sharma:
```suggestion
  /// Returns the offset of the popup's top-left corner in the parent window client area.
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -859,35 +861,77 @@ abstract class PopupWindowController extends BaseWindowController {
...
   ///
-  /// The [constraints] describes the new constraints that the window should
-  /// satisfy. If the constraints disagree with the current size of the window,
-  /// the platform might resize the window to satisfy the new constraints.
+  /// The offset is in logical coordinates.
   ///
-  /// The platform is free to ignore this request.
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  Offset get offsetInParent;
```
loic-sharma: What do y'all think of the names `parentOffset` or `offsetFromParent`?

Our gestures use names like:

* [TapDragUpdateDetails.offsetFromOrigin](https://api.flutter.dev/flutter/gestures/TapDragUpdateDetails/offsetFromOrigin.html)
* [LongPressMoveUpdateDetails.offsetFromOrigin](https://api.flutter.dev/flutter/gestures/LongPressMoveUpdateDetails/offsetFromOrigin.html)

knopp: I'm not sure. `parentOffset` sounds to me like "parent offset relative to child", which would be this inverted, `offsetFromParent` lack the anchor point. Is it offset from parent origin? parent edge? 

To me Offset**In**Parent implies that the offset is relative to parent content area. But I might be wrong about this. @mattkae, any ideas?

packages/flutter/lib/src/widgets/_window.dart
Saviq: Platform is free to ignore anything, in the end, so no ;)

packages/flutter/lib/src/widgets/_window.dart
Saviq: I'd be fine with `offsetFromParent`, `(0,0)` is the implicit enough anchor? I was actually interpreting `offsetInParent` as relative to the positioning anchor point (offset from something _in_ parent), so I don't think that's better.

knopp: I don't feel too strongly about this.

mattkae (dismissed):
Good by me!

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: If we can implement this API reliably on all desktop platforms, please skip the disclaimer.

However please add this disclaimer if:

1. One or more platforms ignore this request completely. Hypothetical example: if this method no-ops on Wayland.
2. There are common scenarios where this request will be ignored. Hypothetical example: if this method no-ops on Windows if the parent window is a satellite window.

If we do need the disclaimer, ideally the disclaimer would explain the scenarios in which this API is known to misbehave.

packages/flutter_test/lib/src/binding.dart
knopp: The activation has been restored.

engine/src/flutter/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
loic-sharma: Thanks!

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -364,6 +372,108 @@ class TooltipWindowControllerMacOS extends TooltipWindowController with _WindowC
...
+  @override
+  void setConstraints(BoxConstraints constraints) {
+    _ensureNotDestroyed();
+    _MacOSPlatformInterface.setWindowConstraints(getWindowHandle(), constraints);
+  }
+
+  final PopupWindowControllerDelegate _delegate;
+  final BaseWindowController _parent;
+  WindowPositioner _positioner;
+  Rect _anchorRect;
```
loic-sharma, line 474: Nit: Please put these fields after the constructor but before the subsequent methods.

See this bit in the style guide: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -364,6 +372,108 @@ class TooltipWindowControllerMacOS extends TooltipWindowController with _WindowC
   Rect _anchorRect;
 }
 
+/// MacOS specific implementation of [PopupWindowController].
+///
+/// {@macro flutter.widgets.windowing.experimental}
+class PopupWindowControllerMacOS extends PopupWindowController with _WindowControllerMixin {
```
loic-sharma, line 378: Just checking - is this automatically `@internal` since it extends `BaseWindowController`? If not, could we add an `@internal` attribute here plus all the other platform-specific controller types?

See: [flutter.dev/go/multi-window-experimental-apis](https://docs.flutter.dev/go/multi-window-experimental-apis)
Sorry, I missed this earlier

loic-sharma (dismissed):
LGTM once this comment is resolved: https://github.com/flutter/flutter/pull/182371/changes#r2897119330

mattkae (dismissed):
LGTM!

mattkae (approved):

loic-sharma (approved):
Re-LGTM

Issue #144397, opened 2024-02-29
macOS build failures are not parsed for common causes


vashworth:
When a macOS build fails, we don't parse the build output for common causes like we do with iOS. We should do something similar for macOS, or better yet - unify the two to use the same build method and do error parsing within it.

macOS:
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/macos/build_macos.dart#L136-L163
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/macos/macos_device.dart#L69-L80
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/commands/build_macos.dart#L62-L76

iOS:
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/ios/devices.dart#L489-L503
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/commands/build_ios.dart#L662-L681
https://github.com/flutter/flutter/blob/8a11703c6cfc1269aa9ef2789621177f7fed0982/packages/flutter_tools/lib/src/ios/simulators.dart#L565-L575

PR #182375, opened 2026-02-13
Improve SwiftPM minimum platform mismatch diagnostics

### LINKED ISSUE #165420: [SwiftPM] Improve the error message if a plugin requires a higher deployment target version
# Background

SwiftPM strictly enforces that your iOS target version is equal to or higher than your dependencies' iOS target version.

If your app target has a minimum deployment of iOS 12 but a SwiftPM plugin requires iOS 15, you'll get an error like: https://github.com/flutter/flutter/issues/162196#issuecomment-2734348292

```
Failed to build iOS app
Could not build the precompiled application for the device.
Target Integrity (Xcode): The package product 'cloud-firestore' requires minimum platform version 13.0 for the iOS platform, but this target supports 12.0

Target Integrity (Xcode): The package product 'firebase-core' requires minimum platform version 13.0 for the iOS platform, but this target supports 12.0
```

The solution to this problem is documented here: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers#how-to-use-a-swift-package-manager-flutter-plugin-that-requires-a-higher-os-version

### Suggestion

The Flutter CLI should catch this error and provide a more helpful error message. For example:

```
$ flutter run
Failed to build iOS app

Target Integrity (Xcode): The package product 'cloud-firestore' requires minimum platform version 13.0 for the iOS platform, but this target supports 12.0

To fix this, raise your app's minimum platform version or remove this dependency.

To increase your app's minimum platform version:

1. Open your app (ios/Runner.xcworkspace or macos/Runner.xcworkspace) in Xcode.
2. In the Project Navigator, select the Runner project > Runner target > General tab
3. Increase your app's target Minimum Deployments setting.
4. If you updated your iOS app's Minimum Deployments, regenerate the iOS project's configuration files:

   flutter build ios --config-only

5. If you updated your macOS app's Minimum Deployments, regenerate the macOS project's configuration files:

   flutter build macos --config-only
```

Part of https://github.com/flutter/flutter/issues/126005

MohammedTarigg:
Fixes #165420.

When using Swift Package Manager plugins, Xcode can fail with a minimum platform mismatch (a package requires a higher iOS/macOS version than the target supports).

This change detects that failure during iOS and macOS builds and prints actionable guidance about increasing the app's minimum deployment target, including which Flutter config-only commands to re-run.

Tests:
- dart analyze (packages/flutter_tools)
- dart test test/general.shard/ios/mac_test.dart
- dart test test/commands.shard/hermetic/build_macos_test.dart

Local E2E repro:
- Performed locally with throwaway fixtures (not checked in).


packages/flutter_tools/lib/src/ios/mac.dart
vashworth, line 1267: I think we want to move the parsing logic to here because the Xcresult does not give the full error message. For example, the xcresult may say `The package product 'firebase-core' requires minimum platform version 15.0 for the iOS platform, but this target supports 13.0`, but the error in stdout is `error: The package product 'firebase-core' requires minimum platform version 15.0 for the iOS platform, but this target supports 13.0 (in target 'FlutterGeneratedPluginSwiftPackage' from project 'FlutterGeneratedPluginSwiftPackage')`

I think we want to give a different message dependency on the target. If the target is `FlutterGeneratedPluginSwiftPackage`, then we can give a message with instructions to upgrade their app minimum version. If it's not, we should probably check if the target is a plugin and then instruct the user to file an issue with the plugin - we could do this in a separate PR, though.

packages/flutter_tools/lib/src/macos/build_macos.dart
vashworth, line 1: Let's keep this fix scoped to iOS for now. We have a different issue on making iOS and macOS use shared logic, and it'd be better to have a more robust solution: https://github.com/flutter/flutter/issues/144397

MohammedTarigg:
Thanks, that makes sense. I scoped this PR to iOS only and moved the minimum-platform detection to iOS stdout parsing with target-based handling.

packages/flutter_tools/lib/src/macos/build_macos.dart
MohammedTarigg: Sounds good. If it’s okay with you, I’d like to take #144397 next and work on the more robust shared iOS/macOS approach next.

MohammedTarigg:
@vashworth In parseVendoredFrameworksFromPbxproj, we currently only read direct children of PBXGroups named Frameworks and then pick PBXFileReferences.
If a pod/project has nested PBXGroups under Frameworks, we might miss those framework refs. Do you think this is worth handling now (recursive traversal), or should we keep current behavior unless we hit a concrete case?

packages/flutter_tools/lib/src/ios/mac.dart
MohammedTarigg: I updated this PR to parse the minimum-platform mismatch from stdout (where the full target context is available), and now I only show the app minimum-version guidance when the target is FlutterGeneratedPluginSwiftPackage.
For non-FlutterGeneratedPluginSwiftPackage targets, I’m not showing that app guidance in this PR. I can follow up with plugin-target detection and a plugin-issue instruction in a separate PR if you’d like.

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
...
+) {
+  if (stdout == null || stdout.isEmpty) {
+    return const <_SwiftPackageManagerMinPlatformMismatch>[];
+  }
+
+  final pattern = RegExp(
+    r"(?:error:\s*)?The package product '[^']+' requires minimum platform version "
+    r'([0-9]+(?:\.[0-9]+)*) for the (iOS|macOS) platform, but this target supports '
+    r"([0-9]+(?:\.[0-9]+)*)(?: \(in target '([^']+)' from project '[^']+'\))?",
+    caseSensitive: false,
```
vashworth:
```suggestion
    r"The package product '([^']+)' requires minimum platform version "
    r'([0-9\.]+) for the (iOS|macOS) platform, but this target supports '
    r"([0-9\.]+)(?: \(in target '([^']+)' from project '[^']+'\))?",
    caseSensitive: false,
```

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
...
+  final mismatches = <_SwiftPackageManagerMinPlatformMismatch>[];
+  for (final RegExpMatch match in pattern.allMatches(stdout)) {
+    final Version? requiredMinVersion = Version.parse(match.group(1));
+    final Version? targetSupportedVersion = Version.parse(match.group(3));
+    if (requiredMinVersion == null || targetSupportedVersion == null) {
+      continue;
+    }
+
+    mismatches.add(
+      _SwiftPackageManagerMinPlatformMismatch(
```
vashworth: Instead of using a class, you can just return the message directly

```dart
return 
"To fix this error, increase your app's minimum platform version from $supportedVersion to at least "
'$highestRequiredVersion or remove the $highestRequiredByProduct dependency.\n\n'
"To increase your app's minimum platform version, follow these instructions:\n"
'  https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers#how-to-use-a-swift-package-manager-flutter-plugin-that-requires-a-higher-os-version';
```

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
...
+    r"(?:error:\s*)?The package product '[^']+' requires minimum platform version "
+    r'([0-9]+(?:\.[0-9]+)*) for the (iOS|macOS) platform, but this target supports '
+    r"([0-9]+(?:\.[0-9]+)*)(?: \(in target '([^']+)' from project '[^']+'\))?",
+    caseSensitive: false,
+  );
+
+  final mismatches = <_SwiftPackageManagerMinPlatformMismatch>[];
+  for (final RegExpMatch match in pattern.allMatches(stdout)) {
+    final Version? requiredMinVersion = Version.parse(match.group(1));
+    final Version? targetSupportedVersion = Version.parse(match.group(3));
```
vashworth:
```suggestion
    final String? requiredByProduct = match.group(1);
    final Version? requiredMinVersion = Version.parse(match.group(2));
    final Version? targetSupportedVersion = Version.parse(match.group(4));
    final String? targetName = match.group(5);
```

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1080,6 +1084,22 @@ Future<bool> _handleIssues(
     logger.printError("Also try selecting 'Product > Build' to fix the problem.");
   } else if (missingPlatform != null) {
     logger.printError(missingPlatformInstructions(missingPlatform), emphasis: true);
+  } else if (swiftPackageManagerMinPlatformMismatches.isNotEmpty) {
+    final Version requiredMinVersion = swiftPackageManagerMinPlatformMismatches
+        .map((e) => e.requiredMinVersion)
+        .reduce((a, b) => a > b ? a : b);
+    final Version supportedVersion = swiftPackageManagerMinPlatformMismatches
+        .map((e) => e.targetSupportedVersion)
+        .reduce((a, b) => a < b ? a : b);
```
vashworth: Instead of reducing, just keep track of the highest required version in `_parseSwiftPackageManagerMinPlatformMismatches` and return a single record.

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
   return pattern.firstMatch(message)?.group(1);
 }
 
+const _flutterGeneratedPluginSwiftPackageTarget = 'FlutterGeneratedPluginSwiftPackage';
```
vashworth: Use `kFlutterGeneratedPluginSwiftPackageName` instead imported from swift_package_manager.dart

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
...
+  );
+
+  final mismatches = <_SwiftPackageManagerMinPlatformMismatch>[];
+  for (final RegExpMatch match in pattern.allMatches(stdout)) {
+    final Version? requiredMinVersion = Version.parse(match.group(1));
+    final Version? targetSupportedVersion = Version.parse(match.group(3));
+    if (requiredMinVersion == null || targetSupportedVersion == null) {
+      continue;
+    }
+
```
vashworth:
```suggestion
if (targetName != kFlutterGeneratedPluginSwiftPackageName) {
  continue;
}
```

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1221,6 +1241,69 @@ String? _parseMissingPlatform(String message) {
...
+  final pattern = RegExp(
+    r"(?:error:\s*)?The package product '[^']+' requires minimum platform version "
+    r'([0-9]+(?:\.[0-9]+)*) for the (iOS|macOS) platform, but this target supports '
+    r"([0-9]+(?:\.[0-9]+)*)(?: \(in target '([^']+)' from project '[^']+'\))?",
+    caseSensitive: false,
+  );
+
+  final mismatches = <_SwiftPackageManagerMinPlatformMismatch>[];
+  for (final RegExpMatch match in pattern.allMatches(stdout)) {
```
vashworth, line 1326: As you loop through, keep track of the highest version and then just return the message using that one. That'll reduce the number of loops you're doing.

packages/flutter_tools/lib/src/macos/build_macos.dart
vashworth: Sure!

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1128,6 +1130,9 @@ Future<bool> _handleIssues(
     logger.printError("Also try selecting 'Product > Build' to fix the problem.");
   } else if (missingPlatform != null) {
     logger.printError(missingPlatformInstructions(missingPlatform), emphasis: true);
+  } else if (swiftPackageManagerMinPlatformMismatchMessage != null) {
+    logger.printError(swiftPackageManagerMinPlatformMismatchMessage, emphasis: true);
+    issueDetected = true;
```
hellohuanlin, line 1137: why is `issueDetected` set here, but not in the other cases below?

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1298,6 +1303,53 @@ String? _parseMissingPlatform(String message) {
...
+  final pattern = RegExp(
+    r"The package product '([^']+)' requires minimum platform version "
+    r'([0-9\.]+) for the (iOS|macOS) platform, but this target supports '
+    r"([0-9\.]+)(?: \(in target '([^']+)' from project '[^']+'\))?",
+    caseSensitive: false,
+  );
+
+  String? highestRequiredByProduct;
+  Version? highestRequiredVersion;
+  Version? highestSupportedVersion;
```
hellohuanlin, line 1325: worth documenting why we only find the highest supported version here. (i assume it's because we are asking developers to bump their minimum version to at least as high as the highest requirement of the plugins)

hellohuanlin (approved):

packages/flutter_tools/lib/src/ios/mac.dart
MohammedTarigg: We set issueDetected = true here because this error is found from stdout, not from xcresult.
So without setting it, we might print the generic fallback output again.

For the other cases below, issueDetected was already set earlier when we parsed xcresult, so we don’t need to set it again.

packages/flutter_tools/lib/src/ios/mac.dart
MohammedTarigg: Yes You’re right.
If multiple plugins have different minimum iOS requirements, we only keep the highest required version because bumping to that version also satisfies all lower ones.
We also keep the current app minimum from that same mismatch so the message can say “increase from X to at least Y.”

packages/flutter_tools/lib/src/ios/mac.dart
hellohuanlin: ah ok. this sounds quite ugly and worth a cleanup. Not the scope of your PR so you don't have to fix it. Could you add a comment to where `issueDetected` is defined, so it's helpful for future readers?

MohammedTarigg: done, i added a comment at issueDetected definition to clarify that it tracks whether we already surfaced a targeted message and should skip the less-accurate stdout fallback diagnostics.

packages/flutter_tools/lib/src/ios/mac.dart
hellohuanlin: Please add a comment there for future readers.

MohammedTarigg: added a comment to say we keep only the highest required version

vashworth (approved):
@MohammedTarigg Thanks for your work on this!

PR #181722, opened 2026-01-30
Remove Material Dependency from `semantics_debugger_test`

rizwan-saleem:
This change refactors `semantics_debugger_test.dart` to eliminate its dependency on **Material** widgets, as part of the test reorganization effort to ensure tests only depend on the library code they are testing.

---

## Changes

- Replace `Material` import with `widgets` import in `semantics_debugger_test.dart`
- Replace `ElevatedButton` with `TestButton` (2 tests)
- Replace `Slider` with `TestSlider` (1 test)
- Replace `Checkbox` with `TestCheckbox` (2 tests)
- Replace `TextField` with `TestTextField` (1 test)
- Replace `MaterialApp` with `TestWidgetsApp` (2 tests)
- Remove `Material` wrapper widgets throughout tests
- Replace `Colors` constants with explicit `Color` values

---

### `TestSlider`
A minimal slider with semantics support for testing.

- Implements `onIncrease` / `onDecrease` semantic actions
- Supports min/max range and divisions
- Uses a **10% adjustment unit** for continuous sliders

### `TestCheckbox`
A minimal checkbox with semantics support for testing.

- Implements checked state and `onTap` semantic action
- Supports enabled and disabled states

---

### `TestTextField`
Updated to better match the Material `TextField` semantics structure.

- Added a `Semantics` wrapper
- Provides enabled state and `onTap` action
- Allows tests to find `Semantics` descendants as expected

---

## Test Results

All **21 tests pass** successfully with the new test widget
Part of  #177414

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[...]

PR #182395, opened 2026-02-13
Standardize on Test* widgets in *_tester.dart files

justinmc:
This PR moves:

 * TestTextField from editable_text_utils.dart to editable_text_tester.dart
 * TestListTile from list_tile_test_utils.dart to list_tile_tester.dart
 * TestButton from utils.dart to button_tester.dart
 
The purpose is to align with other Test* widgets, such as TestWidgetsApp in [widgets_app_tester.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/widgets_app_tester.dart) and TestSemantics in [semantics_tester.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/test/widgets/semantics_tester.dart).

We should continue to follow this pattern in the future, such as in https://github.com/flutter/flutter/pull/181722.


packages/flutter/test/widgets/editable_text_tester.dart
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+
+class _TestTextFieldSelectionGestureDetectorBuilder extends TextSelectionGestureDetectorBuilder {
+  _TestTextFieldSelectionGestureDetectorBuilder({required _TestTextFieldState state})
+    : super(delegate: state);
+
+  @override
+  void onUserTap() {
+    // BasicTestTextField does not have an onTap callback.
```
navaronbracke: "TestTextField" ?

justinmc: Might as well clean that up while I'm at it, good catch.

packages/flutter/test/widgets/editable_text_tester.dart
```diff
@@ -0,0 +1,165 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return _selectionGestureDetectorBuilder.buildGestureDetector(
+      behavior: HitTestBehavior.translucent,
+      child: EditableText(
+        key: editableTextKey,
+        autofillHints: widget.autofillHints,
+        autofocus: widget.autofocus,
+        backgroundCursorColor: _red, // Colors.red, required by editable text.
```
navaronbracke, line 106: Just for correctness, this isn't "Colors.red" from Material, that is `0xFFF44336`, not `0xFFFF0000`

packages/flutter/test/widgets/radio_group_test.dart
```diff
@@ -7,7 +7,7 @@ import 'package:flutter/semantics.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import 'editable_text_utils.dart' show TestTextField;
+import 'editable_text_tester.dart' show TestTextField;
```
navaronbracke: Was there a specific reason for the show combinator here and in other places? (not that it matters much)

navaronbracke (approved):
LGTM, with one suggestion

packages/flutter/test/widgets/radio_group_test.dart
justinmc: I guess it made more sense when it was imported from a more general utils file. I'll push a commit to remove these.

victorsanni (approved):

Renzo-Olivares (approved):
LGTM

PR #182398, opened 2026-02-13
Add await to tester.pump callsites

victorsanni:
Surprised we missed this when reviewing the 0x0 PRs. cc @ahmedsameha1


dkwingsmt (approved):
I can't believe these passed the CI checks. How? Are they ancient lines before we were equipped with analysis?

PR #182402, opened 2026-02-13
[Impeller] libImpeller: Dispose thread local caches on each Vulkan frame

jason-simmons:
The Impeller Vulkan back end caches command pools in thread-local storage. These caches can grow unbounded unless the embedder calls ContextVK::DisposeThreadLocalCachedResources()

This PR ensures that libImpeller Vulkan apps do this on each frame by calling DisposeThreadLocalCachedResources inside the ImpellerVulkanSwapchainAcquireNextSurfaceNew API.

### DIFF
--- a/engine/src/flutter/impeller/toolkit/interop/backend/vulkan/swapchain_vk.cc
+++ b/engine/src/flutter/impeller/toolkit/interop/backend/vulkan/swapchain_vk.cc
@@ -47,6 +47,9 @@ ScopedObject<SurfaceVK> SwapchainVK::AcquireNextSurface() {
     return nullptr;
   }
 
+  auto& context_vk = impeller::ContextVK::Cast(*context_->GetContext());
+  context_vk.DisposeThreadLocalCachedResources();
+
   auto impeller_surface = swapchain_->AcquireNextDrawable();
   if (!impeller_surface) {
     VALIDATION_LOG << "Could not acquire next drawable.";


gaaclarke (commented):
Were these caches created as a means to improve performance?  What sort of performance ramifications will dumping these each frame have?  Is the growth bounded by the largest usage of a frame, as opposed to being additive?  If so maybe a better heuristic would be to dispose of the caches if the cache size is much greater than the actual usage.

gaaclarke (approved):
> In the SurfaceContextVK used by the engine, MarkFrameEnd/ContextVK::DisposeThreadLocalCachedResources is called inside AcquireNextSurface.
> 
> The performance cost is generally insignificant.

Sounds good, this does look like what we are doing in the engine proper.  Even at 120hz a lock probably isn't a big deal even when contested but if we can easily place it somewhere where we aren't holding onto a system resource it seems like an easy problem to avoid.

We have an integration that already covers this code, right?  I'm not too familiar with this setup.

gaaclarke:
@jason-simmons friendly ping, this was approved before the holiday weekend.  But I did have one question about test coverage.

jason-simmons:
The libImpeller library has some tests that are built into the `impeller_unittests` binary.  It includes a suite of playground tests that run on the various Impeller back ends including Vulkan (https://github.com/flutter/flutter/blob/master/engine/src/flutter/impeller/toolkit/interop/impeller_unittests.cc)

However, the playground test framework is using the main Impeller Vulkan `SwapchainVK` implementation, not the libImpeller `SwapchainVK` wrapper affected by this PR.

PR #182411, opened 2026-02-14
Improve (eliminate?) touch latency on iOS

bernaferrari:
Hopefully fix or make it less bad: https://github.com/flutter/flutter/issues/110431

I might need help from second pair of eyes to be sure the issue is either fixed or improved (and that this is the best fix).

I ran dozens of experiments and they all tell me this is an improvement. But the original code from @knopp was a bit confusing to me. 

So I hope he or someone else can review this. I really hope I'm not breaking anything. If you find any issues, let me know. If this PR is bad, let's pretend it never existed.

### DIFF
--- a/engine/src/flutter/shell/platform/darwin/ios/framework/Source/VsyncWaiterIosTest.mm
+++ b/engine/src/flutter/shell/platform/darwin/ios/framework/Source/VsyncWaiterIosTest.mm
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 #import <OCMock/OCMock.h>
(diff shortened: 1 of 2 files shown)


bernaferrari:
Not yet, this universe is still a bit new to me. But I ran some experiments trying 1ms vs min(10% of interval, 2ms) and the min provided almost the same result with, well, less overhead. So I know that it provides a good trade-off. But I don't know yet how much worse it is in performance compared to original. I don't expect much. But I can do that later (as soon as I figure out a reliable way to compare them. I've never had a change on the engine that deals with iOS simulator).

PR #182415, opened 2026-02-14
[Material] modernize legacy switch statements to expressions in app.dart, bottom_navigation_bar.dart, and button.dart

Enderjua:
Following up on the previous merge for Dart 3 switch modernization, this PR continues refactoring legacy switch statements in the Material library. The focus remains on improving type safety through enum exhaustiveness as requested in earlier reviews.

Modernized **10 legacy switches** across `app.dart`, `bottom_navigation_bar.dart`, and `button.dart`

`flutter analyze` passed with no issues and all modified files are processed with `dart format`

## Pre-launch Checklist

- [x] I signed the [CLA]. (I will sign it once the bot prompts me)


packages/flutter/lib/src/material/app.dart
```diff
@@ -857,22 +857,24 @@ class MaterialScrollBehavior extends ScrollBehavior {
...
+            TargetPlatform.android || TargetPlatform.fuchsia || TargetPlatform.iOS => true,
+          },
+    );
+    return switch (axisDirectionToAxis(details.direction)) {
+      Axis.horizontal => child,
+      Axis.vertical => switch (getPlatform(context)) {
+        TargetPlatform.linux ||
+        TargetPlatform.macOS ||
+        TargetPlatform.windows => Scrollbar(controller: details.controller, child: child),
+        TargetPlatform.android || TargetPlatform.fuchsia || TargetPlatform.iOS => child,
```
dkwingsmt: Why can't we make it `details.controller!` here and remove the grand assertion before? The assertion was not very readable and I think we should avoid it if possible.

dkwingsmt (commented):
Generally LGTM except for one comment.

packages/flutter/lib/src/material/app.dart
Enderjua: That makes sense, it definitely cleans up the code. I will replace assert code.

I am thinking...: by using !, we are moving from a debug-only check (assert) to a runtime check. If details.controller ever happens to be null on desktop in a release build, it will now result in a crash rather than failing. Are we comfortable with this strict requirement for Scrollbar on desktop?

Enderjua:
> Hi @Enderjua, can you fix the failing tests?


I don't think these test errors will be corrected without deviating from Tong Mu's suggestion. Can we talk about the question I asked above?

packages/flutter/lib/src/material/app.dart
Piinks: The assertion is important here because desktop platforms require the scroll controller to be provided to the scrollbar so that it can be interactive, it cannot be null. Without the assertion, the resulting error message would be hard to debug because this is the scrollbar being applied under the hood. It is not readily accessible to the user as if it were in the widget tree. The ScrollBehavior provides it by default on desktop, which does not use the primary scroll controller by default. It should never be null, because if the user has not provided a scroll controller, ScrollableState creates and manages a fallback controller, which ends up being passed to ScrollBehavior.buildScrollbar. The property itself is nullable because there are different default behaviors on other platforms.

dkwingsmt: I confirmed with @Piinks, who said that we should keep it as an `assert` here to keep the refactory nature of this PR. In this case, I propose the following change that uses a private function to avoid the repetitiveness. I also included a paragraph of doc by rephrasing @Piinks 's explaination above.
```dart
    ScrollController? getNonNullController(ScrollableDetails details) {
      // `Scrollbar` requires a non-null scroll controller for interactivity on
      // desktop. Asserting here prevents confusing errors that are hard to
      // trace. This value should never be null, as `ScrollableState` manages a
      // fallback. The property itself is nullable because there are different
      // default behaviors on other platforms.
      assert(details.controller != null);
      return details.controller;
    }

    return switch (getPlatform(context)) {
      TargetPlatform.linux || TargetPlatform.macOS || TargetPlatform.windows =>
            Scrollbar(controller: getNonNullController(details), child: child),
      TargetPlatform.android || TargetPlatform.fuchsia || TargetPlatform.iOS => child,
    };

File: packages/flutter/lib/src/material/app.dart [hunk 2]
```diff
@@ -857,22 +857,13 @@ class MaterialScrollBehavior extends ScrollBehavior {
   Widget buildScrollbar(BuildContext context, Widget child, ScrollableDetails details) {
     // When modifying this function, consider modifying the implementation in
     // the base class ScrollBehavior as well.
-    switch (axisDirectionToAxis(details.direction)) {
```
dkwingsmt, line 860: You missed this switch. The logic is changed.

Enderjua:
@dkwingsmt @Piinks Sorry for mention...

I am seeing 7 CI failures (e.g., long_press_test, lifecycle_test, system_navigator_test). I have verified locally that these are completely unrelated to my changes in app.dart, button.dart, or bottom_navigation_bar.dart, and they actually pass on my local machine.

Could you please take a look?

AUTHORS
```diff
@@ -137,5 +137,9 @@ Mairon Slusarz <maironlucaslusarz@gmail.com>
 Ricardo Dalarme <ricardodalarme@outlook.com>
 yiiim <ybz975218925@live.com>
 letrungdo <letrdo@gmail.com>
+<<<<<<< HEAD
```
Piinks: Looks like this needs to be cleaned up after a merge?

packages/flutter/lib/src/material/app.dart
```diff
@@ -857,22 +857,25 @@ class MaterialScrollBehavior extends ScrollBehavior {
...
-          case TargetPlatform.windows:
-            assert(details.controller != null);
-            return Scrollbar(controller: details.controller, child: child);
-          case TargetPlatform.android:
-          case TargetPlatform.fuchsia:
-          case TargetPlatform.iOS:
-            return child;
-        }
+
+    // ÖNCE YARDIMCI FONKSİYONU TANIMLA
```
Piinks, line 861: What is this comment?

packages/flutter/lib/src/material/app.dart
```diff
@@ -882,31 +885,26 @@ class MaterialScrollBehavior extends ScrollBehavior {
...
-              child: child,
-            );
-          case AndroidOverscrollIndicator.glow:
-            break;
-        }
-      case TargetPlatform.fuchsia:
-        break;
-    }
-    return GlowingOverscrollIndicator(
+    final Widget glowIndicator = GlowingOverscrollIndicator(
```
Piinks, line 888: This appears to be created regardless of whether or not it is needed?

Piinks: This might be a case where the switch statement is better than the switch expression.

packages/flutter/lib/src/material/app.dart
Enderjua: I forgot to translate to English the comment.

I will translate to English

PR #182419, opened 2026-02-14
Make onChanged optional and add `enabled` to DropdownButtonFormField

### LINKED ISSUE #57953: Why is DropdownButtonFormField's onChanged required?
## Steps to Reproduce

Try to initialize a `DropdownButtonFormField` without specifying the `onChanged` parameter.

**Expected results:** 

For `onChanged` to not be required and the widget to still work, which is consistent with other FormField widgets, such as `TextFormField`.

**Actual results:** 

Currently the `onChanged` parameter is required and is seemingly used as a way to disable the widget if you set it to null. If you don't care about the onChanged event and just want to use the DropdownButtonFormField in a normal basic form, you still have to set onChanged to an empty function (e.g. `onChanged: (value) {}`) otherwise it won't update the FormState/stay enabled. This feels unintuitive to me since e.g. the `TextFormField` widget does not require you to specify `onChanged` unless you're interested in that event.

While I'm not that experienced with Flutter, it seems like it would make more sense to not use `onChanged` as a way of disabling a `DropdownButtonFormField` and instead let it purely be a way of letting the user do something when the field changes. I don't know if there are issues with that or if this odd current behavior is just a leftover from old logic? Seems easy enough to change it up (I could attempt to submit a PR for it).

**Proposed Changes:**
Add a `bool enabled` parameter to `DropdownButtonFormField` (consistent with TextFormField) and use that to decide whether to set the underlying `DropdownButton.onChanged` to control its enabled'ness.

In the future, it might make sense to add an `enabled` parameter to `DropdownButton` as well, as an explicit way of controlling whether it should be enabled or not.

Mairramer:
Made `onChanged` optional and introduced the `enabled` parameter in `DropdownButtonFormField`, allowing proper disabled/read-only usage without dummy callbacks.

**Breaking change**

Previously, setting `onChanged: null` disabled the widget. This is no longer the case. The disabled state must now be controlled via the `enabled` property.

Most simple usages are handled by the automated fix tool. Code that conditionally relied on `onChanged` being null to disable the widget must be updated to use `enabled`.

If not migrated, the widget will remain interactive but will not trigger a callback.

Closes https://github.com/flutter/flutter/issues/57953

## Pre-launch Checklist

- [x] I signed the [CLA].


Piinks:
Hi @Mairramer! Is this something you would like to move forward with?

Mairramer:
> Hi @Mairramer! Is this something you would like to move forward with?

Yes, I’d like that, but only if the team genuinely believes it’s something worth pursuing and sees potential for me to move forward with it.

justinmc (commented):
I'm on board with this approach. I agree that the situation described in https://github.com/flutter/flutter/issues/57953 is probably confusing a bunch of people.

My main question is whether this will cause breakages or not. Looks like customer tests are passing, but we should run the Google tests. If there are tons of breakages maybe we can rethink how we do this, but if not then I'm all for landing this.

Mairramer:
> I'm on board with this approach. I agree that the situation described in #57953 is probably confusing a bunch of people.
> 
> My main question is whether this will cause breakages or not. Looks like customer tests are passing, but we should run the Google tests. If there are tons of breakages maybe we can rethink how we do this, but if not then I'm all for landing this.

In this case, we would only have a small breaking change related to the conditional use of `onChanged`, such as `condition ? (value) { ... } : null` (or vice versa), because from what I’ve seen, there’s no straightforward way to migrate this automatically.

packages/flutter/lib/fix_data/fix_material/fix_dropdown_button.yaml
justinmc, line 1: Thanks for including a dart fix

justinmc (dismissed):
Approving in order to kick off the Google tests.

justinmc:
This will have to be postponed due to the code freeze, though.

dkwingsmt:
There are Google test failures, and they look real.

justinmc:
@Mairramer I see Google tests that have failed because a previously disabled field is now enabled. Would it be possible to solve this problem in a way that doesn't break anything?

Mairramer:
> @Mairramer I see Google tests that have failed because a previously disabled field is now enabled. Would it be possible to solve this problem in a way that doesn't break anything?

Yes, I believe so, but I’ll need to adjust a few things in the implementation first to avoid introducing breaking changes and to keep the existing behavior intact.

dkwingsmt:
Is it ready for another review? Or if the change is going to take some time, would you mind converting this PR to a draft for now? (from triage)

Mairramer:
> Is it ready for another review? Or if the change is going to take some time, would you mind converting this PR to a draft for now? (from triage)

Yes, it’s ready. The main thing now would be for someone with approval access to verify the Google testing results.

Piinks:
I will approve to kick of Google testing to see if that is resolved, in the meantime, since this will be a breaking change, can you create a migration guide in flutter/website? We can land it when this is ready, but it will be good to have a look at the guidance and iterate there as well in the meantime. :)

Piinks (dismissed):

Mairramer:
@Piinks Migration guide created: flutter/website#13430

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1120,19 +1124,23 @@ class DropdownButton<T> extends StatefulWidget {
...
-  /// is null then the dropdown button will be disabled, i.e. its arrow will be
-  /// displayed in grey and it will not respond to input. A disabled button
-  /// will display the [DropdownButton.disabledHint] widget if it is non-null.
-  /// If [DropdownButton.disabledHint] is also null but [DropdownButton.hint] is
-  /// non-null, [DropdownButton.hint] will instead be displayed.
+  /// The dropdown button is enabled when [enabled] is true.
+  ///
+  /// If [enabled] is false, or the list of [DropdownButton.items] is null or empty,
+  /// then the dropdown button will be disabled. When disabled, the arrow is
+  /// displayed in grey and the button does not respond to input.
```
justinmc, line 1141: Should you also explain what happens if `enabled` is null? Which I think includes the original behavior of disabling when onChanged is null. Or maybe you are trying to encourage people to explicitly set `enabled`?

packages/flutter/lib/src/material/dropdown.dart
```diff
@@ -1340,6 +1348,16 @@ class DropdownButton<T> extends StatefulWidget {
   /// If this property is null, [WidgetStateMouseCursor.adaptiveClickable] will be used.
   final MouseCursor? dropdownMenuItemMouseCursor;
 
+  /// Whether the [DropdownButton] is enabled.
+  ///
+  /// When [enabled] is false, the field is disabled and does not allow user
+  /// interaction or value changes, regardless of the [onChanged] callback.
+  ///
```
justinmc, line 1358: As above, consider explaining what happens when `enabled` is null.

packages/flutter/test/material/dropdown_test.dart
```diff
@@ -4991,4 +5015,50 @@ void main() {
...
+      );
+    }
+
+    // 1. If enabled is false, it is disabled even with onChanged provided.
+    await tester.pumpWidget(buildDropdown(enabled: false, onChanged: (int? v) {}));
+    await tester.tap(find.text('one'));
+    await tester.pumpAndSettle();
+    expect(find.text('two'), findsNothing); // Menu did not open
+
+    // 2. If enabled is true, it is enabled even without onChanged.
```
justinmc: Each pumpWidget is independent from the rest of the test, right? Consider creating a separate testWidgets for each.

justinmc (dismissed):
LGTM with nits 👍 . But due to the code freeze, we should not land this PR. It will have to be ported to flutter/packages after https://github.com/flutter/packages/pull/11669 lands and we open up material_ui.

packages/flutter/lib/src/material/dropdown.dart
Mairramer: In this case, it would be both. However, I’ll include the nullable `enabled` behavior as well.

dkwingsmt (commented):
Trigger google test

dkwingsmt (approved):
Trigger Google test

dkwingsmt:
Thanks for your contribution! This PR has reached completion, however we're not able to merge this PR for the moment due to the Material/Cupertino code freeze (https://github.com/flutter/flutter/issues/184093). As soon as the decoupling is complete, we will post a way to port this PR to the new packages to land. In the meantime, I'll convert this PR to a draft and apply a label so that we can quickly find it when the freeze is lifted.

Piinks:

Mairramer:
Ported to https://github.com/flutter/packages/pull/12610.

PR #182426, opened 2026-02-14
Add dedicated rubber band spring for overscroll recovery in BouncingScrollPhysics

### LINKED ISSUE #181752: Proposal - Missing dedicated "Rubber Band" physics in BouncingScrollPhysics causes low iOS fidelity
Currently, `BouncingScrollPhysics` in Flutter uses a single `SpringDescription` for both momentum scrolling (flings) and the overscroll recovery (snap-back/rubber band effect).

This design forces a compromise:
- If we tune the spring for a natural scroll feel, the snap-back becomes too loose and "floaty" compared to native iOS.
- If we tune it for a snappy return, the scrolling becomes too stiff.

As a result, Flutter's iOS scroll feel lacks the characteristic "heavy and snappy" rubber band effect found in `UIScrollView`, creating an "uncanny valley" effect for users.

## Video
https://github.com/user-attachments/assets/435744b3-0caa-4d91-9f85-7a27c8c50c54

## Describe the solution

Native iOS behaves as if it uses a different physical property when snapping back from an overscroll. Flutter should implement a dedicated spring simulation for this specific phase.

I have analyzed the motion curves of native iOS and found that separating the spring logic significantly improves fidelity.

**Proposed Logic:**
1. Keep the existing spring for standard scrolling to maintain smoothness.
2. Apply a specific **"Rubber Band Spring"** when the scroll position is out of bounds and the gesture ends.
3. Empirically tuned values for this spring: `mass: 1.0`, `stiffness: 1e5 * lambda^2`, `damping: 100001 * lambda`.

> [!NOTE]
> The constants were specially tuned to achieve exponential damping for the spring.

## Additional context

I have measured and compared the scroll position over time between:
1. Native iOS (UIScrollView)
2. Current Flutter
3. **New Flutter implementation**

Image

As shown in the graph, the current Flutter implementation (Cyan) exhibits a **significantly slower recovery** compared to native iOS. It lacks the initial 'snap' (stiffness) and takes considerably longer to settle (~0.81s) compared to the native duration (~0.63s), leading to the perceived "floaty" behavior.

In contrast, the new implementation (Brown) closely aligns with the native curve (Blue) in trajectory.

I have a PR ready with this fix.

MTtankkeo:
Fixes #181752
*(for detailed analysis, measurements, and discussion, please refer to the original issue.)*

This pull request implements a dedicated “rubber band” physics for overscroll snap-back in BouncingScrollPhysics on iOS to better match the native UIScrollView feel.

546867255-4ad56291-9603-40c1-ae5a-3e319017c3c9

- - -

> [!NOTE]
> This change adjusts spring constants and simulation behavior
> to improve iOS scroll fidelity.
>
> Given that this change primarily involves physics parameter tuning,
> it may reasonably qualify as test-exempt. However, I am not entirely
> certain what level of automated testing would be considered appropriate.
>
> Please let me know if additional tests are expected,
> and I am happy to update the PR accordingly.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/widgets/scroll_physics.dart
```diff
@@ -753,9 +778,12 @@ class BouncingScrollPhysics extends ScrollPhysics {
   @override
   Simulation? createBallisticSimulation(ScrollMetrics position, double velocity) {
     final Tolerance tolerance = toleranceFor(position);
-    if (velocity.abs() >= tolerance.velocity || position.outOfRange) {
+    final bool isStationary = velocity.abs() <= tolerance.velocity;
```
navaronbracke, line 797: This new behavior is going to need a test that verifies that the right spring is used in the right circumstances.

dkwingsmt: I'm surprised that releasing from stationary uses a different spring from otherwise. I'd imagine that the spring might depend on whether it is an overscroll, but not the starting velocity. Can you check and confirm?

dkwingsmt (commented):
Fantastic change overall! Thank you so much for the investigation, and especially for the figure for demonstration.

I have commented a question.

Regarding tests, do you think the following test is possible?
* Build a scrollable with this physics, overscroll it, release. For every 0.2s elapsed, check the y position of the content.

packages/flutter/lib/src/widgets/scroll_physics.dart
```diff
@@ -685,6 +685,31 @@ class BouncingScrollPhysics extends ScrollPhysics {
...
+  /// - Fast mode: r2 = -1e5 * lambda (vanishes almost instantly)
+  ///
+  /// Mathematical derivation:
+  /// - r1 + r2 = -damping/mass, so damping = (1e5 + 1) * lambda
+  /// - r1 * r2 = stiffness/mass, so stiffness = 1e5 * lambda^2
+  static final SpringDescription rubberBandSpring = SpringDescription(
+    mass: 1.0,
+    stiffness: 1e5 * _rubberBandLambda * _rubberBandLambda,
+    damping: (1e5 + 1) * _rubberBandLambda,
+  );
```
dkwingsmt: It took me a while to understand this part. I consulted AI to reword documentation. Here's my proposal:
```suggestion
  // Approximation of iOS native rubber band decay rate.
  static const double _rubberBandHalfLifeSeconds = 0.07;

  // Decay constant (lambda) for rubber band spring simulation.
  static final double _rubberBandLambda = math.log(2) / _rubberBandHalfLifeSeconds;

  /// Spring used to animate overscroll bounce from a stationary release in iOS
  /// native style.
  ///
  /// Used in [createBallisticSimulation] depending on the conditions.
  ///
  /// Research indicates that iOS employs a distinct decay function when a
  /// scrollable area is released in an overscroll and stationary state (zero
  /// initial velocity), conforming to an exponential decay model.
  //
  // ## Mathematical derivation
  //
  // A standard spring-damper system follows the second-order differential equation:
  // m*x'' + c*x' + k*x = 0
  //
  // To force this second-order system to behave like a first-order exponential
  // decay x(t) = C * e^(-lambda * t), we configure it as an overdamped spring
  // with two explicitly defined roots (r1 and r2) for its characteristic
  // equation:
  //
  // * r1 = -lambda (the primary root driving the visible exponential decay)
  // * r2 = -100000 * lambda (an extremely large negative root)
  //
  // Because r2 is massive and negative, its corresponding term in the exact
  // mathematical solution (C2 * e^(r2 * t)) decays to zero almost
  // instantaneously. The system movement becomes dominated by r1.
  //
  // Using Vieta's formulas for the characteristic equation r^2 + (c/m)r + (k/m) = 0:
  // * r1 + r2 = -c/m => damping (c) = -(r1 + r2) * m
  // * r1 * r2 = k/m  => stiffness (k) = (r1 * r2) * m
  static final SpringDescription rubberBandSpring = SpringDescription(
    mass: 1.0,
    stiffness: 1e5 * _rubberBandLambda * _rubberBandLambda,
    damping: (1e5 + 1) * _rubberBandLambda,
  );
```

packages/flutter/test/widgets/scroll_physics_test.dart
```diff
@@ -366,4 +366,33 @@ FlutterError
...
+
+    final double xStationary = simStationary.x(0.2);
+    final double xMoving = simMoving.x(0.2);
+
+    expect(simStationary, isA<BouncingScrollSimulation>());
+    expect(simMoving, isA<BouncingScrollSimulation>());
+
+    // Both simulations should still be in motion.
+    expect(xStationary.abs(), greaterThan(0));
+    expect(xMoving.abs(), greaterThan(0));
```
dkwingsmt: Instead of `greaterThan(0)`, I think we should just compare with it with a literal value (with proper precision) to ensure that the curve is as expected.

It would be best if it matches the environment of your device test so that you can plug in the real data you used to draw that chart. :) But it's ok if it doesn't.

dkwingsmt (approved):
LGTM. Thank you for the great work!

Piinks (commented):
Hey @MTtankkeo thanks for contributing this! It looks like it is causing some failures, can you take a look?

PR #182477, opened 2026-02-16
Add labelTextDirection handling in InputDecorator

### LINKED ISSUE #41324: TextField/TextFormField labelText and hintText should be right-aligned with TextDirection.rtl
There is forgotten attribute **write** for the `labelText of(TextFormField or TextField)` from right to left.
I tried with `textAlign`, `textDirection`. Although, it works with `hintText` and the user input, it doesn't work with the `labelText`.

Screenshot
[textfield.txt](https://github.com/flutter/flutter/files/3655101/textfield.txt)

```dart
TextFormField(
        textAlign: Language.lang  == 'NotArabic'  ?TextAlign.left:TextAlign.right,
        validator: (value) {
          return Validators().validateEmail(value);
        },
        controller: _emailController,
        decoration: new InputDecoration(
            hintStyle: new TextStyle(color: Colors.grey[500],),
            labelStyle: new TextStyle(color: Colors.grey[900],),
            hintText: 'hint Email',
            labelText: 'label Email',
            fillColor: Colors.white70),
        textDirection:
            Language.lang == 'NotArabic' ? TextDirection.ltr : TextDirection.rtl,
      ) 
```

Mairramer:
This PR introduces `labelTextDirection` property to InputDecoration.

Fixes https://github.com/flutter/flutter/issues/41324

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -2671,7 +2680,7 @@ class _InputDecoratorState extends State<InputDecorator> with TickerProviderStat
         counter: counter,
         container: container,
       ),
-      textDirection: textDirection,
+      textDirection: decoration.textDirection ?? textDirection,
```
Renzo-Olivares: I think we should be moving this logic to when we first set `textDirection` on line 2574 above. It seems the text direction is taken into account in the padding calculation `final EdgeInsets? resolvedPadding = decoration.contentPadding?.resolve(textDirection);` so it should have the appropriate text direction at that time.

Renzo-Olivares (commented):
Hi @Mairramer, thank you for the contribution! I definitely agree we should try to make textDirection consistent across text field's components like input decorator. I left a small comment but the change and approach looks good to me overall.

packages/flutter/lib/src/material/text_field.dart
```diff
@@ -1688,7 +1692,7 @@ class _TextFieldState extends State<TextField>
           style: style,
           strutStyle: widget.strutStyle,
           textAlign: widget.textAlign,
-          textDirection: widget.textDirection,
+          textDirection: widget.decoration?.textDirection ?? widget.textDirection,
```
Renzo-Olivares: This feels like it might be a breaking change if someone had has TextField.textDirection and TextField.hintTextDirection as different and then they migrate away from hintTextDirection. Previously the text field and input decorator could have a different text direction but after this change they cannot. I don't think this is a blocker since I think unifying the handling here is the right decision but definitely something to keep an eye out for.

Renzo-Olivares (dismissed):
LGTM, thank you for the contribution @Mairramer!

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -3167,6 +3178,16 @@ class InputDecoration {
   /// input field and the current context.
   final TextDirection? hintTextDirection;
 
+  /// The direction to use for both the [labelText] and the [hintText].
+  ///
+  /// This parameter ensures visual consistency across all text elements
+  /// within the input field. If provided, it overrides individual direction
```
Renzo-Olivares: I think this needs to be updated, "Individual direction settings like [hintTextDirection] take precedence over this global configuration."

Renzo-Olivares: It would also be helpful to add some docs to `TextField.textDirection` to make it clear that it setting it does not influence the `textDirection` of the input decoration, to do that a user must set either `InputDecoration.textDirection` and/or `InputDecoration.hintTextDirection`.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -2320,6 +2320,10 @@ class _InputDecoratorState extends State<InputDecorator> with TickerProviderStat
         ? _InputDecoratorDefaultsM3(context)
         : _InputDecoratorDefaultsM2(context);
     final IconButtonThemeData iconButtonTheme = IconButtonTheme.of(context);
+    final TextDirection directionality = Directionality.of(context);
+    final TextDirection effectiveLabelTextDirection =
+        decoration.labelTextDirection ?? directionality;
+    final TextDirection effectiveHintTextDirection = decoration.hintTextDirection ?? directionality;
```
loic-sharma: I think we don't need to determine the effective text direction here. If we just pass the text directions to `Text`, `Text` will determine the effective text direction by looking up the ambient text direction if needed.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -3167,6 +3178,12 @@ class InputDecoration {
   /// input field and the current context.
   final TextDirection? hintTextDirection;
 
+  /// The direction to use for the [labelText].
+  ///
+  /// If null, defaults to a value derived from [Directionality] for the
+  /// input field and the current context.
+  final TextDirection? labelTextDirection;
```
loic-sharma: Could you move this field to be after the `floatingLabelStyle` field? That way the label fields are grouped together.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -2848,6 +2857,7 @@ class InputDecoration {
     this.alignLabelWithHint,
     this.constraints,
     this.visualDensity,
+    this.labelTextDirection,
```
loic-sharma, line 2858: Could you move this argument to be after the `floatingLabelStyle` argument so that the label arguments are grouped together? Same comment for the other constructor below.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -3167,6 +3178,12 @@ class InputDecoration {
   /// input field and the current context.
   final TextDirection? hintTextDirection;
 
+  /// The direction to use for the [labelText].
+  ///
+  /// If null, defaults to a value derived from [Directionality] for the
+  /// input field and the current context.
```
loic-sharma: Let's clarify that this value is only used if `labelText` is used, but not if `label` is used.

```suggestion
  /// input field and the current context.
  ///
  /// This is ignored if [label] is provided instead of [labelText].
  /// If you need to set the direction for a custom [label] widget,
  /// wrap that widget in a [Directionality] widget or use a [Text]
  /// widget with its own [Text.textDirection] property.
```

Could you also update `hintTextDirection` accordingly?

packages/flutter/test/material/input_decorator_test.dart
```diff
@@ -15770,4 +15770,36 @@ void main() {
...
+
+    await tester.pumpWidget(
+      const MaterialApp(
+        home: Scaffold(
+          body: Center(
+            child: InputDecorator(
+              decoration: InputDecoration(
+                hintText: hintText,
+                labelText: labelText,
+                labelTextDirection: TextDirection.rtl,
```
Renzo-Olivares: I think this test should either explicitly wrap with a `Directionality` widget set to `ltr` or leave a comment regarding it being the default. Then `labelTextDirection` and `hintTextDirection` should both be set to rtl to make it clear between the default ltr direction set by `Directionality` and the explicit rtl direction set by the `InputDecoration`.

Renzo-Olivares (commented):
Mostly LGTM just had a small comment regarding the test. Thank you for the changes @Mairramer!

loic-sharma (approved):
Thanks for the excellent contribution!

Renzo-Olivares (approved):
LGTM

mdebbar (approved):
FYI: this PR looks good to me. Thanks for the contribution!

Piinks:

Mairramer:
Ported to https://github.com/flutter/packages/pull/12607.

Issue #182498, opened 2026-02-17
A test that creates StretchingOverscrollIndicator with a null child fails with 'Null check operator used on a null value'


ahmedsameha1:
### Steps to reproduce

Put the following test at the end of the main function in: packages/flutter/test/widgets/overscroll_indicator_test.dart. Then run flutter test --verbose packages/flutter/test/widgets/overscroll_indicator_test.dart
```dart
testWidgets('StretchingOverscrollIndicator does not crash when child is null', (
    WidgetTester tester,
  ) async {
    await tester.pumpWidget(
      const Directionality(
        textDirection: TextDirection.ltr,
        child: StretchingOverscrollIndicator(axisDirection: AxisDirection.down),
      ),
    );
  });
```

### Expected results

I think that the test should pass.

### Actual results

The test fails with: 'Null check operator used on a null value'

### Code sample

<details open><summary>Code sample</summary>

```dart
testWidgets('StretchingOverscrollIndicator does not crash when child is null', (
    WidgetTester tester,
  ) async {
    await tester.pumpWidget(
      const Directionality(
        textDirection: TextDirection.ltr,
        child: StretchingOverscrollIndicator(axisDirection: AxisDirection.down),
      ),
    );
  });
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
Building flutter tool...
Resolving dependencies... (2.1s)
Downloading packages... 
Got dependencies.
[  +41 ms] executing: uname -m
[  +10 ms] Exit code 0 from: uname -m
[   +1 ms] x86_64
[  +20 ms] Artifact Instance of 'MaterialFonts' is not required, skipping update.
[        ] Artifact Instance of 'GradleWrapper' is not required, skipping update.
[        ] Artifact Instance of 'AndroidGenSnapshotArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'AndroidInternalBuildArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'IOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[        ] Artifact Instance of 'LegacyCanvasKitRemover' is not required, skipping update.
[        ] Artifact Instance of 'FlutterSdk' is not required, skipping update.
[...]

PR #182499, opened 2026-02-17
Make sure that a StretchingOverscrollIndicator doesn't crash when it …

ahmedsameha1:
This is my attempt to handle #182498.

### DIFF
--- a/packages/flutter/lib/src/widgets/overscroll_indicator.dart
+++ b/packages/flutter/lib/src/widgets/overscroll_indicator.dart
@@ -816,7 +816,7 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
           final Widget transform = StretchEffect(
             stretchStrength: overscroll,
             axis: widget.axis,
-            child: widget.child!,
+            child: widget.child ?? const SizedBox.shrink(),
           );
 
           // Only clip if the viewport dimension is smaller than that of the
--- a/packages/flutter/test/widgets/overscroll_indicator_test.dart
+++ b/packages/flutter/test/widgets/overscroll_indicator_test.dart
@@ -731,6 +731,18 @@ void main() {
       );
     });
   });
+
+  testWidgets('StretchingOverscrollIndicator does not crash when child is null', (
+    WidgetTester tester,
+  ) async {
+    await tester.pumpWidget(
+      const Directionality(
+        textDirection: TextDirection.ltr,
+        child: StretchingOverscrollIndicator(axisDirection: AxisDirection.down),
+      ),
+    );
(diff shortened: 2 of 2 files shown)


packages/flutter/lib/src/widgets/overscroll_indicator.dart
```diff
@@ -816,7 +816,7 @@ class _StretchingOverscrollIndicatorState extends State<StretchingOverscrollIndi
           final Widget transform = StretchEffect(
             stretchStrength: overscroll,
             axis: widget.axis,
-            child: widget.child!,
+            child: widget.child != null ? widget.child! : const SizedBox.shrink(),
```
victorsanni:
```suggestion
            child: widget.child ?? const SizedBox.shrink(),
```

packages/flutter/test/widgets/overscroll_indicator_test.dart
```diff
@@ -731,6 +731,17 @@ void main() {
...
+  testWidgets('StretchingOverscrollIndicator does not crash when child is null', (
+    WidgetTester tester,
+  ) async {
+    await tester.pumpWidget(
+      const Directionality(
+        textDirection: TextDirection.ltr,
+        child: StretchingOverscrollIndicator(axisDirection: AxisDirection.down),
+      ),
+    );
+  });
```
victorsanni, line 745: Do we need to confirm that no exceptions were thrown?

victorsanni (commented):
My first thought here was is there a real life case where `child` is null? But then `StretchingOverscrollIndicator` is a public API so I don't think it's far-fetched for a user to accidentally configure it with a nullable value that is null as at when the widget is instantiated.

packages/flutter/test/widgets/overscroll_indicator_test.dart
ahmedsameha1: For me, I think we do not need a confirmation for that; I think it is redundant, as in https://github.com/flutter/flutter/issues/6537#issuecomment-3157607193, even for readability, as the description of the test states its aim.
Anyway, I saw the following multiple times in the codebase.
```dart
expect(tester.takeException(), isNull);
```
It is up to you.

victorsanni (approved):

Piinks (approved):
LGTM

packages/flutter/test/widgets/overscroll_indicator_test.dart
victorsanni: Even though it might not serve a purpose, I lean towards explicitly adding it so it is obvious to a reader of the test that it serves to prevent against an exception getting thrown.

PR #182516, opened 2026-02-18
Add a cli flag for toggling HCPP use, and enable it

gmackall:
1. Adds the flag (`--enable-surface-control`)
2. Adds a driver test which tests that we can upgrade all 3 of the create calls (HC, TLHC w VD fallback, TLHC w HC fallback)
3. Respects its use in create call

Does not yet rename it, will do that in a follow up as it's logically independent.


packages/flutter_tools/lib/src/commands/run.dart
```diff
@@ -248,6 +248,7 @@ abstract class RunCommandBase extends FlutterCommand with DeviceBasedDevelopment
     addEnableFlutterGpuFlag(verboseHelp: verboseHelp);
     addEnableVulkanValidationFlag(verboseHelp: verboseHelp);
     addEnableEmbedderApiFlag(verboseHelp: verboseHelp);
+    addEnableSurfaceControlFlag(verboseHelp: verboseHelp);
```
jmagman, line 251: Add to `flutter test` as well.
https://github.com/flutter/flutter/blob/f5825a22a47969ce9fec303dbefee035eadb1acc/packages/flutter_tools/lib/src/commands/test.dart#L82

gmackall: done

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterShellArgs.java
```diff
@@ -132,6 +135,14 @@ public static FlutterShellArgs fromIntent(@NonNull Intent intent) {
     if (intent.getBooleanExtra(ARG_KEY_ENABLE_VULKAN_VALIDATION, false)) {
       args.add(ARG_ENABLE_VULKAN_VALIDATION);
     }
+    if (intent.hasExtra(ARG_KEY_TOGGLE_SURFACE_CONTROL)) {
+      if (intent.getBooleanExtra(ARG_KEY_TOGGLE_SURFACE_CONTROL, true)) {
```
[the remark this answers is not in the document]
gmackall: thanks gemini

packages/flutter_tools/lib/src/runner/flutter_command.dart
```diff
@@ -1292,6 +1292,14 @@ abstract class FlutterCommand extends Command<void> {
     );
   }
 
+  void addEnableSurfaceControlFlag({required bool verboseHelp}) {
+    argParser.addFlag(
+      'enable-surface-control',
+      hide: !verboseHelp,
```
johnmccutchan, line 1298: The flag name and its help text are meaningless to 99% of Flutter developers

gmackall: very fair, am planning to rename it in an immediate follow up pr

dev/bots/suite_runners/run_android_engine_tests.dart
```diff
@@ -107,6 +108,12 @@ Future<void> runAndroidEngineTests({required ImpellerBackend impellerBackend}) a
 
     // Test HCPP Platform Views on Vulkan.
     if (impellerBackend == ImpellerBackend.vulkan) {
+      // Run upgrade_legacy_pv_types first, as it is testing the flag and not the manifest
```
reidbaker: nit can you put the full path here and explain why it matters that the test is testing the flag and not the manifest. It took me a bit to understand what upgrade_legacy_pv_types was referring to.

dev/bots/suite_runners/run_android_engine_tests.dart
```diff
@@ -116,8 +123,8 @@ Future<void> runAndroidEngineTests({required ImpellerBackend impellerBackend}) a
       for (final file in mains) {
         // This statement is attempting to catch all tests inside of the
         // dev/integration_tests/android_engine_test/lib/hcpp
-        // directory.
-        if (!file.path.contains('hcpp')) {
+        // directory, except for upgrade_legacy_pv_types which we already ran.
+        if (!file.path.contains('hcpp') || file.path.contains('upgrade_legacy_pv_types')) {
```
reidbaker: I think having the string helps with understandability but what do you think about having a list of `runFirstTests` and for now having it contain one item. I think it would make the callsite code easier to understand then the explanation for why a test needs to be run first can go adjacent to its entry in the list.

dev/integration_tests/android_engine_test/test_driver/hcpp/upgrade_legacy_pv_types_main_test.dart
```diff
@@ -0,0 +1,84 @@
...
+// found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io' as io;
+
+import 'package:android_driver_extensions/native_driver.dart';
+import 'package:flutter_driver/flutter_driver.dart';
+import 'package:test/test.dart';
+
+/// Smoke test: verifies that when HCPP is enabled, all three
```
reidbaker, line 12: great comment

dev/integration_tests/android_engine_test/test_driver/hcpp/upgrade_legacy_pv_types_main_test.dart
```diff
@@ -0,0 +1,84 @@
...
+          'got $hcppCount. Logcat:\n$logcat',
+    );
+    expect(
+      legacyCount,
+      0,
+      reason:
+          'Expected 0 legacy creations, '
+          'got $legacyCount. Logcat:\n$logcat',
+    );
+  }, timeout: Timeout.none);
```
reidbaker, line 83: Is timeout none correct here?

reidbaker (approved):

dev/bots/suite_runners/run_android_engine_tests.dart
gmackall: I would prefer not to add the full path, as there is no enforcement that it is correct. Someone might reasonably move the file up or down subdirectories, and the comment will become incorrect. Notably this is not true about the name in my opinion, because of the proximity of this comment to the check right below

dev/integration_tests/android_engine_test/test_driver/hcpp/upgrade_legacy_pv_types_main_test.dart
gmackall: Yeah, it's also the timeout used for all the other hcpp tests. It's only a per test timeout, so the target will still fail if it reaches the ci timeout

dev/bots/suite_runners/run_android_engine_tests.dart
gmackall: changed

Issue #182400, opened 2026-02-13
Possible GLSL -> SkSL translation error.


zanderso:
Looking at https://github.com/flutter/flutter/pull/180861, I think it's the right thing to do, but it doesn't play nicely with shaders that aren't trying to target Skia, like the ones from https://github.com/whynotmake-it/flutter_liquid_glass/tree/main/packages/liquid_glass_renderer.

impellerc doesn't distinguish between iOS and macOS when compiling runtime effects for Metal when deciding whether SkSL also needs to be bundled in the iplr: https://github.com/flutter/flutter/blob/master/engine/src/flutter/impeller/compiler/types.cc#L318.

In particular, the SkSL doesn't need to be bundled for iOS since there's no Skia there at all.

Before the build time error, an iOS app would bundle the SkSL runtime effect shaders, and they'd just never be used.

After the build time errors, the shaders in liquid_glass_renderer cause an app to fail to compile.

The `TargetPlatformBundlesSkSL` function and the way it's used for runtime effect shaders might just be a bit funky because for runtime effects, you can just specify on the command line what you want included. This is what the Flutter tool does: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/build_system/tools/shader_compiler.dart#L118

In any case, here's an example of the input glsl, the translation to SkSL, and the SkSL compiler error:

```glsl
float sceneSDF(vec2 p, int numShapes, float shapeData[MAX_SHAPES * 6], float blend) {
    if (numShapes == 0) {
        return 1e9;
    }
    
    float result = getShapeSDFFromArray(0, p, shapeData);
    ...
```

```glsl
 float FLT_flutter_local_sceneSDF(vec2 p, int numShapes, float shapeData[96], float blend) {
    if (numShapes == 0)
    {
        return 1000000000.0;
    }
    int param = 0;
    vec2 param_1 = p;
    float param_2[96] = shapeData;
    float result = FLT_flutter_local_getShapeSDFFromArray(param, param_1, param_2);
...
```

```
[        ]            SkSL Error:
[        ]            error: 121: initializers are not permitted on arrays (or structs containing arrays)
[        ]                float param_2[96] = shapeData;
```

cc @gaaclarke @b-luk

zanderso:
Repro with the example in https://github.com/whynotmake-it/flutter_liquid_glass/tree/main/packages/liquid_glass_renderer/example

[...]

PR #182519, opened 2026-02-18
Don't compile shaders to SkSL unless --sksl arg is present

b-luk:
The main change of this PR is to make `impellerc_main` no longer use `TargetPlatformBundlesSkSL(switches.SelectDefaultTargetPlatform())` to determine whether to compile to SkSL.

Addresses the "sksl unexpectedly compiling when building for ios" mentioned in #182400.

`TargetPlatformBundlesSkSL()` returned `true` when running impellerc targeting *any* runtime shaders. So even though https://github.com/flutter/flutter/pull/163144/changes changed iOS runtime shaders to no longer call impellerc with the `--sksl` runtime target, it would still compile to SkSL because of the `--runtime-stage-metal` runtime target.

After this PR, impellerc will only compile to SkSL if `--sksl` is explicitly provided when calling impellerc.

The rest of the changes in this PR are related changes to improve code organization and to update tests.

- `impellerc_main.cc`:
  - `OutputIPLR()`: Remove the custom logic for compiling to SkSL. SkSL is now treated the same as all the other compile targets in the `for (const auto& platform : switches.PlatformsToCompile())` loop. This is the only change with a behavioral difference in the PR. We no longer use `TargetPlatformBundlesSkSL(switches.SelectDefaultTargetPlatform())` to determine whether to compile to SkSL. Instead, it's treated the same as any other compilation target and will only be targeted when it is specified as an arg to the executable.
  - Remove the `CompileSkSL()` helper function. This had the exact same logic as what is used for all the other compilation targets in the `for (const auto& platform : switches.PlatformsToCompile())` loop. So it is not needed.
  - `OutputDepfile()`: Remove the switch/case. The `TargetPlatform::kUnknown` case is unreachable, so this used the other case 100% of the time.
  - `Main()`: `switches.CreateSourceOptions()` no longer has a default platform target parameter. It doesn't matter which target it's called with here. Arbitrarily call it with `switches.PlatformsToCompile().front()`.

- `types.h/cc`:
  - Remove the `TargetPlatformBundlesSkSL()` function. The only place this was used was for `impellerc_main.cc`'s special case logic for SkSL, which was removed as described above.

- `switches.h/cc`:
  - Remove the optional/default parameter for `CreateSourceOptions()`. This used to fall back to calling `SelectDefaultTargetPlatform()` to determine the default value. I found this to be very unclear behavior. The only place this was actually called with the default parameter is in one function in `shader_bundle.cc` where the selected target platform does not actually matter, so it doesn't even need this somewhat-convoluted logic to select a platform with `SelectDefaultTargetPlatform()`. I changed `CreateSourceOptions()` to require an explicit parameter, which I think makes the function's behavior clearer by removing the confusing default parameter.
  - Remove the `SelectDefaultTargetPlatform()` function. I think it was a little unclear/nonobvious what this was actually returning. It was only used in 3 places, all of which were kind of confusing/unneeded:
    1. In this same file, to pick the default target platform in `switches.CreateSourceOptions()`. As described above, this seems entirely unnecessary.
    1. In `impeller_main.cc`, used in `TargetPlatformBundlesSkSL(switches.SelectDefaultTargetPlatform())` to determine whether to compile to SkSL. As described above, we don't want this behavior. I think it was also kind of confusing.
    1. In `impeller_main.cc`, used for the switch/case in `OutputDepfile()`. As described above, this switch/case is entirely unnecessary.
  - Change the `kKnownRuntimeStages` map to be a vector of pairs. This is iterated through to populate the returned `runtime_stages_` list of `Switches::PlatformsToCompile()`. Making it a vector makes the `runtime_stages_` list maintain the ordering of `kKnownRuntimeStages` (previously, iterating through the map would iterate in alphabetical order of the keys).
    - `impellerc_main.cc`'s `OutputIPLR()` now compiles to targets based on `Switches::PlatformsToCompile()`, without special case logic to always compile to "sksl" first. Changing this to a vector with "sksl" as the first value preserves the original behavior of compiling to "sksl" before any other targets. We do this because certain tests that perform a failed shader compilation check specifically for an SkSL-based error message (e.g. [this one](https://github.com/flutter/flutter/blob/64866862f623ceeb45fd8be4782e8db8b58910c0/packages/flutter_tools/test/integration.shard/shader_compiler_test.dart#L150-L170)). So for these tests, we need to try/fail with the SkSL compiler first, before trying/failing with other compilers which would produce a different error message.

- `shader_bundle.cc`
  - As described above, change the usage of `switches.CreateSourceOptions()` to require an explicit target platform parameter. This particular usage doesn't matter, so use `TargetPlatform::kUnknown` and add an explanatory comment.

- `compiler.cc`
  - In `CreateCompiler()`, Add an `FML_UNREACHABLE` for the `TargetPlatform::kUnknown` case, instead of falling back to using a vulkan compiler. It doesn't make sense to call `CreateCompiler()` with `TargetPlatform::kUnknown`. And currently, it can't happen: The only use of `CreateCompiler()` is in the `Compiler` constructor on [line 432](https://github.com/flutter/flutter/blob/24ce716cfddfef201027c1a5fa2299a8aeffb03e/engine/src/flutter/impeller/compiler/compiler.cc#L432), and there is a check preventing `TargetPlatform::kUnknown` earlier on [line 292](https://github.com/flutter/flutter/blob/24ce716cfddfef201027c1a5fa2299a8aeffb03e/engine/src/flutter/impeller/compiler/compiler.cc#L292).

- `compiler_unittests.cc`, `compiler_test.h`
  - The `INSTANTIATE_{TARGET|RUNTIME_TARGET|SKSL_TARGET}_PLATFORM_TEST_SUITE_P` defines were oddly located in the file in the middle of the tests. Move them to the top of the file.
  - `INSTANTIATE_TARGET_PLATFORM_TEST_SUITE_P`
    - Remove the `kSkSL` target. These tests seem to be specifically for non-runtime targets, so SkSL doesn't belong here. All of these tests had a filter to skip for SkSL, so none of them actually ran for SkSL. These skips are now removed.
    - Add the `kVulkan` target, so all non-runtime platform targets are covered: opengles, openglesdesktop, metaldesktop, metalios, vulkan.
  - `INSTANTIATE_RUNTIME_TARGET_PLATFORM_TEST_SUITE_P`
    - This used to only test with `kRuntimeStageMetal`. For better coverage, I added all other runtime stages to this. It now tests on the metal, gles, gles3, vulkan, and sksl runtime targets.
    - Two tests, `UniformsAppearInJson` and `PositionedUniformsAppearInJson` fail when running with vulkan and with sksl. I added skips for these. I haven't dug deeper, but the failures seem unexpected to me. It's possible that this is revealing a bug with the vulkan and sksl compilers.
  - `INSTANTIATE_UNKNOWN_TARGET_PLATFORM_TEST_SUITE_P`
    - Added this new define for running tests with `TargetPlatform::kUnknown`.
    - Added a `MustFailDueToUnknownPlatform` test for this case. 

- `fixtures/BUILD.gn`, `runtime_stage_unittests.cc`
  - For the `impellerc("runtime_stages")` build target, add `--sksl` to the impellerc flags. This preserves the existing behavior of these targets being compiled for SkSL. They used to compile for SkSL because other runtime targets are specified. But now impellerc only compiles to SkSL when `--sksl` is explicitly specified.
  - Create a new `impellerc("runtime_stages_non_sksl")` target that runs impellerc without `--sksl`. Use this for a new `ContainsExpectedShaderTypesNoSksl` test in the unit test file. That test is the same as the existing `ContainsExpectedShaderTypes` test, but using the non-sksl output from `impellerc("runtime_stages_non_sksl")`.

### Update for commit 2 of the PR:
The original PR had an issue that failed CI because a build rule expected an impellerc output to include C++ reflection data, but the reflection data was not output. I added a sizable commit to address this:

- Fix compiler.gni logic around when to generate reflection state.
  - This used to incorrectly generate reflection state whenever the last shader_target_flags is not "--sksl".
  - Instead, generate reflection state when any non-runtime target is in shader_target_flags.
  - Consolidate some of the if/else logic to reduce duplicate code.
- Remove the TargetPlatformNeedsReflection check in impeller_main.cc.
  - Instead, whether reflection state is generated depends only on the presence or absense of "reflection_{json|header|cc}_name" flags.
  - The logic of whether to include these flags is already in compiler.gni. So it's redundant to also have logic for whether to generate the reflection state here.
  - The TargetPlatformNeedsReflection method had faulty logic.
    - It returned true for everything except SkSL, even though reflection state isn't needed for runtime targets.
    - It was called on the target from Switches::SelectDefaultTargetPlatform. When impellerc is used with multiple runtime targets, this would return the runtime target that is first alphabetically by flag name. So if --sksl is provided along with any other --runtime-stage-* target, SelectDefaultTargetPlatform returns the non-sksl runtime target. Effectively this meant that TargetPlatformNeedsReflection returns true except for when --sksl is the only provided runtime target.
  - Removes the TargetPlatformNeedsReflection function entirely.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/impeller/compiler/compiler.cc
+++ b/engine/src/flutter/impeller/compiler/compiler.cc
@@ -242,7 +242,6 @@ static CompilerBackend CreateCompiler(const spirv_cross::ParsedIR& ir,
     case TargetPlatform::kRuntimeStageVulkan:
       compiler = CreateVulkanCompiler(ir, source_options);
       break;
-    case TargetPlatform::kUnknown:
     case TargetPlatform::kOpenGLES:
     case TargetPlatform::kOpenGLDesktop:
     case TargetPlatform::kRuntimeStageGLES:
@@ -251,6 +250,9 @@ static CompilerBackend CreateCompiler(const spirv_cross::ParsedIR& ir,
       break;
     case TargetPlatform::kSkSL:
       compiler = CreateSkSLCompiler(ir, source_options);
+      break;
+    case TargetPlatform::kUnknown:
+      FML_UNREACHABLE();
   }
   if (!compiler) {
     return {};
--- a/engine/src/flutter/impeller/compiler/compiler_unittests.cc
+++ b/engine/src/flutter/impeller/compiler/compiler_unittests.cc
@@ -15,6 +15,52 @@ namespace impeller {
 namespace compiler {
 namespace testing {
 
+#define INSTANTIATE_TARGET_PLATFORM_TEST_SUITE_P(suite_name)                 \
+  INSTANTIATE_TEST_SUITE_P(                                                  \
+      suite_name, CompilerTest,                                              \
+      ::testing::Values(TargetPlatform::kOpenGLES,                           \
+                        TargetPlatform::kOpenGLDesktop,                      \
+                        TargetPlatform::kMetalDesktop,                       \
+                        TargetPlatform::kMetalIOS, TargetPlatform::kVulkan), \
+      [](const ::testing::TestParamInfo<CompilerTest::ParamType>& info) {    \
+        return TargetPlatformToString(info.param);                           \
+      });
+
+INSTANTIATE_TARGET_PLATFORM_TEST_SUITE_P(CompilerSuite);
+
+#define INSTANTIATE_RUNTIME_TARGET_PLATFORM_TEST_SUITE_P(suite_name)      \
+  INSTANTIATE_TEST_SUITE_P(                                               \
+      suite_name, CompilerTestRuntime,                                    \
+      ::testing::Values(TargetPlatform::kRuntimeStageMetal,               \
+                        TargetPlatform::kRuntimeStageGLES,                \
+                        TargetPlatform::kRuntimeStageGLES3,               \
+                        TargetPlatform::kRuntimeStageVulkan,              \
+                        TargetPlatform::kSkSL),                           \
+      [](const ::testing::TestParamInfo<CompilerTest::ParamType>& info) { \
+        return TargetPlatformToString(info.param);                        \
+      });
+
+INSTANTIATE_RUNTIME_TARGET_PLATFORM_TEST_SUITE_P(CompilerSuite);
+
+#define INSTANTIATE_SKSL_TARGET_PLATFORM_TEST_SUITE_P(suite_name)             \
+  INSTANTIATE_TEST_SUITE_P(                                                   \
+      suite_name, CompilerTestSkSL, ::testing::Values(TargetPlatform::kSkSL), \
+      [](const ::testing::TestParamInfo<CompilerTest::ParamType>& info) {     \
+        return TargetPlatformToString(info.param);                            \
+      });
+
+INSTANTIATE_SKSL_TARGET_PLATFORM_TEST_SUITE_P(CompilerSuite);
+
+#define INSTANTIATE_UNKNOWN_TARGET_PLATFORM_TEST_SUITE_P(suite_name)      \
+  INSTANTIATE_TEST_SUITE_P(                                               \
+      suite_name, CompilerTestUnknownPlatform,                            \
+      ::testing::Values(TargetPlatform::kUnknown),                        \
+      [](const ::testing::TestParamInfo<CompilerTest::ParamType>& info) { \
+        return TargetPlatformToString(info.param);                        \
+      });
+
+INSTANTIATE_UNKNOWN_TARGET_PLATFORM_TEST_SUITE_P(CompilerSuite);
+
 TEST(CompilerTest, Defines) {
   std::shared_ptr<const fml::Mapping> fixture =
       flutter::testing::OpenFixtureAsMapping("check_gles_definition.frag");
@@ -48,9 +94,6 @@ TEST(CompilerTest, ShaderKindMatchingIsSuccessful) {
 }
 
 TEST_P(CompilerTest, CanCompile) {
-  if (GetParam() == TargetPlatform::kSkSL) {
-    GTEST_SKIP() << "Not supported with SkSL";
-  }
   ASSERT_TRUE(CanCompileAndReflect("sample.vert"));
   ASSERT_TRUE(CanCompileAndReflect("sample.vert", SourceType::kVertexShader));
   ASSERT_TRUE(CanCompileAndReflect("sample.vert", SourceType::kVertexShader,
@@ -58,17 +101,11 @@ TEST_P(CompilerTest, CanCompile) {
 }
 
 TEST_P(CompilerTest, CanCompileHLSL) {
-  if (GetParam() == TargetPlatform::kSkSL) {
-    GTEST_SKIP() << "Not supported with SkSL";
-  }
   ASSERT_TRUE(CanCompileAndReflect(
       "simple.vert.hlsl", SourceType::kVertexShader, SourceLanguage::kHLSL));
 }
 
 TEST_P(CompilerTest, CanCompileHLSLWithMultipleStages) {
-  if (GetParam() == TargetPlatform::kSkSL) {
-    GTEST_SKIP() << "Not supported with SkSL";
-  }
   ASSERT_TRUE(CanCompileAndReflect("multiple_stages.hlsl",
                                    SourceType::kVertexShader,
                                    SourceLanguage::kHLSL, "VertexShader"));
@@ -87,18 +124,12 @@ TEST_P(CompilerTest, CanCompileComputeShader) {
 }
 
 TEST_P(CompilerTest, MustFailDueToExceedingResourcesLimit) {
-  if (GetParam() == TargetPlatform::kSkSL) {
-    GTEST_SKIP() << "Not supported with SkSL";
-  }
   ScopedValidationDisable disable_validation;
   ASSERT_FALSE(
       CanCompileAndReflect("resources_limit.vert", SourceType::kVertexShader));
 }
 
 TEST_P(CompilerTest, MustFailDueToMultipleLocationPerStructMember) {
-  if (GetParam() == TargetPlatform::kSkSL) {
-    GTEST_SKIP() << "Not supported with SkSL";
-  }
   ScopedValidationDisable disable_validation;
   ASSERT_FALSE(CanCompileAndReflect("struct_def_bug.vert"));
 }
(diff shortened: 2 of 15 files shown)


engine/src/flutter/impeller/compiler/compiler.cc
```diff
@@ -251,6 +250,9 @@ static CompilerBackend CreateCompiler(const spirv_cross::ParsedIR& ir,
       break;
     case TargetPlatform::kSkSL:
       compiler = CreateSkSLCompiler(ir, source_options);
+      break;
+    case TargetPlatform::kUnknown:
+      FML_UNREACHABLE();
```
gaaclarke, line 255: I thought this was mostly reserved for things that are truly unreachable?  Maybe this is unreachable but one can't validate it by looking at this code where it appears.  I think I'd rather have an FML_CHECK with an apropos error message here.

engine/src/flutter/impeller/compiler/compiler_unittests.cc
```diff
@@ -15,6 +15,52 @@ namespace impeller {
 namespace compiler {
 namespace testing {
 
+#define INSTANTIATE_TARGET_PLATFORM_TEST_SUITE_P(suite_name)                 \
```
gaaclarke: These macros are only used once, so why not just inline them?

engine/src/flutter/impeller/compiler/compiler_unittests.cc
```diff
@@ -202,6 +233,13 @@ inline std::ostream& operator<<(std::ostream& out, const UniformInfo& info) {
 }  // namespace
 
 TEST_P(CompilerTestRuntime, UniformsAppearInJson) {
+  if (GetParam() == TargetPlatform::kSkSL) {
+    GTEST_SKIP() << "Not supported with SkSL";
```
gaaclarke: For these skips, can we provide a couple more words about why they are not supported?  Like, are the unimplemented today, but will be some day?  Are they nonsensical, will they never be implemented, etc.

engine/src/flutter/impeller/compiler/shader_bundle.cc
```diff
@@ -215,7 +215,11 @@ bool GenerateShaderBundle(Switches& switches) {
   ///
 
   auto shader_bundle = GenerateShaderBundleFlatbuffer(
-      switches.shader_bundle, switches.CreateSourceOptions());
+      switches.shader_bundle,
+      switches.CreateSourceOptions(
+          // This TargetPlatform does not matter because it is overridden by
+          // GenerateShaderBackendFB().
+          TargetPlatform::kUnknown));
```
gaaclarke: Maybe it makes sense to override the method then for no arguments instead of sending in data that just gets ignored.

gaaclarke (changes requested):
Awesome, mostly looks good, just a few suggestions.

engine/src/flutter/impeller/compiler/compiler.cc
b-luk: This FML_UNREACHABLE is pretty much the same as the existing FML_UNREACHABLE for `case TargetPlatform::kUnknown:` in `EntryPointMustBeNamedMain()` above:
https://github.com/flutter/flutter/blob/e0e7a7d72db3fef30b221433e259e4e94a4d3aec/engine/src/flutter/impeller/compiler/compiler.cc#L212-L215

I could add a FML_CHECK, but what would I return for this case? If this case ever happens, things are in a bad state and I think the old behavior of returning a GLSL compiler doesn't actually make sense.

engine/src/flutter/impeller/compiler/shader_bundle.cc
b-luk: I changed CreateSourceOptions() to not take a TargetPlatform at all. The caller now needs to explicitly set the TargetPlatform after calling CreateSourceOptions(). I think this is a cleaner interface which accounts for some uses (specifically this one) not knowing the desired TargetPlatform until after CreateSourceOptions() is called.

engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -263,7 +213,8 @@ bool Main(const fml::CommandLine& command_line) {
   // Create at least one compiler to output the SL file, reflection data, and a
   // depfile.
 
-  SourceOptions options = switches.CreateSourceOptions();
+  SourceOptions options =
+      switches.CreateSourceOptions(switches.PlatformsToCompile().front());
```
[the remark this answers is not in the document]
b-luk: Added a comment about this.

engine/src/flutter/impeller/compiler/compiler_unittests.cc
b-luk: Done. Don't know why they were like this originally, but inlining them seems like a good change.

engine/src/flutter/impeller/compiler/compiler_unittests.cc
b-luk: After trying these out again, they do pass with SkSL but fail on Vulkan. Maybe some of the followup changes I made fixed SkSL somehow.

I filed https://github.com/flutter/flutter/issues/182578 to follow up for Vulkan. I don't know what the errors mean off the top of my head, but to me they look like they could possibly be something legit.

engine/src/flutter/impeller/compiler/compiler_unittests.cc
```diff
@@ -202,6 +233,13 @@ inline std::ostream& operator<<(std::ostream& out, const UniformInfo& info) {
...
+  }
+  if (GetParam() == TargetPlatform::kRuntimeStageVulkan) {
+    GTEST_SKIP() << "Not supported with Vulkan";
+  }
```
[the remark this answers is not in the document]
b-luk: Created an issue to follow up, and added comments here.

engine/src/flutter/impeller/compiler/switches.h
```diff
@@ -57,9 +57,11 @@ class Switches {
   /// A vector containing at least one valid platform.
   std::vector<TargetPlatform> PlatformsToCompile() const;
 
-  // Creates source options from these switches for the specified
-  // TargetPlatform.
-  SourceOptions CreateSourceOptions(TargetPlatform target_platform) const;
+  // Creates source options from these switches. The returned options does not
+  // have a set TargetPlatform because that cannot be determined based purely
+  // on the switches. Clients must set a valid TargetPlatform on the returned
+  // options before before it is used.
```
gaaclarke: This should be in the doxygen format (`///`)

gaaclarke (approved):
LGTM! thanks benson

engine/src/flutter/impeller/compiler/switches.h
b-luk: Done

gaaclarke:
The last commit came in after this was added to the merge queue.  I'm not sure how that's going to shake out, not a huge deal in this case but just a heads up @b-luk

PR #181632, opened 2026-01-28
[Android] Add mechanism for setting Android engine flags via Android manifest (take 2)

camsim99:
A continuation of https://github.com/flutter/flutter/pull/177127. Copying from there:

# Overview

> [!NOTE]  
> This PR is based on conversation & feedback on go/flutter-android-harden-engine-shell-arguments.

Adds a mechanism for setting Android engine flags via the manifest. If a flag is specified on the command line and in manifest metadata, the value specified on the command line will take precedence. Documentation is added on this mechanism

Additionally, this PR removes the exposure of`--cache-sksl` command line flag as per [https://github.com/flutter/flutter/issues/140310#issuecomment-2708459007](https://www.google.com/url?q=https://github.com/flutter/flutter/issues/140310%23issuecomment-2708459007&sa=D&source=docs&ust=1761156167162464&usg=AOvVaw3a8ubXTtv3apknY2-P9dKe).

Additionally, this PR adds documentation for the only two supported ways of setting engine flags moving forward -- via the command line or manifest. The `Intent` mechanism will be removed when https://github.com/flutter/flutter/issues/180686 is completed (intended to be a follow up to this PR).

As the unit tests in this PR only cover setting flags via manifest in debug mode, I will follow up this PR with an integration test to test that flags are appropriately respected/ignored in release mode. See https://github.com/flutter/flutter/pull/178383 for my currently working but WIP draft.

Part of https://github.com/flutter/flutter/issues/172553.

# Follow up work:

## Add integration test for this new added mechanism
This will land as an immediate follow-up to this PR. WIP in https://github.com/flutter/flutter/pull/178383.

## Remove support for setting shell arguments via `Intent`
This task will be a follow up to this work + the integration test landing, and will complete work for https://github.com/flutter/flutter/issues/172553. See https://github.com/flutter/flutter/issues/180686 for details.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
[...]

Issue #180686, opened 2026-01-08
[Android] Remove the ability to set engine shell arguments via `Intent` extras


camsim99:
Part of https://github.com/flutter/flutter/issues/172553 -- hardening `FlutterActivity` and `FlutterFragment` against attacks via vulnerabilities of using `Intent` extras to specify engine shell arguments.

When this task is done, the following will be true:

1. Flags **CANNOT** be set via [`Intent`](https://developer.android.com/reference/android/content/Intent), which is exploitable by malicious actors, particularly in add-to-app scenarios. See go/flutter-fragment-flag-security-analysis for an example of such an exploitation.
2. As a consequence of 1, flags that previously could be set dynamically upon the launch of an `Intent` can no longer directly be set dynamically. Instead, they must be set statically in the manifest or on the command line.

This means removing support for setting flags via `Intent` extras for `FlutterActivity` and `FlutterFragment` by developers **and** the Flutter tool.

camsim99:
Closing in favor of https://github.com/flutter/flutter/issues/190461. See that issues for details on the pivot!

PR #182522, opened 2026-02-18
Reland "[Android] Add mechanism for setting Android engine flags via Android manifest (take 2)"

camsim99:
### Overview
Re-lands https://github.com/flutter/flutter/pull/181632 which TLDR does the following:

**Same old same old changes from original PR**
- Refactors all engine flags recognized by the Flutter Android embedding (specifically those previously recognized by [`FlutterShellArgs`](https://github.com/flutter/flutter/blob/4f5478cce38d837e14b7a032a12500d3fc0f1310/engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterShellArgs.java) and all of those used to initialize the engine in `FlutterLoader`) into a new class `FlutterEngineFlags` (`FlutterShellArgs` is kept for now to maintain backwards compatibility; see https://github.com/flutter/flutter/issues/182153 for details)
- Adds the ability to set engine flags (that are recognized by `FlutterEngineFlags`) via the manifest
- Adds warnings for developers that set engine flags via `Intent`s as this will be disabled as part of https://github.com/flutter/flutter/issues/180686.
- Disallows debug/test only flags in release mode

**New changes/fixes**

> [!NOTE]
> This re-land was required because when previously submitted, it caused b/483299339. I have confirmed that this bug has been fixed in this PR; see go/flutter-android-setting-flags-in-manifest-perf-comparison for A/B test results with the previous version of this PR (https://github.com/flutter/flutter/pull/181632) and commit https://github.com/flutter/flutter/pull/182522/commits/d6d4a07ede2052db4a861528ff05df8175bd9d37 of this PR.

- [Critical change] To fix the bug caught by b/483299339, `--merged-platform-ui-thread` flag is added to `FlutterEngineFlag`s 
- [Behavior change, not critical] To avoid more bugs like b/483299339, `FlutterLoader` allows unrecognized flags specified by the command line or `Intent` extras to be used to initialize the engine for now (same behavior as before my original change but I plan to secure this in the future; see https://github.com/flutter/flutter/issues/182557)
- [Small refactor] To speed up manifest argument processing time, `FlutterLoader` loops through all recognized `FlutterEngineFlag`s to check if they are present in an application's manifest versus looping through all metadata in an application's manifest and looking for recognized `FlutterEngineFlag`s (small improvement for apps with large manifests)
- [Small refactor] To speed up command line argument processing time, `FlutterLoader` processes the `--aot-shared-library-name` flag in the same loop as the other flags, reducing the number of loops `ensureInitializationComplete` has to do (minimal improvement perf-wise but kept it in because it's more readable to me)

### Full description
(Mostly copied from https://github.com/flutter/flutter/pull/181632)

> [!NOTE]  
> This PR is based on conversation & feedback on go/flutter-android-harden-engine-shell-arguments.

Adds a mechanism for setting Android engine flags via the manifest. If a flag is specified on the command line and in manifest metadata, the value specified on the command line will take precedence. Documentation is added on this mechanism

Additionally, this PR removes the exposure of`--cache-sksl` command line flag as per [https://github.com/flutter/flutter/issues/140310#issuecomment-2708459007](https://www.google.com/url?q=https://github.com/flutter/flutter/issues/140310%23issuecomment-2708459007&sa=D&source=docs&ust=1761156167162464&usg=AOvVaw3a8ubXTtv3apknY2-P9dKe).

Additionally, this PR adds documentation for the only two supported ways of setting engine flags moving forward -- via the command line or manifest. The `Intent` mechanism will be removed when https://github.com/flutter/flutter/issues/180686 is completed (intended to be a follow up to this PR).

As the unit tests in this PR only cover setting flags via manifest in debug mode, I will follow up this PR with an integration test to test that flags are appropriately respected/ignored in release mode. See https://github.com/flutter/flutter/pull/178383 for my currently working but WIP draft.

Part of https://github.com/flutter/flutter/issues/172553.

# Follow up work:

- [ ] **Migrate flags all un-migrated flags `FlutterEngineFlags`** https://github.com/flutter/flutter/issues/182852
- [ ]  **Add integration test for this new added mechanism.** This will land as an immediate follow-up to this PR. WIP in https://github.com/flutter/flutter/pull/182241.
- [ ] **Remove support for setting shell arguments via `Intent`.** This task will be a follow up to this work + the integration test landing, and will complete work for https://github.com/flutter/flutter/issues/172553. See https://github.com/flutter/flutter/issues/180686 for details.
- [ ] **Only allow known engine flags to be passed to the engine.** This is more of a P2 but would be a nice additional security feature; see https://github.com/flutter/flutter/issues/182557.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
```diff
@@ -474,27 +520,66 @@ public void ensureInitializationComplete(
...
+   *
+   * <p>If the path is safe, it will be added to the beginning of the arguments list of arguments.
+   * The earlier specified path takes precedence over any later specified paths for the AOT shared
+   * library name argument.
+   */
+  private void maybeAddAotSharedLibraryNameArg(
+      @NonNull Context applicationContext,
+      @NonNull String aotSharedLibraryPath,
+      @NonNull List<String> shellArgs) {
+    String safeAotSharedLibraryName = null;
```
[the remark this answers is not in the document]
camsim99: Added a warning for null specified paths instead.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
```diff
@@ -38,49 +39,10 @@
...
-      "io.flutter.embedding.android.EnableVulkanValidation";
-  private static final String IMPELLER_BACKEND_META_DATA_KEY =
-      "io.flutter.embedding.android.ImpellerBackend";
-  private static final String IMPELLER_OPENGL_GPU_TRACING_DATA_KEY =
-      "io.flutter.embedding.android.EnableOpenGLGPUTracing";
-  private static final String IMPELLER_VULKAN_GPU_TRACING_DATA_KEY =
-      "io.flutter.embedding.android.EnableVulkanGPUTracing";
-  private static final String DISABLE_MERGED_PLATFORM_UI_THREAD_KEY =
-      "io.flutter.embedding.android.DisableMergedPlatformUIThread";
-  private static final String ENABLE_SURFACE_CONTROL =
```
gmackall: I think not migrating this flag will cause the [Linux_android_emu_vulkan_stable android_engine_vulkan_tests](https://github.com/flutter/flutter/blob/master/.ci.yaml#L1734) target to fail (which only runs in post submit).

It relies on parsing the flag for the HCPP integration tests
https://github.com/flutter/flutter/blob/692b51763d45cc4d574a06cf3d7b9b36f69c5170/dev/bots/suite_runners/run_android_engine_tests.dart#L108-L124

When the initial version of this pr landed this target was mistakenly in bringup, so the failure wasn't visible, but it has since been restored.

camsim99: Oh wow I'm glad you brought this up because I meant for all of these flags to be migrated to `FlutterEngineFlags`. Will fix!

camsim99: Sooo I actually did convert these but later deleted them because I thought they weren't being used 🤦‍♀️  added back in https://github.com/flutter/flutter/pull/182522/commits/89161405c8bfba3332523ec7d1d9776fc3d2da61

gmackall: Nice, yeah it's a bit confusing because some of the flags are parsed from the tool + manifest but some only from the manifest, so they look unused if you're only looking at tool code

camsim99: Oh I need one more change...I introduced a subtle bug because some flags like the enable surface control flag take a value in the manifest but does not on the command line, so I need to make sure that case is properly handled.

camsim99: Fixed in https://github.com/flutter/flutter/pull/182522/changes/5fca5fd1a9ac75f907c735028db6c9e4f7f60626

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineFlags.java
```diff
@@ -0,0 +1,462 @@
...
+      String key = commandLineArg.startsWith("--") ? commandLineArg.substring(2) : commandLineArg;
+      if (key.endsWith("=")) {
+        key = key.substring(0, key.length() - 1);
+      }
+      if (key.equals(intentKey)) {
+        return flag;
+      }
+    }
+    return null;
+  }
```
mboetger, line 536: Instead of looping over everything, it would probably be faster to look for the command line flag from the intent key: 

```suggestion
    Flag toReturn = getFlagByCommandLineArgument(intentKey);
    if (toReturn == null) {
      toReturn =getFlagByCommandLineArgument("--" + intentKey);
    }
    if (toReturn == null) {
      toReturn = getFlagByCommandLineArgument("--" + intentKey + "=");
    }
```

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
gmackall: Should the info in `docs/engine/Flutter-Android-Engine-Flags.md` be updated? I assume it should indicate that some subset of the flags cannot be set via the command line, right

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -331,6 +333,7 @@ private FlutterEngineGroup.Options addEntrypointOptions(FlutterEngineGroup.Optio
         "No preferred FlutterEngine was provided. Creating a new FlutterEngine for"
             + " this FlutterFragment.");
 
+    warnIfEngineFlagsSetViaIntent(host.getActivity().getIntent());
```
gmackall, line 336: If this PR doesn't touch the tool, won't we be sending this warning for flags that we (flutter) ourselves are sending (as the tool is still using the intent mechanism)? Should we hold off on warning (or at most print the log as a `Log.i`) until we fix our own case of passing engine args via intent through the tool? Because otherwise developers will see this action (probably primarily) in cases where it isn't actionable

e.g. using a flutter tool arg 
https://github.com/flutter/flutter/blob/915a972eb2d69cea5c21e366b36b7e83a6b0b4ba/packages/flutter_tools/lib/src/android/android_device.dart#L626
shouldn't print error level warnings I don't think

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
camsim99: Sure, I can. I can also go through FlutterEngineFlags and comment which flags are settable via the command line and which ones are not. Will update when I get to this.

camsim99: Done. Noted which flags are only available on the command line (https://github.com/flutter/flutter/blob/ebfe8ce7a0f52f834c70a9f782c35cb1d80bc09d/packages/flutter_tools/lib/src/commands/run.dart#L4). Some appear to have different names than the actual engine argument, so I noted that where I saw it. I also did some renaming to clarify that `FlutterEngineFlags` maps the engine flag to the manifest flag and not the command line flag (if it exists).

mboetger (dismissed):
LGTM

engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gmackall: sorry I added this comment as part of a a review instead of single comment, so it didn't appear

camsim99: I do want to keep the warning, just so that I can start giving folks a heads up but fair point. Reworded the warning and changed the log level to info! I will likely rework this warning in the PR where I fix the tool so that people actually start migrating away from using Intents.

gmackall: Yeah I think raising the log level is perfectly fair once we fix our own tool use. My main point was just that someone will pass a flag and see this warning, and probably simply not understand it (as they are not passing anything via intents, and can't be reasonably expected to understand the internals of the flutter tool).

camsim99: Yeah totally fair, I don't want people confused! Thanks for pointing this out!

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/FlutterEngineFlags.java
```diff
@@ -0,0 +1,557 @@
...
+   * Specifies the path to the isolate snapshot data file.
+   *
+   * <p>Allowed in release to support different snapshot configurations. Only settable via the
+   * manifest.
+   */
+  public static final Flag ISOLATE_SNAPSHOT_DATA =
+      new Flag("--isolate-snapshot-data=", "IsolateSnapshotData", true);
+
+  /**
+   * Enables Android SurfaceControl for rendering.
```
gmackall: I think this should be removed now in favor the the below `ENABLE_HCPP`, right?

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/ApplicationInfoLoader.java
```diff
@@ -10,24 +10,19 @@
...
-  public static final String PUBLIC_VM_SNAPSHOT_DATA_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.VM_SNAPSHOT_DATA_KEY;
-  public static final String PUBLIC_ISOLATE_SNAPSHOT_DATA_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.ISOLATE_SNAPSHOT_DATA_KEY;
-  public static final String PUBLIC_FLUTTER_ASSETS_DIR_KEY =
-      FlutterLoader.class.getName() + '.' + FlutterLoader.FLUTTER_ASSETS_DIR_KEY;
+  // TODO(camsim99): Remove support for these flags:
+  // https://github.com/flutter/flutter/issues/179276.
+  // AndroidManifest.xml metadata keys for setting internal respective Flutter configuration values.
   public static final String NETWORK_POLICY_METADATA_KEY = "io.flutter.network-policy";
```
mboetger, line 23: Just fyi, this is not connected to anything.

camsim99: I filed https://github.com/flutter/flutter/issues/179276 to remember to remove it!

mboetger (approved):
LGTM

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/loader/FlutterLoader.java
```diff
@@ -310,59 +287,188 @@ public void ensureInitializationComplete(
...
             }
+            shellArgs.add(arg + value);
+            continue;
+          }
+
+          // Check if a boolean value is specified and if so, use it to determine if the
+          // flags should be added. If not, assume the flag is meant to be added.
+          if (applicationMetaData.getBoolean(metadataKey, true)) {
+            shellArgs.add(arg);
+          }
```
gmackall, line 402: Isn't defaulting to true a behavior change for many of these flags?

`IMPELLER_VULKAN_GPU_TRACING_DATA_KEY` for example, defaulted to false before

Should we instead have a default value defined on the flag to fall back to? I.e. as part of the interface?

PR #182527, opened 2026-02-18
Add RawTooltip.ignorePointer

### LINKED ISSUE #182468: [flutter 3.41.0] Tooltip ignorePointer property not working
### Steps to reproduce

use `Tooltip` with `meesage` instead of `richMessage` or set `ignorePointer` to `true`, after showing the tooltip, hover the tooltip, tooltip keeps visible on hover

Investigation:
before flutter 3.41.0:
`IgnorePointer` is outside the `_ExclusiveMouseRegion`
after flutter 3.41.0:
`_ExclusiveMouseRegion` is outside the `IgnorePointer`

### Expected results

tooltip dissmissed on hover (flutter 3.38.x)

### Actual results

tooltip keeps visible on hover (flutter 3.41.0)

### Code sample

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: .fromSeed(seedColor: Colors.deepPurple),
      ),
    );
  }
}

  final String title;

  @override
}

  @override
    return Scaffold(
      body: Center(
        child: Tooltip(
          message: 'tooltip',
          // or this
          // ignorePointer: true,
          // richMessage: TextSpan(text: 'tooltip'),
          child: Icon(Icons.more_vert),
        ),
      ),
    );
  }
}

```

</details>

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
no error log
```

</details>

### Flutter Doctor output

```console
[Paste your output here]
```

</details>

victorsanni:
https://github.com/flutter/flutter/issues/182468#issuecomment-3917883694

Fixes [[flutter 3.41.0] Tooltip ignorePointer property not working](https://github.com/flutter/flutter/issues/182468)

### DIFF
--- a/packages/flutter/test/material/tooltip_test.dart
+++ b/packages/flutter/test/material/tooltip_test.dart
@@ -2055,6 +2055,39 @@ void main() {
     expect(tester.getSize(find.byType(Tooltip)), Size.zero);
     expect(find.text('X'), findsOne);
   });
+
+  testWidgets('Tooltip should disappear even if mouse is hovering over the tooltip overlay', (
+    WidgetTester tester,
+  ) async {
+    await tester.pumpWidget(
+      const MaterialApp(
+        home: Center(
+          child: Tooltip(message: tooltipText, child: Text('Hover me')),
+        ),
+      ),
+    );
+
+    // Hover over the target widget.
+    final TestGesture gesture = await tester.createGesture(kind: PointerDeviceKind.mouse);
+    await gesture.addPointer(location: Offset.zero);
+    addTearDown(gesture.removePointer);
+
+    final Finder targetFinder = find.text('Hover me');
+    await gesture.moveTo(tester.getCenter(targetFinder));
+    await tester.pump();
+
+    // Wait for the tooltip to actually appear.
+    await tester.pump(const Duration(milliseconds: 100));
+    expect(find.text(tooltipText), findsOneWidget);
+
+    // Move the mouse to the Tooltip overlay.
+    final Finder tooltipOverlayFinder = find.text(tooltipText);
+    await gesture.moveTo(tester.getCenter(tooltipOverlayFinder));
+    await tester.pumpAndSettle();
+
+    // Verify the tooltip overlay is no longer displayed.
+    expect(find.text(tooltipText), findsNothing);
+  });
 }
 
 Future<void> _testGestureTap(WidgetTester tester, Finder tooltip) async {
--- a/packages/flutter/test/material/tooltip_test.dart
+++ b/packages/flutter/test/material/tooltip_test.dart
@@ -2080,7 +2080,7 @@ void main() {
     await tester.pump(const Duration(milliseconds: 100));
     expect(find.text(tooltipText), findsOneWidget);
 
-    // Move the mouse to the Tooltip overlay.
+    // Move the mouse to the tooltip overlay.
     final Finder tooltipOverlayFinder = find.text(tooltipText);
     await gesture.moveTo(tester.getCenter(tooltipOverlayFinder));
     await tester.pumpAndSettle();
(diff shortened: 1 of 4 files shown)


dkwingsmt (approved):
LGTM. Thank you!

packages/flutter/test/material/tooltip_test.dart
```diff
@@ -2055,6 +2055,39 @@ void main() {
     expect(tester.getSize(find.byType(Tooltip)), Size.zero);
     expect(find.text('X'), findsOne);
   });
+
+  testWidgets('Tooltip should disappear even if mouse is hovering over the tooltip overlay', (
```
navaronbracke, line 2059: Do we need a regression test that verifies that `Tooltip` correctly passes the ignore pointer flag to the RawTooltip?

packages/flutter/test/widgets/raw_tooltip_test.dart
```diff
@@ -2740,6 +2741,45 @@ void main() {
...
+    // Hover over the target widget.
+    final TestGesture gesture = await tester.createGesture(kind: PointerDeviceKind.mouse);
+    await gesture.addPointer(location: Offset.zero);
+    addTearDown(gesture.removePointer);
+
+    final Finder targetFinder = find.text('Hover me');
+    await gesture.moveTo(tester.getCenter(targetFinder));
+    await tester.pump();
+
+    // Wait for the tooltip to actually appear.
```
navaronbracke, line 2771: What is the difference between this test and the one named "Tooltip should disappear even if mouse is hovering over the tooltip overlay" ? Should this test not assert that the tooltip is never displayed at all, when hovered over the child?

navaronbracke (changes requested):
Overall LGTM, but I have some questions for the tests.

chunhtai (approved):
LGTM, once the remaining comments are addressed

packages/flutter/test/material/tooltip_test.dart
victorsanni: There are other tests in this file that confirm `Tooltip.ignorePointer` works (both when true and false). Before this PR it also worked-ish (by wrapping the `RawTooltip.tooltipBuilder` in an `IgnorePointer` so that the logic stays in `Tooltip`) except for this edge case because of `_ExclusiveMouseRegion` in `RawTooltip`.

packages/flutter/test/widgets/raw_tooltip_test.dart
victorsanni: > What is the difference between this test and the one named "Tooltip should disappear even if mouse is hovering over the tooltip overlay"

It's the same test, but for `RawTooltip` instead of `Tooltip`. The difference is `Tooltip` has `ignorePointer` set to true (in most cases) by default, while `RawTooltip` has it default to false. But this is really the behavior we are guarding - the tooltip overlay child should ignore mouse pointers if `ignorePointer` is set to true (and so be dismissed even when hovered, as any other surface that is not the child).

> Should this test not assert that the tooltip is never displayed at all, when hovered over the child?

Not really, this is not the intention of `ignorePointer`. `ignorePointer` exists to allow the developer to ignore pointers over the overlay child, not the child. If ignoring pointers over the child is desired, the child itself can be wrapped in an `IgnorePointer` by the developer.

Maybe the documentation doesn't explain this well enough? Is this something we can change?

navaronbracke: I didn't take a look at the widget documentation for this to be honest. We can do the doc change in a follow-up if you prefer.

navaronbracke (approved):
LGTM

Issue #44269, opened 2019-11-06
[TabBar] Setting isScrollable in TabBar to true causes exception when navigating between pages


perclasson:
Steps to reproduce
==============
(this is on the master branch of flutter)

1. Create an app with a start page and a main page with a TabBar that has isScrollable set to true (see this gist: https://gist.github.com/perclasson/912d072cdb9cd17e0fb837d9afe0b309).

2. Go from the start page (by clicking on the button) and you will get an exception on the tab bar page.

Logs 
=====
    
    [VERBOSE-2:ui_dart_state.cc(157)] Unhandled Exception: 'package:flutter/src/widgets/scroll_position.dart': Failed assertion: line 696 pos 12: 'pixels != null': is not true.
    #0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:40:39)
    #1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:36:5)
    #2      ScrollPosition.dispose (package:flutter/src/widgets/scroll_position.dart:696:12)
    #3      ScrollPositionWithSingleContext.dispose (package:flutter/src/widgets/scroll_position_with_single_context.dart:256:11)
    #4      _rootRun (dart:async/zone.dart:1120:38)
    #5      _CustomZone.run (dart:async/zone.dart:1021:19)
    #6      _CustomZone.runGuarded (dart:async/zone.dart:923:7)
    #7      _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:963:23)
    #8      _rootRun (dart:async/zone.dart:1124:13)
    #9      _CustomZone.run (dart:async/zone.dart:1021:19)
    #10     _CustomZone.runGuarded (dart:async/zone.dart:923:7)
    #11     _CustomZone.bindCallbackGuarded.<anonymous closure<…>

Flutter Doctor
==========
    [✓] Flutter (Channel master, v1.10.1-pre.867, on Mac OS X 10.14.6 18G95, locale en-DE)
        • Flutter version 1.10.1-pre.867 at /Users/perc/development/flutter
        • Framework revision 289b4588f4 (6 hours ago), 2019-11-06 02:51:17 -0500
        • Engine revision 9726b4cb99
        • Dart version 2.7.0

    [✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
        • Android SDK at /Users/perc/Library/Android/sdk
        • Android NDK location not configured (optional; useful for native profiling support)
[...]

PR #181774, opened 2026-02-01
Remove material import from toggleable_test.dart + draggable_test.dart + obscured_animated_image_test.dart + sliver_constraints_test.dart

navaronbracke:
This pull request removes the material import from

- toggleable_test.dart
- draggable_test.dart
- obscured_animated_image_test.dart

It also cleans up usages of `PageRoute` for testing.

Part of https://github.com/flutter/flutter/issues/177415
Part of https://github.com/flutter/flutter/issues/181090

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] I followed the [breaking change policy] and added [Data Driven Fixes] where supported.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[Contributor Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md
[Features we expect every widget to implement]: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#features-we-expect-every-widget-to-implement
[CLA]: https://cla.developers.google.com/
[...]

PR #182546, opened 2026-02-18
Remove material imports from navigator_on_did_remove_page_test and scrollable_in_overlay_test

apoorvdarshan:
## Summary
- Replaced `MaterialPage`/`MaterialPageRoute` with a local `_TestPage`/`PageRouteBuilder` in `navigator_on_did_remove_page_test.dart`
- Replaced `TabBar`/`TabController`/`MaterialApp`/`Material` with `SingleChildScrollView`/`Directionality` in `scrollable_in_overlay_test.dart`
- Removed both files from the cross-import allowlist in `check_tests_cross_imports.dart`

Part of https://github.com/flutter/flutter/issues/177414

## Test plan
- [ ] Verify tests pass in CI with the material imports removed
- [ ] Regression behavior for https://github.com/flutter/flutter/issues/44269 is preserved


packages/flutter/test/widgets/navigator_on_did_remove_page_test.dart
navaronbracke, line 1: This test should not be moved. Instead, the uses of MaterialPage should be fixed instead.

https://github.com/flutter/flutter/pull/181774 adds a TestRoute for this purpose.

packages/flutter/test/widgets/scrollable_in_overlay_test.dart
navaronbracke, line 1: This test should not be moved. Instead, fix the test named "scrollable in hidden overlay does not crash when unhidden" to not use a TabBar, while still maintaining the behavior of the regression test.

So probably replace it with a SingleChildScrollView or similar.

navaronbracke (changes requested):

justinmc (dismissed):
LGTM 👍 . Thanks for helping out with this!

justinmc (dismissed):
Re LGTM 👍

apoorvdarshan:
Rebased onto latest `master` and resolved conflicts.

Notes on the resolution:
- `dev/bots/check_tests_cross_imports.dart`: the branch's base had a handful of entries that have since been fixed and removed upstream (`image_filter_test`, `baseline_test`, `editable_text_show_on_screen_test`, `system_context_menu_test`, `scrollable_fling_test`, `snapshot_widget_test`). Dropped those so the `_differencePaths` check doesn't fail.
- `packages/flutter/test/widgets/navigator_on_did_remove_page_test.dart`: upstream already removed the Material dependency here using the new shared `TestPage` helper from `test_page_tester.dart`. Took that version rather than keeping the local `_TestPage` duplicate.
- `scrollable_in_overlay_test.dart` refactor (`TabBar`/`MaterialApp` → `SingleChildScrollView`/`Directionality`) is preserved and is now the only net change this PR adds on top of `master`.

PTAL when convenient.

justinmc (approved):
LGTM 👍

victorsanni (approved):

navaronbracke (approved):

PR #182580, opened 2026-02-19
framework: Use a super-parameter in several missed cases

srawlins:
Work towards https://github.com/dart-lang/sdk/issues/58729

The lint rule will suggest using a super-parameter for a named parameter even if the positional parameters cannot be super-parameters.


justinmc (approved):
LGTM 👍 . I would have thought these cases should already be caught by the linter.

PR #182624, opened 2026-02-19
Manual dart sdk flutter 174bcc79 25ff 4267 8e26 d0e902f18681 1771486449

gaaclarke:
fixes https://github.com/flutter/flutter/issues/182617

manual roll for https://github.com/flutter/flutter/pull/182591

https://dart.googlesource.com/sdk.git/+log/2642761fca94..f8fac50475b8


jtmcdole (approved):

dev/integration_tests/widget_preview_scaffold/lib/src/theme/ide_theme.dart
```diff
@@ -14,6 +14,7 @@ export '_ide_theme_desktop.dart'
 
 /// IDE-supplied theming.
 final class IdeTheme {
+  // ignore: prefer_initializing_formals
   const IdeTheme({this.backgroundColor, this.foregroundColor, bool? isDarkMode})
     : _isDarkMode = isDarkMode;
```
munificent, line 19: The lint will fire on this line. If the ignore comment seems to be working where it is, then SGTM, but you may need to move it to the end of this line or one line above this one. Likewise below.

munificent (approved):

dev/integration_tests/widget_preview_scaffold/lib/src/theme/ide_theme.dart
gaaclarke: I wasn't able to reproduce the failure locally to verify.  Gemini told me this is the right place for it to go.  I originally had it like this:

```dart
  const IdeTheme({this.backgroundColor, this.foregroundColor, bool? isDarkMode})
    // ignore: prefer_initializing_formals
    : _isDarkMode = isDarkMode;
```

Having it at the beginning of the statement makes sense to me since there can be multiple instances of initializers, I'm not sure if the analyzer agrees.

gaaclarke: I moved them, the old position didn't work as suspected.  It is weird because in that one case we have 2 initializers, but only one lint failure is issued.  I'm guessing one ignore is sufficient since only one failure was issued.

dev/integration_tests/widget_preview_scaffold/lib/src/controls.dart
```diff
@@ -13,6 +13,7 @@ class ZoomControls extends StatelessWidget {
   const ZoomControls({
     super.key,
     required TransformationController transformationController,
+    // ignore: prefer_initializing_formals
   }) : _transformationController = transformationController;
```
munificent, line 17: Style nit, but I'd put the comment outside of the parameter list so it's clearer that it applies to the initializer:

```dart
const ZoomControls({
    super.key,
    required TransformationController transformationController,
  }) : // ignore: prefer_initializing_formals
       _transformationController = transformationController;
```

Likewise elsewhere, but it's up to you.

munificent (approved):
LGTM! Thanks for fixing what I broke and unblocking the roll.

dev/integration_tests/widget_preview_scaffold/lib/src/controls.dart
gaaclarke: Oh yea, that's nicer.  I unfortunately didn't have a setup where i could reproduce the problem locally so I couldn't play around with it.  I didn't know the analyzer can recognized ignores if there is non-whitespace before the comment.

PR #182649, opened 2026-02-20
Fix a race in EmbedderTest.CanSpecifyCustomUITaskRunner

jason-simmons:
The UI task runner will be destroyed during the call to FlutterEngineShutdown.  It should continue forwarding tasks to the engine until the destruction callback is invoked.  After that, it should stop using the engine.

The test had been using a mutex to guard the engine, but that does not work in this case because both the platform and UI threads must be able to run tasks during engine shutdown.  If the test's platform thread holds the mutex during the call to engine.reset(), then the UI task runner can not acquire it to run its tasks. The test previously avoided that by calling engine.reset() without holding the mutex.  That creates a race between engine shutdown and the UI task runner's check of engine.is_valid().


engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -206,22 +206,30 @@ TEST_F(EmbedderTest, CanSpecifyCustomUITaskRunner) {
   auto& context = GetEmbedderContext<EmbedderTestContextSoftware>();
   auto ui_thread = std::make_unique<fml::Thread>("test_ui_thread");
   auto ui_task_runner = ui_thread->GetTaskRunner();
+  std::mutex ui_task_runner_mutex;
+  bool ui_task_runner_destroyed = false;
```
gaaclarke, line 210: Please add the IPLR_GUARDED_BY macro to this variable.

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -444,7 +449,7 @@ TEST_F(EmbedderTest, CanSpecifyCustomPlatformTaskRunner) {
   auto platform_task_runner = CreateNewThread("test_platform_thread");
   static std::mutex engine_mutex;
   static bool signaled_once = false;
-  static std::atomic<bool> destruction_callback_called = false;
+  std::atomic<bool> destruction_callback_called = false;
```
gaaclarke, line 458: This variable isn't read anywhere.

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -466,7 +471,7 @@ TEST_F(EmbedderTest, CanSpecifyCustomPlatformTaskRunner) {
         latch.Signal();
       });
   test_task_runner.SetDestructionCallback(
-      [](void* user_data) { destruction_callback_called = true; });
+      [&]() { destruction_callback_called = true; });
```
gaaclarke: The convention at google is to use explicit load() and store() functions with std::atomics.

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests_util.h
```diff
@@ -130,12 +130,17 @@ class EmbedderTestTaskRunner {
 
       real_task_runner->PostTaskForTime(invoke_task, target_time);
     };
-    task_runner_description_.destruction_callback = [](void* user_data) {};
+    task_runner_description_.destruction_callback = [](void* user_data) {
+      auto thiz = reinterpret_cast<EmbedderTestTaskRunner*>(user_data);
+      if (thiz->destruction_callback_) {
+        thiz->destruction_callback_();
```
gaaclarke, line 146: Is this executing on a different thread than SetDestructionCallback?  If so there is a race condition and we'll need a mutex.

gaaclarke (changes requested):
Nice catch.

Test exempt: this is a test

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
jason-simmons: The `IPLR_` macros are defined in Impeller, but this test is in the engine tree outside Impeller.  The engine/shell subsystems are not using the threading annotations.

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
jason-simmons: It's checked by an assertion later in this test.

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
jason-simmons: done

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests_util.h
jason-simmons: Changed this callback to be immutable after construction

engine/src/flutter/shell/platform/embedder/tests/embedder_unittests.cc
gaaclarke: It's a couple of lines you can inline in this file.  It's basically a preprocessor check for clang and an attribute on the variable.

gaaclarke (approved):
LGTM modulo a nit about clang's guarded_by attribute.  Thanks jason.

PR #182674, opened 2026-02-20
Refactor bin/ shell scripts for better performance and safety

kevmoo:
- Utilized native bash parameter expansion and file reading mechanisms like `$(< file)` instead of sub-shelling external invocations like `$(cat file)`.

- Enforced path-safety across internal directory traversals by appending -- to cd and rm commands, insulating scripts against files/folders that might start with a hyphen.

- Restructured chmod application using find -exec ... + to batch parameter arguments, bypassing the overhead of spinning up individual process instances per matched file.

- Cleaned up control flow logic, replacing redundant $? and set +e manipulations with graceful || true fallback patterns and more idiomatic if list; then blocks.


bin/internal/content_aware_hash.sh
```diff
@@ -50,18 +50,15 @@ if [[ "$CURRENT_BRANCH" != "main" && \
   # This is a development branch. Find the merge-base.
   # We will fallback to origin if upstream is not detected.
   REMOTE="origin"
-  set +e
-  git -C "$FLUTTER_ROOT" remote get-url upstream >/dev/null 2>&1
-  if [[ $? -eq 0 ]]; then
+  if git -C "$FLUTTER_ROOT" remote get-url upstream >/dev/null 2>&1; then
```
jtmcdole, line 53: if you really want to clean this up; set GIT_WORK_TREE="$FLUTTER_ROOT" and you can remove the `-C` option from all the git commands

bin/internal/content_aware_hash.sh
```diff
@@ -50,18 +50,15 @@ if [[ "$CURRENT_BRANCH" != "main" && \
...
     REMOTE="upstream"
   fi
 
   # Try to find the merge-base with master, then main.
-  MERGEBASE=$(git -C "$FLUTTER_ROOT" merge-base HEAD "$REMOTE/master" 2>/dev/null)
+  MERGEBASE=$(git -C "$FLUTTER_ROOT" merge-base HEAD "$REMOTE/master" 2>/dev/null || true)
```
jtmcdole, line 58: neat

jtmcdole (approved):
awesome-computer-kid-computer-lgtm

bin/internal/content_aware_hash.sh
kevmoo: but we'd have to set the environment, right? We wouldn't want that to leak into other bits?

jtmcdole: You can set it up to be a local variable that isn't exported.

PR #182698, opened 2026-02-21
refactor: remove material in reorderable_list_test, scroll_notification_test, scroll_physics_test, shortcuts_test, sliver_floating_header_test, snapshot_widget_test

rkishan516:
This PR removes Material imports from reorderable_list_test, scroll_notification_test, scroll_physics_test, shortcuts_test, sliver_floating_header_test, snapshot_widget_test.

part of: https://github.com/flutter/flutter/issues/177415
depends on: https://github.com/flutter/flutter/pull/182805

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/reorderable_list_test.dart
```diff
@@ -1774,14 +1770,14 @@ void main() {
...
-                    color: items[index].isOdd ? Colors.red : Colors.green,
+                    color: items[index].isOdd ? _kRedColor : _kGreenColor,
                     child: ReorderableDragStartListener(
                       index: index,
                       child: Row(
                         crossAxisAlignment: CrossAxisAlignment.start,
                         children: <Widget>[
                           Text('item ${items[index]}'),
-                          const Icon(Icons.drag_handle),
+                          const Icon(IconData(0xe25d, fontFamily: 'MaterialIcons')),
```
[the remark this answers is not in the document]
victorsanni: +1, or similar

packages/flutter/test/widgets/reorderable_list_test.dart
```diff
@@ -1681,51 +1675,53 @@ void main() {
...
-            ),
-          ),
+                    ],
+                  ),
+                ),
+                GestureDetector(
+                  onTap: () => setState(() {
+                    showList = false;
+                  }),
+                  child: const Text('Close drawer'),
```
victorsanni: Are we still working with drawers?

packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -325,16 +323,16 @@ void main() {
     expect(scrollController.offset, 0);
     expect(notification, isNull);
 
-    final TestGesture dragScrollbarGesture = await tester.startGesture(const Offset(790, 45));
+    final TestGesture dragScrollbarGesture = await tester.startGesture(const Offset(797, 45));
```
victorsanni, line 326: Can you explain the numbers change?

packages/flutter/test/widgets/scroll_notification_test.dart
```diff
@@ -325,16 +323,16 @@ void main() {
...
     await tester.pumpAndSettle();
     expect(notification, isA<ScrollStartNotification>());
 
-    await dragScrollbarGesture.moveBy(const Offset(0, -20));
+    await dragScrollbarGesture.moveBy(const Offset(0, 10));
```
victorsanni, line 330: Why are we now moving by positive dy instead of negative as in existing code?

packages/flutter/test/widgets/shortcuts_test.dart
```diff
@@ -1215,7 +1212,19 @@ void main() {
       tester.binding.focusManager.highlightStrategy = FocusHighlightStrategy.alwaysTraditional;
       bool? value = true;
       Widget buildApp() {
-        return MaterialApp(
+        return WidgetsApp(
```
victorsanni: Why can't we use `TestWidgetsApp` here?

packages/flutter/test/widgets/shortcuts_test.dart
```diff
@@ -1229,25 +1238,37 @@ void main() {
...
-                ],
-              ),
+          home: Center(
+            child: ListView(
+              primary: true,
+              children: <Widget>[
+                StatefulBuilder(
+                  builder: (BuildContext context, StateSetter setState) {
+                    // A focusable widget that handles ActivateIntent,
+                    // replacing Checkbox for this widgets-layer test.
```
victorsanni:
```suggestion
                    // A focusable widget that handles ActivateIntent.
```

We should not specify what it is replacing, future code readers will not see a `Checkbox` here and will get confused. If more code changes happen here, it could be hard to find the PR that added the `Checkbox` in the first place. So it is just breadcrumbing to an irrelevant implementation detail, wasting developer time. 

Instead we want to guarantee that the behavior reinforced by the test is preserved. As long as that is guaranteed implementation details are free to change.

packages/flutter/test/widgets/scroll_notification_test.dart
rkishan516: If i am not wrong then Material Scrollbar positioned the thumb center around x=790 (800 - some internal padding) but RawScrollbar positions the thumb center closer to x=797 (800 - half the default thumb width, approximately)

packages/flutter/test/widgets/scroll_notification_test.dart
rkishan516: The drag direction changed from - to + because we replaced Scaffold + Scrollbar with a plain RawScrollbar. The Scaffold added extra layout layers that shifted the scrollbar track's coordinate space, causing a negative drag delta to map to a positive scroll offset. Without Scaffold, the coordinate mapping is straightforward — dragging down (+dy) scrolls down i.e. increases offset, which is the natural scroll. The magnitude also changed to half because removing Scaffold altered the viewport dimensions and thus the drag-to-scroll ratio.

packages/flutter/test/widgets/shortcuts_test.dart
rkishan516: Yup, we can. Replacing.

packages/flutter/test/widgets/reorderable_list_test.dart
rkishan516: No, we are not, I will replace with Close list or hide list.

packages/flutter/test/widgets/shortcuts_test.dart
rkishan516: Updated

victorsanni (approved):
Thanks for the work here!

PR #182702, opened 2026-02-21
refactor: remove material from widget_inspector_test, sliver_cross_axis_group_test, editable_text_show_on_screen_test, scrollable_fling_test, selection_container_test

rkishan516:
This PR removes Material imports from widget_inspector_test, sliver_cross_axis_group_test, editable_text_show_on_screen_test, scrollable_fling_test, selection_container_test.

part of: https://github.com/flutter/flutter/issues/177415

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter/test/widgets/scrollable_fling_test.dart
```diff
@@ -2,17 +2,20 @@
...
-import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
 
+import 'widgets_app_tester.dart';
+
 const TextStyle testFont = TextStyle(color: Color(0xFF00FF00));
 
 Future<void> pumpTest(WidgetTester tester, TargetPlatform platform) async {
+  debugDefaultTargetPlatformOverride = platform;
```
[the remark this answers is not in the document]
navaronbracke: We do need a teardown here, so I agree

packages/flutter/test/widgets/scrollable_fling_test.dart
```diff
@@ -92,6 +95,8 @@ void main() {
     expect(windowsResult, equals(androidResult));
     expect(linuxResult, equals(androidResult));
     expect(linuxResult, equals(androidResult));
+
+    debugDefaultTargetPlatformOverride = null;
```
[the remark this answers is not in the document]
navaronbracke: I would prefer doing this with teardowns, since then it is easier to follow

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
```diff
@@ -860,7 +863,7 @@ void main() {
   );
 
   testWidgets(
```
navaronbracke, line 865: Can we add the original test(s) in SliverAppBar tests over in Material?

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
```diff
@@ -920,20 +926,20 @@ void main() {
       await gesture.moveBy(const Offset(0.0, 10));
       await tester.pump();
 
-      // The snap animation does not go through until the gesture is released.
+      // Floating header appears as user scrolls back.
       expect(renderHeader.geometry!.paintExtent, equals(10));
       expect((renderHeader.parentData! as SliverPhysicalParentData).paintOffset.dy, equals(0.0));
 
-      // Once it is released, the header's paint extent becomes the maximum and the group sets an offset of -50.0.
+      // Without snap, the header stays at the current extent after gesture release.
```
navaronbracke: There is no "snap" parameter for the widget? Or am I reading this wrong?

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
```diff
@@ -970,11 +974,11 @@ void main() {
       expect(renderHeader.geometry!.paintExtent, equals(30.0));
       expect((renderHeader.parentData! as SliverPhysicalParentData).paintOffset.dy, equals(-20.0));
 
-      // Once we lift the gesture up, the animation should finish.
+      // Without snap, the header stays at the current extent after gesture release.
```
navaronbracke: Same here

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -52,7 +53,7 @@ class ClockDemo extends StatelessWidget {
   Widget makeClock(String label, int utcOffset) {
     return Stack(
       children: <Widget>[
-        const Icon(Icons.watch),
+        const Icon(IconData(0xe6ce, fontFamily: 'MaterialIcons')),
```
navaronbracke, line 56: Is there a constant from the widget inspector that we can leverage, instead of the magic number?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -169,7 +170,7 @@ class RenderRepaintBoundaryWithDebugPaint extends RenderRepaintBoundary {
       final paint = Paint()
         ..style = PaintingStyle.stroke
         ..strokeWidth = 1.0
-        ..color = Colors.red;
+        ..color = const Color(0xFFF44336);
```
navaronbracke: Is this specific for a golden test? Why not `Color(0xFFFF0000)`

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -429,8 +430,11 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
         required GlobalKey key,
       }) {
         exitWidgetSelectionButtonKey = key;
-        return Material(
-          child: ElevatedButton(onPressed: onPressed, key: key, child: null),
+        return GestureDetector(
```
navaronbracke: Use TestButton?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -481,25 +485,23 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
-                    onPressed: () {
-                      log.add('bottom');
-                    },
-                    child: const Text('BOTTOM'),
-                  ),
-                ],
-              ),
+            child: ListView(
+              children: <Widget>[
+                GestureDetector(
```
navaronbracke: Same here for TestButton

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -609,8 +611,11 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
         required GlobalKey key,
       }) {
         exitWidgetSelectionButtonKey = key;
-        return Material(
-          child: ElevatedButton(onPressed: onPressed, key: key, child: null),
+        return GestureDetector(
```
navaronbracke: Here too

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -860,8 +865,11 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
           required String semanticsLabel,
           required GlobalKey key,
         }) {
-          return Material(
-            child: ElevatedButton(onPressed: onPressed, key: key, child: null),
+          return GestureDetector(
```
navaronbracke: TestButton?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -4343,7 +4343,7 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
         _CreationLocation location = knownLocations[id]!;
         expect(location.file, equals(file));
         // ClockText widget.
-        expect(location.line, equals(57));
```
navaronbracke, line 4346: Why is there a tiny difference here?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5457,21 +5461,14 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
         expect(crossAxisAlignment, equals('start'));
       });
 
-      // TODO(justinmc): Don't test Cupertino interactions in Widgets tests.
-      // https://github.com/flutter/flutter/issues/177028
       testWidgets('ext.flutter.inspector.getLayoutExplorerNode does not throw StackOverflowError', (
```
navaronbracke, line 5465: This should be moved? The original issue mentioned CupertinoTabScaffold specifically?

cc @justinmc

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5899,20 +5897,25 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
         // ignore: avoid_dynamic_calls
         final detailedChildren = json.decode(details)['children'] as List<Object?>;
 
-        final appBars = <Map<String, Object?>>[];
+        final texts = <Map<String, Object?>>[];
         void visitChildren(List<Object?> children) {
           for (final Map<String, Object?> child in children.cast<Map<String, Object?>>()) {
-            if (child['description'] == 'AppBar') {
-              appBars.add(child);
+            if (child['description'] != null &&
```
navaronbracke: Use an If-case to clean this up? It feels odd to read

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5899,20 +5897,25 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
+        final childrenOfDirectionality = directionality['children']! as List<Object?>;
+        final column = childrenOfDirectionality.first! as Map<String, Object?>;
+        expect(column['description'], 'Column');
+        final childrenOfColumn = column['children']! as List<Object?>;
+        final padding = childrenOfColumn.first! as Map<String, Object?>;
+        expect(padding['description'], 'Padding');
+        final objectId = padding['valueId']! as String;
         final String details = service.getDetailsSubtree(objectId, 'foo2');
         // ignore: avoid_dynamic_calls
         final detailedChildren = json.decode(details)['children'] as List<Object?>;
```
navaronbracke: Can we use an if-case to get rid of the // ignore ?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5899,20 +5897,25 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
         final String summary = service.getRootWidgetSummaryTree('foo1');
         // ignore: avoid_dynamic_calls
         final childrenOfRoot = json.decode(summary)['children'] as List<Object?>;
```
navaronbracke: If case to avoid the // ignore ?

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5487,35 +5484,32 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
-                child: Row(
-                  children: <Widget>[
-                    Flexible(
-                      child: ColoredBox(
-                        color: Colors.green,
-                        child: Tooltip(
-                          message: 'a',
-                          child: ElevatedButton(onPressed: null, child: Text('a')),
-                        ),
+          const Widget widget = Directionality(
```
navaronbracke: Nit: inline this var into the pumpWidget call?

navaronbracke (changes requested):

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
rkishan516: I made it wrong sentence, actually I wanted to say, since there is no snap parameter that's why no change after gesture release.

packages/flutter/test/widgets/scrollable_fling_test.dart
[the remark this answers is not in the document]
rkishan516: Sure will do

packages/flutter/test/widgets/widget_inspector_test.dart
rkishan516: Oh, I think I missed this. We should use single channel color.

packages/flutter/test/widgets/widget_inspector_test.dart
rkishan516: I don't think there is. But will check.

packages/flutter/test/widgets/widget_inspector_test.dart
rkishan516: Removal of import of material and cupertino has caused this.

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
rkishan516: I was thinking to move SliverAppBar to sliver.dart, because thats the only sliver stuck inside material.

packages/flutter/test/widgets/scrollable_fling_test.dart
```diff
@@ -30,6 +33,8 @@ const double dragOffset = 213.82;
 
 void main() {
   testWidgets('Flings on different platforms', (WidgetTester tester) async {
+    addTearDown(() => debugDefaultTargetPlatformOverride = null);
```
justinmc: I would put this teardown inside of pumpTest to avoid some future test calling pumpTest and forgetting to include the teardown.

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
justinmc: Ah that could be a good idea, it looks like SliverAppBar does not depend on AppBar or anything else Material?  I wonder what @Piinks thinks.

For another PR though.

packages/flutter/test/widgets/widget_inspector_test.dart
justinmc: I'm not sure exactly why that would be off the top of my head.

packages/flutter/test/widgets/widget_inspector_test.dart
justinmc: Good catch, yeah @rkishan516 can you copy the original test to Cupertino? Just to make sure we have that specific bug covered. Up to you if you also want to keep your modified test here too.

justinmc (commented):
Thanks @rkishan516! Heads up that there are test failures.

packages/flutter/test/widgets/scrollable_fling_test.dart
rkishan516: Sure, will move.

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
rkishan516: Yup, i will create another PR.

packages/flutter/test/widgets/scrollable_fling_test.dart
```diff
@@ -92,6 +96,8 @@ void main() {
     expect(windowsResult, equals(androidResult));
     expect(linuxResult, equals(androidResult));
     expect(linuxResult, equals(androidResult));
+
+    debugDefaultTargetPlatformOverride = null;
```
navaronbracke, line 100: This can be removed? We have a teardown now

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5897,22 +5897,26 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
+        final column = childrenOfDirectionality.first! as Map<String, Object?>;
+        expect(column['description'], 'Column');
+        final childrenOfColumn = column['children']! as List<Object?>;
+        final padding = childrenOfColumn.first! as Map<String, Object?>;
+        expect(padding['description'], 'Padding');
+        final objectId = padding['valueId']! as String;
         final String details = service.getDetailsSubtree(objectId, 'foo2');
-        // ignore: avoid_dynamic_calls
-        final detailedChildren = json.decode(details)['children'] as List<Object?>;
+        final detailedChildren =
```
navaronbracke, line 5912: Same here

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5897,22 +5897,26 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
 
         final String summary = service.getRootWidgetSummaryTree('foo1');
-        // ignore: avoid_dynamic_calls
-        final childrenOfRoot = json.decode(summary)['children'] as List<Object?>;
-        final childrenOfMaterialApp =
-            (childrenOfRoot.first! as Map<String, Object?>)['children']! as List<Object?>;
-        final scaffold = childrenOfMaterialApp.first! as Map<String, Object?>;
-        expect(scaffold['description'], 'Scaffold');
-        final objectId = scaffold['valueId']! as String;
+        final childrenOfRoot =
```
navaronbracke, line 5900: This can probably be cleaned up using a switch expression? The double forced cast makes it look a little worse now

packages/flutter/test/cupertino/tab_scaffold_test.dart
```diff
@@ -1405,6 +1410,56 @@ void main() {
   });
 }
 
+class _TabScaffoldWidgetInspectorService extends TestWidgetInspectorService {
```
navaronbracke, line 1413: This is really odd for several reasons?

1. TestWidgetInspectorService is public, you don't need to subclass it to access `resetAllState()` ? That method is visibleForTests, so I think you can do `final service = TestWidgetInspectorService();` ?
2. Can you put the set up / tear down directly inside the test `ext.flutter.inspector.getLayoutExplorerNode does not throw StackOverflowError`
3. `service.disposeAllGroups();` can be put in the teardown? (or if resetAllState() call it, then you can omit it?)

navaronbracke (changes requested):

packages/flutter/test/widgets/scrollable_fling_test.dart
rkishan516: I think this is required because between variants teardown doesn't run.

packages/flutter/test/widgets/widget_inspector_test.dart
rkishan516: I feel this pattern is being used at multiple places and converting to switch expression adds more verbosity but not clarity. WDYT?

packages/flutter/test/cupertino/tab_scaffold_test.dart
rkishan516: If i am not wrong

While resetAllState() is indeed @visibleForTesting and doesn't require a subclass, the subclass is needed for other reasons: WidgetInspectorService.instance (setter), setSelection, and toId are all @protected, so they can only be accessed from within a subclass of WidgetInspectorService.

For the same very reason, WidgetInspectorService.instance = service must be called from within the subclass scope (the setter is @protected), so the runTests() static method pattern is necessary.

You're right on this one. resetAllState() already calls disposeAllGroups() as its first line, so the explicit call at the end of the test was redundant. Removing it.

packages/flutter/test/widgets/sliver_cross_axis_group_test.dart
rkishan516: Have created https://github.com/flutter/flutter/pull/182925, SliverAppBar actually depends quite on material, So made new version as RawSliverAppBar.

packages/flutter/test/cupertino/tab_scaffold_test.dart
```diff
@@ -1405,6 +1410,54 @@ void main() {
...
+  static void runTests() {
+    final service = _TabScaffoldWidgetInspectorService();
+    WidgetInspectorService.instance = service;
+
+    tearDown(() {
+      service.resetAllState();
+    });
```
justinmc, line 1423: Should you also reset WidgetInspectorService.instance to whatever it was before?

packages/flutter/test/widgets/scrollable_fling_test.dart
```diff
@@ -2,17 +2,21 @@
...
+import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
 
+import 'widgets_app_tester.dart';
+
 const TextStyle testFont = TextStyle(color: Color(0xFF00FF00));
 
 Future<void> pumpTest(WidgetTester tester, TargetPlatform platform) async {
+  debugDefaultTargetPlatformOverride = platform;
+  addTearDown(() => debugDefaultTargetPlatformOverride = null);
```
justinmc, line 16: I was going to suggest just using a `variant` instead of debugDefaultTargetPlatformOverride, but since I see the test compares the results between different platforms, I think keeping it as-is is the best bet that I can think of.

packages/flutter/test/widgets/widget_inspector_test.dart
justinmc: Looks like this is done 👍

packages/flutter/test/widgets/widget_inspector_test.dart
justinmc: Either way is good for me.

packages/flutter/test/cupertino/tab_scaffold_test.dart
```diff
@@ -1405,6 +1410,54 @@ void main() {
   });
 }
 
+class _TabScaffoldWidgetInspectorService extends TestWidgetInspectorService {
+  static void runTests() {
```
justinmc, line 1415: Can you include a comment similar to the one in the original widget_inspector_test.dart version? "These tests need access to protected members of WidgetInspectorService."

justinmc (approved):
LGTM 👍

packages/flutter/test/cupertino/tab_scaffold_test.dart
rkishan516: Updated by saving previous instance.

packages/flutter/test/cupertino/tab_scaffold_test.dart
rkishan516: Updated.

packages/flutter/test/widgets/widget_inspector_test.dart
```diff
@@ -5487,21 +5485,20 @@ class _TestWidgetInspectorService extends TestWidgetInspectorService {
...
                 child: Row(
                   children: <Widget>[
                     Flexible(
                       child: ColoredBox(
-                        color: Colors.green,
-                        child: Tooltip(
-                          message: 'a',
-                          child: ElevatedButton(onPressed: null, child: Text('a')),
+                        color: Color(0xFF00FF00),
+                        child: SizedBox(
```
navaronbracke, line 5500: Why do we have explicit zero padding here?

navaronbracke (approved):
LGTM with nit

packages/flutter/test/widgets/widget_inspector_test.dart
rkishan516: Its required for `getLayoutExplorerNode` to have Padding in tree.

packages/flutter/test/widgets/button_tester.dart
```diff
@@ -12,13 +12,15 @@ class TestButton extends StatelessWidget {
     this.focusNode,
     this.autofocus = false,
     this.onPressed,
+    this.behavior,
```
victorsanni, line 15: do we need to test this separately, or is it fine that it is already used in some of these tests in this PR?

victorsanni (approved):

packages/flutter/test/widgets/button_tester.dart
rkishan516: Yes, it is being tested.

justinmc (approved):
There was 1 failing golden and I've approved it. It was just a very slight color change.

Issue #69676, opened 2020-11-03
Persistent bottom sheet does not respect SafeArea


amitkot:
When showing a persistent bottom sheet `SafeArea` is ignored and does not provide bottom padding:

<img width="511" alt="Screen Shot 2020-11-03 at 16 18 04" src="https://user-images.githubusercontent.com/654041/97996255-34324b80-1df0-11eb-88c6-23f9417c0e35.png">

Code:
```dart
import 'package:flutter/material.dart';

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) => MaterialApp(home: MyHomePage());
}

class MyHomePage extends StatelessWidget {
  MyHomePage({Key key}) : super(key: key);
  @override
  Widget build(BuildContext context) => Scaffold(
        floatingActionButton: Builder(
          builder: (context) => FloatingActionButton(
            onPressed: () {
              showBottomSheet(
                  context: context,
                  builder: (context) => SafeArea(
                        child: Container(
                          height: 100,
                          width: 200,
                          color: Colors.red,
                        ),
                      ));
            },
          ),
        ),
      );
}
```

```
[✓] Flutter (Channel stable, 1.22.2, on Mac OS X 10.14.6 18G6032, locale en-IL)
    • Flutter version 1.22.2 at /Users/amit/fvm/versions/stable
    • Framework revision 84f3d28555 (3 weeks ago), 2020-10-15 16:26:19 -0700
    • Engine revision b8752bbfff
    • Dart version 2.10.2
```

pedromassangocode:
Reproducible on latest master

<details><summary>flutter doctor -v</summary>

```
[✓] Flutter (Channel master, 1.24.0-8.0.pre.97, on Mac OS X 10.15.7 19H2 darwin-x64, locale en)
    • Flutter version 1.24.0-8.0.pre.97 at /Users/pedromassango/dev/SDKs/flutter_master
    • Framework revision 4c24eb10b8 (72 minutes ago), 2020-11-05 11:18:00 +0100
    • Engine revision e66a720137
    • Dart version 2.12.0 (build 2.12.0-21.0.dev)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.1)
    • Android SDK at /Users/pedromassango/Library/Android/sdk
    • Platform android-30, build-tools 30.0.1
[...]

PR #182727, opened 2026-02-22
Enable SafeArea on Scaffold bottomSheet

iazel:
As explained in https://github.com/flutter/flutter/issues/69676 , for a long time, it hasn't been possible to use `SafeArea` with `BottomSheet` displayed by `Scaffold`.

The issue here is that we remove the bottom padding from `MediaQuery`, which is what `SafeArea` widget uses to understand when to apply padding and of how much.

I can't think of a good reason to disable such option for bottom sheets.

As a note, this _doesn't_ add padding by default, but rather users must still wrap their widgets in a `SafeArea`, hence this should be a backward compatible change.


victorsanni (commented):
Hi, this PR will need a test to prevent regression.

Piinks:

Issue #135221, opened 2023-09-21
Predictive TextField on iOS 17


nikb7:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)
- [X] I have read the [guide to filing a bug](https://flutter.dev/docs/resources/bug-reports)

### Use case

Predictive text - makes it very easy for user to types long messages on app. It utilises the device's inbuilt AI to predict the text that user will type.

### Proposal

Enable the new Predictive text field introduced in iOS 17
Reference - https://support.apple.com/en-in/guide/iphone/iphd4ea90231/ios#:~:text=As%20you%20type%2C%20predictions%20that,keep%20typing%20to%20reject%20it.&text=%2C%20then%20tap%20the%20word%20you%20were%20in%20the%20process%20of%20typing.

hellohuanlin:
Seems to be [this API](https://developer.apple.com/documentation/uikit/uitextinputtraits/4240862-inlinepredictiontype?language=objc)

moffatman:
I don't think `UITextInput` exposes this....

LongCatIsLooong:
As moffatman pointed out there doesn't seem to be additional API for this feature. I think the feature is provided by iOS keyboard implementations so if the user uses the on screen keyboard they'll get this feature automatically.

nikb7:
There are two types of predictive texts
1. Autocomplete suggestions in above keyboards
2. Inlined suggestions within the textfield, attaching a screenshot of the same as well
![image](https://github.com/flutter/flutter/assets/29572153/55381f77-42fe-4a39-8a41-a160b10d7f98)

The inlined suggestion is enabled in the TextField in iOS native apps, but this doesn't seem to work in the flutter app.

stuartmorgan:
> I don't think `UITextInput` exposes this....

We should investigate further; this would definitely be unfortunate.

nikb7:
Hi Team,
Any update for timeline for this feature?

LongCatIsLooong:
This does have public API, it is the one @hellohuanlin mentioned. Our current `UITextInput` does not implement [setAttributedMarkedText(NSAttributedString?, selectedRange: NSRange)](https://developer.apple.com/documentation/uikit/uitextinput/setattributedmarkedtext(_:selectedrange:)) so the prediction is not showing.

PR #182728, opened 2026-02-22
Enable inline text prediction on iOS (issue #135221)

nikb7:
**Description**

This PR adds support for **inline predictive text** on iOS 17+ (the gray suggestion that appears after the cursor as you type). It does two things:

1. **Enable/disable inline prediction** – Apps can turn inline prediction on or off per field via a new `enableInlinePrediction` parameter on `TextField` and `CupertinoTextField` (default remains `true`). The setting is sent through `TextInputConfiguration` to the engine, which sets `UITextInlinePredictionType` on the iOS text input view so the system shows or hides inline suggestions.

2. **Style the prediction/composing region** – A new optional `composingStyle` parameter lets apps control how the composing range (IME and inline prediction) is drawn. When set (e.g. gray with no underline), the suggestion can match the native iOS look instead of the default underline.

**Why:** On iOS 17+, the system keyboard can show inline predictions; Flutter had no way to enable/disable this or style the suggestion text. This change adds that control and styling so Flutter apps can match platform behavior and design.

**Technical notes:**  
- Framework: `TextInputConfiguration.enableInlinePrediction`, `EditableText`/`TextField`/`CupertinoTextField` pass-through, and `TextEditingController.buildTextSpan(composingStyle)` for the composing region.  
- Engine (iOS only): read `enableInlinePrediction` in `configureWithDictionary:`, set `FlutterTextInputView.inlinePredictionType`, and implement `setAttributedMarkedText:selectedRange:` so the system can deliver inline prediction. Semantics objects updated for the same protocol.  
- Other platforms are unchanged: they ignore the new config; `composingStyle` only affects how the composing range is drawn.

Consider adding before/after screenshots: one with inline prediction off or default underline, one with `enableInlinePrediction: true` and a custom `composingStyle` (e.g. gray, no underline) to show the new behavior.

---

**Fixes** #135221

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1100,6 +1104,13 @@ - (void)configureWithDictionary:(NSDictionary*)configuration {
   // The input field needs to be visible for the system autofill
   // to find it.
   self.isVisibleToAutofill = autofill || _secureTextEntry;
+
+  if (@available(iOS 17.0, *)) {
+    NSNumber* enableInlinePrediction = configuration[kEnableInlinePrediction];
```
LongCatIsLooong: Is it supposed to be NSValue here?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1100,6 +1104,13 @@ - (void)configureWithDictionary:(NSDictionary*)configuration {
...
+    BOOL enabled = enableInlinePrediction == nil || [enableInlinePrediction boolValue];
+    self.inlinePredictionType =
+        enabled ? UITextInlinePredictionTypeYes : UITextInlinePredictionTypeNo;
```
LongCatIsLooong: What about default?

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -300,6 +300,7 @@ class TextEditingController extends ValueNotifier<TextEditingValue> {
     required BuildContext context,
     TextStyle? style,
     required bool withComposing,
+    TextStyle? composingStyle,
```
LongCatIsLooong, line 303: This will likely be a breaking change (based on my past attempts to add new parameters to this method).

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2067,6 +2069,19 @@ class EditableText extends StatefulWidget {
   /// {@macro flutter.services.TextInputConfiguration.hintLocales}
   final List<Locale>? hintLocales;
 
+  /// Whether to enable inline predictive text (e.g. iOS 17+ inline suggestions).
+  ///
+  /// Defaults to true. Only affects platforms that support it.
```
LongCatIsLooong: Document which platforms support it (or something like "Whether to enable inline predictive text on iOS").

packages/flutter/lib/src/widgets/editable_text.dart
```diff
@@ -2067,6 +2069,19 @@ class EditableText extends StatefulWidget {
...
+  final bool enableInlinePrediction;
+
+  /// Optional style for the composing (and inline prediction) region.
+  ///
+  /// When set, this style is used for the segment of text in the composing
+  /// range (e.g. IME composition and inline predictive text). When null,
+  /// the default is used (underlined). Applies to both IME composition and
```
LongCatIsLooong: nit: underlined -> [TextDecoration.underline]

engine/src/flutter/shell/platform/darwin/ios/framework/Source/vsync_waiter_ios.mm
```diff
@@ -120,6 +120,15 @@ - (void)onDisplayLink:(CADisplayLink*)link {
   CFTimeInterval duration = link.targetTimestamp - link.timestamp;
   fml::TimePoint frame_target_time = frame_start_time + fml::TimeDelta::FromSecondsF(duration);
 
+  // Clamp frame_start_time to the past so FireCallback's DCHECK(Now() >= frame_start_time) holds.
+  // CADisplayLink timestamps can occasionally be in the future (e.g. first frame, ProMotion).
+  fml::TimePoint now = fml::TimePoint::Now();
```
LongCatIsLooong: This doesn't seem to have to do with text predication?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
nikb7: the config value is a boolean from the channel, hence NSNumber. Tested this flow locally as well to validate this behaviour

packages/flutter/lib/src/widgets/editable_text.dart
nikb7: updated

packages/flutter/lib/src/widgets/editable_text.dart
nikb7: updated

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
nikb7: platform default behaviour will be default - and null/no value will be considered as default

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1595,6 +1613,14 @@ - (void)setMarkedText:(NSString*)markedText selectedRange:(NSRange)markedSelecte
   }
 }
 
+// iOS 17+ inline predictive text uses setAttributedMarkedText:selectedRange:.
+// Forward to the existing setMarkedText implementation by extracting the plain string.
```
hellohuanlin, line 1617: I'm surprised that `setMarkedText` works out of box. Is the inline predictive text eventually rendered by the framework or iOS system?

nikb7: is it eventually rendered by flutter framework itself - as composing text. And we can override the style attributes of the text as well

nikb7:
I will be adding these attributes to ThemeData class as well - so it can easily controlled centrally.
Also it will better to keep default value for inline as false - so there is no change in behaviour of textfield for existing users.

packages/flutter/lib/src/cupertino/text_field.dart
```diff
@@ -836,6 +840,15 @@ class CupertinoTextField extends StatefulWidget {
   /// {@end-tool}
   final TextMagnifierConfiguration? magnifierConfiguration;
 
+  /// {@macro flutter.services.TextInputConfiguration.enableInlinePrediction}
+  final bool? enableInlinePrediction;
+
+  /// Optional style for the composing (and inline prediction) region.
+  ///
+  /// When set, applied to the composing range (IME and inline predictive text).
+  /// When null, the default is used ([TextDecoration.underline]).
```
dkwingsmt: How come you can use one style to represent both IME and predictive text? AFAI understand, IME text uses regular color with underline, while predictive text uses grey color and no underline.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -702,6 +703,17 @@ class TextInputConfiguration {
...
   final List<Locale>? hintLocales;
 
+  /// {@template flutter.services.TextInputConfiguration.enableInlinePrediction}
+  /// Whether to enable inline predictive text.
+  ///
+  /// This is an iOS-specific feature (iOS 17 and later). When true, inline
+  /// prediction may be shown; when false, it is disabled; when null, on iOS
+  /// the platform default is used and other platforms ignore this.
+  ///
+  /// Defaults to null (use platform default).
```
dkwingsmt:
```suggestion
  /// This feature is specific to iOS 17 and later. It has no effect on other platforms.
  ///
  /// By default, this property is null, meaning the platform setting is used.
  /// Setting this flag overrides the platform setting.
  /// When true, inline prediction is shown; when false, it is disabled.
```

packages/flutter/lib/src/widgets/editable_text.dart
dkwingsmt: If that's so we should solve this once and for all: Create a new data class `class BuildTextSpanInfo` that contains all these parameters, and use this class as the parameter instead. This way adding new parameters no longer breaks.

dkwingsmt (commented):
Generally LGTM with minor comments.

packages/flutter/lib/src/cupertino/text_field.dart
LongCatIsLooong: The iOS text input implementation does not distinguish these two. UX-wise I don't think it's too big of a deal since to users they both represent uncommitted changes suggested by iOS.

packages/flutter/lib/src/widgets/editable_text.dart
LongCatIsLooong: Ah it looks like you're trying to expose the configuration on TextField/CupertinoTextField. The widget-level API makes sense but I'm still worried this change (on TextEditingController) will be breaking.

Since this is a separate / potentially breaking feature from inline prediction (and people can subclass TextEditingController to achieve the same thing, and interestingly this is going to be a breaking change for them),  I feel we should hold off adding this argument if it's indeed breaking.

packages/flutter/lib/src/material/input_decorator.dart
```diff
@@ -4319,6 +4319,8 @@ class InputDecorationTheme extends InheritedTheme with Diagnosticable {
     Color? suffixIconColor,
     BoxConstraints? suffixIconConstraints,
     TextStyle? counterStyle,
+    TextStyle? composingStyle,
```
QuncCccccc, line 4322: New properties should be added in `InputDecorationThemeData` only. Check https://github.com/flutter/flutter/pull/168981 for more context:) Eventually the xxxTheme class should only contain `data` and `child` fields and any other properties should only exist in xxxThemeData:)

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -702,6 +703,17 @@ class TextInputConfiguration {
   /// {@endtemplate}
   final List<Locale>? hintLocales;
 
+  /// {@template flutter.services.TextInputConfiguration.enableInlinePrediction}
+  /// Whether to enable inline predictive text.
+  ///
+  /// This feature is specific to iOS 17 and later. It has no effect on other platforms.
+  ///
+  /// By default, this property is null, meaning the platform setting is used.
```
loic-sharma: Maybe consider referring to `UITextInlinePredictionTypeDefault` to clarify what this means.

```suggestion
  /// By default, this property is null, which corresponds to
  /// `UITextInlinePredictionTypeDefault` on iOS.
```

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -702,6 +703,17 @@ class TextInputConfiguration {
...
+  /// Setting this flag overrides the platform setting.
+  /// When true, inline prediction is shown; when false, it is disabled.
```
loic-sharma: Maybe to be super duper extra clear this only affects iOS, consider:

```suggestion
  /// Setting this flag overrides the platform setting on iOS:
  /// when true, inline prediction is shown; when false, it is disabled.
```

packages/flutter/lib/src/widgets/editable_text.dart
loic-sharma: I agree with @LongCatIsLooong's comment. Also I'm concerned that:

* Inline predictive text will have incorrect styling but will be on by default on iOS 17+.
* We're coupling the styling for inline predictive text and the composition string, even though these are styled differently on iOS.

I'd suggest that in this PR:
1. We remove the APIs to customize the `composingStyle`.
2. We make inline text prediction off by default on all platforms. iOS apps with `enablePredictiveText: null` have inline predictive text disabled. iOS apps must use `enablePredictiveText: true` to enable it.

Then in one or more subsequent PRs:
1. Add APIs to customize the `composingStyle` and/or the `inlinePredictionStyle`
2. Update `CupertinoTextField` to customize these styles to match iOS's styling for inline predictive text
3. Make `enablePredictiveText: null` use the platform default for inline text prediction on iOS 17+. 

What do y'all think?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1595,6 +1613,14 @@ - (void)setMarkedText:(NSString*)markedText selectedRange:(NSRange)markedSelecte
...
+- (void)setAttributedMarkedText:(NSAttributedString*)attributedString
+                  selectedRange:(NSRange)selectedRange API_AVAILABLE(ios(17.0)) {
+  NSString* markedText = attributedString ? [attributedString string] : @"";
+  [self setMarkedText:markedText selectedRange:selectedRange];
```
loic-sharma, line 1621: From my understanding, this works because currently the iOS text input plugin considers any marked text as the composition string:

https://github.com/flutter/flutter/blob/b07acf9bb92f104596639b433bdc0fc3957a1b5c/engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm#L2212-L2227

However, this forces the framework to use the same style for inline predictive text and the composition string. That is not desirable since these strings have different styling on iOS.

@hellohuanlin @LongCatIsLooong What do y'all think of introducing new values to the text input editing state to capture the inline predictive text's base / extent? Something like `inlinePredictionBase` and `inlinePredictionExtent`. This would let track the composition string and the inline predictive text separately, which would let us style these separately.

hellohuanlin: Yes, reusing the composition text feels a bit too magical for me (https://github.com/flutter/flutter/pull/182728#discussion_r2850292886)

I prefer having a more explicit setup as you described, since they are semantically different concepts.

packages/flutter/lib/src/widgets/editable_text.dart
nikb7: makes sense @loic-sharma , I will split this PR and share

loic-sharma:
Thanks! I'll convert this PR to a draft since it contains styling changes that aren't ready to be reviewed until after [this](https://github.com/flutter/flutter/pull/183650) lands.

Please feel free to mark this as ready for review when ready!

Piinks:

nikb7:
I am closing this PR. I had created another PR for enabling inline prediction and that has been merged - https://github.com/flutter/flutter/pull/183650

PR #182740, opened 2026-02-23
Move Linux FFI window code into own module

robert-ancell:
To reduce the size of the window module

### DIFF
--- a/packages/flutter/lib/src/widgets/_window_linux.dart
+++ b/packages/flutter/lib/src/widgets/_window_linux.dart
@@ -14,8 +14,6 @@
 //
 // See: https://github.com/flutter/flutter/issues/30701.
 
-import 'dart:convert';
-import 'dart:ffi' as ffi;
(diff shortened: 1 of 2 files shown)


stuartmorgan-g:

mattkae (changes requested):
We should do this, but like I mentioned offline, we should add a bunch of documentation to `_window_linux_ffi.dart`, similar to how we do it in `_window.dart`

loic-sharma:
Hi @robert-ancell, we discussed this a bit more internally. So generally the team's preference would be to keep everything in a single file until it is truly untenable. We have several Dart files that are >5k lines long.

For extreme cases, we'd be OK breaking the "Avoid using part of" rule. However, @dkwingsmt suggested using a directory structure to make the hierarchy between the part file clear. For example:

```
├── foo_parts/
  └── foo_bar.dart
└── foo.dart
```

PR #182753, opened 2026-02-23
Add ability to access plugins from other plugins

### LINKED ISSUE #121527: Allow plugins to access published plugin values via the plugin registrar/binding
## Use case

When developing a plugin, or writing other native code, it is sometimes necessary to access the native implementation of another plugin. A good example of this is `webview_flutter`, which has `WebViewFlutterAndroidExternalApi` (Android) and `FWFWebViewFlutterWKWebViewExternalAPI` (iOS) classed for accessing the platform WebViews associated with a given handle. Both use either the `FlutterPluginRegistry` or the `FlutterEngine`, neither of which are easily accessible from a Plugin.

## Proposal

In my mind, the simplest solution that will have the smallest impact is to add properties onto the objects that are already passed to plugins during initialization.

iOS - Add a method to `FlutterPluginRegistrar` to allow upward access to the `FlutterPluginRegistry`

Android - Add a method to `FlutterPlugin.FlutterPluginBinding` to access the `FlutterEngine`

fuzzybinary:
Plugins occasionally need to communicate with other native plugins.

On iOS, plugins are able to publish values that should be accessible to other plugins. The easiest way to allow to these published value is to add an accessor in FlutterPluginRegistrar, similar to the one that exists in FlutterPluginRegistry (valuePublishedByPlugin).

On Android, we can ask for the plugin class directly by added and accessor to `FlutterPluginBinding` called `getPlugin`, which delegates to the engine's `FlutterPlugins`.

Fixes #121527

## Pre-launch Checklist

- [x] I signed the [CLA].


hellohuanlin (commented):
You added android tests, but not iOS tests

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/plugins/FlutterPlugin.java
```diff
@@ -163,6 +163,11 @@ public FlutterAssets getFlutterAssets() {
       return flutterAssets;
     }
 
+    @Nullable
+    public FlutterPlugin getPlugin(@NonNull Class<? extends FlutterPlugin> pluginClass) {
```
hellohuanlin, line 175: missing doc comment

hellohuanlin (dismissed):
ios part looks good.

engine/src/flutter/shell/platform/android/io/flutter/embedding/engine/plugins/FlutterPlugin.java
reidbaker: +1

reidbaker (dismissed):
Approve % public doc. 

@mboetger I think this pr is a good idea are there any issues it causes us migrating to the embedder api?

vashworth:
From triage: @hellohuanlin Can you make a g3fix?

gmackall (approved):

reidbaker (approved):

PR #182783, opened 2026-02-23
Refactor impellerc_main to better organize some of its logic

b-luk:
No-op change to better organize some of the logic in impellerc_main. This makes some of the followup work for https://github.com/flutter/flutter/issues/182400 easier to do.

- Instead of creating each compiler inside the `OutputIPLR` function, create them outside of `OutputIPLR` and pass the list of compilers into `OutputIPLR`. This way, the created compilers can be re-used for the alternate `OutputSLFile` step and for the subsequent `OutputSPIRV`, `OutputReflectionData`, and `OutputDepfile` steps.
- Moves the logic to output the SPIRV file into a `OutputSPIRV` helper function instead of having it inlined in `Main`.
- Move/fix some of the existing comments about the 4 steps performed in `Main`.

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -32,41 +34,25 @@ static Reflector::Options CreateReflectorOptions(const SourceOptions& options,
...
-    auto stage_data = reflector->GetRuntimeStageShaderData();
-    if (!stage_data) {
-      std::cerr << "Runtime stage information was nil." << std::endl;
-      return false;
-    }
-
-    stages.AddShader(stage_data);
+  for (const auto& compiler : compilers) {
+    stages.AddShader(compiler->GetReflector()->GetRuntimeStageShaderData());
   }
```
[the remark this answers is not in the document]
b-luk: Done

engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -32,40 +32,29 @@ static Reflector::Options CreateReflectorOptions(const SourceOptions& options,
...
-
-    auto reflector = compiler.GetReflector();
-    if (reflector == nullptr) {
-      std::cerr << "Could not create reflector." << std::endl;
-      return false;
-    }
-
-    auto stage_data = reflector->GetRuntimeStageShaderData();
+  for (const auto& compiler : compilers) {
+    auto stage_data = compiler->GetReflector()->GetRuntimeStageShaderData();
```
gaaclarke: please remove type deduction, it's important here since you may be accidentally introducing a copy.

https://google.github.io/styleguide/cppguide.html#Type_deduction

engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -32,40 +32,29 @@ static Reflector::Options CreateReflectorOptions(const SourceOptions& options,
   return reflector_options;
 }
 
-static bool OutputIPLR(
-    const Switches& switches,
-    const std::shared_ptr<fml::Mapping>& source_file_mapping) {
+static std::shared_ptr<Compiler> CreateCompiler(
```
gaaclarke: nit: i know it's done elsewhere in this file but static functions is kind of a c way or doing things, not a c++.  it should be anonymous namespaces.

engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -204,51 +197,73 @@ bool Main(const fml::CommandLine& command_line) {
     return false;
   }
 
-  if (switches.iplr && !OutputIPLR(switches, source_file_mapping)) {
-    return false;
+  std::vector<std::shared_ptr<Compiler>> compilers;
```
gaaclarke, line 204: nit: put a .reserve() call in here since we know the size.

engine/src/flutter/impeller/compiler/impellerc_main.cc
```diff
@@ -204,51 +197,73 @@ bool Main(const fml::CommandLine& command_line) {
...
+  for (const auto& platform : switches.PlatformsToCompile()) {
+    auto compiler = CreateCompiler(platform, source_file_mapping, switches);
```
gaaclarke: remove `auto`

b-luk: Done

engine/src/flutter/impeller/compiler/impellerc_main.cc
b-luk: Done

engine/src/flutter/impeller/compiler/impellerc_main.cc
b-luk: Done. Moved all the static functions into an anonymous namespace.

engine/src/flutter/impeller/compiler/impellerc_main.cc
b-luk: I changed it to be `std::shared_ptr<RuntimeStageData::Shader>`. But can you clarify what you mean when you say auto can accidentally introduce a copy?

My understanding is that auto would be `std::shared_ptr<RuntimeStageData::Shader>` here, right? So I think it creates a copy of the `shared_ptr`. But that is fine because the ptr is lightweight to copy, and the underlying `RuntimeStageData::Shader` is not copied. Is that correct? Replacing the auto with `std::shared_ptr<RuntimeStageData::Shader>` doesn't change anything, does it? Or am I misunderstanding, and I should I be doing something else?

gaaclarke: `std::shared_ptr` in aggregate it can cause performance problems in systems programming.  It introduces atomics for synchronizing ref counts.  This is why languages like Rust opted to have 2 types of ref counting (see Arc vs Rc).

In this case the getter is returning a std::shared_ptr by value, so the deduced type is not hiding an extra copy.  We've had instances where `auto` has hid them before so it's best to stick to the C++ style guide, especially when capturing intermediate values.

For fun check this out: https://www.volatileint.dev/posts/auto-type-deduction-gauntlet/

gaaclarke (approved):
lgtm! thanks

PR #182788, opened 2026-02-23
Custom KGP task and migration to AGP api for geting kgp version

reidbaker:
Create a unified `PrintTask` that uses a standard Gradle `Property<String>` to lazily hold and print a message.
Modify `javaVersion`, `kgpVersion`, and `printBuildVariants` to use this new unified task.
Modify existing tests in `FlutterPluginTest` and `FlutterPluginUtilsTest` to mock the newly required task registration and provide data.
Add new unit tests in `PrintTaskTest.kt` to cover the `PrintTask` behavior under both simple string and lazy mapped provider usages.

KgpVersion is migrated to use the *internal* AGP api for getting kgp version. 

Related to #180137

Format command that can be run from package_tools/gradle on only files that are staged for commit. `ktlint --editorconfig=../../../dev/bots/test/analyze-test-input/.editorconfig --baseline=../../../dev/bots/test/analyze-test-input/ktlint-baseline.xml $(git diff --name-only --cached| grep \.kt | sed 's/packages\/flutter_tools\/gradle\///g') --format`

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/flutter_tools/gradle/src/main/kotlin/VersionFetcher.kt
```diff
@@ -44,9 +45,16 @@ internal object VersionFetcher {
...
     internal fun getKGPVersion(project: Project): Version? {
-        // TODO(gmackall): AGP has a getKotlinAndroidPluginVersion(), and KGP has a
-        //                 getKotlinPluginVersion(). Consider replacing this implementation with one of
-        //                 those.
+        // AGP and Kgp have methods for getting kotlin version.
+        // AGP's method is internal, we try to use it anyway.
+        // KGP's version in org.jetbrains.kotlin.gradle.plugin.DefaultKotlinBasePlugin is not
+        // available when this method is called.
+        // When testing call `setAgpKotlinVersionToNull(project)`.
+        val agpDefinedKgpVersion = getKotlinAndroidPluginVersion(project)
```
reidbaker, line 53: This method was added in https://cs.android.com/android-studio/platform/tools/base/+/0793d77987f447b8aad9dd801cd9f10f38228267. Which is after the publish date of AGP 8.1 and is why one of the tests is failing. 
The error AGP version right now is 8.1.1 so this api cannot be used without bumping the agp version. 

When I get back I should be able to bisect the jars for AGP to find when the method was added. 
https://maven.google.com/web/index.html?q=com.android.tools.build#com.android.tools.build:gradle

reidbaker:
Marked as draft since the new api usage is older than our oldest supported agp version.

packages/flutter_tools/gradle/src/main/kotlin/tasks/PrintKgpTask.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Task to print the current Kotlin Gradle Plugin version used by the project.
+ */
+abstract class PrintKgpTask : DefaultTask() {
+    init {
+        description = "Print the current kgp version used by the project."
+    }
+
+    @TaskAction
+    fun run() {
+        println("KGP Version: ${VersionFetcher.getKGPVersion(project)}")
```
gmackall: This seems to be deprecated behavior that will be removed in gradle 10, as the gemini comment is alluding to:
https://docs.gradle.org/current/userguide/upgrading_version_8.html#task_project
> Calling [Task.getProject()](https://docs.gradle.org/current/javadoc/org/gradle/api/Task.html#getProject--) from a task action at execution time is now deprecated and will be made an error in Gradle 10. This method can still be used during configuration time.

I believe this run method takes place during the execution phase, right - that is the point of using `project.tasks.register`, is that we avoid doing this work during configuration (and therefore it happens during execution).

packages/flutter_tools/gradle/src/main/kotlin/VersionFetcher.kt
gmackall: Do you know what version is the minimum this is in?

packages/flutter_tools/gradle/src/main/kotlin/tasks/PrintKgpTask.kt
```diff
@@ -0,0 +1,19 @@
...
+    }
```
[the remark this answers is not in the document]
reidbaker: This does not happen often  and does not happen on the critical path.

gmackall (commented):
Re this comment https://github.com/flutter/flutter/pull/182788/changes#r3041587350 

Is this addressed? Or otherwise do we plan to land this and then resolve before Gradle 10?

reidbaker:
> Re this comment https://github.com/flutter/flutter/pull/182788/changes#r3041587350
> 
> Is this addressed? Or otherwise do we plan to land this and then resolve before Gradle 10?

it is now addressed.

gmackall (commented):
Is there a reason why we are building a different pattern for `PrintKgpTask` and the Java/build variants case? Abstractly it seems to me like the combo of `PrintTaskDeferred`'s `closureInput: T` and `messageClosure: (T) -> String` is essentially just a provider. It seems like we could instead build a general purpose version of the KGP pattern here:

```kt
package com.flutter.gradle.tasks

import org.gradle.api.DefaultTask
import org.gradle.api.provider.Property
import org.gradle.api.tasks.Input
import org.gradle.api.tasks.TaskAction

/** Prints a message computed lazily, before the task action runs. */
abstract class PrintTask : DefaultTask() {
    @get:Input
    abstract val message: Property<String>

    @TaskAction
    fun run() = println(message.get())
}
```

And then use it for all three cases:
```kt
internal fun addTaskForJavaVersion(project: Project) {
    project.tasks.register<PrintTask>("javaVersion") {
        description = "Print the current java version used by gradle. see: " +
            "https://docs.gradle.org/current/javadoc/org/gradle/api/JavaVersion.html"
        message.set(project.provider { VersionFetcher.getJavaVersion().toString() })
    }
}

internal fun addTaskForKGPVersion(project: Project) {
    project.tasks.register<PrintTask>("kgpVersion") {
        description = "Print the current kgp version used by the project."
        message.set(
            project.provider { "KGP Version: " + (VersionFetcher.getKGPVersion(project)?.toString() ?: "null") }
        )
    }
}

internal fun addTaskForPrintBuildVariants(project: Project) {
    val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
    val variants = project.objects.listProperty(String::class.java)
    androidComponents.onVariants { variant -> variants.add(variant.name) }

    project.tasks.register<PrintTask>("printBuildVariants") {
        description = "Prints out all build variants for this Android project"
        message.set(variants.map { names -> names.joinToString("\n") { name -> "BuildVariant: $name" } })
    }
}
```

I think this accomplishes the same goals - from the [lazy configuration docs](https://docs.gradle.org/current/userguide/lazy_configuration.html):
> The method [Property.set(Provider)](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html#set-org.gradle.api.provider.Provider-) specifies a Provider for the value for the property, overwriting whatever value may have been present. This allows you to wire together Provider and Property instances before the values are configured.

I think this makes our general purpose task easier to read as well, as the current version has an added level of abstraction that this removes.

reidbaker:
> Is there a reason why we are building a different pattern for `PrintKgpTask` and the Java/build variants case? Abstractly it seems to me like the combo of `PrintTaskDeferred`'s `closureInput: T` and `messageClosure: (T) -> String` is essentially just a provider. It seems like we could instead build a general purpose version of the KGP pattern here:
> 
> ```kotlin
> package com.flutter.gradle.tasks
> 
> import org.gradle.api.DefaultTask
> import org.gradle.api.provider.Property
> import org.gradle.api.tasks.Input
> import org.gradle.api.tasks.TaskAction
> 
> /** Prints a message computed lazily, before the task action runs. */
> abstract class PrintTask : DefaultTask() {
>     @get:Input
>     abstract val message: Property<String>
> 
>     @TaskAction
>     fun run() = println(message.get())
> }
> ```
> 
> And then use it for all three cases:
> 
> ```kotlin
> internal fun addTaskForJavaVersion(project: Project) {
>     project.tasks.register<PrintTask>("javaVersion") {
>         description = "Print the current java version used by gradle. see: " +
>             "https://docs.gradle.org/current/javadoc/org/gradle/api/JavaVersion.html"
>         message.set(project.provider { VersionFetcher.getJavaVersion().toString() })
>     }
> }
> 
> internal fun addTaskForKGPVersion(project: Project) {
>     project.tasks.register<PrintTask>("kgpVersion") {
>         description = "Print the current kgp version used by the project."
>         message.set(
>             project.provider { "KGP Version: " + (VersionFetcher.getKGPVersion(project)?.toString() ?: "null") }
>         )
>     }
> }
> 
> internal fun addTaskForPrintBuildVariants(project: Project) {
>     val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
>     val variants = project.objects.listProperty(String::class.java)
>     androidComponents.onVariants { variant -> variants.add(variant.name) }
> 
>     project.tasks.register<PrintTask>("printBuildVariants") {
>         description = "Prints out all build variants for this Android project"
>         message.set(variants.map { names -> names.joinToString("\n") { name -> "BuildVariant: $name" } })
>     }
> }
> ```
> 
> I think this accomplishes the same goals - from the [lazy configuration docs](https://docs.gradle.org/current/userguide/lazy_configuration.html):
> 
> > The method [Property.set(Provider)](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html#set-org.gradle.api.provider.Provider-) specifies a Provider for the value for the property, overwriting whatever value may have been present. This allows you to wire together Provider and Property instances before the values are configured.
> 
> I think this makes our general purpose task easier to read as well, as the current version has an added level of abstraction that this removes.

IIRC because this pr is quite old. The reason why I had it different was because of the amount of work that was done in the different calls and that gradles cache configuration docs ban using project state (like project.provider). 
https://docs.gradle.org/current/userguide/configuration_cache_requirements.html 
Let me see if there is another way to do what you are asking.

gmackall (dismissed):

gmackall (approved):

PR #182789, opened 2026-02-23
[A11y][Web] Support isAccessibilityFocusBlocked flag on web

hannah-hyj:
... using aria-hidden. 

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -890,9 +890,9 @@ abstract class SemanticRole {
 
     if (semanticsObject.isFlagsDirty) {
       if (semanticsObject.flags.isAccessibilityFocusBlocked) {
-        element.setAttribute('aria-hidden', 'true');
+        setAttribute('aria-hidden', 'true');
       } else {
-        element.removeAttribute('aria-hidden');
+        removeAttribute('aria-hidden');
       }
     }
 
--- a/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -888,6 +888,14 @@ abstract class SemanticRole {
       updateValidationResult();
     }
 
+    if (semanticsObject.isFlagsDirty) {
+      if (semanticsObject.flags.isAccessibilityFocusBlocked) {
+        element.setAttribute('aria-hidden', 'true');
+      } else {
+        element.removeAttribute('aria-hidden');
+      }
+    }
+
     final List<SemanticBehavior>? behaviors = _behaviors;
     if (behaviors != null) {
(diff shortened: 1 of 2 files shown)


engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -888,6 +888,14 @@ abstract class SemanticRole {
       updateValidationResult();
     }
 
+    if (semanticsObject.isFlagsDirty) {
+      if (semanticsObject.flags.isAccessibilityFocusBlocked) {
+        setAttribute('aria-hidden', 'true');
```
chunhtai, line 893: this will hide the entire subtree, not just this node. If this is set by AccessibilityFocusBlockType.blockNode this will be broken i believe

mdebbar: Good point, @chunhtai.

Also, a few things stand out from the aria-hidden [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden):

> Warning: Do not use aria-hidden="true" on focusable elements.

They don't explain why, but this warning is highlighted in red to indicate it's important. For focusable nodes, it could be better to use the `disabled` or `inert` attribute.

> Using aria-hidden="false" will not re-expose the element to assistive technology if any of its parents specify aria-hidden="true".

Once `aria-hidden: true` is set, there's no way to unhide the children.

hannah-hyj: >Warning: Do not use aria-hidden="true" on focusable elements. 

My understanding is a focusable element should not be a11y unfocusable.  We have the same check in the framework side, A semantics node should not be focusable and  a11y unfocusable , but it can be unfocusable but a11y focusable. So it's wont be a problem. 

>Once aria-hidden: true is set, there's no way to unhide the children.

That's unfortunate. on the framework side we block a11y focus on node or subtree.  I think we should either 1. annotate on the framework side that web only support block a11y focus for a subtree, blocking a node will block its children too  or 2. just don't support block a11y focus on web. using the old way (dropping semantics nodes from the tree) in framework instead.  

I prefer option 1 on this as blocking a subtree is the more common use case and the function to block a subtree is enough to replace BlockSemantics.

This is the design doc for  isAccessibilityFocusBlocked to replace BlockSemantics: https://docs.google.com/document/d/1niJnShCR3-PIMYo1a9lWdVySBqr_wmXYPvBXntvcc0c/edit?tab=t.0

Piinks:
Hey @hannah-hyj! Is this still something we want to land?

hannah-hyj:
Closing this PR as I don't have immediate plan to land it.

marwfair:
Any plans on getting back to this, or should a new issue be opened?

marwfair:
I created a new issue for this: https://github.com/flutter/flutter/issues/191484

Issue #179804, opened 2025-12-12
Expose vm services for running accessibility guidelines on a running App


chunhtai:
### Use case

As doc flutter.dev/go/devtool-a11y-features (actual redirect TBD) mentioned, we need a VM services for devtool to run accessibility guideline of the current screen and the return result.

### Proposal

This will first requires https://github.com/flutter/flutter/issues/32057 to be addressed first. Once it does we should create a VM services to expose the endpoint

PR #182731, opened 2026-02-22
`SizedBox.square()`

nate-thegrate:
From the [style guide](https://github.com/flutter/flutter/blob/5603e19b4b3ebe7151e54d66577aa4baa0925845/docs/contributing/Style-guide-for-Flutter-repo.md#prefer-specialized-functions-methods-and-constructors):

> ### Prefer specialized functions, methods and constructors
>
> Use the most relevant constructor or method, when there are multiple options.
>
> Example:
>
> ```dart
> // BAD:
> const EdgeInsets.TRBL(0.0, 8.0, 0.0, 8.0);
>
> // GOOD:
> const EdgeInsets.symmetric(horizontal: 8.0);
> ```

<br>

This pull request implements that advice with respect to the [**SizedBox.square**](https://main-api.flutter.dev/flutter/widgets/SizedBox/SizedBox.square.html) constructor.

The vast majority of changes here are very bland; I'll make comments about the exciting parts.

PR #182791, opened 2026-02-23
Add vmservices for accessibilityEvaluation

chunhtai:
part of https://github.com/flutter/flutter/issues/179804

still needs to wire up devtools side

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
+++ b/packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
@@ -82,13 +82,10 @@ abstract class AccessibilityEvaluation {
 @internal
 class MinimumTapTargetEvaluation extends AccessibilityEvaluation {
   /// Create a new [MinimumTapTargetEvaluation].
-  const MinimumTapTargetEvaluation({required this.size, required this.link});
+  const MinimumTapTargetEvaluation({required this.size});
 
   /// The minimum allowed size of a tappable node.
   final Size size;
-
-  /// A link describing the tap target evaluations for a platform.
-  final String link;
 
   /// The gap between targets to their parent scrollables to be considered valid
   /// tap targets.
@@ -151,8 +148,7 @@ class MinimumTapTargetEvaluation extends AccessibilityEvaluation {
         Violation(
           node,
           '$node: expected tap target size of at least $size, '
-          'but found $candidateSize\n'
-          'See also: $link',
+          'but found $candidateSize\n',
         ),
       );
     }
@@ -248,7 +244,22 @@ class LabeledTapTargetEvaluation extends AccessibilityEvaluation {
 @internal
 class MinimumTextContrastEvaluation extends AccessibilityEvaluation {
   /// Create a new [MinimumTextContrastEvaluation].
-  const MinimumTextContrastEvaluation();
+  const MinimumTextContrastEvaluation({
+    required this.minNormalTextContrastRatio,
+    required this.minLargeTextContrastRatio,
+  });
+
+  /// The minimum contrast ratio for normal text.
+  ///
+  /// Normal text is text that is smaller than [kLargeTextMinimumSize] (18.0) or
+  /// smaller than [kBoldTextMinimumSize] (14.0) if bold.
+  final double minNormalTextContrastRatio;
+
+  /// The minimum contrast ratio for large text.
+  ///
+  /// Large text is text that is at least [kLargeTextMinimumSize] (18.0) or at
+  /// least [kBoldTextMinimumSize] (14.0) if bold.
+  final double minLargeTextContrastRatio;
 
   /// The minimum text size considered large for contrast checking.
   ///
@@ -464,46 +475,9 @@ class MinimumTextContrastEvaluation extends AccessibilityEvaluation {
     final double fontSizeOrDefault = fontSize ?? _kDefaultFontSize;
     if ((bold && fontSizeOrDefault >= kBoldTextMinimumSize) ||
         fontSizeOrDefault >= kLargeTextMinimumSize) {
-      return kMinimumRatioLargeText;
+      return minLargeTextContrastRatio;
     }
-    return kMinimumRatioNormalText;
-  }
-}
-
-/// {@macro flutter.widgets.accessibility_evaluations.internal}
-///
-/// An evaluation which verifies that all nodes that contribute semantics via text
-/// meet **WCAG AAA** contrast levels.
-///
-/// The AAA level is defined by the Web Content Accessibility Guidelines:
-/// https://www.w3.org/WAI/WCAG22/Understanding/contrast-enhanced
-///
-/// This evaluation enforces a stricter contrast ratio:
-///  * Normal text must have a contrast ratio of at least 7.0
-///  * Large or bold text must have a contrast ratio of at least 4.5
-@internal
-class MinimumTextContrastEvaluationAAA extends MinimumTextContrastEvaluation {
-  /// Create a new [MinimumTextContrastEvaluationAAA].
-  const MinimumTextContrastEvaluationAAA();
-
-  /// The minimum contrast ratio for large text (bold ≥14px or ≥18px).
-  ///
-  /// Defined by WCAG AAA standard http://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html
-  static const double kAAAMinimumRatioLargeText = 4.5;
-
-  /// The minimum contrast ratio for normal text.
-  ///
-  /// Defined by WCAG AAA standard http://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html
-  static const double kAAAMinimumRatioNormalText = 7.0;
-
-  @override
-  double _targetContrastRatio(double? fontSize, {required bool bold}) {
-    final double fontSizeOrDefault = fontSize ?? MinimumTextContrastEvaluation._kDefaultFontSize;
-    if ((bold && fontSizeOrDefault >= MinimumTextContrastEvaluation.kBoldTextMinimumSize) ||
-        fontSizeOrDefault >= MinimumTextContrastEvaluation.kLargeTextMinimumSize) {
-      return kAAAMinimumRatioLargeText;
-    }
-    return kAAAMinimumRatioNormalText;
+    return minNormalTextContrastRatio;
   }
 }
 
--- a/packages/flutter/lib/src/widgets/binding.dart
+++ b/packages/flutter/lib/src/widgets/binding.dart
@@ -36,6 +36,7 @@ import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
 
 import '../foundation/_features.dart';
+import '_accessibility_evaluations.dart';
 import '_window.dart';
 import 'app.dart';
 import 'debug.dart';
@@ -752,6 +753,46 @@ mixin WidgetsBinding
           debugProfileBuildsEnabledUserWidgets = value;
         },
       );
+
+      registerServiceExtension(
+        name: WidgetsServiceExtensions.accessibilityEvaluations.name,
+        callback: (Map<String, String> parameters) async {
+          final String? type = parameters['type'];
+          if (type == null) {
+            throw Exception('type parameter is required');
+          }
+
+          switch (type) {
+            case 'MinimumTextContrastEvaluation':
+              if (parameters case {
+                'minNormalTextContrastRatio': final String minNormalTextContrastRatio,
+                'minLargeTextContrastRatio': final String minLargeTextContrastRatio,
+              }) {
+                final EvaluationResult result = await MinimumTextContrastEvaluation(
+                  minNormalTextContrastRatio: double.parse(minNormalTextContrastRatio),
+                  minLargeTextContrastRatio: double.parse(minLargeTextContrastRatio),
+                ).evaluate(this);
+                return _formatEvaluationResult(result.violations);
+              }
+              throw Exception('Invalid arguments');
+            case 'MinimumTapTargetEvaluation':
+              if (parameters case {'targetSize': final String targetSize}) {
+                final EvaluationResult result = await MinimumTapTargetEvaluation(
+                  size: Size.square(double.parse(targetSize)),
+                ).evaluate(this);
+                return _formatEvaluationResult(result.violations);
+              }
+              throw Exception('Invalid arguments');
+            case 'LabeledTapTargetEvaluation':
+              final EvaluationResult result = await const LabeledTapTargetEvaluation().evaluate(
+                this,
+              );
+              return _formatEvaluationResult(result.violations);
+            default:
+              throw Exception('unknown type: $type');
+          }
+        },
+      );
     }
 
     assert(() {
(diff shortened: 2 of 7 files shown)


packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
```diff
@@ -82,14 +82,11 @@ abstract class AccessibilityEvaluation {
...
 class MinimumTapTargetEvaluation extends AccessibilityEvaluation {
   /// Create a new [MinimumTapTargetEvaluation].
-  const MinimumTapTargetEvaluation({required this.size, required this.link});
+  const MinimumTapTargetEvaluation({required this.size});
 
   /// The minimum allowed size of a tappable node.
   final Size size;
 
-  /// A link describing the tap target evaluations for a platform.
-  final String link;
```
chunhtai, line 91: This is a clean up, I think the link or other supplement information should be provided by the subclass or consumer class that leverage this class.

packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
```diff
@@ -464,46 +466,9 @@ class MinimumTextContrastEvaluation extends AccessibilityEvaluation {
...
-/// meet **WCAG AAA** contrast levels.
-///
-/// The AAA level is defined by the Web Content Accessibility Guidelines:
-/// https://www.w3.org/WAI/WCAG22/Understanding/contrast-enhanced
-///
-/// This evaluation enforces a stricter contrast ratio:
-///  * Normal text must have a contrast ratio of at least 7.0
-///  * Large or bold text must have a contrast ratio of at least 4.5
-@internal
-class MinimumTextContrastEvaluationAAA extends MinimumTextContrastEvaluation {
```
chunhtai, line 485: This should be defined by the consumer classes like accessibilityGuidline or devtool's ui

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -752,6 +736,64 @@ mixin WidgetsBinding
...
+      registerServiceExtension(
+        name: WidgetsServiceExtensions.accessibilityEvaluations.name,
+        callback: (Map<String, String> parameters) async {
+          final String? type = parameters['type'];
+          if (type == null) {
+            return _formatError('type parameter is required');
+          }
+
+          switch (type) {
+            case 'MinimumTextContrastEvaluation':
```
chunhtai, line 766: not sure if there is a better way to define these string since this endpoint will be called by devtools

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -773,6 +815,21 @@ mixin WidgetsBinding
     }());
   }
 
+  Map<String, String> _formatError(String message) {
+    return <String, String>{'error': message};
+  }
+
+  Map<String, List<Map<String, String>>> _formatEvaluationResult(List<Violation> violations) {
+    return <String, List<Map<String, String>>>{
+      'result': violations.map((Violation violation) {
```
bkonyi, line 819: Nit: this can be written without the map and the `result` key (it's already part of the JSON RPC response):

```dart
return List<Map<String, String>>[
  for (final violation in violations)
    {
      'nodeId': violation.node.id.toString(),
      'message': violation.reason,
    }
};
```

Or with some object destructuring:

```dart
return List<Map<String, String>>[
  for (final Violation(:node, :reason) in violations)
    {
      'nodeId': node.id.toString(),
      'message': reason,
    }
}
```

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+          child: Semantics(label: 'button', onTap: _emptyCallback),
+        ),
+      ),
+    ),
+  );
+}
+
+void _emptyCallback() {}
+
+Future<Map<String, dynamic>> _runEvaluation(WidgetTester tester, Map<String, String> params) async {
```
bkonyi: Nit: in general, avoid using `dynamic` as it can lead to runtime errors due to objects of unexpected types leaking in. Prefer using `Object?` or `Object` and explicitly casting them to the right type to make it more clear that the value should be of a specific type.

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+
+  tearDownAll(() {
+    debugEnabledFeatureFlags.clear();
+    debugEnabledFeatureFlags.addAll(originalFeatureFlags);
+  });
+
+  testWidgets(
+    'accessibilityEvaluations service extension returns violations for MinimumTapTargetEvaluation',
+    (WidgetTester tester) async {
+      final SemanticsHandle handle = tester.ensureSemantics();
```
bkonyi, line 57: Nit: consider using `addTearDown(...)` here and elsewhere to dispose of this handle.

```dart
addTearDown(handle.dispose);
```

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+import 'widgets_app_tester.dart';
+
+class AccessibilityEvaluationTestBinding extends AutomatedTestWidgetsFlutterBinding {
+  static AccessibilityEvaluationTestBinding? _instance;
+
+  static AccessibilityEvaluationTestBinding ensureInitialized() {
+    if (_instance == null) {
+      AccessibilityEvaluationTestBinding();
+    }
+    return _instance!;
```
bkonyi: This can be more concise:

```dart
static AccessibilityEvaluationTestBinding ensureInitialized() {
  return _instance ??= AccessibilityEvaluationBinding();
}
```

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  @override
+  void initInstances() {
+    super.initInstances();
+    _instance = this;
+  }
+
+  final Map<String, ServiceExtensionCallback> extensions = <String, ServiceExtensionCallback>{};
```
bkonyi: Nit: Is this LHS type necessary?

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+
+  static AccessibilityEvaluationTestBinding ensureInitialized() {
+    if (_instance == null) {
+      AccessibilityEvaluationTestBinding();
+    }
+    return _instance!;
+  }
+
+  @override
+  void initInstances() {
```
bkonyi: Is this needed if we're invoking `ensureInitialized()` at the beginning of `main()`?

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -773,6 +815,21 @@ mixin WidgetsBinding
     }());
   }
 
+  Map<String, String> _formatError(String message) {
+    return <String, String>{'error': message};
```
bkonyi: I think we should be throwing an exception rather than returning a valid result. `registerServiceExtension` has a catch block that returns an actual JSON RPC error. This approach will result in the request succeeding with the JSON response looking something like:

```dart
{
  'id': 123,
  'result': {
    'error': 'message'
  }
}
```

packages/flutter/lib/src/widgets/binding.dart
```diff
@@ -752,6 +736,64 @@ mixin WidgetsBinding
...
+
+      registerServiceExtension(
+        name: WidgetsServiceExtensions.accessibilityEvaluations.name,
+        callback: (Map<String, String> parameters) async {
+          final String? type = parameters['type'];
+          if (type == null) {
+            return _formatError('type parameter is required');
+          }
+
+          switch (type) {
```
bkonyi, line 765: I think this can be simplified using pattern matching and  object destructuring. For example:

```dart
  switch (type) {
    case 'MinimumTextConstrastEvaluation':
      if (parameters case {
        'minNormalTextContrastRatio': final double minNormalTextContrastRatio,
        'minLargeTextContrastRatio': final double minLargeTextContrastRatio,
      }) {
        final EvaluationResult result = await MinimumTextContrastEvaluation(
          minNormalTextContrastRatio: minNormalTextContrastRatioDouble,
          minLargeTextContrastRatio: minLargeTextContrastRatioDouble,
        ).evaluate(this);
        return _formatEvaluationResult(result.violations);
      }
      throw Exception('Invalid arguments');
    case 'MinimumTapTargetEvaluation':
      // ...
  }
```

packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
```diff
@@ -248,7 +244,13 @@ class LabeledTapTargetEvaluation extends AccessibilityEvaluation {
 @internal
 class MinimumTextContrastEvaluation extends AccessibilityEvaluation {
   /// Create a new [MinimumTextContrastEvaluation].
-  const MinimumTextContrastEvaluation();
+  const MinimumTextContrastEvaluation({
+    required this.minNormalTextContrastRatio,
+    required this.minLargeTextContrastRatio,
+  });
+
+  final double minNormalTextContrastRatio;
```
bkonyi, line 256: These fields need some documentation.

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,392 @@
...
+        ),
+      );
+
+      // Run MinimumTapTargetEvaluation
+      final Map<String, dynamic> tapTargetResult = await _runEvaluation(tester, <String, String>{
+        'type': 'MinimumTapTargetEvaluation',
+        'targetSize': '48.0',
+      });
+
+      expect(tapTargetResult.keys, contains('result'));
```
bkonyi: `expect(tapTargetResult, contains('result'));` should also work here.

packages/flutter/lib/src/widgets/binding.dart
chunhtai: Will this work? ServiceExtensionCallback expects a Future<Map<String, dynamic>> as return value

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
chunhtai: Unfortunately, the test checks for handle dispose before teardown is called, so this will crash the app.

filed https://github.com/flutter/flutter/issues/182978

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
chunhtai: yes for public api they have to declare the type

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
chunhtai: good call

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
loic-sharma: You should be able to use `.new()` to omit the RHS: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#prefer-using-dot-shorthands-to-initialize-top-level-variables-and-fields

Up to you though.

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,424 @@
...
+  testWidgets(
+    'accessibilityEvaluations service extension returns violations for MinimumTapTargetEvaluation',
+    (WidgetTester tester) async {
+      final SemanticsHandle handle = tester.ensureSemantics();
+
+      // Pump a widget with a small tap target
+      await tester.pumpWidget(
+        TestWidgetsApp(
+          home: Center(
+            child: SizedBox(
```
loic-sharma: Consider using `SizedBox.square` here and below. nate-thegrate sent a patch earlier to clean these up (https://github.com/flutter/flutter/pull/182731)

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,424 @@
...
+              width: 40.0,
+              height: 40.0,
+              child: Semantics(label: 'small button', onTap: _emptyCallback),
```
loic-sharma: Just curious, you can't do something like this?

```suggestion
              child: Semantics(label: 'small button', onTap: () {}),
```

I'm totally fine with the code as-is though, feel free to keep it.

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
```diff
@@ -0,0 +1,424 @@
...
+      final SemanticsHandle handle = tester.ensureSemantics();
+      await _pumpTestWidget(tester);
+
+      expect(() async {
+        await _runEvaluation(tester, <String, String>{
+          'type': 'MinimumTextContrastEvaluation',
+          'minNormalTextContrastRatio': 'foo',
+          'minLargeTextContrastRatio': '3.0',
+        });
+      }, throwsA(isA<FormatException>()),
```
loic-sharma: Does the formatter get mad if you put this on a separate line?

packages/flutter/test/widgets/accessibility_evaluations_service_extension_test.dart
bkonyi: Oh that's interesting! Disregard then :)

packages/flutter/lib/src/widgets/binding.dart
bkonyi: Ugh, yup, you're right. The implementation of `registerServiceExtension` is more restrictive than it needs to be. In theory, any JSON serializable type should be supported as a return value, as [that's what `ServiceExtensionResponse.result(...)` accepts](https://api.flutter.dev/flutter/dart-developer/ServiceExtensionResponse/ServiceExtensionResponse.result.html).

This isn't a big deal, it just means there's some unnecessary layers of collections and the RPC response will look weird.

I'd say it'd be good to fix the `registerServiceExtension` implementation to allow for arbitrary JSON encodable responses, but it looks like we inject the `method` and set the `type` to `_extensionType` in the response, assuming it's a `Map`. Looking through code search, we really don't need to do this as nothing I can find cares about these values in the response outside of tests and they don't add any value, so I think it'd be safe to remove them and allow for non-`Map` results.

I know that's extra work, so I'll leave it up to you if that's a modification you're willing to pursue.

packages/flutter/lib/src/widgets/binding.dart
bkonyi: Are there other service extension implementations that define public constants  anywhere for parameter names or types? Is there any enum associated with these names that could be used?

packages/flutter/lib/src/widgets/binding.dart
chunhtai: If we do change the return response, we may need to migrate the consumer of these endpoint. Do you know what other services we need to look into besides devtools and mcp?

I think this should probably be another pr, filed https://github.com/flutter/flutter/issues/183017

bkonyi (approved):

PR #182815, opened 2026-02-24
Fix/ios codesign xattr for Building in Xcode 26

coder-ishan:
## Fix iOS/macOS codesign failure caused by extended attributes

Fixes #181103

### Problem

On macOS Sequoia (15.x) with Xcode 26, iOS and macOS builds fail during codesigning with:

```
resource fork, Finder information, or similar detritus not allowed
```

The root cause is the `com.apple.provenance` extended attribute, which newer macOS versions automatically add to files during copy operations. The existing `removeFinderExtendedAttributes` function only stripped `com.apple.FinderInfo`, leaving `com.apple.provenance` behind — which `codesign` rejects.

### Changes

**1. Renamed and expanded xattr removal (`ios/mac.dart`)**

Renamed `removeFinderExtendedAttributes` → `removeExtendedAttributes` to reflect its broader scope. The function now removes both `com.apple.FinderInfo` and `com.apple.provenance`, while intentionally preserving other attributes like `com.apple.xcode.CreatedByBuildSystem` that Xcode relies on.

**2. Fixed xattr target scope for iOS frameworks (`ios.dart`)**

Changed `_signFramework` to strip xattrs from the entire `.framework` directory (`binary.parent`) rather than just the binary file. `codesign` validates the whole bundle, so attributes on any file within the framework can cause failures.

**3. Added xattr stripping to macOS native asset codesigning (`native_assets_host.dart`)**

The `codesignDylib` function had no xattr removal at all. Added a call to the shared `removeExtendedAttributes()` helper before codesigning, ensuring consistent behavior across both iOS and macOS native asset paths.

**4. Updated all tests**

- Updated `ios_test.dart` to expect xattr commands targeting the `.framework` directory instead of the binary
- Updated `native_assets_test.dart` to expect the two targeted `xattr -r -d` commands (one per attribute) before each codesign invocation

### Why targeted removal instead of `xattr -cr`

The issue description suggests `xattr -cr` (clear all attributes) as an alternative. This PR intentionally removes only specific known-problematic attributes because:

- `com.apple.xcode.CreatedByBuildSystem` is used by Xcode to manage build directories — clearing it could cause subtle build system issues
- Targeted removal is the existing pattern in this codebase and makes the intent explicit
- New problematic attributes can be added to the set in `removeExtendedAttributes()` as needed

### Future refactoring

`removeExtendedAttributes()` currently lives in `ios/mac.dart` but is now used by both iOS and macOS code paths (including `native_assets/macos/native_assets_host.dart` and `macos/xcdevice.dart`). A future cleanup should move it to a shared location (e.g., `base/` or `build_system/targets/darwin.dart`) to better reflect its cross-platform usage and avoid the `ios/` → `macos/` import direction.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/lib/src/build_system/targets/ios.dart
+++ b/packages/flutter_tools/lib/src/build_system/targets/ios.dart
@@ -914,8 +914,10 @@ Future<void> _createStubAppFramework(
 }
 
 Future<void> _signFramework(Environment environment, File binary, BuildMode buildMode) async {
+  // Strip xattrs from the entire .framework directory, not just the binary.
+  // codesign validates the whole bundle. See https://github.com/flutter/flutter/issues/181103
   await removeExtendedAttributes(
-    binary,
+    binary.parent,
     ProcessUtils(processManager: environment.processManager, logger: environment.logger),
     environment.logger,
   );
--- a/packages/flutter_tools/lib/src/isolated/native_assets/macos/native_assets_host.dart
+++ b/packages/flutter_tools/lib/src/isolated/native_assets/macos/native_assets_host.dart
@@ -13,6 +13,7 @@ import '../../../base/process.dart';
 import '../../../build_info.dart';
 import '../../../build_system/targets/darwin.dart';
 import '../../../globals.dart' as globals;
+import '../../../ios/mac.dart';
 import '../native_assets.dart';
 
 /// Create an `Info.plist` in [target] for a framework with a single dylib.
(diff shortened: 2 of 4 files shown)


coder-ishan:
I have tested this patch by using it with a project on my device.It works now, but future refactoring suggested can be done on approval and after discussions

vashworth:
@coder-ishan Looks like https://github.com/flutter/flutter/issues/181103 is closed? Are you still getting that issue? Can you provide reproduction steps?

PR #182865, opened 2026-02-24
Update package contrib docs for batched release

stuartmorgan-g:
Now that some packages are using batched release mode, update the contributor docs about version and CHANGELOG updates to address both models.


chunhtai (approved):
LGTM

PR #182868, opened 2026-02-24
Add await to callsites of BasicMessageChannel.send

victorsanni:
Impl:

```dart
  /// Sends the specified [message] to the platform plugins on this channel.
  ///
  /// Returns a [Future] which completes to the received response, which may
  /// be null.
  Future<T?> send(T message) async {
    return codec.decodeMessage(await binaryMessenger.send(name, codec.encodeMessage(message)));
  }
```


dev/benchmarks/platform_channels_benchmarks/lib/main.dart
```diff
@@ -69,7 +69,7 @@ Future<double> _runBasicStandardParallel(
   final counter = _Counter();
   watch.start();
   for (var i = 0; i < parallel; ++i) {
-    basicStandard.send(payload).then((Object? result) {
+    await basicStandard.send(payload).then((Object? result) {
       _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
     });
```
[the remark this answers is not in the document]
dkwingsmt: Use `await Future.forEach()` or `Future.wait` instead.

examples/flutter_view/lib/main.dart
```diff
@@ -54,8 +54,8 @@ class _MyHomePageState extends State<MyHomePage> {
     return _emptyMessage;
   }
 
-  void _sendFlutterIncrement() {
-    platform.send(_pong);
+  Future<void> _sendFlutterIncrement() async {
```
dkwingsmt: This is technically incorrect, since `FloatingActionButton.onPressed` expects a `void` return value. The correct way to do this is to catch message errors and report to `FlutterError`. Since this is just an example, it's acceptable if the compiler doesn't complain (which I'm not sure).

packages/flutter/lib/src/widgets/navigator.dart
dkwingsmt: This future should have a `.catch` that reports error.

dkwingsmt (commented):
Generally ok.

examples/flutter_view/lib/main.dart
victorsanni: I added an ignore here, is this ok?

dkwingsmt: Yeah, sure.

dev/benchmarks/platform_channels_benchmarks/lib/main.dart
```diff
@@ -69,9 +68,11 @@ Future<double> _runBasicStandardParallel(
   final counter = _Counter();
   watch.start();
   for (var i = 0; i < parallel; ++i) {
-    basicStandard.send(payload).then((Object? result) {
-      _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
-    });
+    await Future.wait([
```
dkwingsmt: `Future.wait` is meant to run tasks in parallel and wait for the completion of all of them.

You should replace the entire for loop with something like (psuedocode):
```dart
await Future.wait(List.generate(parallel, (_) => () async {
    await basicStandard.send(payload);
    await _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
}));
```

packages/flutter/lib/src/widgets/navigator.dart
victorsanni: Is the current impl satisfactory?

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -4267,6 +4267,7 @@ class SemanticsNode with DiagnosticableTreeMixin {
     if (!attached) {
       return;
     }
+    // ignore: unawaited_futures
```
victorsanni: This is a public API, so changing its signature might be breaking.

dev/benchmarks/platform_channels_benchmarks/lib/main.dart
```diff
@@ -68,11 +67,13 @@ Future<double> _runBasicStandardParallel(
   final completer = Completer<int>();
   final counter = _Counter();
   watch.start();
-  for (var i = 0; i < parallel; ++i) {
-    basicStandard.send(payload).then((Object? result) {
+  await Future.wait(
+    Iterable.generate(parallel, (_) async {
+      await basicStandard.send(payload);
+      // ignore: unawaited_futures
```
victorsanni: Ignoring because recursive method is intended to run in parallel

dkwingsmt: Are you sure we can't have await here? I think this still runs in parallel even if you use await, since Iterable.generate runs everything at the same time. In other words, each `(_) async { ... }` runs in parallel.

packages/flutter/lib/src/semantics/semantics.dart
dkwingsmt: Then let's make it report to `FlutterError` in `.catchError`.

dev/benchmarks/platform_channels_benchmarks/lib/main.dart
```diff
@@ -41,20 +41,19 @@ class _Counter {
...
+) async {
   counter.count += 1;
   if (counter.count == count) {
     completer.complete(counter.count);
   } else if (counter.count < count) {
-    basicStandard.send(payload).then((Object? result) {
-      _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
-    });
+    await basicStandard.send(payload);
+    _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
```
dkwingsmt: `await` (or `return`)

packages/flutter/lib/src/semantics/semantics.dart
```diff
@@ -4267,7 +4267,25 @@ class SemanticsNode with DiagnosticableTreeMixin {
     if (!attached) {
       return;
     }
-    SystemChannels.accessibility.send(event.toMap(nodeId: id));
+    SystemChannels.accessibility
+        .send(event.toMap(nodeId: id))
+        .then(
+          (dynamic _) {},
+          onError: (Object error, StackTrace stack) {
```
dkwingsmt, line 4279: FYI there's a standalone `catchError` method that's equivalent. Of course, the current implementation is acceptable too.

victorsanni: Yeah I used that method previously but realized `onError` is more terse, `catchError` requires an `Object` (or `null`) to be returned, so after throwing the error there is an extra step to explicitly return `null`.

dev/benchmarks/platform_channels_benchmarks/lib/main.dart
```diff
@@ -68,11 +67,12 @@ Future<double> _runBasicStandardParallel(
...
-    basicStandard.send(payload).then((Object? result) {
-      _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
-    });
-  }
+  await Future.wait(
+    Iterable.generate(parallel, (_) async {
+      await basicStandard.send(payload);
+      await _runBasicStandardParallelRecurse(basicStandard, counter, count, completer, payload);
+    }),
+  );
```
dkwingsmt: I'm sorry but I just realized that we might have been doing something wrong. `_runBasicStandardParallelRecurse` is using its `completer` parameter as the async control, and shouldn't need additional `Future` return value.

(I suspect that this entire function is basically reinventing `Future.wait` from scratch.)

In short, I think the original implementation is probably completely correct and we shouldn't change it at all. It doesn't have anything that violate the lint either, right? All futures either has `.then` or is awaited.

packages/flutter/lib/src/semantics/semantics.dart
victorsanni: Actually while writing tests for this PR (with help from Gemini) it seems that catchError might be necessary in cases where `.then.onValue` itself awaits a Future that throws, not only the original Future which calls the `.then`. 

Here `onValue` is empty so it might be better to keep `onError`, but in the navigator use case `.send` is called in the `onValue` of the original Future, so we need catchError to catch all errors from both the original Future and Futures awaited in `onValue`.

dkwingsmt (approved):

PR #182870, opened 2026-02-25
Add await or ignore lint to invokeMethod callsites

victorsanni:


dev/integration_tests/android_views/lib/motion_events_page.dart
```diff
@@ -213,11 +213,11 @@ class MotionEventsBodyState extends State<MotionEventsBody> {
   }
 
   void listenToFlutterViewEvents() {
-    channel.invokeMethod<void>('pipeFlutterViewEvents');
-    viewChannel?.invokeMethod<void>('pipeTouchEvents');
+    channel.invokeMethod<void>('pipeFlutterViewEvents');  // ignore: unawaited_futures
```
victorsanni: I didn't add `await` here because I'm wondering if the `Timer` is evidence the exclusion is intentional?

dev/integration_tests/hybrid_android_views/lib/motion_events_page.dart
```diff
@@ -194,9 +194,9 @@ class MotionEventsBodyState extends State<MotionEventsBody> {
   }
 
   void listenToFlutterViewEvents() {
-    viewChannel!.invokeMethod<void>('pipeTouchEvents');
+    viewChannel!.invokeMethod<void>('pipeTouchEvents'); // ignore: unawaited_futures
```
victorsanni, line 197: Same with the `Timer` here?

packages/flutter/lib/src/services/live_text.dart
```diff
@@ -28,7 +28,7 @@ abstract final class LiveText {
   /// If any [TextInputConnection] is currently active, calling this method will tell the text field
   /// to start Live Text input. If the current device doesn't support Live Text input,
   /// nothing will happen.
-  static void startLiveTextInput() {
-    SystemChannels.textInput.invokeMethod('TextInput.startLiveTextInput');
+  static Future<void> startLiveTextInput() async {
```
victorsanni, line 31: Changing the signature of a public method? Should we ignore instead?

packages/flutter/lib/src/services/system_chrome.dart
```diff
@@ -758,7 +758,7 @@ abstract final class SystemChrome {
     scheduleMicrotask(() {
       assert(_pendingStyle != null);
       if (_pendingStyle != _latestStyle) {
-        SystemChannels.platform.invokeMethod<void>(
+        SystemChannels.platform.invokeMethod<void>(  // ignore: unawaited_futures
```
victorsanni: Added ignore because this is in a microtask?

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2592,40 +2592,40 @@ class _PlatformTextInputControl with TextInputControl {
 
   @override
   void attach(TextInputClient client, TextInputConfiguration configuration) {
-    _channel.invokeMethod<void>('TextInput.setClient', <Object>[
+    _channel.invokeMethod<void>('TextInput.setClient', <Object>[  // ignore: unawaited_futures
```
victorsanni: Added ignore here (and elsewhere in this file) because awaiting will require changing the method signature

packages/flutter/lib/src/widgets/platform_menu_bar.dart
```diff
@@ -318,7 +318,7 @@ class DefaultPlatformMenuDelegate extends PlatformMenuDelegate {
     // Currently there's only ever one window, but the channel's format allows
     // more than one window's menu hierarchy to be defined.
     final windowMenu = <String, Object?>{'0': representation};
-    channel.invokeMethod<void>(_kMenuSetMethod, windowMenu);
+    channel.invokeMethod<void>(_kMenuSetMethod, windowMenu); // ignore: unawaited_futures
```
victorsanni: awaiting will mean changing method signature?

packages/flutter/lib/src/services/live_text.dart
dkwingsmt: I think this is ok since this method can't be overridden. The worst that can happen is that apps will get new lint warnings for unwaited futures.

packages/flutter/lib/src/services/system_chrome.dart
dkwingsmt: Unawaited futures in production code should be reported to FlutterError in a `.catch`.

packages/flutter/lib/src/services/text_input.dart
dkwingsmt: Similarly, report to `FlutterError` in a `.catch`.

packages/flutter/lib/src/services/undo_manager.dart
```diff
@@ -90,8 +90,8 @@ class UndoManager {
     throw MissingPluginException();
   }
 
-  void _setUndoState({bool canUndo = false, bool canRedo = false}) {
-    _channel.invokeMethod<void>('UndoManager.setUndoState', <String, bool>{
+  Future<void> _setUndoState({bool canUndo = false, bool canRedo = false}) async {
+    await _channel.invokeMethod<void>('UndoManager.setUndoState', <String, bool>{
```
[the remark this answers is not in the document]
dkwingsmt: The current change simply moves the `unawaited_futures` warning to `setUndoState`. We should either change `setUndoState`'s return value, or report the error in a `.catch`.

packages/flutter/lib/src/widgets/platform_menu_bar.dart
dkwingsmt: Yeah we probably don't want to break it. Report the error in `.catch` then.

packages/flutter/lib/src/widgets/platform_view.dart
```diff
@@ -801,12 +801,12 @@ class _AndroidViewState extends State<AndroidView> {
     }
   }
 
-  void _onFocusChange(bool isFocused) {
+  Future<void> _onFocusChange(bool isFocused) async {
```
dkwingsmt: This probably breaks. Report the error in `.catch` instead.

packages/flutter/test/services/platform_channel_test.dart
```diff
@@ -148,7 +148,7 @@ void main() {
         },
       );
       expect(
-        () => channel.invokeMethod<dynamic>('sayHello', 'hello'),
+        () => channel.invokeMethod<dynamic>('sayHello', 'hello'),  // ignore: unawaited_futures
```
dkwingsmt: This might be the same case as above. Simply removes `() =>` might work.

dkwingsmt (commented):
All instances of unawaited future in production code should be reported to `FlutterError` with a `.catchError` instead.

packages/flutter/lib/src/widgets/platform_view.dart
victorsanni: It has a `catchError` already, so we don't need to do anything here.

dev/integration_tests/hybrid_android_views/lib/motion_events_page.dart
dkwingsmt: Are you saying `Timer` should also be awaited? I don't think so...

packages/flutter/lib/src/rendering/binding.dart
```diff
@@ -485,10 +485,10 @@ mixin RendererBinding
     );
   }
 
-  void _handleWebFirstFrame(Duration _) {
+  Future<void> _handleWebFirstFrame(Duration _) async {
     assert(kIsWeb);
```
dkwingsmt, line 489: I'd be surprised if this is even allowed. If it is, we should somehow have a lint that avoids assigning `Future<void> ()` to `void ()`.

This `invokeMethod` should report errors in a `catchError` instead.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2593,39 +2593,47 @@ class _PlatformTextInputControl with TextInputControl {
   @override
   void attach(TextInputClient client, TextInputConfiguration configuration) {
     _channel.invokeMethod<void>('TextInput.setClient', <Object>[
+      // ignore: unawaited_futures
```
dkwingsmt: All errors in this file should be reported instead of ignored.

packages/flutter/lib/src/services/undo_manager.dart
```diff
@@ -91,6 +91,7 @@ class UndoManager {
   }
 
   void _setUndoState({bool canUndo = false, bool canRedo = false}) {
+    // ignore: unawaited_futures
```
dkwingsmt: Report instead of ignore.

packages/flutter/lib/src/widgets/platform_view.dart
```diff
@@ -1285,11 +1285,11 @@ class _PlatformViewLinkState extends State<PlatformViewLink> {
     }
   }
 
-  void _handleFrameworkFocusChanged(bool isFocused) {
+  Future<void> _handleFrameworkFocusChanged(bool isFocused) async {
```
dkwingsmt: Same as above.

packages/flutter/lib/src/widgets/platform_view.dart
dkwingsmt: The `catchError` only ignores `MissingPluginException` and does not report other errors.

We should keep `_onFocusChange`'s return type `void` (and hence removing `async`, because `Focus` does not process `_onFocusChange`'s returning future.

The two `catchError`s should report the error to `FlutterError` in their `else` branch.

packages/flutter/lib/src/widgets/platform_view.dart
```diff
@@ -944,14 +944,14 @@ abstract class _DarwinViewState<
 
   Future<ControllerT> createNewViewController(int id);
 
-  void _onFocusChange(bool isFocused, ControllerT controller) {
+  Future<void> _onFocusChange(bool isFocused, ControllerT controller) async {
```
dkwingsmt: Likewise: Keep the return value `void`, remove `async`, and report the future error in `catchError`.

dev/integration_tests/hybrid_android_views/lib/motion_events_page.dart
victorsanni: My understanding is `Timer` is used to lag the function calls, but I am not sure if the intention is to lag from the moment the Future is scheduled or from when it completes.

packages/flutter/lib/src/rendering/binding.dart
victorsanni: But we do this below: https://github.com/flutter/flutter/pull/182870/changes#r2850174911

dev/integration_tests/hybrid_android_views/lib/motion_events_page.dart
dkwingsmt: I see your point. I think it doesn't matter here since it's in a test where we're safe to ignore all futures. So the current change is good by adding ignore to both `invokeMethod`. Or is there any other concern you're talking about?

packages/flutter/lib/src/rendering/binding.dart
dkwingsmt: First, for `startLiveTextInput`, there are two ways to design its API.
* Either we make `startLiveTextInput` return void, indicating that code who uses it is not allowed to know the result of this `startLiveTextInput` call and `startLiveTextInput` should report to `FlutterError` in a `catchError` within itself.
* Or we make `startLiveTextInput` return a future, indicating that code who uses it is responsible of handling the result of this `startLiveTextInput` call, and is recommended to report to `FlutterError` in a `catchError` after this call.

We're making a design decision here, because we can choose how users should handle it. What I suggested in that comment suggested the 2nd choice, although I'm ok with either. If we choose the 2nd choice, we should make sure our own invocations process the errors - there's exactly one use in `editable_text.dart`.

As for `_handleWebFirstFrame` (and other similar cases below), the callback is handled by an API that does not handle the async result at all (in this case, `addPostFrameCallback`; in other cases, `Focus`). Therefore we have to choose the 1st way and provide a callback that handles these errors.

packages/flutter/lib/src/services/text_input.dart
```diff
@@ -2592,81 +2592,221 @@ class _PlatformTextInputControl with TextInputControl {
...
+        ])
+        .then(
+          (void _) {},
+          onError: (Object error, StackTrace stack) {
+            FlutterError.reportError(
+              FlutterErrorDetails(
+                exception: error,
+                stack: stack,
+                library: 'services library',
+                context: ErrorDescription('while attaching the text input client'),
```
dkwingsmt: nit: Extract a private helper function that takes a future and reports its error, since there are so many with the same pattern.

Honestly we should have a function globally. It can probably be a static method of `FlutterError`. But for now we can make it privately local here.

packages/flutter/lib/src/widgets/platform_view.dart
dkwingsmt: This should also be reported to FlutterError instead of returning the future, since `Focus` does not handle futures.

dkwingsmt (commented):
Mostly OK. There's one place remaining.

packages/flutter/lib/src/widgets/platform_view.dart
```diff
@@ -806,33 +806,31 @@ class _AndroidViewState extends State<AndroidView> {
...
-          // We land the framework part of Android platform views keyboard
-          // support before the engine part. There will be a commit range where
-          // clearFocus isn't implemented in the engine. When that happens we
-          // just swallow the error here. Once the engine part is rolled to the
-          // framework I'll remove this.
-          // TODO(amirh): remove this once the engine's clearFocus is rolled.
-          return;
-        }
+      _controller.clearFocus().catchError((Object error, StackTrace stack) {
+        FlutterError.reportError(
```
dkwingsmt: This changes the behavior. Should we continue returning silently for `MissingPluginException` and only report the other error kinds?

packages/flutter/lib/src/widgets/platform_view.dart
```diff
@@ -806,33 +806,31 @@ class _AndroidViewState extends State<AndroidView> {
...
-            // We land the framework part of Android platform views keyboard
-            // support before the engine part. There will be a commit range where
-            // setPlatformViewClient isn't implemented in the engine. When that
-            // happens we just swallow the error here. Once the engine part is
-            // rolled to the framework I'll remove this.
-            // TODO(amirh): remove this once the engine's clearFocus is rolled.
-            return;
-          }
+        .catchError((Object error, StackTrace stack) {
+          FlutterError.reportError(
```
dkwingsmt: ditto

packages/flutter/lib/src/widgets/platform_view.dart
victorsanni: Done

dkwingsmt (approved):
LGTM

PR #182872, opened 2026-02-25
Add `UnlabaledLeafNodeEvaluation`

### LINKED ISSUE #94485: Developers should be discouraged from having unlabeled leaf a11y nodes
For example, the following should trigger a test error:

```dart
await tester.pumpWidget(
  MaterialApp(
    home: Scaffold(
      body: Column(
        children: [
          const Text(
          ),
          Text(
            '5',
            style: Theme.of(context).textTheme.headline4,
          ),
          Image.network('https://via.placeholder.com/150'),
        ],
       ),
    ),
  ),
);
```

The error should tell the developer that they should either provide a semantic label for the image, or exclude it from semantics. Otherwise screen readers may focus on the image and not have an appropriate thing to read out, which is frustrating for a11y users.

I think we should do this in `testWidgets`, with a walk of the a11y tree that checks for unlabeled/hinted/valued leaf nodes. It could be done as an a11y matcher, but ideally this is something tests would get opted into to catch errors that developers tend to miss in this regard.

/cc @goderbauer @chunhtai

QuncCccccc:
Fixes https://github.com/flutter/flutter/issues/94485

This PR is to add an `UnlabeledLeafNodeEvaluation` to check whether there is a leaf node that has no label/tooltip/hint/value.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
+++ b/packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
@@ -584,3 +584,58 @@ Iterable<Element> _collectElementsByText(Element root, String text) {
   });
   return result;
 }
+
+/// {@macro flutter.widgets.accessibility_evaluations.internal}
+///
+/// An evaluation which enforces that all leaf semantics nodes have a label,
+/// value, hint, or tooltip.
+@internal
+class UnlabeledLeafNodeEvaluation extends AccessibilityEvaluation {
+  const UnlabeledLeafNodeEvaluation();
+
+  @override
+  FutureOr<EvaluationResult> _evaluate(WidgetsBinding binding) {
+    final violations = <Violation>[];
+    for (final RenderView view in binding.renderViews) {
+      violations.addAll(_traverse(view.owner!.semanticsOwner!.rootSemanticsNode!));
+    }
+    return EvaluationResult(violations);
+  }
+
+  List<Violation> _traverse(SemanticsNode node) {
+    final violations = <Violation>[];
+    bool hasChildren = false;
+    node.visitChildren((SemanticsNode child) {
+      hasChildren = true;
+      violations.addAll(_traverse(child));
+      return true;
+    });
+
+    if (node.isMergedIntoParent ||
+        node.isInvisible ||
+        node.flagsCollection.isHidden) {
+      return violations;
+    }
+
+    // If not merging descendants and has children, it's not a leaf.
+    if (hasChildren && !node.mergeAllDescendantsIntoThisNode) {
+      return violations;
+    }
+
+    final SemanticsData data = node.getSemanticsData();
+    if (data.label.trim().isEmpty &&
+        data.value.trim().isEmpty &&
+        data.hint.trim().isEmpty &&
+        data.tooltip.trim().isEmpty) {
+      violations.add(
+        Violation(
+          node,
+          '$node: expected leaf semantics node to have a label, value, hint, or tooltip, '
+          'but none was found.',
+        ),
+      );
+    }
+
+    return violations;
+  }
+}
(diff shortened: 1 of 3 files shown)


packages/flutter/test/widgets/accessibility_evaluations_test.dart
```diff
@@ -245,4 +245,121 @@ void main() {
       handle.dispose();
     });
   });
+
+  group('UnlabeledLeafNodeEvaluation', () {
```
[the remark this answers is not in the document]
QuncCccccc: Added.

packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
```diff
@@ -610,3 +610,56 @@ Iterable<Element> _collectElementsByText(Element root, String text) {
...
+    if (hasChildren && !node.mergeAllDescendantsIntoThisNode) {
+      return violations;
+    }
+
+    final SemanticsData data = node.getSemanticsData();
+    if (data.label.trim().isEmpty &&
+        data.value.trim().isEmpty &&
+        data.hint.trim().isEmpty &&
+        data.tooltip.trim().isEmpty) {
+      violations.add(
```
chunhtai, line 691: should also check for whether this node has actions or focusable.

packages/flutter/test/widgets/accessibility_evaluations_test.dart
```diff
@@ -245,4 +245,152 @@ void main() {
...
+            ),
+          ),
+        ),
+      );
+      final EvaluationResult result = await evaluation.evaluate(tester.binding);
+      expect(result.violations, isEmpty); // Should pass (merged node has label from child)
+      handle.dispose();
+    });
+
+    testWidgets('Fails if node merges descendants but has no label', (WidgetTester tester) async {
```
chunhtai, line 399: also add some tests for case like hidden or not focusable

packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
QuncCccccc: Copied over the `_isImportantForAccessibility` to this file and added a TODO to remove that method in the future.

packages/flutter/test/widgets/accessibility_evaluations_test.dart
QuncCccccc: done!

packages/flutter/lib/src/widgets/_accessibility_evaluations.dart
```diff
@@ -610,3 +610,58 @@ Iterable<Element> _collectElementsByText(Element root, String text) {
...
+        Violation(
+          node,
+          '$node: expected leaf semantics node to have a label, value, hint, or tooltip, '
+          'but none was found.',
+        ),
+      );
+    }
+
+    return violations;
+  }
```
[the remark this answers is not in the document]
chunhtai: We prefer function without side effect in these class

packages/flutter_test/lib/src/controller.dart
```diff
@@ -334,6 +334,11 @@ class SemanticsController {
   ///
   /// * [flutter/engine/AccessibilityBridge.java#SemanticsNode.isFocusable()](https://github.com/flutter/flutter/blob/main/engine/src/flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java#L2641)
   /// * [flutter/engine/SemanticsObject.mm#SemanticsObject.isAccessibilityElement](https://github.com/flutter/flutter/blob/main/engine/src/flutter/shell/platform/darwin/ios/framework/Source/SemanticsObject.mm#L449)
+  //
+  // TODO(quncheng): If this method is modified, please also update the copy of
+  // this method located in `packages/flutter/lib/src/widgets/_accessibility_evaluations.dart`.
+  // This private method will be removed once the feature flag
+  // `isAccessibilityEvaluationsEnabled` is turned on.
```
chunhtai:
```suggestion
  // `isAccessibilityEvaluationsEnabled` is removed.
```

chunhtai (approved):
LGTM

PR #182903, opened 2026-02-25
Adding the SatelliteWindowController interface to `_window.dart`

mattkae:
## What's new?
- Added the concept of a `SatelliteWindowController` and `SatelliteWindow` widget to the `_window.dart` API
- Wrote tests for it
- Throwing "unsupported" for all existing `WindowingOwner`s when trying to create it
- Stubbed it in the example application
- Implemented a "happy" implementation in the `flutter_test` package

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/flutter/lib/src/widgets/_window_macos.dart
```diff
@@ -156,6 +156,20 @@ class WindowingOwnerMacOS extends WindowingOwner {
...
+  SatelliteWindowController createSatelliteWindowController({
+    required SatelliteWindowControllerDelegate delegate,
+    required BaseWindowController parent,
+    required Rect anchorRect,
+    required WindowPositioner positioner,
+    Size? preferredSize,
+    BoxConstraints? preferredConstraints,
+    String? title,
+  }) {
+    throw UnimplementedError('Satellite windows are not yet implemented on MacOS.');
```
loic-sharma: Silly nitpick, can we change all of these to `macOS` in this file?

```suggestion
    throw UnimplementedError('Satellite windows are not yet implemented on macOS.');
```

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  /// This field enforces a minimum and maximum size on the window. If the
+  /// user attempts to resize the window beyond these constraints, the platform
+  /// will enforce the constraints according to its own policy. For example, the
+  /// platform might clip the content to fit within the resized window, or it might
+  /// prevent the window from being resized altogether. If null, the window will
+  /// be unconstrained.
+  ///
+  /// If both [preferredSize] and [preferredConstraints] are null,
+  /// then the platform will use its own default size for the window.
+  /// {@endtemplate}
```
loic-sharma: Should this be `{@macro flutter.widgets.windowing.constraints}`? This template is already defined above.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  @protected
+  SatelliteWindowController.empty();
+
+  /// The parent controller of this satellite.
+  ///
+  /// The satellite will be destroyed if its parent is destroyed.
+  BaseWindowController get parent;
```
loic-sharma, line 1063: Looks like this is missing `@internal` + the experimental macro

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+///
+/// ** See code in examples/api/lib/widgets/windows/satellite.0.dart **
+/// {@end-tool}
+///
+/// Children of a [SatelliteWindow] widget can access the [SatelliteWindowController]
+/// via the [WindowScope] inherited widget.
+///
+/// {@macro flutter.widgets.windowing.experimental}
+@internal
+abstract class SatelliteWindowController extends BaseWindowController {
```
loic-sharma, line 985: Do you need to update `_WindowManagerState.build` to handle this new controller? 

I'm surprised it's not causing a build error since it looks like that should be an exhaustive switch:

https://github.com/canonical/flutter/blob/8b484855657f4f9c8e8135ab715466d95e8624f5/packages/flutter/lib/src/widgets/_window.dart#L2327-L2344

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  ///
+  /// The [parent] argument specifies the parent window of this satellite.
+  ///
+  /// The [anchorRect] argument specifies the rectangle in the parent's coordinate
+  /// space to which the tooltip is anchored.
+  ///
+  /// The [positioner] argument specifies how the satellite should be positioned
+  /// relative to the [anchorRect]. The positioner is only applied the first
+  /// time that the window is shown. Afterwards, the user may move and resize
+  /// the window to their preference.
```
loic-sharma: Given `positioner` and `anchorRect` are only applied when the window is first shown, should we rename these to `initialPositioner` and `initialAnchorRect`?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+/// A satellite window is an auxiliary window to a regular or dialog window.
+/// Satellite windows maintain their position relative to their parent and are
+/// hidden when the application loses focus. They are also hidden when their parent
+/// becomes fullscreen or maximized. They may be resized by the user.
+///
+/// A satellite may be reparented. For example, if an application has two documents
+/// open, the satellite may choose to reparent to the active document such that
+/// closing one document will not cause the satellite to close. This behavior
+/// may be implemented at the toolkit level. Reparenting a satellite will not
+/// change the current absolute position of the satellite.
```
loic-sharma: Just curious, why can't we change the satellite's position when it is reparented? Is that something we can't support on all platforms?

---

Previously the docs state:

> Satellite windows maintain their position relative to their parent

What happens if I reparent a satellite window, and then move the new parent? Will the satellite window maintain its position relative to its new parent? I expect that might result in funky behavior.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+    if (!isWindowingEnabled) {
+      throw UnsupportedError(_kWindowingDisabledErrorMessage);
+    }
+  }
+}
+
+/// A controller for a satellite window.
+///
+/// A satellite window is an auxiliary window to a regular or dialog window.
+/// Satellite windows maintain their position relative to their parent and are
```
loic-sharma: This statement seems at odds with the constructor's docs below:

```
  /// ... The positioner is only applied the first
  /// time that the window is shown. Afterwards, the user may move and resize
  /// the window to their preference.
```

Is one of these docs incorrect? If both are correct, I'd consider explaining how the satellite maintains its relative position without using the positioner.

Also, it looks like only the constructor explains that the user can move the window manually. I'd also call this out on the class declaration's docs too as that's what most people will read first.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -1769,6 +2121,7 @@ class WindowScope extends InheritedModel<_WindowControllerAspect> {
               DialogWindowController() => false,
               TooltipWindowController() => false,
               PopupWindowController() => false,
+              SatelliteWindowController() => false,
```
justinmc, line 2124: I think you have another switch that you need to add Satellite to, in _WindowManagerState.

examples/api/lib/widgets/windows/satellite.0.dart
```diff
@@ -0,0 +1,139 @@
...
+// ignore_for_file: invalid_use_of_internal_member
+// ignore_for_file: implementation_imports
+import 'package:flutter/material.dart';
+import 'package:flutter/src/widgets/_window.dart';
+import 'package:flutter/src/widgets/_window_positioner.dart';
+
+void main() {
+  try {
+    runWidget(
+      RegularWindow(
```
justinmc, line 17: When I try to run this, I get errors that binding is accessed before being set up, coming from instantiating RegularWindow here.

examples/api/test/widgets/windows/satellite.0_test.dart
```diff
@@ -0,0 +1,15 @@
...
+
+import 'package:flutter_api_samples/widgets/windows/satellite.0.dart'
+    as example;
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  testWidgets('Calling satellite main returns normally', (
+    WidgetTester tester,
+  ) async {
+    expect(() => example.main(), returnsNormally);
```
justinmc, line 13: Nit: I would check for the `Unsupported` text.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+///
+/// A satellite window is an auxiliary window to a regular or dialog window.
+/// Satellite windows maintain their position relative to their parent and are
+/// hidden when the application loses focus. They are also hidden when their parent
+/// becomes fullscreen or maximized. They may be resized by the user.
+///
+/// A satellite may be reparented. For example, if an application has two documents
+/// open, the satellite may choose to reparent to the active document such that
+/// closing one document will not cause the satellite to close. This behavior
+/// may be implemented at the toolkit level. Reparenting a satellite will not
```
justinmc: Nit: What do you mean by "toolkit" exactly? Consider clarifying.

packages/flutter/lib/src/widgets/_window.dart
justinmc: I ran into the same problem and left another comment. I think it is causing a build error.

examples/api/lib/widgets/windows/satellite.0.dart
```diff
@@ -0,0 +1,139 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final List<Widget> children = <Widget>[
+      ElevatedButton(
+        key: _key,
+        onPressed: () {
+          setState(() {
+            _satelliteController ??= SatelliteWindowController(
+              parent: WindowScope.of(context),
+              anchorRect: _getAnchorRect()!,
```
justinmc: Is there any noticeable frame lag using this approach? Say if the ElevatedButton is animated across the screen while the satellite window is open.

packages/flutter/lib/src/widgets/_window.dart
justinmc: Usually "initial" parameters like this worry me when they related to widgets, but in this case where they refer to the first time that the window is opened by the platform, I think that's very clear and it's ok. I like @loic-sharma's name suggestions.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  /// will enforce the constraints according to its own policy. For example, the
+  /// platform might clip the content to fit within the resized window, or it might
+  /// prevent the window from being resized altogether. If null, the window will
+  /// be unconstrained.
+  ///
+  /// If both [preferredSize] and [preferredConstraints] are null,
+  /// then the platform will use its own default size for the window.
+  /// {@endtemplate}
+  ///
+  /// The [title] argument configures the window's initial title.
```
justinmc: Why is the word "initial" used here, can the title change over time?

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  SatelliteWindowController.empty();
+
+  /// The parent controller of this satellite.
+  ///
+  /// The satellite will be destroyed if its parent is destroyed.
+  BaseWindowController get parent;
+
+  /// The current title of the window.
+  ///
+  /// This might differ from the requested title.
```
justinmc: If I understand correctly, maybe reword to: "The title shown in the window is controlled by the platform and may differ from this given `title`."

justinmc (commented):
* I can't get the example to run, see my comments below.

 * I see this is beyond the stuff that you guys decided to call [MVP in your doc](https://docs.google.com/document/d/1eQG-IS7r4_S9_h50MY_hSGwUVtgSTRSLzu7MLzPNd2Y/edit?tab=t.0#heading=h.4y3d6rrizqfa), are there any implications for pre/post MVP work and our plans for shipping it in general?

 * Overall this PR looks fine but there's one thing that I want to think through. Let's say I'm building an app on only one platform, say macOS. SatelliteWindow uses [NSPanel](https://developer.apple.com/documentation/appkit/nspanel) under the hood on macOS, which looks like it takes 3 boolean parameters (isFloatingPanel, becomesKeyOnlyIfNeeded, and worksWhenModal). Let's say I need to individually set one of those booleans for my use case. What are my options? How much of the multiwindow code will be unusable for me and how much can I still manage to use?

packages/flutter/lib/src/widgets/_window.dart
mattkae: Whoops! Yes

packages/flutter/lib/src/widgets/_window.dart
mattkae: I wonder why it wasn't complaining! I will add it

packages/flutter/lib/src/widgets/_window.dart
mattkae: We can change the satellite's position, but we figured that would be too jarring for users.

The best example is a word processing application. Imagine you have two windows open each with their own document, but sharing the same satellite window toolbar. When one document is closed, you would expect the satellite toolbar to remain open in the same position instead of snapping to a new anchor position next to the other document.

In this case, the satellite window will now maintain its position relative to the remaining document.

packages/flutter/lib/src/widgets/_window.dart
mattkae: Both are correct :)

1. The user may move and resize the satellite independently
2. When the parent moves, the satellite is moved relative to it

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,248 @@ abstract class PopupWindowController extends BaseWindowController {
...
+  ///
+  /// The [delegate] argument can be used to listen to the window's
+  /// lifecycle. For example, it can be used to save state before
+  /// a window is closed.
+  ///
+  /// {@macro flutter.widgets.windowing.experimental}
+  @internal
+  factory SatelliteWindowController({
+    required BaseWindowController parent,
+    required Rect anchorRect,
```
knopp: Should the `anchorRect` be nullable? I think per spec `null` anchor rect means anchoring the satallite to the outer window frame (as it is implemented in my prototype).

packages/flutter/lib/src/widgets/_window.dart
mattkae: Bad merge, fixed :+1:

examples/api/lib/widgets/windows/satellite.0.dart
mattkae: Keep in mind that there is no implementation of satellites yet, so this will not work

packages/flutter/lib/src/widgets/_window.dart
mattkae: I meant "library", I will clarify

examples/api/lib/widgets/windows/satellite.0.dart
mattkae: This is the approach that we've used for popups and tooltips, so I just went with it here.

We haven't seen a huge performance cost in the example app, but perhaps it is possible to incur one. Is there a better approach here?

packages/flutter/lib/src/widgets/_window.dart
mattkae: I agree that "initial" is useless here

examples/api/lib/widgets/windows/satellite.0.dart
knopp: Are you referring to how well position tracker can keep up with the coordinates? I think there shouldn't be any major lag - you can try that when resizing the window with popover open.

mattkae: We can leave it `null` for now, because null means "use the whole parent window"

examples/api/lib/widgets/windows/satellite.0.dart
justinmc: Ah of course, I thought I had just missed that PR 😅

examples/api/lib/widgets/windows/satellite.0.dart
justinmc: I just tried the example using tooltip and resizing the window and @knopp is right that there is no noticeable lag, so I'm good with this either way. They appear to be using the same approach.

https://github.com/flutter/flutter/blob/9db591bda44a730d22e3c518da71e13615fd7ac2/examples/multiple_windows/lib/app/tooltip_button.dart#L65

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,241 @@ abstract class PopupWindowController extends BaseWindowController {
...
+/// satellite.
+///
+/// Upon construction, the window is created for the platform with the provided
+/// properties.
+///
+/// This class does not interact with the widget tree. Instead, it is typically
+/// provided to the [SatelliteWindow] widget, who does the work of rendering the
+/// content inside of this window.
+///
+/// The user of this class is responsible for managing the lifecycle of the window.
```
loic-sharma, line 963: I'd also add a sentence that explains how destroying the parent `BaseWindowController` affects this `SatelliteWindowController`.

It'd likely be worth adding a similar comment to dialog, tooltip, and popups window controllers too.

mattkae: I added it just to satellite for now.

I am going to revise some of the documentation for the other windows after this, as satellites have brought up a bunch of nits that we should address there.

packages/flutter/lib/src/widgets/_window.dart
```diff
@@ -890,6 +890,245 @@ abstract class PopupWindowController extends BaseWindowController {
...
+/// When the window is no longer needed, the user should call [destroy] on this
+/// controller to release the resources associated with the window.
+///
+/// If the parent window of the satellite is destroyed, then the satellite will
+/// be destroyed as well. [SatelliteWindowControllerDelegate.onWindowDestroyed]
```
justinmc: Does the developer still need to call `destroy` on the satellite window when it or its parent is closed by the platform? (I mean say by the use clicking the X in the window)

justinmc (approved):
LGTM 👍 

Thanks for the explanations of the API here @mattkae and @knopp, I think the problem is a lot smaller than I was assuming. As discussed offline I think the best way to prove this is to include an example of the use of something like the raw NSWindow* used alongside the rest of the usual multiwindow infrastructure like RegularWindow etc. But that can wait for another PR, once we have exposed the raw window handles.

Landing this PR won't delay the removal of the feature flag, will it? I just want to make sure that adding satellites in the framework isn't going to delay everything else by expanding the scope of the engine work.

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: SGTM, thanks!

loic-sharma (approved):
LGTM!

packages/flutter/lib/src/widgets/_window.dart
loic-sharma: Gotcha, that makes sense.

Would it make sense to add a `move({required Rect anchorRect, required WindowPositioner positioner})` method to `SatelliteWindowController`? This would update the satellite window position once. I imagine this would be useful in reparenting scenarios if you do want this slightly jarring experience. I'm OK punting this until later, just a thought.

packages/flutter/lib/src/widgets/_window.dart
mattkae: Nope. I can call that out :)

packages/flutter/lib/src/widgets/_window.dart
mattkae: Yeah let's punt it for now, and see if it comes up. Intuition tells me that the user probably won't want to do that, but who knows. We have that type of method for tooltips and popups.

Issue #179452, opened 2025-12-04
SceneDelegate adoption and dynamic links on cold start


Y-Smirnov:
### Steps to reproduce

Hello Flutter Team,

Dynamic links (universal links) seem not to work as expected on cold start when adopting **SceneDelegate**.

Testing steps:
- Create new Flutter app (no extra plugins), new app comes without **SceneDelegate**
- Open the app in Xcode, setup  associated domains
- Ensure at this point everything is working as expected, cold and warm starts by testing with the **AppDelegate** method:
```
override func application(
      _ application: UIApplication,
      continue userActivity: NSUserActivity,
      restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void
    ) -> Bool
```

- Update **AppDelegate** and **Info.plist** as described [here](https://docs.flutter.dev/release/breaking-changes/uiscenedelegate#migration-guide-for-flutter-apps) to adopt UISceneDelegate
- Create **SceneDelegate.swift** and update **Info.plist** accordingly
- Add method to scene delegate:
`override func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {}`

### Expected results

- The scene(_:continue userActivity) method should be invoked when the app is opened from a dynamic link on cold start, as it did with application(_:continue userActivity).
- Plugins that rely on dynamic links and have not adopted SceneDelegate should still receive the fallback to the AppDelegate method

### Actual results

- The scene(_:continue userActivity) method is invoked only when the app is already running
- Plugins that have not yet adopted **SceneDelegate** also do not receive fallback to the AppDelegate method on cold start

### Code sample

<details open><summary>Code sample</summary>
</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.38.3, on macOS 15.6.1 24G90 darwin-arm64, locale en-PT) [3.8s]
    • Flutter version 3.38.3 on channel stable at /Users/yevheniismirnov/Library/Flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 19074d12f7 (2 weeks ago), 2025-11-20 17:53:13 -0500
    • Engine revision 13e658725d
    • Dart version 3.10.1
    • DevTools version 2.51.1
[...]

PR #182904, opened 2026-02-25
User activitiy fallbacks for unmigrated plugins

LongCatIsLooong:
- Add cold-start continuing user activity / shortcut fallbacks for unmigrated plugins in apps that support UIScene (#179452)
~- ~Makes sure `applicationWillFinishLaunching` / `applicationDidFinishLaunching` are only sent at most once. In the event that we need to forward `sceneWillConnect` connection options to unmigrated plugins but `applicationDidFinishLaunching` has already been called, the method will not be called again, and the connection option fallback will kick in regardless of the (previous) return value of `applicationDidFinishLaunching` (as if it was not cold-start, since in pre-scene world `application:performActionForShortcutItem:completionHandler:` is called for both cold-start and resume).~

## Pre-launch Checklist

- [x] I signed the [CLA].


engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPlugin.h
```diff
@@ -37,6 +40,9 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Called if this has been registered for `UIApplicationDelegate` callbacks.
  *
+ * This method is only called once per application. Plugins registered later than
```
LongCatIsLooong: I guess here it should be at most once since it's possible that we only call didFinishLaunching (which happens in the sceneWillConnect fallback). But that is only supposed to happen to unmigrated plugins.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
                isFallbackForScene:(BOOL)isFallback {
+  if (_state >= kAppLifecycleDidLaunchSent) {
+    return YES;
+  }
   for (NSObject<FlutterApplicationLifeCycleDelegate>* delegate in _delegates) {
     if (!delegate || (isFallback && [self pluginSupportsSceneLifecycle:delegate])) {
       continue;
     }
     if ([delegate respondsToSelector:@selector(application:didFinishLaunchingWithOptions:)]) {
+      _state = kAppLifecycleDidLaunchSent;
```
LongCatIsLooong: Unfortunately if this method is called with `isFallback: YES` and then `isFallback: NO`, migrated plugins may never receive the `didFinishLaunching` event because we promised that this event will only be sent once. 

Does it make sense to ignore `isFallback` and always send the event to every delegate?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -30,16 +41,19 @@ - (void)handleDidBecomeActive:(NSNotification*)notification
...
-@property(nonatomic, assign) BOOL didForwardApplicationWillLaunch;
-@property(nonatomic, assign) BOOL didForwardApplicationDidLaunch;
 @end
 
 @implementation FlutterPluginAppLifeCycleDelegate {
   UIBackgroundTaskIdentifier _debugBackgroundTask;
 
   // Weak references to registered plugins.
   NSPointerArray* _delegates;
+  // Wether this FlutterPluginAppLifeCycleDelegate class has sent
```
vashworth:
```suggestion
  // Whether this FlutterPluginAppLifeCycleDelegate class has sent
```

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -19,6 +19,17 @@
...
 
+typedef NS_ENUM(NSInteger, FlutterPluginAppLifeCycleDelegateState) {
+  /// The FlutterPluginAppLifeCycleDelegate was just initialized and neither
+  /// willFinishLaunching or didFinishLaunching has been sent to the plugins.
+  kAppLifecycleInitialized,
+  /// The willFinishLaunching event has been sent to the plugins
+  /// (but not didFinishLaunching).
+  kAppLifecycleWillLaunchSent,
+  /// The didFinishLaunching event has been sent to the plugins.
+  kAppLifecycleDidLaunchSent,
```
hellohuanlin: 1. in standard objc naming convention these enum cases should be prefixed with the type name rather than `k`;
2. since you are comparing the raw value in your code (e.g. `_state < kAppLifecycleDidLaunchSent`), it's more idiomatic to explicitly assign values to each case (e.g. `= 0`, `= 1`). This reduces the risk of enum members being reordered in the future.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -119,35 +134,17 @@ - (void)addDelegate:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
 }
 
 - (void)sceneFallbackDidFinishLaunchingApplication:(UIApplication*)application {
-  // If the application:didFinishingLaunchingWithOptions: event has already been sent to plugins, do
-  // not send again.
-  if (self.didForwardApplicationDidLaunch) {
```
hellohuanlin, line 124: why is this check removed? this seems to be an important check

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -119,35 +134,17 @@ - (void)addDelegate:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
...
 
 - (void)sceneFallbackWillFinishLaunchingApplication:(UIApplication*)application {
-  // If the application:willFinishLaunchingWithOptions: event has already been sent to plugins, do
-  // not send again.
-  if (self.didForwardApplicationWillLaunch) {
-    return;
-  }
-  // If the application:didFinishingLaunchingWithOptions: event has already been sent to plugins, do
-  // not send willFinishLaunchingWithOptions since it should happen before, not after.
-  if (self.didForwardApplicationDidLaunch) {
```
hellohuanlin, line 139: same here

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -119,35 +134,17 @@ - (void)addDelegate:(NSObject<FlutterApplicationLifeCycleDelegate>*)delegate {
...
-    return;
-  }
-  // Send nil launchOptions since UIKit sends nil when UIScene is enabled.
+  // Send empty launchOptions since UIKit sends nil when UIScene is enabled.
   [self application:application willFinishLaunchingWithOptions:@{}];
 }
 
 - (BOOL)application:(UIApplication*)application
     didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
-  if (_delegates.count > 0) {
```
hellohuanlin: why `_delegates.count` check is removed here?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
     return NO;
   }
-  if (![self application:application
+
+  // Send appDidFinishLaunching, but only if we've never sent it.
+  //
+  // If there is an unmigrated plugin returned NO, do not attempt continuing
+  // user activities. This is based on the documented behavior of the
+  // (now deprecated)
+  // -[UIApplication application:continueUserActivity:restorationHandler:]
```
hellohuanlin: > This is based on the documented behavior of the (now deprecated) [UIApplication application:continueUserActivity:restorationHandler:]

can you elaborate what "the documented behavior" is, and how it is related to "If there is an unmigrated plugin returned NO, do not attempt continuing..."

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
+  // method.
+  if (_state < kAppLifecycleDidLaunchSent &&
+      ![self application:application
           didFinishLaunchingWithOptions:convertedLaunchOptions
                      isFallbackForScene:YES]) {
-    return YES;
+    return NO;
```
hellohuanlin: what does the return value of `sceneWillConnectFallback` mean?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
+
+  for (NSUserActivity* activity in connectionOptions.userActivities) {
+    if ([self application:application
+            continueUserActivity:activity
+              restorationHandler:nil
+              isFallbackForScene:YES]) {
+      return YES;
+    }
+  }
+  return shortcutHandled;
```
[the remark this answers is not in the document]
hellohuanlin: this seems to be a valid concern?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
+                                                          // synchronously whether they consume the
+                                                          // item.
+                                                        }
+                                                       isFallbackForScene:YES];
+
+  for (NSUserActivity* activity in connectionOptions.userActivities) {
+    if ([self application:application
+            continueUserActivity:activity
+              restorationHandler:nil
+              isFallbackForScene:YES]) {
```
hellohuanlin, line 193: to double check if i understand the problem here, this is to call `continueUserActivity` method on legacy plugins that haven't migrated to UIScene?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -160,26 +157,59 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
+      return YES;
+    }
+  }
+  return shortcutHandled;
 }
 
 - (BOOL)application:(UIApplication*)application
     didFinishLaunchingWithOptions:(NSDictionary*)launchOptions
                isFallbackForScene:(BOOL)isFallback {
```
hellohuanlin, line 202: Does `isFallbackForScene` mean that for un-migrated old plugins in a already migrated app, we call methods like `did/willFinishLaunching` originated from corresponding scene callbacks (e.g. sceneWillConnect)

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
hellohuanlin: >  if this method is called with isFallback: YES and then isFallback: NO

how will this happen?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegateTest.mm
```diff
@@ -150,6 +150,125 @@ - (void)testSceneWillConnectFallbackSkippedNoOptions {
...
+                         completionHandler:OCMArg.any])
+      .andReturn(YES);
+
+  [delegate sceneWillConnectFallback:mockOptions];
+  OCMVerify(times(1), [mockPlugin application:UIApplication.sharedApplication
+                          didFinishLaunchingWithOptions:expectedApplicationOptions]);
+
+  OCMVerify(times(1), [mockPlugin application:UIApplication.sharedApplication
+                          continueUserActivity:mockUserActivity
+                            restorationHandler:OCMArg.any]);
```
hellohuanlin, line 236: to check if i understand: the original behavior calls `didFinishLaunching` twice, and the new behavior after your change is to call `didFinishLaunching` once, then followed by `continueUserActivity` once?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: IIRC this is method is called by `willConnectToSession` and the return value is supposed to mean whether the event is used by a plugin.  

And the return value of `didFinishLaunchingWithOptions` indicates whether the app can handle the user activity (https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:didfinishlaunchingwithoptions:))

But for plugins I assume `didFinishLaunchingWithOptions` means "do not send this event to other plugins"? I can't find the answer in the documentation. Maybe @vashworth knows?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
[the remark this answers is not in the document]
LongCatIsLooong: This mirrors the current implementation for handling deep links in FlutterSceneLifeCycle.mm.

```objc
  for (NSUserActivity* userActivity in connectionOptions.userActivities) {
    if ([self handleDeeplink:userActivity.webpageURL
                       flutterEngine:engine
            relayToSystemIfUnhandled:YES]) {
      return;
    }
  }
```

Although I don't know the reason for skipping the rest of activities if one activity gets handled.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegateTest.mm
LongCatIsLooong: That was correct. I updated the implementation so `didFinishLaunching` will still get called one than once.

LongCatIsLooong:
Reverted the logic that tried to prevent `didFinishLaunching` from being called more than once due to https://github.com/flutter/flutter/pull/182904#issuecomment-4015679948

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
```diff
@@ -164,12 +164,35 @@ - (BOOL)sceneWillConnectFallback:(UISceneConnectionOptions*)connectionOptions {
...
+                                                       isFallbackForScene:YES];
+
+  for (NSUserActivity* activity in connectionOptions.userActivities) {
+    if ([self application:application
+            continueUserActivity:activity
+              restorationHandler:nil
+              isFallbackForScene:YES]) {
+      return YES;
+    }
+  }
```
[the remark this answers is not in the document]
LongCatIsLooong: See the [comment](https://github.com/flutter/flutter/pull/182904#discussion_r3089169193) above.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: I think it's the other way around. If the plugin is not migrated but the app is, this translates the scene event to the closest application equivalent and sends that to the plugin.

vashworth (dismissed):

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: wait I can't find this line any more, I think this is from an earlier commit and the typo no longer exists.

LongCatIsLooong:
Hi @hellohuanlin I reverted the changes the for `didFinishLaunching` because it broke a plugin: https://github.com/flutter/flutter/pull/182904#issuecomment-4283216394. Would you like to take another look?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
hellohuanlin: i assume the answer is yes here?

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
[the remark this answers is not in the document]
hellohuanlin: Can you add a comment there with your research, for the sake of future readers who will likely wonder the same

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
hellohuanlin: aren't we talking about the same? (confused here)

hellohuanlin (commented):
Can you add an integration test?

LongCatIsLooong:
Ok I tried to add an integration test but it takes more than 60 seconds to run because "com.apple.springboard" keeps animating itself (and `tap` will wait for the springboard to idle and timeout after 60s). The solution seems to be using `shouldWaitForQuiescence = false`. The API is private so I'd prefer to not use it even it's just in tests. The existing unit test should cover the bug reported in the original issue and the fallbacks should be temporary anyway.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: Yes, app migrated, plugin unmigrated.

engine/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm
LongCatIsLooong: Oh I misread your comment. Yeah this argument indicates whether the caller (the app) is calling this method as a fallback to a scene event.

hellohuanlin (approved):

PR #182913, opened 2026-02-25
Exclude arm64 if any dependencies do and print warning when using Xcode 26

### LINKED ISSUE #182748: [iOS] - Framework 'Pods_Runner' not found when targeting simulator with dependency that does not support arm64 simulators
### Steps to reproduce

Hi, I've just updated flutter to 3.41.1 but an error is showing now when launching on IOS.

It was running fine before the update

### Actual results

It should work as before

### Logs

```
Error (Xcode): Framework 'Pods_Runner' not found

Error (Xcode): Linker command failed with exit code 1 (use -v to see invocation)

Could not build the application for the simulator.
Error launching application on iPhone 16.
```

### Flutter Doctor output

<details open>
<summary>Doctor output</summary>

```console

[✓] Flutter (Channel stable, 3.41.1, on macOS 26.2 25C56 darwin-arm64, locale it-IT) [238ms]
    • Flutter version 3.41.1 on channel stable at /Users/sasoftware/development/flutter
    • Framework revision 582a0e7c55 (10 days ago), 2026-02-12 17:12:32 -0800
    • Engine revision 3452d735bd
    • Dart version 3.11.0
    • DevTools version 2.54.1
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [2,8s]
    • Android SDK at /Users/sasoftware/Library/Android/sdk
    • Platform android-36, build-tools 35.0.0
    • ANDROID_HOME = /Users/sasoftware/Library/Android/sdk
    • Java binary at: /Users/sasoftware/Library/Java/JavaVirtualMachines/jbr-17.0.12/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment JBR-17.0.12+1-1207.37-nomod (build 17.0.12+1-b1207.37)

[✓] Xcode - develop for iOS and macOS (Xcode 26.2) [959ms]
    • Build 17C52
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [4ms]

[✓] Connected device (3 available) [9,1s]
    • iPhone 16 (mobile) • 0845E509-267D-4F1B-A233-20E3DC802914 • ios            • com.apple.CoreSimulator.SimRuntime.iOS-18-1 (simulator)
    • macOS (desktop)    • macos                                • darwin-arm64   • macOS 26.2 25C56 darwin-arm64
    • Chrome (web)       • chrome                               • web-javascript • Google Chrome 145.0.7632.109

[✓] Network resources [239ms]

• No issues found!

```

</details>

vashworth:
If any Pod dependencies exclude the arm64 architecture for simulators, Flutter excludes it too. This will force Xcode to build it in an x86 architecture. However, iOS 26 introduced arm-only simulators. So when using Xcode 26, print a warning with the list of plugins/targets that are excluding arm and direct them to request arm support. Then if the build fails due to targeting an arm-only simulator, give a guided message to switch to an x86 simulator.

Fixes https://github.com/flutter/flutter/issues/182748.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_tools/lib/src/ios/mac.dart
+++ b/packages/flutter_tools/lib/src/ios/mac.dart
@@ -336,6 +336,7 @@ Future<XcodeBuildResult> buildXcodeProject({
     project: project,
     targetOverride: targetOverride,
     buildInfo: buildInfo,
+    printWarnings: true,
   );
   if (app.project.usesSwiftPackageManager) {
     final String? iosDeploymentTarget = buildSettings['IPHONEOS_DEPLOYMENT_TARGET'];
@@ -1019,6 +1020,7 @@ _XCResultIssueHandlingResult _handleXCResultIssue({
         'Unable to find a destination matching the provided destination specifier',
       ) &&
       message.contains('platform:iOS Simulator, arch:x86_64,') &&
+      !message.contains('platform:iOS Simulator, arch:arm64,') &&
       !message.contains(
         'The requested device could not be found because no available devices matched the request.',
       )) {
@@ -1175,7 +1177,7 @@ Future<bool> _simulatorSupportsIntel(Device device) async {
     return true;
   }
   final String runtime = await device.sdkNameAndVersion;
-  final RunResult result = globals.processUtils.runSync([
+  final RunResult result = await globals.processUtils.run([
     ...globals.xcode!.xcrunCommand(),
     'simctl',
     'list',
--- a/packages/flutter_tools/lib/src/ios/mac.dart
+++ b/packages/flutter_tools/lib/src/ios/mac.dart
@@ -796,6 +796,7 @@ Future<void> diagnoseXcodeBuildFailure(
   required FileSystem fileSystem,
   required FlutterDarwinPlatform platform,
   required FlutterProject project,
+  Device? device,
 }) async {
   final XcodeBuildExecution? xcodeBuildExecution = result.xcodeBuildExecution;
   if (xcodeBuildExecution != null &&
@@ -824,6 +825,7 @@ Future<void> diagnoseXcodeBuildFailure(
     platform: platform,
     logger: logger,
     fileSystem: fileSystem,
+    device: device,
   );
 
   if (!issueDetected && xcodeBuildExecution != null) {
@@ -1013,6 +1015,18 @@ _XCResultIssueHandlingResult _handleXCResultIssue({
       hasProvisioningProfileIssue: false,
       modifiedPrecompiledSource: true,
     );
+  } else if (message.contains(
+        'Unable to find a destination matching the provided destination specifier',
+      ) &&
+      message.contains('platform:iOS Simulator, arch:x86_64,') &&
+      !message.contains(
+        'The requested device could not be found because no available devices matched the request.',
+      )) {
+    return _XCResultIssueHandlingResult(
+      requiresProvisioningProfile: false,
+      hasProvisioningProfileIssue: false,
+      unableToFindArmDestination: true,
+    );
   }
   return _XCResultIssueHandlingResult(
     requiresProvisioningProfile: false,
@@ -1028,11 +1042,13 @@ Future<bool> _handleIssues(
   required FlutterDarwinPlatform platform,
   required Logger logger,
   required FileSystem fileSystem,
+  Device? device,
 }) async {
   var requiresProvisioningProfile = false;
   var hasProvisioningProfileIssue = false;
   var issueDetected = false;
   var modifiedPrecompiledSource = false;
+  var unableToFindArmDestination = false;
   String? missingPlatform;
   final duplicateModules = <String>[];
   final missingModules = <String>[];
@@ -1059,6 +1075,7 @@ Future<bool> _handleIssues(
         missingModules.add(handlingResult.missingModule!);
       }
       modifiedPrecompiledSource = handlingResult.modifiedPrecompiledSource;
+      unableToFindArmDestination = handlingResult.unableToFindArmDestination;
       issueDetected = true;
     }
   } else if (xcResult != null) {
@@ -1133,8 +1150,40 @@ Future<bool> _handleIssues(
       'the cache.\n'
       '════════════════════════════════════════════════════════════════════════════════',
     );
+  } else if (unableToFindArmDestination &&
+      xcodeBuildExecution != null &&
+      xcodeBuildExecution.environmentType == EnvironmentType.simulator &&
+      device != null) {
+    final bool simulatorSupportsIntel = await _simulatorSupportsIntel(device);
+    if (!simulatorSupportsIntel) {
+      logger.printError(
+        '════════════════════════════════════════════════════════════════════════════════\n'
+        'The selected simulator is incompatible with the current build settings.\n'
+        'Please use a simulator that supports x86_64, such as a simulator prior to iOS 26 or '
+        'download the universal variant of the iOS 26 simulator using '
+        '"xcodebuild -downloadPlatform iOS -architectureVariant universal".\n'
+        '════════════════════════════════════════════════════════════════════════════════',
+      );
+    }
   }
   return issueDetected;
+}
+
+Future<bool> _simulatorSupportsIntel(Device device) async {
+  final Version? xcodeVersion = globals.xcode?.currentVersion;
+  if (xcodeVersion != null && xcodeVersion.major < 26) {
+    return true;
+  }
+  final String runtime = await device.sdkNameAndVersion;
+  final RunResult result = globals.processUtils.runSync([
+    ...globals.xcode!.xcrunCommand(),
+    'simctl',
+    'list',
+    'runtimes',
+    runtime,
+    '--json',
+  ]);
+  return result.stdout.contains('x86_64');
 }
 
 /// Returns true if a Package.swift is found for the plugin and a podspec is not.
@@ -1271,6 +1320,7 @@ class _XCResultIssueHandlingResult {
     this.duplicateModule,
     this.missingModule,
     this.modifiedPrecompiledSource = false,
+    this.unableToFindArmDestination = false,
   });
 
   /// An issue indicates that user didn't provide the provisioning profile.
@@ -1292,6 +1342,8 @@ class _XCResultIssueHandlingResult {
   /// An issue indicates that a source file, such as a header in the Flutter framework, has
   /// changed since last built. This requires "flutter clean" to resolve.
   final bool modifiedPrecompiledSource;
+
+  final bool unableToFindArmDestination;
 }
 
 const _kResultBundlePath = 'temporary_xcresult_bundle';
--- a/packages/flutter_tools/lib/src/xcode_project.dart
+++ b/packages/flutter_tools/lib/src/xcode_project.dart
@@ -5,6 +5,8 @@
 /// @docImport 'ios/mac.dart';
 library;
 
+import 'package:yaml/yaml.dart' as yaml;
+
 import 'base/error_handling_io.dart';
 import 'base/file_system.dart';
 import 'base/logger.dart';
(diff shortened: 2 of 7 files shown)


packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1013,6 +1015,18 @@ _XCResultIssueHandlingResult _handleXCResultIssue({
...
+        'Unable to find a destination matching the provided destination specifier',
+      ) &&
+      message.contains('platform:iOS Simulator, arch:x86_64,') &&
+      !message.contains(
+        'The requested device could not be found because no available devices matched the request.',
+      )) {
+    return _XCResultIssueHandlingResult(
+      requiresProvisioningProfile: false,
+      hasProvisioningProfileIssue: false,
+      unableToFindArmDestination: true,
```
hellohuanlin, line 1030: to verify that i understand what it is doing - this is saying unable to find arm destination **in the list of destinations supported by the scheme**, right? (rather than **... in the list of available simulators**).

packages/flutter_tools/lib/src/ios/mac.dart
```diff
@@ -1013,6 +1015,18 @@ _XCResultIssueHandlingResult _handleXCResultIssue({
       hasProvisioningProfileIssue: false,
       modifiedPrecompiledSource: true,
     );
+  } else if (message.contains(
+        'Unable to find a destination matching the provided destination specifier',
+      ) &&
+      message.contains('platform:iOS Simulator, arch:x86_64,') &&
```
hellohuanlin, line 1022: should this also check the message does not contain `platform:iOS Simulator, arch:arm64,`?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+      // Xcode isn't installed, don't try to check.
+      return true;
     }
     final String? buildSettings = await xcodeProjectInterpreter.pluginsBuildSettingsOutput(
       podXcodeProject,
     );
-
     if (buildSettings == null || buildSettings.isEmpty) {
-      return;
+      return false;
```
hellohuanlin: why does this return `false` while the previous 2 return `true`

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+            })
+            .join('\n');
+        globals.logger.printWarning(
+          'The following target(s) do not support arm64 architecture, which is a requirement for '
+          'Apple Silicon iOS 26+ simulators:\n'
+          '$list\n\n'
+          'Please contact plugin maintainers to request arm64 support to continue to be able to '
+          'use the plugin on a simulator.',
+        );
+        _warnedAboutExcludingArm = true;
```
hellohuanlin: why do we need this check? In what scenario will this function be called multiple times?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+      } else {
+        // If it's not a plugin, it may de a transitive dependency of a plugin
+        final String? parentPlugin = _pluginUsingDependency(
+          targetName: currentTarget,
+          pluginNames: pluginNames.toList(),
+          cocoapodsTree: cocoapodsTree,
+        );
+        if (parentPlugin != null) {
+          foundTargets.add((target: currentTarget, plugin: parentPlugin));
+        } else if (currentTarget.startsWith('Pods-')) {
```
hellohuanlin, line 592: what does this case mean?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+        _warnedAboutExcludingArm = true;
+      }
     }
 
+    return !buildSettings.contains(RegExp('EXCLUDED_ARCHS.*arm64'));
+  }
+
+  /// Returns a list of targets and their associated plugin (if found) that exclude arm64 architecture.
+  Future<List<({String target, String? plugin})>> _targetsExcludingArm(String buildSettings) async {
+    final Map<String, List<String>> cocoapodsTree = _cocoapodTree();
```
hellohuanlin: is it only a problem with cocoapods and not swiftpm?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+    required String targetName,
+    required List<String> pluginNames,
+    required Map<String, List<String>> cocoapodsTree,
+  }) {
+    final String? pluginName = _findPluginForDependency(targetName, cocoapodsTree, pluginNames);
+    if (pluginName != null) {
+      return pluginName;
+    }
+    if (targetName.contains('-')) {
+      // If the target has a - in the name, the first part is often the pod name
```
hellohuanlin: can you give an example of this? and why one search by prefix and the other does not?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+    return foundTargets;
+  }
 
-    if (pluginsExcludingArmArch.isNotEmpty) {
-      final String list = pluginsExcludingArmArch.map((String n) => '  - $n').join('\n');
+  /// Returns the plugin that uses the given target.
+  String? _pluginUsingDependency({
+    required String targetName,
+    required List<String> pluginNames,
+    required Map<String, List<String>> cocoapodsTree,
```
hellohuanlin: it's a dependency graph (not tree) right?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+          if (podsList is List<dynamic>) {
+            // ignore: specify_nonobvious_local_variable_types
+            for (final dynamic pod in podsList) {
+              if (pod is yaml.YamlMap) {
+                final name = pod.keys.first as String;
+                final dynamic value = pod.value[name];
+                if (value is yaml.YamlList) {
+                  podDependencies[name.split(' ').first] = value
+                      .map((dep) => (dep as String).split(' ').first)
+                      .toList();
```
hellohuanlin, line 655: why is it only getting the first token of the name?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+          (dep) => dep.startsWith('$transitiveDependency/'),
+        );
+      } else {
+        podHasTransitiveDependency = podDependencies.contains(transitiveDependency);
+      }
+      if (podHasTransitiveDependency) {
+        if (pluginNames.contains(podName)) {
+          return podName;
+        }
+        return _findPluginForDependency(podName, cocoapodTree, pluginNames);
```
hellohuanlin, line 689: should this pass in `searchByPrefix` into the recursion?

hellohuanlin (commented):
The overall logic makes sense. Just a few questions since I am not familiar with the details. 

@okorohelijah should also review since you are more familiar

packages/flutter_tools/lib/src/xcode_project.dart
okorohelijah: if the `buildsetting` is null or empty, shouldnt the error message be different?

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
         continue;
       }
 
-      final Iterable<String> tokens = settingsLine.split(' ');
-      if (tokens.contains('arm64')) {
-        pluginsExcludingArmArch.add(currentTarget);
+      if (pluginNames.contains(currentTarget)) {
+        foundTargets.add((target: currentTarget, plugin: currentTarget));
+      } else {
+        // If it's not a plugin, it may de a transitive dependency of a plugin
```
okorohelijah:
```suggestion
        // If it's not a plugin, it may be a transitive dependency of a plugin
```

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,77 +491,191 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+        }
       }
     }
+    return foundTargets;
+  }
 
-    if (pluginsExcludingArmArch.isNotEmpty) {
-      final String list = pluginsExcludingArmArch.map((String n) => '  - $n').join('\n');
+  /// Returns the plugin that uses the given target.
+  String? _pluginUsingDependency({
```
okorohelijah, line 604: what happens if a dependency that doesnt support ARM is added via spm instead of cocoapods?

okorohelijah (changes requested):

packages/flutter_tools/lib/src/ios/mac.dart
vashworth: This should be sufficient because Xcode does not support x86_64-only simulators anymore, but it's easy to add so sure

packages/flutter_tools/lib/src/ios/mac.dart
vashworth: Correct, the list of destinations provided by the scheme

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: This is how it was originally, so was just restoring that:

https://github.com/flutter/flutter/blob/2383b644984a4a15595978e0e4be11a2c407dbf2/packages/flutter_tools/lib/src/xcode_project.dart#L501

Considering that Xcode is moving toward not supporting x86, we can probably change this to default to true.

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: Yes, this is called within `updateGeneratedXcodeProperties`, which can be called multiple times

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: Yes. SwiftPM does not allow you to exclude architectures

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: When a project as CocoaPods integrated, there are a couple Pod targets: `Pods-Runner`, `Pods-RunnerTests`. The Pod targets are generated based on the pod dependencies and therefore the user should not change them.

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: SwiftPM does not allow you to exclude arches, so there is not way to verify. It will be caught by `diagnoseXcodeBuildFailure` in mac.dart

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: Good point, I'll rename

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: Resource bundles are prefixed with the name of the pod and then the name of the resource

For example, the target for this pod resource is named `PromisesObjC-FBLPromises_Privacy`: https://github.com/google/promises/blob/d6d659a6cbdd4108788781d178a72fd34ce07e6e/PromisesObjC.podspec#L32

vashworth: I'll add a comment

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: It also contains the version info sometimes e.g. `GoogleMLKit/BarcodeScanning (~> 9.0.0)`

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: No, it specifically should not. After the initial prefix match, it should match by exact name since it's a valid pod name now

packages/flutter_tools/lib/src/xcode_project.dart
vashworth: There's nothing really actionable for the user to do, but I added a trace log

okorohelijah (approved):
LGTM👍🏾

packages/flutter_tools/lib/src/xcode_project.dart
hellohuanlin: gotcha. worth adding a comment there since it isn't clear when reading it. probably too late so feel free to ignore.

packages/flutter_tools/lib/src/xcode_project.dart
```diff
@@ -489,79 +491,207 @@ def __lldb_init_module(debugger: lldb.SBDebugger, _):
...
+      globals.logger.printTrace('Failed to parse podfile.lock: $e');
+    }
+
+    return podDependencies;
+  }
+
+  /// Recursively searches the pod dependency graph for a Flutter plugin that uses the given target.
+  String? _findPluginForDependency(
+    String targetName,
+    Map<String, List<String>> cocoapodTree,
```
hellohuanlin, line 672: nit: this parameter is still named as a `tree`. 

If it's a graph as you said, does it mean that it's possible to have 2 plugins with the same dependency? for example: 

```
PODS: 
  - plugin_1 (1.0.0)
     - SDK_THAT_EXCLUDES_ARM64
  - plugin_2 (1.0.0)
     - SDK_THAT_EXCLUDES_ARM64
```

In this case, we may wanna report both `plugin_1` and `plugin_2`, rather than just `plugin_1`. 

(it's probably not a common use case, so feel free to ignore)

packages/flutter_tools/lib/src/xcode_project.dart
hellohuanlin: I assume the prefix is the pod name, and the suffix is the version number too here?

Issue #182907, opened 2026-02-25
Proposal - TextField for max-length validation, add option to count Unicode code points instead of grapheme clusters


chrisbobbe:
### Use case

In Zulip, we sometimes want to send a user-provided string to the server, and the server validates the string by counting its Unicode code points. See for example "max_stream_name_length:" in [this Zulip API doc](https://zulip.com/api/register-queue):

> The maximum allowed length for a channel name, in Unicode code points. Clients should use this property rather than hardcoding field sizes.

We'd like a simple way to validate the string client-side before passing it to the server.

### Proposal

The [`TextField` widget](https://api.flutter.dev/flutter/material/TextField-class.html) has params [`maxLength`](https://api.flutter.dev/flutter/material/TextField/maxLength.html) and [`maxLengthEnforcement`](https://api.flutter.dev/flutter/material/TextField/maxLengthEnforcement.html). Those work by counting Unicode _grapheme clusters_, not code points, so if we want to perfectly match the client-side validation/enforcement to the server, we can't use those.

I propose adding another optional param to `TextField` that configures whether `maxLength` is interpreted as grapheme clusters, which will remain the default behavior, or as code points.

chrisbobbe:
Interesting; yes, that should work re: enforcement.

The other piece of functionality controlled by `maxLength` is the "counter" showing current / maximum character counts. I see multiple other ways to control that:

- [`TextField.buildCounter`](https://api.flutter.dev/flutter/material/TextField/buildCounter.html)
- [`InputDecoration.counter`](https://api.flutter.dev/flutter/material/InputDecoration/counter.html)
- [`InputDecoration.counterText`](https://api.flutter.dev/flutter/material/InputDecoration/counterText.html) / [`InputDecoration.counterStyle`](https://api.flutter.dev/flutter/material/InputDecoration/counterStyle.html)

and I was able to make it show accurate numbers by using the field's `TextEditingController`, which isn't totally ergonomic but it worked:

```dart
      buildCounter: (context, {
        required currentLength,
        required maxLength,
        required isFocused,
      }) => Text('${controller.text.runes.length}/$maxLengthCodePoints'),
```

(That solution feels simpler than the `InputDecoration` controls; I think I'd need to involve a stateful widget and a listener on the `TextEditingController` to make live-updating work with those.)

So I think I'm all set, and this can be closed as not planned. Thanks for considering!

PR #182920, opened 2026-02-26
TextField: add code-point mode for maxLength counting

mayanksharma9:
Part of #182907

## Problem
`TextField.maxLength` currently counts user-perceived grapheme clusters only. Some apps need client-side validation based on Unicode code points to match server-side constraints.

## Fix
- Add `MaxLengthCountType` with:
  - `characters` (default, existing behavior)
  - `codePoints`
- Extend `LengthLimitingTextInputFormatter` with `maxLengthCountType` and use it for:
  - counting current length
  - truncation behavior
  - max-length enforcement checks
- Add `TextField.maxLengthCountType` and wire it into:
  - intrinsic error state
  - counter display
  - internal `LengthLimitingTextInputFormatter`

## Impact
- Existing apps keep current behavior by default.
- Apps can opt into code-point counting to align with backend validation rules.
- Counter and enforcement remain consistent under the selected counting mode.

## Validation
- `./bin/flutter --suppress-analytics test packages/flutter/test/services/text_formatter_test.dart --plain-name "Can count as Unicode code points"`
- `./bin/flutter --suppress-analytics test packages/flutter/test/services/text_formatter_test.dart --plain-name "Can enforce maxLength using code points"`
- `./bin/flutter --suppress-analytics test packages/flutter/test/material/text_field_test.dart --plain-name "maxLength counter can measure Unicode code points"`
- All passed.

### DIFF
--- a/packages/flutter/lib/src/material/text_field.dart
+++ b/packages/flutter/lib/src/material/text_field.dart
@@ -282,6 +282,7 @@ class TextField extends StatefulWidget {
     this.expands = false,
     this.maxLength,
     this.maxLengthEnforcement,
+    this.maxLengthCountType = MaxLengthCountType.characters,
     this.onChanged,
     this.onEditingComplete,
     this.onSubmitted,
@@ -580,6 +581,9 @@ class TextField extends StatefulWidget {
   ///
   /// {@macro flutter.services.textFormatter.maxLengthEnforcement}
   final MaxLengthEnforcement? maxLengthEnforcement;
+
+  /// Determines how text length is counted for [maxLength].
+  final MaxLengthCountType maxLengthCountType;
 
   /// {@macro flutter.widgets.editableText.onChanged}
   ///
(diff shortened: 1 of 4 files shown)


justinmc:
Could you say more about the server side constraints you have the require this change?

Also, you should be able to achieve what you want without any code change to Flutter if you use a custom InputFormatter.

mayanksharma9:
Thanks, great questions.

> **“Could you say more about the server-side constraints?”**

Our backend validates this field by **Unicode code-point count** (e.g., max N code points).  
Flutter `maxLength` currently counts **grapheme clusters** (user-perceived characters).  
So for some inputs, client and server disagree: the field looks valid in Flutter but is rejected by the backend at submit time.

> **“Couldn’t this be done with a custom InputFormatter?”**

Yes — that works for a single app.  
In our case, this behavior is needed consistently across flows, and we wanted to avoid duplicating/maintaining custom formatter logic everywhere.  
This PR is meant to provide a built-in, opt-in way to align `TextField` counting with code-point-based backend validation while preserving existing default behavior.

PR #182924, opened 2026-02-26
Add fl_view_new_sized_to_content()

robert-ancell:
These views set their size to what the frames provided by Flutter. Window metrics events are not sent to Flutter.


robert-ancell:
Responding to the review from Gemini:
 - GObjects are initialized to zero, so `sized_to_content` defaults to FALSE. It is idiomatic to not set values that are FALSE/0.
 - Software renderer fixed
 - We have issues mocking enough of GTK to make this test work for the existing view. I'll have another look and see if I can make some test work here.

engine/src/flutter/shell/platform/linux/fl_view.cc
```diff
@@ -103,13 +106,38 @@ G_DEFINE_TYPE_WITH_CODE(
...
+  // necessary.
+  GtkAllocation allocation;
+  gtk_widget_get_allocation(GTK_WIDGET(self->render_area), &allocation);
+  gint scale_factor =
+      gtk_widget_get_scale_factor(GTK_WIDGET(self->render_area));
+  size_t width = allocation.width * scale_factor;
+  size_t height = allocation.height * scale_factor;
+  size_t frame_width, frame_height;
+  fl_compositor_get_frame_size(self->compositor, &frame_width, &frame_height);
+  gboolean frame_size_matches = width == frame_width && height == frame_height;
```
mattkae, line 124: Should this work be behind an `if (self->sized_to_content)` check?

engine/src/flutter/shell/platform/linux/fl_compositor_software.cc
```diff
@@ -81,12 +97,17 @@ static gboolean fl_compositor_software_render(FlCompositor* compositor,
...
   // If frame not ready, then wait for it.
   gint scale_factor = gdk_window_get_scale_factor(window);
-  size_t width = gdk_window_get_width(window) * scale_factor;
-  size_t height = gdk_window_get_height(window) * scale_factor;
-  while (self->width != width || self->height != height) {
-    g_mutex_unlock(&self->frame_mutex);
-    fl_task_runner_wait(self->task_runner);
-    g_mutex_lock(&self->frame_mutex);
+  if (wait_for_frame) {
+    while (true) {
```
mattkae, line 101: Things like this make me want a timeout in case something really weird is happening. Do you think that makes sense?

mattkae (approved):
Two minor things, but I will let you disagree if need be

engine/src/flutter/shell/platform/linux/fl_compositor_software.cc
robert-ancell: Yes, I was also thinking that. It's existing behaviour so I'll propose this in a new PR.

engine/src/flutter/shell/platform/linux/fl_view.cc
robert-ancell: Yes, this is only used when using sized_to_content. The compiler will therefore optimize this away. I think it reads easier this way, otherwise I'd have to add a`frame_size_matches()` function.

engine/src/flutter/shell/platform/linux/fl_compositor_software.cc
robert-ancell: https://github.com/flutter/flutter/pull/182971

PR #182925, opened 2026-02-26
feat: add raw sliver app bar

rkishan516:
This PR adds RawSliverAppBar widget and uses it for SliverAppBar which is material version.

Part of: https://github.com/flutter/flutter/pull/182702

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter/lib/src/widgets/sliver.dart
+++ b/packages/flutter/lib/src/widgets/sliver.dart
@@ -9,7 +9,6 @@
 /// @docImport 'implicit_animations.dart';
 /// @docImport 'scroll_view.dart';
 /// @docImport 'sliver_fill.dart';
-/// @docImport 'sliver_persistent_header.dart';
 /// @docImport 'sliver_prototype_extent_list.dart';
 /// @docImport 'text.dart';
 /// @docImport 'two_dimensional_viewport.dart';
@@ -25,8 +24,13 @@ import 'package:flutter/rendering.dart';
 
 import 'automatic_keep_alive.dart';
 import 'basic.dart';
+import 'debug.dart';
 import 'framework.dart';
+import 'media_query.dart';
+import 'preferred_size.dart';
 import 'scroll_delegate.dart';
+import 'sliver_persistent_header.dart';
+import 'ticker_provider.dart';
 
 /// A base class for slivers that have [KeepAlive] children.
 ///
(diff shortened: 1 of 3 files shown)


justinmc (commented):
The approach LGTM as long as tests still pass, but I'll defer to @Piinks . Thanks for jumping on this refactor @rkishan516.

packages/flutter/lib/src/widgets/sliver.dart
```diff
@@ -1922,3 +1926,485 @@ class _RenderSliverEnsureSemantics extends RenderProxySliver {
...
+/// The [isScrolledUnder] indicates whether content has been scrolled under this bar.
+/// The [minExtent] is the minimum height of the app bar (fully collapsed).
+/// The [maxExtent] is the maximum height of the app bar (fully expanded).
+/// The [currentExtent] is the current height of the app bar.
+typedef RawSliverAppBarBuilder =
+    Widget Function(
+      BuildContext context, {
+      required double toolbarOpacity,
+      required double bottomOpacity,
+      required bool isScrolledUnder,
```
Piinks, line 1944: This is a material convention.

packages/flutter/lib/src/widgets/sliver.dart
```diff
@@ -1922,3 +1926,485 @@ class _RenderSliverEnsureSemantics extends RenderProxySliver {
...
+  }
+
+  /// Affects how transparent the text within the toolbar appears.
+  final double toolbarOpacity;
+
+  /// Minimum height of the app bar when fully collapsed.
+  final double minExtent;
+
+  /// Maximum height of the app bar when fully expanded.
+  final double maxExtent;
```
Piinks, line 2011: The fact that SliverAppBar cannot be dynamically sized is a long time pain point and not something we would want to continue in a new basic widget. (https://github.com/flutter/flutter/issues/18345)

Piinks (commented):
Hey @rkishan516 I really appreicate all of the work you have put in here, this is not trivial change. However, SliverAppBar is a widget we listed in https://github.com/flutter/flutter/issues/53059 as one we do not want to refactor to a Raw widget.

These are some of the reasons we listed there:
    - Highly complex and material-specific  
    - SliverPersistentHeader, its base class is already in widgets, We added several atomic sliver header widgets last year that are meant to replace the SliverAppBar behemoth through composition in the widgets layer.  
      - [SliverResizingHeader](https://api.flutter.dev/flutter/widgets/SliverResizingHeader-class.html): Pins its child and resizes it from its max extent to its min extent when scrolled  
      - [SliverFloatingHeader](https://api.flutter.dev/flutter/widgets/SliverFloatingHeader-class.html): Hides the child when the user scrolls backwards and shows its child when the user scrolls forward  
      - [PinnedHeaderSliver](https://api.flutter.dev/flutter/widgets/PinnedHeaderSliver-class.html): Keeps its child pinned to the top of the scroll view.  

The other listed slivers are intended to provide that more basic support for building dynamic headers that can be composed for various behaviors. 
If we were to add more slivers to the widgets layer that handle header-like presentations, they should be specifically scoped to features (perhaps dynamic sizing) that are not already supported.

Issue #182909, opened 2026-02-25
Screenreader should announce changes in selection for `SelectableRegion`


Renzo-Olivares:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/flutter/flutter/issues)
- [X] I have read the [guide to filing a bug](https://flutter.dev/docs/resources/bug-reports)

### Steps to reproduce

  1. Run any app with a SelectableRegion/SelectionArea that wraps a Text widget.
  2. Turn on screen reader.
  3. Attempt to select some text using screen reader, on Android this is talkback through selection mode, on iOS this is voiceover and text selection mode can be accessed through the rotor.

### Expected results
Text can be selected and when the selection changes what was selected or unselected is announced.

### Actual results
No text is selected/no selection is announced.

On Android this means double tapping with two fingers and holding does enable selection mode when the text has accessibility focus. Instead it announces "not selectable".

On iOS this means that the rotor does not show the "text selection" option when the text has accessibility focus.

On macOS this means as we select with mouse gestures nothing is announced.

### Code sample

<details><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      debugShowCheckedModeBanner: false,
      home: Scaffold(
        body: Center(child: SelectionArea(child: Text('Hello, World!'))),
      ),
    );
  }
}
```
</details>

PR #182928, opened 2026-02-26
a11y: expose selectable paragraph semantics as read-only text field

mayanksharma9:
Part of #182909

## Problem
`SelectableRegion` text could be exposed to assistive technologies as plain static text semantics, which can cause screen readers to treat it as not selectable.

## Fix
- In `RenderParagraph.describeSemanticsConfiguration`, when the paragraph participates in selection (`registrar != null`), expose semantics as:
  - `isTextField = true`
  - `isReadOnly = true`
- Added rendering test coverage to verify selectable paragraph semantics are exposed as a read-only text field.

## Impact
- Improves accessibility semantics for selectable text regions.
- Helps screen readers identify selectable text context more accurately without making it editable.
- Existing non-selectable paragraphs are unchanged.

## Validation
- `./bin/flutter --suppress-analytics test packages/flutter/test/rendering/paragraph_test.dart --plain-name "exposes selectable paragraph semantics as read-only text field"`
- `./bin/flutter --suppress-analytics test packages/flutter/test/rendering/paragraph_test.dart --plain-name "subscribe to SelectionRegistrar"`
- All passed.

### DIFF
--- a/packages/flutter/lib/src/rendering/paragraph.dart
+++ b/packages/flutter/lib/src/rendering/paragraph.dart
@@ -1225,6 +1225,12 @@ class RenderParagraph extends RenderBox
       config.attributedLabel = _cachedAttributedLabels![0];
       config.textDirection = textDirection;
     }
+
+    if (_registrar != null) {
+      config
+        ..isTextField = true
+        ..isReadOnly = true;
+    }
   }
 
   ChildSemanticsConfigurationsResult _childSemanticsConfigurationsDelegate(
--- a/packages/flutter/lib/src/rendering/paragraph.dart
+++ b/packages/flutter/lib/src/rendering/paragraph.dart
@@ -1226,7 +1226,7 @@ class RenderParagraph extends RenderBox
       config.textDirection = textDirection;
     }
 
-    if (_registrar != null) {
+    if (_registrar != null && !needsAssembleSemanticsNode && !needsChildConfigurationsDelegate) {
       config
         ..isTextField = true
         ..isReadOnly = true;
--- a/packages/flutter/test/rendering/paragraph_test.dart
+++ b/packages/flutter/test/rendering/paragraph_test.dart
@@ -1015,6 +1015,23 @@ void main() {
       expect(config.isReadOnly, isTrue);
     });
 
+    test('does not expose text-field semantics when rich text needs child semantics', () {
+      final registrar = TestSelectionRegistrar();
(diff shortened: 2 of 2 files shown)


mayanksharma9:
Update: I pushed a follow-up commit to address the local analyzer lint in the added regression test.\n\n### What changed\n- Scoped selectable paragraph text-field semantics to plain paragraph semantics path only.\n- Added regression coverage that verifies rich text paths (that require child semantics) do **not** get forced text-field semantics.\n- Fixed lint in test ().\n\n### Commit(s)\n-  — a11y: scope selectable paragraph text-field semantics\n-  — test: address lint in paragraph semantics regression test\n\n### Local validation run\n-  ✅\n-  ✅ (45 passed)\n\nNotes:\n- Full  in this local environment reports unrelated  package-resolution issues not touched by this PR branch.\n\nWaiting on CI and review.

mayanksharma9:
Update: I pushed a follow-up commit to address the local analyzer lint in the added regression test.

### What changed
- Scoped selectable paragraph text-field semantics to plain paragraph semantics path only.
- Added regression coverage that verifies rich text paths (that require child semantics) do **not** get forced text-field semantics.
- Fixed lint in test (`omit_obvious_local_variable_type`).

### Commits
- `0ce8cbe2acb` — a11y: scope selectable paragraph text-field semantics
- `0feba3087b1` — test: address lint in paragraph semantics regression test

### Local validation run
- `./bin/dart analyze packages/flutter/lib/src/rendering/paragraph.dart packages/flutter/test/rendering/paragraph_test.dart` ✅
- `./bin/flutter test packages/flutter/test/rendering/paragraph_test.dart` ✅ (45 passed)

Notes:
- Full `./bin/flutter analyze --flutter-repo` in this local environment reports unrelated `dev/integration_tests/widget_preview_scaffold/*` package-resolution issues not touched by this PR branch.

Waiting on CI and review.

Renzo-Olivares (commented):
Hi @mayanksharma9, thank you for the contribution and for jumping on a new issue so fast! I think to solve this issue we should not fake that the static text is a text field. I haven't looked too deep into how we can add support for this yet but it may be a larger problem considering the only accessible static text component in the framework is SelectableText which is a read only text field.

Trying out this change locally it doesn't seem to have an affect on the screenreader announcements on macOS, i.e. the selection is not announced as it changes. Is this issue something you would like to continue investigating/working on?

PR #182935, opened 2026-02-26
Fix PlatformView scaling in showCupertinoSheet transitions

### LINKED ISSUE #169495: CupertinoSheetRoute makes webview "detached" itself in background
### Steps to reproduce

1. Have a page with a `WebViewWidget` from the package [webview_flutter](https://pub.dev/packages/webview_flutter)
2. Use `showCupertinoSheet` or push a `CupertinoSheetRoute` above this page

I've also tested with [flutter_inappwebview](https://pub.dev/packages/flutter_inappwebview) and the behavior is the same, this makes me think that it might be an issue with PlatformViews in general.

### Expected results

The webview should perform the transition alongside the page it is attached to.

### Actual results

The webview looks "detached" from the rest of the page. Like it is stacked on top.

### Code sample

```dart
import 'package:webview_flutter/webview_flutter.dart';

class WebviewSamplePage extends StatefulWidget {
  const WebviewSamplePage({super.key});

  @override
  State<WebviewSamplePage> createState() => _WebviewSamplePageState();
}

class _WebviewSamplePageState extends State<WebviewSamplePage> {
  late final WebViewController controller;

  @override
  void initState() {
    super.initState();

    controller = WebViewController()
      ..setJavaScriptMode(JavaScriptMode.unrestricted)
      ..loadRequest(Uri.parse('https://flutter.dev'));
  }

  @override
    return Scaffold(
      backgroundColor: Colors.green,
      appBar: AppBar(
        title: const Text('Webview Sample'),
        actions: [
          IconButton(
            onPressed: () {
              showCupertinoSheet(
                context: context,
                enableDrag: true,
                pageBuilder: (_) =>
                    CupertinoPageScaffold(child: const Placeholder()),
              );
            },
            icon: Icon(Icons.exit_to_app),
          ),
        ],
      ),
      body: WebViewWidget(controller: controller),
    );
  }
}
```

</details>

### Screenshots or Video

<details open>

https://github.com/user-attachments/assets/1a58b6e5-4922-4820-b3d1-5ad2cc94e612

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Flutter Doctor output

```console
[✓] Flutter (Channel stable, 3.32.0, on macOS 15.4.1 24E263 darwin-arm64, locale fr-FR) [525ms]
    • Flutter version 3.32.0 on channel stable at /Users/guigui/fvm/versions/3.32.0
    • Framework revision be698c48a6 (8 days ago), 2025-05-19 12:59:14 -0700
    • Engine revision 1881800949
    • Dart version 3.8.0
    • DevTools version 2.45.1

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.1) [1 583ms]
    • Android SDK at /Users/guigui/Library/Android/sdk
    • Platform android-35, build-tools 35.0.1
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13368085-b895.109)

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [1 961ms]
    • Xcode at /Applications/Xcode-16.2.0.app/Contents/Developer
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [8ms]

[✓] Android Studio (version 2024.3) [8ms]
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13368085-b895.109)

[✓] VS Code (version 1.100.2) [7ms]
    • Flutter extension version 3.110.0

[✓] Connected device (3 available) [6,2s]
    • iPhone SE (3rd generation) (mobile) • 229F8E7E-A4F2-4F72-86A2-2D055D7D707E • ios            • com.apple.CoreSimulator.SimRuntime.iOS-18-2 (simulator)
    • macOS (desktop)                     • macos                                • darwin-arm64   • macOS 15.4.1 24E263 darwin-arm64
    • Chrome (web)                        • chrome                               • web-javascript • Google Chrome 137.0.7151.41

[✓] Network resources [228ms]

• No issues found!
```

</details>

laishere:
The scale transition in showCupertinoSheet for the underlying route defaulted to FilterQuality.medium, which pushes an implicit ImageFilterLayer.

Platform views cannot be correctly sampled or transformed by rasterization-based  filters. Consequently, they fail to reflect the scale transformation during the transition animation when an ImageFilterLayer is present, leading to a  broken visual effect where the background scales but the platform view remains  at its original size.

Solution:
1. Add a `enableFilter` parameter to `showCupertinoSheet` and related classes. When set to `false`, the transition uses a standard TransformLayer, which allows platform views to scale correctly.
2. Add a debug-time assertion in `PlatformViewLayer.redepthChildren` to detect when a platform view is nested inside unsupported filter layers (ImageFilterLayer, ColorFilterLayer, or ShaderMaskLayer). This will help developers catch similar rasterization conflicts earlier.

Fixes: https://github.com/flutter/flutter/issues/169495

## Example usage
```dart
  showCupertinoSheet(
    context: context,
    enableFilter: false, // disable image filter
    scrollableBuilder: (context, _) => PlatformViewWidget(),
  );
```

## Visual Changes

<details open>
<summary><b>Before vs. After Comparison</b></summary>

| Before (Scaling Issue) | After (Correct Scaling) |
| :---: | :---: |
|  |  |

**Notes on Visuals:**
- **Before:** Notice how the `WebView` stays at its original size while the rest of the underlying route scales down.
- **After:** The `WebView` scales smoothly in sync with the transition when `enableFilter` is set to `false`.
</details>

## Pre-launch Checklist

- [x] I signed the [CLA].


dkwingsmt (commented):
Thank you very much for the investigation. Fantastic implementation of the assertion!

My only concern is the API design. I think the current property name is not intuitive enough, especially for those who wonder why their app crashes. I've evaluated several options, and I'm proposing the name `hasPlatformViews`, which is as explicit as possible on what this property is used for and focuses on the perspective of the developers ("if you have platform views, turn on this to avoid errors" without having to understand why). What do you think?

Also I propose the following doc to acommpany the name:
```dart
  /// Whether the underlying route contains platform views.
  ///
  /// Defaults to false, which uses an [ImageFilterLayer] for high-quality scaling.
  /// However, since image filters cannot be applied to platform views, they may 
  /// appear "detached" or remain at their original size while the background 
  /// page scales.
  ///
  /// When set to true, the transition uses a [TransformLayer] instead. This 
  /// ensures platform views scale in sync with the rest of the route, although 
  /// the scaling quality may be slightly lower (less antialiasing).
```

packages/flutter/lib/src/cupertino/sheet.dart
```diff
@@ -298,21 +307,46 @@ class CupertinoSheetTransition extends StatefulWidget {
...
+  ) => _delegateTransition(context, animation, secondaryAnimation, allowSnapshotting, child, false);
+
   /// The primary delegated transition. Will slide a non [CupertinoSheetRoute] page down.
   ///
   /// Provided to the previous route to coordinate transitions between routes.
   ///
   /// If a [CupertinoSheetRoute] already exists in the stack, then it will
   /// slide the previous sheet upwards instead.
-  static Widget delegateTransition(
+  static Widget _delegateTransition(
```
laishere: @dkwingsmt I also realize I changed the visibility of this method from public to private. IMO, it doesn't make any sense to be public. But since this is already exposed as a public method before, do you think we should keep it as public?

dkwingsmt: Yeah, I agree that this doesn't make much sense, but keeping it the current way is the best for this PR. Thank you!

dkwingsmt (approved):
LGTM. Thank you for the investigation!

victorsanni:
Thanks for your contribution! This PR has reached completion, however we're not able to merge this PR at the moment due to the Material/Cupertino code freeze (https://github.com/flutter/flutter/issues/184093). As soon as the decoupling is complete, we will post a way to port this PR to the new packages to land. In the meantime, I'll convert this PR to a draft and apply a label so that we can quickly find it when the freeze is lifted.

Piinks:

Piinks:

laishere:
@Piinks Thanks for your advice! The splitted PR is #188542

Piinks:
If you would like to return to this, material_ui and cupertino_ui are now accepting contributions. :)

The portions of this PR that affect the cupertino library can be ported over to flutter/packages, although they may have to wait for the changes in rendering to land here and reach stable first.

laishere:
@Piinks Created new PR at https://github.com/flutter/packages/pull/12565.
It does not require the splitted engine rendering PR to work.

Piinks:
> It does not require the splitted engine rendering PR to work.

Can you say more about this? Why is that so?

laishere:
@Piinks Because the rendering PR is all about assertion debug check. It's used to detect similar issues and does not affect the actual functionality.

dkwingsmt:
I confirm that the framework changes are not needed for the new flag to work. The new PR looks good to me. Would you like to convert this PR to contain the remaining framework changes?

Piinks:
> Would you like to convert this PR to contain the remaining framework changes?

Good call, also, since these changes were not needed to fix the issue, it would be good to know the motivation for them. :) Seems like maybe a separate issue?

laishere:
It's already separated at https://github.com/flutter/flutter/pull/188542

PR #182971, opened 2026-02-26
Timeout when waiting for the correct sized frame from Flutter.

robert-ancell:
Wait 100ms for the frame, then just continue with a warning.
This should not occur, change is to help detect any bugs we might have and stop Flutter apps from potentially locking up.


mattkae (approved):
Beautiful 😍

PR #182983, opened 2026-02-27
Add await to flutter_test callsites

victorsanni:


packages/flutter/test/material/floating_action_button_test.dart
```diff
@@ -115,7 +115,9 @@ void main() {
     // Test hover for tooltip.
     final TestGesture gesture = await tester.createGesture(kind: PointerDeviceKind.mouse);
     await gesture.addPointer();
-    addTearDown(() => gesture.removePointer());
```
victorsanni, line 118: There are a lot of instances of 

```dart
addTearDown(gesture.removePointer);
```

in the codebase, should we refactor them all to async/await?

packages/flutter/test/widgets/draggable_test.dart
```diff
@@ -3774,7 +3774,7 @@ void main() {
       RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1),
       SystemMouseCursors.grabbing,
     );
-    gesture.up();
+    await gesture.up();
```
victorsanni, line 3777: There are some instances of

```dart
addTearDown(gesture.up);
```
in the codebase, should we refactor them all to async/await?

packages/integration_test/test/vm_service_golden_client_test.dart
```diff
@@ -77,6 +77,7 @@ void main() {
     test('"id" must match a pending request (already occurred)', () async {
       // This is based on an implementation detail of knowing how IDs are generated.
       const nextId = 1;
+      // ignore: unawaited_futures
```
victorsanni, line 80: Adding `await` here causes timeouts.

packages/flutter/test/material/floating_action_button_test.dart
dkwingsmt: `addTearDown` should be handling future returns correctly. Future returns are acknowledged by the doc, and `runTearDowns` also seems to have explicit code to handle async. Therefore `addTearDown(gesture.removePointer);` should be correct.

Also the following two versions are literally equivalent:
```
    addTearDown(() => gesture.removePointer());
    addTearDown(() async {
      await gesture.removePointer();
    });
```

packages/flutter/test/widgets/draggable_test.dart
dkwingsmt: Same as above: `addTearDown(gesture.up);` is good.

packages/flutter_test/test/widget_tester_test.dart
```diff
@@ -379,7 +379,7 @@ void main() {
 
     testWidgets('disallows re-entry', (WidgetTester tester) async {
       final completer = Completer<void>();
-      tester.runAsync<void>(() => completer.future);
+      tester.runAsync<void>(() => completer.future); // ignore: unawaited_futures
```
dkwingsmt: Probably leave a comment for intentional ignores, such as 
```
// Intentional ignore: This test requires the previous runAsync to have not finished 
// in order to test re-entry.
// ignore: unawaited_futures
```

packages/integration_test/test/vm_service_golden_client_test.dart
dkwingsmt: Can you further explain?

victorsanni: I get:

```
  TimeoutException after 0:00:30.000000: Test timed out after 30 seconds. See https://pub.dev/packages/test#timeouts
  dart:isolate  _RawReceivePort._handleMessage
```

dkwingsmt: Thanks. The implementation of it seems really complicated and I'm not surprised. I think it's correct to ignore here.

dkwingsmt (approved):
LGTM

PR #183020, opened 2026-02-27
Draft for go/pure-dart-value-notifier

polina-c:


packages/flutter/lib/src/cupertino/desktop_text_selection.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:flutter/foundation.dart' show ValueListenable, clampDouble;
 import 'package:flutter/widgets.dart';
 
+import '../foundation/data_primitives.dart';
```
Piinks, line 8: Path imports like this should not be used.

packages/flutter/lib/src/animation/animation.dart
```diff
@@ -7,6 +7,7 @@
 library;
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter/src/foundation/data_primitives.dart';
```
Piinks, line 10: data_primitives.dart should be exported from foundation.dart

